PR #5, opened 2018-09-06
fix: Update duration spec to account for end of month

nathanielc:


docs/SPEC.md
```diff
@@ -200,29 +200,35 @@ Examples:
...
 
 Durations can be added to date times to produce a new date time.
-Adding and subtracting durations to date times normalizes the dates.
-For example July 32 converts to August 1.
 
 Addition and subtraction of durations to date times do not commute and are left associative.
-Addition and subtraction of durations to date times applies months, days and seconds in that order and then normalizes.
+Addition and subtraction of durations to date times applies months, days and seconds in that order.
+When months are added to a date times and the resulting date is past the end of the month, the day is rolled back to the last day of the month.
+When months are added to a date times and the starting date is the last day of the month, the day is rolled forward or backward to the last day of the month.
```
jsternberg: Is this the same behavior excel has? This in particular may be surprising behavior to some and may violate the principle of least surprise.

One thing that may be useful is to have a `last_day_of_week()` function and have it return 31 or 32 as its value. Then tell people to use that if they want to specify the last day of the month.

nathanielc: I had a similar concerns over this being surprising. Can you give an example of how I could use the `last_day_of_month` function?

jsternberg: I'm not sure because I don't know how we're going to deal with the call site very well. Maybe it would be one of the time based functions we're thinking of. Two examples:

```
midnight(now()) // get midnight of the current day for the selected time zone
last_day_of_month(now()) // get the last day of the month for the current month
range(start: last_day_of_month(now()-1m))
```

pauldix: So is the issue that this might be surprising because `2018-02-28 + 1m` would become `2018-03-31` while `2018-01-31 + 1m` would become `2018-02-28`?

jsternberg: Yes. I think it would be better to internally store the date as "February 31" and have that materialize as the last day of February (which would usually be February 28) and then I think this handles iteration better too.

pauldix: Yeah, I don't think the internal representation matters (at least it shouldn't for the spec). What's important is the materialized behavior. So I'd expect `2018-01-31 + 2m` to materialize as `2018-03-31`.

nathanielc: > When months are added to a date times and the starting date is the last day of the month, the day is rolled forward or backward to the last day of the month.

The reason for this is precisely for iteration. Look at this example:

```
    intervals(every:1mo, period: 1mo)(start:2018-01-31, stop: 2018-05-31)
    // [2018-01-31, 2018-02-28)
    // [2018-02-28, 2018-03-31)
    // [2018-03-31, 2018-04-30)
    // [2018-04-30, 2018-05-31)

```

Without the last day of the month rule the results would look like this:
```
    intervals(every:1mo, period: 1mo)(start:2018-01-31, stop: 2018-05-31)
    // [2018-01-31, 2018-02-28)
    // [2018-02-28, 2018-03-28)
    // [2018-03-28, 2018-04-28)
    // [2018-04-28, 2018-05-28)

```

The invariant I am trying to preserve is that adding durations behaves the same given a static date as well as during iteration. Specifically I should be able to do this in Flux and get the same result:

```

start = 2018-01-31

first = start + 1mo // 2018-02-28
second = first + 1mo // 2018-03-31
third = second + 1mo // 2018-04-30
forth = third + 1mo // 2018-05-31
```

nathanielc: In contrast though we could define iteration as being index based, so the above example could look like this without the last day of the month rule.


```

start = 2018-01-31

first = start + 1*1mo // 2018-02-28
second = start + 2*1mo // 2018-03-31
third = start + 3*1mo // 2018-04-30
forth = start + 4*1mo // 2018-05-31
```

docmerlin: How would I say "do this on the 30th of the next month" if the current action is on the 30th and the current month is 30 days long and the next month is 31 days long?

nathanielc: @docmerlin Describing specific dates can be accomplished using some date helper methods: For example we could have a `setDayOfMonth` function and use it like this to specify the 30th of the next month.

```
next30th = (date) => setDayOfMonth(date:date+1mo,day:30)
```

Now if you want to describe an interval that begins and ends on the 30th of each month that would be impossible with the last day of the month rule.

If we remove the last day of the month rule, then describing an interval that begins and ends on the last day of the month would be impossible, using pure duration addition.

So its a trade off of which use case we want to support. I am leaning towards removing the last day of the month rule because it doesn't solve all use cases and violates the principle of least surprise.

Then you could write a custom function to handle the intervals of last day of the month like this: (using some made up syntax around loop iteration and generators)


```
monthEndToEnd = (start,stop) => {
    i = 0 
    for { 
        begin = setDayOfMonth(date:start + i * 1mo, day:31) // always use the last day of the month
        if begin >= stop {
           return
        }
        end = setDayOfMonth(date:begin + 1mo, day:31) // always use the last day of the month
        emit {start: begin, stop: end} // psuedo generator syntax to emit interval
        i++
    }
}
```

By using the `setDayOfMonth` function internally it becomes possible to implement the last day of the month behavior.

Along similar line the interval function could be implemented like this


```
intervals = (every,period,offset) => (start,stop) => {
    i = 0 
    for { 
        begin = start + offset + i * every
        if begin >= stop {
           return
        }
        end = begin + period
        emit {start: begin, stop: end} // psuedo generator syntax to emit interval
        i++
    }
}
```

Seeing the contrast between the two was helpful for me. Thoughts?

docmerlin: 👍 
I like the idea of providing them a way to do a last day of month (via the setdayofmonth), without it being magic.

docs/SPEC.md
```diff
@@ -200,29 +200,34 @@ Examples:
...
+When months are added to a date times and the resulting date is past the end of the month, the day is rolled back to the last day of the month.
+
 
 Examples:
 
     2018-01-01T00:00:00Z + 1d       // 2018-01-02T00:00:00Z
     2018-01-01T00:00:00Z + 1mo      // 2018-02-01T00:00:00Z
-    2018-01-01T00:00:00Z + 2mo30d   // 2018-03-30T00:00:00Z
-    2018-01-01T00:00:00Z + 1mo30d   // 2018-03-02T00:00:00Z, February 30th is normalized to March 2 in 2018 since it is not a leap year.
+    2018-01-01T00:00:00Z + 2mo      // 2018-03-01T00:00:00Z
```
jsternberg: Can there be a mention in here about one of the other comments you made? I found it interesting and it may not be immediately obvious the behavior. You had this:

```
first = 2018-01-31T00:00:00Z // January 31
second = first + 1mo // February 31st is rolled back to the last day of the month
third = second + 1mo // The duration keeps track that it was originally January 31 so this is March 31 again instead of March 28

You also say "February 31th" instead of "31st" on line 216.

nathanielc: @jsternberg The above is no longer true, removing the rolling forward rule breaks this. There is no hidden state in dates. If you want that behavior you have to multiply the duration first.

first = 2018-01-31T00:00:00Z // January 31
second = first + 1mo // February 31st is rolled back to the last day of the month Feb 28th
third = first + 2*1mo // March 31st

jsternberg: Sure.

PR #9, opened 2018-09-06
Fix distinct bug when column is outside table key

aanthony1243:
bug was minor.  added plenty of tests to verify correct behavior.


functions/distinct.go
```diff
@@ -223,46 +223,47 @@ func (t *distinctTransformation) Process(id execute.DatasetID, tbl flux.Table) e
 
 	return tbl.Do(func(cr flux.ColReader) error {
 		l := cr.Len()
+		j := execute.ColIdx(t.column, cr.Cols())
```
nathanielc: This lookup can happen outside the Do loop.

PR #11, opened 2018-09-07
fix(csv): include annotations on error messages.

goller:
This came up from chronograf, where annotations were expected and not received in error bodies.


nathanielc:
Closing in favor of #32

PR #19, opened 2018-09-11
Type Inference

jlapacik:
This PR adds type inference to the Flux language. 

Review:

- [x] Type Annotation
- [ ] Constraint Generation

Approved:

- [ ] Type Annotation
- [ ] Constraint Generation

### DIFF
--- a/semantic/annotate.go
+++ b/semantic/annotate.go
@@ -13,6 +13,17 @@ func (tv TypeVar) FreeTypeVar() []TypeVar {
 	return []TypeVar{tv}
 }
 
+func (tv TypeVar) String() string {
+	return tv.name
+}
+
+func (tv TypeVar) Equal(sub Substitutable) bool {
+	if tvar, ok := sub.(TypeVar); ok {
+		return tv.name == tvar.name
+	}
+	return false
+}
+
 // TypeVarGenerator generates type variables
 type TypeVarGenerator struct {
 	prefix string
--- a/semantic/annotate.go
+++ b/semantic/annotate.go
@@ -1,0 +1,70 @@
+package semantic
+
+import (
+	"strconv"
+)
+
+// TypeVar is a type variable
+type TypeVar struct {
+	name string
+}
+
+// TypeVarGenerator generates type variables
+type TypeVarGenerator struct {
+	prefix string
+	suffix int
+}
+
+// NewTypeVar returns the a new type variable from a generator
+func (v *TypeVarGenerator) NewTypeVar() TypeVar {
+	name := v.prefix + strconv.Itoa(v.suffix)
+	v.suffix++
+	return TypeVar{name: name}
+}
+
+// TypeAnnotationVisitor visits the nodes in a semantic
+// graph and assigns all sub-expressions type variables.
+type TypeAnnotationVisitor struct {
+	// tenv represents the type environment of a program.
+	// It maps a semantic node to a unique type variable.
+	tenv map[Node]TypeVar
+	// the next type variable to be assigned. The first
+	// type variable is t0. Then t1, t2, t3 and so on.
+	next *TypeVarGenerator
+}
+
+// Visit assigns a new type variable to a semantic node
+func (v *TypeAnnotationVisitor) Visit(node Node) Visitor {
+	if n, ok := node.(Expression); ok {
+		v.tenv[n] = v.next.NewTypeVar()
+		return v
+	}
+	switch n := node.(type) {
+	case *Identifier:
+		v.tenv[n] = v.next.NewTypeVar()
+	case *Property:
+		v.tenv[n] = v.next.NewTypeVar()
+	case *FunctionParam:
+		v.tenv[n] = v.next.NewTypeVar()
+	}
+	return v
+}
+
+// Done is required to implement the Visitor Interface
+func (v *TypeAnnotationVisitor) Done() {}
+
+// NewTypeAnnotationVisitor returns a new TypeAnnotationVisitor
+func NewTypeAnnotationVisitor() *TypeAnnotationVisitor {
+	return &TypeAnnotationVisitor{
+		tenv: make(map[Node]TypeVar, 64),
+		next: &TypeVarGenerator{
+			prefix: "t",
+			suffix: 0,
+		},
+	}
+}
+
+// TypeEnvironment returns the mapping between nodes and type variables
+func (v *TypeAnnotationVisitor) TypeEnvironment() map[Node]TypeVar {
+	return v.tenv
+}
--- a/semantic/annotate.go
+++ b/semantic/annotate.go
@@ -1,12 +1,16 @@
 package semantic
 
-import (
-	"strconv"
-)
+import "strconv"
 
 // TypeVar is a type variable
 type TypeVar struct {
 	name string
+}
+
+// FreeTypeVar returns all the free (not bound) type variables of an expression.
+// In the case of a type variable, it just returns the same type variable.
+func (tv TypeVar) FreeTypeVar() []TypeVar {
+	return []TypeVar{tv}
 }
 
 // TypeVarGenerator generates type variables
@@ -37,14 +41,8 @@ type TypeAnnotationVisitor struct {
 func (v *TypeAnnotationVisitor) Visit(node Node) Visitor {
 	if n, ok := node.(Expression); ok {
 		v.tenv[n] = v.next.NewTypeVar()
-		return v
 	}
-	switch n := node.(type) {
-	case *Identifier:
-		v.tenv[n] = v.next.NewTypeVar()
-	case *Property:
-		v.tenv[n] = v.next.NewTypeVar()
-	case *FunctionParam:
+	if n, ok := node.(*Identifier); ok {
 		v.tenv[n] = v.next.NewTypeVar()
 	}
 	return v
--- a/semantic/constraint.go
+++ b/semantic/constraint.go
@@ -29,7 +29,7 @@ func (a arraySignature) FreeTypeVar() []TypeVar {
 	return a.elementType.FreeTypeVar()
 }
 func (a arraySignature) String() string {
-	return "[ " + a.elementType.String() + " ]"
+	return "[" + a.elementType.String() + "]"
 }
 func (a arraySignature) Equal(sub Substitutable) bool {
 	if arr, ok := sub.(arraySignature); ok {
@@ -54,10 +54,10 @@ func (o objectSignature) String() string {
 	builder := strings.Builder{}
 	builder.WriteString("{ ")
 	for k, v := range o.properties {
-		param := fmt.Sprintf("%s: %s,", k, v.String())
+		param := fmt.Sprintf("%s:%s, ", k, v.String())
 		builder.WriteString(param)
 	}
-	builder.WriteString(" }")
+	builder.WriteString("}")
 	return builder.String()
 }
 func (o objectSignature) Equal(sub Substitutable) bool {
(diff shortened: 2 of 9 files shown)


semantic/type_inference.md
```diff
@@ -0,0 +1,138 @@
...
+
+            x   -> t0
+            1+2 -> t1
+            1   -> t2
+            2   -> t3
+
+2. Constraint generation phase
+
+    Input: Map of expressions to type variables  
+    Output: Map of type variables to type expressions
```
nathanielc: I expected the output to be a set of constraints not a map. Below your example shows it as a set.

semantic/types.go
```diff
@@ -65,6 +65,7 @@ const (
 	Array
 	Object
 	Function
+	TypeVar
```
nathanielc: This seems odd. Why is a type variable a first class type within Flux itself? No value can have a type of TypeVar.

semantic/annotate.go
```diff
@@ -0,0 +1,126 @@
+package semantic
+
+// Annotate annotates the nodes of a semantic graph with type variables
+func Annotate(node Node, venv *VariableEnvironment, tenv *TypeEnvironment) {
+	var tvar Type
+	switch n := node.(type) {
+	// Assign a new type variable for every literal
+	case *BooleanLiteral:
```
nathanielc: The semantic package has helper methods for walking the graph. Are those useful here? If not why not? I'd like to try and use the walk functions for all graph traversal that way only one place needs to be updated when we add new node types.

semantic/types.go
```diff
@@ -113,6 +115,53 @@ func (k Kind) ReturnType() Type {
...
+}{
+	prefix: "t",
+	suffix: 0,
+}
+
+// NewTypeVar constructs a new type variable to be assigned to an expression
+// Type variable names progress as such - t0, t1, t2, t3, ...
+func NewTypeVar() Type {
+	name := typeVarCache.prefix + strconv.Itoa(typeVarCache.suffix)
+	typeVarCache.suffix++
```
nathanielc: This is not memory safe. If multiple Annotation calls are running concurrently then this will race and possibly crash.

 Each Annotation phase should create its own typeVarCache, that way the type vars are all unique. We should not add locking here as there is no need for the vars to be global across different annotation phases.

semantic/env.go
```diff
@@ -0,0 +1,65 @@
+package semantic
+
+// TypeEnvironment maps expressions to their types
+type TypeEnvironment struct {
+	m map[Node]Type
```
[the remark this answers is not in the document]
jlapacik: Identifiers aren't expressions. That's why I kept it Node. Although we have IdentifierExpressions, so maybe there's no need to annotate Identifiers?

semantic/annotate.go
jlapacik: This was due entirely to having to pass different `VariableEnvironment`s to `Annotate` depending on the type of the expression. However I realize that `VariableEnvironment` is an unnecessary data structure, and so by getting rid of it, we can use the `semantic.Visitor` interface. This will definitely reduce the code bloat. I will update.

semantic/types.go
jlapacik: Agreed, will update.

semantic/types.go
jlapacik: I don't have a good answer for this at the moment so this will probably change.

jlapacik:
Type annotation has been updated based on review comments. In particular the annotation method has been updated to use the `semantic.Visitor` interface and the backing data structure for producing new type variables is now encapsulated inside of this visitor. @nathanielc you are welcome to have another look.

semantic/annotate.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// Visit assigns a new type variable to a semantic node
+func (v *TypeAnnotationVisitor) Visit(node Node) Visitor {
+	if n, ok := node.(Expression); ok {
+		v.tenv[n] = v.newTypeVar()
+	}
+	switch n := node.(type) {
+	case *CallExpression:
+		if n.pipe != nil {
```
nathanielc: Instead of this we should update the Visitor to walk the CallExpression.pipe if it exists.

semantic/annotate.go
```diff
@@ -0,0 +1,67 @@
...
+	tenv map[Node]Type
+	// the next type variable to be assigned. The first
+	// type variable is t0. Then t1, t2, t3 and so on.
+	next *typeVar
+}
+
+// Visit assigns a new type variable to a semantic node
+func (v *TypeAnnotationVisitor) Visit(node Node) Visitor {
+	if n, ok := node.(Expression); ok {
+		v.tenv[n] = v.newTypeVar()
```
nathanielc: Should this return here?

semantic/annotate.go
```diff
@@ -0,0 +1,67 @@
...
+		tenv: make(map[Node]Type, 64),
+		next: &typeVar{
+			prefix: "t",
+			suffix: 0,
+		},
+	}
+}
+
+// newTypeVar returns the next type variable to be assigned
+func (v *TypeAnnotationVisitor) newTypeVar() Type {
```
nathanielc: A TypeVar is not a Type. I think we need a new type called `typeVar`. The existing `typeVar` struct should probably be renamed to `typeVarGenerator` since its purpose it to create a new type var.

Then the `typeVar` name is free to become its own type, probably something like this:

```
type typeVarGenerator struct {
	prefix string
	suffix int
}

func (g *typeVarGenerator) Generate() typeVar {
   //...
}
type typeVar struct {
    name string
   // maybe need more here once we move on to unification...
}
```

And then the `tenv map[Node]Type` would become `tenv map[Node]typeVar`.

Thoughts?

semantic/annotate.go
jlapacik: Yes I'll update.

semantic/annotate.go
jlapacik: No, it could be a call expression in which case it needs to perform a pipe check.

semantic/annotate.go
jlapacik: @nathanielc So I'm currently thinking through this. Are you saying remove `TypeVar` and instead create a `typeVar` that implements the `Type` interface (so it is still considered a type) but when you call `Kind()` on a `typeVar` if it doesn't know its concrete type it will just return `nil`?

I agree it doesn't make sense to have `TypeVar` be a primitive type as it is only going to be used internally during type inference. My reason for doing it originally was with the substitution phase in mind. During substitution we will be building up our types. In this phase type variables might be assigned to concrete types like int, bool, etc. But they might also be assigned to not yet complete types (types that are composed of type variables). For example:

    t0 = f(t1, t2) -> int

What is `f(t1, t2) -> int` here? It's not a type variable, but it's also not a function type because its inputs are two type variables. If `typeVar` implements `Type`, then both sides of the above equation are types, and we can then perform substitutions on types.

Or does it not make sense for `typeVar` to implement `Type`?

semantic/annotate.go
jlapacik: Never mind. Yes by changing `semantic.Walk` to also visit the pipe expression if it exists, we may just return after this if statement.

semantic/annotate.go
nathanielc: yeah, I do not think `typeVar` should implement Type. It should have its own methods for reporting what the concrete type is. Maybe just a `Type() Type` method on the `typeVar` struct? So instead of `typeVar` being a type it can tell you its Type.

semantic/annotate.go
```diff
@@ -0,0 +1,69 @@
...
+		next: &tvarGenerator{
+			prefix: "t",
+			suffix: 0,
+		},
+	}
+}
+
+// NewTypeVar returns the next type variable to be assigned
+func (v *TypeAnnotationVisitor) NewTypeVar() TypeVar {
+	name := v.next.prefix + strconv.Itoa(v.next.suffix)
```
nathanielc: nit, can you make this method a method of the typeVarGenerator type instead?

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+func (f funcTypeExpression) FreeTypeVar() []TypeVar {
+	// TODO: Handle duplicates
+	freeTypeVars := make([]TypeVar, 0, 8)
+	for _, v := range f.params {
+		freeTypeVars = append(freeTypeVars, v.FreeTypeVar()...)
+	}
+	return append(freeTypeVars, f.returnType.FreeTypeVar()...)
+}
```
nathanielc: Why can't the `FunctionExpression` itself implement the FreeTypeVar interface?

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+		//     Statement
+		//     Return Statement
+		// }
+		//
+		// -> typeof(f) = (typeof(a), typeof(b)) => typeof(Return Statement)
+		funcType := funcTypeExpression{
+			params:     make(map[string]Substitutable, len(n.Params)),
+			returnType: n.returnTypeVar,
+		}
+		c.cons = append(c.cons, Constraint{
```
nathanielc: This seems like a special case of Variable Declaration.

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+		// --------------------------------
+		// x = (op) expression
+		//
+		// -> typeof(x) = bool
+		// -> typeof(expression) = bool
+		//
+		// Unary expressions yield boolean types
+		c.cons = append(c.cons, Constraint{
+			left:  tv,
+			right: Bool,
```
nathanielc: Unary expressions can apply to integers as well. Negation is a unary operator.

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+				},
+				Constraint{
+					left:  tv,
+					right: c.tenv[n.Right],
+				},
+				Constraint{
+					left:  c.tenv[n.Right],
+					right: c.tenv[n.Left],
+				},
+			)
```
nathanielc: All the above cases look the same. Is that right? If so can you collapse them into a single case? You can comma separate the cases like so:

`case ast.MultiplicationOperator,ast.DivisionOperator,...:`

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+				Constraint{
+					left:  tv,
+					right: Bool,
+				})
+		case ast.NotEqualOperator:
+			c.cons = append(c.cons,
+				Constraint{
+					left:  tv,
+					right: Bool,
+				})
```
nathanielc: Same here. Also should the operands must be the same type too?

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+		case ast.StartsWithOperator:
+		case ast.InOperator:
+		case ast.RegexpMatchOperator:
```
nathanielc: Regex can add a constraint that the right hand side is a Regex.

semantic/constraint.go
```diff
@@ -0,0 +1,334 @@
...
+		case ast.NotRegexpMatchOperator:
```
nathanielc: Same here.

semantic/constraint.go
jlapacik: That is right. I didn't know you could do that.

semantic/constraint.go
jlapacik: According to this, comparison operands may be of different types.

https://github.com/influxdata/flux/blob/4f007663549585e92c885e714da02dcb4a782c4a/semantic/binary_types.go#L35

semantic/constraint.go
jlapacik: Sorry, my comment is misleading. There is no variable declaration as part of this rule. I included a variable declaration for readability. It should read `typeof(function expression) = (typeof(a), typeof(b)) => ...`.

jlapacik:
The last commit https://github.com/influxdata/flux/pull/19/commits/e8e6604e1bd4f2e9a6c9a167b668f7184bf280bc is not stable. It introduces new semantic nodes for function parameters and a modification to the `Walk` function for a semantic graph. These features aren't in a working state and were just experiments.

nathanielc:
Closing in favor of #203 which is just an extension of this PR

PR #21, opened 2018-09-11
Updated aggregates to no longer have a time column.  Many tests updated

aanthony1243:
Some end-to-end tests show how to use duplicate to get old behavior


functions/integral.go
```diff
@@ -14,6 +14,8 @@ const IntegralKind = "integral"
 
 type IntegralOpSpec struct {
 	Unit flux.Duration `json:"unit"`
+	TimeSrc string   `json:"timeSrc"`
+	TimeDst string   `json:"timeDst"`
```
nathanielc: Why are we adding it to the integral function?

aanthony1243: the timeSrc is used for computation in the integral.  We could remove timeDst, I'm not sure if that provides context for the value or not.

aanthony1243: @nathanielc I made an update that I believe addresses your concerns.  We still need to identify the input timeCol because that column is required for computing the integral.  The timeDST is not needed.

nathanielc: OK, that makes sense

jsternberg: Do we have a time destination for derivative? The time is likely an important part of this function, but I'm not sure how we treat derivative which would be similar.

PR #22, opened 2018-09-12
chore: Remove platform dependency

nathanielc:


nathanielc:
The only parts that still depend on platform are the from.go implementation. See https://github.com/influxdata/flux/issues/23 for how to go about that.

PR #32, opened 2018-09-13
fix: decode the error annotation in the csv format

jsternberg:
This fixes the csv decoder so it recognizes and decodes the error
annotation.

Also fixes an error in the multi result decoder because it didn't share
the `csv.Reader` instance across results which caused buffered data to
be discarded.

### DIFF
--- a/csv/result.go
+++ b/csv/result.go
@@ -34,6 +34,7 @@ const (
 
 	resultLabel = "result"
 	tableLabel  = "table"
+	errorLabel  = "error"
 
 	commentPrefix = "#"
 
@@ -72,7 +73,7 @@ type ResultDecoderConfig struct {
 }
 
 func (d *ResultDecoder) Decode(r io.Reader) (flux.Result, error) {
-	return newResultDecoder(r, d.c, nil)
+	return newResultDecoder(newCSVReader(r), d.c, nil)
 }
 
 // MultiResultDecoder reads multiple results from a single csv file.
@@ -94,8 +95,9 @@ func NewMultiResultDecoder(c ResultDecoderConfig) *MultiResultDecoder {
 
 func (d *MultiResultDecoder) Decode(r io.ReadCloser) (flux.ResultIterator, error) {
 	return &resultIterator{
-		c: d.c,
-		r: r,
+		c:  d.c,
+		r:  r,
+		cr: newCSVReader(r),
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


csv/result.go
```diff
@@ -31,6 +31,7 @@ const (
 	datatypeAnnotation = "datatype"
 	groupAnnotation    = "group"
 	defaultAnnotation  = "default"
+	errorAnnotation    = "error"
```
nathanielc: The `error` is not an annotation. It will never appear in the first column of the csv data if annotations are being used.

csv/result.go
```diff
@@ -296,6 +297,18 @@ func readMetadata(r *csv.Reader, c ResultDecoderConfig, extraLine []string) (tab
 		if n != len(line) {
 			return tableMetadata{}, errors.Wrap(csv.ErrFieldCount, "failed to read annotations")
 		}
+		if len(line) > 0 && line[0] == errorAnnotation {
```
nathanielc: This will not work when annotations are enabled, which is expected when decoding.

csv/result_test.go
```diff
@@ -467,6 +468,16 @@ func TestResultDecoder(t *testing.T) {
 				}},
 			},
 		},
+		{
+			name:          "simple error message",
+			encoderConfig: csv.DefaultEncoderConfig(),
+			encoded: toCRLF(`error,reference
```
nathanielc: To decode it is required that the csv be annotated. Can you make these tests use the `annotated` versions of csv encoding. Then it will be clear that `error` is not an annotation but rather a column header.

csv/result.go
jsternberg: I'll rename this. It's not really an annotation as I use it.

csv/result.go
jsternberg: I'm not really sure I understand why this won't work. I also don't see a way to turn off annotations to begin with. Can you give me an example of what you mean so I can add it to the tests?

jsternberg: Found it in the spec.

csv/result_test.go
jsternberg: I've updated things so the error is now encoded with annotations and I decode that now instead.

It seems that the decoder has a `NoHeader` attribute that changes how things work, but it seems intended to mean no header column rather than no annotations so it doesn't appear there's a way to detect that there were no annotations.

csv/result.go
nathanielc: Yes, the decoder needs the annotation in order to completely decode a result. As such the decoder assumes they exist. 

So instead of testing with:

```
error,reference
msg,ref
```

The test needs to be like:

```
#datatype,string,string
#group,false,false
#default,,
,error,reference
,msg,ref
```

I just wrote that by hand so it may have a typo, but that is the general idea.

PR #42, opened 2018-09-19
fix(interpreter): don't panic when function calls unknown function

mark-rushakoff:
This is specifically to handle the case of `() => foo()` where `foo` is
an unknown function. The type of that function is evaluated as part of a
binary expression.


mark-rushakoff:
This probably shouldn't be merged, as #43 includes many more cases that this approach doesn't cover.

nathanielc:
Closing

PR #46, opened 2018-09-20
feat(functions): implementing various new data sources for flux.

aanthony1243:
feat(functions): introduced the connect/fetch/decode interface for writing sources

feat(fluxcli) with generic from support, http server for Chronograph

feat(fluxcli): add logging

feat(functions): Add generator function

feat(functions): implement fromSQL to support PostgreSQL, MySQL Includes support for Timescale.


docs/from_sql_setup.md
```diff
@@ -0,0 +1,221 @@
+# Quick Steps to Setup SQL Databases for Flux Testing
+
+## Timescale
```
stuartcarnie: We need to rework or remove the README from `master` for the SQL setup

cmd/fluxcli/main.go
```diff
@@ -0,0 +1,316 @@
...
+		return 0, err
+	}
+	results := flux.NewResultIteratorFromQuery(qry)
+	defer results.Cancel()
+
+	encoder := d.Encoder()
+	return encoder.Encode(w, results)
+}
+
+func NewQuerier() *Querier {
```
nathanielc: How does this daemon have any data to return? Is the expectation that you use sources that reach out to other systems? How are those sources configured?

cmd/fluxcli/main.go
```diff
@@ -0,0 +1,316 @@
+package main
```
stuartcarnie: Should this command be called `fluxcli` or `flux`?

cmd/fluxcli/main.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	q     = flag.String("q", "", "flux script")
+	httpd = flag.String("http", "", "run http server on specified port")
+)
+
+func main() {
+	flag.Parse()
```
stuartcarnie: Can be a separate PR, but I would like to rework `fluxcli` to use `spf13/cobra`:

```
fluxcli [command]

    exec   execute a flux query
    http   start a HTTP server (for REPL or Chronograf)

```

functions/from_sql.go
```diff
@@ -0,0 +1,231 @@
...
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+	_ "github.com/lib/pq"
+	"github.com/pkg/errors"
+	"reflect"
+	"time"
```
stuartcarnie: nit: we need to run `goimports` over all the changes

cmd/fluxcli/main.go
aanthony1243: yes, that's correct.  it assumes that the source you use will be able to read from its intended source without any support from this binary.  Not sure what would happen is a user tried to use influxdb.from

stuartcarnie: the built-in sources should be self contained, such that calling `fromCSV` or `fromSQL` in a Flux query should just execute.

For example, the path to `fromCSV` will serve up files from the local file system

stuartcarnie: @aanthony1243 I think `from` to InfluxDB will be removed from this repository, however, there isn't anything stopping us adding `fromInfluxDB` in the future that talks over gRPC

functions/generate.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/compiler"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+)
+
+const GenerateKind = "generate"
```
stuartcarnie: This is not conventional; all other sources start with `from`

functions/generate.go
```diff
@@ -0,0 +1,218 @@
...
+		},
+	}
+
+	for _, col := range cols {
+		b.AddCol(col)
+	}
+
+	cols = b.Cols()
+
+	colIndex := map[string]int{}
```
jlapacik: Can remove `colIndex`. Not used below.

docs/from_sql_setup.md
aanthony1243: I agree it doesn't really have a "place"  but it is a very nice bit of text that should be made available to the community somehow...maybe a blog?

aanthony1243:
this PR was split into #68 #67  #62 #61

PR #61, opened 2018-09-24
feature(functions): introduced the connect/fetch/decode interface for…

aanthony1243:
… writing sources

### DIFF
--- a/functions/from_iterator.go
+++ b/functions/from_iterator.go
@@ -1,0 +1,103 @@
+package functions
+
+import (
+	"context"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+)
+
+// Source Decoder is an interface that generalizes the process of retrieving data from an unspecified data source.
+//
+// Connect implements the logic needed to connect directly to the data source.
+//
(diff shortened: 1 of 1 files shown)


functions/from_iterator.go
```diff
@@ -0,0 +1,104 @@
...
+// Source Decoder is an interface that generalizes the process of retrieving data from an unspecified data source.
+//
+// Connect implements the logic needed to connect directly to the data source.
+//
+// Fetch implements a single fetch of data from the source (may be called multiple times).  Should return false when
+// there is no more data to retrieve.
+//
+// Decode implements the process of marshaling the data returned by the source into a flux.Table type.  Buffer size
+// may be controlled by ensuring that the rows are sorted by column _time,  and adding columns _start and _stop to
+// each row that indicate time boundaries for when the buffer should be flushed.
```
nathanielc: The rows themselves do not need to be sorted by time. Rather the tables need to be sorted and grouped by the _stop column. The SPEC says that tables need to be sorted by their group key. So here really we just need to state that the tables are grouped by _stop. 

For sources that do not have a time component, they can either not set the _stop column, which will buffer all data, or they can implement the Source interface directly and submit the UpdateWatermark and family events directly.

PR #62, opened 2018-09-24
Added fluxcli binary cmd

aanthony1243:


goller:
@aanthony1243 this is very similar to my work in platform.  Could this be moved there?

stuartcarnie:
@goller the expectation is that flux has an independent, standalone binary, with support for CSV and other contributed sources, directly from this repository to be readily go installable

goller:
Gotcha.

Is there any need to keep the API for platform and fluxcli in-sync?  Perhaps fluxcli API lives its own life?

aanthony1243:
I think it's just supposed to be a source-agnostic client that can get users started with flux.  The fact that it's chronograf-compatible was just a quick feature we built-in for a compelling demo on the hack day.  I think we can maintain this independently of the product API

PR #65, opened 2018-09-26
update spec for buckets function

aanthony1243:
closes https://github.com/influxdata/platform/issues/769

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1208,7 +1208,7 @@ It takes no input parameters and produces an output table with the following col
 * organization (string): the organization this bucket belongs to
 * organizationID (string): the internal ID of the organization
 * RetentionPolicy (string): the name of the retention policy, if present
-* RetentionPeriod (string): the duration of time for which data is held in this bucket
+* RetentionPeriod (duration): the duration of time for which data is held in this bucket
 
 Example: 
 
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -1198,6 +1198,22 @@ Example:
...
+#### Buckets
+
+Buckets is a type of data source that retrieves a list of buckets that the caller is authorized to access.  
+It takes no input parameters and produces an output table with the following columns: 
+
+* name: the name of the bucket
+* id: the internal ID of the bucket
+* organization: the organization this bucket belongs to
+* organizationID: the internal ID of the organization
+* Retention Policy: the name of the retention policy, if present
```
nathanielc: Do retention policies have names? I didn't think they did.

docs/SPEC.md
```diff
@@ -1198,6 +1198,22 @@ Example:
...
+* RetentionPeriod: the duration of time for which data is held in this bucket
```
nathanielc: These cols look good, could you add the data type of the column in the description somewhere?

docs/SPEC.md
```diff
@@ -1198,6 +1198,22 @@ Example:
...
+
+Example: 
+
+    buckets() |> filter(fn: (r) => r["organization"] == "my-org")
```
nathanielc: nit, could you use the `r.organization` syntax? I prefer it for the examples as it is less noisy.

docs/SPEC.md
aanthony1243: in what context?  in v1.X they do: 

> show retention policies
name    duration shardGroupDuration replicaN default
----    -------- ------------------ -------- -------
autogen 0s       168h0m0s           1        true

PR #67, opened 2018-09-26
Feat/from generator

aanthony1243:
generator source written during hackday.  Was authored as `generate(...)` but we changed it to `fromGenerator` to be consistent with other sources.  This could change in the future, but is probably the right call for now.

### DIFF
--- a/functions/from_generator.go
+++ b/functions/from_generator.go
@@ -195,11 +195,6 @@ func (s *GeneratorSource) Decode() (flux.Table, error) {
 
 	cols = b.Cols()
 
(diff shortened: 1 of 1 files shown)


functions/from_generator.go
```diff
@@ -0,0 +1,218 @@
...
+		},
+	}
+
+	for _, col := range cols {
+		b.AddCol(col)
+	}
+
+	cols = b.Cols()
+
+	colIndex := map[string]int{}
```
desa: I think we cal remove lines 198-201 since we use `execute.ColIdx` now

PR #72, opened 2018-09-28
Uint64 platform.IDs

leodido:
PR https://github.com/influxdata/platform/pull/539 depends on this branch (which in turn depends on that platform branch, but not using that branch atm).

It translates some flux code to use uint64 `platform.ID`s.


leodido:
No more needed since https://github.com/influxdata/flux/pull/115 merged in.

Thanks to anyone involved.

:tada:

PR #74, opened 2018-09-28
Logical and physical plan nodes with common interface

wolffcm:


planner/physical.go
```diff
@@ -1,98 +1,34 @@
...
-type PhysicalPlan interface {
-	physical()
-	Predecessors() []PhysicalPlan
-	Successors() []PhysicalPlan
-	Cost() Cost
+// PhysicalProcedureSpec is similar to its logical counterpart but must provide a method to determine cost
+type PhysicalProcedureSpec interface {
+	Kind() plan.ProcedureKind
+	Copy() plan.ProcedureSpec
+	Cost(inStats []Statistics) (cost Cost, outStats Statistics)
```
jlapacik: Why does `Cost()` produce `Statistics`? Should `PhysicalProcedureSpec` also have a `Statistics()` method?

PR #75, opened 2018-09-28
major refactor of functions package into builtins package.  split out sources to an inputs folder.

aanthony1243:
new structure:
functions/ : top level package.
   inputs/ : Holds all input sources in the flux repo
   transformations/ : Holds all builtin procedure code in the flux repo
   testdata/ : Holds end-to-end tests for inputs, transformations, outputs
   doc.go : documentation for package functions
   functions_test.go : driver for end-to-end tests
   outputs/ :   planned, but not present because `flux` repo has no outputs presently.


builtin/functions/keys.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
+	. "github.com/influxdata/flux/builtin/inputs"
```
nathanielc: I do not like the use of this kind of import as it makes it harder to follow the code to its source.
Can you update the code here to use the `inputs` package name?

builtin/inputs/from_test.go
```diff
@@ -104,38 +106,40 @@ func TestFrom_NewQuery(t *testing.T) {
...
+//		},
+//	}
+//	for _, tc := range tests {
+//		tc := tc
+//		t.Run(tc.Name, func(t *testing.T) {
+//			t.Parallel()
+//			pquerytest.BucketAwareQueryTestHelper(t, tc)
+//		})
+//	}
+//}
```
nathanielc: Can you fix this test case to pass again?

aanthony1243: depends on code from platform, which won't build because I moved too many things here.  So we need to refactor platform then this test should be OK

builtin/functions/keys.go
aanthony1243: yep, I thought there might be an objection or two, no problem.

pauldix:
Structure makes. Although, why have test data live in its own directory rather than alongside where it is used i.e. `inputs`, `transformations`, `outputs`?

aanthony1243:
@pauldix that's a good idea.  Currently all of the CSV tests are for transformations so I should move them in there.    I was just thinking that we need some end to end tests for the inputs and outputs too, but they will probably require a different approach than what we do for the transformations.

PR #82, opened 2018-10-02
Add Flux function `union` to SPEC

wolffcm:
I treated `join()` as a template where applicable, and included an example so semantics are clear.

Question: should the parameter name be called `streams` instead of `tables`?  This question applies to `join()` too.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2146,6 +2146,69 @@ Examples:
     | 0002  | "temp" | 75 | "temp" | 56 |
     | 0003  | "temp" | 72 | "temp: | 55 |
 
+#### Union
+
+Union concatenates two input streams into a single output stream.  In tables that have identical
+schema and group keys, contents of the tables will be concatenated in the output stream.
+
+Example:
+
+* SF_Weather
+* Group Key for both tables {"_field"}
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "temp" | 70 |
+    | 0002  | "temp" | 75 |
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "humidity" | 81 |
+    | 0002  | "humidity" | 82 |
+
+* NY_Temperature
+* Group Key for both tables {"_field"}
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "temp" | 55 |
+    | 0002  | "temp" | 56 |
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "pressure" | 29.82 |
+    | 0002  | "pressure" | 30.01 |
+
+`union(tables: {sf: SF_Temperature, ny: NY_Temperature})` produces:
+
+* Group Key for all tables {"_field"}
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "temp" | 70 |
+    | 0002  | "temp" | 75 |
+    | 0001  | "temp" | 55 |
+    | 0002  | "temp" | 56 |
+
+    | _time | _field | _value |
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -2146,6 +2146,69 @@ Examples:
...
+
+##### Options
+
+| Name   | Type | Required | Default Value | Possible Values |
+| ------ | ---- | -------- | ------------- | --------------- |
+| tables | map  | yes      |  none         |  N/A            |
+
+##### Output Schema
+
+The output schema of union shall be identical to the input schema.
```
aanthony1243: ...shall be the union of all input schemas.

docs/SPEC.md
```diff
@@ -2146,6 +2146,69 @@ Examples:
...
+    | ----- | ------ | ------ |
+    | 0001  | "temp" | 55 |
+    | 0002  | "temp" | 56 |
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "pressure" | 29.82 |
+    | 0002  | "pressure" | 30.01 |
+
+`union(tables: {sf: SF_Temperature, ny: NY_Temperature})` produces:
```
aanthony1243: @jlapacik   what happens if there's > 2 tables for join?   

@wolffcm  do we want to support N >= 2 tables for union?  Seems like it wouldn't be much more complex than the N=2 case.

jlapacik: @aanthony1243 join does not support joining more that two streams at the moment.

wolffcm: It seems like we could handle more than 2 streams with no change at all to the processing algorithm.  Is there any reason not to support an arbitrary number of of tables?

jlapacik: I don't think so. I would say do it if it's trivial.

aanthony1243: @jlapacik probably not urgent, but the spec leaves it undefined what happens if you pass >2 tables to join

docs/SPEC.md
```diff
@@ -2146,6 +2146,69 @@ Examples:
...
+    | ----- | ------ | ------ |
+    | 0001  | "humidity" | 81 |
+    | 0002  | "humidity" | 82 |
+
+    | _time | _field | _value |
+    | ----- | ------ | ------ |
+    | 0001  | "pressure" | 29.82 |
+    | 0002  | "pressure" | 30.01 |
+
+##### Options
```
jlapacik: Can you remove this options sub-section and replace it with: "Union has the following properties: ... " under the initial description. All specs should follow this convention but Join does not.

docs/SPEC.md
```diff
@@ -2146,6 +2146,69 @@ Examples:
     | 0002  | "temp" | 75 | "temp" | 56 |
     | 0003  | "temp" | 72 | "temp: | 55 |
 
+#### Union
+
+Union concatenates two input streams into a single output stream.  In tables that have identical
+schema and group keys, contents of the tables will be concatenated in the output stream.
+
```
jlapacik: Can you list the properties (or in this case property) of Union under this description?

docs/SPEC.md
jlapacik: @aanthony1243 there are also a few other things I need to change in the join spec. I'll put up a follow-up PR.

nathanielc: >  Should it use a simple list of streams instead?

I was thinking that union should just take a list of tables.

nathanielc:
> Question: should the parameter name be called streams instead of tables? This question applies to join() too.

The parameter name should be `tables`.

jlapacik:
@wolffcm The example tables don't seem to be getting rendered correctly. Perhaps remove the tabs?

wolffcm:
@jlapacik Okay, I fixed the markdown so the tables render.  I had to indent with only three spaces for some reason, even though other tables indent with 4 and render just fine.

Issue #83, opened 2018-10-02
Add support for joining more than 2 streams


jpacik:
The **join** operator currently only supports two input streams. However the Flux language spec says that **join** can support two or more input streams.

PR #86, opened 2018-10-02
update join spec to follow set convention

jlapacik:
Previously the join spec did not follow the typical template that other function specs follow. In particular it did not list its properties after its initial description. This PR updates the join spec to follow the convention of other function specs. It also adds links to features that have yet to be implemented.


jlapacik:
@wolffcm usually you link the issue for the unimplemented behavior in the spec like this:

[IMPL#83](https://github.com/influxdata/flux/issues/83) Add support for joining more than 2 streams

PR #87, opened 2018-10-03
logical-to-physical planner

jlapacik:
This PR adds the following features:

1. A basic logical-to-physical planner that takes a logical plan and set of transformations that map logical query operations to physical query specs.

2. It introduces several patterns for sub-plan matching.

3. It defines (but does not implement) several physical leaf operators. I.e different `from` procedure specs.

All features are not final. This PR is meant as a first rough draft for how things could look.


planner/planner_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/plan/plantest"
+	"github.com/influxdata/flux/planner"
+)
+
+// from() |> range() -> FromRange
+func TestPlanFromRange(t *testing.T) {
+	// Instantiate planner
+	physicalPlanner := planner.NewLogicalToPhysicalPlanner([]planner.Rule{
```
wolffcm: It would be great to be able to start from Flux text and convert it to a logical plan.  That would make these tests much easier to read.

planner/transformation_rules.go
```diff
@@ -0,0 +1,118 @@
...
+	if !rule.visitor.conjuctive || len(rule.visitor.tagFilters) == 0 {
+		return nil, false
+	}
+
+	tagFilters := make(map[string]string, len(rule.visitor.tagFilters))
+	for k, v := range rule.visitor.tagFilters {
+		tagFilters[k] = v
+	}
+
+	return &PhysicalPlanNode{
```
wolffcm: So the rule doesn't set any of its predecessors or successors?  This doesn't seem right to me.  I think the rule needs to at least be able to set the successors of the new node it creates.

planner/planner.go
```diff
@@ -9,5 +9,98 @@ type Planner interface {
...
+		if node, matched := p.matchRule(pred, rule); matched {
+			return node, matched
+		}
+	}
+	return nil, false
+}
+
+func (p LogicalToPhysicalPlanner) AddRules(rules []Rule) {
+	for _, rule := range rules {
+		p.rules[rule.Pattern().Root()] = rule
```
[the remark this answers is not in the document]
jlapacik: You're right. This was supposed to be a map of slices.

planner/transformation_rules.go
```diff
@@ -0,0 +1,118 @@
...
+
+// FromRangeTransformationRule rewrites the sequence from() |> range()
+type FromRangeTransformationRule struct{}
+
+// Pattern returns a pattern of the form:
+//     range
+//       |
+//     from
+func (rule FromRangeTransformationRule) Pattern() Pattern {
+	return &TreePattern{
```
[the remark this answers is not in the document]
jlapacik: I agree. This was a first attempt and I'm not really satisfied with it either.

planner/transformation_rules.go
jlapacik: So the way it is right now, the planner handles setting predecessors and successors instead of the rule itself. As I'm writing this I realize that this approach is very simple for rewriting leaf nodes. However non-leaf nodes aren't quite as simple. You may be right. This logic might have to live with the rule itself. We can discuss more.

wolffcm:
There are some issues we're working on, but let's merge this to the feature branch to fix them.

PR #90, opened 2018-10-03
Fixed row/col counters, add test to demonstrate fix

aanthony1243:
counters were being allocate 1 per stream.  They needed to be allocated one per table, identified by the group key.  Replaced single variable with a map keyed on the group key values.


functions/pivot.go
```diff
@@ -231,11 +236,10 @@ func (t *pivotTransformation) Process(id execute.DatasetID, tbl flux.Table) erro
 		}
 		t.colKeyMaps[groupKeyString] = make(map[string]int)
 		t.rowKeyMaps[groupKeyString] = make(map[string]int)
-		t.nextCol = len(cols)
-		t.nextRow = 0
+		t.nextRowCol[groupKeyString] = rowCol{nextCol: len(cols), nextRow: 0}
 	}
 
-	tbl.Do(func(cr flux.ColReader) error {
+	err := tbl.Do(func(cr flux.ColReader) error {
```
lyondhill: Nice :)

functions/testdata/pivot_task_test.in.csv
```diff
@@ -0,0 +1,14 @@
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,string,string,string,string
+#group,false,false,true,true,false,false,true,true,true,true,true
+#default,_result,,,,,,,,,,
+,result,table,_start,_stop,_time,_value,_field,taskID,orgID,state,_measurement
+,,0,2018-10-02T17:55:11.520461Z,2018-10-03T17:55:11.520461Z,2018-10-03T17:55:11.01114Z,02bac3c908f37000,runID,02bac3c8f0f37000,02bac3c8d6c5b000,started,records
```
lyondhill: my test case actually might have a intermittent field requestedAt. Could we add that into the test case here and make sure it works with fields that are not present in every table?

```
Table: keys: [taskID, orgID, state]
         taskID:string            orgID:string            state:string                      _time:time      requestedAt:string            runID:string     scheduledFor:string  
----------------------  ----------------------  ----------------------  ------------------------------  ----------------------  ----------------------  ----------------------  
      02badb212b203000        02badb21104d3000                 success  2018-10-03T19:37:10.770820000Z    1970-01-01T00:00:01Z        02badb215be03000    2018-10-03T19:37:13Z  
      02badb212b203000        02badb21104d3000                 started  2018-10-03T19:37:10.577552000Z                                02badb212ba03000    2018-10-03T19:37:11Z  
      02badb212b203000        02badb21104d3000                 success  2018-10-03T19:37:10.670483000Z                                02badb2142e03000    2018-10-03T19:37:12Z  
```

PR #91, opened 2018-10-04
Sources promql

aurrelhebert:
Hello, 

Based on the [https://community.influxdata.com/t/connect-flux-to-other-tsdb/](https://community.influxdata.com/t/connect-flux-to-other-tsdb/), I implemented a first connector to a Prometheus backend. This PR is handling the pushdown logic for the Range and filter function.

Work in progress: 

- [ ] White list URL
- [ ] Manage a basic auth


pauldix:
Thanks for submitting this! I should mention that after I had a chat with the Prometheus developers at PromCon, it was suggested that using Prometheus' remote read endpoint is a much better fit for this kind of query. It could also attempt to read in chunks rather than all at once, but that may be more difficult to figure out. It's certainly something we need built into the primitives of adding a new source.

pauldix:
@aurrelhebert the remote read interface still supports sending the matcher. So you'll still be able to push down the filter and the range. Well, maybe not filter on the actual value?

aurrelhebert:
Hi Paul, 
   I made a quick poc this afternoon on using the remote-read with a Prometheus matcher (completed with the filter). I kept for now the query range to still allow the use of PromQL queries. 
   I will continue to work on this PR, to improve how I handle the chunks and reformat the HTTP client to be more generic (eventually to support later on others TSDB).

aurrelhebert:
Just rebased the branch to match code structure of PR #75

PR #93, opened 2018-10-04
Flux Union function implementation

wolffcm:


docs/SPEC.md
```diff
@@ -2203,6 +2203,10 @@ Union concatenates two or more input streams into a single output stream.  In ta
 schema and group keys, contents of the tables will be concatenated in the output stream.  The output schemas of 
 the Union operation shall be the union of all input schemas.
 
+Note that while the tables in the stream produced by Union will be ordered, the rows within those tables
+may appear in any order, due to Flux's parallel execution model. The output of Union may be piped through
+a Sort function to address this, if desired.
```
nathanielc: I know this was merged, but I'd like to see a small update to this bit.

Instead of discussing the implementation details of parallel execution lets simply specify the behavior.

Maybe something like this:

```
Union does not preserve the sort order of the rows within tables. A sort operation may be added if a specific sort order is needed. 
```

wolffcm: Sure, will do.

PR #96, opened 2018-10-05
top-level QueryPlan struct

jlapacik:
This PR creates a top-level structure for holding a query plan DAG. It also un-exports `Edge` fields. These fields do not need to be exported as the new testing framework won't need them to be.


planner/types.go
```diff
@@ -18,11 +18,41 @@ type PlanNode interface {
...
-	AddPredecessor(PlanNode)
+	AddSuccessors(...PlanNode)
+	AddPredecessors(...PlanNode)
 	RemovePredecessor(PlanNode)
 }
 
+// QueryPlan holds the roots of the query plan DAG.
+// Roots correspond to nodes that produce final results.
+type QueryPlan struct {
+	roots map[PlanNode]struct{}
```
wolffcm: I've never seen this before.  Is this the idiomatic way to have something like `std::unordered_set` in go?

planner/types.go
```diff
@@ -18,11 +18,41 @@ type PlanNode interface {
...
+}
+
+// NewQueryPlan instantiates a new query plan
+func NewQueryPlan(nodes []PlanNode) *QueryPlan {
```
wolffcm: This argument would be better off called `roots` I think, lest callers think that they need to pass a slice of every node in the plan.

planner/types.go
```diff
@@ -33,39 +63,39 @@ type ProcedureSpec interface {
...
 
 func (e *Edges) Successors() []PlanNode {
-	return e.Succ
+	return e.successors
 }
 
-func (e *Edges) AddSuccessor(node PlanNode) {
-	e.Succ = append(e.Succ, node)
+func (e *Edges) AddSuccessors(nodes ...PlanNode) {
+	e.successors = append(e.successors, nodes...)
```
wolffcm: I think maybe these methods should also add the back-edges... if a node is adding a successor then the successor also needs to add a predecessor.  Not sure what the right set of helper methods are, but I think it will become more apparent as we do more rewriting.

planner/types.go
jlapacik: I'm not sure. I've seen it done `map[PlanNode]bool` too, but `struct{}` has a size of 0.

planner/types.go
jlapacik: True, I'll fix.

PR #99, opened 2018-10-05
Flux spec to logical plan tests

jlapacik:


planner/logical_test.go
```diff
@@ -0,0 +1,180 @@
...
+		return &planner.FilterProcedureSpec{
+			Fn: spec.Fn.Copy().(*semantic.FunctionExpression),
+		}, nil
+	},
+}
+
+func TestFluxSpecToLogicalPlan(t *testing.T) {
+	testcases := []struct {
+		name string
+		spec *flux.Spec
```
wolffcm: How about starting from Flux text, translating it to a Spec and then to the logical plan in the `Run` method below?  That way these tests would be somewhat easier to read.

planner/plannertest/cmp.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux/planner"
+	"github.com/influxdata/flux/semantic/semantictest"
+)
+
+// CompareLogicalPlans recursively compares two logical query plans.
+// It returns an error if it encounters a non-matching node.
+// TODO(jlapacik): Need a way to consistently order nodes
+func CompareLogicalPlans(want, got []planner.PlanNode, seen map[planner.PlanNode]bool) error {
```
wolffcm: I wonder if defining a custom `Comparer` in the `cmp` package would take care of a lot of the compare logic here?

planner/plannertest/cmp.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux/planner"
+	"github.com/influxdata/flux/semantic/semantictest"
+)
+
+// CompareLogicalPlans recursively compares two logical query plans.
+// It returns an error if it encounters a non-matching node.
+// TODO(jlapacik): Need a way to consistently order nodes
```
wolffcm: We should define a `Walk` method similar to Specs.  Edges are ordered, so there's no reason not to have a deterministic traversal order.

planner/plannertest/cmp.go
```diff
@@ -0,0 +1,114 @@
...
+
+		if err := ComparePhysicalPlans(wantNode.Predecessors(), got[i].Predecessors(), seen); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// CreateLogicalPlanTree creates a logical plan tree (not a DAG)
+func CreateLogicalPlanTree(spec planner.LogicalProcedureSpec, predecessors ...*planner.LogicalPlanNode) *planner.LogicalPlanNode {
```
wolffcm: I see what you mean about creating an expected plan that is a DAG and not a tree.

I'm thinking that the best way to write test cases that lock down translation from Flux text to the logical plan would use a struct like this:

```go
type TestCase struct {
    name string,
    fluxText string,
    logicalNodes []LogicalPlanNode,
    edges [][2]int,
}
```
Tests can then define a slice of logical nodes, and describe edges using offsets into the slice.  It would be up to the method that runs each test case to create the edges.  What do you think?

planner/logical_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/functions"
+	"github.com/influxdata/flux/planner"
+	"github.com/influxdata/flux/planner/plannertest"
+	"github.com/influxdata/flux/semantic"
+)
+
+var create = map[flux.OperationKind]planner.CreateLogicalProcedureSpec{
```
wolffcm: So you're creating this map here... What would it take to use the map that gets created by all the registration that goes on in the `functions` package?  I guess that's the next step.

planner/logical_test.go
jlapacik: That's true. Would these still be considered unit tests if we were to do that? Does is matter if they're not unit tests?

planner/plannertest/cmp.go
jlapacik: Edges of logical plan are ordered?

planner/logical_test.go
wolffcm: Hmm.  That's a good question.  Personally, I think it's okay to start from a Flux program in unit tests, because it's easier to read, and it provides a convenient way to produce a Spec whose graph has the topology you want to test.  It's pretty much a 1:1 conversion from Spec operations to logical nodes, so it's not adding a whole lot of complexity.

planner/plannertest/cmp.go
jlapacik: I tried defining a custom `Comparer` however, in the event that two plans aren't the same, the above function will give a much richer diff than a custom `Comparer`.

planner/plannertest/cmp.go
```diff
@@ -0,0 +1,161 @@
...
+
+		// Recurse on predecessors
+		if err := cmpPhysicalPlans(p[i].Predecessors(), q[i].Predecessors(), visited, level+1); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type cmpPair struct {
```
wolffcm: I'm not sure what's idiomatic for Go, but I would define this struct further up in this file (between the exported and non-exported functions) to help the code read a bit easier.

planner/plannertest/cmp.go
```diff
@@ -0,0 +1,161 @@
...
+
+		if _, ok := q[i].(*planner.PhysicalPlanNode); !ok {
+			return fmt.Errorf("encountered a non-physical plan node with spec type %T", q[i].ProcedureSpec())
+		}
+
+		left := p[i].(*planner.PhysicalPlanNode)
+		right := q[i].(*planner.PhysicalPlanNode)
+
+		// Must have the same IDs
+		if p[i].ID() != q[i].ID() {
```
wolffcm: It seems like many of these checks are common between logical and physical nodes.  Physical nodes are like logical nodes with some additional attributes.  I think this could be refactored with a function that does the common checks.

PR #101, opened 2018-10-08
Modify plan-traversal algorithm

wolffcm:
Also:
- Added a new `ID()` method to `PlanNode`
- Added ability to add rules that match any node
- Added plan traversal test

### DIFF
--- a/planner/plan_traversal_test.go
+++ b/planner/plan_traversal_test.go
@@ -1,8 +1,9 @@
-package planner
+package planner_test
 
 import (
 	"fmt"
 	"context"
+	"github.com/influxdata/flux/planner"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


planner/plan_traversal_test.go
```diff
@@ -0,0 +1,183 @@
+package planner
```
jlapacik: Convention is that tests are placed in a separate package. Is there a reason for having them in the same package?

planner/planner.go
```diff
@@ -5,102 +5,122 @@ type Planner interface {
...
+				newNode, changed := p.matchRules(node)
+				anyChanged = anyChanged || changed
+				if node != newNode {
+					updateSuccessors(inputPlan, node, newNode)
+				}
+
+				// append to stack in reverse order so lower-indexed children
+				// are visited first.
+				for i := len(newNode.Predecessors()); i > 0; i-- {
+					nodeStack = append(nodeStack, newNode.Predecessors()[i - 1])
```
jlapacik: What about successors? If a node (call it Node) is rewritten, there could conceivably be a rule that previously did not match a successor of Node but now does. Hence that successor should be visited again, right?

planner/planner.go
```diff
@@ -5,102 +5,122 @@ type Planner interface {
...
+// Plan may change its argument and/or return a new instance of QueryPlan, so the correct way to call Plan is:
+//     plan, err = planner.Plan(plan)
+func (p *LogicalToPhysicalPlanner) Plan(inputPlan *QueryPlan) (*QueryPlan, error) {
+
+	visited := make(map[PlanNode]struct{})
+
+	nodeStack := make([]PlanNode, len(inputPlan.Roots()))
+	copy(nodeStack, inputPlan.Roots())
+
+	for anyChanged := true; anyChanged == true; {
```
jlapacik: Nice.

planner/plan_traversal_test.go
wolffcm: No, no reason.  I did not realize that it was a convention.  I'll change it.

PR #102, opened 2018-10-08
Feat/bucket lookup all

aanthony1243:
Affects both Platform and IDPE, so we need to test those before we merge.

### DIFF
--- a/functions/inputs/storage/storage.go
+++ b/functions/inputs/storage/storage.go
@@ -20,7 +20,7 @@ type HostLookup interface {
 
 type BucketLookup interface {
 	Lookup(orgID platform.ID, name string) (platform.ID, bool)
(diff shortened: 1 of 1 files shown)


functions/inputs/storage/storage.go
```diff
@@ -20,6 +20,7 @@ type HostLookup interface {
 
 type BucketLookup interface {
 	Lookup(orgID platform.ID, name string) (platform.ID, bool)
+	FindAllBuckets(orgID platform.ID) ([]*platform.Bucket, int)
```
nathanielc: What purpose does the int serve?

aanthony1243: the underlying service interface returns a count of the number of buckets, so I pass it up.  it's not essential.

aanthony1243:
migrated to platform repo

PR #112, opened 2018-10-09
Merge master into feat-planner

wolffcm:
This just merges the latest master into feat-planner.  We haven't yet made any changes in code outside of the new planner package, so there were no conflicts.

Doing this now ahead of making changes to `functions`, `control`, and `execute` to use the new planner.


wolffcm:
My plan traversal test is failing... I'll fix it.

PR #121, opened 2018-10-10
Updated planner to check specifically for specs that require bounds are set

aanthony1243:
previously we checked the results to be sure they were bounded.  But some of our sources don't require/create bounds.


plan/physical.go
```diff
@@ -226,13 +226,12 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
...
-	// by any parent procedure node.
-	for name, yield := range p.plan.Results {
-		if pr, ok := p.plan.Procedures[yield.ID]; ok {
-			if pr.Bounds == nil {
+	// check for bounded procedure specs and make sure they have non-zero bounds.
+	for name, proc := range p.plan.Procedures {
+		pr, ok := proc.Spec.(BoundedProcedureSpec)
+		proc.Spec.Kind()
+		if ok {
+			if pr.TimeBounds().Start.IsZero() && pr.TimeBounds().Stop.IsZero() {
```
wolffcm: So before we were just looking at the result-producing procedures and making sure that they were bounded based on the propagation we did in the previous loop near line 164.  (Tangentially related: `shift` should affect how we propagate bounds, right?)

The only procedures that implement `BoundedProcedureSpec` are `from` and `range`.  Range should always have non-zero bounds, correct?  So the effect of this change is to make it so that `from` must have had a range pushed down into it.  Other sources like `fromGenerator`, `fromCSV` will therefore be allowed to be unbounded.

It seems like the check you want to make is specific to `from`.  I realize you can't reference `FromKind` directly from the `plan` package, so it seems like what we really want is an interface that procedure specs can optionally implement that does any required checks before planning is complete.

Since we're switching planners soon anyways, I'm okay with the approach you've taken here.

aanthony1243: I assumed that since the previous version was checking results at this point, then it should be true for the whole path of the plan that the bounds are either set explicitly  or else inherited from a parent.   `from` can push down bounds from one of `range` `first` and `last`.  

This check is, in fact, stronger than the previous check.  It _could_ have been the case that the push down rule failed and so `from` would not have any bounds set, though the `range` would pass its bounds down the line to the results.  But our true goal here is to bound `from` to limit what comes across the network from storage.

PR #126, opened 2018-10-12
Integrate new query plans with the execution engine

jlapacik:


execute/executor.go
```diff
@@ -119,77 +117,106 @@ type triggeringSpec interface {
...
+		// No need to add transports to predecessors as results pull data from downstream.
+		yield, ok := spec.(plan.YieldProcedureSpec)
+
+		if !ok {
+			return fmt.Errorf("expected yield, instead got %T", spec)
+		}
+
+		numPreds := len(node.Predecessors())
+
+		if numPreds != 1 {
```
wolffcm: There can be only one predecessor for a yield?  I wonder why.

execute/executor.go
```diff
@@ -94,21 +93,20 @@ func (e *executor) createExecutionState(ctx context.Context, p *plan.PlanSpec, a
 		deps:      e.deps,
 		alloc:     a,
 		resources: p.Resources,
-		results:   make(map[string]flux.Result, len(p.Results)),
+		results:   make(map[string]flux.Result, len(p.Roots())),
```
wolffcm: I think I like the name `Results` better as its unambiguous... it could be possible someone interprets `Roots` as the sources.

execute/executor.go
```diff
@@ -275,8 +302,8 @@ func (ec executionContext) Allocator() *Allocator {
 func (ec executionContext) Parents() []DatasetID {
 	return ec.parents
 }
-func (ec executionContext) ConvertID(id planner.ProcedureID) DatasetID {
-	return DatasetID(plan.ProcedureID(id))
+func (ec executionContext) ConvertID(id plan.ProcedureID) DatasetID {
```
wolffcm: So I guess we still need procedure ids?  Seems like there should be a way to do away with them.

planner/types.go
```diff
@@ -66,13 +73,43 @@ func (plan *PlanSpec) Replace(root, with PlanNode) {
 	plan.roots[with] = struct{}{}
 }
 
+// TopDownWalk will execute f for each plan node in the PlanSpec,
+// only visiting a node after all its predecessors have been visited.
```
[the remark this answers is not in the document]
jlapacik: Yes I'll fix this.

planner/types.go
```diff
@@ -66,13 +73,43 @@ func (plan *PlanSpec) Replace(root, with PlanNode) {
...
+func (plan *PlanSpec) TopDownWalk(f func(node PlanNode) error) error {
+	var nodes []PlanNode
+
+	plan.BottomUpWalk(func(node PlanNode) error {
+		nodes = append(nodes, node)
+		return nil
+	})
+
+	for i, j := 0, len(nodes)-1; i < j; i, j = i+1, j-1 {
```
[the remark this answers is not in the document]
jlapacik: Yeah that's a good idea.

execute/executor.go
wolffcm: Yes, you're right.  The user can put in a `union` if they want to do this.  I was confused.

planner/types.go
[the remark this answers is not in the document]
jlapacik: I spoke too soon. Is `WalkTopDown` supposed to visit a node "before all its predecessors" or "after all its successors"?

PR #128, opened 2018-10-13
remove stream context and resource errors

jlapacik:
* Remove stream context for right now and replace with bounds
* Have controller insert planning resources into physical plan
* Validate logical plan has correct number of yields

### DIFF
--- a/functions/transformations/range.go
+++ b/functions/transformations/range.go
@@ -8,7 +8,6 @@ import (
 	plan "github.com/influxdata/flux/planner"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
-	"github.com/pkg/errors"
 )
 
 const RangeKind = "range"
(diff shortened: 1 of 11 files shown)


functions/transformations/range.go
```diff
@@ -140,12 +139,12 @@ func createRangeTransformation(id execute.DatasetID, mode execute.AccumulationMo
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
 
-	bounds := a.StreamContext().Bounds()
-	if bounds == nil {
-		return nil, nil, errors.New("nil bounds supplied to range")
+	bounds := execute.Bounds{
```
wolffcm: I guess we'll need to do something similar in the `platform` repo too, in `createFromSource`.

planner/logical.go
```diff
@@ -163,6 +170,29 @@ func (v *fluxSpecVisitor) visitOperation(o *flux.Operation) error {
...
+			if !ok {
+				return nil, errors.New("query must have explicit yields for multiple result")
+			}
+
+			if name, ok := names[spec.YieldName()]; ok {
+				return nil, fmt.Errorf("found duplicate yield name %q", name)
+			}
+
+			if len(root.Predecessors()) != 1 {
+				return nil, errors.New("yield procedures must have exactly one predecessor")
```
wolffcm: Is this unreachable?  I guess since `flux.Spec` is part of the public API, it's possible to start with a graph where nodes have the wrong number of predecessors.  Probably we'll want checks like this for each kind of node.

PR #129, opened 2018-10-13
fix how specs are constructed from table objects

### LINKED ISSUE #103: flux.Compile ignores multiple yields
In this Flux program
```
from(bucket: "foo") |> filter(fn: (r) => r._field == "cpu") |> yield(name: "1")
from(bucket: "foo") |> filter(fn: (r) => r._field == "fan") |> yield(name: "2")
```
The `query.Spec` produced by `flux.Compile` will omit operations for the second statement.  I've traced the issue to how we compare table objects in `TableObject.Equal` in `flux/compile.go`.  The table objects compare as equal even though they will contain different content at run time.

What's the right way to fix this?

jlapacik:
Fixes #103 

This commit fixes how the compiler compares TableObjects
and, as a result, how it constructs a Flux Spec. This
fixes an error where queries with multiple yields were
having all but one of those yields stripped during compilation.

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -420,14 +420,14 @@ func (t *TableObject) Equal(rhs values.Value) bool {
 	if t.Len() != r.Len() {
 		return false
 	}
-	for _, k := range t.keys() {
-		val1, ok1 := t.Get(k)
-		val2, ok2 := r.Get(k)
-		if !ok1 || !ok2 || !val1.Equal(val2) {
-			return false
-		}
-	}
-	return true
+	var isEqual = true
+	// Range over both TableObjects and
+	// compare their properties for equality
+	t.Range(func(k string, v values.Value) {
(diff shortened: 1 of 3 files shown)


compile.go
```diff
@@ -420,14 +420,14 @@ func (t *TableObject) Equal(rhs values.Value) bool {
...
-	}
-	return true
+	var isEqual = true
+	// Range over both TableObjects and
+	// compare their properties for equality
+	t.Range(func(k string, v values.Value) {
+		w, ok := r.Get(k)
+		isEqual = isEqual && ok && v.Equal(w)
+	})
+	return isEqual
```
wolffcm: If `r` has a key that `t` does not, then they will still compare as equal, right?  Seems like you want to also require that `r` and `t` have the same number of keys.

PR #130, opened 2018-10-15
Add rule registration to planners

wolffcm:


planner/registration.go
```diff
@@ -44,3 +44,27 @@ func RegisterProcedureSpec(k ProcedureKind, c CreateProcedureSpec, qks ...flux.O
...
 	}
 }
+
+type CreateRule = func() Rule
+
+var ruleNameToLogicalRule = make(map[string]Rule)
+var ruleNameToPhysicalRule = make(map[string]Rule)
+
+// RegisterLogicalRule registers the rule created by createFn with the logical planner.
+func RegisterLogicalRule(createFn CreateRule) {
```
jlapacik: Why not just register a rule as `RegisterLogicalRule(rule Rule)`?

wolffcm: Hmm, yes, I think you are right that's easier all around.  I will change it.

PR #134, opened 2018-10-16
remove yields from physical plan

jlapacik:
* Remove yields during physical planning
* Build execution graph assuming no yield nodes
* Update `plantest` with a `LogicalPlanSpec` and `PhysicalPlanSpec` for testing both phases

### DIFF
--- a/planner/heuristic_planner.go
+++ b/planner/heuristic_planner.go
@@ -51,8 +51,10 @@ func (p *heuristicPlanner) Plan(inputPlan *PlanSpec) (*PlanSpec, error) {
 
 		visited := make(map[PlanNode]struct{})
 
-		nodeStack := make([]PlanNode, len(inputPlan.Results()))
-		copy(nodeStack, inputPlan.Results())
+		nodeStack := make([]PlanNode, 0, len(inputPlan.Roots))
+		for root := range inputPlan.Roots {
+			nodeStack = append(nodeStack, root)
+		}
 
 		anyChanged = false
 		for len(nodeStack) > 0 {
--- a/planner/physical.go
+++ b/planner/physical.go
@@ -1,6 +1,10 @@
 package planner
 
-import "math"
+import (
+	"errors"
+	"fmt"
+	"math"
+)
 
 // PhysicalPlanner performs transforms a logical plan to a physical plan,
 // by applying any registered physical rules.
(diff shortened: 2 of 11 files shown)


planner/heuristic_planner.go
```diff
@@ -51,8 +51,10 @@ func (p *heuristicPlanner) Plan(inputPlan *PlanSpec) (*PlanSpec, error) {
 
 		visited := make(map[PlanNode]struct{})
 
-		nodeStack := make([]PlanNode, len(inputPlan.Results()))
-		copy(nodeStack, inputPlan.Results())
+		nodeStack := make([]PlanNode, 0, len(inputPlan.Roots))
+		for root := range inputPlan.Roots {
```
wolffcm: Does this not do the same thing as `copy`?  I thought that `copy` just did an elementwise assignment, but maybe I was wrong.

planner/physical.go
```diff
@@ -34,17 +38,58 @@ func (pp *physicalPlanner) Plan(spec *PlanSpec) (*PlanSpec, error) {
 		return nil, err
 	}
 
+	// Convert yields into result list
+	// TODO: Implement this via a transformation rule
```
wolffcm: Interesting idea to remove `yield` nodes via a transformation rule.  Rules would need to be able to update the `PlanSpec.Results` with the yield name...

wolffcm: I've been pondering whether or not plan nodes should contain a pointer to the `PlanSpec` instance that owns them, this makes me think they should.

planner/types.go
```diff
@@ -27,74 +27,48 @@ type PlanNode interface {
...
-	return plan
-}
-
-func newPlanSpec(results []PlanNode) *PlanSpec {
-	r := make(map[PlanNode]struct{}, len(results))
-	for _, root := range results {
-		r[root] = struct{}{}
-	}
-	return &PlanSpec{results: r}
+	Roots     map[PlanNode]struct{}
```
wolffcm: So `Roots` and `Results` are very similar.  They are likely to be confused as this is now.  Are you thinking that once the physical plan and logical plan are represented via different structures, `Roots` goes in the logical and `Results` in the physical?

planner/types.go
```diff
@@ -27,74 +27,48 @@ type PlanNode interface {
...
-	return plan.now
-}
-
-// Replace replaces one of the result nodes of the query plan
-func (plan *PlanSpec) Replace(result, with PlanNode) {
-	delete(plan.results, result)
-	plan.results[with] = struct{}{}
+// Replace replaces one of the root nodes of the query plan
+func (plan *PlanSpec) Replace(root, with PlanNode) {
+	delete(plan.Roots, root)
```
wolffcm: If this method is called during physical planning, then `Roots` and `Results` will become out of sync, right?

planner/physical.go
```diff
@@ -34,17 +38,58 @@ func (pp *physicalPlanner) Plan(spec *PlanSpec) (*PlanSpec, error) {
...
+	final, err := removeYields(transformedSpec)
```
wolffcm: Curious why you decided to remove yield nodes just before completing physical planning?  Seems like it could be done first thing during logical planning, which might simplify plan representation.

planner/heuristic_planner.go
jlapacik: No you're correct, but copy is for slices. Here I'm copying map keys to a slice.

planner/physical.go
jlapacik: That actually isn't necessary. The planner just has to update the roots (or results) any time a transformation produces a new root node. That logic can be contained in `Plan()`.

planner/types.go
jlapacik: Yes exactly. In fact, I was thinking about just doing that with this PR but I didn't want to complicate things. But I can do that now if you think it's best.

planner/types.go
jlapacik: True. I'm thinking ahead to when we have `LogicalPlanSpec` and `PhysicalPlanSpec`. Only the latter will have `Results`.

planner/physical.go
jlapacik: Yes you're right I should probably change this. However I do want to remove this fairly quickly and replace it with a rule.

PR #140, opened 2018-10-16
Add unit tests that exercise rules; fix bugs found along the way

wolffcm:


planner/types.go
```diff
@@ -213,6 +216,84 @@ func (e *edges) ClearSuccessors() {
 	e.successors = e.successors[0:0]
 }
 
+func (e *edges) ClearPredecessors() {
+	e.predecessors = e.predecessors[0:0]
```
jlapacik: Is there a reason not to just set this to nil?

planner/logical_test.go
```diff
@@ -125,3 +129,196 @@ func TestFluxSpecToLogicalPlan(t *testing.T) {
...
+}
+
+func mergeFilterSpecs(a, b *transformations.FilterProcedureSpec) planner.ProcedureSpec {
+	fn := a.Fn.Copy().(*semantic.FunctionExpression)
+
+	aExp, aOK := a.Fn.Body.(semantic.Expression)
+	bExp, bOK := b.Fn.Body.(semantic.Expression)
+
+	if !aOK || !bOK {
+		panic("function body not expression")
```
jlapacik: So does this mean return statements are not allowed?

planner/types.go
```diff
@@ -213,6 +216,84 @@ func (e *edges) ClearSuccessors() {
...
+	copy(newEdges.predecessors, e.predecessors)
+	copy(newEdges.successors, e.successors)
+	return *newEdges
+}
+
+// MergeLogicalPlanNodes merges top and bottom plan nodes into a new plan node, with the
+// given procedure spec.  The returned node will have its predecessors set to the predecessors
+// of "bottom", however, it's successors will not be set---it will be the responsibility of
+// the planner to attach the merged node to its successors.
+func MergeLogicalPlanNodes(top, bottom PlanNode, procSpec ProcedureSpec) PlanNode {
```
jlapacik: Can you put a visual description of this one like you did for `SwapPlanNodes` showing that it's a linear sequence of nodes? Those are really helpful, I think.

planner/types.go
wolffcm: Well, setting it to `nil` could free the memory that backs the slice.  I think this will leave the memory around, so the slice has a length of 0 but an unchanged capacity.  So the next time a predecessor is added, no allocation will be required.  (at least I think that's how it works)

planner/logical_test.go
wolffcm: Yes.  I was not thorough since this is just a unit test.  I'll add a comment to that effect.

PR #142, opened 2018-10-17
give plan nodes bounds

jlapacik:
This PR gives a `PlanNode` time bounds. Originally I had tried to make bounds plan node statistics. However, because the execution engine uses the bounds associated with each plan node, the execution engine would also have to have access to input statistics for each node. To go that direction would require more design, so instead I made bounds a property of the node itself. Thoughts?

### DIFF
--- a/planner/bounds.go
+++ b/planner/bounds.go
@@ -1,0 +1,110 @@
+package planner
+
+import "github.com/influxdata/flux/values"
+
+// EmptyBounds is a time range containing only a single point
+var EmptyBounds = &Bounds{
+	Start: values.Time(0),
+	Stop:  values.Time(0),
+}
+
+// Bounds is a range of time
+type Bounds struct {
+	Start values.Time
+	Stop  values.Time
+}
+
+// BoundsAwareProcedureSpec is any procedure
+// that modifies the time bounds of its data.
+type BoundsAwareProcedureSpec interface {
+	TimeBounds(predecessorBounds *Bounds) *Bounds
+}
+
+// ComputeBounds computes the time bounds for a
+// plan node from the bounds of its predecessors.
+func ComputeBounds(node PlanNode) error {
+	var bounds *Bounds
+
+	for _, pred := range node.Predecessors() {
+
+		if pred.Bounds() != nil && bounds == nil {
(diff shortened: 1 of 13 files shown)


planner/bounds.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if pred.Bounds() != nil && bounds == nil {
+			bounds = pred.Bounds()
+		}
+		if pred.Bounds() != nil && bounds != nil {
+			bounds = bounds.Union(pred.Bounds())
+		}
+	}
+
+	s, bdd := node.ProcedureSpec().(BoundedProcedureSpec)
```
wolffcm: I'm wondering what the behavior for the flux function `shift` should be here.  It's not bounded in the same way that `range` is, it just translates the time columns in its input by a fixed amount.

Maybe instead of `BoundedProcedureSpec` as an interface we need something like this
```go
type BoundsAwareProcedureSpec interface {
    TimeBounds(predBounds *plan.Bounds) plan.Bounds
}
```
Both `range` and `shift` could implement this, but `range` would intersect, and `shift` would shift.

The behavior you have here is also the behavior of the old planner, so if you don't want to do this now, maybe that's okay.

planner/logical.go
```diff
@@ -134,6 +146,10 @@ func createLogicalPlan(spec *flux.Spec, a Administration) (*PlanSpec, error) {
 		return nil, err
 	}
 
+	if err := v.plan.BottomUpWalk(ComputeBounds); err != nil {
```
wolffcm: I would have expected this to be in the physical planner, computed just before returning the final plan.  The reason is, there may be rules we apply that refine the time bounds on the plan nodes, for example, if we can push a `range` down through a `window`.

planner/bounds.go
jlapacik: I completely overlooked `shift`. You are right, procedures should implement their own `TimeBounds` method. This is an easy fix. I'll add it.

planner/logical.go
jlapacik: Yes you're right, I'll move this to end of planning.

wolffcm:
Looks great.  I like the tests.

PR #144, opened 2018-10-17
 Added error handling to the AddCol routine, cascaded error handling throughout repo

aanthony1243:
first phase in correcting the group by bug, to add error handling to AddCol.  More to come, but it makes sense to do these PR's in phases since they reach so many files.  

The majority of changes are just catching errors where there were no errors returned previously.


functions/transformations/data_test.go
```diff
@@ -53,21 +53,39 @@ func init() {
...
+		return
+	}
+	_, err = normalTableBuilder.AddCol(flux.ColMeta{Label: "t1", Type: flux.TString})
+	if err != nil {
+		return
+	}
+	_, err = normalTableBuilder.AddCol(flux.ColMeta{Label: "t2", Type: flux.TString})
+	if err != nil {
+		return
+	}
```
stuartcarnie: nit: you could replace all this code with `AddTableCols` function

PR #148, opened 2018-10-18
refactor: rename DataType to ColType and add a ColumnType method to semantic.Type

jsternberg:
This renames and adds a GoDoc for ColType to make it clear that the
DataType is only for column data types. It also adds a ColumnType method
to the semantic types so that it is possible to take a value and
determine the type that should be used for a column.

### DIFF
--- a/semantic/types.go
+++ b/semantic/types.go
@@ -8,6 +8,8 @@ import (
 	"sort"
 	"strconv"
 	"sync"
+
+	"github.com/influxdata/flux"
 )
 
 // Type is the representation of a Flux type.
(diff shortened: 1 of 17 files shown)


semantic/types.go
```diff
@@ -32,6 +34,10 @@ type Type interface {
 	// It panics if the type's Kind is not Array.
 	ElementType() Type
 
+	// ColumnType returns the column type for a basic data type.
+	// It returns flux.TInvalid if the Kind is not a valid column type.
+	ColumnType() flux.ColType
```
nathanielc: I would rather the semantic package not have a dependency on the flux package. The flux package represents concepts that are global to the entire Flux system and the semantic package is a very narrow package. Could you move this method to a helper function in the flux package? Perhaps where the ColType is defined?

jsternberg: I can move it to a helper function, but I do think it would be useful to somehow have this as an interface method to reduce the api that we are exposing. I would like it to be possible to take a `semantic.Type` and retrieve the column information about it, but moving it to a separate package then requires another package to know the internal implementation of the semantic package so it can do the type-switch on the `semantic.Kind` object. I'll move it anyway and we can discuss this again at some other point if it ever comes up.

PR #150, opened 2018-10-19
push range into from

### LINKED ISSUE #139: Push Range and Filter into From Source

jlapacik:
Closes #139

### DIFF
--- a/functions/inputs/from.go
+++ b/functions/inputs/from.go
@@ -35,6 +35,7 @@ func init() {
 	flux.RegisterFunction(FromKind, createFromOpSpec, fromSignature)
 	flux.RegisterOpSpec(FromKind, newFromOp)
 	plan.RegisterProcedureSpec(FromKind, newFromProcedure, FromKind)
+	plan.RegisterPhysicalRule(MergeFromRange{})
 	execute.RegisterSource(FromKind, createFromSource)
 }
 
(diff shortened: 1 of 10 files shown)


functions/inputs/from_test.go
```diff
@@ -142,3 +147,226 @@ func TestFromOpSpec_BucketsAccessed(t *testing.T) {
...
+			got := make([]fromTestAttrs, 0)
+			pp.BottomUpWalk(func(node plan.PlanNode) error {
+				got = append(got, fromTestAttrs{
+					ID:   node.ID(),
+					Spec: node.ProcedureSpec(),
+				})
+				return nil
+			})
+
+			if !cmp.Equal(want, got, semantictest.CmpOptions...) {
```
wolffcm: Is there any reason not to use helper methods in  `plan.plantest` here to compare the expected and  actual plans?

functions/inputs/from.go
```diff
@@ -128,13 +113,48 @@ type FromProcedureSpec struct {
...
 	GroupKeys   []string
 
 	AggregateSet    bool
 	AggregateMethod string
 }
 
+// FromRangeTransformationRule pushes a `range` into a `from`
+type FromRangeTransformationRule struct{}
+
+// Name returns the name of the rule
```
wolffcm: You can register this rule in the `init` method above so the planner picks it up by default.  Then we can begin testing the new planner against an actual running database.

plan/physical.go
```diff
@@ -88,7 +88,17 @@ func removeYields(plan *PlanSpec) (*PlanSpec, error) {
 			}
 
 			newRoot := root.Predecessors()[0]
-			newRoot.RemoveSuccessor(root)
+			newSucc := make([]PlanNode, 0, len(newRoot.Successors()))
+
+			for _, succ := range newRoot.Successors() {
+				newSucc = append(newSucc, succ)
+				if succ == root {
```
wolffcm: Don't you want to place this check before you add `succ` to `newSucc`?

functions/inputs/from.go
```diff
@@ -128,13 +113,48 @@ type FromProcedureSpec struct {
...
 	OrderByTime bool
-	GroupMode   GroupMode
+	GroupMode   functions.GroupMode
 	GroupKeys   []string
 
 	AggregateSet    bool
 	AggregateMethod string
 }
 
+// FromRangeTransformationRule pushes a `range` into a `from`
```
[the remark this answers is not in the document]
wolffcm: Oops, this is actually exactly what you've done.  The diff made me think this was closer to the top of the file.

functions/inputs/from.go
```diff
@@ -128,13 +113,48 @@ type FromProcedureSpec struct {
...
+	return plan.Pat(transformations.RangeKind, plan.Pat(FromKind))
+}
+
+// Rewrite attempts to rewrite a `from -> range` into a `FromRange`
+func (rule FromRangeTransformationRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool) {
+	from := node.Predecessors()[0]
+	fromSpec := from.ProcedureSpec().(*FromProcedureSpec)
+	rangeSpec := node.ProcedureSpec().(*transformations.RangeProcedureSpec)
+
+	// Range can only be pushed into from once
```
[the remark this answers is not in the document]
jlapacik: Yes I was thinking that after several transformations you might get a situation where you have two sequential range operations but you want them to have union semantics. However after thinking this through, that could never happen unless we perform an incorrect rewrite. So yes, they should be intersected.

plan/physical.go
jlapacik: This is embarrassing.

functions/inputs/from_test.go
jlapacik: So I'm not sure if we want to have one helper method for comparing an entire plan anymore. Especially for physical plans we will be adding statistics and costs, and we might want to just test for specific attributes of a plan instead of all attributes. What do you think?

Another reason I didn't use the test helper was now that `PlanNodes` have bounds, after planning the bounds will be calculated, but there's not really an easy way to construct a plan with bounds from a list of nodes and edges. In order to do that we would need to make all of the attributes of a `PlanNode` public.

So I guess what I'm trying to say is that I'm still not satisfied with testing. It's not as easy and straight forward as I want it to be.

jlapacik:
The two most recent commits make the changes that were requested from review. They also add the notion of a rule registrar to the `plan` package. Builtin functions now register their rules with the default registrars, but a planner, in order to use those rules, must specify (via planner option) to use the default registrars. This allows one to define a custom registrar for tests.

This fixes a previous problem where tests, because they were importing the `transformations` and `inputs` packages, were automatically having all builtin rules added to their rule sets. Any time a new rule is registered in any of these packages, this could potentially cause unexpected behavior in previously written tests. Explicit rule registrars make removes this behavior.

functions/inputs/from_test.go
wolffcm: I am of the opinion that rules-testing should be as concise and easy to write and to read as possible.  I think specifying the input as Flux is nice.  Specifying the expected output using Go syntax has thus far been painful.  I think you're right that we probably want each test to lock down the specific attributes that the test is concerned with, and disregard others.  Maybe a DSL would help do this.  It's worth filing an issue an talking about it as a team.

PR #159, opened 2018-10-22
Rule to push filters into `from`

wolffcm:


functions/inputs/from.go
```diff
@@ -178,9 +178,41 @@ func (rule MergeFromRangeRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool)
...
+		mergedSpec.FilterSet = true
+		mergedSpec.Filter = filterSpec.Fn.Copy().(*semantic.FunctionExpression)
+	}
+
+	merged, err := plan.MergePhysicalPlanNodes(node, node.Predecessors()[0], mergedSpec)
+	if err != nil {
+		return nil, false, err
+	}
+
+	return merged, true, nil
```
jlapacik: `Rewrite` might need to check if a `range` has been pushed down to `from`. I think storage can support tag filters without a range, but not field filters. Currently we don't know if a predicate is a tag predicate or a field predicate (aside from `_value`). So for right now, we probably can't push down unless a range has already been pushed.

semantic/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	if aOK && bOK {
+		fn.Body = &LogicalExpression{
+			Operator: ast.AndOperator,
+			Left:     aExp,
+			Right:    bExp,
+		}
+		return fn
+	}
+
+	// TODO(nathanielc): This code is unreachable while the current PushDownRule Match function is inplace.
```
jlapacik: This code is now reachable. However, storage can only support expressions as predicates. I.e no return statements and such. Since this is used in an actual rewrite rule [https://github.com/influxdata/flux/pull/159/files#diff-fa60b7ee55bcb82c7acd49cc6bca844aR204](url), I think there should be a check (in the rule) that the filter is an expression, and if not, don't perform the rewrite.

functions/inputs/from.go
```diff
@@ -178,9 +178,41 @@ func (rule MergeFromRangeRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool)
...
-	return merged, true
+
+	return merged, true, nil
+}
+
+type FromFilterMergeRule struct {
+}
+
+func (FromFilterMergeRule) Name() string {
+	return "mergeFilterFrom"
```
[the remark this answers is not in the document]
wolffcm: Yeah, I didn't follow the suggestion that I gave you.  I don't have a strong opinion other than that we should be consistent.  How about:
1. The rule's `Name` method should just return the name of the struct type.  I.e., a capitalized first letter.  (This differs from the "kinds" i.e., `fromKind` and such which don't capitalize the first letter.)
2. The struct type should follow the formula (these names all start with a verb):
 a. `MergeXxxYyyRule` for rules that merge nodes
 b. `PushXxxThroughYyyRule` for push-through rules
 c. `RemoveXxxRule` for rules that just remove nodes

Feel free to suggest any preferences you have.

functions/inputs/from.go
wolffcm: I need to explore this a bit more, especially after our recent discussion.

semantic/utils.go
wolffcm: I will at the very least make sure that we only push down filters that are simple expressions and not statements.

functions/inputs/from.go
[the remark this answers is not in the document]
jlapacik: Perfect. And I'll add `PullXxxThroughYyyRule`.

functions/inputs/from.go
```diff
@@ -154,9 +155,141 @@ func (rule MergeFromRangeRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool)
...
+	case *semantic.IntegerLiteral:
+		return true, nil
+	case *semantic.BooleanLiteral:
+		return true, nil
+	case *semantic.FloatLiteral:
+		return true, nil
+	case *semantic.RegexpLiteral:
+		return true, nil
+	case *semantic.MemberExpression:
+		if idExpr, ok := e.Object.(*semantic.IdentifierExpression); ok && idExpr.Name == "r" {
```
jlapacik: Could you modify this to capture the input argument to the filter function so that it works with functions that take an argument other that `r`. Such as `(x) => ...`.

functions/inputs/from_test.go
```diff
@@ -28,6 +30,14 @@ func fluxTime(t int64) flux.Time {
 	}
 }
 
+func makeFilterFn(exprs ...semantic.Expression) *semantic.FunctionExpression {
+	body := semantic.ExprsToConjunction(exprs...)
+	return &semantic.FunctionExpression{
+		Params: []*semantic.FunctionParam{{Key: &semantic.Identifier{Name: "r"}}},
```
jlapacik: If you made "r" an input to this function and tested with different "r" values, I think you'd find find a bug in `isPushableExpr`. This isn't necessary if you fix `isPushableExpr`.

functions/inputs/from_test.go
```diff
@@ -59,6 +69,23 @@ func TestFrom_PlannerTransformationRules(t *testing.T) {
 		from  = &inputs.FromProcedureSpec{}
 		mean  = &transformations.MeanProcedureSpec{}
 		count = &transformations.CountProcedureSpec{}
+
+		pushableExpr1 = &semantic.BinaryExpression{Operator: ast.EqualOperator,
+			Left:  &semantic.MemberExpression{Object: &semantic.IdentifierExpression{Name: "r"}, Property: "_measurement"},
+			Right: &semantic.StringLiteral{Value: "cpu"}}
+
+		pushableExpr2 = &semantic.BinaryExpression{Operator: ast.EqualOperator,
```
jlapacik: Is `"cpu" == r._value` pushable?

plan/logical_test.go
```diff
@@ -165,14 +165,19 @@ func (MergeFiltersRule) Pattern() plan.Pattern {
 			plan.Any()))
 }
 
-func (MergeFiltersRule) Rewrite(pn plan.PlanNode) (plan.PlanNode, bool) {
+func (MergeFiltersRule) Rewrite(pn plan.PlanNode) (plan.PlanNode, bool, error) {
```
jlapacik: This doesn't need to happen now, but we should probably define this with filter.

plan/logical_test.go
```diff
@@ -210,14 +215,15 @@ func (PushFilterThroughMapRule) Pattern() plan.Pattern {
 			plan.Any()))
 }
 
-func (PushFilterThroughMapRule) Rewrite(pn plan.PlanNode) (plan.PlanNode, bool) {
+func (PushFilterThroughMapRule) Rewrite(pn plan.PlanNode) (plan.PlanNode, bool, error) {
```
jlapacik: As we add more rewrite rules this file seems a little redundant to me. Maybe we don't need it anymore when we have these corresponding rules in filter.go and map.go?

semantic/utils.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// Suppose partitonFn returns true for p and r, and false for q:
+//
+//      AND           passExpr     failExpr
+//     /   \
+//    AND   r    =>     AND           q
+//   /   \             /   \
+//  p     q           p     r
+//
+func PartitionPredicates(expr Expression, partitionFn func(expression Expression) (bool, error)) (passExpr, failExpr Expression, err error) {
```
jlapacik: Filter is the only function that uses these utilities. Should they live with filter?

functions/inputs/from.go
wolffcm: Sure I can do that.  I actually thought that it had to be named `r`.  But I just tried it with a different name and it worked.

plan/logical_test.go
jlapacik: No that's fine. It makes sense.

semantic/utils.go
wolffcm: I was thinking that they would end up being used in more places eventually.  Example: when we push a filter down through a map, maybe we can push down some predicates and not others (`filter -> map` would become `filter -> map -> filter`).  Adam's interfaces for pushing things into sources would use these utilities too.

Even taking that into account, you're right that these utilities are still filter-centric.  But I think soon we'll want lots of tools for analyzing and manipulating scalar expressions, and it will be helpful if there's just one place to look rather than having things spread over various files in `transformations`.

That was my intent, anyway.  I'm open to other viewpoints on this.

plan/logical_test.go
wolffcm: Yes, I agree this test will become redundant as it is now.

functions/inputs/from_test.go
wolffcm: Actually I'm finding that filtering on `_value` causes a panic in storage (for which I will file an issue).  So I'll need to guard against it until its fixed.

jlapacik: Interesting. Does the order of an equality predicate matter?

semantic/utils.go
jlapacik: 👍

functions/inputs/from_test.go
wolffcm: Actually I'm finding that it does... if I transpose the operands such that the constant is on the left, I'll get an empty result, and the other way I get a panic.

PR #168, opened 2018-10-23
update planner docs

jlapacik:


docs/Planner.md
```diff
@@ -1,63 +1,86 @@
...
 
-## Interface
+## Plans
+--------
 
-The Planner inter is defined as:
+A Flux query plan is represented as a Directed Acyclic Graph (DAG).
+The nodes of the graph represent individual query operations to perform.
+The edges of the graph represent the flow of data from one operation to another.
+In a Flux query plan, data flows vertically from the leaves of the DAG to the roots.
```
wolffcm: It would be good to mention that we think of the direction of data flow as going upwards from sources to terminal yield nodes.  This helps contextualize terms like "push down" and "pull up".

docs/Planner.md
```diff
@@ -1,63 +1,86 @@
...
-The leaves of the plan DAG represent sources of data and the data flows from bottom up through the tree.
-Again this is inverted from the Query DAG where data flows top down.
+## Planning
+-----------
 
-## Data Frames
+There are two types of query plans - logical and physical.
+The nodes of a logical plan represent logical operations.
+The nodes of a physical plan represent physical operations.
+Logical operations define how data will be transformed.
```
wolffcm: I've heard it described this way: logical operations describe *what* happens to the data (join, sort, etc) and physical operations specify *how* the operations happen, i.e., the actual algorithms (merge join, bubble sort, etc).

docs/Planner.md
```diff
@@ -1,63 +1,86 @@
...
+Physical operations define how data transformations are implemented.
 
-Data frames are a set of data and their lineage is known.
-Meaning it is known what parent data frames and operations are needed to construct the data frame.
-Using this concept of lineage allows a data frame to be reconstructed if it is loss due to node failure or if its parent data frames are modified.
+A Flux query spec is translated directly into a logical plan.
```
wolffcm: Would be good to talk about the high-level flow here a bit, like `flux.Spec -> logical plan -> physical plan`.

docs/Planner.md
```diff
@@ -1,63 +1,86 @@
...
 
-This distinction is necessary to precisely define the lineage of a data frame.
+	// Rewrite an operation into an equivalent one
+	Rewrite(PlanNode) (PlanNode, bool)
+}
+```
 
+A rewrite rule defines an equivalence class of queries.
+That is, for any data instance, a query expressed before and after a rewrite rule is applied will always define the same data transformation.
+A rule must match a specified pattern in order to rewrite an operation.
```
wolffcm: An example would be nice here.

I think we'll want a separate document that goes into more depth on writing rules, since open source contributors may want to write them.

docs/Planner.md
```diff
@@ -1,63 +1,86 @@
...
+Physical operations define how data transformations are implemented.
 
-Data frames are a set of data and their lineage is known.
-Meaning it is known what parent data frames and operations are needed to construct the data frame.
-Using this concept of lineage allows a data frame to be reconstructed if it is loss due to node failure or if its parent data frames are modified.
+A Flux query spec is translated directly into a logical plan.
 
-### Windowing
+## Optimization
+---------------
```
wolffcm: Planning and optimization are basically synonyms, I think... consider merging these sections?

docs/Planner.md
wolffcm: Perhaps include the interfaces for `LogicalPlanner` and `PhysicalPlanner`?

PR #171, opened 2018-10-23
feat: array builder interface for constructing an array

jsternberg:
The array builder interfaces roughly matches the arrow interface and is
a reasonable beginning point for constructing tables in a columnar
method.

### DIFF
--- a/array/bool.go
+++ b/array/bool.go
@@ -6,3 +6,13 @@ type Boolean interface {
 	Value(i int) bool
 	BooleanSlice(start, stop int) Boolean
 }
+
+// BooleanBuilder represents an abstraction over building a bool array.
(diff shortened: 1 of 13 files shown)


array/bool.go
```diff
@@ -6,3 +6,14 @@ type Boolean interface {
 	Value(i int) bool
 	BooleanSlice(start, stop int) Boolean
 }
+
+// BooleanBuilder represents an abstraction over building a bool array.
+type BooleanBuilder interface {
+	BaseBuilder
+	Append(v bool)
+	AppendNull()
```
stuartcarnie: nit: you could move `AppendNull` to the `BaseBuilder` interface, as it is not specific

PR #172, opened 2018-10-23
refactor(execute): update column reader to use the array interfaces

jsternberg:
This modifies `execute.ColReader` to use the interfaces in the `array` package. This will make it easier to integrate arrow with the query engine when we make an array implementation of it.

Some of the array interfaces in arrow do not expose the underlying array to the data, but they are used in some of the functions. For those, they will do a typecast directly on the staticarray implementation to access the data. These will need to be removed before we change to arrow implementations.


jsternberg:
This is no longer relevant.

PR #174, opened 2018-10-24
feat: add a new table builder that builds on the array package

jsternberg:
The new table builder interface focuses on simplicity and ensuring that
column-based table construction is prioritized over other methods. We will
build other interfaces for building tables on top of this, but the underlying
data is column based so the table builder interface should be too.

It builds on the array package which, itself, builds on the arrow interface.

### DIFF
--- a/execute/executetest/transformation.go
+++ b/execute/executetest/transformation.go
@@ -19,11 +19,24 @@ func ProcessTestHelper(
 ) {
 	t.Helper()
 
-	d := NewDataset(RandomDatasetID())
-	c := execute.NewTableBuilderCache(UnlimitedAllocator)
-	c.SetTriggerSpec(execute.DefaultTriggerSpec)
+	ProcessTest(t, data, want, wantErr, func() (execute.DataCache, execute.Transformation) {
+		d := NewDataset(RandomDatasetID())
+		c := execute.NewTableBuilderCache(UnlimitedAllocator)
+		c.SetTriggerSpec(execute.DefaultTriggerSpec)
+		return c, create(d, c)
+	})
+}
 
-	tx := create(d, c)
+func ProcessTest(
+	t *testing.T,
+	data []flux.Table,
+	want []*Table,
+	wantErr error,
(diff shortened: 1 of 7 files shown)


influxql/decoder.go
```diff
@@ -0,0 +1,188 @@
+package influxql
```
aanthony1243: I don't think we want something so specifically `influxql`  in `flux`.   This code is probably a better fit in  functions/inputs  as a source decoder.  It would be OK to add a sub-package (even an internal on?) there to keep this file intact, and then write the source implementation using this as a helper package.

influxql/decoder.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return &resultIterator{resp: &resp, a: dec.a}, nil
+}
+
+type resultIterator struct {
+	resp *Response
+	a    *execute.Allocator
+}
+
+func (ri *resultIterator) More() bool {
```
aanthony1243: The SourceDecoder interface is similar to what you've done here, but instead defines  `Connect/Fetch/Decode` functions.  We had considered implementing more of an iterator pattern like this but decided to keep it simple.  What do you think?

influxql/decoder.go
aanthony1243: also:  if you really want to implement this as a 2.0 style decoder, we probably need to keep this on `Platform` not `flux`

influxql/decoder.go
```diff
@@ -0,0 +1,188 @@
...
+				// Skip the time column if one was found.
+				continue
+			}
+
+			// Construct a new table.
+			b := tablebuilder.New(r.a)
+
+			// If we have a time column, do it here.
+			if timeCol >= 0 {
+				if err := b.Times("_time").Do(func(b array.TimeBuilder) error {
```
aanthony1243: This seems to be exposing the data at too low-level.  

A function author should not be thinking "I need to remember to Reserve space for my column".  This seems like a long way around compared to the existing method: `builder.AppendTime(timeIndex, values.Time(t.UnixNano))`

PR #179, opened 2018-10-25
feat: staticarray supports using the allocator for tracking memory usage

jsternberg:
The staticarray package did not use the allocator so it would not track
memory usage. The table builder will track usage so this adds that
functionality to these utilities so the two are equivalent.

### DIFF
--- a/array/float.go
+++ b/array/float.go
@@ -1,14 +1,23 @@
 package array
 
-// Float represents an abstraction over a float array.
-type Float interface {
-	Base
+// FloatRef is a reference to a Float.
+type FloatRef interface {
+	BaseRef
 	Value(i int) float64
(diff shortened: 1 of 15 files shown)


array/float.go
```diff
@@ -1,16 +1,25 @@
...
 	Value(i int) float64
-	FloatSlice(start, stop int) Float
+	FloatSlice(start, stop int) FloatRef
 
 	// Float64Values will return the underlying slice for the Float array. It is the size
 	// of the array and null values will be present, but the data at null indexes will be invalid.
 	Float64Values() []float64
 }
 
+// FloatRef represents an abstraction over a float array.
```
jlapacik: Float

PR #180, opened 2018-10-25
feat: Add aggregateWindow helper function

nathanielc:
Also fixes issue with window that did not clip infinite bounds to the
window bounds.


functions/transformations/window.go
```diff
@@ -415,9 +418,7 @@ func (t *fixedWindowTransformation) clipBounds(bnds *execute.Bounds) {
 
 func (t *fixedWindowTransformation) getWindowBounds(now execute.Time) []execute.Bounds {
 	if t.w.Every == infinityVar.Duration() {
-		return []execute.Bounds{
-			{Start: execute.MinTime, Stop: execute.MaxTime},
-		}
+		return []execute.Bounds{t.bounds}
```
aanthony1243: do we need to be robust against sources that may not set bounds?  Not sure what the new planner will do about that, do we need to be robust against it here?

nathanielc: No, the planner ensures the all window functions are bounded.

nathanielc: Or, rather, the planner tracks the bounds of all transformations, and the window function during creation will error if there are no bounds.

PR #181, opened 2018-10-25
fix: to_http will return any errors that happen during execution

jsternberg:
The `to_http` function did not report some errors during execution.

This commit also adds a utility package for wait groups that involve
errors. You can launch a goroutine on a function that returns an error
and it will return the first error that happens when invoking wait. In
the normal standard library, this functionality can be difficult to
achieve because it then has to rely heavily on closures.

### DIFF
--- a/functions/outputs/to_http.go
+++ b/functions/outputs/to_http.go
@@ -11,11 +11,11 @@ import (
 	"runtime"
 	"sort"
 	"strings"
-	"sync"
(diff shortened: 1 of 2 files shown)


jsternberg:
There's more, but this one had a confusing enough structure that I wanted to also make sure the `syncutil` was cleared separately. I'll be having another PR for the bulk of the functions package.

PR #182, opened 2018-10-25
fix: simplify and fix a problem in the keys function

jsternberg:
The `keys` function would sort the table columns. The problem is it
would return the column information from the table and it would modify
the mutable copy of the table. This changes things so it copies the
information out before it iterates over the columns.

Likely, it would be better if this function did not sort the data.
There's a sort function for that and it would likely be much easier if
this returned the columns in order rather than as sorted.

This also simplifies the use of except. A small map will not cause any
performance issues and it clarifies the intent so it is much easier to
follow and verify the logic. In cases without except, no additional
memory is allocated for the map.

### DIFF
--- a/functions/transformations/keys.go
+++ b/functions/transformations/keys.go
@@ -3,7 +3,6 @@ package transformations
 import (
 	"fmt"
 	"sort"
-	"strings"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
@@ -139,16 +138,10 @@ type keysTransformation struct {
 }
 
 func NewKeysTransformation(d execute.Dataset, cache execute.TableBuilderCache, spec *KeysProcedureSpec) *keysTransformation {
-	var except []string
-	if len(spec.Except) > 0 {
-		except = append([]string{}, spec.Except...)
-		sort.Strings(except)
-	}
-
 	return &keysTransformation{
 		d:      d,
 		cache:  cache,
-		except: except,
+		except: spec.Except,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


functions/transformations/keys.go
```diff
@@ -162,42 +161,48 @@ func (t *keysTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
-	// add remaining
-	for ; i < len(cols); i++ {
-		execute.AppendKeyValues(tbl.Key(), builder)
-		builder.AppendString(colIdx, cols[i].Label)
+	// Append the keys to the column index.
+	if err := builder.AppendStrings(colIdx, keys); err != nil {
+		return err
 	}
 
 	// TODO: this is a hack
```
wolffcm: It's unrelated to your change, but I guess this "hack" at the end has to happen because the execution framework expects the input table to actually be read?

jsternberg:
The `keys` function is handled separately in #182 because I noticed some other problems with it that I wanted reviewed separately.

functions/transformations/keys.go
```diff
@@ -162,42 +161,48 @@ func (t *keysTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 		return fmt.Errorf("keys found duplicate table with key: %v", tbl.Key())
 	}
```
wolffcm: I guess it's no longer necessary to sort the `except` array when creating the transformation above, here:
https://github.com/influxdata/flux/blob/3fc3eeaa76952f5d732ab17089d5de5249142fbe/functions/transformations/keys.go#L144

functions/transformations/keys.go
jsternberg: Yes. I think we should probably create an issue for this.

functions/transformations/keys.go
jsternberg: Good catch. I'll update that.

PR #184, opened 2018-10-25
fix: handle errors everywhere in functions and things that use the table builder

### LINKED ISSUE #135: Chore: clean up uncaught errors in `flux` functions.
Many of our functions fail to catch errors returned by `tbl.Do(...)`.  There are other cases as well.

jsternberg:
This fixes things to properly read and handle the errors that are
returned from `TableBuilder` and the `table.Do()` function.

This commit does not change things to handle errors in the entire
project so `errcheck` still reports existing unhandled errors, but it
handles all of the ones related to those mentioned above.

Fixes #135.


jsternberg:
The `keys` function is handled separately in #182 because I noticed some other problems with it that I wanted reviewed separately.

nathanielc:
LGTM

PR #188, opened 2018-10-26
merge new planner into master

jlapacik:


control/controller.go
```diff
@@ -318,6 +319,7 @@ func (c *Controller) processQuery(q *Query) (pop bool, err error) {
 			return true, errors.New("failed to transition query into executing state")
 		}
 		q.alloc = new(execute.Allocator)
+		// TODO: pass the plan to the executor here
```
wolffcm: This comment can go away.

functions/inputs/from.go
```diff
@@ -86,13 +91,293 @@ type FromProcedureSpec struct {
...
+
+// Pattern returns the pattern that matches `from -> range`
+func (rule MergeFromRangeRule) Pattern() plan.Pattern {
+	return plan.Pat(transformations.RangeKind, plan.Pat(FromKind))
+}
+
+// Rewrite attempts to rewrite a `from -> range` into a `FromRange`
+func (rule MergeFromRangeRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool, error) {
+	from := node.Predecessors()[0]
+	fromSpec := from.ProcedureSpec().(*FromProcedureSpec)
```
jsternberg: It would be nice if we did matching rules on the actual types and the kind was only used for JSON encoding. I always feel a bit uncomfortable with encoding the kind into logic like this, but it's everywhere else so this comment is by no means a suggestion to change this.

I would like to revisit it in the future though.

jsternberg:
Even if the commits are messy, this is likely best to merge without attempting any rebase or squashing any commits.

docs/Planner.md
```diff
@@ -1,63 +1,123 @@
...
-The leaves of the plan DAG represent sources of data and the data flows from bottom up through the tree.
-Again this is inverted from the Query DAG where data flows top down.
+```go
+type LogicalPlanner interface {
+	Plan(*flux.Spec) (*PlanSpec, error)
+}
 
-## Data Frames
+type PhysicalPlanner interface {
+	Plan(*PlanSpec) (*PlanSpec, error)
```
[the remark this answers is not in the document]
wolffcm: Practically, today it's not possible to use just the logical planner and skip physical planning, because a each `from` needs a `range` to be pushed into it, and that push down is done as a physical optimization.

Some more depth: right now all our procedure specs are both logical procedure specs and physical procedure specs. That is, they all have a `Cost` method and can be translated directly into an instance of `Transformation` or `Source`.  But that won't always be the case; at some point we'll have (for example) a `JoinProcedureSpec` which is logical (no `Cost` method and can't be converted to a `Transformation`) and the physical planner will need to choose either `MergeJoinProcedureSpec` or `NestedLoopProcedureSpec` for the physical plan.

(I need to add the above to the docs.)

I think you're right that the logical planner should return something more opaque.  Josh and I have talked about breaking up `PlanSpec` into logical and physical counterparts.

As to why the logical/physical divide is exposed on the interface level, that's a good question.  Honestly, I was following the example of what was there in the `control` package already.  I can think of good reasons for the divide, like maybe wanting to cache logical plans to use later on, but that seems to be a ways off.

functions/inputs/from.go
wolffcm: I'm open to using something else.  Ultimately, we want our patterns to be able to not just recognize a procedure of a particular type, but some class of procedures within a taxonomy (like any source that can have a filter pushed into it, any aggregate function, etc).

docs/Planner.md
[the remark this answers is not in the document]
jsternberg: 👍 I like the answer and I'm really looking forward to a further API-level differentiation between the two. I can easily see now why the difference would be needed, but we do need to have them as separate APIs. The influxql query engine suffered for a long time because the outward facing  APIs made a lot of assumptions about the previous steps that were invoked without having them encoded as part of the type system so it's important that we do that before having a stable API.

Issue #192, opened 2018-10-29
toHTTP function seems to never execute


mark-rushakoff:
I have a query that returns results:

```
'from(bucket:"demo-bucket-in-1") |> range(start:-1h) |> last()'
Result: _result
Table: keys: [_start, _stop, _field, _measurement]
                   _start:time                      _stop:time           _field:string     _measurement:string                      _time:time                  _value:float
------------------------------  ------------------------------  ----------------------  ----------------------  ------------------------------  ----------------------------
2018-10-29T18:36:12.028749000Z  2018-10-29T19:36:12.028749000Z                       n                 counter  2018-10-29T19:15:50.616071000Z                          1148
```

But when I use `toHTTP`, no HTTP requests are made. 

```
'from(bucket:"demo-bucket-in-1") |> range(start:-1h) |> toHTTP(url:"http://localhost:12345")'
```

I'm trying to send the result to a simple HTTP server that just prints out whatever requests arrive.

Through printf debugging, I've determined that `(*ToHTTPTransformation).Process` is never being called. I don't understand enough of what else is going on in that part of the code, to make much further progress without doing a lot more learning first.

I'm using commit `39f39ce9135326d126c224b8b2b6c8f72a5600d6`, which is what is currently on platform master. I haven't tried bisecting other commits to find if there was some point where it did work.

jlapacik:
@mark-rushakoff how are you running the above query? Through the repl? What happens if you put a `yield()` on the end of the query like so:

    from(bucket:"demo-bucket-in-1")
        |> range(start:-1h)
        |> toHTTP(url:"http://localhost:12345")
        |> yield()

mark-rushakoff:
Using `influx query`, with influx build from platform master today.  `./influx  query -t ... --org-id 02dc43328ac01000 'from(bucket:"demo-bucket-in-1") |> range(start:-1h) |> toHTTP(url:"http://localhost:12345")'`.

Putting `|> yield()` on the end of that query appears to fix the issue so that the HTTP request is executed. But it's surprising to me that I would need to do that.

jlapacik:
This is an issue with the Flux REPL. The REPL always looks for a call to `yield()` in order to execute a query:

https://github.com/influxdata/flux/blob/340b9f699992434e350508b98d756bd3a14ab9e7/repl/repl.go#L182

This is unnecessary since all queries will go through planning where any implicit yields will be made explicit. I will file an issue to modify this check by the REPL and link here.

Issue #50, opened 2018-09-24
Improve REPL interface


nathanielc:
Some feedback we received about the REPL:

Autocomplete uses TAB instead of ENTER, this UX causes users to frequently submit partially typed queries.

The only way to quit the REPL is to send EOF (Ctrl-D), this was confusing to most users, perhaps using SIGINT (Ctrl-C) when the input line is empty should quit. We want to preserve SIGINT for canceling running queries so making it also quit will require understanding that state.

There is no way to input multiple lines. Since the Flux language naturally prefers multiple lines we need a convenient way to input multiple lines. Perhaps a SHIFT-ENTER submits the query, and a normal ENTER just adds a new line? This should also play nice with `yield`.
 
The hidden behavior of  `run` was confusing, some feedback was to change the behavior of the REPL and execute queries if there was no variable assignment. This way `yield` remains implicit in most cases. This would remove the need to have a `run` method as simply entering the identifier of a saved query would run it. If you are building up a query then use variable assignment or multi-line input methods. 

We should probably use a real readline backend for the REPL so both vim and emacs bindings can exist as is expected by users.

wolffcm:
I recently made some changes to how we propagate errors from the Flux controller.  As a result the error formatting in the platform REPL is a little more wonky than before, unfortunately:
```
> from(bucket: "telegraf")
Error: <invalid> failed to create physical plan: result 'from0' is unbounded. Add a 'range' call to bound the query
```
The `<invalid>` in the message above is new, and is how `platform.Error` messages are formatted.  "Invalid" reflects that a 400 error code was returned by the query service, which is now returning `platform.Error` when queries fail.

Since the REPL code lives in the Flux repo and cannot depend on platform we'll need to figure out the best way to make error messages pretty when they are returned from a remote server.

RexBarker:
Need to improve the documentation page for the REPL usage.  I found the exit method of cntrl-D by accident by using my Home Simpson method of mashing the keyboard

PR #197, opened 2018-10-29
repl does not check for yield nodes

### LINKED ISSUE #194: No need for REPL to check for yields
Originates from https://github.com/influxdata/flux/issues/192.

Currently the flux REPL checks that a query has an explicit call to `yield()` before executing it.

https://github.com/influxdata/flux/blob/340b9f699992434e350508b98d756bd3a14ab9e7/repl/repl.go#L182

This is unnecessary as all queries passed to the REPL will be planned before they are executed and the planning phase explicitly generates any implied calls to `yield()`. Therefore we should remove this check from the REPL which will allow users to run queries without having to place a `yield()` at the end of every one.

jlapacik:
Fixes #194 #192 .
Partially addresses #50 .

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -16,7 +16,6 @@ import (
 	prompt "github.com/c-bata/go-prompt"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
-	"github.com/influxdata/flux/functions/transformations"
 	"github.com/influxdata/flux/interpreter"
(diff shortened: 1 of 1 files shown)


jlapacik:
@nathanielc only values returned by the interpreter are executed. Variable declarations are **not** returned by the interpreter. Expression statements are evaluated and the resulting values **are** returned. If this value is a table object, it is passed to the planner and executed.

PR #198, opened 2018-10-30
fix: remove a cyclical import that got added as part of a bad merge

jsternberg:
A combination of an error refactor (that also fixed import orderings in
some files) and the planner change resulted in an accidental cyclical
import.


jsternberg:
This was already done last night and I didn't see it.

PR #200, opened 2018-10-30
feat: add an influxql json decoder as a library

### LINKED ISSUE #138: Create InfluxDB 1.x JSON source/decoder

jsternberg:
The json decoder will decode the influxql json format and allow it to be
used as a result decoder. This will make it easy to integrate it as an
alternative source method similar to `fromCSV`.

Fixes #138.

### DIFF
--- a/influxql/decoder.go
+++ b/influxql/decoder.go
@@ -1,0 +1,195 @@
+package influxql
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"sort"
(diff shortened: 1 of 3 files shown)


jlapacik:
Is it possible to go the reverse direction? From Flux table to influxql Result?

jsternberg:
It is, but it is in the platform repository as part of the transpiler interface there. It also has slightly different logic because the two have different structures.

PR #201, opened 2018-10-30
generated csv test cases

jlapacik:
Generated queries and test data covering the following:

* predicates with lots of ANDs/ORs
* grouping/windowing with aggregates over generated float series

**Note:** The csv input data is duplicated for a lot of the queries. Would like a way for multiple queries to use the same csv input file.


jlapacik:
I'm closing this PR. These test cases belong in platform and will use the `fromJSON` function #210 .

PR #203, opened 2018-10-31
Add Type Inference support

nathanielc:
Type inference is ready.
There are a few TODOs:

- [x] The complete package API needs to change
- [ ] I think we should merge the Type and PolyType interfaces, but that should come as a later change. See #208 

Depending on how much the complete API needs to change and how easy it is for Chronograf to update we can merge this sooner or later.


semantic/poly_types.go
```diff
@@ -0,0 +1,764 @@
...
+)
+
+type TypeExpression interface {
+	FreeVars(*Constraints) TvarSet
+	Type(map[Tvar]KindConstraint) (Type, error)
+	MonoType() (Type, bool)
+	resolvePolyType(map[Tvar]KindConstraint) (PolyType, error)
+}
+
+type PolyType interface {
```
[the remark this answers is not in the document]
nathanielc: Yes, I have just pushed up a new commit that adds docs, it also update the type_inference.md file to explain how type inference works.

compiler/runtime.go
```diff
@@ -7,8 +7,25 @@ import (
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
+	"github.com/pkg/errors"
 )
 
+type Func interface {
```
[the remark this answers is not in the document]
nathanielc: This change does not modify the Func interface except for where types were concerned. 

I am not sure I follow the concern about type casting, the point of this interface is so that you do not have to type cast at all. The `compiler` follows the types through and handles concrete types during the flow of data where it can. The `scope` being the obvious exception.

compiler/runtime.go
```diff
@@ -24,138 +41,137 @@ type Evaluator interface {
...
-func (c compiledFn) Eval(scope Scope) (values.Value, error) {
-	if err := c.validate(scope); err != nil {
+func (c compiledFn) Eval(input values.Object) (values.Value, error) {
+	if err := c.buildScope(input); err != nil {
 		return nil, err
 	}
 	switch c.Type().Kind() {
 	case semantic.String:
-		return values.NewString(c.root.EvalString(scope)), nil
+		return values.NewString(c.root.EvalString(c.inputScope)), nil
```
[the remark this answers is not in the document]
nathanielc: This code is the definition of `Eval` I am not sure which `Eval` you are referring to.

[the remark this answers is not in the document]
jsternberg: I mean that it should just call `c.root.Eval` instead of doing a type switch. They appear to do the same thing and reducing the number of type switches seems good to me.

compiler/runtime.go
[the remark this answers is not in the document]
jsternberg: Ah, I did not see this was moved code.

My comment still applies, but in a more general way and isn't directly related to this PR. I thought this was a new interface.

semantic/type_inference.md
```diff
@@ -38,7 +38,7 @@ The following rules will be used to solve for the types in the above program:
 #### Algorithm
 --------------
 
-The type inference algorithm that Flux uses is based on algorithm W.
+The type inference algorithm that Flux uses is based on Wand's algorithm and .
```
jlapacik: and algorithm W?

semantic/poly_types.go
```diff
@@ -0,0 +1,778 @@
...
+		}
+	}
+	return fvs
+}
+func (l Tvar) unifyType(kinds map[Tvar]Kind, r PolyType) (Substitution, error) {
+	switch r := r.(type) {
+	case Tvar:
+		if l == r {
+			return nil, nil
+		}
```
zeebo: Maybe this happens somewhere else, but it looks like it's missing an occurs check. How does `(x) => x(x:x)` type check?

semantic/poly_types.go
```diff
@@ -0,0 +1,778 @@
...
+		typR, ok := r.properties[f]
+		if !ok {
+			continue
+		}
+		s, err := unifyTypes(kinds, typL, typR)
+		if err != nil {
+			properties[f] = invalid{}
+		}
+		subst.Merge(s)
+		properties[f] = subst.ApplyType(typL)
```
zeebo: Does this actually handle `invalid` correctly? Are these two statements supposed to be in an else, or does the returned substitution actually map `typL` to `invalid` if there's an error?

semantic/poly_types.go
nathanielc: The occurs check is here https://github.com/influxdata/flux/pull/203/files#diff-594e3760d97b20e5e28945e9d3c70f19R165

That if condition is not about occurs rather it  ignores any unification of a type variable with itself.

semantic/poly_types.go
nathanielc: Ah good catch, these are supposed to be in an else. I'll add a test case for the type error condition.

semantic/poly_types.go
zeebo: Of course the occurs check doesn't matter when unifying two variables. 👍

PR #210, opened 2018-11-01
add fromJSON source for reading influxql JSON

jlapacik:
`fromJSON` is a source function for reading influxql JSON text. It assumes a singe influxql result. If more that one result is present in the JSON, it will throw an error. The execution of the source is not tested here but will be tested heavily using the transpiler in the platform repo. **Note** it is my intention that this function be used internally for generated testing (similar to `fromCSV`). I don't see it being used externally.

### DIFF
--- a/functions/inputs/from_json.go
+++ b/functions/inputs/from_json.go
@@ -1,0 +1,193 @@
+package inputs
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/influxql"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/pkg/errors"
+)
+
+const FromJSONKind = "fromJSON"
+const bufferSize = 8192
+
+func init() {
+	fromJSONSignature := semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"json": semantic.String,
+			"file": semantic.String,
+		},
+		Required: nil,
+		Return:   flux.TableObjectType,
+	}
+	flux.RegisterFunction(FromJSONKind, createFromJSONOpSpec, fromJSONSignature)
+	flux.RegisterOpSpec(FromJSONKind, newFromJSONOp)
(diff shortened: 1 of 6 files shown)


functions/inputs/from_json.go
```diff
@@ -0,0 +1,193 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/influxql"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/pkg/errors"
+)
+
+const FromJSONKind = "fromJSON"
```
aanthony1243: I'd be concerned that `fromJSON` will be perceived as a general tool of some sort.   namespacing would be the best solution to this problem, but I wonder if we should have this be   `fromInfluxJSON` for now to indicate a fixed JSON structure.

functions/inputs/from_json.go
```diff
@@ -0,0 +1,193 @@
...
+	results flux.ResultIterator
+	id      execute.DatasetID
+	ts      []execute.Transformation
+}
+
+func (c *JSONSource) AddTransformation(t execute.Transformation) {
+	c.ts = append(c.ts, t)
+}
+
+func (c *JSONSource) Run(ctx context.Context) {
```
aanthony1243: This is all technically correct, but can we re-tool this to use the SourceDecoder type?  This will be another good example of using to encourage adoption of it:  https://github.com/influxdata/flux/blob/ab9815acfd8e9d429835a39b3e6e8d02cfd7c238/functions/inputs/source_iterator.go#L21

I think @jsternberg had some resistance to implementing that interface and instead went for a pure iterator-type interface for his decoder.  I'd be willing to have a discussion about updating the SourceDecoder to be a more traditional iterator type with  `More()` / `Next()` instead of `fetch`/`decode`

functions/inputs/from_json.go
jlapacik: I was wondering that myself. I'll change it.

functions/inputs/from_json.go
jlapacik: @aanthony1243 Could this be done in a follow on PR, since this is more of an internal tool for testing rather than a general one?

aanthony1243: yeah, we can drop this point for now.

PR #215, opened 2018-11-02
chore: fix current megacheck errors

mark-rushakoff:
I spent some time trying to get megacheck into CI as well, but I ran
into continual problems, and I'm not able to spend any more time on it.

The current version of megacheck is not module aware. I tried a few
different approaches of copying the source to a new, traditional GOPATH
and running megacheck there, but something went wrong every time I did
it.

Then I found about about go-tools' `next` branch, which is module aware.
However, that version runs comment style checkers which aren't present
in megacheck, and I couldn't figure out how to disable that check.

So this commit only fixes the current megacheck errors without touching
the Circle config.


functions/transformations/filter_test.go
```diff
@@ -346,7 +346,7 @@ func TestFilter_NewQuery(t *testing.T) {
 			Name: "from with database filter with no parens including regex and field",
 			Raw: `from(bucket:"mybucket")
 						|> filter(fn: (r) =>
-							r["t1"]==/val1/
+							r["t1"]==/^val1/
```
nathanielc: Did megacheck complain about this? That seems odd since this is an opaque string literal according to Go.

parser/parser_test.go
```diff
@@ -198,7 +198,7 @@ func TestParse(t *testing.T) {
 		},
 		{
 			name: "regex match operators",
-			raw:  `"a" =~ /.*/ and "b" !~ /c/`,
+			raw:  `"a" =~ /.*/ and "b" !~ /c$/`,
```
nathanielc: Again this seems odd that megacheck cares

mark-rushakoff:
The complaints were that the assertions of `regexp.MustCompile("val")` didn't contain any metacharacters. I think the goal of the warning was that, if this was a regex you were actually calling from your code, you should prefer `strings.Contains` instead for simplicity. 

I arbitrarily chose to update the flux source to contain a metacharacter to suppress that warning. At the time, that seemed better than comment-suppressing the warning.

The options are:
- Leave the changes as-is
- Restore the flux code in test to its original state and line-comment-suppress each use of `regexp.MustCompile("val")`
- Restore the flux code in test to its original state, and file-comment-suppress the "no metacharacters in regex" warning

Let me know which option you prefer.

PR #228, opened 2018-11-03
feat: initial work on a scanner using ragel

jsternberg:
The current parser is a mix of the lexing and parsing components and it
is decently slow and doesn't follow the spec completely. This is the
initial process of creating a well tested scanner for flux.

It uses ragel for the finite state machine generation. It supports
lexing the entirety of the language. It includes two methods for
scanning the input because, in the grammar, regular expressions and the
division operator will conflict with each other. The `ScanNoRegex`
method is intended to be used in contexts where we are not expecting a
regular expression (such as in an expression where the next token should
be an operator).

This package is placed into internal so that it is not used and it can
be gradually modified if needed while developing the parser. It may be
moved to a non-internal package in the future.


internal/scanner/scanner.rl
```diff
@@ -0,0 +1,122 @@
...
+	escaped_char = "\\" ( "n" | "r" | "t" | "\\" | '"' );
+	unicode_value = any_count_line | escaped_char;
+	byte_value = "\\x" xdigit{2};
+	string_lit = '"' ( unicode_value | byte_value )* '"';
+
+	regex_escaped_char = "\\" ( "/" | "\\");
+	regex_lit = "/" ( any_count_line | byte_value | regex_escaped_char )+ "/";
+
+	# The newline is optional so that a comment at the end of a file is considered valid.
+	single_line_comment = "//" [^\n]* newline?;
```
aanthony1243: this is all worth it right here.  current parser won't do this.

jsternberg:
I made a small modification to this so that an EOF token prevents the scanner from going backwards. In the future, I want to extend this so that data can be added to allow for interactive command line use and the EOF token along with `Unread` are going to be important for that since these two are different lexically: `opt` and `opt `. The second has whitespace delimiting that the token is complete while the first could be the beginning of `option`.

internal/scanner/scanner_test.go
```diff
@@ -0,0 +1,308 @@
...
+	}{
+		{
+			name: "variable declaration",
+			s:    `x = 5`,
+			want: []token.Token{
+				token.IDENT,
+				token.ASSIGN,
+				token.INT,
+			},
+		},
```
nathanielc: Can you add one longer test case here? Something that looks like a query? That way we have some coverage over a typical query. May just a `from |> range |> filter |> count |> yield` kind of query.

I would also like to see a test case that has an ILLEGAL token somewhere in the middle. Something as simple as `x = 5 @ y = 1` to show that the scanner can continue scanner after seeing an ILLEGAL token. There is a test case below for multi byte illegal tokens, this case is about being able to continue after seeing an illegal token.

internal/scanner/scanner.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"unicode/utf8"
+
+	"github.com/influxdata/flux/internal/token"
+)
+
+//go:generate ruby unicode2ragel.rb -e utf8 -o unicode.rl
+//go:generate ragel -I. -Z scanner.rl -o scanner.gen.go
+//go:generate sh -c "go fmt scanner.gen.go > /dev/null"
```
nathanielc: These generate commands now mean that both ragel and ruby are dependencies of the Flux repo. Do we have that documented anywhere? Is there a specific version of ragel that we need to use?

jsternberg: I was planning to rewrite the ruby one in the future to use go. There's no specific version of ragel that I'm aware of. I don't think they have had a release for a long time.

I'll add a note that they're dependencies until I get to making them optional.

dgryski: Ragel 6 is still the stable supported release.  Ragel 7 is in development (with a slightly broken Go backend.)

internal/scanner/scanner_test.go
jsternberg: I have added these tests.

PR #230, opened 2018-11-05
Add source & table stats.

benbjohnson:
This pull request adds stats for `Source` & `Table` so they can be propagated up from lower level storage. These stats are required for reporting values & bytes scanned.

Required by https://github.com/influxdata/platform/pull/1265

This is my first PR to `flux` so feedback definitely welcome.


benbjohnson:
@aanthony1243 I refactored this PR so it merges the scanned values & scanned bytes stats into the existing `flux.Statistics` struct. Can you recheck it to make sure it's still 👍 ?

PR #241, opened 2018-11-05
fix(control): fix when a query is requeued multiple times

### LINKED ISSUE #239: Failed to transition query into requeueing state
This error is occurring in the query log.

jsternberg:
When a query is requeued multiple times, it will return an error on the
second attempt. This is because it tries to transition into the
requeueing state, but it's already in that state so there is nowhere to
go and it thinks a fatal error has happened.

This fix patches things so if it is already in the requeueing state it
does nothing.

Fixes #239.

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -615,6 +615,9 @@ func (q *Query) tryRequeue() bool {
 
 		q.state = Requeueing
 		return true
+	} else if q.state == Requeueing {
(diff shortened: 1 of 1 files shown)


jsternberg:
The controller is pretty opaque for testing. It's one of the reasons we want to refactor it at some point.

PR #242, opened 2018-11-06
fix: range should have the default logic for stop when creating the procedure

### LINKED ISSUE #238: Failed to create logical plan: cannot pass zero time to 'range'
This error is occurring in the query log.

The error means that one of the time parameters to range is the zero time. 

Maybe update the error message to be more clear. It was misunderstood as the range being a zero width range.

jsternberg:
The `range()` function would fill in the default value for `stop`, which
was `now()` as the current time within the `createRangeOpSpec` function.
This function is only called by those go who through the compiler so the
defaults never get filled in when directly submitting the spec.

This moves that logic to when we create the procedure node so that the
spec gets the same default value.

Fixes #238.

### DIFF
--- a/functions/transformations/range.go
+++ b/functions/transformations/range.go
@@ -57,9 +57,6 @@ func createRangeOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operat
 		return nil, err
 	} else if ok {
 		spec.Stop = stop
-	} else {
-		// Make stop time implicit "now"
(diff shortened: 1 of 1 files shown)


jsternberg:
I'm going to close this for now. I'm not sure it's the issue and it might be a change with a larger surface-area than I anticipated. The influxql compiler fills in a stop time so I'm going to investigate that.

Issue #216, opened 2018-11-02
"by" parameter in highestMax, highestAverage, etc., not working


sanderson:
When I pass an array of strings to the `by` parameter in theses functions:

- `highestMax`
- `highestAverage`
- `highestCurrent`
- `lowestMax`
- `lowestAverage`
- `lowestCurrent`

I get this (or something similar to this) error back:

```
Error: error calling function "highestMax": error calling function "_highestOrLowest": error calling function "_sortLimit": error calling function "group": error calling function "reducer": error calling function "max": cannot access property "0" on value of type array
```

PR #244, opened 2018-11-06
type scope for interpreter

jlapacik:
This PR addresses a bug that was discovered through #216 .

Previously when evaluating a Flux program, type inference would infer the types of all expressions in the program and store them in a `TypeSolution`. However a `TypeSolution` was tied to a single instance of the Flux interpreter, and Flux programs use at least two distinct instances of the interpreter - one for builtin evaluation and one for program evaluation. Whenever a new interpreter was instantiated the previous `TypeSolution` was effectively discarded.

This PR adds a `TypeScope` object that is passed to a new Flux interpreter as a way to keep the `TypeSolution` information from previous builtin evaluations.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -15,30 +15,31 @@ type Interpreter struct {
 	values  []values.Value
 	options *Scope
 	globals *Scope
-
-	typeSol semantic.TypeSolution
+	types   *TypeScope
 }
 
 // NewInterpreter instantiates a new Flux Interpreter whose builtin values are not mutable.
 // Options are always mutable.
-func NewInterpreter(options, builtins map[string]values.Value) *Interpreter {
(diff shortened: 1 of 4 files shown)


interpreter/interpreter.go
```diff
@@ -15,30 +15,31 @@ type Interpreter struct {
...
 // NewInterpreter instantiates a new Flux Interpreter whose builtin values are not mutable.
 // Options are always mutable.
-func NewInterpreter(options, builtins map[string]values.Value) *Interpreter {
+func NewInterpreter(options, builtins map[string]values.Value, types *TypeScope) *Interpreter {
 	optionScope := NewScopeWithValues(options)
 	globalScope := optionScope.NestWithValues(builtins)
 	interpreter := &Interpreter{
 		options: optionScope,
 		globals: globalScope.Nest(),
+		types:   types,
```
nathanielc: This line should nest the type scope as it is the non mutable version of NewInterpreter.

PR #252, opened 2018-11-07
refactor: move controller transitions to a single method

jsternberg:
Moving the controller transitions to a single method makes it easier to
reason about how the transitions affect the overall state of the system
and take that logic out of the locations where transitions happen.

This doesn't fix any of the code that is buggy, but it should help build
a cleaner base for further refactors of the controller and future
testing of transitions.


control/controller.go
```diff
@@ -530,6 +461,97 @@ func (q *Query) State() State {
...
+		}
+
+		if !okToTransition {
+			return false
+		}
+	}
+
+	// We are transitioning to a new state. Close the current span (if it exists).
+	if q.currentSpan != nil {
+		q.currentSpan.Finish()
```
stuartcarnie: Food for thought, logic like this is a good use case for `goto`:

```go
	if len(currentState) > 0 {
		// Find the current state in the list of current states.
		for _, st := range currentState {
			if q.state == st {
				goto TRANSITION
			}
		}

		return false
	}

TRANSITION:
	// We are transitioning to a new state. Close the current span (if it exists).
	if q.currentSpan != nil {
		q.currentSpan.Finish()
```

PR #253, opened 2018-11-07
more restrictions for function type unification

### LINKED ISSUE #226: Type Inference: pipe argument ignored when inside function body
The following

    g = () => 5
    f = () => 4 |> g()

Passes type inference when it should fail because `g` does not accept a pipe argument.

jlapacik:
This PR fixes #226 . However it seems to add a lot of mental gymnastics to the logic for unifying function types in the presence of pipe arguments. Perhaps this logic could be moved to the constraint phase of type inference? Not sure yet.


jlapacik:
@nathanielc can you give this another review? The flow is now consistent with the above steps

PR #257, opened 2018-11-07
cleanup: remove cancel requests from the controller

jsternberg:
This section of code was never used. When we finally work on a api for
canceling running queries, we'll revisit this to flesh it out. It is
very unlikely that channels are needed to handle this.

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -43,12 +43,11 @@ import (
 // Controller provides a central location to manage all incoming queries.
 // The controller is responsible for queueing, planning, and executing queries.
 type Controller struct {
-	newQueries    chan *Query
-	lastID        QueryID
-	queriesMu     sync.RWMutex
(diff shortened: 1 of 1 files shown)


nathanielc:
I believe this code is being used. What makes you think it is dead? We support query cancellation, is this somehow a dead code path during cancellation?

jsternberg:
@nathanielc it's an unexposed channel and it isn't used by any public methods. It's only usage according to GoLand is being read from within the controller loop. The `Cancel` method doesn't use it.

Also, upon reading the code, there is a 100% chance that this would deadlock if it were actually invoked in 0.7.0. It won't deadlock now, but that's because I recently (post 0.7) fixed the cause of the deadlock.

jsternberg:
These are getting removed in a recent controller refactor so this branch isn't worth merging.

See #1115 for details.

PR #258, opened 2018-11-08
feat: shutdown method for the controller

### LINKED ISSUE #245: Query controller should be closable/cancelable
Currently, NewController starts `c.run` in its own goroutine and that method does not return, so every controller created in test leaks a goroutine. It would be better if there was a way to shut down a controller.

jsternberg:
Fixes #245.

### DIFF
--- a/control/controller_test.go
+++ b/control/controller_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/influxdata/flux"
 	_ "github.com/influxdata/flux/builtin"
+	"github.com/influxdata/flux/internal/pkg/syncutil"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/mock"
 	"github.com/influxdata/flux/plan"
@@ -32,6 +33,14 @@ func TestController_CompileQuery_Failure(t *testing.T) {
 	}
 
 	ctrl := New(Config{})
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer func() {
+		if err := ctrl.Shutdown(ctx); err != nil {
+			t.Fatal(err)
+		}
+		cancel()
+	}()
 
 	// Run the query. It should return an error.
 	if _, err := ctrl.Query(context.Background(), compiler); err == nil {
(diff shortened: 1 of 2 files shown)


control/controller_test.go
```diff
@@ -219,3 +260,55 @@ func TestController_BlockedExecutor(t *testing.T) {
...
+	// There should be 10 active queries.
+	if want, got := 10, len(ctrl.Queries()); want != got {
+		t.Fatalf("unexpected query count -want/+got\n\t- %d\n\t+ %d", want, got)
+	}
+
+	// Mark each of the queries as done.
+	for _, q := range queries {
+		q := q
+		wg.Do(func() error {
+			<-q.Ready()
```
jlapacik: I'm assuming this gets closed when `Shutdown` is called?

control/controller_test.go
```diff
@@ -189,6 +222,14 @@ func TestController_BlockedExecutor(t *testing.T) {
 	ctrl := New(Config{})
 	ctrl.executor = executor
 
+	cctx, ccancel := context.WithTimeout(context.Background(), time.Second)
+	defer func() {
+		if err := ctrl.Shutdown(cctx); err != nil {
```
jlapacik: Does this work because cctx is the top level context?

control/controller_test.go
```diff
@@ -219,3 +260,55 @@ func TestController_BlockedExecutor(t *testing.T) {
...
+	}
+
+	// There should be 10 active queries.
+	if want, got := 10, len(ctrl.Queries()); want != got {
+		t.Fatalf("unexpected query count -want/+got\n\t- %d\n\t+ %d", want, got)
+	}
+
+	// Mark each of the queries as done.
+	for _, q := range queries {
+		q := q
```
jlapacik: Why is it necessary to shadow here?

control/controller_test.go
jsternberg: No, it is just used to prevent a deadlock from stopping the cleanup. So this context isn't part of the test.

control/controller_test.go
jsternberg: This is because of how for loops interact with closures.

https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

control/controller_test.go
jsternberg: Yes, but it might be good if I were to add a timeout to this.

PR #259, opened 2018-11-08
fix: revert the change to query result iterator

jsternberg:
The query result iterator should not take a context. The context passed
to Query should be the one that controls whether or not a query gets
canceled.

This reverts the API back, but keeps the cancel stuff removed since
cancel was never used for actually canceling a query. The context was
always used for that.

### DIFF
--- a/result_iterator.go
+++ b/result_iterator.go
@@ -1,7 +1,6 @@
 package flux
 
 import (
-	"context"
 	"sort"
 )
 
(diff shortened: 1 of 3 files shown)


result_iterator.go
```diff
@@ -48,15 +47,11 @@ func (r *queryResultIterator) More() bool {
 	}
 
 	if r.results == nil {
-		select {
-		case <-r.ctx.Done():
+		results, ok := <-r.query.Ready()
+		if !ok {
```
wolffcm: It doesn't seem like `r.done` gets set to `true` until `Release` is actually called.  This might be surprising to future-us?  Maybe `r.done` should be called `r.released`?

jsternberg: Done.

PR #262, opened 2018-11-09
fix: array indexing

### LINKED ISSUE #216: "by" parameter in highestMax, highestAverage, etc., not working

jlapacik:
This PR fixes #216 .

Previously during parsing, array access would define a `MemberExpression` node in the AST. However during program evaluation, when the interpreter would see a `MemberExpression`, it would assume that this meant accessing an object - not an array. As a result, not only did array access throw a runtime error, but in some special cases, like #216, an incorrect type error would be thrown, because the type system also assumed `MemberExpressions` corresponded to object access.

This PR defines defines array access and object access as separate nodes in the AST. It also adds the necessary constraints during type inference to safely catch any type errors associated with array access.


parser/types.go
```diff
@@ -119,10 +119,18 @@ func incompletePipeExpr(call interface{}, text []byte, pos position) (*ast.PipeE
 func memberexprs(head, tail interface{}, text []byte, pos position) (ast.Expression, error) {
 	res := head.(ast.Expression)
 	for _, prop := range toIfaceSlice(tail) {
-		res = &ast.MemberExpression{
-			Object:   res,
-			Property: prop.(ast.Expression),
-			BaseNode: base(text, pos),
+		if integer, ok := prop.(*ast.IntegerLiteral); ok {
```
jsternberg: What happens when you have an expression that returns an integer?

jlapacik: Good catch. I did not consider this case. I think the proper way to handle it is probably to have a single `MemberExpression` AST node and have type inference figure out if it represents array access or object access. This will mean re-working a lot of the changes here.

jsternberg: 👍 

This is likely going to require a lot of reworking. Can we close this PR since this method is unlikely to work and we'll create a new branch/new PR  for the new method? I think what you described will work much better personally.

jlapacik: Yes closing this PR.

PR #265, opened 2018-11-09
feat: add Fuzz method for go-fuzz into the parser

jsternberg:
The Fuzz method is added to a separate file with the `gofuzz` tag so it
is not included in non-gofuzz builds. The Fuzz method is not considered
part of the public API.


jsternberg:
Not at the moment. The fuzzing itself is currently planned to be in a separate repository and controlled from there.

PR #274, opened 2018-11-12
feat: initial token package

jsternberg:
It covers the various primitives for tokenizing a file and for using
those primitives in the scanner and parser.


jsternberg:
It isn't according to the spec and it's in a non-expression context so we'll read an ident to determine if option is used.

PR #276, opened 2018-11-13
flux array indexing

### LINKED ISSUE #216: "by" parameter in highestMax, highestAverage, etc., not working

jlapacik:
This PR fixes #216 . Previously access to both array and object values was done through a `MemberExpression`. This PR defines a new lexical expression called an `IndexExpression` used for accessing the values of an array. It also adds a new kind constraint to the type inference system called an `ArrayKind`. This constraint ensures that any expression used to index an array resolves to an integer type.


parser/flux.peg
```diff
@@ -75,10 +75,10 @@ MemberExpressions
 
 MemberExpressionProperty
   = "." __ property:Identifier {
-      return property, nil
+      return membexpr(property)
     }
     / "[" __ property:Primary __ "]" __ {
-      return property, nil
+      return indexpr(property)
```
nathanielc: This does not seem complete to me.

The `[]` do not differentiate index vs member expressions. For example `r["name"]` is a member expression as well as `r.name`. But `r[0]` or `r[i]` are index expressions. 

This should be possible to parse since if after the `[` a string literal follows then it is a member expression otherwise it is an index expression.

parser/parser_test.go
```diff
@@ -526,9 +526,92 @@ func TestParse(t *testing.T) {
...
+								Callee: &ast.Identifier{Name: "f"},
+							},
+							Index: &ast.IntegerLiteral{Value: 3},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "index with member expressions",
```
nathanielc: This is the test case that would change based on the above. All of these are member expressions.

semantic/analyze.go
```diff
@@ -295,23 +297,33 @@ func analyzeMemberExpression(member *ast.MemberExpression) (*MemberExpression, e
...
+	object, err := analyzeExpression(e.Object)
+	if err != nil {
+		return nil, err
+	}
+	index, err := analyzeExpression(e.Index)
+	if err != nil {
+		return nil, err
+	}
+	if n, ok := index.(*StringLiteral); ok {
+		return &MemberExpression{
```
nathanielc: Ah I see how it is working now... I still think we should make the distinction at the parser level.

semantic/constraints.go
```diff
@@ -306,14 +306,33 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
 		return ptv, nil
+	case *IndexExpression:
+		ptv := v.cs.f.Fresh()
+		t, err := v.lookup(n.Object)
+		if err != nil {
+			return nil, err
+		}
+		tv, ok := t.(Tvar)
+		if !ok {
+			return nil, errors.New("object must be a type variable")
```
nathanielc: this should read `array must be a type variable`.

semantic/constraints.go
```diff
@@ -306,14 +306,33 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
+		}
+		idx, err := v.lookup(n.Index)
+		if err != nil {
+			return nil, err
+		}
```
nathanielc: I expected an AddTypeConst call here with `idx` and Int. Does this happen inside of the ArrayKind or something? If it does can we move it out? I would expect the ArrayKind to only deal with the element type.

ast/ast.go
```diff
@@ -362,11 +364,31 @@ func (e *MemberExpression) Copy() Node {
...
-	ne.Property = e.Property.Copy().(Expression)
+	ne.Property = e.Property.Copy().(*Identifier)
 
 	return ne
 }
 
+// IndexExpression represents indexing into an array
+type IndexExpression struct {
+	BaseNode
+	Object Expression `json:"object"`
```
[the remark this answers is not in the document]
jsternberg: Is there a reason we still have `IndexExpression`? Is there any difference between it and `MemberExpression`?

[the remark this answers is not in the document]
jlapacik: In the AST a bracketed expression such as `r["name"]` or `r[0]` is considered an index expression. A dot signifies a member expression.

parser/flux.peg
jlapacik: `r["name"]` should be considered a member expression. I wasn't sure where this conversion should take place. I decided to put it in the semantic package. My reasoning was  so that the AST nodes represented specified lexical elements, whereas semantic nodes represented the meaning of those elements. But I'm totally fine having the conversion done inside of the parser if that makes more sense.

semantic/analyze.go
jlapacik: Ah ok. Oops, disregard my comment above.

semantic/constraints.go
jlapacik: Ah yes I forgot to change that.

semantic/constraints.go
jlapacik: The reason I coupled the type constraint and kind constraint together was just in case changes where made to the `ArrayKind` constraint later on. But I agree, the type constraint should be separate.

ast/ast.go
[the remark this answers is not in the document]
nathanielc: @jsternberg Also the type inference needs to know the difference. Javascript didn't need to differentiate them since types are checked at runtime (We modeled the ast off the Javascript ast). With Flux we need to be able to infer the types of the expressions so the difference is important.

jlapacik:
@nathanielc `ArrayKind` is now defined only by its element type and a separate type constraint has been added to the constraint generation phase to ensure that index types are integers. Also the optimization that converts an index expression of the form `r["name"]` to a member expression is now performed during parsing.

parser/types.go
```diff
@@ -119,29 +119,46 @@ func incompletePipeExpr(call interface{}, text []byte, pos position) (*ast.PipeE
...
+			res = &ast.MemberExpression{
+				Object:   res,
+				Property: expr.property,
+				BaseNode: base(text, pos),
+			}
+		case *indexExpression:
+			if lit, ok := expr.key.(*ast.StringLiteral); ok {
+				res = &ast.MemberExpression{
+					Object:   res,
+					Property: &ast.Identifier{Name: lit.Value},
```
nathanielc: I would like to preserve the invariant that an Identifier's Name is always a valid identifier. Using the string literal value as the Name breaks that. See my comment above about using *Identifier vs Expression as the Property.  

I think a change to the parser grammar is required to make this work.

There should be three production rules:

* One for `Expr "." Identifier` -> MemberExpression
* One for `Expr "[" StringLiteral "]"` -> MemberExpression
* One for `Expr "[" Expression "]" ` -> IndexExpression (the integer literal case fall into this one)

The StringLiteral is more specific than an Expression so it should appear first. This should remove the need to have the special `memberExpression` and `indexExpression` types in the parser.

compiler/compiler.go
```diff
@@ -162,6 +162,20 @@ func compile(n semantic.Node, typeSol semantic.TypeSolution, builtIns Scope, fun
...
+			return nil, err
+		}
+		idx, err := compile(n.Index, typeSol, builtIns, funcExprs)
+		if err != nil {
+			return nil, err
+		}
+		return &arrayEvaluator{
+			t:     monoType(typeSol.TypeOf(n)),
+			array: arr,
+			index: int(idx.EvalInt(builtIns)),
```
nathanielc: This wont work when the index is an expression instead of a literal. The Eval call needs to happen as part of the Eval call of the arrayEvaluator.

compiler/runtime.go
```diff
@@ -981,6 +981,61 @@ func (e *memberEvaluator) EvalFunction(scope Scope) values.Function {
 	return v.Function()
 }
 
+type arrayEvaluator struct {
+	t     semantic.Type
+	array Evaluator
+	index int
```
nathanielc: As noted above the index here should be an Evaluator. Then in the Eval methods below Eval the index as an Int.

parser/types.go
```diff
@@ -119,29 +119,30 @@ func incompletePipeExpr(call interface{}, text []byte, pos position) (*ast.PipeE
...
-		m.Object = object.(ast.Expression)
-	}
+func memberExpr(p interface{}) (*ast.MemberExpression, error) {
+	return &ast.MemberExpression{Property: p.(*ast.Identifier)}, nil
+}
 
-	if property != nil {
-		m.Property = property.(*ast.Identifier)
-	}
+func memberExprFromIndexExpr(p interface{}) (*ast.MemberExpression, error) {
```
nathanielc: Can you rename this to something like `memberExprFromStringLiteral` and rename `memberExpr` to something like `memberExprFromIdentifier`?

parser/types.go
```diff
@@ -119,29 +119,30 @@ func incompletePipeExpr(call interface{}, text []byte, pos position) (*ast.PipeE
...
-		res = &ast.MemberExpression{
-			Object:   res,
-			Property: prop.(ast.Expression),
-			BaseNode: base(text, pos),
+		switch expr := prop.(type) {
+		case *ast.MemberExpression:
+			expr.BaseNode = base(text, pos)
+			expr.Object = res
+			res = expr
+		case *ast.IndexExpression:
```
nathanielc: I am not sure why `memberexprs` changes? How does an IndexExpression get passed to this method?

semantic/constraints.go
```diff
@@ -306,14 +306,31 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
+		v.cs.AddKindConst(tv, ArrayKind{ptv})
+		v.cs.AddTypeConst(idx, Int, n.Index.Location())
+		return ptv, nil
 	case *ArrayExpression:
-		at := array{typ: NewObjectPolyType(nil, nil, AllLabels())}
+		at := array{typ: v.cs.f.Fresh()}
```
nathanielc: Using a type variable makes a lot more sense :), With this change the loop below does not need to be inside a len() > 0 check and does not need to treat the first element as special. Instead the loop can now just unify against `at.typ` directly. Can you make that simplification?

parser/types.go
jlapacik: The grammar defines a member expression as an identifier followed by a sequence of `MemberExpressionProperty`s. These properties can either be of the form `.identifier` or `[expression]`.

PR #283, opened 2018-11-14
feat: copy parser tests to the internal parser package

jsternberg:
This will allow us to easily test the new parser functionality to match
the old one and ensure the project can be done in multiple commits
instead of as one gigantic commit.


jsternberg:
It's just a copy and paste job in preparation for my work today, but thanks!

PR #285, opened 2018-11-14
feat: add error node to the flux ast

jsternberg:
The error node can go anywhere in the AST and signals there was an
error.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -55,6 +55,7 @@ type Node interface {
 }
 
 func (*Program) node() {}
+func (*Error) node()   {}
 
 func (*BlockStatement) node()      {}
 func (*ExpressionStatement) node() {}
@@ -100,6 +101,25 @@ func (b BaseNode) Location() SourceLocation {
 	return *b.Loc
 }
 
+// Error represents an error within the AST. It is created
+// by the parser when the parser gets an invalid sequence.
+type Error struct {
+	BaseNode
+	Message string `json:"message"`
+}
+
+// Type is the abstract type.
+func (*Error) Type() string { return "Error" }
+
+func (e *Error) Copy() Node {
+	ne := *e
+	return &ne
+}
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
 // Program represents a complete program source tree
 type Program struct {
 	BaseNode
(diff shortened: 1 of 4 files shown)


nathanielc:
The type ast.Node is the most generic type for an ast node, but unfortunately that does not mean it can go anywhere in the ast. For example the Object expression has a list of Property nodes, I would not be possible to set the ast.Error as the properties for an object node. Similarly concrete types are used throughout the ast nodes. 

I don't think we should change all fields to use ast.Node instead of concrete types as we loose a lot of type expressiveness with that.

My first thought as to how this could work is to make the Error type not an ast.Node but rather part of the BaseNode type, and then have the ast.Node interface expose the Error via a method. Thoughts?

jsternberg:
I didn't modify `ast.Node`. I just made it so `ast.Error` implemented all of the possible interfaces so it could be placed anywhere. The property is specific to `ast.Error`.

jsternberg:
To add to the above, a `PrintErrors` function could call `ast.Visit` and then just do a type switch on `*ast.Error` to print all of the errors with line numbers attached.

nathanielc:
> I didn't modify ast.Node. I just made it so ast.Error implemented all of the possible interfaces so it could be placed anywhere. The property is specific to ast.Error.

My point is that even though ast.Error implements all of the interfaces it does not mean it can go anywhere in the AST. Various ast nodes use concrete types not interfaces for their fields. 

So instead of replacing ast node with error nodes I was thinking that each ast node could contain an error.  Then the PrintErrors method uses ast.Visit but just checks each visited node for an error. Its a similar concept but avoids the concrete type vs interface issue.

jsternberg:
I'll modify it to be an interface that implements `ast.Node` and exposes another method. At the moment, I'm hesitant to include too much because I don't know exactly where I'm going to end up.

jsternberg:
So looking things over, it's likely easiest to piggy-back on the error interface for now. The `Error` node implements `Error() string` and so that will probably be easiest to use until I have more details while I'm working on the parser.

The interface I was going to construct was just the source location and an error string right now anyway, so that's already there.

nathanielc:
@jsternberg I am not sure what you mean by piggy-back off the error interface. Can you clarify? Maybe push up a change to demonstrate what you mean?

jsternberg:
@nathanielc I'm going to write the `GetErrors` method to show you rather than try to explain it. I created a PR for adding one of the required methods to the AST package in #287.

nathanielc:
Closing this PR as its is out of date and we are discussing error handing here #288

PR #287, opened 2018-11-15
feat: add a method for walking the flux ast

jsternberg:
The walk method matches the one in the semantic package of the same name
and follows the same pattern exactly.

### DIFF
--- a/internal/ast/astutil/walk.go
+++ b/internal/ast/astutil/walk.go
@@ -1,0 +1,257 @@
+package astutil
+
+import "github.com/influxdata/flux/ast"
+
+// A Visitor's Visit method is invoked for each node encountered by Walk.
+// If the result visitor w is not nil, Walk visits each of the children
+// of node with the visitor w, followed by a call of w.Done(n).
+// This will not visit nil nodes.
+type Visitor interface {
+	Visit(ast.Node) Visitor
+	Done(ast.Node)
+}
+
+func Walk(v Visitor, n ast.Node) {
+	switch n := n.(type) {
+	case *ast.ArrayExpression:
+		if n == nil {
+			return
+		}
+		if w := v.Visit(n); w != nil {
+			for _, element := range n.Elements {
+				w.Visit(element)
+			}
+		}
+
+	case *ast.ArrowFunctionExpression:
+		if n == nil {
+			return
+		}
+		if w := v.Visit(n); w != nil {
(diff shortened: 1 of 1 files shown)


jsternberg:
#288 is an example of using this for collecting errors in the flux AST.

ast/walk.go
```diff
@@ -0,0 +1,132 @@
...
+		for _, decl := range n.Declarations {
+			v.Visit(decl)
+		}
+
+	case *VariableDeclarator:
+		v.Visit(n.ID)
+		v.Visit(n.Init)
+	}
+
+	v.Visit(nil)
```
nathanielc: I dislike this pattern of using `nil` to signal the visit is complete. I'd rather have an explicit method for it. See semantic/walk.go for an example.

Also in semantic/walk.go we made it impossible to walk nil nodes, do we want to do something similar here?

jsternberg: I've been confused about the pattern too. I would say that walking nil nodes is fine as is the case of a nil `*ast.Identifier`, but I don't understand the `nil` sentinel value since it seems mostly useless to me as it doesn't really handle nesting IMO.

Is there any reason why we need it at all? I'm not convinced `Done()` is very helpful either.

ast/walk.go
```diff
@@ -0,0 +1,133 @@
...
+		for _, decl := range n.Declarations {
+			v.Visit(decl)
+		}
+
+	case *VariableDeclarator:
+		v.Visit(n.ID)
+		v.Visit(n.Init)
+	}
+
+	v.Done(n)
```
nathanielc: This call to Done needs be on the original `v` not the child v. Meaning as it is defined Done is called on a node only after all of its children have been visited (which the current code does), but Done needs to be called on the same visitor that had Visit called for a given node.

internal/ast/astutil/walk.go
```diff
@@ -0,0 +1,257 @@
...
+
+func Walk(v Visitor, n ast.Node) {
+	switch n := n.(type) {
+	case *ast.ArrayExpression:
+		if n == nil {
+			return
+		}
+		if w := v.Visit(n); w != nil {
+			for _, element := range n.Elements {
+				w.Visit(element)
```
nathanielc: Note, these calls to w.Visit do not recurse, so the Walk implementation as it stands now only visits direct children, but not any grand children.

PR #288, opened 2018-11-15
feat(ast): add a function for checking for errors within the AST

jsternberg:
The `Check` function will annotate the AST with errors in the AST
construction. The parser will be modified to make a best effort to
produce an AST, but it will produce an invalid AST when it gets an
invalid input.

The `GetErrors` method will traverse the AST and report any errors it
finds within the AST construction that were added a result of `Check`.

Related to #305.

### DIFF
--- a/ast/errors.go
+++ b/ast/errors.go
@@ -1,0 +1,92 @@
+package ast
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"strconv"
+)
+
+// Check will inspect each node and annotate it with any AST errors.
+// It will return the number of errors that were found.
+func Check(root Node) int {
+	v := errorVisitor{}
+	Walk(&v, root)
+	return v.count
+}
+
+// check will inspect a single node and annotate it with any AST errors.
+// For convenience, it returns the number of errors that were identified.
+func check(n Node) int {
+	// TODO(jsternberg): Fill in the details for how we retrieve errors.
+	switch n := n.(type) {
+	case *BadStatement:
+		n.Errors = append(n.Errors, Error{
+			Msg: fmt.Sprintf("invalid statement: %s", n.Text),
+		})
+		return len(n.Errors)
+	}
+
+	return 0
+}
+
+// GetError will return the first error within an AST.
+func GetError(n Node) error {
+	errs := GetErrors(n)
+	if len(errs) == 0 {
+		return nil
+	}
+	return errs[0]
+}
+
+// GetErrors will return each of the errors within an AST.
+func GetErrors(n Node) (errs []error) {
+	Walk(CreateVisitor(func(node Node) {
+		if nerrs := node.Errs(); len(nerrs) > 0 {
+			for _, err := range nerrs {
+				errs = append(errs, err)
+			}
+		}
+	}), n)
+	return errs
+}
+
+// PrintErrors will format the errors within the AST and output them
+// to the writer.
+func PrintErrors(w io.Writer, root Node) {
+	var buf bytes.Buffer
+	Walk(CreateVisitor(func(node Node) {
+		if errs := node.Errs(); len(errs) > 0 {
+			loc := node.Location()
+			for _, err := range errs {
+				buf.WriteString("error")
+				if loc.Start.Line > 0 {
+					buf.WriteByte(':')
+					buf.WriteString(strconv.FormatInt(int64(loc.Start.Line), 10))
+				}
+				if loc.Start.Column > 0 {
+					if buf.Len() > 0 {
+						buf.WriteByte(':')
+					}
+					buf.WriteString(strconv.FormatInt(int64(loc.Start.Column), 10))
+				}
+				buf.WriteString(": ")
+				buf.WriteString(err.Msg)
+				buf.WriteByte('\n')
+				_, _ = w.Write(buf.Bytes())
+				buf.Reset()
+			}
+		}
+	}), root)
+}
+
+type errorVisitor struct {
+	count int
+}
+
+func (ev *errorVisitor) Visit(n Node) Visitor {
+	ev.count += check(n)
+	return ev
+}
+
+func (ev *errorVisitor) Done(n Node) {}
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -5,7 +5,7 @@ import (
 	fastparser "github.com/influxdata/flux/internal/parser"
 )
 
-// NewAST parses Flux query and produces an ast.Program
-func NewAST(flux string) (*ast.Program, error) {
-	return fastparser.NewAST([]byte(flux)), nil
+// NewAST parses Flux query and produces an ast.Program.
+func NewAST(flux string) *ast.Program {
+	return fastparser.NewAST([]byte(flux))
 }
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -48,6 +48,7 @@ type Node interface {
 	node()
 	Type() string // Type property is a string that contains the variant type of the node
 	Location() SourceLocation
+	Errs() []Error
 	Copy() Node
 
 	// All node must support json marshalling
@@ -58,6 +59,7 @@ func (*Program) node()           {}
 func (*PackageClause) node()     {}
 func (*ImportDeclaration) node() {}
 
+func (*BadStatement) node()        {}
 func (*Block) node()               {}
 func (*ExpressionStatement) node() {}
 func (*ReturnStatement) node()     {}
@@ -91,7 +93,8 @@ func (*UnsignedIntegerLiteral) node() {}
 
 // BaseNode holds the attributes every expression or statement should have
 type BaseNode struct {
-	Loc *SourceLocation `json:"location,omitempty"`
+	Loc    *SourceLocation `json:"location,omitempty"`
+	Errors []Error         `json:"errors,omitempty"`
 }
 
 // Location is the source location of the Node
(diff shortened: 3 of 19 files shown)


jsternberg:
If we're going to use this kind of method, another possibility is to build in this functionality to the `semantic` package. For influxql, we originally built validation and compilation as separate steps and this ended up resulting in code being duplicated between the two sections and it added bugs. I found it useful to refactor influxql so that the validation step and compilation steps were together so that there was actual work done in validation. It makes it harder to write the code incorrectly because a meaningful result is produced.

So it might be better to build this functionality into the semantic package since it is already there. The AST nodes would still have to be added, but `GetErrors` would disappear and instead the logic for each of those would go in the corresponding `analyze` function in semantic.

I think I like that idea better actually.

nathanielc:
I disagree that the AST package should push validation to the semantic package. The AST package should have a `Validate` method that reports errors if any exist given an ast.Node.

As an example use of the ast.Validate method would be inside of the ast.Format method. The ast.Format method should refuse to format invalid AST nodes as doing so would be a destructive action. This workflow should not depend on the semantic package.

The semantic package is going to need the same, meaning it is going to need to be able to provide some semantic analysis of AST even if the AST itself contains errors. This is because we still want to provide completion hints based on type information even if there is a syntax error somewhere in a source file. 


I think the simplest solution is to be able to attach an ast.Error node as a child of every other ast.Node. This way we can encode errors anywhere we find them in the source file. I expect this to be similar to how we handle comments as well. Every ast.Node can have comments associated with them, the difference being that comments can be formatted with the source while errors cannot.

For example the error case brought up today:

    () asdf asdf asdf => 2

This currently parses because we ignore tokens we do not understand. Instead of ignoring these tokens we can create an ast.Error node and attach it to the ast.FunctionExpression node as an additional child. Then when asked if the ast is valid it would return an error about the extra identifiers in the function expression. But in the meantime we can perform  semantic analysis on the function expression and provide completion hints based on the type of the function.

 The query engine will refuse to execute any invalid Flux scripts, but during the editing process we can work with errors as part of the normal AST and semantic graph.

ast/errors.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (ev *errorVisitor) Visit(n Node) Visitor {
+	if !ev.more() {
+		return nil
+	}
+
+	// TODO(jsternberg): Fill in the details for how we retrieve errors.
+	switch n := n.(type) {
+	case *BadStatement:
```
nathanielc: What is the plan for how to encode errors into the AST? This is the most important question to me.

Also remember that the AST needs to be able to serialize to JSON, which includes its errors. So whatever mechanism we choose needs to easily marshal/unmarshal to JSON.

ast/errors.go
```diff
@@ -0,0 +1,121 @@
...
+	"strconv"
+	"strings"
+)
+
+// GetErrors will collect all of the errors from the construction of the AST.
+// If there were no errors, this will return nil. If the limit is set to
+// positive number, it will limit the number of errors to the given number. If
+// it is set to a negative number, it will return the last N errors. A value of
+// zero will return all errors.
+func GetErrors(root Node, limit int) []error {
```
nathanielc: What is the motivation for the limit logic being the last errors? Why do we need the limit logic at all? An AST is expected to fit in RAM, since its error are a subset seems reasonable to always return all errors.

ast/errors.go
jsternberg: My initial plan would be for it to be dependent on the node itself. `BadStatement` is the easiest, but I figured that if you had an expression like `foo(a b)` it might create a property node like this:

```
{
    "type": "CallExpression",
    "callee": {
        "type": "Identifier",
        "name": "foo"
    },
    "arguments": [
        {
             "type": "ObjectExpression",
             "properties": [
                  {
                      "type": "Property",
                      "key": {
                          "type": "Identifier",
                          "name": "a"
                      },
                      "separator": {
                           "missing": true
                      },
                      "value": {
                          "type": "Identifier",
                          "name": "b"
                      }
                  }
             ]
        }
    ]
}
```

In this way, the errors are part of the contruction of the AST and don't have to be explicitly added. They are just there by fact of how it exists. It also means that the information of where the separators are is embedded as part of the AST itself for analysis programs. This is similar to how the Go AST records positions for symbols such as for [ParenExpr](https://godoc.org/go/ast#ParenExpr) where the `Lparen` and `Rparen` are part of the AST construction. In this case, we would just differentiate that it is possible for the position to be invalid and indicate that it is missing. The zero value would still be valid though and indicate that position information is missing so the errors are explicit.

ast/errors.go
jsternberg: This is more for a command line tool. It is pretty common for a compiler to limit itself to the first 10 errors. The  limit  logic is for the first errors and a negative number gets you the last errors, but I could just remove that.

nathanielc: OK, yeah, when such a CLI tool exists that tool can limit the errors it displays. Lets remove it from here.

ast/errors.go
nathanielc: Does this mean we need to create error types for all possible syntax errors? That does not seem possible. For example how would you express this syntax error in the AST using the above plan `foo(a:)`?

jsternberg: That wasn't my plan. There would be a few error nodes, such as `BadStatement` and `BadExpression` (which exist in `go/ast` too), but many of the errors would be derived from the AST that's currently produced. So the above could either be represented by having the `Value` set to `nil` (which is not allowed in a call expression context) or the value could be set to `BadExpression` with a detailed message.

This would limit the number of error AST's and I think would make the error generation more flexible. I don't think we want very many `BadXXX` types. For examples, a misformed expression such as `% a` could be represented by a `UnaryExpression` with `%` as the operator kind. Since this is not allowed, it would generate an error from the error visitor. A similar thing would be true of multiple identifiers. If we see something like `foo(a: b c)` it would be represented as the property value being a `BinaryExpression` between two expressions with no operator kind. When the error visitor sees that, it says, "well you're missing some kind of operator here".

ast/errors.go
jsternberg: Removed this. We can always add it back later.

ast/errors.go
nathanielc: How are you thinking UnaryExpression will work when it has `%` as its operator? Were you thinking that this above switch statement would switch on UnaryExpressions and then check for valid operators and if it doesn't find one it would create an error? Or something else?

jsternberg: Yes. I was expecting this switch statement to contain the equivalent `Validate` logic for each node, but I wanted to keep it hidden just so we had more flexibility with error handling so there wouldn't be a specific `Validate` method. I also wanted it in one location just so that we could see all error handling text in a single location rather than hunting around for the correct implementation.

Although that's an implementation detail so we could switch to using `Validate` instead pretty easily or we could switch the logic to an unexposed `validate` method. Before doing that though, I would like to experiment with this code layout.

nathanielc: OK I think I understand the proposed design, but I think we are going to need a different design. Specifically we do not want the AST nodes to have behavior associated with them, rather we need them to be declarative in nature. This is because once you encode the AST to JSON and then hand that encoded JSON to an external program you can no longer leverage any behavior associated with the AST nodes, be it a GetErrors or a validate method. As such we need the errors to be explicitly encoded into the AST.

A concrete use case is Chronograf, they currently send the Flux source text to the parser via an HTTP request and get a JSON encoded AST in return. That AST needs to contain the error information. This way the Flux editor in Chronograf can communicate those errors to the user. If the error information can only be retrieved by passing the AST to a method then it is inaccessible to external programs like Chronograf. While we could extend the JSON representation of the AST to contain a list of errors how would you unmarshal that representation? Ignore the list of errors? At that point the JSON structure has duplicated information that can get out of sync. Encoding the error information directly into the AST means that the AST remains declarative (i.e. just data no behavior) and normalized (no duplicate data).

jsternberg: After an offline conversation, we came to a different design. The `Errors` attribute is going to be roughly like this:

```go
package ast

// Check will inspect each node and annotate each Node with any AST errors.
func Check(n ast.Node) int

// PrintErrors will recurse through the node and print any encountered errors to the io.Writer.
// The node must have been previously compiled.
func PrintErrors(w io.Writer, n ast.Node)

type BaseNode struct {
	Loc *SourceLocation `json:"location,omitempty"`
	Errors []Error `json:"errors,omitempty"`
}

// Error is a detailed informational for the error. This exists so that the error section
// is extensible.
type Error struct {
	Message string `json:"message"`
}
```

The basic premise is that `Check` is called to determine the errors within the AST. The parser _will_ produce AST's that are invalid when given invalid input, but it will not fill out errors. This way, we can separate the logic of errors away from the parser so the parser just focuses on doing its best. It also means the logic for checking an AST can be reused for human/machine-constructed AST's and AST's constructed by the parser.

We're going to give this method an attempt to see if it works well.

parser/parser.go
```diff
@@ -7,5 +7,10 @@ import (
 
 // NewAST parses Flux query and produces an ast.Program
 func NewAST(flux string) (*ast.Program, error) {
-	return fastparser.NewAST([]byte(flux)), nil
+	n := fastparser.NewAST([]byte(flux))
+	if ast.Check(n) > 0 {
+		errs := ast.GetErrors(n)
+		return nil, errs[0]
```
nathanielc: This should not return an error here. We want to be able to return invalid ASTs from the parser. As such I do not think it should run the Check step here either, rather if you need to Check an AST you do so explicitly. In other words assume an ast Node has not been Checked unless you did it explicitly.

jsternberg: Done.

PR #294, opened 2018-11-16
fix: add arrow token to the scanner and rename the pipe tokens

jsternberg:
The `<-` token got misnamed to `ARROW` because that's what it's called
in Go, but that doesn't apply here. The `=>` token was missed and should
be called `ARROW`, but it wasn't present in the spec so I forgot to add
it.

Renamed the two pipe tokens to play off of the names of each other.


docs/SPEC.md
```diff
@@ -103,7 +103,7 @@ The following keywords are reserved and may not be used as identifiers:
 
 The following character sequences represent operators:
 
-    +   ==   !=   (   )
+    +   ==   !=   (   )   =>
```
aanthony1243: I see what you meant on slack.  2 rows of 13 might be ok but a bit wide.  4 rows of 7/6?

jsternberg: I know! They're all so nicely grouped together by function as it is! It's driving me crazy!

We can fix this later. I'm going to merge so we have a functional scanner and I'll take care of my need for nicely formatted columns later.

PR #298, opened 2018-11-16
docs: Add rough draft of virtual machine design

nathanielc:
This PR adds a document on a proposed virtual machine implementation of Flux.
It is written as if the virtual machine already exists. I will use this PR comment to discuss how this differs from the current design and its strengths and weakness.

The current design at a high level looks like this:

```
Flux Source Code -> AST -> Semantic Graph -> Query Spec -> Plan -> Execution
```
The parser produces the AST, the analyzer produces the semantic graph, the interpreter produces the spec, the planner produces the plan and the query engine executes the plan.

Currently external systems can submit programmatically generated queries by generating  Flux source code or by generating a query spec to execute a query. For example the transpiler directly produces the query spec and submits it for execution.

The proposed change is this:

```
Flux source code -> AST -> Semantic Graph -> IR -> Execution
```

The parser produces the AST, the analyzer produces the semantic graph, the compiler produces the IR and the virtual machine executes the IR.

Obvious questions:

Q: Why the design change?
A: We want Flux to be able to describe control flow and data dependencies between transformations that a simple graph cannot describe. Using an IR allows us to describe more complex data dependencies like a `topN` query. Using an IR allows us to describe control flow, meaning execute different sub queries based on the result of a previous query. For example, anytime you use an identifier in a filter or map function, that is not just a simple constant expression, that is difficult to describe in a graph, but not in an IR. 

Q: Where did the planner go?
A: The planner is now encapsulated in the "compiler". The planner will consume IR and produce an optimized IR. The IR abstraction we have chosen maps closely to the types the planner is already manipulating. 

Q: With the query spec gone, how do external system submit programmatically generated queries?
A: External systems can use the semantic graph to submit programmatically generated queries. This means the transpiler will need to produce semantic graph instead of a query spec. We will not accept IR from external systems, the semantic graph will be the canonical source for describing Flux programs in addition to Flux source code.

Q: Is this design complete?
A: No, this is a very rough draft of the design. Please tear it apart so we can more fully specify it.

Q: Does this design make embedding Flux easier?
A: Maybe? This design is not specifically about improving the embedding story around Flux although it was discussed during design. This design will have a large runtime to handle most of the heavy lifting as such porting that runtime to other languages/systems will not be a small task.

Q: Do this mean we have to rewrite all the Go we have written as Flux/IR?
A: No, the IR abstraction is the same level as abstraction that the query spec provided. As such most of the code that has to change is the execution engine to be a virtual machine instead of just a direct Go implementation. In short the current Go code we have just becomes the virtual machine runtime. The nice thing about this is it provides a nice bridge between these designs and allows for the runtime to be updated without changing the core virtual machine.

### DIFF
--- a/docs/VirtualMachine.md
+++ b/docs/VirtualMachine.md
@@ -1,0 +1,160 @@
+# Virtual Machine Design
+
+This document details the workings and design of the virtual machine that executes a Flux script.
+
+## Execution Stack
+
+A Flux script goes through several transformations from text to execution.
+
+1. Flux source code is created via any means.
+2. The parser produces an abstract syntax tree (AST) from the source code.
+3. The semantic analyzer produces a semantic graph from the AST.
+4. The compiler accepts the semantic graph and produces an intermediate representation (IR) to be executed.
+5. The flux virtual machine executes the IR.
+
+NOTE: Should we call it bytecode or IR?
+Should we call it a VM or an interpreter?
+
+## Execution model
+
+Flux has a specific execution model.
+The model depends on the definitions of various data types and operations on those types.
+The Flux spec defines some of those types.
+This document reviews those types as well as defines a few new types that are specific to the implementation.
+
+### Data types and operations
+
+#### Table
+
+A Flux table consists of rows and columns.
+Columns have a label and a type.
+The table has zero or more rows, where each row defines a value for each column or a null is used to indicate a missing value.
+
+A table additionaly has a group key, which is a list of columns and their values.
+All rows within a table have the same value for all columns that are part of the group key.
+
+#### Stream
+
+A stream is an unbounded set of tables.
+Each table must have a different group key.
+
+#### Input
+
+An input is a source of a stream.
+
+Inputs are reprented as functions that produce a stream.
+
+#### Output
+
+An output consumes a stream and send the data to an external system.
+
+Output are reprented as functions that consume a stream and produce a stream.
+Outputs are transparent proxies, passing their input data downstream unmodified.
+As a result outputs themselves have no effect on the Flux program, in other words they only have side effects.
+
+#### Transformation
+
+A transformation is not a data type but rather an operation defined on a stream.
+A transformation consumes a stream of tables and produces a new stream of tables.
+
+Transformations may make use of a data cache to allow for multiple edits to a table before materializing the table into the new stream.
+
+Transformations are reprented as functions that consume a stream and produce a stream.
+
+#### Data cache
+
+A data cache is a set of tables that have yet to be materialized.
+Tables in a data cache may receive edits.
+The cache identifies each table by its group key.
+
+#### Triggers
+
+A trigger indicates that a table in the data cache should be materialized and delivered downstream.
+
+A trigger can fire multiple times until it is finished.
+Once a trigger is finished it will never fire again.
+
+A data cache uses triggers to know when it can evit and materialize tables.
+
+Triggers fire based on events.
+For example a trigger may fire after receiving an update watermark event.
+See the Flux spec for complete details on events available to triggers.
+
+#### Message bus
+
+The message bus is a queue where trigger events and data messages can be placed.
+Each transformation has its own message bus.
+A transformation consumes messages off its bus and processes them.
+
+### Data flow
+
+With the various data types and operations defined lets be explicit about how data flows through a Flux program.
+
+A Flux program describes a graph through which data flows.
+This graph may have complex data dependencies between transformations.
+Sources produce an inital stream of tables, each table is placed on the message bus for the downstream transformations of the source.
+Each transformation is linked to its downstream transformations via the message bus.
+Sources produces trigger events about the stream which are also placed on the message bus.
+The trigger events cause tables to be materialized from the data cache, if it exists. 
+The materialized tables are placed on the message bus of the downstream transformations.
+If a transformation does not need to make edits to a table it does not need to use the data cache but can simply place its generated tables directly on the downstream message bus.
+
+## Compiler
+
+The compiler is responsible for transforming a semantic graph into an IR.
+Internally the compiler performs several passes in order to produce optimized IR.
+
+1. Initial Pass - Convert the semantic graph into non-optimized IR.
+2. Planner - The planner rewrites the IR to optimize for the structure of the Flux script.
+3. Final Pass - The final pass rewrites the IR for any final optimizations that can be performed.
+
+
+The planner tends to perform more macro optimizations, like merging entire data transformations together.
+The final pass tends to perform more micro optimizations, like rewriting a few instructions to allow for better cache locality.
+
+## Intermediate Representation
+
+The Flux IR consists of a set of instructions and arguments to those instructions.
+The IR is a high level IR, specifically instructions are not required to directly map to hardware operations.
+
+The data types the IR can describe are also high level.
+For example an instruction may take an entire table as its input argument.
+
+The IR describes the both the control flow and the data dependencies between function calls.
+The IR is represented as imperative code.
+As such data dependencies are encoded in the order of the instructions and control flow is encoded through condition jumps to different sections of code.
+
+Here are a few examples of the IR instructions:
+
+* AddInt - Add two integers
+* Stddev - Compute the stddev of a table of data
+* Join - Joins N streams of tables
+
+The IR is not a language to describe how to process the data with tables.
+In other words the IR is not used to describe the implementation of the transformations themselves rather it is used to describe the data flow between transformations.
+The implementations of the transformations is left to the virtual machine runtime.
+
+## Virtual Machine
+
+The virtual machine will consume a list of IR instructions and execute them.
+Many instructions will call out to the Flux runtime.
+
+### Runtime
+
+The Flux runtime provides the heavy data processing for the virtual machine.
+
+The following operations will be implemented in the runtime.
+
+* Transformations
+* Data cache
+* Message bus
+
+
+## Debugger
+
+Flux will provide a debugger to allow a user to step through their Flux source code.
+The compiler can be instructed to generate debug IR.
+This IR will allow a debugger to step through the function calls of a Flux program and inspect the data being passed.
+Since the IR describes data flow and conditional logic it is the correct abstract to debug a Flux program since the details of how a transformation is implemented should not be the concern of a user.
+Large amounts of data can be passed between functions, the debugger will expose ways to page through the data in a streaming manner.
+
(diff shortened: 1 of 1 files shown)


docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+As a result outputs themselves have no effect on the Flux program, in other words they only have side effects.
+
+#### Transformation
+
+A transformation is not a data type but rather an operation defined on a stream.
+A transformation consumes a stream of tables and produces a new stream of tables.
+
+Transformations may make use of a data cache to allow for multiple edits to a table before materializing the table into the new stream.
+
+Transformations are reprented as functions that consume a stream and produce a stream.
```
wolffcm: Some transformations may consume more than one stream, such as `join` and `union`.  Maybe there are others, too?

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### Data cache
+
+A data cache is a set of tables that have yet to be materialized.
```
wolffcm: I find this statement confusing.  I think of "materialization" as happening anytime a table is stored with all its data accessible.  As a counterexample, in a sequence of operations that can be pipelined such that they operate on a single row at a time, I would say the table would *not* be materialized between such operations.

By "materialized" do you mean made available to downstream transformations?

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+
+A data cache uses triggers to know when it can evit and materialize tables.
+
+Triggers fire based on events.
+For example a trigger may fire after receiving an update watermark event.
+See the Flux spec for complete details on events available to triggers.
+
+#### Message bus
+
+The message bus is a queue were trigger events and data messages can be placed.
```
wolffcm: Typo: were -> where

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+The planner tends to perform more macro optimizations, like merging entire data transformations together.
+The final pass tends to perform more micro optimizations, like rewriting a few instructions to allow for better cache locality.
+
+## Intermediate Representation
+
+The Flux IR consists of a set of instructions and arguments to those instructions.
+The IR is a high level IR, specifically instructions are not required to directly map to hardware operations.
+
+The data types the IR can describe are also high level.
+For example an instruction may take an entire table as its input argument.
```
wolffcm: Or even a stream of tables, right?

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+
+The IR describes the both the control flow and the data dependencies between function calls.
+The IR is represented as imperative code.
+As such data dependencies are encoded in the order of the instructions and control flow is encoded through condition jumps to different sections of code.
+
+Here are a few examples of the IR instructions:
+
+* AddInt - Add two integers
```
wolffcm: I would think that in an imperative IR, most instructions would involve an assignment to a variable, except for stuff like control flow.  Also, I think it will simplify planning greatly if function calls can have other function calls as arguments.

Maybe this is your intention, and this is question of defining terms like instruction, statement, expression, etc?

docs/VirtualMachine.md
nathanielc: I am not sure about this one, I was thinking that we let the virtual machine direclty manage how data flows from transformation to transformation. So the largest unit an instruction would consume would be a table.

wolffcm: Interesting.  This suggests that the IR would at least need to be able to do the equivalent of:
```
for each table in a stream {
    <call some function on a table> 
}
```
So there would need to be control flow out of the gate.  I wonder if the control flow statements that eventually make their way into Flux would be inspired by what is first only implemented in IR.

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+* Stddev - Compute the stddev of a table of data
+* Join - Joins N streams of tables
```
jlapacik: So IR instructions can take variable length arguments?

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+
+A Flux script goes through several transformations from text to execution.
+
+1. Flux source code is created via any means.
+2. The parser produces an abstract syntax tree (AST) from the source code.
+3. The semantic analyzer produces a semantic graph from the AST.
+4. The compiler accepts the semantic graph and produces an intermediate representation (IR) to be executed.
+5. The flux virtual machine executes the IR.
+
+NOTE: Should we call it bytecode or IR?
```
pauldix: I think they're kind of interchangeable, but if the <thing> is going to be optimized by the compiler/planner, then I think maybe IR is best

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+Should we call it a VM or an interpreter?
```
pauldix: VM if it's an actual virtual machine. To me interpreter means that it's walking the AST and executing along the way.

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+Columns have a label and a type.
+The table has zero or more rows, where each row defines a value for each column or a null is used to indicate a missing value.
+
+A table additionaly has a group key, which is a list of columns and their values.
+All rows within a table have the same value for all columns that are part of the group key.
+
+#### Stream
+
+A stream is an unbounded set of tables.
+Each table must have a different group key.
```
pauldix: Is this strictly true? If the stream is unbounded, you could have the same table/schema/group key come up multiple times during different time frames. Like if you have group keys A & B and you have data that is streaming in from T 1-10, but you get in batches of 5. Then in the stream you'd have `A,T1-T5 : B,T1-T5 : A,T6-T10 : B,T6-T10`

docs/VirtualMachine.md
pauldix: Might also have transformations that split a single stream into N streams.

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+Sources produce an inital stream of tables, each table is placed on the message bus for the downstream transformations of the source.
+Each transformation is linked to its downstream transformations via the message bus.
+Sources produces trigger events about the stream which are also placed on the message bus.
+The trigger events cause tables to be materialized from the data cache, if it exists. 
+The materialized tables are placed on the message bus of the downstream transformations.
+If a transformation does not need to make edits to a table it does not need to use the data cache but can simply place its generated tables directly on the downstream message bus.
+
+## Compiler
+
+The compiler is responsible for transforming a semantic graph into an IR.
```
pauldix: Would it make sense in the future to have the compiler written in Flux? I'm thinking of when we want to create Flux implementations in other languages. Then we'd only need to write the parser and the VM in the target language.

docs/VirtualMachine.md
```diff
@@ -0,0 +1,160 @@
...
+Internally the compiler performs several passes in order to produce optimized IR.
+
+1. Initial Pass - Convert the semantic graph into non-optimized IR.
+2. Planner - The planner rewrites the IR to optimize for the structure of the Flux script.
```
pauldix: I'm thinking #2 and #3 would ideally be written in Flux (or maybe in the IR?). #1 could still be in Go.

docs/VirtualMachine.md
nathanielc: We do not support nor have we spec'ed out how unbounded streams will work. So according to the current Flux spec this is true. But as you say we will need to workout what this means for a streaming use case.

docs/VirtualMachine.md
stuartcarnie: I'd call it a VM, as the source code must be compiled to some intermediate representation / byte code before it can be executed. Conceivably, I could create DSL "foo" that compiles to Flux IR and the VM would execute my code.

docs/VirtualMachine.md
stuartcarnie: I'm not sure complex instructions like `Stddev` and `Join` should be represented as IR instructions. My expectation is that these are `Call` instructions to a runtime library. Thus the VM has an associated "kernel" of standard APIs that we can extend over time.

docs/VirtualMachine.md
nathanielc: The VM we are talking about building will not be capable of implementing a compiler, therefore Flux will not be a self-hostable language. That said we can revisit that design later as any direction towards a VM based implementation of Flux allows for a future self-hostable language.

To be specific as to some things that are missing from the Flux and the VM that would mean writing the compiler in Flux is difficult.  (Not an exhaustive list, just some initial thoughts)

* No memory management, we do not want to write a garbage collector or otherwise manage memory. The only memory management that is going to happen is around tables as units of memory which means that  queries can be well controlled and isolated, but arbitrary code will not. I expect the predicate and map functions will just have one memory free at the end of their execution. 
* Flux's type system is  too small to easily represent things like an AST. Flux has no named types for example.
* Similar to the type system Flux has only limited dynamic data structures, i.e a tables and streams. Objects, arrays and primitive values are all static, meaning they cannot grow. These dynamic data structures would be needed to write a compiler in Flux.


TL;DR Flux is a small language, adding the features to it to make it self-host the language would greatly increase the scope of the language.  That said moving towards a VM moves use towards being able to self-host Flux in the future.

docs/VirtualMachine.md
aanthony1243: conceptually, to get any meaningful output from a query on an unbounded stream there must be some checkpoint where in the context of the query checkpoint/trigger, all tables are finite and complete.  

then there are 2 definitions to think about  for an "unbounded"  stream.  
1.  unless the data itself defines it (e.g. all group keys are boolean)  you must assume that at any time a new table may appear.

2.  are aggregators, selectors, etc defined over the life of the query, since the last checkpoint, or something else completely?

docs/VirtualMachine.md
```diff
@@ -0,0 +1,269 @@
...
+        tmp0 = Eq row._measurement "cpu" # handwaving how object property access happens
+        tmp1 = Eq row._field "usage_idle" # handwaving how object property access happens
+        tmp2 = And tmp0 tmp1
+        Return tmp2
+
+    meanLoop:
+        mbr = MsgRange $0 mean # Call out to intrinsic "mean" function for each table on the message bus
+        Return mbr
+
+
```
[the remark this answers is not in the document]
nathanielc: @wolffcm This is great! We need to make it easy for the planner to understand and optmize the IR ast the same level of abstraction that the DAG provided. I like this idea of a `mat` instruction that needs to be lowered before execution, i.e step 3 of the compiler, but the planner can use it as its atomic unit of rewriting the IR.

Thanks for the detailed write up and examples.

nathanielc:
I am going to merge this rough design. Then I am going to update the SPEC's wording to reflect this change. Most of the details here are implementation specific but some things should be called out in the SPEC. I'll work on a PR for that.

Issue #302, opened 2018-11-19
Range ignored if put after pivoting operation


affo:
The query:

```
from(bucket: "telegraf")
    |> pivot(rowKey:["_time"], columnKey: ["_field"], valueColumn: "_value")
    |> range(start: -1h)
```

raises the error:

```
failed to create physical plan: result 'from0' is unbounded. Add a 'range' call to bound the query,
```

While:

```
from(bucket: "telegraf")
    |> range(start: -1h)
    |> pivot(rowKey:["_time"], columnKey: ["_field"], valueColumn: "_value")
```

Executes normally.

Could be related to https://github.com/influxdata/flux/issues/139.

nathanielc:
The `fromRows` builtin use range last which causes this error. We need to change fromRows to do the range first.

nathanielc:
Can a range push trhough a pivot? If so we need to add the rewrite rule for it. If not then we need to update the `fromRows` function to add its own range call.

jlapacik:
I don't believe range can be pushed through a pivot according to the current pivot spec. The column schema that pivot produces depends on the number of unique values in pivot's `columnKey`. The more unique values, the more columns the output tables will have.

A range before a pivot will restrict the number of unique values for the `columnKey`. A range after pivot however could potentially produce null columns (columns with only null values) - something that cannot occur if range comes before pivot.

The only way I see that `range |> pivot` could be equivalent to `pivot |> range` is when pivot's `rowKey` contains the `_time` column **and** if pivot were to strip any null columns from its output by default.

@aanthony1243 do you agree? Am I leaving anything out?

affo:
I see your point @jlapacik.

Just to be sure that I got the point with `null` values, let me provide an example. Please correct me if some assumption is wrong.

`null` values can happen only when some point is missing a value for a `rowKey`, `columnKey` pair.
So, for instance, suppose the `rowKey` is `_time` and the `columnKey` is `_field`, it could happen that a point with a new field causes `null`s in every previous point for that field. Ranging and excluding that point would cause a `null` column.

Correct?

Now, some considerations:

Given that `pivot` is very general, it could produce any kind of "weird" result (suppose you specify `_value` as `columnKey` or `rowKey`...). So, I agree with you, `pivot` is too general to be swapped a priori with `range`.

[...]

PR #307, opened 2018-11-20
Dropping fromRows for toRows

affo:
Related to #302.

`fromRows` is buggy, it needs to be split into `from` and `toRows`.


functions/transformations/transformations_test.go
```diff
@@ -22,6 +22,7 @@ var skipTests = map[string]string{
 	"string_interp":             "string interpolation not working as expected in flux (https://github.com/influxdata/platform/issues/404)",
 	"to":                        "to functions are not supported in the testing framework (https://github.com/influxdata/flux/issues/77)",
 	"task_per_line":             "bug in group by caused by heterogeneous table schemas (https://github.com/influxdata/flux/issues/100)",
+	"fromRows":                  "buggy, dropping support (https://github.com/influxdata/flux/issues/302)",
```
nathanielc: Can we fix this test as part of this change?

nathanielc:
>  I think this is OK but toRows is not generally applicable except for results from influx storage. it will probably be a good fit once we have namespaces. should we keep for now?

I think this should inform what we name it. And yes when we have namespaces we will want to place it in an influx specific namespace.

We can pick a descriptive name for it now, then when we have namespaces we can rename it accordingly. For example `influxToRows` could become `influx.toRows`, or maybe we use `pivot`, `influxPivot` or `influx.pivot` when we have namespaces. Thoughts? Just brainstorming here, but we definitely need a better name.

affo:
I opened an issue about a `select` statement and its relation with `toRows`. What do you guys think?

https://github.com/influxdata/flux/issues/312

Could `select` be an appropriate name? I thought about this because `toRows` performs a IFQL-style `SELECT *`.

affo:
`select` implementation is postponed.

For a better name, here are my proposals:
 - `transpose` (very algebric)
 - `rows` (without "to" it seems less of an output)
 - `lines` (recalls the line protocol)
 - `pivotField` (describes what it does)

The name will be prefixed with `influx`.

@aanthony1243 can you comment on this?

aanthony1243:
maybe...   `influxFieldsByRow` ?   the default is fields-by-column

affo:
What do you think of `influxFieldsAsCols`?

It seems much clearer to me, don't you think?  
By default, field values are stored in rows 😄 

> For now, I left the .csv files for fromRows and skipped the tests, because fromRows leaves _start and _stop fields while toRows not.  
I saw that they are used in window, for instance. Should I keep them?  
Can I suppose that different fields for the same point have the same triple _start, _stop, _time?  
If so I could just use the 3 of them as rowKey in the pivoting operation.

Can you comment on this?

aanthony1243:
`influxFieldsAsCols`  sounds good to me.  

if `_start` and `_stop` are dropped by pivot, it's because they are not part of the table key.  If you want them retained by the `pivot` call, then add a `group(by: ["_start", "stop", "_measurement", "_field"])`  before the pivot.  

I don't know if `fromRows`  will ever come back exactly as it was.  we can probably drop the tests.

aanthony1243:
I think @nathanielc thumbs up is good enough, I dismissed his review this is OK to merge!

affo:
Should I also update a SPEC, or should I file an issue as a reminder for the next release?

aanthony1243:
@affo yes, submit a PR to update the spec please.

Issue #306, opened 2018-11-19
ResultIterator should be canceled when parent context is done


mark-rushakoff:
In our implementations in platform to handle tasks, we have code that could be simplified to roughly this:

```go
func ExecuteQuery(ctx context.Context, req *query.Request) {
	var svc query.QueryService = getQueryService()
	it, err := svc.Query(ctx, req)
	if err != nil {
		return err
	}
	defer it.Release()

	for it.More() { // it.More may block
		res := it.Next()
		if err := doSomethingWith(res); err != nil {
			return err
		}
	}
	return nil
```

We want this function to return when the passed-in context is canceled, but `it.More` may be in a blocking call.  See: https://github.com/influxdata/flux/blob/14b3100e9ede054cb420c5ea4f0880e539c9986a/result_iterator.go#L51

The temporary fix that I would like to apply is:

```go
func ExecuteQuery(ctx context.Context, req *query.Request) {
	var svc query.QueryService = getQueryService()
	it, err := svc.Query(ctx, req)
	if err != nil {
		return err
	}
	defer it.Release()

	// NEW CODE: Handle context closed on separate goroutine by releasing iterator:
	go func() {
		<-ctx.Done() // In this code path we know the context is always eventually canceled.
		it.Release() // ResultIterator godoc specifies: "It is safe to call Release multiple times."
	}()
	// END NEW CODE

	for it.More() { // it.More may block
		res := it.Next()
		if err := doSomethingWith(res); err != nil {
			return err
		}
	}
	return nil
```

However, the ResultIterator godoc also specifies:

>  all of the methods are expected to be called within the same goroutine.

And for platform's http.QueryService, calling Release from two goroutines causes a data race. I would prefer to see that constraint relaxed specifically for the Release method, but perhaps that is out of scope with context-aware result iterators.

I think the simplest additional API to the ResultIterator interface would be a `WithContext(context.Context) ResultIterator` method, which returns a new ResultIterator that respects the passed-in context's Done method; if ctx.Done() is closed, then:

- subsequent calls to ResultIterator.More() will return false
- the Result returned by an in-progress call to ResultIterator.Next() is valid, but its TableIterator will not return any new Tables after the current Table.
- ResultIterator.Err() will return ctx.Err()

I would also be satisfied with a new `ContextAwareResultIterator` type with method signatures that accept Context:

```go
type ContextAwareResultIterator interface {
    More(context.Context) bool
    Next(context.Context) Result
    Release()
    Err() error
[...]

PR #311, opened 2018-11-20
releasing ResultIterator resources is thread-safe

jlapacik:
Related to #306 .

The ResultIterator godoc specifies that all of its methods are expected to be called within the same goroutine. However for one of its implementations, `queryResultIterator`, the `More()` method could become blocked:
https://github.com/influxdata/flux/blob/14b3100e9ede054cb420c5ea4f0880e539c9986a/result_iterator.go#L51

The `Release()` method in this case is intended to unblock `More()`. But this is only possible if `Release()` is called from a goroutine separate from the one in which `More()` was called. This constraint should probably be relaxed as to allow blocking calls to `More()` to become unblocked.

### DIFF
--- a/result_iterator.go
+++ b/result_iterator.go
@@ -6,7 +6,10 @@ import (
 
 // ResultIterator allows iterating through all results synchronously.
 // A ResultIterator is not thread-safe and all of the methods are expected to be
-// called within the same goroutine. A ResultIterator may implement Statisticser.
+// called within the same goroutine with the exception of the Release method.
+// Release is the only method that is thread-safe and may be called from a
+// goroutine different than that of the other methods. A ResultIterator may
+// implement Statisticser.
 type ResultIterator interface {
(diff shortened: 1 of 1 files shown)


jsternberg:
As an FYI, the `Release()` isn't meant to unblock `More()`. The only thing that can cancel it is if the original query that is passed in has its `Cancel()` method called or if the original context is canceled.

This was to differentiate the `ResultIterator` interface from the `Query` interface. `ResultIterator` would be used in any contexts that were synchronous and the asynchronous `Query` interface would be used in other situations. Previously, the method was vaguely defined of exactly what the old `Cancel()` method should do so I thought it would be best to add a godoc comment and make it explicit. This way, it also stresses that `Release()` has to be called even if `More()` returns false. The old one implied you didn't have to call `Cancel()` if `Done()` returned false, but this didn't interface correctly with the controller.

Issue #263, opened 2018-11-09
end-to-end test fails on CI occasionally


jlapacik:
CI will catch the following error occasionally using `go test -race`.

    --- FAIL: Test_QueryEndToEnd (44.04s)
    --- FAIL: Test_QueryEndToEnd/selector_preserve_time.flux (0.01s)
        failed to run query: failed to compile query: type error 4:5-4:23: cannot unify object with semantic.Nature

It has only happened to me twice but both times it occurred on the above end-to-end test.

mark-rushakoff:
Reproducible with `go test -count=100 -run=Test_QueryEndToEnd/selector_preserve_time.flux ./functions/transformations/` -- `-race` isn't necessary. Usually have to run it a few times to get it to fail.

I put a panic where the error was originating, and got this stack trace from master @ ab39f6de2c4df52e874cdb9f2463ff9cbde18ab7:

```
goroutine 259 [running]:
testing.tRunner.func1(0xc0049c4000)
	/Users/mr/gotip/src/github.com/golang/go/src/testing/testing.go:815 +0x387
panic(0x17b8d80, 0x19b6f00)
	/Users/mr/gotip/src/github.com/golang/go/src/runtime/panic.go:513 +0x1b7
github.com/influxdata/flux/semantic.object.unifyType(0xc003d1dad0, 0x1f53da8, 0x0, 0x0, 0xc003d245f0, 0x1, 0x1, 0xc0051780f0, 0x19ead00, 0xc005183ff0, ...)
	/Users/mr/go/src/github.com/influxdata/flux/semantic/poly_types.go:574 +0x2f0
github.com/influxdata/flux/semantic.unifyTypes(0xc0051780f0, 0x19eaee0, 0xc003d28900, 0x19ead00, 0xc005183ff0, 0x1, 0x0, 0x0)
	/Users/mr/go/src/github.com/influxdata/flux/semantic/solve.go:146 +0x4f
github.com/influxdata/flux/semantic.function.unifyType(0xc003d1daa0, 0x1f53da8, 0x0, 0x0, 0x19eaee0, 0xc003d28940, 0x18d85d5, 0x6, 0xc0051780f0, 0x19eae20, ...)
	/Users/mr/go/src/github.com/influxdata/flux/semantic/poly_types.go:384 +0x631
[...]

Issue #27, opened 2018-09-12
Introduce Schema introspection function


nathanielc:
It will be valuable for operations to be able to inspect the table schema on the fly. 

We propose adding a standard object type for doing so. The `schema` object will be as follows:

```
type schema = {
    colums: []column // list of all columns
    grouped:  []string // list of only the group labels
}
type column = {
   label : string, // column label
   type: string, // name for the column type
   grouped: bool, // whether the column is part of the group key
}
```

Flux functions can then make use of the schema record for defining dynamic behaviors on the schema

Take for example the `group` function that wants a list of column labels.

A user could write this to add a column to the list of grouped columns without needing to know the existing group columns

```
group(by: (schema) => schema.grouped + ["host"])
```

jsternberg:
There's a typo in the schema. It should be `columns`.

I also think that `groupKeys` or `keys` are better names than `grouped` and that the column boolean attribute should be `isKey`. Those are just preferences though.

aanthony1243:
new question:  in #28   we propose passing schema.columns   to a filter()  function.    This would require overloading the existing filter()  function, or else  having a filterList()  function or similar.

PR #313, opened 2018-11-21
Fix in GroupBy spec

affo:
Fixed formatting and added `all` and `none` parameters for `group` operation.  
I don't know if I got well the meaning of `group(all: true)` is it really equivalent to not grouping at all?

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1237,6 +1237,7 @@ The tables schema will include the following columns:
     the exclusive upper time bound of all records
 
 Additionally any tags on the series will be added as columns.
+The default group key for the tables is every column except `_time` and `_value`.
 
 
 From has the following properties:
@@ -2069,32 +2070,62 @@ It produces tables with new group keys based on the provided properties.
 
 Group has the following properties:
 
-*  `by` list of strings
-    Group by these specific columns.
-    Cannot be used with `except`.
-*  `except` list of strings
-    Group by all other columns except this list.
-    Cannot be used with `by`.
+*  `columns` list of strings
+    List of columns used to calculate the new group key.
+    The default is `[]`.
+*  `mode` string
+    The grouping mode, can be one of `"by"` or `"except"`.
+    The default is `"by"`.
+    
+When using `"by"` mode, the specified `columns` are the new group key.  
+When using `"except"` mode, the new group key is the difference between the columns of the table under exam
+and `columns`.  
 
-Examples:
-    group(by:["host"]) // group records by their "host" value
-    group(except:["_time", "region", "_value"]) // group records by all other columns except for _time, region, and _value
-    group(by:[]) // group all records into a single group
-    group(except:[]) // group records into all unique groups
+__Examples__
+
+_By_
 
 ```
 from(bucket: "telegraf/autogen") 
     |> range(start: -30m) 
-    |> group(by: ["host", "_measurement"])
+    |> group(columns: ["host", "_measurement"])
 ```
-All records are grouped by the "host" and "_measurement" columns. The resulting group key would be ["host, "_measurement"]
+
+Or:
+
+```
+...
+    |> group(columns: ["host", "_measurement"], mode: "by")
+```
+
+Records are grouped by the `"host"` and `"_measurement"` columns.  
+The resulting group key is `["host", "_measurement"]`, so a new table for every different `["host", "_measurement"]`
+value is created.  
+Every table in the result contains every record for some `["host", "_measurement"]` value.  
+Every record in some resulting table has the same value for the columns `"host"` and `"_measurement"`.
+
+_Except_
 
 ```
 from(bucket: "telegraf/autogen")
     |> range(start: -30m)
-    |> group(except: ["_time"])
+    |> group(columns: ["_time"], mode: "except")
 ```
-All records are grouped by the set of all columns in the table, excluding "_time". For example, if the table has columns ["_time", "host", "_measurement", "_field", "_value"] then the group key would be ["host", "_measurement", "_field", "_value"]
+
+Records are grouped by the set of all columns in the table, excluding `"_time"`.  
+For example, if the table has columns `["_time", "host", "_measurement", "_field", "_value"]` then the group key would be
+`["host", "_measurement", "_field", "_value"]`.
+
+_Single-table grouping_
+
+```
+from(bucket: "telegraf/autogen")
+    |> range(start: -30m)
+    |> group()
+```
+
+Records are grouped into a single table.  
+The group key of the resulting table is empty.
 
 #### Keys
 


docs/SPEC.md
```diff
@@ -2067,34 +2067,43 @@ from(bucket:"telegraf/autogen")
...
 
 *  `by` list of strings
     Group by these specific columns.
-    Cannot be used with `except`.
 *  `except` list of strings
     Group by all other columns except this list.
-    Cannot be used with `by`.
+*  `none` boolean
+    Group by nothing, so, group all records into a single group.
+*  `all` boolean
```
nathanielc: Group all means that you are grouping by all columns and as such each unique row will be in its own group.

docs/SPEC.md
```diff
@@ -2067,34 +2067,43 @@ from(bucket:"telegraf/autogen")
...
-    group(by:["host"]) // group records by their "host" value
-    group(except:["_time", "region", "_value"]) // group records by all other columns except for _time, region, and _value
-    group(by:[]) // group all records into a single group
-    group(except:[]) // group records into all unique groups
+
+*    `group(by: ["host"])`: group records by their `"host"` value.
+*    `group(except: ["_time", "region", "_value"])`: group records by the values in all other columns except for
+     `"_time"`, `"region"`, and `"_value"`.
+*    `group(none: true)`: group all records into a single group
+*    `group(all: true)`: ignores grouping. Equivalent to `group()`.
```
nathanielc: `group()` Should be invalid.

affo: Indeed it works, it defaults to `group(all: true)`: https://github.com/influxdata/flux/blob/master/functions/transformations/group.go#L80

I can avoid the example in the SPEC.

docs/SPEC.md
affo: Oh yes, got it. Indeed, it was pretty obvious 😄 . My doubt arose from the fact that I ran some queries with group all and the result was the same as no grouping, but maybe it is because the data was already grouped by every column by telegraf itself.

Let me be more precise: I issued this query on the `syslog` bucket generated by telegraf:

```
from(bucket: "telegraf")
    |> filter(fn: (r) => r._measurement == "syslog")
    |> range(start: -24h)
    |> group(all: true)
```

And I got (showing only the table header) this:

```
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string,string
#group,false,false,true,true,false,false,true,true,true,true
#default,_result,,,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement,cpu,host
```

Is it true that "all" doesn't include `_time` and `_value`?

affo: Maybe, the best to tell it is like here: https://github.com/influxdata/flux/blob/master/functions/group_mode.go#L11

"GroupModeAll produces a separate table for each series."

stuartcarnie: Also note that the work I did in influxdata/platform#1312 provides new implementations of these strategies when merging multiple remote results sets with additional unit tests.

docs/SPEC.md
nathanielc: @affo that is a bug. Good find. Lets remove it from the SPEC examples and file a bug on the repo.

docs/SPEC.md
affo: Thank you Stuart, I'll update the spec accordingly.

affo: Ok, I needed to run some tests to clarify some bits.

* What `group(none: true)` does is crystal clear;
* For what concerns `group(all: true)`, it performs a little different from what you say (at least from my results):

> the partition key == all the tag values of the series key

Sorry for being pedantic, but I would say that the partition key is composed of the values of `_field`, `_measurement`, and the tag values.  
If the "series key" is composed of the values of `_field` and `_measurement` values, then we are telling the same thing! 😄 

If it is so, I suggest to fix the "Concepts" section in the documentation, because it tells that a series is "collection of data that share a retention policy, measurement, and tag set" (https://docs.influxdata.com/influxdb/v1.7/concepts/key_concepts/#series). It becomes quite natural to think of the "series key" as measurement and tag set, but not field.

docs/SPEC.md
```diff
@@ -2064,37 +2064,72 @@ from(bucket:"telegraf/autogen")
...
+
+_All_
+
+```
+from(bucket: "telegraf/autogen")
+    |> range(start: -30m)
+    |> group(all: true)
+```
+
+Group records by every column.
```
affo: @stuartcarnie 

As you can see, I opted for different levels of abstraction in the description.

affo:
@stuartcarnie @nathanielc 

I don't know why tests failed...
Is this update fine for you?

mark-rushakoff:
```
--- FAIL: Test_QueryEndToEnd (5.10s)
    --- FAIL: Test_QueryEndToEnd/selector_preserve_time.flux (0.00s)
        transformations_test.go:104: failed to run query: failed to compile query: type error 4:5-4:23: cannot unify object with semantic.Nature
FAIL
FAIL	github.com/influxdata/flux/functions/transformations	6.208s
```

That's #263. I'll restart the circle job for you. If you don't have permissions to do that on circle, talk to Nathaniel about getting the right permissions.

affo:
Yep, I don't have permissions. Thank you!

docs/SPEC.md
```diff
@@ -2064,37 +2064,72 @@ from(bucket:"telegraf/autogen")
...
+
+Records are grouped into a single table.  
+The group key of the resulting table is empty.
+
+_All_
+
+```
+from(bucket: "telegraf/autogen")
+    |> range(start: -30m)
+    |> group(all: true)
```
nathanielc: @stuartcarnie Does group by `all` mean something different when it is pushed down into storage vs if it is bare and executed in the Flux engine?

nathanielc: In other words, why does group by all not include the `_value` and `_time` columns? I think we have an inconsistency in generic Flux vs InfluxDB Flux here.

docs/SPEC.md
wolffcm: It's worth noting in this discussion that `group(all: true)` is not implemented at all in the `Process` method and happens to behave the same as `group(none: true)` when evaluated in Flux.  I've been working on this issue to address it:
https://github.com/influxdata/flux/issues/322
So we will likely see differing behavior depending on whether grouping is evaluated in storage or in flux.  Note that if there's several consecutive `group` calls, we will only push the first one to storage, and evaluate the rest in Flux.

@stuartcarnie @affo The way `group(all: true)` is being described here I would expect it to be a no-op, or am I misunderstanding?  I think I agree with @nathanielc that `group(all: true)` should group by all the columns, including `_value` and `_time`.  Maybe we can't push down the grouping into storage for that case though.

When we're all available we should talk about what the right behavior should be for both Flux and storage.

docs/SPEC.md
stuartcarnie: @nathanielc it should not; `group(all:true)` simply produces a table per series

nathanielc: @stuartcarnie As per @wolffcm comment, about `group(all:true)` not being implemented in Flux, but only as a push down. How would I implement it as a Flux transformation? In other words once outside of the database how do I group per "series"? What is a series in that context?

I am thinking that maybe the `group(all:true)` option should not exist at all on the Flux function. But that the default grouping behavior of the `from` function is per series. In the context of InfluxDB is `group(all:true)` the same as `group(except:["_time", "_value"])`? If it is then maybe we remove the whole `group(all: true)` concept.

docs/SPEC.md
aanthony1243: it looks to me that the `flux` syntax was overloaded to encode the storage-level RPC flags for GroupMode.  The semantics should come from there.  the issue with `group(all:true)`  is that it's only well-defined if we have a default set of influxDB storage columns.  instead of re-interpreting it in the `process` method, we should probably return an error instead, reporting that that mode is only compatible when pushed down to an influxdb storage request.  

I had never considered this before, or else I might have pushed for a different function signature.  Something like  group(auto: "none")   to get GroupModeNone  and group(auto: "influxdb-default") to get GroupModeAll.  basically, if there's automagic ways of doing grouping we should put them here.   This could also help us with an existing feature request to additively construct a group key, e.g.,  `group(auto: "concat-existing", by: ["newCol1", "newCol2"])`.  Though that's more of a hack than what we should be doing (schema introspection #27 ).

nathanielc:
OK, after some discussion I think we have come to a definition...

Group will only have three parameters:

* by
* except
* none

The `all` parameter is not needed and should be removed from the SPEC.

nathanielc:
NOTE: We should update the SPEC to document that the default grouping for the `from` function is `group(except:["_time","_value"])`.

docs/SPEC.md
affo: Sorry for the delay.

> It's worth noting in this discussion that `group(all: true)` is not implemented at all in the `Process` method and happens to behave the same as `group(none: true)` when evaluated in Flux. I've been working on this issue to address it:
> #322
> So we will likely see differing behavior depending on whether grouping is evaluated in storage or in flux. Note that if there's several consecutive `group` calls, we will only push the first one to storage, and evaluate the rest in Flux.
> 

This sheds light on the suspicious behavior I detected in #322. When `group(all: true)` is the first group operation it is performed by storage, otherwise it is not performed because flux implements it this way. Great!

> @stuartcarnie @affo The way `group(all: true)` is being described here I would expect it to be a no-op, or am I misunderstanding? I think I agree with @nathanielc that `group(all: true)` should group by all the columns, including `_value` and `_time`. Maybe we can't push down the grouping into storage for that case though.
> 
> When we're all available we should talk about what the right behavior should be for both Flux and storage.

With a rewrite rule to use only the last grouping, it should be performed by storage and fix the problem, at least if it is possible to perform it on storage. If the user specifies transformations and then re-groups, the grouping is performed by flux and the problem araises.

nathanielc:
OK we had more discussion and we found more issues when exploring the implementation here https://github.com/influxdata/flux/pull/338

Basically having two mutually exclusive arguments caused validation issues and ambiguities in meaning. So instead we propose a design change for how group should behave.

Group will have two arguments

* columns - a list of columns - no default
* mode - one of `by` or `expect` - defaults to `by`.

Therefore, the old function becomes this
```
group(none:true) -> group(columns:[])
group(by:["a"]) -> group(columns:["a"])
group(expect:["b"]) -> group(columns:["b"], mode="except")
group(all:true) -> group(columns:[], mode="except")
```

@affo @wolffcm Thoughts?

wolffcm:
I like it.  A few questions:

So is it expected that `group(columns: [])` behave the same as `group()`?  I.e., `group()` would now be legal again.

What should happen with `highestMax`, etc?  It would be possible to leave the public-facing API as-is, or to change it to something like this:
```
highestMax = (n, columns=["_value"], groupColumns, groupMode="by", tables=<-) => ...
```

aanthony1243:
would it be confusing to have `group(by: ["x"], mode: "by")` ?   This would make a majority of use-cases backward compatible.

wolffcm:
I guess I still have some questions about the rules for named function parameters.  Thinking about it more, here's my understanding now:

We don't have optional parameters per se, we just have parameters for which there are default values.  If you leave out a parameter, the function will be called with the parameter's default value.

However, not all parameters have a default value, and if you leave out a parameter that doesn't have a default value, that is an error.  In other words, parameters do not default to the zero-value if the function declaration does specify a default value for the parameter.  (I had been implicitly making this assumption)

wolffcm:
@aanthony1243 It would be nice to maintain backward compatibility (and less work to not have to update tests etc), but I think in the long run it would be confusing.

nathanielc:
> @aanthony1243 It would be nice to maintain backward compatibility (and less work to not have to update tests etc), but I think in the long run it would be confusing.

Now is the time to break it. Which means we should break highestMax etc as well to keep them in sync with group.

@wolffcm Your summary of parameters is correct. There is no implicit default of the zero value.

nathanielc:
I almost forgot.,,

> So is it expected that group(columns: []) behave the same as group()? I.e., group() would now be legal again.

I was thinking that columns has no default, and therefore `group()` is invalid, meaning that grouping by nothing is explicit. I can be convinced otherwise but I prefer explicit over implicit.

wolffcm:
> > So is it expected that group(columns: []) behave the same as group()? I.e., group() would now be legal again.
> 
> I was thinking that columns has no default, and therefore `group()` is invalid, meaning that grouping by nothing is explicit. I can be convinced otherwise but I prefer explicit over implicit.

Yes, I agree too now that I've a better understanding of default values.

affo:
I like the new arguments

I am missing why

> validation issues

? @nathanielc 

There was something wrong or is it more about ambiguity?  
I read the discussion, but I couldn't track the problem in validation.

nathanielc:
The validation issue comes down to not being able to tell the difference between a parameter having its default value set and parameter be passed with its default value.

In the old args is this valid `group(by:[], except:[])`? What about this `group(by:["a"], except:[])`?

In Flux  this `group(by:["a"], except:[])` is equivalent in every way to this `group(by:["a"])`. 

So making by and except mutually exclusive parameters is not really possible. All you can really do is set defaults. As a result it was not possible to determine a difference between `group()` and `group(by:[], except:[])` and we want both to be invalid since its not clear whether that should group by all or none. 

So changing the args to the new system avoid all these issues.

affo:
Crystal clear, thank you!  
I've lost the point of defaults, but now I understand.

The new definition is elegant, simple, and fixes the issue. Great!

Tomorrow I'll update the spec.

nathanielc:
Another detail, group came up in the demo discussion. We decided that allowing `group()` to be used for group by none is a good pattern. This means that `columns` should have a default value of `[]`

docs/SPEC.md
```diff
@@ -2064,37 +2067,67 @@ from(bucket:"telegraf/autogen")
 ```
 #### Group
 
-Group groups records based on their values for specific columns.
+_Group_ groups records based on their values for specific columns.
```
nathanielc: Nit, we do not use `_` for any other function so lets not do it here.

docs/SPEC.md
```diff
@@ -1250,6 +1250,9 @@ Example:
 
     from(bucket:"telegraf/autogen")
     from(bucketID:"0261d8287f4d6000")
+    
+__NOTE__  
+The default group key for the tables produced by `from` is every column except `_time` and `_value`.
```
nathanielc: Can you move this line up with the rest of the description of the `from` function? That way it does not need to be its own NOTE but rather is just part of the normal description of the `from` function.

nathanielc:
@affo Please rebase and use the conventional commit message before merging. Otherwise this is good to go.

Issue #286, opened 2018-11-14
Programmatically updating options in Flux source


mark-rushakoff:
To support an easier REST API and UI for tasks, there is a desire for taking an existing Flux script and a new option or set of options to produce a new Flux script.

For example, given a Flux script like:

```
option task = {
  every: 1m,
  delay: 5s, // Sometimes data comes in late, but never more than 5 seconds late.
  name: "mytask"
}

from(bucket: "x") |> range(start:-1h) |> max() |> to(bucket: "max_over_last_hour")
```

Then with a request to update the "every" option to "20m", the resulting Flux script would be:

```
option task = {
  every: 20m,
  delay: 5s, // Sometimes data comes in late, but never more than 5 seconds late.
  name: "mytask"
}

from(bucket: "x") |> range(start:-1h) |> max() |> to(bucket: "max_over_last_hour")
```

That is, the resultant script has identical comments and formatting, and the only change was the value of the `every` option was changed from 1m to 20m.

I'm not sure exactly what the flux API needs to be for this. Are options only overwritten? Should there be a way to delete options?

nathanielc:
@mark-rushakoff The way I envision these kinds of workflows is to leverage the AST.

The process would look something like this:

```
sourceAST = parser.Parse(source)

// Walk the AST and edit its options
editor := new(OptionEditor)
ast.Walk(editor, sourceAST)

newSource := ast.Format(editor.AST)

// do what you need with the new source code
```

It is also possible to perform a similar work flow using the semantic graph, but that should only be needed if you need the type information in order to perform your edits.

NOTE: The ability to convert AST to text is not available yet.

mark-rushakoff:
https://github.com/influxdata/platform/issues/1996 is the issue tracking the changes in platform's task code, which will use the related changes coming out of #323.  Once https://github.com/influxdata/platform/issues/1996 is done then I think we can close this issue. It's possible that there may be more Flux iterations needed, but currently anticipating that the existing changes will suffice.

mark-rushakoff:
Looks like we were able to do what we needed to do in https://github.com/influxdata/influxdb/pull/10954, so closing this.

PR #323, opened 2018-11-26
AST format & option editor

affo:
Issue: https://github.com/influxdata/flux/issues/286

* Followed the approach of JSON conversion for string formatting
* Implemented `Walk` for `ast` (mostly copy/paste from `semantic`)
* Implemented `OptionEditor`


ast/walk.go
```diff
@@ -0,0 +1,180 @@
+package ast
+
+import "fmt"
+
+func Walk(v Visitor, node Node) {
+	walk(v, node)
+}
+
+type Visitor interface {
```
nathanielc: Can you add a comment describing how visitors work?

ast/format_test.go
```diff
@@ -0,0 +1,74 @@
...
+			}
+
+			got, err := semantic.New(newProgram)
+			if err != nil {
+				t.Fatal(errors.Wrapf(err, "original program is not a valid flux query: %s", stringResult))
+			}
+
+			if !cmp.Equal(want, got, semantictest.CmpOptions...) {
+				t.Errorf("to string conversion error:\nin:\t\t%s\nout:\t\t%s\n", content, stringResult)
+			}
```
nathanielc: Can you remove the testing code around the semantic package? We want these tests for be focused on testing ast.Format only. This test should compare that the input file is the same as the formatted output.

ast/format.go
```diff
@@ -0,0 +1,301 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+func Format(n Node) string {
+	return formatNode(n)
+}
+
+func formatChildren(children interface{}, sep string) string {
```
nathanielc: Can you update the format implementation to also use the ast.Walk/Visitor paradigm? This way you can avoid the use of reflection and not need to have a psuedo walk implementation as part of the format method.

One way to do this is to use many different visitor types, one for each of the methods that you already have here. 

Additionally can you use the `strings.Builder` type instead of many calls to `fmt.Sprintf`? This will be significantly faster and should help using the walk paradigm as the builder can be passed to children nodes as you walk.

nathanielc:
See https://github.com/influxdata/flux/pull/287 for good comments on the ast.Walk types.

ast/format.go
affo: Ok, sounds good. I was thinking of it, but I decided to follow an approach similar to JSON. However, what you are saying has much sense to me. Thank you.

affo: I spent the day on this, but it seems that visiting to extract the query is quite challenging.

The problem lies in the fact that every node must wait for its children to complete their piece of query before concatenating their results (or wait that they append their result to the StringBuilder).

In order to do this, I added hooks to the visitor in order to be called once a child has finished and also to produce different visitors for each child.  In this way the parent visitor can produce the correct visitor for each children and append pieces of string before and after children do it.

For example, an arrow function needs a visitor that:
 - appends a `(` on visit;
 - lets `Params` append their result using a proper visitor for `PropertyExpression`;
 - appends `)=>`;
 - lets `Body` append its result using a proper visitor for `Node`.

The resulting code becomes much more complex (and less readable) than the one I have already pushed. Plus, the Visitor interface becomes more complex.

I haven't yet tested the current implementation, but, before spending more time on this, do you think it would be better if I simply adapt this implementation to another one that uses a StringBuilder?

If you prefer, I can go on and workout a better solution to use the walk-visitor pattern.

ast/format_test.go
affo: I used semantic for two reasons:

 - verify that the query is valid;
 - two statements can be expressed in different ways, but semantically, they mean the same thing. E.g.: `filter(fn: (r) => body)` is the same as `filter(fn: (r) => (body))`, or the float `1.0` is the same as `1.0000`, while they are different strings.

Should I input queries such that I can reproduce an identical output and compare strings?

nathanielc: The above statements are true, but this test doesn't need to test those things, they are tested elsewhere, for example in the semantic package. Specifically the AST can represent an invalid query, but it should still be able to be formatted correctly.

For example:
```
filter(fn: false)
```

That is valid syntax and should be able to be formatted correctly, but it is not a valid query. But this test is only concerned with formatting not whether the script itself makes any sense.

Therefore this test should be testing that the output is as expected, which means we need to test the actual string output of Format. As for your question:


> Should I input queries such that I can reproduce an identical output and compare strings?

There are two ways to do this:

1. Input correctly formatted scripts and ensure that the output is identical
2. Have each test case have an input and output script. The input script can be formatted in the non "standard" way while the output will always be formatted according to the "standard"

I prefer option 2. It is easier to do this if you do not use files as the source data, rather just a list of structs that have two strings. For example you could follow this pattern here https://github.com/influxdata/flux/blob/master/semantic/types_test.go#L15

affo:
I think I did the same thing in the tests for the option editor.

Now I changed the test to compare the AST of the original program against the AST of the one obtained by formatting the first AST. Anyhow, I can change to go with string comparison.

I'll wait for the second pass of the review :)

nathanielc:
@affo I have been playing around with the code getting more familiar with it. I think your original solution is the correct approach. I am now convinced that using ast.Walk is not the correct way to implement the Format method. There are too many places where their need to be hooks for each child. While we could add many more nodes to the AST to solve this, it would create a lot of bloat that would be too noisy. 

Thank you for exploring this.

Can you now use a combination of the two implementations? Specifically:

* Do not use ast.Walk for Format, rather like you had it initially let each format method recurse to its children.
* Do use the string.Builder for the output string, if you have small string concatenations that is not a large concern
* Do not use reflection, if you have to copy paste some code to avoid it that is fine.

Again thank you for exploring this, while a bit circular we have learned a lot about how to best implement this.

affo:
Absolutely no problem!  
It has been interesting to explore that solution and It was also an opportunity for mental stretching 😄 .

affo:
Do we need prettified output?

affo:
Oh, I am sorry!

Ugly:

```
from(bucket:"testdb")|>range(start:2018-05-20T19:53:26Z)|>filter(fn:(r)=>r.name=~/.*0/)|>group(by:["_measurement","_start"])|>map(fn:(r)=>{_time:r._time,io_time:r._value})
```

Pretty:

```
from(bucket: "testdb")
         |> range(start: 2018-05-20T19:53:26Z)
         |> filter(fn: (r) => r.name =~ /.*0/)
         |> group(by: ["_measurement", "_start"])
         |> map(fn: (r) => {_time: r._time, io_time: r._value})
```

I've just implemented the code for obtaining the second version, I am wondering if it is useful/preferable.

nathanielc:
Thanks! The pretty version is what we want. This formatter is going to be setting the defacto standard for how to format Flux code, just like `go fmt` does for Go. 

As for what is pretty? You get the luxury of doing the first pass, so just do something that looks good to you. We can bike-shed on formatting issues later. I don't want this PR to get slowed down by those discussions.

ast/format.go
```diff
@@ -0,0 +1,356 @@
...
+	return f.String()
+}
+
+// strings.Builder's methods never returns a non-nil error.
+func (f *formatter) writeString(s string) {
+	_, err := f.WriteString(s)
+	if err != nil {
+		panic(err)
+	}
+}
```
nathanielc: WriteString is documented to always return a `nil` error. https://golang.org/pkg/strings/#Builder.WriteString Can you update this method to ignore both return values and make a comment about why it is safe?

ast/format.go
```diff
@@ -0,0 +1,356 @@
...
+
+func (f *formatter) writeRune(r rune) {
+	_, err := f.WriteRune(r)
+	if err != nil {
+		panic(err)
+	}
```
nathanielc: Same here

ast/format.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (f *formatter) formatProgram(n *Program) {
+	sep := '\n'
+	for i, c := range n.Body {
+		f.formatNode(c)
+		if i < len(n.Body)-1 {
```
nathanielc: just a nit so not important to change but I find this pattern easier to read.

```
for i, c := range n.Body {
    if i != 0 {
         f.writeRune(sep)
    }
    f.formatNode(c)
}
```

ast/format.go
```diff
@@ -0,0 +1,356 @@
...
+	f.formatNode(n.Test)
+	f.writeRune('?')
+	f.formatNode(n.Consequent)
+	f.writeRune(':')
+	f.formatNode(n.Alternate)
+}
+
+func (f *formatter) formatMemberExpression(n *MemberExpression) {
+	f.formatNode(n.Object)
+	f.writeRune('.')
```
nathanielc: This needs to check if the Property is an Identifier or a StringLiteral. If it a string literal you need to output the `["this is a property"]` format.

ast/format.go
```diff
@@ -0,0 +1,356 @@
...
+	f.writeString(strconv.FormatUint(n.Value, 10))
+}
+
+func (f *formatter) formatPipeLiteral(_ *PipeLiteral) {
+	f.writeString("<-")
+}
+
+func (f *formatter) formatRegexpLiteral(n *RegexpLiteral) {
+	f.writeRune('/')
+	f.writeString(n.Value.String())
```
nathanielc: We need to escape all `/` in the regexp string.

ast/format_test.go
```diff
@@ -0,0 +1,130 @@
...
+var skip = map[string]string{
+	"array_expr":        "without pars -> bad syntax, with pars formatting removes them",
+	"conditional":       "how is a conditional expression defined in spec?",
+	"duration_multiple": "how are multiple duration values meant to be used?",
+}
+
+func TestFormat(t *testing.T) {
+	testCases := []struct {
+		name  string
+		query string
```
nathanielc: These test cases look great!

Small nit, can you rename `query` to `script` since I want the test writer thinking that these could be any Flux scripts, not just queries?

affo:
I hadn't actually implemented any pretty formatting in the code of yesterday! I only had it locally. Today I updated it also to handle proper indentation. Hope you like it 😄

ast/format.go
affo: much better, absolutely

affo:
I have doubts about the current implementation.  
About formatting:

 - I couldn't find how some expressions should be rendered (e.g. logical)
 - I can't distinguish between nil and default values in function definition

Anyways, I put skipped tests for the things that are not clear to me.

About option editing, should the editor specifically target `task` option? At the moment it edits the first OptionStatement it finds.

ast/format_test.go
```diff
@@ -0,0 +1,222 @@
...
+import (
+	"testing"
+
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/parser"
+	"github.com/pkg/errors"
+)
+
+var skip = map[string]string{
+	"array_expr":           "without pars -> bad syntax, with pars formatting removes them",
```
nathanielc: The `()` should not be needed.

ast/format_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"conditional":          "how is a conditional expression defined in spec?",
```
nathanielc: Conditional expressions are not yet defined. We can leave this as skipped for now.

ast/format_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"duration_multiple":    "how are multiple duration values meant to be used?",
```
nathanielc: This test case seems valid. What is the question? Durations should be formatted as largest to smallest duration units.

ast/format_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"nil_value_as_default": "cannot distinguish nil values from default values in function definition",
```
nathanielc: Why not? A `nil` should mean no default value. The `[]` should not be parsing as `nil`. There was a bug about this recently maybe you need to rebase your branch off master to get the fix?

ast/format_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"multi_var_decl":       "how is a variable declaration with multiple declarations represented?",
```
nathanielc: Each variable declaration should be on its own line. Currently the parser has no syntax for reading multiple declarations into a single VariableDeclaration node so we can ignore that case for now.

ast/option_editor.go
```diff
@@ -0,0 +1,83 @@
+package ast
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+	"time"
+)
+
+func NewOptionEditor(keyMap map[string]interface{}) *OptionEditor {
```
nathanielc: Can you use `values.Value` here instead of `interface{}`?

ast/option_editor.go
```diff
@@ -0,0 +1,83 @@
...
+		literal = &RegexpLiteral{Value: val}
+	case Duration:
+		// TODO should we allow to specify more than one duration?
+		literal = &DurationLiteral{Values: []Duration{val}}
+	case time.Time:
+		literal = &DateTimeLiteral{Value: val}
+	default:
+		errString := fmt.Sprintf("ERROR, cannot create literal for %v", reflect.TypeOf(val))
+		literal = &StringLiteral{Value: errString}
+	}
```
nathanielc: Can you make this switch into its own function? One that takes a `values.Value` and produces an ast.Node for its literal value.

affo:
Sorry, the option editor edits _every_ option statement it finds.

Should I add some sort of namespacing to better specify the option that you need to edit?
By now the option editor only takes a `map[string]string`. Maybe a map of map of strings would be more appropriate:

```
{
  task: {every: 1h}
  location: {start: ...}
}
```

For example, if you have:

```
option location = loadLocation(name:"America/Denver")
```

ast/option_editor.go
affo: If I import `values` I get an import cycle: ast -> values -> ast.

Is there any workaround? Maybe externalize types in another module?

ast/format_test.go
affo: After rebase it works, thank you

ast/format_test.go
affo: Now I got it, forget about it 😄

ast/format_test.go
affo: Rechecked, It returns `bad syntax` without parenthesis.

nathanielc:
> And you specify a different name, then you get both names edited. location is not implemented for now... Should we worry about this case?

Yes we need to worry about it.

@affo I think at this point we should do two things:

1. Break this PR into three separate PRs, the ast.Walk bits, the formatting bits, and the option editing bits. This way we can review them separately and get the merged when they are ready. For example both the ast.Walk and ast.Format are good to go
2. We should ping @mark-rushakoff on the option editing PR to ensure that he has input into what the API looks like since he is the one going to be using it.

As for:

> And you specify a different name, then you get both names edited. location is not implemented for now... Should we worry about this case?

Yes, we need to handle these cases. In fact I thought the `map` was keyed by the option name, as you point out not all options are objects, so the option editor cannot assume that it can update the keys within an option. Lets have a larger discussion about how the option editor should work on the new PR.

affo:
Split in 3 PRs:

 - #383 
 - #386 
 - #387

Issue #328, opened 2018-11-27
Log queries that are abandoned within the controller and ensure they are cleaned up


jsternberg:
The controller returns a query that must have `Done()` called on it to be properly cleaned up by the controller.

Unfortunately, coding error may cause a query to not have `Done()` called on it. In order to ensure the proper functioning of the server in this error situation, it would be good to add a finalizer to the returned query that would ensure that `Done()` gets called. This finalizer should also log a message to indicate that something is very wrong and needs to be addressed by the developers (us).

nathanielc:
@jsternberg With the changes to the controller is this still relevant? If not please close.

jsternberg:
I believe this is still relevant.

PR #331, opened 2018-11-27
feat(control/controltest): add controltest.Controller

mark-rushakoff:
controltest.Controller associates a finalizer with returned queries,
which panics if Query.Done is not called.

When triggered, this panics with a stack trace of where the query was
created, which pinpoints exactly where there is a missing Done call.

This is not a complete solution, because short tests may not trigger a
garbage collection; see also #328.

### DIFF
--- a/control/controltest/controller.go
+++ b/control/controltest/controller.go
@@ -1,0 +1,58 @@
+// Package controltest provides a controller for use in tests.
+package controltest
+
+import (
+	"context"
+	"runtime"
+	"runtime/debug"
+
+	"github.com/influxdata/flux"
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
Can do. If it ever turns out we need to run this check in production, it should be straightforward to add the option back to the original controller. I think we did something similar in influxdb to discover cursors not being closed, but those were probably just logging instead of panicking.

I'll add a control/controltest package with a new type to encapsulate this behavior, and update this PR.

mark-rushakoff:
@nathanielc updated, PTAL.

PR #338, opened 2018-11-29
Make "group" behave according to SPEC

### LINKED ISSUE #316: Group operator should require exactly one argument
Group operator should fail when no argument is provided.  
At the moment, `group()` defaults to `group(all: true)`.

### LINKED ISSUE #322: Group operation chaining produces unexpected result with group all
If we chain multiple `group` operations, we get the expected result, so that the last `group` operation is the only one that matters.

I.e., this:

```
from(bucket: "test")
    |> range(start: -1y)
    |> group(none: true)
    |> group(by: ["tag1"])
    |> group(all: true)
    |> ...
    |> group(by: ["tag1", "tag2"])
```

is equivalent to:

```
from(bucket: "test")
    |> range(start: -1y)
    |> group(by: ["tag1", "tag2"])
```

However, if the last group is `group(all: true)`, it has the same effect as a `group(none: true)`, we get only one table with an empty group key.  

In other words, this:

```
from(bucket: "test")
    |> range(start: -1y)
    |> group(by: ["tag1"])
    |> group(all: true)
```

Is _not_ equivalent to:

```
from(bucket: "test")
    |> range(start: -1y)
    |> group(all: true)
```

And produces a single table with no group key.

### LINKED ISSUE #335: Remove group all:true
The group all:true parameter is not needed and should be removed from the function signature and implementation.

See https://github.com/influxdata/flux/pull/313 for discussion.

wolffcm:
These changes remove the `all` keyword argument from `group`, and make it so `group` requires exactly one argument.

This makes it so `group` behaves as discussed here:
https://github.com/influxdata/flux/pull/313

Fixes #316 
Fixes #335 
Fixes #322

### DIFF
--- a/functions/transformations/group.go
+++ b/functions/transformations/group.go
@@ -20,7 +20,6 @@ const GroupKind = "group"
 type GroupOpSpec struct {
 	By     []string `json:"by"`
 	Except []string `json:"except"`
-	All    bool     `json:"all"`
 	None   bool     `json:"none"`
 }
 
@@ -30,7 +29,6 @@ func init() {
 			"by":     semantic.NewArrayPolyType(semantic.String),
 			"except": semantic.NewArrayPolyType(semantic.String),
 			"none":   semantic.Bool,
-			"all":    semantic.Bool,
 		},
 		nil,
 	)
@@ -53,11 +51,6 @@ func createGroupOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operat
 	} else if ok && val {
 		spec.None = true
 	}
-	if val, ok, err := args.GetBool("all"); err != nil {
-		return nil, err
-	} else if ok && val {
-		spec.All = true
-	}
 
 	if array, ok, err := args.GetArray("by", semantic.String); err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


wolffcm:
@affo  Good questions.

> Do you think it is better to remove unused group modes (All, Default)?

Right now the group modes in Flux are tightly coupled with those used by storage (with a simple cast to convert between the two):
https://github.com/influxdata/platform/blob/master/query/functions/inputs/storage/storage.go#L172-L183
https://github.com/influxdata/platform/blob/master/query/functions/inputs/from.go#L82

I think we need a separate issue to make them more distinct types with an explicit function that maps one to the other.  @stuartcarnie Does this make sense to you?

> For curiosity: is group(by: []) the same of group(none: true)? Is it possible to perform it, or it is malformed?

I think that `group(by: [])` is well-formed and should be allowed, since there's nothing inconsistent about it.  It should behave the same as `group(none: true)`.  I guess we provide the latter since it maybe easier to read (although the former is fewer keystrokes...)

affo:
@wolffcm Thank you for the detailed answer!

Apart from curiosity, I approve the PR.

stuartcarnie:
@wolffcm yes, I agree. I am thinking more specific RPCs such as `GroupByRead` and `GroupExceptRead`

stuartcarnie:
That aligns with my thinking that we want RPC calls to model a physical operation. Some benefits with this approach include:

* obvious what types of operations the planner can push down
    * as a corollary, we can optimize RPC request / response formats per operation
* optimized metadata RPC calls

PR #345, opened 2018-11-29
bug(csv): csv multi-result decoder was not handling the case where an…

aanthony1243:
… empty annotation group is followed by another group.  fix checks for empty tables before erroring.


aanthony1243:
@affo  this fix is on master if you want to try your inflummoxed tests again

PR #347, opened 2018-11-29
fix: do not panic when the right hand side is not a call expression

jsternberg:


jsternberg:
@wolffcm I think it's a temporary thing. The PEG parser has a comment that it should be any valid expression, but that's a separate issue to fix that I don't really understand yet.

PR #363, opened 2018-11-30
Updates to parser

nathanielc:


nathanielc:
@jsternberg This is still WIP, its just what I got done poking at it.

internal/parser/grammar.md
```diff
@@ -105,7 +105,7 @@ For the parser, the above grammar undergoes a process to have the left-recursion
     CallExpression = lparen ParameterList rparen .
     IndexExpression = lbrack Expression rbrack .
     UnaryExpression = PrimaryExpression
-                    | PrefixOperator PrimaryExpression .
+                    | PrefixOperator PostfixExpression .
```
jsternberg: I think one other part has to change. It looks like the current EBNF gives higher precedence to unary operations than the postfix and it should be of lower precedence. I think `PipeExpression` has to be changed to `PipeExpression = UnaryExpression { PipeExpressionSuffix }`.

jsternberg:
I think it would be good to split out the individual parts of this. In particular, the comparison operators can be split out and merged immediately and the `parseExpressionSuffix` in ident statement also since that's a pretty big oversight.

jsternberg:
I've opened #368 for the unary expressions and postfix expressions precedence mixup.

nathanielc:
Closing in favor of #371 adn #370

PR #375, opened 2018-12-04
docs: update SPEC for official grammar

nathanielc:
This PR only updates the SPEC, there are several renames that need to
occur in order for the code to match this update to the SPEC.
Those changes will follow in separate PRs.

Closes https://github.com/influxdata/platform/issues/246


nathanielc:
PRs #378 #379 and #381 update the code for the various changes in the SPEC.

PR #378, opened 2018-12-04
refactor: rename ArrowFunction* -> Function

nathanielc:
There is only one type of function expression remove the redundant
description Arrow from the name.


aanthony1243:
I always felt like we'd get confused when we start bringing in the apache arrow bits.  If we need to specialize function types later, we can rename this to `anonymous` function or something like that.

Issue #340, opened 2018-11-29
Update SPEC for influxFieldsAsCols


affo:
`fromRows` has been dropped #307

PR #384, opened 2018-12-05
docs(spec): InfluxFieldsAsCols specification

affo:
Issue #340

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2289,20 +2289,25 @@ If no value is found, the value is set to `null`.
 
 [IMPL#353](https://github.com/influxdata/platform/issues/353) Null defined in spec but not implemented.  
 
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -2289,21 +2289,38 @@ If no value is found, the value is set to `null`.
...
+  |> filter(fn: (r) => r._measurement == "cpu")
+  |> range(start: -1h)
+  |> influxFieldsAsCols()
+  |> keep(columns: ["_time", "cpu", "usage_idle", "usage_user"])
+```
+
+Range must be applied before InfluxFieldsAsCols. Queries like this one:
+
 ```
+// WRONG
```
aanthony1243: this is true for nearly all `flux` functions.  I don't think we need to say this here.

PR #387, opened 2018-12-05
AST Option Editor

affo:
Must be merged to master after #386

### DIFF
--- a/ast/edit/option_editor.go
+++ b/ast/edit/option_editor.go
@@ -1,0 +1,173 @@
+package edit
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+// `OptionFn` is a function that, provided with an `OptionStatement`, returns
+// an `Expression` or an error. It is used by `Option` functions to edit
+// AST's options statements.
+type OptionFn func(opt *ast.OptionStatement) (ast.Expression, error)
+
+// `Option` passes the `OptionStatement` in the AST rooted at `node` that has the
+// specified identifier to `fn`.
+// The function can have side effects on the option statement
+// and/or return a non-nil `Expression` that is set as value for the option.
+// If the value returned by the edit function is `nil` (or an error is returned) no new value is set
+// for the option statement (but any, maybe partial, side effect is applied).
+// `Option` returns whether it could find and edit the option (possibly with errors) or not.
+func Option(node ast.Node, optionIdentifier string, fn OptionFn) (bool, error) {
+	oe := &optionEditor{identifier: optionIdentifier, optionFn: fn, err: nil}
+	ast.Walk(oe, node)
+
+	if oe.err != nil {
+		return oe.found, oe.err
+	}
+
+	return oe.found, nil
+}
+
+// Creates an `OptionFn` for setting the value of an `OptionStatement`.
+func OptionValueFn(expr ast.Expression) OptionFn {
+	return func(opt *ast.OptionStatement) (ast.Expression, error) {
+		return expr, nil
+	}
+}
+
+// Creates an `OptionFn` for updating the values of an `OptionStatement` that has an
+// `ObjectExpression` as value. Returns error if the child of the option statement is not
+// an object expression, or if some key in the provided map is not a property of the object.
+func OptionObjectFn(keyMap map[string]values.Value) OptionFn {
+	return func(opt *ast.OptionStatement) (ast.Expression, error) {
+		obj, ok := opt.Assignment.Init.(*ast.ObjectExpression)
+		if !ok {
+			return nil, fmt.Errorf("value is is %s, not an object expression", opt.Assignment.Init.Type())
+		}
+
+		// check that every specified property exists in the object
+		keys := make(map[string]bool, len(obj.Properties))
+		for _, p := range obj.Properties {
+			keys[p.Key.Name] = true
+		}
+
+		for k := range keyMap {
+			if !keys[k] {
+				return nil, fmt.Errorf("cannot find property '%s' in object expression", k)
+			}
+		}
+
+		for _, p := range obj.Properties {
+			value, found := keyMap[p.Key.Name]
+			if found {
+				p.Value = CreateLiteral(value)
+			}
+		}
+
+		return nil, nil
+	}
+}
+
+// `CreateLiteral` creates a AST `Expression` from a `Value`. Supported types for value are Bool, Int, UInt, Float,
+// String, Time, Duration, Regexp, Array, and Object. In the case of Object, the returned expression gets
+// properties sorted by their keys in increasing order.
+func CreateLiteral(v values.Value) ast.Expression {
+	var literal ast.Expression
+	switch v.Type().Nature() {
+	case semantic.Bool:
+		literal = &ast.BooleanLiteral{Value: v.Bool()}
+	case semantic.UInt:
+		literal = &ast.UnsignedIntegerLiteral{Value: v.UInt()}
+	case semantic.Int:
+		literal = &ast.IntegerLiteral{Value: v.Int()}
+	case semantic.Float:
+		literal = &ast.FloatLiteral{Value: v.Float()}
+	case semantic.String:
+		literal = &ast.StringLiteral{Value: v.Str()}
+	case semantic.Time:
+		literal = &ast.DateTimeLiteral{Value: v.Time().Time()}
+	case semantic.Duration:
+		literal = &ast.DurationLiteral{
+			Values: []ast.Duration{
+				{
+					Magnitude: int64(v.Duration()),
+					Unit:      "ns",
+				},
+			},
+		}
+	case semantic.Regexp:
+		literal = &ast.RegexpLiteral{Value: v.Regexp()}
+	case semantic.Array:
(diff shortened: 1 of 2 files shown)


affo:
👍

ast/edit/option_editor_test.go
```diff
@@ -0,0 +1,82 @@
...
+	|> range(start: 2018-05-23T13:09:22.885021542Z)`,
+			editor: edit.NewOptionEditor(map[string]values.Value{
+				"bucket": values.New("foo"),       // should ignore this
+				"start":  values.New("yesterday"), // should ignore this
+				"name":   values.New("bar"),
+				"every":  values.New(duration("2h")),
+				"delay":  values.New(duration("42m")),
+				"cron":   values.New("buz"),
+				"retry":  values.New(int64(10)),
+			}),
```
nathanielc: Not all options are objects, so these keys would not make sense in that case. For example the `now` option is actually a function. By default it is defined as:

```
option now = () => systemTime()
```


Or more simply the the `location` may need to be changed to a function call.

For example:

```
option location = loadLocation("Custom/Location")
```

Actually thinking about this I think that maybe the `values.Value` is the wrong interface to use, but rather we should use `ast.Expression`. Meaning you can replace an options `Init` with an arbitrary expression. Then we can make the `getLiteral` function public so that if a consumer has a values.Value they can create an ast.Expression from it.

affo:
> Meaning you can replace an options Init with an arbitrary expression

Yeah, yesterday I was thinking about the same thing.

I preferred to offer a more general interface with an `EditFn` that edits an option and can return an expression to directly set its value.

I also offered more functions that are particular applications of the more general `Edit`. What do you think about it?

I still feel there is something wrong with packaging...

I needed to make a new package to break the circular dependency `semantic -> ast -> semantic` caused by using `Value`.

ast/edit/option_editor.go
```diff
@@ -0,0 +1,141 @@
...
+			Values: []ast.Duration{
+				{
+					Magnitude: int64(v.Duration()),
+					Unit:      "ns",
+				},
+			},
+		}
+	case semantic.Regexp:
+		literal = &ast.RegexpLiteral{Value: v.Regexp()}
+	default:
```
nathanielc: This function should support creating array and object literals as well.

ast/edit/option_editor.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+	"github.com/pkg/errors"
+)
+
+var OPT_NOT_FOUND = errors.New("option not found")
```
nathanielc: While somewhat common in Go its best to avoid these "sentinel" errors as it creates a dependency on the package to be able to handle errors. See this blog post for some good discussion on the issue https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully.

ast/edit/option_editor.go
```diff
@@ -0,0 +1,141 @@
...
+					return nil, err
+				}
+
+				p.Value = lit
+			}
+		}
+
+		return nil, nil
+	}
+	return Edit(node, optionIdentifier, editFn)
```
nathanielc: The EditObject function should return an EditFn instead of calling Edit directly. Then it can easily be reused.

Same for the EditValue function.

ast/edit/option_editor.go
affo: I read it and it made me conclude that I shouldn't return an error for that specific case

GitCop:
There were the following issues with your Pull Request

 * Commit: 227db751fa4f3396cadcd5c44881863b45f25d50
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 731e563fa1fc3f65bcba4ed191716abc5c5beeae
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**



---
This message was auto-generated by https://gitcop.com

ast/edit/option_editor.go
```diff
@@ -0,0 +1,153 @@
...
+/*
+`Edit` passes the `OptionStatement` in the AST rooted at `node` that has the
+specified identifier to `editFn`.
+The function can have side effects on the option statement
+and/or return a non-nil `Expression` that is set as value for the option.
+If the value returned by the edit function is `nil` (or an error is returned) no new value is set
+for the option statement (but any, maybe partial, side effect is applied).
+`Edit` returns whether it could find and edit the option (possibly with errors) or not.
+*/
+func Edit(node ast.Node, optionIdentifier string, editFn EditFn) (bool, error) {
```
nathanielc: nit, but the convention is to use `//` style comments even if they are large.

ast/edit/option_editor.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+type EditFn func(opt *ast.OptionStatement) (ast.Expression, error)
```
nathanielc: Can you add a comment to this type? Also it should have `Option` in its name and it does not need `edit` in its name since we are in the `edit` package. Maybe `OptionFn`?

ast/edit/option_editor.go
```diff
@@ -0,0 +1,153 @@
...
+			if found {
+				p.Value = GetLiteral(value)
+			}
+		}
+
+		return nil, nil
+	}
+}
+
+func GetLiteral(v values.Value) ast.Expression {
```
nathanielc: Can you comment this function? Also I think it needs a better name maybe `CreateLiteral` or `ConstructLiteral`?

ast/edit/option_editor.go
```diff
@@ -0,0 +1,153 @@
...
+		arr.Range(func(i int, el values.Value) {
+			arrExpr.Elements[i] = GetLiteral(el)
+		})
+
+		literal = arrExpr
+	case semantic.Object:
+		obj := v.Object()
+		objExpr := &ast.ObjectExpression{Properties: make([]*ast.Property, obj.Len())}
+		ip := new(int)
+		*ip = 0
```
nathanielc: This logic can be simplified. The `make` function accept three arguments when creating a slice: `make(type, length, capacity)`.

So you can create the Properties slice with enough capacity but zero length, then you can append to the slice which will grow its length, but not is capacity.

Generic example:
```
s := make([]int, 0, len(mymap))
for k ,v := range mymap {
    s = append(s, Pair{k,v})
}
```

This is a very common pattern when copying a map into a slice.

If you are interested in more details on how slices work https://blog.golang.org/go-slices-usage-and-internals

ast/edit/option_editor_test.go
```diff
@@ -0,0 +1,242 @@
...
+	retry: 10,
+}`},
+			editExpected: true,
+			edit: func(node ast.Node) (bool, error) {
+				return edit.Edit(node, "task", edit.EditObjectFn(map[string]values.Value{
+					"foo":   values.New("foo"), // should ignore this
+					"bar":   values.New("bar"), // should ignore this
+					"buz":   values.New("buz"), // should ignore this
+					"every": values.New(duration("2h")),
+					"delay": values.New(duration("42m")),
```
nathanielc: This can be easily written as `42*time.Minute`, this way you do not have to create the helper function to parse it. And it still reads clearly.

ast/edit/option_editor_test.go
```diff
@@ -0,0 +1,242 @@
...
+	delay: 2520000000000ns,
+	cron: "buz",
+	retry: 10,
+}`},
+			editExpected: true,
+			edit: func(node ast.Node) (bool, error) {
+				return edit.Edit(node, "task", edit.EditObjectFn(map[string]values.Value{
+					"foo":   values.New("foo"), // should ignore this
+					"bar":   values.New("bar"), // should ignore this
+					"buz":   values.New("buz"), // should ignore this
```
nathanielc: It should be an error to provide keys to edit that do nothing.

ast/edit/option_editor_test.go
```diff
@@ -0,0 +1,242 @@
...
+						"b": values.NewInt(2),
+					}),
+				})
+
+				literal := edit.GetLiteral(values.NewObjectWithValues(map[string]values.Value{
+					"x": x,
+					"y": y,
+					"z": z,
+				}))
+				return edit.Edit(node, "foo", edit.EditValueFn(literal))
```
nathanielc: Its not clear to me why this produces multiple edits. Can you explain that?

ast/edit/option_editor.go
nathanielc: Can you rename this function to `Option`? That way it reads like `edit.Option` when you use it and it lets us define other "edit" methods in the future.

Alternatively if you find that using a verb as the package name is confusing we could rename this package to `options` and then it becomes `options.Edit`. I am fine with either, so no as there is not stutter (i.e. edit.EditOptions), and option is somewhere in the name.

If you do rename the package to `options` then the `GetLiteral` function doesn't really belong in the package anymore and should be moved.

ast/edit/option_editor_test.go
affo: It's because `x, y, z` can happen in any order because they are passed as a map. Would it be better if I execute the comparison directly on the ast and not on formatting?

PR #390, opened 2018-12-05
fix(execute): terminal output functions must produce results

jlapacik:
Fixes https://github.com/influxdata/platform/issues/1311

Output or side effect producing functions operate as pass-throughs.
They don't transform the data they consume. They perform their
intended side effect and then pass the data along to the next
transformation.

Previously if an output function was a terminal node in a plan,
a result transformation was not generated for that plan. All
queries must have an associated result transformation, otherwise
the query context will be cancelled immediately and execution
will be cancelled before any work for the query has been done.

With this fix, a result transformation is generated for every
terminal node in a plan. Queries with terminal output functions
will now perform their side effects before the query returns.

Note that terminal output functions will return a result
consisting of all tables that were consumed by the output.

### DIFF
--- a/execute/executetest/output.go
+++ b/execute/executetest/output.go
@@ -1,0 +1,78 @@
+package executetest
+
+import (
+	"errors"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+)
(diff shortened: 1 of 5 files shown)


jlapacik:
Working on adding a test right now.

PR #396, opened 2018-12-06
feat(transformations): add function assertEquals to transformations

### LINKED ISSUE #358: implement assertEQ function
related to #107 assertEQ takes two streams as input and outputs whether the two streams are equal.  

TODO: 

- [x] update spec
- [x] implement builtin procedure
- [x] Implement test

aanthony1243:
* added assertEquals, as well as extensions to the execute package to support the feautre.
* added builtin helper function testingTest to cover common usage of assertEquals
* updated functions/transformations/simple_max.flux to demo testingTest usage.
* added unit tests for assertEquals

closes #358

### DIFF
--- a/execute/allocator.go
+++ b/execute/allocator.go
@@ -48,6 +48,7 @@ func (a *Allocator) AppendBools(slice []bool, vs ...bool) []bool {
 	return s
 }
 
+// GrowBools increases the size of the input slice by n
 func (a *Allocator) GrowBools(slice []bool, n int) []bool {
 	newCap := len(slice) + n
 	if newCap < cap(slice) {
@@ -60,6 +61,14 @@ func (a *Allocator) GrowBools(slice []bool, n int) []bool {
 	diff := cap(s) - cap(slice)
 	a.account(diff, boolSize)
 	return s
+}
+
+// SliceBools reduces the input slice to slice[start:stop]
+func (a *Allocator) SliceBools(slice []bool, start, stop int) []bool {
+	origLen := len(slice)
+	slice = slice[start:stop]
+	a.Free(origLen-len(slice), boolSize)
+	return slice
 }
 
 // Ints makes a slice of int64 values.
@@ -79,6 +88,7 @@ func (a *Allocator) AppendInts(slice []int64, vs ...int64) []int64 {
 	return s
 }
 
+// GrowInts increases the size of the input slice by n
 func (a *Allocator) GrowInts(slice []int64, n int) []int64 {
 	newCap := len(slice) + n
 	if newCap < cap(slice) {
@@ -91,6 +101,14 @@ func (a *Allocator) GrowInts(slice []int64, n int) []int64 {
 	diff := cap(s) - cap(slice)
 	a.account(diff, int64Size)
 	return s
+}
+
+// SliceInts reduces the input slice to slice[start:stop]
+func (a *Allocator) SliceInts(slice []int64, start, stop int) []int64 {
+	origLen := len(slice)
+	slice = slice[start:stop]
+	a.Free(origLen-len(slice), int64Size)
+	return slice
 }
 
 // UInts makes a slice of uint64 values.
@@ -110,6 +128,7 @@ func (a *Allocator) AppendUInts(slice []uint64, vs ...uint64) []uint64 {
 	return s
 }
 
+// GrowUInts increases the size of the input slice by n
 func (a *Allocator) GrowUInts(slice []uint64, n int) []uint64 {
 	newCap := len(slice) + n
 	if newCap < cap(slice) {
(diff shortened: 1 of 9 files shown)


execute/table.go
```diff
@@ -953,6 +1018,45 @@ func (b ColListTableBuilder) RawTable() *ColListTable {
...
+
+	if stop < start || stop > b.table.nrows {
+		return fmt.Errorf("invalid start/stop parameters: %d/%d", start, stop)
+	}
+
+	for i, col := range b.table.cols {
+		switch col.Meta().Type {
+		case flux.TBool:
+			col := b.table.cols[i].(*boolColumn)
+			col.data = b.alloc.SliceBools(col.data, start, stop)
```
nathanielc: I see why you are using the allocator to track that the memory is not being used anymore. To reduce complexity and avoid possible accounting errors I think we should not use the allocator to manage slices.

But rather when we call Free inside the `Clear` call for the various table implementations we should use the capacity of the array not the length. This should keep accounting simple and allow for slicing the array in the meantime. 

Also to confirm my understanding that slicing cannot cause new allocations here is the relevant bit in the Go spec https://golang.org/ref/spec#Slice_types.

functions/transformations/assert_equals.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+		if ok, err := execute.TablesEqual(cacheTable, tbl); !ok {
+			return fmt.Errorf("test %s: tables not equal", t.name)
+		} else if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
nathanielc: I am not sure how to go about this but the current implementation will ignore missing tables. Since the execute.TablesEqual is only called when two matching tables are found if an entire table is missing this will not report the error.

functions/transformations/assert_equals.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	} else {
+		cacheTable, err := builder.Table()
+		if err != nil {
+			return err
+		}
+		if ok, err := execute.TablesEqual(cacheTable, tbl); !ok {
+			return fmt.Errorf("test %s: tables not equal", t.name)
+		} else if err != nil {
+			return err
```
nathanielc: Should the error check happen first? The way it is written this line is unreachable, since any error returned from execute.TablesEqual is also !ok,

functions/transformations/assert_equals.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (s *AssertEqualsOpSpec) Kind() flux.OperationKind {
+	return AssertEqualsKind
+}
+
+func init() {
+	assertEqualsSignature := semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"name": semantic.String,
+			"got":  flux.TableObjectType,
```
nathanielc: The examples show that `got` is the pipe forward argument, can you update the signature here for that?

Should we rename `got` to `tables` to make it the same as other chaining functions?

execute/table.go
```diff
@@ -312,15 +313,79 @@ func AppendMappedRecordExplicit(i int, cr flux.ColReader, builder TableBuilder,
...
+
+// TablesEqual takes two flux tables and compares them.  Returns false if the tables have different keys, different
+// columns, or if the data in any column does not match.  Returns true otherwise.  This function will consume the
+// ColumnReader so if you are calling this from the a Process method, you may need to copy the table if you need to
+// iterate over the data for other calculations.
+func TablesEqual(left, right flux.Table) (bool, error) {
+	eq := false
+	// rbuffer will buffer out rows from the right table, always holding just enough to do a comparison with the left
+	// table's ColReader
+	rowBuffer := NewColListTableBuilder(right.Key(), &memory.Allocator{})
```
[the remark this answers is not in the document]
aanthony1243: I thought about that. will make the change.

execute/allocator.go
```diff
@@ -62,6 +63,14 @@ func (a *Allocator) GrowBools(slice []bool, n int) []bool {
 	return s
 }
 
+// SliceBools reduces the input slice to slice[start:stop]
+func (a *Allocator) SliceBools(slice []bool, start, stop int) []bool {
+	origLen := len(slice)
+	slice = slice[start:stop]
+	a.Free(origLen-len(slice), boolSize)
```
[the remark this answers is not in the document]
aanthony1243: only reason I did this was because that's how all other changes are made to the ColListTable columns.  But you are right about the backing array.  This call to Free is probably not safe.  

We could copy out the data and discard the backing array, what do you think of that?

functions/transformations/assert_equals.go
aanthony1243: added test to Finish()

functions/transformations/assert_equals.go
aanthony1243: I think it will be common enough to write a test `assertEquals(got: s1, want: s2)`  that `got` is better.  If it's a big issue, we can remove the pipe-forwarding from the spec.

PR #397, opened 2018-12-06
feat(semantic): add package and import support to the semantic graph

nathanielc:


GitCop:
There were the following issues with your Pull Request

 * Commit: d9bd37821d949d31f27351ebc25140681f7f92ac
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 5441e27eac6aea975c60856b75d66d9e3172d54a
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**



---
This message was auto-generated by https://gitcop.com

semantic/package.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Package represents the type and name of a package.
+type Package struct {
+	Name string
+	Type PolyType
+}
+
+// CreatePackage constructs a Package from the node. The node must contain a Program node
+// with a valid PackageClause
+func CreatePackage(n Node, importer Importer) (Package, error) {
```
jlapacik: Will packages and imports be replacing externs completely? If so, should this function specifically take a `Program` type instead of `Node`?

nathanielc: They will not, atleast I do not expect them to. There are some builtin values in the top level namespace that are not part of any package. Maybe we can model those as a package in the future but not yet.

semantic/graph.go
```diff
@@ -137,13 +141,55 @@ func (p *Program) Copy() Node {
...
+func (*ImportDeclaration) NodeType() string { return "ImportDeclaration" }
+
+func (d *ImportDeclaration) Copy() Node {
+	if d == nil {
+		return d
+	}
+	nd := new(ImportDeclaration)
+	*nd = *d
+
+	nd.As = d.As.Copy().(*Identifier)
```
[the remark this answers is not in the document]
nathanielc: It is safe to call Copy when `As` is nil.

The implementation of `Identifier.Copy` explicitly handles this case:

```
func (i *Identifier) Copy() Node {
	if i == nil {
		return i
	}
	ni := new(Identifier)
	*ni = *i

	return ni
}
```

In fact all of the Copy methods do. Also note that `i` is returned instead of a bare `nil`. This is because `nils` in Go are typed and `i`'s type is `*Identifier` and a bare `nil` would have type `Node`, but since we type assert the result as an `*Identifier` it would panic if we  returned a `nil` with type `Node`.

PR #399, opened 2018-12-07
fix: add import/package nodes to ast.Walk

nathanielc:


ast/walk.go
```diff
@@ -39,48 +39,89 @@ func (v *visitor) Visit(node Node) Visitor {
...
-func walk(v Visitor, n Node) {
-	if n == nil {
-		return
-	}
-
-	switch n := n.(type) {
+func walk(v Visitor, node Node) {
+	// We need to check if node is nil,
+	// because of Go's type system we need to compare against the concrete type of n,
+	// which means we need to check in each case statement.
```
affo: Woah, good to know!

affo:
I also checked also tests.  
Thank you for the example, I didn't think about it!

ast/walk.go
nathanielc: For some more details on this gotcha https://www.gmarik.info/blog/2016/understanding-golang-nil-value/

PR #400, opened 2018-12-07
fix(semantic): collision between external and fresh type vars

### LINKED ISSUE #263: end-to-end test fails on CI occasionally

jlapacik:
Fixes #263.

Type inference replaces any external free type variables with new
fresh type variables. The result is a new but completely equivalent
type expression. Previously during the substitution process
collisions could occur between the names of old type variables and
the names of the new freshly generated type variables. This would
render the new type expression invalid.

To remedy this problem, fresh type variables are now generated
before being substituted into the original type expression. If
there is a collision, that type variable is discarded and a new
one is generated.

Note: Not sure how to create a test case for this. Although the flaky test from #263 that this
PR fixes is no longer flaky.

### DIFF
--- a/semantic/constraints.go
+++ b/semantic/constraints.go
@@ -96,7 +96,11 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
 		ftv := n.ExternType.freeVars(nil)
 		subst := make(Substitution, len(ftv))
 		for _, tv := range ftv {
-			subst[tv] = v.cs.f.Fresh()
+			f := v.cs.f.Fresh()
+			for ftv.contains(f) {
+				f = v.cs.f.Fresh()
+			}
+			subst[tv] = f
 		}
 		t := subst.ApplyType(n.ExternType)
 		// Check if this type knows about its kind constraints
--- a/semantic/inference_test.go
+++ b/semantic/inference_test.go
@@ -448,6 +448,65 @@ identity(x:identity)(x:2)
 			},
 		},
 		{
+			name: "extern type variables",
+			node: &semantic.Extern{
+				Assignments: []*semantic.ExternalVariableAssignment{
+					{
+						Identifier: &semantic.Identifier{Name: "f"},
+						ExternType: semantic.NewFunctionPolyType(
+							semantic.FunctionPolySignature{
+								Return: semantic.Tvar(3),
+							},
+						),
+					},
+					{
+						Identifier: &semantic.Identifier{Name: "g"},
+						ExternType: semantic.NewFunctionPolyType(
+							semantic.FunctionPolySignature{
+								Return: semantic.Tvar(5),
+							},
+						),
+					},
+				},
+				Block: &semantic.ExternBlock{
+					Node: &semantic.Program{
+						Body: []semantic.Statement{
+							&semantic.NativeVariableAssignment{
+								Identifier: &semantic.Identifier{Name: "a"},
+								Init:       &semantic.IdentifierExpression{Name: "f"},
+							},
+							&semantic.NativeVariableAssignment{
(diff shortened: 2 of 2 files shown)


jsternberg:
Do you know how those type collisions happen? I'm concerned about the `for` loop and its impact on performance and potential for hitting more than one collision.

jlapacik:
@jsternberg previously a collision could happen like this:

* External function type: (n: t30, ^tables: t32) -> t340
* Substitution: {t30: t32, t32: t33, t340: t34}

If the substitution occurs in the following order we have a problem:

    t30: t32 => (n: t32, ^tables: t32) -> t340 // Uh oh, n and tables should not be the same type
    t32: t33 => (n:t33, ^tables: t33) -> t340
    t340: t34 => (n:t33, ^tables: t33) -> t34

The solution is to have the set of pre-substitution type variables and the set of post-substitution type variables not intersect.

jlapacik:
I agree that multiple calls to `contains` does have a negative performance impact.

jsternberg:
Is there a way we can unify the generation of the substitution variables so we aren't using a pool of names that can conflict? I don't know too much about the type inference, but it feels like there might be a better solution to the problem.

semantic/constraints.go
```diff
@@ -96,7 +96,11 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
 		ftv := n.ExternType.freeVars(nil)
 		subst := make(Substitution, len(ftv))
 		for _, tv := range ftv {
-			subst[tv] = v.cs.f.Fresh()
+			f := v.cs.f.Fresh()
+			for ftv.contains(f) {
+				f = v.cs.f.Fresh()
+			}
+			subst[tv] = f
```
nathanielc: It should be possible to create a test case that enter the loop. That would be a good enough test case for me. Have a look at the extern type inference test cases that already exist. Since you can just pick the type variables the extern type use, you should be able to create a conflict that requires this loop to fix.

jlapacik:
@jsternberg type variables are represented as incrementing integers. The issue is exclusive to external variables. The types of external variables are computed in a separate type inference process. To be more specific, the types of Flux built-in functions and scripts are computed separately. Type inference is then run a second time, this time over the actual Flux program that calls the built-ins. In this second call to the type inference process external variables (and only external variables) will already have some type information. We need to make sure this type information doesn't conflict with any type information we compute for the rest of the program.

nathanielc:
> Is there a way we can unify the generation of the substitution variables so we aren't using a pool of names that can conflict? I don't know too much about the type inference, but it feels like there might be a better solution to the problem.

I looked into this initially and concluded it was too complex a solution. The problem is that these external type variables can come from anywhere and so coordinating all external locations to use the same semantic.Fresher was too complicated and error prone. As a result the solution was to transform all external type variables to new ones. This fix seems appropriate given that.

jsternberg:
I guess the part that to me that seems weird is the for loop. If we are renaming types from other locations to ones that work in the local context, shouldn't we only need to replace the type once?

It makes sense to rename external type variables. That's how linkers work too.

jlapacik:
If t33 intersects with any of the type variables, we generate t34 but still need to check if it conflicts. If so we generate t35 and so on. I could probably make this more efficient.

jsternberg:
I was thinking about it a bit more and I think this is fine. I retract my objection.

PR #401, opened 2018-12-07
refactor(inputs): better error message for unbounded from

### LINKED ISSUE #302: Range ignored if put after pivoting operation

affo:
Sometimes it happens that the planner cannot push `range` to storage and complains that the query is unbounded:

```
result from0 is unbounded. Add a 'range' call to bound the query
```

Now the message helps the user to better understand what's happening and proposes solutions:

```
impossible to bound results from "telegraf/autogen" (from0).
Add a 'range' call to bound the query or move any 'range' call as close as possible to 'from'
```

Fixes #302

### DIFF
--- a/functions/inputs/from.go
+++ b/functions/inputs/from.go
@@ -168,7 +168,13 @@ func (s *FromProcedureSpec) TimeBounds(predecessorBounds *plan.Bounds) *plan.Bou
 
 func (s FromProcedureSpec) PostPhysicalValidate(id plan.NodeID) error {
 	if !s.BoundsSet || (s.Bounds.Start.IsZero() && s.Bounds.Stop.IsZero()) {
-		return fmt.Errorf(`result '%s' is unbounded. Add a 'range' call to bound the query`, id)
+		var bucket string
+		if len(s.Bucket) > 0 {
+			bucket = s.Bucket
+		} else {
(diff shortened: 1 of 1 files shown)


affo:
Great, I see your point. I preferred:

```
from0: results from "telegraf/autogen" must be bounded
```

 - It doesn't include any reference to requests
 - It doesn't require the notion of time, so the possibility to add `limit`, for instance, is left to the user,
 - It doesn't explain how to bound the query (the first message did)
 - It includes the bucket name for better clarity.

While doing this, I tried to run some queries with `limit`, `last`, and `first`, but they do not bound the query. It seems they are not pushed to storage. Should we add some physical rewrite rules for them?

Issue #339, opened 2018-11-29
Add planner rewrite rule to merge consecutive group nodes


wolffcm:
If there are several consecutive `group` operations chained together, then the planner should be able to ignore all but the last one.  The rewrite rules for `from` provide a relevant example:
https://github.com/influxdata/flux/blob/master/functions/inputs/from.go#L181-L238

However in this case the rewrite rule should be a logical instead of a physical rule.

A rewrite rule that matches `group |> group` and just returns just the second `group` would be applied several times by the planner until there is just one consecutive `group`.

Note that at some point the `group` may define its behavior in relative terms using a schema function as described here:
https://github.com/influxdata/flux/issues/27
So this rule should be careful to only be applied when the grouping is being defined in absolute terms (i.e., with the existing group modes of `by`, `except` or `none`).

PR #403, opened 2018-12-07
feat(transformations): rule to chain group operations

affo:
Implements #339

### DIFF
--- a/functions/transformations/group.go
+++ b/functions/transformations/group.go
@@ -37,6 +37,7 @@ func init() {
 	flux.RegisterFunction(GroupKind, createGroupOpSpec, groupSignature)
 	flux.RegisterOpSpec(GroupKind, newGroupOp)
 	plan.RegisterProcedureSpec(GroupKind, newGroupProcedure, GroupKind)
+	plan.RegisterLogicalRules(MergeGroupRule{})
 	execute.RegisterTransformation(GroupKind, createGroupTransformation)
 }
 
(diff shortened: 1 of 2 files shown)


functions/transformations/group.go
```diff
@@ -216,3 +217,27 @@ func (t *groupTransformation) UpdateProcessingTime(id execute.DatasetID, pt exec
...
+
+// returns the pattern that matches `group |> group`
+func (r MergeGroupRule) Pattern() plan.Pattern {
+	return plan.Pat(GroupKind, plan.Pat(GroupKind, plan.Any()))
+}
+
+func (r MergeGroupRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool, error) {
+	firstGroup := node.Predecessors()[0]
+	newSpec := node.ProcedureSpec().Copy()
+
```
wolffcm: I think it would be safest to not apply this rule unless the group mode is `GroupModeBy` or `GroupModeExcept`.  Right now, those are the only group modes possible, but that could change in the future.  There has been some discussion of allowing users a way to express "group by everything the input groups by, but add or subtract these columns."  So I think adding a check here would guard against wrong answers in the future.

functions/transformations/group_test.go
```diff
@@ -590,3 +592,133 @@ func TestGroup_Process(t *testing.T) {
...
+			Before: &plantest.PlanSpec{
+				Nodes: []plan.PlanNode{
+					plan.CreateLogicalNode("from", from),
+					plan.CreateLogicalNode("group", groupBy),
+				},
+				Edges: [][2]int{
+					{0, 1},
+				},
+			},
+			After: &plantest.PlanSpec{
```
wolffcm: I just added a concise way to express that a rule should not apply to a plan, by omitting the `After` field and specifying `NoChange: true`.

functions/transformations/group.go
affo: Ok, now I got the last lines in the issue description about schema introspection 👍

affo: I only checked for the second (last) group not to be by or except, because, in the case of a relative (or fancy) grouping, a subsequent one will always override its behavior. Do you agree?

wolffcm: Yes, makes perfect sense.

PR #404, opened 2018-12-08
Fix/keep

### LINKED ISSUE #330: Call to flux function keep() produces "column doesn't exist" error
This query produces an error:
```
from(bucket:"telegraf")
|> range(start: -1m)
|> keep(columns: ["_measurement"])
```
Error:
```
Error: drop error: column "fstype" doesn't exist
```
I guess that it computes the columns to drop form the first table, and when those columns are not in subsequent tables, it complains?

jlapacik:
Fixes #330 .

### DIFF
--- a/functions/transformations/schema_mutators.go
+++ b/functions/transformations/schema_mutators.go
@@ -205,15 +205,7 @@ func NewDropKeepMutator(qs flux.OperationSpec) (*DropKeepMutator, error) {
 	return m, nil
 }
 
-func (m *DropKeepMutator) checkColumns(tableCols []flux.ColMeta) error {
-	if m.DropCols != nil {
-		for c := range m.DropCols {
-			if err := checkCol(c, tableCols); err != nil {
-				return errors.Wrap(err, "drop error")
-			}
-		}
-	}
-
+func (m *DropKeepMutator) checkKeepColumns(tableCols []flux.ColMeta) error {
 	if m.KeepCols != nil {
 		for c := range m.KeepCols {
 			if err := checkCol(c, tableCols); err != nil {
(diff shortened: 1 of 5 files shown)


functions/transformations/schema_mutators.go
```diff
@@ -206,20 +206,23 @@ func NewDropKeepMutator(qs flux.OperationSpec) (*DropKeepMutator, error) {
...
+	// then we must be performing the `drop` operation. Note that drop
+	// columns will be present even if we are performing the `keep`
+	// operation. So we must check the keep columns first.
 	if m.KeepCols != nil {
 		for c := range m.KeepCols {
 			if err := checkCol(c, tableCols); err != nil {
 				return errors.Wrap(err, "keep error")
 			}
 		}
+	} else if m.DropCols != nil {
```
wolffcm: Looking at the code that does the mutation here:

https://github.com/influxdata/flux/blob/master/functions/transformations/schema_mutators.go#L267-L271

I think the calls to `checkColumns()` and `keepToDropCols()` should be transposed.

It seems like we are performing the check on the set of `dropCols` for the previous table, and that's what's causing this issue.  Your fix works, but it's still weird that we are calling `checkCols` on the stale set of drop columns.

jlapacik: After discussing with @aanthony1243 we decided that dropping a non-existent column should not be an error. So the `m.DropCols != nil` check has been removed and now we only check columns in the case of `keep`.

functions/transformations/schema_mutators.go
```diff
@@ -206,20 +206,23 @@ func NewDropKeepMutator(qs flux.OperationSpec) (*DropKeepMutator, error) {
...
+		for c := range m.DropCols {
+			if err := checkCol(c, tableCols); err != nil {
```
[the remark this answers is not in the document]
jlapacik: @aanthony1243 this has been resolved. An error no longer occurs when one of the input tables does not have a column specified by `drop`.

PR #444, opened 2018-12-10
chore: update semantic-pull-request bot to validate all commits

nathanielc:
Use this commit to test that validation works,

by providing and

invalid

body or footer


zeke:
@nathanielc it makes sense to me that this is passing: there's only one commit and it has a semantic prefix: https://github.com/influxdata/flux/pull/444/commits/20ab00190cc70887247de9c42b0958397885606f

PR #445, opened 2018-12-10
chore(changelog): add command to generate a changelog

nathanielc:


cmd/changelog/cmd/generate.go
```diff
@@ -0,0 +1,300 @@
...
+func parseCommit(commit *object.Commit) (Commit, error) {
+	var (
+		header,
+		body,
+		footer string
+	)
+	lines := strings.Split(commit.Message, "\n\n")
+	switch {
+	case len(lines) >= 3:
+		footer = strings.TrimSpace(strings.Join(lines[2:], "\n\n"))
```
mark-rushakoff: I hadn't thought too hard about the difference between parsing the body vs the footer.  I don't think there's a simple way to tell when there's just a body with no footer, as opposed to a footer with no body, as opposed to two paragraphs of body with no footer. 

If I'm reading this correctly, it doesn't look like we're using the body or footer anywhere after we set those fields on the Commit. I'd lean towards just merging them as a single field, perhaps `BodyAndFooter`.

cmd/changelog/cmd/generate.go
```diff
@@ -0,0 +1,300 @@
...
+	Footer  string
+
+	Breaking bool
+
+	Hash      string
+	ShortHash string
+	Author    string
+}
+
+// headerPattern matches the type, scope and subject of a commit message header.
```
mark-rushakoff: s/subject/description

cmd/changelog/cmd/generate.go
```diff
@@ -0,0 +1,300 @@
...
+var headerPattern = regexp.MustCompile(`^(\w+)(\(\w+\))?:(.*)$`)
+
+const breaking = "BREAKING CHANGE:"
```
mark-rushakoff: The spec says "A breaking change MUST consist of the uppercase text BREAKING CHANGE, followed by a colon and a space," but it's probably fine if we're more relaxed here and don't require the space.

nathanielc: yeah I had the space initially, but since we can't validate those before merge, I figured we would use a more relaxed search since as noted in a commit a false negative is more expensive than a false positive.

cmd/changelog/cmd/generate.go
```diff
@@ -0,0 +1,300 @@
...
+		PreviousVersion: prevVer,
+		Date:            time.Now().Format("2006-01-30"),
+		Breaking:        breaking,
+		Features:        features,
+		Fixes:           fixes,
+		CommitURL:       commitURL,
+	}, nil
+}
+
+// findPreviousRelease searches for the highest semantic version tag.
```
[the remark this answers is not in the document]
nathanielc: Yeah, we can add that later if we need it.

PR #446, opened 2018-12-10
fix(control): make controller return planner failures

### LINKED ISSUE #149: Flux incorrectly responds with 200 OK for an invalid query
Per [the spec](https://github.com/influxdata/flux/blob/master/docs/SPEC.md#errors):

> When an error occurs before any results are materialized then the HTTP status code will indicate an error and the error details will be encoded in the csv table.

The current encoder and results interfaces do not provide a means for distinguishing client vs server errors. For example, the following query:

```flux
```

results in a client error with a `200 OK` response:

```csv
#datatype,string,string
#group,true,true
#default,,
,error,reference
,failed to create physical plan: invalid time bounds from procedure from: bounds contain zero time,
```

Upon examination of the implementation, as found in Platform

https://github.com/influxdata/platform/blob/11f869fb37fb44d2cf59c9e245c6f3821c933993/http/query_service.go#L91-L103

a client error is determined if `n == 0`, however, CSV data has already been written to encode the error. It is unclear how to distinguish client vs server query errors and respond accordingly.

wolffcm:
Previously planner errors would be returned at runtime, even though
execution had not yet begun.  This change makes it so that errors will
be returned (with http error code 400) for unplannable queries.

Fixes #149

### DIFF
--- a/control/controller_test.go
+++ b/control/controller_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/influxdata/flux"
 	_ "github.com/influxdata/flux/builtin"
+	"github.com/influxdata/flux/functions/inputs"
 	"github.com/influxdata/flux/internal/pkg/syncutil"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/mock"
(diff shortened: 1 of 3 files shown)


control/controller_test.go
```diff
@@ -2,6 +2,7 @@ package control
 
 import (
 	"context"
+	"github.com/influxdata/flux/functions/inputs"
```
jsternberg: This import is in the wrong order.

PR #448, opened 2018-12-11
refactor(changelog): move the changelog command to an internal package

jsternberg:
This signals that changelog is meant as an internal command for managing
our own changelog rather than part of flux. If you run `go install
./cmd/...` it will now not compile the changelog command.


nathanielc:
The import paths need to be updated.

PR #479, opened 2018-12-12
feat(arrow): add utility wrappers for arrow types

jsternberg:
The utility wrappers take care of integrating the flux memory management
with the arrow types along with wrapping over the detail of which
specific arrow array type we use for each column type.

This should provide an easier base for integrating arrow within the flux
engine.


arrow/int.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (a *Int) Value(i int) int64 {
+	return a.data.Value(i)
+}
+
+func (a *Int) Int64Values() []int64 {
+	return a.data.Int64Values()
+}
+
+func (a *Int) Free() {
```
stuartcarnie: Why are these named `Free` and `Copy` vs `Retain` and `Release`?

jsternberg: I found `Retain` and `Release` personally confusing since I felt you had to then do the reference counting yourself. The original design had a wrapper called `BaseRef` that had the `Copy()` method but did not have `Free()` to signal that this wasn't a copy of the array that you  owned so, if you wanted to keep it, you needed to call `Copy()`. That got removed since it was part of the `array` interface.

I still like the overall idea though as `Copy` and `Free` are things that are easier to keep bug free in my opinion and I think it matches more closely with how we currently use it. The arrays once they are created are immutable so a "Copy" is free to be a shallow copy.

stuartcarnie: @jsternberg the behavior of `Retain` and `Release` [are documented](https://github.com/apache/arrow/tree/master/go#reference-counting). Also worth noting is they are based off NSFoundation by Apple. Microsoft used a similar pattern, `AddRef` and `Release` for COM.

jsternberg: We talked about this on Slack and I added that the reason I preferred the different API was so I could later add code that checked to make sure a single instance didn't free itself multiple times. I added that code to this PR so it is more obvious that a double free will not cause a hard-to-trace bug in the code as an added protection.

We may review this again in the future if the extra garbage that is generated causes a problem, but it should be minimal.

nathanielc: As per my review, I think we should not modify the Go Arrow API and use Retain/Release directly.

arrow/int.go
```diff
@@ -0,0 +1,104 @@
...
+		return
+	}
+
+	b.builder.Reserve(len(vs))
+	for _, v := range vs {
+		b.builder.UnsafeAppend(v)
+	}
+}
+
+func (b *IntBuilder) BuildIntArray() *Int {
```
stuartcarnie: I missed this in my first pass; this should be prefixed with `New` to indicate ownership, per the [Reference Counting](https://github.com/apache/arrow/tree/master/go#reference-counting) documentation

jsternberg: I've changed the prefix.

nathanielc:
@jsternberg Thanks for the benchmarks I am going to play around with them locally to get a better feel for things.

nathanielc:
@jsternberg OK, I messed around with the benchmarks locally and found a few things.

* The wrappers as they stand have about a ~18% overhead. 
* Using an embedded version of the Int64 wrapper was 15% faster than the pointer version of the wrapper and has ~2% overhead against the raw Arrow types.

Since this is performance critical code and the intent of the wrappers is to expose correct types and not hide the Arrow buffers I say we switch to using the embedded struct implementation.

Can you switch to using the embedded struct?

Also can you update the benchmarks to also test reading data from the built arrays? Locally I added a second loop that summed all the values from the constructed array. Something like that would be great. Just make sure to use the value in some way that the compiler can't optimize away.

jsternberg:
I tried to do the follow up PR for this (integrate arrow) with just the raw types since that seemed to be preferred. If I'm right that the utility wrappers would end up being useful, I think we can add them later without too much work.

That PR is at #488.

PR #480, opened 2018-12-12
docs: add a proposal for a standard library

### LINKED ISSUE #365: Design std lib for Flux
Once we have namespaces we can organize the built in code into various standard packages.

There are many existing issues that refer to placing code into a namespace. Please link them here so we can design out the entire stdlib.

wolffcm:
Fixes #365

### DIFF
--- a/docs/StandardLibrary.md
+++ b/docs/StandardLibrary.md
@@ -2,25 +2,15 @@
 
 The Flux standard library will define the set of functions, constants, and other
 values that are available out of the box with Flux, providing the functionality that is
-most frequently used.
+most frequently used. Everything in the Flux standard library is part of a package.  
+The top-level package is called "universe."
 
-Some functions in the standard library will be available without qualifying
-them with a namespace.  These will be the most common functions, to help Flux
-code be more concise.  Such functions are said to be in the "top level package."
+The Flux compiler provides the concept of a "prelude," which is the set of packages whose members
+may be used in a Flux program without namespace qualification.  For example,
+in the vanilla Flux CLI, the prelude includes the `universe` and `influxdata/influxdb`
+packages. 
 
-The remaining functions in the standard library will be divided up into
-discrete packages that are organized by topic.
-
-What's presented here is just a departure point for further discussion.
-
-## The Top Level Package
-
-Those functions that provide functionality that is very frequently
-used and should not need to be qualified with a namespace in Flux code.
-(Consider also the concept of "prelude")
-
-### Transformations
-Should some of these be in the `math` package instead?  `pearsonr` for example.
+## The `Universe` Package
 
 Fundamental transformations:
 - `bottom`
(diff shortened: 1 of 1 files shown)


docs/StandardLibrary.md
```diff
@@ -0,0 +1,334 @@
...
+The remaining functions in the standard library will be divided up into
+discrete packages that are organized by topic.
+
+What's presented here is just a departure point for further discussion.
+
+## The Top Level Package
+
+Those functions that provide functionality that is very frequently
+used and should not need to be qualified with a namespace in Flux code.
+(Consider also the concept of "prelude")
```
nathanielc: The more I think about it the more I like the concept of "prelude", it helps to remove lots of edge cases.

docs/StandardLibrary.md
```diff
@@ -0,0 +1,334 @@
...
+Others?
+
+## I/O Packages
+
+Should vendor-specific packages like `influx` be part of the standard library?  
+If Flux is to be used as a stand-alone language independent of InfluxDB, perhaps
+they should not.
+
+### Package `csv`
+- `from` (seems lonely here; could be in top level package?)
```
nathanielc: I think this is fine. Also I expect we could have some kind of `io.from` that takes a raw byte stream and produces a stream of tables. Then the `csv` package will have a `encoder` and `decoder` types.

And `csv.from` would be defined in terms of the `io` package:

    package csv
    from = (file,content) => io.from(encoder: decoder(file:file, contents:contents))

Or something similar. Obviously the Flux type system doesn't support this yet but we can figure that out separately.

In short I expect the csv package to grow.

docs/StandardLibrary.md
```diff
@@ -0,0 +1,334 @@
...
+- `from`
+- `to`
+
+### Package `influx`
+- `buckets`
+- `databases`
+- `fieldsAsCols`
+- `from`
+- `fromLP` [IMPL#70](https://github.com/influxdata/flux/issues/70)
+- `fromJSON`
```
nathanielc: As per my comment above, I think it would be nice to separate encoding/decoding from the transport of the data. Something to think about here.

Could you do this?

    http.from(url:"...", decoder: influxdb.line)

docs/StandardLibrary.md
```diff
@@ -0,0 +1,334 @@
...
+What's presented here is just a departure point for further discussion.
+
+## The Top Level Package
+
+Those functions that provide functionality that is very frequently
+used and should not need to be qualified with a namespace in Flux code.
+(Consider also the concept of "prelude")
+
+### Transformations
+Should some of these be in the `math` package instead?  `pearsonr` for example.
```
nathanielc: I think the math package should be restricted to working with raw numeric types. As such pearsonr would not be part of the math package, because it deals with tables and streams.

That said pearsonr could be placed into a different package, perhaps a "stats" package?

docs/StandardLibrary.md
wolffcm: I missed the all hands in September, so I'm not that familiar with the idea.  The "prelude" would be an option that is provided to the compiler, that is a list of packages that do not need namespace qualification?

docs/StandardLibrary.md
wolffcm: I like it.  Curious about what the package structure would look like if we went this route.  Would we need a set of transport packages, each of which provide a `from` and `to` that require encoders or decoders as arguments?
```
package http
package file
package kafka
...
```
As well as a set of encoder/decoder packages?  They could also provide `from` and `to` functions that use a default transport for the encoding.
```
package csv
package influx // both LP and Influx JSON
...
```
(Unclear to me where `sql` would fit in this kind of abstraction)

docs/StandardLibrary.md
wolffcm: I like this idea but I predict some confusion as to where some stream functions should live, since `mean` for example is a statistic but is so common that it may be reasonably expected to be top level.

One could imagine having it live in both places:
```
// Top level package
mean = stats.mean
```

wolffcm:
@nathanielc Should this be merged, closed, or updated?

PR #481, opened 2018-12-13
Feat/namespaces

### LINKED ISSUE #324: Implement Namespaces in the interpreter
Implement the `import` and `package` statements and scoping rules inside of the interpreter and the compiler.

See SPEC for details on how namespaces are defined.

How do namespaces interact with type inference?

jlapacik:
This PR adds support for packages and imports to the interpreter.
The Flux interpreter evaluates a single package. It is capable of
evaluating several programs belonging to the same package, however
a Flux package consisting of multiple programs is not well defined
and so the current behavior will most likely change.

An interpreter instance will error if given two programs from two
different packages to evaluate.

Closes #324 .

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -15,12 +15,129 @@ var topScope = NewScopeWithValues(map[string]values.Value{
 	"false": values.NewBool(false),
 })
 
+// Package implementation
+type packageObject struct {
+	name        string
+	scope       *Scope
+	private     map[string]bool
+	sideEffects []values.Value
+}
+
+func (p *packageObject) Name() string {
+	return p.name
+}
+
+func (p *packageObject) SideEffects() []values.Value {
+	return p.sideEffects
+}
+
+func (p *packageObject) Type() semantic.Type {
+	vals := p.scope.Values()
+	size := len(vals) - len(p.private)
+	typs := make(map[string]semantic.Type, size)
+	for k, v := range vals {
+		if !p.private[k] {
+			typs[k] = v.Type()
+		}
+	}
+	return semantic.NewObjectType(typs)
+}
+
+func (p *packageObject) PolyType() semantic.PolyType {
+	vals := p.scope.Values()
+	size := len(vals) - len(p.private)
+	typs := make(map[string]semantic.PolyType, size)
+	names := make(semantic.LabelSet, 0, size)
+	for k, v := range vals {
+		if !p.private[k] {
+			names = append(names, k)
+			typs[k] = v.PolyType()
+		}
+	}
+	return semantic.NewObjectPolyType(typs, nil, names)
+}
+
+func (p *packageObject) Get(name string) (values.Value, bool) {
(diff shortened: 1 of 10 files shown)


interpreter/interpreter.go
```diff
@@ -189,6 +271,11 @@ func (itrp *Interpreter) doVariableAssignment(declaration *semantic.NativeVariab
 		}
 	}
 	scope.Set(declaration.Identifier.Name, value)
+	// If this is a top level variable assignment,
+	// add it to the interpreter's package object.
+	if _, ok := itrp.pkg.types[declaration]; ok {
+		itrp.pkg.value.Set(declaration.Identifier.Name, value)
```
nathanielc: Would it be possible to use the scope as the values.Object implementation instead of copying the values into a separate object?

This way the `packageObject` could implment the values.Object interface directly and all it would need is a pointer to the correct top level scope. This would eliminate the need to track which var assignments are top level versus which are not. 

This would mean that above instead of this:

	scope.Set(name, pkg.Value())

You would have this:

	scope.Set(name, pkg)


If you do use the scope though you have to track which identifiers in the scope are public vs private. The only private identifiers would be the imported packages themselves.

Thoughts?

interpreter/package_test.go
```diff
@@ -0,0 +1,336 @@
...
+			program: ` 
+package foo
+import baz "path/to/bar"
+a = baz.x
+`,
+			want: values.NewObjectWithValues(
+				map[string]values.Value{
+					"a": values.NewInt(10),
+				}),
+		},
```
nathanielc: Why does this test not fail? The import is not defined.

jlapacik: Each test case has a list of imports that it defines via the `imports` field. Am I missing something?

interpreter/interpreter.go
jlapacik: I like it. Less memory consumption, and it will greatly simplify the `PolyType()` method. My only concern is like you say keeping track of private vs public identifiers. But I think the benefits outweighs the extra check in the `Get()` method.

interpreter/package_test.go
nathanielc: No, at first I thought this test case didn't have any imports defined. I misread it. It looks good. Maybe just a small nit to tab in the strings to be indented with the test case. Since we are not testing formatting or otherwise here it will be easier to read that way.

interpreter/interpreter.go
```diff
@@ -15,23 +15,145 @@ var topScope = NewScopeWithValues(map[string]values.Value{
 	"false": values.NewBool(false),
 })
 
+// Package implementation
+type packageObject struct {
+	name        string
+	scope       *Scope
+	private     map[string]struct{}
```
nathanielc: Can you make this a `map[string]bool`? Then you do have to do `_,ok:=pkg.private[foo]` you can just do `ok := pkg.private[foo]`, the zero value is returned when the key does not exists, which in this case means `false`.

PR #487, opened 2018-12-13
feat(plan): plan validation

### LINKED ISSUE #161: Planner needs a plan validator
Lots of things can go wrong with plans, and it will be useful to have a validator that can be run at any stage of planning to find potential problems.

Here are some things to check for:
- Are the edges well-formed?  Does each predecessor have a corresponding successor edge and vice-versa?  
- Do nodes have the expected number of predecessors and successors?  Join should have two predecessors, filter one, and so on
- For the `from` operation, were we able to push a range into it?  If not, planning needs to fail

affo:
Fixes #161

### DIFF
--- a/plan/logical_test.go
+++ b/plan/logical_test.go
@@ -571,3 +571,47 @@ func TestLogicalPlanner(t *testing.T) {
 		})
 	}
 }
+
+func TestLogicalIntegrityCheckOption(t *testing.T) {
+	script := `
+from(bucket: "telegraf")
+	|> filter(fn: (r) => r._measurement == "cpu")
+	|> yield(name: "result")
+`
+
+	spec, err := compile(script, time.Unix(0, 0))
+	if err != nil {
+		t.Fatalf("could not compile flux query: %v", err)
+	}
+
+	intruder := plantest.CreateLogicalMockNode("intruder")
+	k := plan.ProcedureKind(transformations.FilterKind)
+	// no integrity check enabled, everything should go smoothly
+	planner := plan.NewLogicalPlanner(
+		plan.OnlyLogicalRules(
+			plantest.SmashPlanRule{Intruder: intruder, Kind: k},
+			plantest.CreateCycleRule{Kind: k},
+		),
+		plan.DisableIntegrityChecks(),
+	)
+	_, err = planner.Plan(spec)
+	if err != nil {
+		t.Fatalf("unexpected fail: %v", err)
+	}
+
+	// let's smash the plan
+	planner = plan.NewLogicalPlanner(
+		plan.OnlyLogicalRules(plantest.SmashPlanRule{Intruder: intruder, Kind: k}))
+	_, err = planner.Plan(spec)
+	if err == nil {
+		t.Fatal("unexpected pass")
+	}
+
+	// let's introduce a cycle
+	planner = plan.NewLogicalPlanner(
+		plan.OnlyLogicalRules(plantest.CreateCycleRule{Kind: k}))
+	_, err = planner.Plan(spec)
+	if err == nil {
+		t.Fatal("unexpected pass")
+	}
+}
--- a/plan/physical.go
+++ b/plan/physical.go
@@ -54,6 +54,11 @@ func (pp *physicalPlanner) Plan(spec *PlanSpec) (*PlanSpec, error) {
 
 	// Ensure that the plan is valid
 	if !pp.disableValidation {
+		err := transformedSpec.CheckIntegrity()
+		if err != nil {
+			return nil, err
+		}
+
 		err = validatePhysicalPlan(transformedSpec)
 		if err != nil {
 			return nil, err
@@ -121,6 +126,7 @@ func OnlyPhysicalRules(rules ...Rule) PhysicalOption {
 	})
 }
 
+// Disables validation in the physical planner
 func DisableValidation() PhysicalOption {
 	return physicalOption(func(p *physicalPlanner) {
 		p.disableValidation = true
--- a/plan/plantest/rules.go
+++ b/plan/plantest/rules.go
@@ -27,6 +27,81 @@ func (sr *SimpleRule) Name() string {
 	return "simple"
 }
 
+// SmashPlanRule adds an `Intruder` as predecessor of the given `Node` without
+// marking it as successor of it. It breaks the integrity of the plan.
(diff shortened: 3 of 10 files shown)


plan/types.go
```diff
@@ -67,6 +67,47 @@ func (plan *PlanSpec) Replace(root, with PlanNode) {
...
+//     N1 <-------
+//               |--> R
+//     N2 --------
+//
+// Nor this one (R2 is only successor of N1, and, thus, it won't be visited):
+//
+//     N1 <----> R1
+//     |
+//     --------> R2
+//
```
affo: Should we perform deeper checks and cover the cases specified in the function doc?  
We could also check for cycles.

affo:
I started with a super small modification to start tackling your first point in the issue.

It took me a while to get into it, but it seems that it not so trivial to check what I describe in the doc of `CheckIntegrity`, because, by only tracking `Roots` in `PlanSpec`, we don't have the totality of the graph and cannot check for some issues that could arise.

To detect them I should adequate `PlanSpec` to keep every node ever added to it.

wolffcm:
What you have here is a good start.

I see what you are saying about not being able to catch every kind of problem.  I do not think that the plan spec should have a pointer to every node in the graph.  This means that if a node becomes orphaned and nothing points to it, it will be garbage collected and go away.  Hopefully our unit tests for rules are thorough enough to catch this.

What about this as a way to do the integrity check?
- Do a normal walk of the graph, which will follow predecessor edges.  Do edge symmetry checks and also check for cycles.
- During that walk, collect all of the nodes that have no predecessors (the sources)
- Do another walk of the graph with a custom function that starts from the sources and follows the successor edges.  Perform the same symmetry edge checks and cycle checks.

I think this is as much integrity as we can check without tracking everything.  Does this seem reasonable?

wolffcm:
Actually it occurs to me that the `PlanSpec.walk` method uses a `visited` set to ensure each node is visited just once, since the graph is a DAG.  Some nodes may be reachable by more than one path.  I don't think you can use either `BottomUpWalk` or `TopDownWalk` at all to do these checks.

affo:
> Does this seem reasonable?

More than reasonable, indeed. I am implementing it.

> Actually it occurs to me that the PlanSpec.walk method uses a visited set to ensure each node is visited just once, since the graph is a DAG. Some nodes may be reachable by more than one path. I don't think you can use either BottomUpWalk or TopDownWalk at all to do these checks.

I see, I couldn't get why you excluded using `walk` since the beginning, but now I see. Thank you!

affo:
I implemented a breadth-first because I feel it more intuitive and in order to add a new walk pattern in `plan.walk`.  
I can switch to a depth-first implementation, if needed.

plan/logical_test.go
```diff
@@ -571,3 +571,82 @@ func TestLogicalPlanner(t *testing.T) {
...
+	return "smashPlanRule"
+}
+
+func (smashPlanRule) Pattern() plan.Pattern {
+	return plan.Pat(transformations.FilterKind, plan.Any())
+}
+
+func (smashPlanRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool, error) {
+	node.AddPredecessors(plantest.CreateLogicalMockNode("intruder"))
+	return node, false, nil
```
wolffcm: nice :trollface:

plan/walk.go
```diff
@@ -84,3 +105,89 @@ func walk(node PlanNode, preFn, postFn func(PlanNode) error, visited map[PlanNod
...
+
+			visited[node] = struct{}{}
+		}
+
+		// build new level
+		newLevel := make([]PlanNode, 0)
+		for _, node := range level {
+			for _, pred := range node.Predecessors() {
+				if _, ok := visited[pred]; ok {
+					cycle = true
```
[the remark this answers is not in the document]
affo: you are right. Thank you

affo:
Now a test is failing because of the `FromDistinctRule`.

It seems that `updateSuccessors` clears the successors of both `newNode` and `oldNode` at the same time (https://github.com/influxdata/flux/blob/master/plan/heuristic_planner.go#L122), because they point at the same node (`distinct`), thus leaving `distinct` without successor, but `group2` with `distinct` ad predecessor.

It seems that the only cases in which you return the same node as the one passed to the rewrite rule (and tell the planner that something changed) is in `FromDistinctRule` and `FromFilterRule` (in other cases you return new nodes obtained from merging...). However `filter` is always the last operation in the tests, so there is no problem in clearing its successors and the problem remained hidden.

Long story short, is adding

```golang
if newNode == oldNode {
    return
}
``` 

In `updateSuccessors` a proper fix in your opinion?

wolffcm:
Yes, I think that's the right way to handle that.

Also, congratulations, seems like your validation work has found a bug, which is exactly what it's for 🎉 

Some background: at some point our planner will need to model the cost of different ways of implementing a query, e.g., it may need to choose between doing a hash join or a merge join to implement a join.  When it does this it will need to make a copy of a part of the plan, and choose which one to link to the successor.  So I intentionally made it so that the rules' `Rewrite` methods don't link to successors.  It should be up to the planner to do that as needed (which you found it was doing incorrectly)

plan/physical.go
```diff
@@ -90,8 +98,9 @@ func validatePhysicalPlan(plan *PlanSpec) error {
 
 type physicalPlanner struct {
 	*heuristicPlanner
-	defaultMemoryLimit int64
-	disableValidation  bool
+	defaultMemoryLimit     int64
+	disableValidation      bool
+	disableIntegrityChecks bool
```
wolffcm: Can `disableIntegrityChecks` be the same option as `disableValidation`?  The only reason to disable the checks is enable testing.

plan/plantest/rules.go
```diff
@@ -27,6 +27,70 @@ func (sr *SimpleRule) Name() string {
...
+	}
+
+	return plan.Pat(k, plan.Any())
+}
+
+func (spp SmashPlanRule) Rewrite(node plan.PlanNode) (plan.PlanNode, bool, error) {
+	if len(spp.Kind) > 0 || node == spp.Node {
+		node.AddPredecessors(spp.Intruder)
+	}
+	return node, false, nil
```
wolffcm: It's worth noting in a comment here that you're returning `false` even though the plan has changed.

PR #488, opened 2018-12-13
WIP: feat(execute): integrate arrow arrays into the table builder

jsternberg:


execute/table.go
```diff
@@ -327,57 +330,47 @@ func TablesEqual(left, right flux.Table, alloc *memory.Allocator) (bool, error)
...
-			}
+		rightBuffer := NewColListTableBuilder(right.Key(), alloc)
+		if err := AddTableCols(right, rightBuffer); err != nil {
+			return false, err
+		}
+		if err := AppendTable(right, rightBuffer); err != nil {
+			return false, err
 		}
 
+		fmt.Println(leftBuffer.NRows(), rightBuffer.NRows())
```
nathanielc: nit, please remove this.

result.go
```diff
@@ -31,6 +32,10 @@ type Table interface {
 	// The function f will be called zero or more times.
 	Do(f func(ColReader) error) error
 
+	// DoArrow calls f to process the data contained within the table.
+	// It uses the arrow buffers.
+	DoArrow(f func(ArrowColReader) error) error
```
nathanielc: This design looks good. The idea being that nothing changes right now and then we slowly move each transformation over to using this method. Once they are all complete we can then remove `Do` and `ColReader` and rename `DoArrow` and `ArrowColReader` to simply `Do` and `ColReader`.

PR #489, opened 2018-12-14
refactor(plan): make physical planning convert logical nodes to physical

wolffcm:
Currently the physical planner never actually produces physical nodes.  It just inherits the logical nodes from logical planning.  This commit addresses that, by adding a rule that does the conversion.

Currently all of our procedure specs are both logical and physical because they implement a `Cost` method, so this is just a refactoring with no change in behavior.  The next step will be to separate `from` into distinct logical and physical procedure specs.

### DIFF
--- a/plan/physical.go
+++ b/plan/physical.go
@@ -3,6 +3,8 @@ package plan
 import (
 	"fmt"
 	"math"
+
+	"github.com/pkg/errors"
 )
 
 // PhysicalPlanner performs transforms a logical plan to a physical plan,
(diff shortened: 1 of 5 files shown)


plan/physical.go
```diff
@@ -112,12 +121,50 @@ func OnlyPhysicalRules(rules ...Rule) PhysicalOption {
...
+func (physicalConverterRule) Rewrite(pn PlanNode) (PlanNode, bool, error) {
+	if _, ok := pn.(*PhysicalPlanNode); ok {
+		// Already converted
+		return pn, false, nil
+	}
+
+	ln := pn.(*LogicalPlanNode)
+	pspec, ok := ln.Spec.(PhysicalProcedureSpec)
+	if !ok {
+		// A different rule will do the conversion
```
jlapacik: So currently if we add a logical procedure that is not a physical procedure, the rewrite won't happen and and error will occur in the physical planner?

wolffcm: Yes, this is exactly why I had to add the default cost methods to generated yields.  Lots of the end-to-end tests that didn't have explicit yields were failing to plan.

Issue #486, opened 2018-12-13
Specify evaluating packages consisting of multiple source files


jlapacik:
The evaluation of a Flux package consisting of multiple source files is currently not well defined. Given a package `foo` consisting of files `a.flux`, `b.flux`, and `c.flux`, in what order should the files be evaluated by the interpreter?

Currently all statements in a source file are executed by the interpreter sequentially. But what if `a.flux` has a variable `a` that depends on a variable `b` defined in `b.flux`? Variable initialization order needs to be well defined at the package level.

jlapacik:
My thinking is that there should be a `Package` data structure defined in the flux `ast` package that would look something like this:

```go
type Package struct {
    name  string
    files map[string]Program
}
```

Each file belonging to a package would be parsed individually and added to the `files` map. The new signature of `semantic.New()` would look like this:

```go
// New creates a semantic graph from the provided AST
func New(pkg *ast.Package) (*Program, error)
```

That is, it would take an `ast.Package` and return a `semantic.Program` where the package level variable assignments have been re-ordered according to #533 . Note that location information of an ast/semantic node should be updated to included its associated filename. 

I think variables should remain mutable in the way it is currently defined in the SPEC. I think mutability is a useful feature especially once control flow is added to the language. In order to ensure that package level variables are initialized consistently, we should specify a consistent evaluation order (lexical) for the files of a package. And we should expect the user to follow certain best practices such as not re-declaring variables at the package level, while at the same time allowing them the capability.

Thoughts?

nathanielc:
I like the general idea.

This API feels odd:

```go
// New creates a semantic graph from the provided AST
func New(pkg *ast.Package) (*Program, error)
```

Maybe it should be?

```go
// New creates a semantic graph from the provided AST
func New(pkg *ast.Package) (*Package, error)
```

Maybe we translate the concepts as such:

* An ast.Program becomes an ast.File
* An ast.Package is added that is a set of ast.Files
* A semantic.Program becomes a semantic.File
* a semantic.Package is added that  is a set of semantic.Files

With these changes the concept of a program is no more, rather everything is in a package. Then the interpreter can execute a single package which makes sense based on the spec.

I think mutability makes programing in Flux easier, but it opens lots of questions. Here are some of my questions concerns:

If we allow for mutability then Flux becomes a partially order execution model. Meaning the order of statements only matter sometimes but not always. 

How would you resolve conflicts, like a variable is defined separately in two different files? Maybe that should be an error?

[...]

PR #533, opened 2018-12-14
feat(semantic): dependency analysis and assignment re-ordering

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

Relates to #486 .

This PR adds variable dependency analysis to the semantic package. It re-orders the variable assignments in declaration order after resolving dependencies. This is **one** way to handle the issue of evaluating packages that consist of multiple source files. Whether or not it's the right way is up for debate.

### DIFF
--- a/semantic/order.go
+++ b/semantic/order.go
@@ -1,0 +1,161 @@
+package semantic
+
+import "fmt"
+
+// OrderVarDependencies re-orders the variable assignments in a Flux
+// program according to declaration order after dependency analysis.
+// For example, given the following Flux package:
+//
+//    package foo
+//
+//    a = b + c
+//    b = f()
+//    c = f()
+//    d = 0
+//    f = () => d + 1
+//
+// The equivalent declaration ordering would be as follows:
+//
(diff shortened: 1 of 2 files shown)


nathanielc:
@jlapacik I liek this approach, it is simple and straightforward. A few questions?

How does this work when you have multiple files, do we need to change the concept of a Program to include multiple files?

If we accept this change we need to make a change to the SPEC. Specifically we need to specify that variables are immutable. This is something I was thinking we should do anyways as it simplifies a lot of static analysis and compiler logic. We would also need to update the implementation to explicitly disallow redeclaration instead of just constraining declarations to the same type. 

Thoughts? 

My reasoning for requiring immutable variables is because it means it is safe to reorder them, otherwise the order for a specific variable matters while cross variables doesn't, which seems confusing.

jlapacik:
@nathanielc I have put up a possible answer to your questions on #486 .

jlapacik:
Closing. This work will be tracked using #909 .

PR #541, opened 2018-12-17
fix(inputs): moving from to reimplement BucketAccessed

### LINKED ISSUE #114: Re-implement BucketsAccessed for `from`
was removed when refactoring `from` execution code to platform, while keeping the OpSpec on flux.  It's a genuine cyclic dependency that we need to work around.

affo:
Fixes #114 

Must be merged after #554 

This is only a tentative commit to see if this is a correct approach.  
I created a `MockFrom` operation and changed tests accordingly.  
I saw that controller tests are bound to the "correct" implementation of from (filter push down and so on...), so it should stay in `platform` with `from` and `from_test`.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


control/controller_test.go
```diff
@@ -1,454 +0,0 @@
-package control
```
wolffcm: So where is this test moving to?

functions/inputs/from.go
```diff
@@ -3,529 +3,75 @@ package inputs
...
-	"github.com/influxdata/flux/functions"
-	"github.com/influxdata/flux/functions/transformations"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
-	"github.com/influxdata/flux/values"
-	"github.com/pkg/errors"
 )
 
-const FromKind = "from"
+const MockFromKind = "mockFrom"
```
wolffcm: Ultimately I think it makes sense to have `from` live in the `platform` repo since the `flux` repo should be source-agnostic.  However, the `influxdb` repo (1.x version of the product) shares some code with `platform`, but has a separate implementation of `from`.  So whatever change is made should work for both 2.0 and 1.x.  There are longer-term plans to merge `influxdb` and `platform` into one repo I think?  I think we need some broader discussion on this.

I think it makes sense to have a `mockFrom` that lets us test `flux` in isolation.  But I don't think that it belongs in this directory with product code.  We could make it so that an `init` method in the `querytest` package will register `mockFrom`, so maybe it should live in `flux/querytest`.

Tests that need `mockFrom` to be available when compiling Flux can then just import `querytest` in addition to `flux/builtin` to get everything they need.

control/controller_test.go
affo: I thought that the entire controller + tests could go to `platform`

functions/inputs/from.go
affo: 👍

wolffcm: By the way, you can make it so that `mockFrom` does not need to have a range pushed into it.  That may make it easier to use.

There is at least one controller test that makes use of `from` needing a `range` pushed into it to introduce a planner failure.  I guess in that case you'll need to define a `mockRangeNeedingFrom` operation or something similar.

wolffcm:
@affo I mentioned in #554 that I think that it would be better to move both logical and physical to `platform`, and to have a dedicated mock for the `flux` repo.  

The `from` that will eventually be used in the Flux CLI will be quite different from the one we use to talk directly to InfluxDB.  I realize it's more work to move the whole thing, but it seems like the right design moving forward to me.

affo:
Created a `builtin` module for testing builtins in `querytest`. Maybe also @aanthony1243 's testing functions could go there.

The idea is to keep flux builtins and testing-specific builtins separated.

affo:
PR in `platform`: https://github.com/influxdata/platform/pull/2043

Still missing one point: how can platform unmarshal `FromOpSpec` given that it is using another one (`bucketAwareFromOpSpec` https://github.com/influxdata/platform/pull/2043/commits/6f96ca7c0edaecfaf5f9e815fa648ca21d69047d#diff-d6fc83da5d483b5697c651dedf781c2cR44)?

Should I keep the same OpSpec and convert to `platform.ID` here https://github.com/influxdata/platform/pull/2043/commits/6f96ca7c0edaecfaf5f9e815fa648ca21d69047d#diff-d6fc83da5d483b5697c651dedf781c2cR53?

In this case, how can I handle the decode error?

affo:
No problem at all.

I'll update this PR by moving the `FromOpSpec` to `platform`.  
Then, I'll wait for #554 to be merged and rebase so that I can further fix it.

I'll keep the PR on `platform` up-to-date just to make sure that it is ready when this one is ready.

I can postpone the `BucketAccessed` thing and create a new PR for that. Anyhow, it is very straightforward.

aanthony1243:
@affo I think @wolffcm is off for a while now, are there questions about this still?   I'm working on Dec 21 tomorrow,  let's talk in my AM to see if you an wrap this up before the new year or if we have to wait.

affo:
@aanthony1243 No questions by now. Anyways I'll ping you in your AM.

affo:
Closing in favor of #920

PR #548, opened 2018-12-17
docs: update grammar spec with new object literal syntax

jlapacik:
Closes https://github.com/influxdata/platform/issues/212 .

### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -565,7 +565,8 @@ Object literals construct a value with the object type.
 
     ObjectLiteral = "{" PropertyList "}" .
     PropertyList  = [ Property { "," Property } ] .
-    Property      = identifier ":" Expression .
+    Property      = identifier [ ":" Expression ]
+                  | string_lit ":" Expression .
 
 ##### Array literals
 
--- a/internal/parser/grammar.md
+++ b/internal/parser/grammar.md
@@ -79,7 +79,8 @@ The parser directly implements the following grammar.
     Block                      = "{" StatementList "}" .
     ExpressionList             = [ Expression { "," Expression } ] .
     PropertyList               = [ Property { "," Property } ] .
-    Property                   = identifer ":" Expression .
+    Property                   = identifier [ ":" Expression ]
+                               | string_lit ":" Expression .
     ParameterList              = [ Parameter { "," Parameter } ] .
     Parameter                  = identifer [ "=" Expression ] .
 


internal/parser/grammar.md
```diff
@@ -79,7 +79,8 @@ The parser directly implements the following grammar.
     Block                      = "{" StatementList "}" .
     ExpressionList             = [ Expression { "," Expression } ] .
     PropertyList               = [ Property { "," Property } ] .
-    Property                   = identifer ":" Expression .
+    Property                   = identifier [ ":" Expression ]
```
nathanielc: This means that you can mix implicit key properties with explicit. We do not want to allow them to be mixed. Its either all or none.

Specifically this should be invalid:

    {foo, bar: baz}

But these are fine:

    {foo, bar}
    {foo: foo, bar: baz}

jsternberg: I talked about this with Joshua and that logic is out of the domain for context free languages so it would get pushed to wherever we would handle it for errors.

jlapacik: If the above is not a parser error, then it would have to be handled in a separate static analysis step before type inference.

docs/SPEC.md
```diff
@@ -563,13 +563,9 @@ Literals construct a value.
 
 Object literals construct a value with the object type.
 
-    ObjectLiteral = "{" PropertyList "}" .
+    ObjectLiteral = "{" PropertyList | IdentList "}" .
```
jsternberg: You have to turn this alternation into another production because the operator precedence implies this is a left brace and a property list or an identity list followed by a right brace.

internal/parser/grammar.md
jlapacik: @jsternberg can you explain why the new grammar additions violate the context-free grammar definition? The example given by @nathanielc looks correct to me.

jsternberg: It would be more difficult to implement and, in my opinion at least, doesn't really get you anything other than more nodes. It's much more simple to just check the AST to see if it matches. I guess it would technically be possible, but you're trying to build something that's more easily built into validation checks than a context-free grammar.

There's also no reason why an additional ast.Node would be needed. The formatter would just say,  "Well there's no value for these keys, so the value must be the key."

So if you have a property list with implicit keys, it would just have a `nil` ast.Node for every value. If the formatter sees that, it knows to use the implicit keys version. There are then minimal changes to the AST. If you really wanted to, you could make a special "implicit" node that would get set so it shows up in the JSON, but I don't really think that's needed.

jlapacik: I don't have a really strong opinion either way. I guess one thing I don't care too much for is having differences between the parser grammar and the Flux grammar.

nathanielc: OK, lets define the grammar to allow for mixed modes. Then later we can enforce it as a validation step on the AST if we feel its helpful necessary. But lets not make that validation part of the initial implementation. Maybe its fine to allow them to be mixed, and we rely on best practice to not mix them.

Using `nil` as the Property value seems like a good solution.

PR #554, opened 2018-12-18
refactor(inputs): split FromProcedureSpec into logical and physical s…

### LINKED ISSUE #270: Separate "from" into separate logical and physical operations
Right now all of our procedures are both logical and physical.  We want to make logical and physical operations to be distinct.  The `from` procedure would be a good place to start.

wolffcm:
…pecs

The idea here is to have `FromProcedureSpec` be separate from `PhysicalProcedureSpec`.  The latter contains info on anything that physical planner rules have pushed down, and the push down rules only operate on the physical spec.

I had to add a converter rule that is specific to `from` nodes.  My thinking is that this rule will go away eventually, and that pushing down a `range` node will be required to create a `PhysicalFromProcedureSpec`.  It will be the job of the logical planner to ensure that `range` nodes are adjacent to the `from` source so that this can happen.

I also broke out the logical planner's `Plan` method into two methods, one to create the initial logical plan from the `flux.Spec`, and one to actually apply rules.

Fixes #270

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -187,7 +187,11 @@ func (c *Controller) compileQuery(q *Query, compiler flux.Compiler) error {
 
 	if q.tryPlan() {
 		// Plan query to determine needed resources
-		lp, err := c.lplanner.Plan(&q.spec)
+		ip, err := c.lplanner.CreateInitialPlan(&q.spec)
+		if err != nil {
+			return errors.Wrap(err, "failed to create initial logical plan")
+		}
+		lp, err := c.lplanner.Plan(ip)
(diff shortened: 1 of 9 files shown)


wolffcm:
@affo I'm not sure that using the logical `from` as the mock in the `flux` repo is the way to go.  The mock needs to be viable as both a logical and a physical node to be of use, I think.  It would still be better to have a dedicated mock, and to move both logical and physical `from` to `platform`.

PR #556, opened 2018-12-18
doc(spec): include type info in group key

### LINKED ISSUE #439: Explore including the type information into the group key to avoid schema collisions.
The idea is to avoid the mismatched type errors that we commonly encounter.

Update the SPEC with the proposed design.

affo:
Fixes #439 

Let's open a discussion here.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1193,7 +1193,7 @@ Examples:
 
 #### LoadLocation
 
-LoadLoacation loads a locations from a time zone database.
+LoadLocation loads a locations from a time zone database.
 
 LoadLocation has the following parameters:
 
@@ -1207,7 +1207,7 @@ Examples:
     loadLocation(name:"America/Chicago")
     loadLocation(name:"Africa/Tunis")
 
-[IMPL#157](https://github.com/influxdata/flux/issues/157) Implement LoadLoacation function
+[IMPL#157](https://github.com/influxdata/flux/issues/157) Implement LoadLocation function
 
 ## Data model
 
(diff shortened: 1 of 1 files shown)


aanthony1243:
This issue has become less important recently.  I don't want to close this, but it might be a  while before we come back to it.

affo:
Understood.  

I actually opened it not to forget about it 😄 We'll come back to it when needed.

affo:
For what concerns grouping, in principle, there shouldn't be any problem for the user.
He/she shouldn't rely on any specific number of tables inside the stream.
Even when grouping by nothing (i.e., `group()`), it's ok to say that the result contains
as many tables as the different schemas we have in input (more or less).

For what concerns implementation, it would affect operations that cause a regrouping:
 - `group`;
 - `map`;
 - `drop`/`keep`;
 - `set`;
 - `pivot`;
 - `union` (should already handle it).

Those are the ones that could output _less_  groups than in input (i.e., a merge).
Aggregators, selectors, and others produce per table results.  
For instance, `window` is not there, because it adds columns to the group key, rather
then removing them, and so, it causes existing groups to split and not to merge.

It seems that I should touch either:
 - `GroupKey` interface;
 - `TableBuilder`;
 - `GroupLookUp`.

The biggest problem is that, sometimes, the schema of the output table is not known when we need the table builder for some key, so we cannot add the info of the schema of the table we are building prior to building the table itself. The `TableBuilder` and `GroupLookUp` should become aware of the changes of the schema in order to properly adjust the cache, somehow. The code could become messy.

In terms of code, it would be easier to create 2 columns (e.g. `value_string` and `value_float`) once the collision is detected. However this is another approach for the one described by the current commit to the SPEC, and maybe changing the name to columns is not the best idea for users.

This is not exhaustive, but something :)

nathanielc:
I am going to close this PR for now. We need to answer some higher level questions about how the type system can allow or otherwise represent these kinds of schemas.

PR #558, opened 2018-12-18
refactor(internal): added a new utility for generating test data

jsternberg:
The utility is similar to inch in that it generates data, but it
generates data for multiple types (instead of only float) and can be
customized to mix data types and add additional tags.

It will output the generated data as a CSV file to be consumed as input
for tests. This way, we can generate small unique datasets for tests in
a more consistent way than previously.

### DIFF
--- a/internal/cmd/csvgen/main.go
+++ b/internal/cmd/csvgen/main.go
@@ -1,0 +1,463 @@
+package main
+
+import (
+	"container/heap"
+	"flag"
+	"fmt"
+	"math"
+	"math/rand"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/csv"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/values"
+)
+
+var Flags struct {
+	Start          string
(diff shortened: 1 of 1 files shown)


internal/cmd/csvgen/main.go
```diff
@@ -0,0 +1,463 @@
...
+		c, err := strconv.Atoi(cardinality[i])
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse %s as an integer: %s", cardinality[i], err)
+		}
+		m[k] = c
+	}
+	return m, nil
+}
+
+func main() {
```
aanthony1243: I'd really like to see this implemented as a source type inside of `flux` instead.  reasons in my full review, but most of the code here could go into a `Decode` method as shown here:  https://github.com/influxdata/flux/blob/c45e00c1dd69f8143edf0daff5e1a5efa98a446f/functions/inputs/from_generator.go#L164

jsternberg:
@aanthony1243 I separated the data generation part of this into #828.

My goal is to use the data generation part of this with `ast.Format` and integrate it to make something similar to `prepcsvtests`, but with data generation and have it work properly with the new testing framework.

jsternberg:
Closing this in favor of #828 and another PR will follow for using the library in that PR.

PR #561, opened 2018-12-18
feat(parser): add tests for multi-line and escaped strings

### LINKED ISSUE #544: Validate multi-line string literal support
Update the SPEC if the grammar is not complete

nathanielc:
Byte values in strings are now fully unescaped.

Fixes #544 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -747,6 +747,7 @@ func (i *PipeLiteral) Copy() Node {
 // StringLiteral expressions begin and end with double quote marks.
 type StringLiteral struct {
 	BaseNode
+	// Value is the unescaped value of the string literal
 	Value string `json:"value"`
 }
 
--- a/ast/format.go
+++ b/ast/format.go
@@ -376,9 +376,32 @@ func (f *formatter) formatIdentifier(n *Identifier) {
 }
 
 func (f *formatter) formatStringLiteral(n *StringLiteral) {
+	if n.Loc != nil && n.Loc.Source != "" {
+		// Preserve the exact literal if we have it
+		f.writeString(n.Loc.Source)
+		return
+	}
+	// Write out escaped string value
 	f.writeRune('"')
-	f.writeString(n.Value)
+	f.writeString(escapeStr(n.Value))
 	f.writeRune('"')
+}
+
+func escapeStr(s string) string {
+	if !strings.ContainsAny(s, `"\`) {
+		return s
+	}
+	var builder strings.Builder
+	// Allocate for worst case where every rune needs to be escaped.
+	builder.Grow(len(s) * 2)
+	for _, r := range s {
+		switch r {
+		case '"', '\\':
+			builder.WriteRune('\\')
+		}
+		builder.WriteRune(r)
+	}
+	return builder.String()
 }
 
 func (f *formatter) formatBooleanLiteral(n *BooleanLiteral) {
--- a/ast/format_test.go
+++ b/ast/format_test.go
@@ -3,6 +3,7 @@ package ast_test
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/parser"
 	"github.com/pkg/errors"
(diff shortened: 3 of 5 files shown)


jsternberg:
I think it's to retain the original text. As an example, both of these produce the same string literal, but have different source representation.

```
a = "a
"
b = "a\n"
```

@nathanielc is there any way you could use the `Source` attribute that is inside of the position? It seems dirty to me that we would have a property that was solely used for recreating the original. But, we technically have the exact literal that was used pre-parsing (which is what you are trying to produce) already in the node. Maybe that would be a better way of implementing this?

ast/ast.go
```diff
@@ -747,7 +747,11 @@ func (i *PipeLiteral) Copy() Node {
 // StringLiteral expressions begin and end with double quote marks.
 type StringLiteral struct {
 	BaseNode
+	// Value is the unescaped value of the string literal
 	Value string `json:"value"`
+	// Literal may be empty.
+	// When not empty it represents the literal string found in the source including quotes.
+	Literal string `json:"literal,omitempty"`
```
jsternberg: This should already be contained in the `BaseNode` within the `Source` attribute.

nathanielc:
@affo @jsternberg I have made the changes.

ast/format.go
```diff
@@ -376,9 +376,28 @@ func (f *formatter) formatIdentifier(n *Identifier) {
...
+		// Preserve the exact literal if we have it
+		f.writeString(n.Loc.Source)
+	} else {
+		f.writeRune('"')
+		f.writeString(escapeStr(n.Value))
+		f.writeRune('"')
+	}
+}
+
+func escapeStr(s string) string {
```
jsternberg: From what I've seen in the Go standard library, it's more efficient to call `strings.Index` to determine if there are any problematic characters to begin with or you can use `strings.Count` which would be even better. Then if there are characters that require an escape, you use this path (and with `strings.Count` you can know the exact size).

Otherwise, you use the value directly.

ast/format.go
```diff
@@ -376,9 +376,28 @@ func (f *formatter) formatIdentifier(n *Identifier) {
 }
 
 func (f *formatter) formatStringLiteral(n *StringLiteral) {
-	f.writeRune('"')
-	f.writeString(n.Value)
-	f.writeRune('"')
+	if n.Loc != nil && n.Loc.Source != "" {
+		// Preserve the exact literal if we have it
+		f.writeString(n.Loc.Source)
+	} else {
```
jsternberg: This violates the Go style guide: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

nathanielc: How? That guide is about errors, there is no error handling in the code. There are two mutually exclusive branches.

ast/format.go
nathanielc: Since it saves an allocation, it seems like a good optimization. Although this is not very performance critical code. I'll make the change as it is straightforward.

ast/format.go
jsternberg: You're right. I skimmed a bit too fast. At the same time, we generally try to return instead of using else in most other sections of the code as we apply that rule to more than just error handling.

Can you apply that same pattern here too even if there aren't errors?

nathanielc:
@jsternberg Updated. There is no `strings.CountAny` method so I used `strings.ContainsAny` instead, that will save an allocation for the happy path. Further optimizing to pre-allocate the exact amount is not likely to make any significant difference.

ast/format.go
```diff
@@ -376,9 +376,31 @@ func (f *formatter) formatIdentifier(n *Identifier) {
...
+		f.writeRune('"')
+		f.writeString(escapeStr(n.Value))
+		f.writeRune('"')
+	}
+}
+
+func escapeStr(s string) string {
+	if !strings.ContainsAny(s, `"\"`) {
```
jsternberg: Is this the correct argument to this? This appears to have `"` in the string twice. I think this should be `"\`.

nathanielc:
@jsternberg updated.

PR #566, opened 2018-12-19
Refactor/selector arrow

jlapacik:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


execute/selector.go
```diff
@@ -139,23 +141,23 @@ func (t *indexSelectorTransformation) Process(id DatasetID, tbl flux.Table) erro
 		return fmt.Errorf("unsupported selector type %v", valueCol.Type)
 	}
 
-	return tbl.Do(func(cr flux.ColReader) error {
+	return tbl.DoArrow(func(cr flux.ArrowColReader) error {
 		switch valueCol.Type {
 		case flux.TBool:
 			selected := s.(DoBoolIndexSelector).DoBool(cr.Bools(valueIdx))
-			return t.appendSelected(selected, builder, cr)
+			return t.appendSelected(selected, builder, arrow.ColReader(cr))
```
jsternberg: `appendSelected` should be modified to use the arrow column reader. I don't think there is anybody else calling it  and `arrow.ColReader` should be used sparingly because it is being deleted when we're done with this transition.

jlapacik: @jsternberg I should have written a comment or TODO here. Modifying `appendSelected` would indirectly require a modification to `execute.AppendRecord` and `execute.ValueForRow` which multiple functions use. So I decided it was probably best left untouched for right now.

jsternberg: Whenever you run into those kinds of functions, can you create an arrow version of them and say something like `AppendArrowRecord`?

That way we can do the same thing we will do with `DoArrow`. The main reason is because with this method, we have to manually go in and find the usages of `ColReader` and change them all. If we duplicate the `execute` helper methods, we can use GoLand to determine that nobody  uses the original method and then use refactor to change everyone over at the same time in a more mechanical process.

functions/transformations/sample_test.go
```diff
@@ -30,46 +30,6 @@ func TestSample_Process(t *testing.T) {
 		want   [][]int
 		fromor *transformations.SampleSelector
 	}{
-		{
```
jsternberg: These tests seem important. I had begun this work but hadn't finished it and I think we may need to make it easy to test things like a column reader with multiple calls since it will happen.

This is asking a lot and I know, but it's also work that we'll have to do at some point. Can you create a testing utility method to break up.

I think this can be done by figuring out how slicing works. The arrow library supports it, but I'm not 100% on when `Retain()` or `Release()` calls are required. I think this is a rough way of doing it.

```
func BoolSlice(vs *array.Boolean, i, j int) *array.Boolean {
    data := array.NewSliceData(vs.Data(), i, j)
    return array.NewBooleanData(data)
}
```

I think if you add that to our `arrow` utility package, that will work for data slices. It needs to be released after use,  but I'm not sure any other calls are needed.

Issue #523, opened 2018-12-14
Convert state tracking functions to use DoArrow


nathanielc:
This includes

* stateDuration
* stateCount

jsternberg:
I'm going to do this at the same time as #497.

jsternberg:
Fixed via #574.

Issue #509, opened 2018-12-14
Convert map to use DoArrow


nathanielc:

jsternberg:
It's easiest to do this in combination with filter since they share the same code.

jsternberg:
#497 for the related issue.

jsternberg:
Fixed via #574.

PR #574, opened 2018-12-19
refactor(functions): convert filter, map, and state tracking to use arrow

### LINKED ISSUE #497: Convert filter to use DoArrow

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

These three functions all use row predicates so they are related to
each other. They have been converted over to use arrow buffers instead
of the old column reader.

Fixes #497, #509, and #523.

### DIFF
--- a/functions/transformations/state_tracking.go
+++ b/functions/transformations/state_tracking.go
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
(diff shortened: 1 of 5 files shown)


functions/transformations/state_tracking.go
```diff
@@ -5,6 +5,8 @@ import (
 	"log"
 	"time"
 
+	"github.com/influxdata/flux/values"
```
wolffcm: This import is out of order with the others.

jsternberg: Fixed.

PR #575, opened 2018-12-19
refactor(functions): convert set function to use arrow

### LINKED ISSUE #518: Convert set to use DoArrow

jsternberg:
- [ ] docs/SPEC.md updated
- [ ] Test cases written

Fixes #518.


execute/table.go
```diff
@@ -213,6 +213,54 @@ func AppendCol(bj, cj int, cr flux.ColReader, builder TableBuilder) error {
...
+			}
+		}
+		return nil
+	case flux.TInt:
+		return builder.AppendInts(bj, cr.Ints(cj).Int64Values())
+	case flux.TUInt:
+		return builder.AppendUInts(bj, cr.UInts(cj).Uint64Values())
+	case flux.TFloat:
+		return builder.AppendFloats(bj, cr.Floats(cj).Float64Values())
+	case flux.TString:
```
affo: I saw that last commits in `arrow` contain the `String` type. Could you add a TODO here just to remember to switch to the new implementation once its done?

jsternberg: I think that's probably a good idea, but I'd like to avoid it temporarily unless I have to repeat this code in more than one place.

PR #578, opened 2018-12-20
refactor(functions): convert shift to use arrow

jlapacik:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


affo:
Oh, I see. So there is no difference in doing what you did or this: https://github.com/influxdata/flux/blob/2243e125f6902bea7b75d736975b5329166b8a30/functions/tests/testdata/pivot.in.csv?

I switched to the second one because I saw some tests like this and others like yours, and it looked better to me 😅 

Anyways, no problem :)

jlapacik:
I don't think there's a difference, I just noticed more examples without the extra headers so I decided to do it that way. I'm actually not sure which way is preferred (if there is a preferred way).

affo:
Nathaniel was just confused about my use of headers, so I suppose that it should be done without headers 😄 

Here's where I saw the other approach: https://github.com/influxdata/flux/pull/575/files.

jlapacik:
I guess we're a bit inconsistent then. Both versions parse, but I don't want to confuse anyone either 😁.

PR #585, opened 2018-12-20
fix(functions): embed plan.DefaultCost in input and output functions

### LINKED ISSUE #584: running "buckets()" produces "failed to create physical plan" error
This also likely affects the other input/output functions too: `toHTTP()`, `toKafka()`.

wolffcm:
This addresses issues about "failed to create physical plan" in
input/output functions.  This was not caught earlier because there
are no end-to-end tests for input/output functions.

Fixes #584

### Done checklist
- [x] docs/SPEC.md updated (no change needed)
- [x] Test cases written (need end-to-end tests for i/o functions)

### DIFF
--- a/functions/inputs/buckets.go
+++ b/functions/inputs/buckets.go
@@ -37,6 +37,7 @@ func (s *BucketsOpSpec) Kind() flux.OperationKind {
 }
 
 type BucketsProcedureSpec struct {
+	plan.DefaultCost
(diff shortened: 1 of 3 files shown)


wolffcm:
@affo That's right. If there were tests then I would have caught this earlier. Adam is working on  solving the problem of having end to end tests for sources.

PR #608, opened 2019-01-02
refactor(arrow): add append utilities for appending arrow arrays to a builder

jsternberg:


arrow/table.go
```diff
@@ -0,0 +1,63 @@
...
+
+func AppendFloats(tbl TableAppender, j int, vs *array.Float64) error {
+	return tbl.AppendFloats(j, vs.Float64Values())
+}
+
+func AppendStrings(tbl TableAppender, j int, vs *array.Binary) error {
+	var buf [1]string
+	for i := 0; i < vs.Len(); i++ {
+		buf[0] = vs.ValueString(i)
+		fmt.Println("appending", buf[0])
```
jlapacik: Can you remove this print statement?

jsternberg: My mistake. That's how I debugged the slice problem.

PR #570, opened 2018-12-19
refactor(functions): converting limit to use arrow arrays

affo:
Fixes #508 

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

PR #616, opened 2019-01-03
refactor(functions): convert difference to use arrow arrays

### LINKED ISSUE #495: Convert difference to use DoArrow

affo:
Closes #495

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


functions/transformations/difference.go
```diff
@@ -178,68 +180,68 @@ func (t *differenceTransformation) Process(id execute.DatasetID, tbl flux.Table)
...
+	return tbl.DoArrow(func(cr flux.ArrowColReader) error {
 		l := cr.Len()
 
 		if l != 0 {
 			for j, c := range cols {
 				d := differences[j]
 				switch c.Type {
 				case flux.TBool:
-					if err := builder.AppendBools(j, cr.Bools(j)[firstIdx:]); err != nil {
+					if err := appendSlicedCol(j, firstIdx, c.Type, cr, builder); err != nil {
```
jlapacik: Instead of `appendSlicedCol` could you not have just used the `arrow.AppendBools` utility directly here?

functions/transformations/difference_test.go
```diff
@@ -223,33 +223,36 @@ func TestDifference_Process(t *testing.T) {
 				},
 			}},
 		},
-		{
-			name: "float with tags",
-			spec: &transformations.DifferenceProcedureSpec{
-				Columns: []string{execute.DefaultValueColLabel},
+		// TODO(affo): won't pass until https://github.com/apache/arrow/issues/3270 get's fixed
```
jlapacik: I missed this on my first pass, but it looks like this is the only transformation that slices arrow arrays. I think we should wait for https://github.com/apache/arrow/pull/3275 and https://github.com/apache/arrow/pull/3283 to land so we don't introduce breaking changes.

affo: There is also limit actually #570 

But its tests passes because string columns have always the same value I think.

Anyhow that's ok.

PR #619, opened 2019-01-03
docs: variables, options, packages, and imports

### LINKED ISSUE #486: Specify evaluating packages consisting of multiple source files

jlapacik:
Closes #486 .

### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -360,7 +360,7 @@ An option represents a storage location for any value of a specified type.
 Options are mutable.
 An option can hold different values during its lifetime.
 
-Below is a list of all options that are currently implemented in the Flux language:
+Below is a list of some built-in options that are currently implemented in the Flux language:
 
 * now
 * task
@@ -571,7 +571,7 @@ Examples:
 
 #### Option assignment
 
-    OptionAssignment = "option" identifier "=" Expression
+    OptionAssignment = "option" [ identifier "." ] identifier "=" Expression
 
 An option assignment creates an option bound to an identifier and gives it a type and a value.
 Options may only be assigned in a package block.
@@ -846,23 +846,20 @@ If a package has imports, all imported packages are initialized before any optio
 
 ### Variable initialization
 
-Variables are initialized in declaration order but only after all of the variables they depend on have been initialized.
-For packages consisting of multiple source files, declaration order depends on the order in which the files are presented to an interpreter.
-However package variables are initialized with the same value regardless of said file order.
+Variables are initialized and assigned the same value regardless of the order they are declared in the source.
 A variable cannot have a direct or indirect dependency on itself.
 
 ### Option initialization
 
-Options are initialized in declaration order.
 An option cannot have dependencies on any other options assigned in the same package block.
 
 ### Package imports
 
-Imported packages are initialized in declaration order.
+Packages imported in the same file block are initialized in declaration order.
 A package cannot be imported twice in the same file block.
-A package will only be initialized once even if it is imported multiple times in a package.
-A package may be imported purely for its side effects however any such package cannot be imported twice within the same package.
-A package cannot import two separate packages that modify the same option.
+A package will only be initialized once even if it is imported multiple times (across file blocks) in a package.
+Initializing imported packages must be deterministic.
+To be more precise, after all imported packages are initialized, each imported option must be assigned the same value regardless of the order in which package files are presented to an interpreter.
 
 ## Built-ins
 
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -350,8 +350,51 @@ The regular expression syntax is defined by [RE2](https://github.com/google/re2/
 
 ### Variables
 
-A variable holds a value.
-A variable can only hold values defined by its type.
+A variable represents a storage location for a single value.
+Variables are immutable.
+Once a variable is given a value, it holds that value for the remainder of its lifetime.
+
+### Options
+
+An option represents a storage location for any value of a specified type.
+Options are mutable.
+An option can hold different values during its lifetime.
+
+Below is a list of all options that are currently implemented in the Flux language:
+
+* now
+* task
+* location
+
+##### now
+
+The `now` option is a function that returns a time value to be used as a proxy for the current system time.
+
+    // Query should execute as if the below time is the current system time
+    option now = () => 2006-01-02T15:04:05-07:00
+
+##### task
+
+The `task` option is used by a scheduler to schedule the execution of a Flux query.
+
+    option task = {
+        name: "foo",        // name is required
+        every: 1h,          // task should be run at this interval
+        delay: 10m,         // delay scheduling this task by this duration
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -350,8 +350,51 @@ The regular expression syntax is defined by [RE2](https://github.com/google/re2/
...
+Variables are immutable.
+Once a variable is given a value, it holds that value for the remainder of its lifetime.
+
+### Options
+
+An option represents a storage location for any value of a specified type.
+Options are mutable.
+An option can hold different values during its lifetime.
+
+Below is a list of all options that are currently implemented in the Flux language:
```
nathanielc: Since options can be arbitrarily created, I think we should word this as here are "some" built in options. This way we do not imply that this is the list of all options.

docs/SPEC.md
```diff
@@ -480,57 +523,67 @@ A _block_ is a possibly empty sequence of statements within matching brace brack
...
+    x = 5.4
+    f = () => {
+        n = "a"
+        m = "b"
+        return a + b
+    }
+
+#### Option assignment
 
+    OptionAssignment = "option" identifier "=" Expression
```
nathanielc: We need to update this grammar to allow for a qualified identifier, i.e. `options alert.state = ...`

docs/SPEC.md
```diff
@@ -839,6 +833,32 @@ Side effects can occur in two ways.
...
+## Package initialization
+
+A package with no imports is initialized by first initializing and assigning all declared options and then subsequently initializing and assigning all declared variables.
+In particular all package options are initialized before any variables.
+Options cannot depend on any variables declared in the same package block.
+If a package has imports, all imported packages are initialized before any options or variables.
+
+### Variable initialization
+
+Variables are initialized in declaration order but only after all of the variables they depend on have been initialized.
```
nathanielc: I think this should just say the variables initialization and declaration order is not important. The spec does not need to state how we accomplish it.

docs/SPEC.md
```diff
@@ -839,6 +833,32 @@ Side effects can occur in two ways.
...
+For packages consisting of multiple source files, declaration order depends on the order in which the files are presented to an interpreter.
+However package variables are initialized with the same value regardless of said file order.
+A variable cannot have a direct or indirect dependency on itself.
+
+### Option initialization
+
+Options are initialized in declaration order.
```
nathanielc: Same here the order is unimportant and so we should not specify what it is.

docs/SPEC.md
```diff
@@ -839,6 +833,32 @@ Side effects can occur in two ways.
...
+An option cannot have dependencies on any other options assigned in the same package block.
+
+### Package imports
+
+Imported packages are initialized in declaration order.
```
nathanielc: This is the only declaration order that matters.

docs/SPEC.md
```diff
@@ -839,6 +833,32 @@ Side effects can occur in two ways.
...
+A package cannot be imported twice in the same file block.
+A package will only be initialized once even if it is imported multiple times in a package.
+A package may be imported purely for its side effects however any such package cannot be imported twice within the same package.
+A package cannot import two separate packages that modify the same option.
```
nathanielc: This part is not quite true. We want to be able to import packages that modify the same option. 

  I think we should not be so specific about the restrictions on package imports rather state that any package imports must be defined in an unambiguous manner. 

So instead of saying that packages cannot be imported twice for side effects etc, we simply state that package imports must be deterministic. Then define what we mean by deterministic, i.e. for any packages that modify the same options their order is known. Then define the order as known for import in the same file but not across files.

docs/SPEC.md
```diff
@@ -480,57 +523,67 @@ A _block_ is a possibly empty sequence of statements within matching brace brack
...
+
+An option assignment creates an option bound to an identifier and gives it a type and a value.
+Options may only be assigned in a package block.
+An identifier assigned to an option may be reassigned a new value but not a new type.
+An option keeps the same type for the remainder of its lifetime.
+
+Examples:
+
+    option n = 1
+    option n = 2
```
nathanielc: Can you give an example of a qualified option here?

docs/SPEC.md
```diff
@@ -839,6 +833,29 @@ Side effects can occur in two ways.
...
+If a package has imports, all imported packages are initialized before any options or variables.
+
+### Variable initialization
+
+Variables are initialized and assigned the same value regardless of the order they are declared in the source.
+A variable cannot have a direct or indirect dependency on itself.
+
+### Option initialization
+
+An option cannot have dependencies on any other options assigned in the same package block.
```
nathanielc: Does this mean that options can depend on options from other packages?

docs/SPEC.md
```diff
@@ -839,6 +833,29 @@ Side effects can occur in two ways.
...
+
+### Package imports
+
+Packages imported in the same file block are initialized in declaration order.
+A package cannot be imported twice in the same file block.
+A package will only be initialized once even if it is imported multiple times (across file blocks) in a package.
+Initializing imported packages must be deterministic.
+To be more precise, after all imported packages are initialized, each imported option must be assigned the same value regardless of the order in which package files are presented to an interpreter.
```
nathanielc: I think we could simplify these three sections into a single ordered list followed by some additional notes an constraints. I am thinking something like this:

```
## Package initialization

Variables options and imports are all initialized in a specific order:

1. All imported packages are initialized following this package initialization process. 
2. All options in the current package block are assigned regardless of order. An option cannot have dependencies on any other options assigned in the same package block.
3. All variables in the current package block are assigned regardless of order. A variable cannot have a direct or indirect dependency on itself.


A package cannot be imported twice in the same file block. A package will only be initialized once across all files blocks in all packages. 

Initializing imported packages must be deterministic. 
Specifically after all imported packages are initialized, each option must be assigned the same value.
Packages imported in the same file block are initialized in declaration order. 
Packages imported across different file blocks have no known order. 
When a set of imports modify the same option, they must be ordered by placing them in the same file block. 

```

I have tried to capture all the constraints restrictions that we need. 


> A package cannot be imported twice in the same file block. 

Do we need this? Seems like it would just be a noop if it occurred.

docs/SPEC.md
jlapacik: Good question. The way the spec is worded, yes. Allowing options to have dependencies on external options certainly won't introduce any nondeterminism so it's possible. It doesn't affect the implementation at all, but it could introduce some unexpected behavior for the user that's difficult to debug. I'll defer to you on whether we should keep it.

jlapacik:
@nathanielc can this be merged?

PR #620, opened 2019-01-03
feat(ast): add File and Package nodes to the AST

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
The CI is because of a bug in the Copy method. Also we do not have test coverage over the Copy methods I so I am adding that now.

internal/parser/parser.go
```diff
@@ -30,16 +30,17 @@ type Scanner interface {
...
+// ParseFile parses Flux source and produces an ast.File.
+func ParseFile(f *token.File, src []byte) *ast.File {
 	p := &parser{
 		s: &scannerSkipComments{
 			Scanner: scanner.New(f, src),
 		},
 		src: src,
 	}
-	return p.parseProgram()
+	file := p.parseFile()
```
jsternberg: Can you pass the filename as a parameter in `parseFile`? That's the general pattern that I've kept with the parser so far.

internal/token/fileset.go
```diff
@@ -4,6 +4,16 @@ import (
 	"github.com/influxdata/flux/ast"
 )
 
+type FileSet struct {
+	files []*File
+}
+
+func (f *FileSet) AddFile(filename string, size int) *File {
```
jsternberg: It might be reasonable to wrap around the `go/token` package again. The intention of `FileSet` is that a `token.Pos` can be used to lookup which file it came from and, since I took this API from `go/token`, I think it would be helpful not to change that.

But, for the sake of this PR, I do not think this is urgent and so it doesn't need to be changed in this PR.

jsternberg:
Also, the filename probably needs to be added to `SourceLocation`. This is so we can include the filename in errors.

internal/token/fileset.go
nathanielc: This makes sense, I mimiced that API but left the implementation hollow. I think a separate PR to get FileSet and File all working makes sense here.

PR #622, opened 2019-01-03
refactor(parser): panic when expect gets an incorrect token with gofuzz

jsternberg:


internal/parser/parser_default.go
```diff
@@ -0,0 +1,31 @@
+//+build !gofuzz
+
+package parser
+
+import "github.com/influxdata/flux/internal/token"
+
+// expect will continuously scan the input until it reads the requested
+// token. If a token has been buffered by peek, then the token type
+// must match expect or it will panic. This is to catch errors in the
```
jlapacik: `expect` doesn't appear to panic if it receives a token type different from the buffered token type?

jsternberg: I need to update that doc. I removed that when it caused panics that in some wrong places.

jsternberg: Fixed the doc.

PR #627, opened 2019-01-05
Feat/qualified options

jlapacik:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

The spec is updated in #619 . This PR just adds namespace qualified options to Flux and nothing more. Changes are reflected from the parser through to the interpreter. There's no more need for separate global and option scope objects anymore in the interpreter but to keep current PR small this will come as a follow on PR.

### DIFF
--- a/ast/edit/option_editor.go
+++ b/ast/edit/option_editor.go
@@ -44,9 +44,13 @@ func OptionValueFn(expr ast.Expression) OptionFn {
 // an object expression, or if some key in the provided map is not a property of the object.
 func OptionObjectFn(keyMap map[string]values.Value) OptionFn {
 	return func(opt *ast.OptionStatement) (ast.Expression, error) {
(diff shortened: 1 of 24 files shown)


ast/edit/option_editor.go
```diff
@@ -152,15 +156,16 @@ type optionEditor struct {
 
 func (v *optionEditor) Visit(node ast.Node) ast.Visitor {
 	if os, ok := node.(*ast.OptionStatement); ok {
-		if os.Assignment.ID.Name == v.identifier {
+		// Cannot edit namespace qualified options
```
nathanielc: Why not? We are editing the AST it seems practical to be able to edit qualified options.

jlapacik: Wasn't sure if it was allowed so decided not to change it. I can modify this now or merge and modify as a separate PR.

PR #636, opened 2019-01-07
feat(execute/table): TableBuilder interface and ColListTableBuilder i…

aanthony1243:
…mplementation support creation of nil values

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -700,6 +700,7 @@ type TableBuilder interface {
 	SetString(i, j int, value string) error
 	SetTime(i, j int, value Time) error
 	SetValue(i, j int, value values.Value) error
+	SetNil(i, j int) error
 
 	// Append will add a single value to the end of a column.  Will set the number of
 	// rows in the table to the size of the new column. It's the caller's job to make sure
@@ -711,6 +712,7 @@ type TableBuilder interface {
 	AppendString(j int, value string) error
 	AppendTime(j int, value Time) error
 	AppendValue(j int, value values.Value) error
+	AppendNil(j int) error
 
 	// AppendBools and similar functions will append multiple values to column j.  As above,
 	// it will set the numer of rows in the table to the size of the new column.  It's the
@@ -721,6 +723,7 @@ type TableBuilder interface {
 	AppendFloats(j int, values []float64) error
 	AppendStrings(j int, values []string) error
 	AppendTimes(j int, values []Time) error
+
 	// TODO(adam): determine if there's a useful API for AppendValues
 	// AppendValues(j int, values []values.Value)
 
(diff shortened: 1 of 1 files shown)


execute/table.go
```diff
@@ -721,10 +723,11 @@ type TableBuilder interface {
 	AppendFloats(j int, values []float64) error
 	AppendStrings(j int, values []string) error
 	AppendTimes(j int, values []Time) error
+
 	// TODO(adam): determine if there's a useful API for AppendValues
 	// AppendValues(j int, values []values.Value)
 
-	// GrowBools and similar functions will extend column j by n zero-values for the respective type.
+	// GrowBools and similar functions will extend column j by n nil-values for the respective type.
```
jsternberg: I think it would probably be better to keep this function the same. That way, you aren't setting a bunch of values on a map that may not need to be set.

execute/table.go
```diff
@@ -750,10 +753,16 @@ type TableBuilder interface {
...
+type rowColIndex struct {
+	row int
+	col int
+}
+
 type ColListTableBuilder struct {
 	key     flux.GroupKey
 	colMeta []flux.ColMeta
 	cols    []columnBuilder
+	nils    map[rowColIndex]bool
```
jsternberg: I didn't see this before. Can this be an array of map values?

execute/table.go
```diff
@@ -754,6 +757,7 @@ type ColListTableBuilder struct {
 	key     flux.GroupKey
 	colMeta []flux.ColMeta
 	cols    []columnBuilder
+	nils    []map[int]bool
```
nathanielc: This would be much more efficient and easier to follow if it was a `map[struct{i ,j int}]bool` I know I said that this is basically throw away code but managing a list of maps is cumbersome vs just a single map.

jsternberg: I personally find the list easier because the thing that builds the array goes column by column.

execute/table.go
```diff
@@ -1194,6 +1216,63 @@ func (b *ColListTableBuilder) AppendValue(j int, v values.Value) error {
...
 
+func (b *ColListTableBuilder) SetNil(i, j int) error {
+	if j < 0 || j > len(b.cols) {
+		return fmt.Errorf("set nil: column does not exist, index out of bounds: %d", j)
+	}
+	if i < 0 || i > b.cols[j].Len() {
+		return fmt.Errorf("set nil: row  does not exist, index out of bounds: %d", i)
+	}
+
+	if len(b.nils) != len(b.cols) {
```
jsternberg: This should be done in `AddCol`. Basically when you add the new column, do `append(b.nils, nil)` to give it a nil map.

PR #637, opened 2019-01-07
design(functions/fill): add design for fill operation

aanthony1243:
### Done checklist
- [X] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1238,6 +1238,21 @@ Example:
 
 **Note:** The `yield` function produces side effects.
 
+#### Fill
+
+Fill will scan a stream for null values and replace them with a non-null value.  
+
+The output stream will be the same as the input stream, with all null values in the column replaced.  
+
+Fill has the following properties: 
+
+* `column`  string
+    the column to fill
+* `value`        bool, int, uint, float, string, time
+    the constant value to use in place of nulls.  The type must match the type of the valueColumn.  
+* `usePrevious`  bool
+    if set, then assign the value set in the previous non-null row.  Cannot be used with `value`.  
+
 #### AssertEquals
 
 AssertEquals is a function that will test whether two streams have identical data.  It also outputs the data from the tested stream unchanged, so that this function can be used to perform in-line tests in a query.  


docs/SPEC.md
```diff
@@ -1218,6 +1218,21 @@ Example:
 
 **Note:** The `yield` function produces side effects.
 
+#### Fill
+
+Fill will scan a stream for nil values and replace them with a non-null value.
```
rbetts: s/non-null/non-nil/?

docs/SPEC.md
```diff
@@ -1218,6 +1218,21 @@ Example:
 
 **Note:** The `yield` function produces side effects.
 
+#### Fill
+
+Fill will scan a stream for nil values and replace them with a non-null value.  
+
+The output stream will be the same as the input stream, with all null values replaced.
```
rbetts: nil/null

docs/SPEC.md
```diff
@@ -1218,6 +1218,21 @@ Example:
...
+Fill will scan a stream for nil values and replace them with a non-null value.  
+
+The output stream will be the same as the input stream, with all null values replaced.  
+
+Fill has the following properties: 
+
+* `valueColumn`  string
+    the column to fill
+* `value`        bool, int, uint, float, string, time
+    the constant value to use in place of nulls
```
rbetts: Does the `value` type have to match the column type?

docs/SPEC.md
aanthony1243: the rest of the spec uses `null`.  I'm ok with changing it, but it'd need to change everywhere.

aanthony1243: ...except for this PR, where I use one of each for every other sentence.   Fixed all instances to null

docs/SPEC.md
```diff
@@ -1218,6 +1218,21 @@ Example:
 
 **Note:** The `yield` function produces side effects.
 
+#### Fill
+
+Fill will scan a stream for null values and replace them with a non-null value.  
+
+The output stream will be the same as the input stream, with all null values replaced.
```
nathanielc: `with all null values in the column replaced`

docs/SPEC.md
```diff
@@ -1218,6 +1218,21 @@ Example:
...
 
+#### Fill
+
+Fill will scan a stream for null values and replace them with a non-null value.  
+
+The output stream will be the same as the input stream, with all null values replaced.  
+
+Fill has the following properties: 
+
+* `valueColumn`  string
```
nathanielc: This should just be `column` to be consistent.

PR #640, opened 2019-01-08
feat(functions): columns function

### LINKED ISSUE #626: Create a columns function
Create a function called `columns` that will produce a table that has one row for each column in the input table, whose name is placed in the `_value` column.

This is to replace functionality that is lost by fixing `keys` to produce a more intuitive result, per this issue: https://github.com/influxdata/flux/issues/625

affo:
Fixes #626 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2238,6 +2238,38 @@ from(bucket: "telegraf/autogen")
 Records are grouped into a single table.  
 The group key of the resulting table is empty.
 
+#### Columns
+
+Columns lists the column labels of input tables.
+For each input table, it outputs a table with the same group key columns, plus a new column containing the labels of the input table's columns.
+Each row in an output table contains the group key value and the label of one column of the input table.
+So, each output table has the same number of rows as the number of columns of the input table.
+
+Columns has the following properties:
+
+| Name   | Type   | Description                                                        |
+| ----   | ----   | -----------                                                        |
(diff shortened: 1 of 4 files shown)


affo:
For the weird final call to `Do`, see #643

wolffcm:
Looks good to me.  @nathanielc may wish to review the change to the SPEC for `columns` as well as `keys` (#641)

docs/SPEC.md
```diff
@@ -2144,6 +2144,32 @@ from(bucket: "telegraf/autogen")
 Records are grouped into a single table.  
 The group key of the resulting table is empty.
 
+#### Columns
+
+Columns lists the column labels of input tables.
+For each input table, it outputs a table with the same group key columns, plus a `_value` column containing the labels of the input table's columns.
+Each row in an output table contains the group key value and the label of one column of the input table.
+So, each output table has the same number of rows as the number of columns of the input table.
```
nathanielc: This looks great. The `_value` should be a parameter to the function with a default value of `_value.

affo: Ok, shouldn't do the same for `keys`?

nathanielc:
Make sure to use `BREAKING CHANGE: ` in the commit message to indicate the breaking change. Thanks!

PR #641, opened 2019-01-08
fix(functions): fix keys to output group key

### LINKED ISSUE #625: Behavior of keys function is unintuitive
One would expect that the `keys` function would just produce one row for each group key column of a table.  But the SPEC says otherwise:
> Keys outputs a table with the input table's group key columns, plus a _value column containing the names of the input table's columns.

Per consensus discussion on Slack, `keys` should be made to produce one row for each of the input table's group key columns (instead of one row per column).  The `except` parameter would also be made redundant so it can be removed.  The SPEC should also be updated to reflect the new behavior.

affo:
Fixes #625 

Must be merged after #506 

Maybe this could be turned into a built-in once we have schema introspection functions.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2247,9 +2247,9 @@ So, each output table has the same number of rows as the number of columns of th
 
 Columns has the following properties:
 
-| Name   | Type   | Description                                                        |
-| ----   | ----   | -----------                                                        |
-| column | string | Name of the column used to store the labels. Defaults to `_value`. |
+| Name   | Type   | Description                                                                               |
+| ----   | ----   | -----------                                                                               |
+| column | string | Column is the name of the output column to store the column labels. Defaults to `_value`. |
 
 Example:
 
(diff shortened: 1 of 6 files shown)


affo:
For the weird final call to `Do`, see #643

wolffcm:
Looks good to me.  @nathanielc may wish to review the change to the SPEC.

affo:
Great. I add Nathaniel as reviewer.

affo:
@nathanielc

The function that uses `keys` is `_tagKeys`, and I suppose that this fix makes it more close to IFQL behavior given that returns only the group key. Maybe that query was wrong before, because it needed something like:

```
...
  |> keys(except: ["_value", "_time"])
...
```

Now the same result is achieved without `except`.

I'd say that, just to be more close to IFQL, if the purpose is to return _only tags_, the query should be:

```
_tagKeys = (bucket, predicate=(r) => true) =>
  from(bucket:bucket)
    |> range(start:-24h)
    |> filter(fn:predicate)
    |> keys()
    |> filter(fn: (r) => r._value != "_measurement")
    |> filter(fn: (r) => r._value != "_start")
    |> filter(fn: (r) => r._value != "_stop")
    |> filter(fn: (r) => r._value != "_time")
    |> filter(fn: (r) => r._value != "_field")
```

`_tagValues`, instead creates issues when used in `_measurements`, because of mixed types issues. It should be something like:

```
_tagValues = (bucket, tag, predicate=(r) => true) =>
  from(bucket:bucket)
    |> range(start:24h)
    |> filter(fn: predicate)
    |> keep(columns: ["_start", "_stop", "_time", "_field", tag])
    |> group(columns:[tag])
    |> distinct(column:tag)
    |> group(columns:["_stop","_start"])
```

`keep` removes the problem of mixed types. Even if adding it makes `distinct` drop `_start` and `_stop` columns. I don't know why.

Anyways I'll paste these new versions as a comment in that PR.

stdlib/tests/testdata/meta_query_keys.flux
```diff
@@ -37,32 +37,22 @@ outData = "
...
 ,,3,usage_iowait,cpu,cpu-total,host.local,_field
 ,,3,usage_iowait,cpu,cpu-total,host.local,_measurement
-,,3,usage_iowait,cpu,cpu-total,host.local,_start
-,,3,usage_iowait,cpu,cpu-total,host.local,_stop
 ,,3,usage_iowait,cpu,cpu-total,host.local,cpu
 ,,3,usage_iowait,cpu,cpu-total,host.local,host
 ,,4,usage_irq,cpu,cpu-total,host.local,_field
 ,,4,usage_irq,cpu,cpu-total,host.local,_measurement
-,,4,usage_irq,cpu,cpu-total,host.local,_start
-,,4,usage_irq,cpu,cpu-total,host.local,_stop
```
nathanielc: Instead of changing the output of the test here, can you instead add _start and _stop to the group key in the input data as that is closer to the way normal meta queries will work.

docs/SPEC.md
```diff
@@ -2240,31 +2240,31 @@ The group key of the resulting table is empty.
...
 
-Keys has the following parameters: 
+Keys has the following properties:
 
-| Name   | Type     | Description                                                                               |
-| ----   | ----     | -----------                                                                               |
-| except | []string | Expect is a list of keys not to include in the output. Defaults to `["_time", "_value"]`. |
+| Name   | Type   | Description                                                                  |
+| ----   | ----   | -----------                                                                  |
+| column | string | Name of the column used to store the group key labels. Defaults to `_value`. |
```
nathanielc: nit, I have been following Go comment convention with these parameter comments. https://blog.golang.org/godoc-documenting-go-code Specifically the part where the comment starts with the name of the thing being commented and is a complete sentence. 

For example this could change to:

    Column is the name of the output column to store the group key labels. Defaults to `_value`.

nathanielc: I missed this on the `columns` function review, its been merged already but if you could fix it in the PR that would be great. Its a small but important change.

affo: Uops, missed this comment!

PR #686, opened 2019-01-08
refactor(executetest): support arrow and nil values in ConvertTable

jsternberg:


execute/executetest/table.go
```diff
@@ -380,29 +380,39 @@ func ConvertTable(tbl flux.Table) (*Table, error) {
...
+	err := tbl.DoArrow(func(cr flux.ArrowColReader) error {
 		l := cr.Len()
 		for i := 0; i < l; i++ {
 			row := make([]interface{}, len(blk.ColMeta))
 			for j, c := range blk.ColMeta {
-				var v interface{}
 				switch c.Type {
 				case flux.TBool:
-					v = cr.Bools(j)[i]
+					if col := cr.Bools(j); !col.IsNull(i) {
```
aanthony1243: fwiw there's also `col.IsValid(i)`  if you want to get rid of the `!`

execute/executetest/table.go
```diff
@@ -380,29 +380,39 @@ func ConvertTable(tbl flux.Table) (*Table, error) {
...
-					v = cr.Times(j)[i]
+					if col := cr.Times(j); !col.IsNull(i) {
+						row[j] = values.Time(col.Value(i))
+					}
 				default:
 					panic(fmt.Errorf("unknown column type %s", c.Type))
 				}
-				row[j] = v
 			}
 			blk.Data = append(blk.Data, row)
```
aanthony1243: All of this looks OK, you leave the interface value empty/nil  in the case where the flux.Table  has a null value.  But what happens to this table next?   I feel like we also need to update the implementations of  Do()  and DoArrow above to be safe.

jsternberg: That's in a different PR. Let me rebase this one and use `IsValid`.

PR #690, opened 2019-01-08
refactor(functions): support null values in the min selector

### LINKED ISSUE #661: Update min to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #661.

### DIFF
--- a/functions/transformations/min.go
+++ b/functions/transformations/min.go
@@ -131,33 +131,39 @@ func (s *MinSelector) selectRow(idx int, cr flux.ArrowColReader) {
 
 func (s *MinIntSelector) DoInt(vs *array.Int64, cr flux.ArrowColReader) {
 	minIdx := -1
-	for i, v := range vs.Int64Values() {
-		if !s.set || v < s.min {
-			s.set = true
-			s.min = v
-			minIdx = i
+	for i := 0; i < vs.Len(); i++ {
+		if vs.IsValid(i) {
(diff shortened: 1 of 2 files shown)


functions/transformations/min.go
```diff
@@ -131,33 +131,69 @@ func (s *MinSelector) selectRow(idx int, cr flux.ArrowColReader) {
...
+	if vs.NullN() == 0 {
+		for i, v := range vs.Float64Values() {
+			if !s.set || v < s.min {
+				s.set = true
+				s.min = v
+				minIdx = i
+			}
+		}
+	} else {
+		for i := 0; i < vs.Len(); i++ {
```
nathanielc: Is there a meaningful difference between 

    for i, v := range vs.Float64Values()

and 

    for i := 0; i < vs.Len(); i++ {

Does is some how relate to how nulls are handled? If not then I suggest we use a consistent loop iteration mechanism.

jsternberg: i don't know honestly. I assumed it might relate to function calls and whether it could be inlined or not. I'm not sure if that's a concern or not at the moment.

nathanielc: My understanding is that the IsValid call should be inlined independent of the loop iteration method.

jsternberg: Changed to only use the second method.

PR #691, opened 2019-01-08
feat(csv): treat omitted values with no defaults as null in csv

### LINKED ISSUE #632: CSV decoding/encoding with Nil values
Specify how to encode/decode nil values in the CSV format.

wolffcm:
Fixes #632.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/csv/result.go
+++ b/csv/result.go
@@ -45,6 +45,8 @@ const (
 	uintDatatype   = "unsignedLong"
 
 	timeDataTypeWithFmt = "dateTime:RFC3339"
+
+	nullValue = ""
 )
 
 // ResultDecoder decodes a csv representation of a result.
@@ -391,7 +393,15 @@ func readMetadata(r *csv.Reader, c ResultDecoderConfig, extraLine []string) (tab
 				cols[j].fmt = desc
 			}
 		}
-		if defaults[j] != "" {
+		if defaults[j] == nullValue {
+			defaultValues[j] = values.NewNull(flux.SemanticType(cols[j].ColMeta.Type))
+		} else if defaults[j] == "" {
+			// for now, the null value is always represented with "", so this is
+			// unreachable.
+			// When we support the #null annotation we'll want to distinguish
+			// between "" (for strings) and null here.
+			panic("unreachable")
+		} else {
 			v, err := decodeValue(defaults[j], cols[j])
 			if err != nil {
 				return tableMetadata{}, errors.Wrapf(err, "column %q has invalid default value", label)
@@ -603,16 +613,15 @@ func (d *tableDecoder) init(line []string) error {
 	for j, c := range d.meta.Cols {
 		if d.meta.Groups[j] {
 			var value values.Value
-			if d.meta.Defaults[j] != nil {
-				value = d.meta.Defaults[j]
-			} else if record != nil {
+			if record != nil && record[j] != "" {
+				// TODO: consider treatment of nullValue here
 				v, err := decodeValue(record[j], c)
 				if err != nil {
 					return err
 				}
 				value = v
 			} else {
-				return fmt.Errorf("missing value for group key column %q", c.Label)
+				value = d.meta.Defaults[j]
 			}
 			keyCols = append(keyCols, c.ColMeta)
 			keyValues = append(keyValues, value)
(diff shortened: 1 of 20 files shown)


csv/result.go
```diff
@@ -397,6 +397,8 @@ func readMetadata(r *csv.Reader, c ResultDecoderConfig, extraLine []string) (tab
 				return tableMetadata{}, errors.Wrapf(err, "column %q has invalid default value", label)
 			}
 			defaultValues[j] = v
+		} else {
+			defaultValues[j] = values.NewNull(flux.SemanticType(cols[j].ColMeta.Type))
```
nathanielc: Can you add a constant for the `nullValue` and set it to `""` for now? We want to make it clear how the null value interacts with the default value. The if statement above it would become switch statement with three cases:

```go
switch defaults[j] {
    case nullValue:
        // Set explicit null as default
        defaultValues[j] = values.NewNull(flux.SemanticType(cols[j].ColMeta.Type))
    case "":
        // Do nothing there is no default
    default:
        v, err := decodeValue(defaults[j], cols[j])
        // ...
}
```
Technically the second case will not be possible until `nullValue` is dynamic. But it makes it clear that its still possible to not have a default value.

execute/executetest/table.go
```diff
@@ -48,9 +48,14 @@ func (t *Table) Normalize() {
 			if len(t.Data) > 0 {
 				t.KeyValues[j] = t.Data[0][idx]
 			}
-			v := values.New(t.KeyValues[j])
-			if v.Type() == semantic.Invalid {
-				panic(fmt.Errorf("invalid value: %s", t.KeyValues[j]))
+			var v values.Value
+			if t.KeyValues[j] == nil {
+				v = values.NewNull(flux.SemanticType(t.ColMeta[idx].Type))
```
jsternberg: Is there a way to make it so `values.New(nil)` works  instead of adding a new function?

csv/result.go
```diff
@@ -1104,23 +1104,38 @@ func encodeValue(value values.Value, c colMeta) (string, error) {
 	}
 }
 
-func encodeValueFrom(i, j int, c colMeta, cr flux.ColReader) (string, error) {
+func encodeValueFrom(i, j int, c colMeta, cr flux.ArrowColReader) (string, error) {
+	var v string
```
nathanielc: We should use the `nullValue` constant here too

execute/table.go
```diff
@@ -1192,6 +1192,10 @@ func (b *ColListTableBuilder) SetValue(i, j int, v values.Value) error {
 }
 
 func (b *ColListTableBuilder) AppendValue(j int, v values.Value) error {
+	if v.IsNull() {
```
aanthony1243: This is so useful and safety, I want to propose making AppendBool, AppendInt, etc.  all private functions, and forcing us to use  AppendValue(j, values.New(false)), etc.

values/values.go
```diff
@@ -148,6 +152,13 @@ func New(v interface{}) Value {
 	}
 }
 
+func NewNull(t semantic.Type) Value {
```
aanthony1243: this is great.  I'm getting closer to voting that we can only add values.Value  types to table builders, to make sure we don't have any errors of omission re: Nulls.

execute/executetest/table.go
aanthony1243: I thought of that too, but a values.Value has to have a type.

wolffcm: We want `null` values to have a type, so I don't see a way to do it just by passing `nil`.  If `New` accepted a pointer to a value, then it would be possible by passing a null pointer... but that seems kind of hacky to me.

execute/table.go
wolffcm: I suspect we'll still want to expose the methods that accept raw types for performance reasons, since `values.Value` uses `interface{}` and reflection under the hood.  Were it not for this I would totally agree with you.

csv/result.go
wolffcm: I can't write the code like you've done it here because Go complains about duplicate cases, but I will add `nullValue` and comments to this effect.

PR #692, opened 2019-01-08
feat(execute/aggregate): aggregates process empty/all-null tables by …

aanthony1243:
…creating a null row

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/execute/aggregate_test.go
+++ b/execute/aggregate_test.go
@@ -172,6 +172,116 @@ func TestAggregate_Process(t *testing.T) {
 			},
 		},
 		{
+			name:   "empty table",
+			config: execute.DefaultAggregateConfig,
+			agg:    sumAgg,
+			data: []*executetest.Table{
+				{
+					KeyCols: []string{"_start", "_stop"},
+					ColMeta: []flux.ColMeta{
+						{Label: "_start", Type: flux.TTime},
+						{Label: "_stop", Type: flux.TTime},
+						{Label: "_time", Type: flux.TTime},
+						{Label: "_value", Type: flux.TFloat},
+					},
+					KeyValues: []interface{}{
+						execute.Time(100),
+						execute.Time(200),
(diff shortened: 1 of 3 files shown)


execute/aggregate_test.go
```diff
@@ -1,15 +1,14 @@
...
-	"sort"
-	"testing"
-
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
 	"github.com/influxdata/flux/functions/transformations"
+	"sort"
```
jsternberg: Imports here got messed up.

execute/aggregate_test.go
```diff
@@ -171,6 +170,224 @@ func TestAggregate_Process(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:   "empty table",
```
jsternberg: Can you name this something different to indicate that there is a table with data in here? Or remove the first table. I looked at this originally and got confused why the table wasn't empty, but I see now there are two tables. I have the same comment for the other tests too.

execute/aggregate_test.go
```diff
@@ -171,6 +171,224 @@ func TestAggregate_Process(t *testing.T) {
...
+						{Label: "_stop", Type: flux.TTime},
+						{Label: "_value", Type: flux.TFloat},
+					},
+					Data: [][]interface{}{
+						{execute.Time(100), execute.Time(200), nil},
+					},
+				},
+			},
+		},
+		{
```
jsternberg: Same thing from above here too. It might be easier to just remove the first table to reduce confusion. I'm not sure it adds anything to the test.

PR #693, opened 2019-01-08
refactor(functions): support null values in the max selector

### LINKED ISSUE #663: Update max to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #663.

### DIFF
--- a/functions/transformations/max.go
+++ b/functions/transformations/max.go
@@ -131,33 +131,39 @@ func (s *MaxSelector) selectRow(idx int, cr flux.ArrowColReader) {
 
 func (s *MaxIntSelector) DoInt(vs *array.Int64, cr flux.ArrowColReader) {
 	maxIdx := -1
-	for i, v := range vs.Int64Values() {
-		if !s.set || v > s.max {
-			s.set = true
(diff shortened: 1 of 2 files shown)


functions/transformations/max.go
```diff
@@ -131,33 +131,39 @@ func (s *MaxSelector) selectRow(idx int, cr flux.ArrowColReader) {
 
 func (s *MaxIntSelector) DoInt(vs *array.Int64, cr flux.ArrowColReader) {
 	maxIdx := -1
-	for i, v := range vs.Int64Values() {
-		if !s.set || v > s.max {
-			s.set = true
-			s.max = v
-			maxIdx = i
+	for i := 0; i < vs.Len(); i++ {
```
wolffcm: I guess you could check to see if there are any nulls first and if there are none, have a tighter loop for that case.  But this is probably fine.

jsternberg: Related to this comment in the min version of this PR: https://github.com/influxdata/flux/pull/690#discussion_r246133915

PR #777, opened 2019-01-09
refactor(functions): add test confirming that limit works with null values

### LINKED ISSUE #664: Update limit to understand Null

affo:
Fixes #664 

Is it ok to return the `N` records let them be null or not?  
Should I add an option for discarding null values from the count?

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/limit_test.go
+++ b/stdlib/universe/limit_test.go
@@ -6,8 +6,8 @@ import (
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
-	"github.com/influxdata/flux/stdlib/universe"
 	"github.com/influxdata/flux/querytest"
+	"github.com/influxdata/flux/stdlib/universe"
 )
 
 func TestLimitOperation_Marshaling(t *testing.T) {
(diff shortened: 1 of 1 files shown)


arrow/table.go
```diff
@@ -12,49 +12,105 @@ type TableAppender interface {
...
 	}
+
+	return nil
+}
+
+func setNils(tbl TableAppender, j int, vs array.Interface) error {
+	if vs.NullN() > 0 {
+		for i := 0; i < vs.Len(); i++ {
+			if vs.IsNull(i) {
+				if err := tbl.SetNil(i, j); err != nil {
```
aanthony1243: this won't be correct under some conditions, since the column-size may be != `len(vs)` if we just appended data to an existing table.   what we need is for the table-builder to be null-aware for all exported API's.

functions/transformations/limit_test.go
```diff
@@ -54,6 +78,35 @@ func TestLimit_Process(t *testing.T) {
...
+					{execute.Time(2), 1.0},
+				},
+			}},
+			want: []*executetest.Table{{
+				ColMeta: []flux.ColMeta{
+					{Label: "_time", Type: flux.TTime},
+					{Label: "_value", Type: flux.TFloat},
+				},
+				Data: [][]interface{}{
+					{execute.Time(2), nil},
```
aanthony1243: I think this looks correct. Limit should return nulls.  Filter can remove nulls if they don't want them. resubmit this test in a separate PR after we settle the Builder interface?

arrow/table.go
affo: Yeah, in the last commit I pushed a better solution, but it appends elements one by one... Definitely need to put these utilities directly in the builder so that you can get the row offset for that particular column and set nils accordingly.

functions/transformations/limit_test.go
affo: I can also keep this PR open and rebase/fix once the PR for the builder will be merged. Do you prefer to accept this as it is with the new code for setting nils?

The change in the table builder won't affect this function because it only depends on `AppendColsArrow` that will use the new builder interface.

affo:
There is a bug discovered in limit: must be merged after #793

affo:
@aanthony1243 rebased to support new table builder and your changes to limit. I also removed the edits to arrow utils. The PR is now only an addition for null tests

PR #783, opened 2019-01-09
refactor: rename DoArrow to Do

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

This removes the old `ColReader` and renames `ArrowColReader` to
`ColReader`. Anywhere that `Arrow` was used in a name is now removed as
it is the default.


jsternberg:
I have put up PRs for both platform and idpe that use this change.

jsternberg:
I'm going to start the merge train.

PR #788, opened 2019-01-09
refactor(interpreter): variable scoping

jlapacik:
BREAKING CHANGES: function signature for flux.Eval

### Done checklist
- [x] docs/SPEC.md updated (See https://github.com/influxdata/flux/pull/619)
- [x] Test cases written


compile.go
```diff
@@ -787,31 +788,107 @@ func ToQueryTime(value values.Value) (Time, error) {
 	}
 }
 
-var builtinImporter = &importer{
-	Packages: make(map[string]interpreter.Package),
+type pkg struct {
```
nathanielc: How does this type differ from the `interpreter.packageObject` type?

The stdlib uses a map of `*pkg` types while the `Interpreter.Package()` method returns an `*interpreter.packageObject` type. Why do we need both types? It seems that only use of `Interpreter.Package` is to get at the side effects as for testing.

interpreter/interpreter.go
```diff
@@ -892,19 +664,33 @@ type Value interface {
 
 type function struct {
 	e     *semantic.FunctionExpression
-	scope *Scope
+	scope Scope
 
-	typ      semantic.Type
-	polyType semantic.PolyType
+	types     map[semantic.Node]semantic.Type
+	polyTypes map[semantic.Node]semantic.PolyType
```
nathanielc: Are these basically a cache of all child nodes of the function expression so that later when we need to look up types we have a copy?

interpreter/interpreter.go
```diff
@@ -148,89 +116,71 @@ func (p *packageObject) Equal(rhs values.Value) bool {
...
+		},
+		Assignments: make([]*semantic.ExternalVariableAssignment, 0, scope.Size()),
 	}
-	// Add declarations for values in scope
-	addExternalAssignments(extern, itrp.options)
-	addExternalAssignments(extern, itrp.globals)
+
+	// TODO(jlapacik): Does it make sense to range over all variables currently in scope?
+	// Perhaps type inference should happen before evaluation as a separate step?
+	scope.Range(func(k string, v values.Value) {
```
nathanielc: This seems good enough for now. We just need to tell the semantic inference logic that these names exist in scope and their types.

interpreter/interpreter_test.go
```diff
@@ -165,11 +167,9 @@ func TestEval(t *testing.T) {
             six = six()
             nine = nine()
 
-            answer = (not (fortyTwo() == six * nine)) or fail()
+            not (fortyTwo() == six * nine) or fail()
 			`,
 			want: []values.Value{
-				values.NewFloat(6.0),
-				values.NewFloat(9.0),
```
nathanielc: Why did this test need to change?

interpreter/package_test.go
```diff
@@ -301,9 +299,6 @@ func TestInterpreter_EvalPackage(t *testing.T) {
 				map[string]values.Value{
 					"x": values.NewInt(10),
 				}),
-			sideEffects: []values.Value{
-				values.NewInt(10),
-			},
```
nathanielc: Why does `x = 10` not create a side effect anymore?

interpreter/scope.go
```diff
@@ -0,0 +1,19 @@
...
+	Nest() Scope
+
+	Size() int
+
+	Range(f func(k string, v values.Value))
+
+	SetReturn(values.Value)
+
+	Return() values.Value
+}
```
nathanielc: Can you add some doc comments to the interface and its methods?

scope.go
```diff
@@ -0,0 +1,76 @@
+package flux
```
nathanielc: Why is this in the `flux` package? Is it used outside the `interpreter` package at all?

interpreter/interpreter_test.go
jlapacik: Variable declarations aren't side effects so I got rid of them to reduce confusion.

interpreter/package_test.go
jlapacik: I thought it was confusing to have all top level statements be considered side effects when really the only statement that should be considered side effect producing is an expression statement.

scope.go
jlapacik: The interpreter only uses the Scope interface. The only place where we create a concrete implementation of the Scope interface is for defining the prelude in the `flux` package so I decided to place it there.

interpreter/interpreter.go
jlapacik: Yes this is in place of the former TypeScope. This is just temporary and will be removed eventually.

compile.go
```diff
@@ -47,54 +47,72 @@ func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spe
 	}
 
 	s, _ := opentracing.StartSpanFromContext(ctx, "parse")
-	itrp := NewInterpreter()
-	itrp.SetOption(nowOption, nowFunc(now))
-	if err := Eval(itrp, q); err != nil {
+
+	functionCalls, scope, err := Eval(q, SetOption(nowOption, nowFunc(now)))
```
nathanielc: nit, `functionCalls` is a bit of a misnomer, should we just call it `sideEffects`?

compile.go
```diff
@@ -47,54 +47,72 @@ func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spe
...
 	if o.verbose {
 		log.Println("Query Spec: ", Formatted(spec, FmtJSON))
 	}
 	return spec, nil
 }
 
-// Eval evaluates the flux string q and update the given interpreter
-func Eval(itrp *interpreter.Interpreter, q string) error {
-	astPkg := parser.ParseSource(q)
+func Eval(flux string, opts ...scopeMutator) ([]values.Value, interpreter.Scope, error) {
```
nathanielc: nit, can you make `scopeMutator` public? Since it it part of the public API here.

interpreter/package_test.go
```diff
@@ -29,114 +25,13 @@ func (imp *importer) Import(path string) (semantic.PackageType, bool) {
...
-		if !ok || !v.Equal(val) {
-			equal = false
-			return
-		}
-	})
-	return equal
-}
-
+// TODO(jlapacik): re-work these tests
+/*
```
nathanielc: What needs to be done here?

interpreter/scope.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type Scope interface {
+	// Lookup a name in the current scope
+	Lookup(name string) (values.Value, bool)
+
+	// Bind a variable in the current scope
+	Set(name string, v values.Value)
+
+	// Create a new scope by nesting the current scope
```
nathanielc: Can you specify how `values.Object` is used? I.e. it  populates the initial values of the nested scope, and may be nil.

compile.go
jlapacik: Yes I meant to change that.

interpreter/package_test.go
jlapacik: These tests defined their imports as Flux source. The interpreter used to evaluate them and then add them to an importer. But the interpreter returns side effects now, not an entire package. Not sure how to rework them without a lot of extra verbosity, but they seem useful so I've commented them out for right now to return to them later.

PR #789, opened 2019-01-09
refactor(functions): support null values in percentile

### LINKED ISSUE #660: Update percentile to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #660.

### DIFF
--- a/functions/transformations/percentile.go
+++ b/functions/transformations/percentile.go
@@ -11,6 +11,7 @@ import (
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
 	"github.com/influxdata/tdigest"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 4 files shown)


functions/transformations/percentile.go
```diff
@@ -5,6 +5,8 @@ import (
 	"math"
 	"sort"
 
+	"github.com/influxdata/flux/values"
```
wolffcm: This import is out of order.

functions/transformations/percentile_test.go
```diff
@@ -374,6 +374,33 @@ func TestPercentileSelector_Process(t *testing.T) {
...
+					{Label: "_time", Type: flux.TTime},
+					{Label: "_value", Type: flux.TFloat},
+					{Label: "t1", Type: flux.TString},
+					{Label: "t2", Type: flux.TString},
+				},
+				Data: [][]interface{}{
+					{nil, nil, "a", nil},
+				},
+			}},
+		},
```
wolffcm: Would be good to have a test with null values in the `_value` column in a non-empty input table.

functions/transformations/percentile.go
jsternberg: Fixed.

functions/transformations/percentile_test.go
jsternberg: Done.

PR #790, opened 2019-01-09
feat(execute/table): Refactored the table builder interfaces to support null value creation.

aanthony1243:
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -196,35 +196,17 @@ func AppendCol(bj, cj int, cr flux.ColReader, builder TableBuilder) error {
 
 	switch c.Type {
 	case flux.TBool:
-		vs := cr.Bools(cj)
-		for i := 0; i < vs.Len(); i++ {
-			if err := builder.AppendBool(bj, vs.Value(i)); err != nil {
-				return err
-			}
-		}
-		return nil
+		return builder.AppendBools(bj, cr.Bools(cj))
 	case flux.TInt:
-		return builder.AppendInts(bj, cr.Ints(cj).Int64Values())
+		return builder.AppendInts(bj, cr.Ints(cj))
 	case flux.TUInt:
-		return builder.AppendUInts(bj, cr.UInts(cj).Uint64Values())
+		return builder.AppendUInts(bj, cr.UInts(cj))
 	case flux.TFloat:
-		return builder.AppendFloats(bj, cr.Floats(cj).Float64Values())
+		return builder.AppendFloats(bj, cr.Floats(cj))
 	case flux.TString:
-		vs := cr.Strings(cj)
-		for i := 0; i < vs.Len(); i++ {
-			if err := builder.AppendString(bj, vs.ValueString(i)); err != nil {
-				return err
-			}
-		}
-		return nil
+		return builder.AppendStrings(bj, cr.Strings(cj))
 	case flux.TTime:
-		vs := cr.Times(cj)
-		for i := 0; i < vs.Len(); i++ {
-			if err := builder.AppendTime(bj, values.Time(vs.Value(i))); err != nil {
-				return err
-			}
-		}
-		return nil
+		return builder.AppendTimes(bj, cr.Times(cj))
 	default:
 		PanicUnknownType(c.Type)
 	}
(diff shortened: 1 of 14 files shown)


functions/transformations/difference.go
```diff
@@ -187,7 +187,7 @@ func (t *differenceTransformation) Process(id execute.DatasetID, tbl flux.Table)
 				d := differences[j]
 				switch c.Type {
 				case flux.TBool:
-					if err := arrow.AppendBools(builder, j, arrow.BoolSlice(cr.Bools(j), firstIdx, l)); err != nil {
+					if err := builder.AppendBools(j, arrow.BoolSlice(cr.Bools(j), firstIdx, l)); err != nil {
```
jsternberg: I'm not sure how this got in originally, but it's a memory leak. You have to release the slice after creating it.

functions/transformations/keys.go
```diff
@@ -167,8 +168,7 @@ func (t *keysTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 		}
 	}
 
-	// Append the keys to the column index.
-	if err := builder.AppendStrings(colIdx, keys); err != nil {
+	if err := builder.AppendStrings(colIdx, arrow.NewString(keys, nil)); err != nil {
```
jsternberg: This is a memory leak.

execute/table.go
```diff
@@ -244,54 +228,36 @@ func AppendRecord(i int, cr flux.ColReader, builder TableBuilder) error {
...
-		case flux.TTime:
-			var val Time
-			if colMap[j] >= 0 {
-				val = values.Time(cr.Times(colMap[j]).Value(i))
+		var val values.Value
+		if colMap[j] >= 0 {
+			val = ValueForRow(cr, i, colMap[j])
+			err = builder.AppendValue(j, val)
+		} else {
+			switch c.Type {
```
jsternberg: Is there a reason why `AppendNil` can't be used in this section?

functions/transformations/data_test.go
```diff
@@ -95,12 +96,12 @@ func init() {
 		}
 	}
 
-	_ = normalTableBuilder.AppendTimes(0, times)
-	_ = normalTableBuilder.AppendTimes(1, startTimes)
-	_ = normalTableBuilder.AppendTimes(2, stopTimes)
-	_ = normalTableBuilder.AppendFloats(3, normalValues)
-	_ = normalTableBuilder.AppendStrings(4, t1)
-	_ = normalTableBuilder.AppendStrings(5, t2)
+	_ = normalTableBuilder.AppendTimes(0, arrow.NewInt(times, nil))
```
jsternberg: These are also memory leaks.

affo:
@aanthony1243 @jsternberg 

Hey guys, I am not against this change as discussed in Slack. I leave the discussion to you and follow the thread.

However, @jsternberg I can't see the point of having utilities to convert a call that uses go slices to one that uses arrow arrays, given that we should evolve to an arrow-oriented implementation and avoid the use of slices. Am I missing something?   
Are you suggesting that we perform this step post-alpha and, for now, we only fix utils?

At the moment, I wrote down a fix to utils in #777   
However, from the `arrow` package in flux, I can't write, like @aanthony1243 did:

```
col := b.cols[j].(*intColumnBuilder)
nullOffset := len(col.data)
```

And use the offset of that specific column to set nils. So, I must append values one by one.  
Should we move those utils into `execute`?

aanthony1243:
I see the issues and I have some ideas.  we can talk offline and then update this PR.

affo:
For difference memory leaks, see #792

jsternberg:
I talked with @aanthony1243 offline and both of us seemed to be in agreement that the current table builder API needed to be revisited and we would revisit it after the alpha.

There were a  few points that are important:

1. The arrow builder should be exposed directly.
2. The utility methods are a bit of an anti-pattern since nobody knows they exist.
3. The table builder being an interface is likely an anti-pattern.

I'll give this another once over but let's just merge it so we have null support for the alpha and we'll revisit it later. We're not freezing any Go APIs yet anyway.

execute/table.go
aanthony1243: nope. I will fix it.

functions/transformations/difference.go
aanthony1243: will let this get fixed in a currently open PR for difference.

PR #792, opened 2019-01-10
refactor(functions): null support for difference

### LINKED ISSUE #676: Update difference to understand Null

affo:
Fixes #676 and #681 

There is quite a breaking change in the difference behavior which is specified in SPEC.

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


docs/SPEC.md
```diff
@@ -2571,23 +2571,80 @@ from(bucket: "telegraf/autogen")
...
 #### Difference
 
-Difference computes the difference between subsequent non null records.
+Difference computes the difference between subsequent records.  
+Every user-specified column of numeric type is subtracted while others are kept intact.
 
 Difference has the following properties:
 
 * `nonNegative` bool
     nonNegative indicates if the derivative is allowed to be negative.
```
wolffcm: *nit* This was here already but it should say `...if the difference is allowed to be...`

docs/SPEC.md
```diff
@@ -2571,23 +2571,80 @@ from(bucket: "telegraf/autogen")
...
+
+ - If some value is the first ever seen, then there is nothing to subtract to it, so the result of the difference is null;
+ - the difference between two non-null values is their algebraic difference; or null, if the result is negative and `nonNegative: true`;
+ - null minus some value is always null;
+ - some value `v` minus null is `v` minus the last non-null value seen before `v`; or null if `v` is the first non-null value seen.
+
+Example of difference:
+
+    |   A  |   B  |   C  |
+    |:----:|:----:|:----:|
```
wolffcm: This table doesn't seem to render like the others in this document.

affo: yep, you are right. Fixed

docs/SPEC.md
affo: Sorry, I can't get what you are saying 😅

wolffcm: There was a copy/paste error, it says "derivative" but it should say "difference".

affo: Oh! Even reading it multiple times I couldn't see it! Thank you

affo:
Rebased. Now incorporates changes to table builder.

This fixes also the memory leak.

PR #796, opened 2019-01-10
refactor(functions): support null values in distinct

### LINKED ISSUE #675: Update distinct to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #675.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2608,9 +2608,10 @@ from(bucket: "telegraf/autogen")
     |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_user")
     |> difference()
 ```
+
 #### Distinct
 
-Distinct produces the unique values for a given column.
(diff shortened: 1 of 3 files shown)


docs/SPEC.md
```diff
@@ -2610,7 +2610,7 @@ from(bucket: "telegraf/autogen")
 ```
 #### Distinct
 
-Distinct produces the unique values for a given column.
+Distinct produces the unique values for a given column. Null values in the given column are dropped.
```
wolffcm: This differs from SQL behavior which will put a null in the output if there was one or more nulls in the column.  I thought that @nathanielc said in Slack thread that we should keep a single null...

jsternberg: I didn't see his comment. I'll fix that.

jsternberg: Fixed.

PR #804, opened 2019-01-10
feat(stdlib): organizes builtin code into Flux packages

nathanielc:
BREAKING CHANGE: This change moves and renames  Go packages to
mirror the Flux package structure.
Several bugs were also exposed during this change:
The function MonoType didn't preserve the pipe argument when being converted
to a poly type.
The ObjectKind MonoType method always returned false even when a valid
MonoType was constructed.
Package names where being set erroneously preventing then from being
imported.
Not all builtins have been moved into a namespace. Future PRs will
complete the migration.

Summary of code org changes:

* functions -> stdlib
* Instead of organizing the functions/stdlib by their type (inputs/outputs/transformations) they are organized based on their Flux package.

Notes on the prelude:

 The prelude is defined to have two packages currently, `universe` and `influxdata/influxdb`. The effect of the prelude is that all identifiers from those two packages are imported into the toplevel scope. This means that everything can be defined in a package, but we can pick what packages we want to "auto import" so that users do not have to qualify everything.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


options/now.go
```diff
@@ -2,9 +2,9 @@ package options
 
 import (
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/functions/transformations"
+	"github.com/influxdata/flux/stdlib/universe"
 )
 
 func init() {
-	flux.RegisterBuiltInOption("now", transformations.SystemTime())
+	flux.RegisterBuiltInOption("now", universe.SystemTime())
```
jlapacik: Perhaps this belongs in universe?

stdlib/csv/from_test.go
```diff
@@ -1,52 +1,53 @@
...
 		{
 			Name:    "from",
-			Raw:     `from(csv:"telegraf", chicken:"what is this?")`,
+			Raw:     `import "csv" csv.from(csv:"telegraf", chicken:"what is this?")`,
 			WantErr: true,
 		},
 		{
 			Name: "fromCSV text",
-			Raw:  `fromCSV(csv: "1,2") |> range(start:-4h, stop:-2h) |> sum()`,
+			Raw:  `import "csv" csv.from(csv: "1,2") |> range(start:-4h, stop:-2h) |> sum()`,
```
jlapacik: These might read better if the import statement was on its own line.

compile.go
```diff
@@ -263,17 +267,17 @@ func RegisterPackage(pkg *ast.Package) {
...
+func RegisterPackageValue(pkgpath, name string, value values.Value) {
 	if finalized {
 		panic(errors.New("already finalized, cannot register builtin package value"))
 	}
-	packg, ok := stdlib.pkgs[path]
+	packg, ok := stdlib.pkgs[pkgpath]
 	if !ok {
-		packg = interpreter.NewPackage(name)
-		stdlib.pkgs[path] = packg
+		packg = interpreter.NewPackage(path.Base(pkgpath))
```
jlapacik: This could possibly cause issues if we introduce sub-packages, but this works for right now and is likely changing soon once we have the `builtin` syntax.

nathanielc: A package's name is not path, but rather the value of the `package <ident>` statement. The packages are keyed in the map by their path which is unique. The package is name is used as the default for the identifier when importing the package. 

We already have subpackages and this is working.

nathanielc:
Thanks. I agree there are a few things that all need to be moved into a package:

* Options
* System.time
* All the existing pure Flux functions like top/bottom or cov etc.

With this PR the major parts are moved so those changes can come later and be smaller changes.

compile.go
jlapacik: My mistake. You're right.

PR #806, opened 2019-01-10
refactor(functions): make range support nulls

### LINKED ISSUE #658: Update range to understand Null

### LINKED ISSUE #794: Flux implementation of range creates invalid group key
The current implementation of `range` in Flux will write values into group columns that do not match the group key of the table created by the builder.

This is only hit when using a source that is not InfluxDB, like `fromCSV`.

wolffcm:
Add support to range() for nulls in the time, start, stop or value columns.
Also fix a bug in the Flux implementation of range where group key would
be set but different values would be written to result table.

Fixes #658.
Fixes #794.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/range.go
```diff
@@ -248,16 +260,13 @@ func (t *rangeTransformation) Process(id execute.DatasetID, tbl flux.Table) erro
 		// If [start, stop) (where start <= stop) from the group key is contained in the
 		// range transformation bounds [keyStart, keyStop], we can skip the whole table.
 		// Still want to skip if start >= keyStart and t.bounds.Stop == keyStop
-		forwardTable = t.bounds.Contains(keyStart) && (t.bounds.Contains(keyStop) || t.bounds.Stop == keyStop)
-	}
-
-	if forwardTable {
-		return execute.AppendTable(tbl, builder)
+		if t.bounds.Contains(keyStart) && (t.bounds.Contains(keyStop) || t.bounds.Stop == keyStop) {
```
jsternberg: I'm curious. What happens if `_start` and `_stop` are lying and there is a `_time` that isn't in the range? Or there is a null time (which should be discarded) inside of a range that is otherwise valid?

wolffcm: In that case, the lie would be propagated.  The alternative is to never forward tables... I can see reasons for doing it either way.  What do you think?

wolffcm: So, I realized a couple things:
1. transformations can only iterate over the input tables one time.  So the logic to check if table forwarding would be allowed would need to be put into the call to `tbl.Do` at the end of the `Process` method.
2. `execute.AppendTable` just calls `tbl.Do` anyway, so there's not a whole lot of performance benefit to doing the forwarding... it's not as if O(n) becomes O(1).

As such I just removed the table-forwarding logic.

wolffcm: And there is an E2E test failing after rebasing on master, I need to figure that out.

jsternberg: This is something worth exploring later. We should be capable of table forwarding although I do believe you are correct about its current state so we'll revisit forwarding when it becomes  practical.

wolffcm:
@jsternberg Does this look okay now?  I had to update an E2E test because `keys` now reflects that `_start` and `_stop` are in the group key (`range` wasn't adding them before).  I also made it so if we are adding the start/stop columns to the group key, then we put them first, similar to how storage does here:
https://github.com/influxdata/influxdb/blob/master/storage/reads/reader.go#L439

PR #815, opened 2019-01-11
refactor(functions): add null support to derivative

### LINKED ISSUE #631: Implement Nil support for Derivative

wolffcm:
Fixes #631.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2680,14 +2680,15 @@ from(bucket: "telegraf/autogen")
 
 #### Derivative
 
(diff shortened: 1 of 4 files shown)


wolffcm:
@jlapacik Storage (and go Arrow too) use templating to help avoid repeating code:
https://github.com/influxdata/influxdb/blob/master/tsdb/cursors/arrayvalues.gen.go
We don't do anything like that in Flux yet, but maybe we should.

PR #818, opened 2019-01-11
fill function to replace null values in a column

aanthony1243:
the end to end test is a WIP
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


stdlib/tests/testdata/fill.flux
```diff
@@ -0,0 +1,187 @@
...
+  |> filter(fn: (r) => r._field == field)
+  |> yield(name:"inbool")
+  |> fill(column: "_value", value: value)
+
+t_fill_bool = (table=<-) => t_fill(table:table, field:"boolF", value:false)
+
+testingTest(name: "fill",
+            input: testLoadStorage(csv: inData),
+            want: testLoadMem(csv: outData)|>filter(fn:(r) => r._field == "boolF")|>yield(name:"outBool"),
+            test: t_fill_bool)
```
affo: Why that filtering?

stdlib/universe/fill.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, errors.New("unknown type in fill op-spec")
+	}
+
+	return pspec, nil
+}
+
+func (s *FillProcedureSpec) Kind() plan.ProcedureKind {
+	return FillKind
+}
+func (s *FillProcedureSpec) Copy() plan.ProcedureSpec {
```
affo: It seems you are actually not copying fields.

stdlib/universe/fill.go
```diff
@@ -0,0 +1,251 @@
...
+		spec:  spec,
+	}
+}
+
+func (t *fillTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey) error {
+	return t.d.RetractTable(key)
+}
+
+func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	builder, created := t.cache.TableBuilder(tbl.Key())
```
affo: Wouldn't it be better to check as soon as possible type consistency for the column type and the value type, instead of waiting for appending a value of the wrong type?

affo:
@aanthony1243 Looks great!

Question: what if I fill some value in the group key? Does is automatically triggers a re-grouping?

stdlib/universe/fill_test.go
```diff
@@ -0,0 +1,364 @@
...
+
+	"github.com/influxdata/flux/plan"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/querytest"
+	"github.com/influxdata/flux/stdlib/universe"
+)
+
```
affo: Aren't you testing NewQuery?

stdlib/tests/testdata/fill.flux
aanthony1243: it's  WIP.  will remove it now.

affo:
Great!

What if you are using the previous value on a group key column and causing a regrouping?  
Should you use different builders in your code and not rely on the value passed by the user.

Another question: why is `value` required if there is `usePrevious` shouldn't they be mutually exclusive? We should need exactly one of the two. Do . you agree?

stdlib/universe/fill.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (s *FillProcedureSpec) Kind() plan.ProcedureKind {
+	return FillKind
+}
+func (s *FillProcedureSpec) Copy() plan.ProcedureSpec {
+	ns := new(FillProcedureSpec)
+
+	*ns = *s
+	ns.Value = s.Value
```
affo: This is my ignorance 😅 
You don't need to copy primitiva values because they are copied at the line before when assigning the content of objects?

aanthony1243: yeah, and probably the value is OK too since they are only primitives but I though to just be safe.

jsternberg: I agree this copy isn't needed. `values.Value` is supposed to be an immutable interface so, in the case that the pointer gets copied, it should still be fine since the value itself is immutable.

stdlib/universe/fill_test.go
```diff
@@ -0,0 +1,423 @@
+package universe_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/influxdata/flux/values"
```
jsternberg: GoLand is bad at imports.

stdlib/tests/testdata/fill_bool.flux
```diff
@@ -0,0 +1,47 @@
...
+,,1,2018-12-19T22:13:30Z,2018-12-19T22:14:20Z,m1,f1,server02,2018-12-19T22:14:00Z,true
+,,1,2018-12-19T22:13:30Z,2018-12-19T22:14:20Z,m1,f1,server02,2018-12-19T22:14:10Z,true
+,,1,2018-12-19T22:13:30Z,2018-12-19T22:14:20Z,m1,f1,server02,2018-12-19T22:14:20Z,false
+"
+
+option now = () => 2018-12-19T22:15:00Z
+
+t_fill_bool = (table=<-) => table
+  |> range(start: -5m)
+  |> fill(column: "_value", value: false)
```
jsternberg: Can you remove the `column` argument here too?

PR #820, opened 2019-01-11
feat(ast): implement and require builtin statements

nathanielc:
BREAKING CHANGE: The option string is now a keyword.
In addition to adding and implementing the builtin keyword, option was
changed to be a keyword.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

This change partially implements the BuiltinStatment specification. The type expressions are not implemented.


compile_internal_test.go
```diff
@@ -0,0 +1,97 @@
...
+			switch {
+			case err == nil && tc.err == nil:
+				// Test passes
+			case err == nil && tc.err != nil:
+				t.Errorf("expected error %v", tc.err)
+			case err != nil && tc.err == nil:
+				t.Errorf("unexpected error %v", err)
+			case err != nil && tc.err != nil:
+				if err.Error() != tc.err.Error() {
+					t.Errorf("differring error messages -want/+got:\n%s", cmp.Diff(tc.err.Error(), err.Error()))
```
jlapacik: differing.

PR #836, opened 2019-01-14
refactor(functions): add null support to stddev

### LINKED ISSUE #649: Update stddev to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #649.

### DIFF
--- a/stdlib/universe/stddev.go
+++ b/stdlib/universe/stddev.go
@@ -100,7 +100,11 @@ func (a *StddevAgg) NewStringAgg() execute.DoStringAgg {
 }
 func (a *StddevAgg) DoInt(vs *array.Int64) {
 	var delta, delta2 float64
-	for _, v := range vs.Int64Values() {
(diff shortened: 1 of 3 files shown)


stdlib/universe/stddev.go
```diff
@@ -67,7 +67,8 @@ func (s *StddevProcedureSpec) Copy() plan.ProcedureSpec {
 }
 
 type StddevAgg struct {
-	n, m2, mean float64
+	n        int64
```
nathanielc: Why was `n` changed to an `int64` I do not see any uses of `n` as an `int64`? It seems to be being converted to a `float64` on each use.

jsternberg: I thought it was better to help maintain accuracy during comparisons. The two are functionally  the same so I figured keeping it as the proper type (an integer for counting) would reduce potential errors.

jsternberg: I have reverted it back.

PR #840, opened 2019-01-14
test(stdlib): added a test confirming that sample works with null values

### LINKED ISSUE #657: Update sample to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #657.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1778,7 +1778,7 @@ The following properties define how the sample is selected.
 
 | Name | Type | Description                                                                                                                                                                  |
 | ---- | ---- | -----------                                                                                                                                                                  |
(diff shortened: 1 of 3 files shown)


jsternberg:
It looks like a mistake in the validation. There's no requirement in the code for position to be less than N (it will still work), but it does look like that was the intention. I updated the spec creation to check for that argument.

It seems we do most of those validations there, but that won't validate when a spec itself is input. We should revisit moving those validations to the creation of the procedure spec later.

affo:
Yeah, for validation I agree with you.

For this function, wouldn't it be better to leave the code as your previous version and remove the `pos < n` requirement? Should we discuss in Slack?

jsternberg:
#847 for the follow up item.

PR #842, opened 2019-01-15
test(universe): validate columns null support

### LINKED ISSUE #821: Update columns to understand Null
Columns function will be available once #640 is merged.

affo:
Fixes #821 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


affo:
@nathanielc Nope, because the function works only on the column labels that can't be null, and thank to including arrow arrays and null support in the TableBuilder, many PRs for null support became mainly test addition.

PR #844, opened 2019-01-15
refactor(universe): add null support to unique

### LINKED ISSUE #646: Update unique to understand Null

affo:
Fixes #646 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/unique.go
+++ b/stdlib/universe/unique.go
@@ -134,6 +134,7 @@ func (t *uniqueTransformation) Process(id execute.DatasetID, tbl flux.Table) err
 		floatUnique  map[float64]bool
 		stringUnique map[string]bool
 		timeUnique   map[execute.Time]bool
+		nullUnique   bool
 	)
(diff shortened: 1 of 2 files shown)


stdlib/universe/unique.go
```diff
@@ -156,41 +157,83 @@ func (t *uniqueTransformation) Process(id execute.DatasetID, tbl flux.Table) err
...
+						continue
+					}
+					nullUnique = true
+				} else {
+					v := vs.Value(i)
+					if boolUnique[v] {
+						continue
+					}
+					boolUnique[v] = true
 				}
```
affo: It was not necessary to check for null for the tests to work, but I thought we shouldn't rely on default values for null elements.

jsternberg: That is correct.

jsternberg: That is correct.

PR #848, opened 2019-01-15
refactor(stdlib): add null support to sort

### LINKED ISSUE #652: Update sort to understand Null

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

When sorting, null values are always first so they are less than in
ascending mode and greater than in descending mode. This is to match
with the behavior of SQL.

Fixes #652.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2163,15 +2163,17 @@ from(bucket: "telegraf/autogen") |> set(key: "mykey", value: "myvalue")
 Sorts orders the records within each table.
 One output table is produced for each input table.
 The output tables will have the same schema as their corresponding input tables.
+When sorting, nulls will always be first. When `desc: false` is set, then nulls are less than every other value. When `desc: true`, nulls are greater than every value.
 
 Sort has the following properties:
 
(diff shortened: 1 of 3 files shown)


docs/SPEC.md
```diff
@@ -2172,13 +2173,15 @@ Sort has the following properties:
 | desc    | bool     | Desc indicates results should be sorted in descending order.                              |
```
affo: Can you put the default value for `desc`?

jsternberg: Done.

execute/table.go
```diff
@@ -1684,9 +1731,22 @@ func (c *uintColumnBuilder) Len() int {
 	return len(c.data)
 }
 func (c *uintColumnBuilder) Equal(i, j int) bool {
+	if inil, jnil := c.nils[i], c.nils[j]; inil || jnil {
+		return inil == jnil
+	}
```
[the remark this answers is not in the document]
jsternberg: I refactored all of the column builder code to use a mixin when appropriate.

PR #849, opened 2019-01-15
feat(examples): add example on how to use flux as a library

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,5 @@
+# Flux Examples
+
+This folder contains a list of usage examples related to the Flux repo.
+
+- [library](library): Use Flux as a standalone library in your Go programs.
--- a/examples/library/README.md
+++ b/examples/library/README.md
@@ -1,0 +1,39 @@
+# Flux as a Library Example
+
+This folder contains a [main.go](main.go) file that shows how flux can be used as a library in your programs.
+
+## Components
+
+The main components you need are:
+
+- The interpreter;
+- The scope (aka Prelude);
+- The builtin library and your additional functions, if you want to define them;
+- The language specification compiler;
+- A querier.
+
+## What does this example do?
+
+This example takes all the components described above from the Flux repo and uses them to
+compile, execute and print the results of this query.
+
+```flux
+import g "generate"
+g.from(start: 1993-02-16T00:00:00Z, stop: 1993-02-16T00:03:00Z, count: 5, fn: (n) => 1)
+```
+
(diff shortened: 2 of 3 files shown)


fntlnz:
@nathanielc @affo PTAL

examples/library/main.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		spec, err := flux.ToSpec([]values.Value{t}, nowTime.Time().Time())
+		if err != nil {
+			panic(err)
+		}
+		compiler := lang.SpecCompiler{
+			Spec: spec,
+		}
+
+		querier := NewQuerier()
```
affo: I think you can use:

```
import "github.com/influxdata/flux/cmd/flux/cmd"

cmd.NewQuerier()
```

fntlnz: thanks for the suggestion @affo - changed that

fntlnz:
good call @nathanielc - please take another look now

andig:
Great pr for an tutorial. It would be nice to add an example on how to pass data into the flux library for processing, too.

andig:
@fntlnz too bad 😢

affo:
This is a great example and we have only lost track of it.  

@fntlnz If now it requires some changes to work out, I can perform them on this branch.

@nathanielc could you please review this?

andig:
> It would be nice to add an example on how to pass data into the flux library for processing, too.

@affo would that be something you could add on top?

affo:
@andig sure.

But what do you mean with passing data?

```
import "csv"

data = "
#datatype,string,long,long,string
#group,false,false,false,true
#default,_result,,,
,result,table,value,tag
,,0,10,a
,,0,10,a
,,1,20,b
,,1,20,b
,,2,30,c
,,2,30,c
,,3,40,d
,,3,40,d
"

csv.from(csv: data) |> yield(name: "res")
```

This is a valid flux script.

Do you mean encoding a table from go by using an `Encoder`? Yep, I could do that.

Anyway, this PR is about this example. @andig if you want to see another example, can you please file an issue for that in Flux? Feel free to tag me :)

andig:
> But what do you mean with passing data?

@affo you example is already enough to reverse anything else that might be needed by looking at the `csv.from()` function. Much appreciated, thank you!

affo:
@andig the PR got merged to master 🥂 

That folder is there to be enhanced with other examples. If you want to see other examples, feel free to open issues in Flux and tag me in!

If you want to contribute with yours too, you are welcome 😎 😄

Issue #643, opened 2019-01-08
Avoid empty `Do` calls in functions


affo:
In functions `keys` and `columns` there is a call to `Do` on the input table that does nothing, just to "ensure that the iterators work properly" (see #640 and #641).

It is not clear whether this call is still necessary and, if so, in which cases.

We should investigate if it is necessary, and, if so, provide a more idiomatic way to update iterators status.

jsternberg:
We can replace the empty calls to `Do` with `Done` now that `Done` is implemented

PR #850, opened 2019-01-15
refactor(stdlib): add null support to join function

### LINKED ISSUE #668: Update join to understand Null

wolffcm:
Fixes #668.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -280,7 +280,9 @@ func (t *mergeJoinTransformation) Process(id execute.DatasetID, tbl flux.Table) 
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
-	t.cache.insertIntoBuffer(id, tbl)
+	if err := t.cache.insertIntoBuffer(id, tbl); err != nil {
+		return err
+	}
 
 	// Check if enough data sources have been seen to produce an output schema
 	if !t.cache.isBufferEmpty(t.leftID) && !t.cache.isBufferEmpty(t.rightID) && !t.cache.postJoinSchemaBuilt() {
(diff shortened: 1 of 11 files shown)


stdlib/universe/join.go
```diff
@@ -808,6 +828,23 @@ func (c *MergeJoinCache) buildPostJoinSchema() {
...
+// equalJoinKeys compares two keys for equality.
+// Null values are not considered equal when joining (unlike when grouping).
+func equalJoinkeys(left, right flux.GroupKey) bool {
+	for _, v := range left.Values() {
+		if v.IsNull() {
+			return false
+		}
+	}
+	for _, v := range right.Values() {
+		if v.IsNull() {
```
jlapacik: Is this necessary after you added the `isNull()` check to `values.Equal()`? That is, can you reduce this function to just `left.Equal(right)`?

stdlib/universe/join.go
```diff
@@ -697,7 +699,25 @@ func (c *MergeJoinCache) insertIntoBuffer(id execute.DatasetID, tbl flux.Table)
...
+
+	// Optimization: if any group key columns overlap join key columns,
+	// and there are any nulls in those columns, we can discard this table,
+	// since null != null for joining purposes.
+	k := tbl.Key()
+	for j, col := range k.Cols() {
+		if c.on[col.Label] {
+			if k.IsNull(j) {
+				// Discard the table and return.
+				err := tbl.Do(func(flux.ColReader) error {
```
jlapacik: Why is the explicit call to `tbl.Do` necessary? Can we not just return right here, not caching the table?

stdlib/universe/join.go
wolffcm: Yes, you're right.  I'll fix it.

stdlib/universe/join.go
wolffcm: Apparently we need to iterate over the table for it to be consumed.  Lorenzo filed an issue for this:
https://github.com/influxdata/flux/issues/643
I'm not quite clear on why this is exactly, though.  There are a few different implementions of `flux.Table`, I think some of them may require their contents to be iterated exactly once for the pipeline to work right.

I'll add a comment to the code.

PR #851, opened 2019-01-15
Feat/var immutability

jpacik:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

Adds semantic checks for variable reassignments and option declarations below the package block.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -143,14 +143,6 @@ func TestEval(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			name: "reassign nested scope",
-			query: `
-			s = () => 6
-			s = s()
-			`,
-			wantErr: true,
-		},
-		{
 			name: "binary expressions",
 			query: `
 			six = six()
@@ -219,16 +211,6 @@ func TestEval(t *testing.T) {
 			`,
 		},
 		{
-			name: "scope closing mutable",
-			query: `
-			x = 5
-            plusX = (r) => r + x
-            plusX(r:2) == 7 or fail()
-			x = 1
-            plusX(r:2) == 3 or fail()
-			`,
-		},
-		{
 			name: "nested scope mutations not visible outside",
 			query: `
 			x = 5
@@ -238,8 +220,6 @@ func TestEval(t *testing.T) {
             }
             xinc() == 6 or fail()
             x == 5 or fail()
-            x = 1
-            xinc() == 2 or fail()
 			`,
 		},
 		{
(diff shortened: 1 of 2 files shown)


interpreter/interpreter_test.go
```diff
@@ -218,30 +210,6 @@ func TestEval(t *testing.T) {
...
-            xinc = () => {
-                x = x + 1
-                return x
-            }
-            xinc() == 6 or fail()
-            x == 5 or fail()
-            x = 1
-            xinc() == 2 or fail()
-			`,
-		},
```
nathanielc: Instead of deleting these tests can you update them to test the opposite behavior?

semantic/check.go
```diff
@@ -0,0 +1,120 @@
...
+type varDecVisitor struct {
+	names  map[string]bool
+	varFn  func(*NativeVariableAssignment)
+	option bool
+}
+
+func (v varDecVisitor) Visit(node Node) Visitor {
+	switch n := node.(type) {
+	case *OptionStatement:
+		v.option = true
```
nathanielc: Could you just return `nil` here so that you do not walk down the option statement children? That would remove the need to track whether you were in a option statement or not.

semantic/check_test.go
```diff
@@ -0,0 +1,1161 @@
...
+		},
+		{
+			// package foo
+			// f = (a) => {
+			//   a = 1
+			//   return a
+			// }
+			//
+			name: "reassign parameter",
+			skip: "reassigning a function parameter inside a function block is a variable redeclaration and should error",
```
nathanielc: Why is this one skipped?

jlapacik: I was just creating an issue for this. Currently the function body is in a separate block nested inside of the block where the function parameters are assigned. This means you can reassign (in this case it would be shadowing) a function's parameter inside its body. Technically this is not correct; the parameters should be declared in the same block as the body and so should not be able to be reassigned.

semantic/check.go
jlapacik: You still have to walk the children unfortunately. I created a test case highlighting this exact issue. Basically you could declare an option as a function value where the function body has a variable redeclaration. If you returned `nil` here, this case would pass the check.

jlapacik: https://github.com/influxdata/flux/blob/df5ed3aae2276425f763a18f56daae94b4bdc65d/semantic/check_test.go#L685

nathanielc: Great

semantic/check_test.go
nathanielc: OK, so the fix will likely be to update the ScopedVisitor to not create a new scope for parameters vs the body. Sounds good, makes sense as a separate issue.

PR #852, opened 2019-01-15
fix(repl): clear return for each repl command

### LINKED ISSUE #833: REPL displays previous result for import statements
Sets to reproduce:

```
> import "csv"
> csv
pkg{from}
> import "http"
pkg{from}
> http
pkg{to}
```

My guess is that the return value on the scope is not getting cleared so its being printed again.

nathanielc:
Fixes #833

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -153,6 +153,8 @@ func (r *REPL) executeLine(t string) (values.Value, error) {
 		return nil, err
 	}
 
+	r.scope.SetReturn(nil)
+
 	if _, err := r.interpreter.Eval(semPkg, r.scope, flux.StdLib()); err != nil {
 		return nil, err
 	}


repl/repl.go
```diff
@@ -153,6 +153,8 @@ func (r *REPL) executeLine(t string) (values.Value, error) {
 		return nil, err
 	}
 
+	r.scope.SetReturn(nil)
```
jlapacik: Correct me if I'm wrong, but is this necessary? Why not just use the value returned by `interpreter.Eval` and not mess with the scope? `interpreter.Eval` will return any produced side effects. This means TableObjects as well as any value resulting from any expression statements.

nathanielc: I had a similar thought, but their could be many returned side effects and there is no consistent way to know which side effect is the result of the expression. So for now I just made the change I knew would fix the issue. 

I'll create a new issue to address using side effects instead of the scope return value for the REPL.

jlapacik: Ok sounds good to me.

nathanielc: https://github.com/influxdata/flux/issues/867

PR #853, opened 2019-01-15
docs(SPEC): fix examples with group(by)

mark-rushakoff:
Fix the outdated example to use group(columns) instead.


jsternberg:
The PR has nothing to do with the error and can't cause any errors so I'm going to merge this even with the failure.

PR #854, opened 2019-01-15
fix: skip the testdata folder for builtin generate

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The testdata folder is meant to contain testdata and is not a valid
package. This modifies generate to skip that folder.

This also regenerates the scanner. For whatever reason, ragel generated
a different file. This will require further investigation.


jsternberg:
This is incorrect. After a discussion with @jlapacik, #856 is the correct solution.

PR #860, opened 2019-01-15
feat(plan): add rule to remove filter true nodes

nathanielc:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/filter.go
+++ b/stdlib/universe/filter.go
@@ -35,6 +35,9 @@ func init() {
 	flux.RegisterOpSpec(FilterKind, newFilterOp)
 	plan.RegisterProcedureSpec(FilterKind, newFilterProcedure, FilterKind)
 	execute.RegisterTransformation(FilterKind, createFilterTransformation)
+	plan.RegisterPhysicalRules(
+		RemoveTrivialFilterRule{},
+	)
 }
 
 func createFilterOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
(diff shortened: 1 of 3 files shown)


stdlib/universe/filter.go
```diff
@@ -170,3 +173,32 @@ func (t *filterTransformation) UpdateProcessingTime(id execute.DatasetID, pt exe
...
+func (MergeFilterAnyRule) Name() string {
+	return "MergeFilterAnyRule"
+}
+
+func (MergeFilterAnyRule) Pattern() plan.Pattern {
+	return plan.Pat(FilterKind, plan.Any())
+}
+
+func (MergeFilterAnyRule) Rewrite(filterNode plan.PlanNode) (plan.PlanNode, bool, error) {
+	if len(filterNode.Predecessors()) != 1 {
```
wolffcm: You have just one predecessor pattern (`plan.Any()`), so if there were not exactly one predecessor the pattern would not match.  So this check is already enforced by the pattern.

stdlib/universe/filter_test.go
```diff
@@ -580,6 +582,143 @@ func TestFilterOperation_Marshaling(t *testing.T) {
...
+			// from -> filter => from -> filter
+			Rules: []plan.Rule{universe.MergeFilterAnyRule{}},
+			Before: &plantest.PlanSpec{
+				Nodes: []plan.PlanNode{
+					plan.CreatePhysicalNode("from", from),
+					plan.CreatePhysicalNode("filter", filterOther),
+				},
+				Edges: [][2]int{{0, 1}},
+			},
+			After: &plantest.PlanSpec{
```
wolffcm: These tests can be made a bit more concise by using `RuleTestCase.NoChange` to indicate that no rewrite should be done.

stdlib/universe/filter.go
nathanielc: Great I wasn't sure about that so I figured I'd be safe, but I'll update this and the docs.

stdlib/universe/filter_test.go
nathanielc: great

stdlib/universe/filter.go
nathanielc: After a reread of the docs its clear now. I was thinking that more parameters to the Pat function went further up the tree but now I see that it just means more predecessors period. The docs are clear I just had to read them twice.

stdlib/universe/filter.go
```diff
@@ -170,3 +173,29 @@ func (t *filterTransformation) UpdateProcessingTime(id execute.DatasetID, pt exe
 func (t *filterTransformation) Finish(id execute.DatasetID, err error) {
 	t.d.Finish(err)
 }
+
+// MergeFilterAnyRule removes Filter nodes whose predicate always evaluates to true.
+type MergeFilterAnyRule struct{}
```
jlapacik: Do you mind renaming this rule to something like `RemoveConstantFilterRule` or `RemoveTrivialFilterRule`? I think something like that would be more descriptive.

PR #862, opened 2019-01-16
fix(execute): skip lambda evaluation when referencing nulls

affo:
`rowFn` evaluation returns an error when referencing a null value. This prevents a panic when a lambda references a null value and logs the error instead.

This PR also adds testing for `rowFn`.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/row_fn.go
+++ b/execute/row_fn.go
@@ -114,11 +114,56 @@ func ConvertFromKind(k semantic.Nature) flux.ColType {
 }
 
 func (f *rowFn) eval(row int, cr flux.ColReader) (values.Value, error) {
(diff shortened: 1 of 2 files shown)


affo:
@nathanielc I have doubts on the testing part. I don't know if I could achieve a better result with the correct options to `cmp`.

jsternberg:
Related to #941. It might be best to fix this instead of trying to avoid null values in evaluation.

affo:
Yep, I agree with you. This didn't get merged in the alpha, I hoped so, but we couldn't make it.

affo:
Great 👍

PR #864, opened 2019-01-16
refactor(execute): change sort order of group keys to match InfluxDB

e-dard:
This PR changes the sort order of group keys to match the order of series keys emitted from InfluxDB.

In InfluxDB series keys are sorted in lexicographic order. Currently in flux however grouping keys are sorted first by the number of columns and then only lexicographically if both keys have the same number of columns.

For example, InfluxDB stores the following series keys in this order:

```
_f=temp,_m=cpu,region=east,server=a
_f=temp,_m=cpu,region=east,server=a
_f=value,_m=cpu,region=west,server=a,yaba=dabadoo
_f=value,_m=cpu,region=west,server=b
```

However, Because flux currently considers the number of columns in keys before looking at those column names, it will re-sort these keys to the following:

```
_f=temp,_m=cpu,region=east,server=a
_f=temp,_m=cpu,region=east,server=a
_f=value,_m=cpu,region=west,server=b
_f=value,_m=cpu,region=west,server=a,yaba=dabadoo
```

This can lead to significant performance overhead when dealing with many series keys.

In https://github.com/influxdata/influxdb/issues/10988 @russorat  has been seeing some really poor performance when running "meta query" style queries (essentially just getting series keys and identifying the unique set of measurement or tag keys.

For example. given a bucket cardinality of only `~5,000` the following query takes **~40s** to execute:

```
from(bucket: "default")  |> 
range(start: -30d)  |> 
keys()  |> group()  |> distinct()  |> 
keep(columns: ["_value"])  |> sort()  |> limit(n: 200)
```

This profile identifies significant amounts of sorting as the problem: [profile002.pdf](https://github.com/influxdata/flux/files/2764604/profile002.pdf)

With these proposed changes, the time to execute the above query drops from `~40s` to `~1.9s`, which is an improvement of about `21X`.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/group_key.go
+++ b/execute/group_key.go
@@ -132,13 +132,18 @@ func groupKeyEqual(a, b flux.GroupKey) bool {
 	return true
 }
 
+// groupKeyLess determines if the former key is lexicographically less than the
+// latter.
 func groupKeyLess(a, b flux.GroupKey) bool {
 	aCols := a.Cols()
 	bCols := b.Cols()
-	if av, bv := len(aCols), len(bCols); av != bv {
-		return av < bv
+	min := len(aCols)
+
+	if min > len(bCols) {
+		min = len(bCols)
 	}
-	for j, c := range aCols {
+
+	for j := 0; j < min; j++ {
 		if av, bv := aCols[j].Label, bCols[j].Label; av != bv {
 			return av < bv
 		}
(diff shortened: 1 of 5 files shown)


wolffcm:
The E2E test failing on this branch is `show_all_tag_keys`?  I noticed that it was also failing on my branch after rebasing on master.  Maybe it's flaky somehow?

nathanielc:
I just pushed a commit to this branch that fixes the `show_all_tag_keys` test.

nathanielc:
@wolffcm I just ran the test case 1000 times on my machine an it passed every time. I also ran it with the race detector FWIW

wolffcm:
@nathanielc Yeah, seems like it's not the test, it's something on my branch.  Thanks, good to know that the test is deterministic at least.

PR #871, opened 2019-01-16
feat(stdlib): adds various v1 meta queries helper functions

nathanielc:
### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/v1/v1.flux
+++ b/stdlib/influxdata/influxdb/v1/v1.flux
@@ -1,5 +1,40 @@
 package v1
 
+// Json parses an InfluxDB 1.x json result into a table stream.
 builtin json
 
+// Databases returns the list of available databases, it has no parameters.
 builtin databases
+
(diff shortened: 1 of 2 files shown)


stdlib/influxdata/influxdb/v1/v1.flux
```diff
@@ -1,5 +1,40 @@
 package v1
 
+// Json parses an InfluxDB 1.x json result into a table stream.
 builtin json
 
+// Databases returns the list of available databases, it has no parameters.
 builtin databases
+
+// TagValues returns the unique values for a given tag.
```
jsternberg: This seems like godoc, but the function names are `camelCase` and not `CamelCase`.

nathanielc: Yeah, starting a sentence with a lower case felt wrong so I emulated Go doc but used a capital letter. We we build tooling to automate working with docs it will just need to understand how to map between them.

Issue #321, opened 2018-11-22
Cursor type error when grouping causes schema collision


affo:
I am querying an InfluxDB nightly installation using the sandbox and chronograph. The bucket is `telegraf` and the measurement is `syslog` (populated using telegraf's plugin).

The schema is:

```
_start dateTime:RFC3339
_stop dateTime:RFC3339
_time dateTime:RFC3339
_value long
_field string
_measurement string
appname string
facility string
host string
hostname string
severity string
```

This is the base query:

```
from(bucket: "telegraf")
    |> filter(fn: (r) => r._measurement == "syslog")
    |> range(start: <start>)
```

By grouping by some tag (e.g. `severity` or `facility`) and tuning the `start` of the range I am able to produce this error:

```
{"error":"csv encoder error: expected integer cursor type, got *reads.stringMultiShardArrayCursor"}
```

Even if the query without `group` runs fine.

affo:
Again, with query:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "syslog")
  |> range(start: -3d)
  |> group(by: ["facility"])
  //|> distinct(column: "facility")
  |> keyValues(keyColumns: ["facility"])
```

got

```
expected integer cursor type, got *reads.stringMultiShardArrayCursor
```

The query with `distinct` works. Ideas?

aanthony1243:
because you are not filtering by a `_field`, you will group different fields together into the same table by "facility".  A column must have a single type in the same table.  

This is another 'usability' issue.

affo:
O ok, another error message that should be improved

nathanielc:
This should be solved with #1411

russorat:
closing in favor of https://github.com/influxdata/flux/issues/1411

doivosevic:
The error message is still there and very confusing when trying to do a simple thing such as grouping in grafana. Could the error message be improved?

PR #878, opened 2019-01-17
fix(execute): check for schema collision when appending columns to a table

affo:
Must be merged after #877, the second commit is the relevant one.

Fixes the Flux side of #321 by checking column types when columns are added to an existing table and returning a meaningful error if they have the same label and different type.

Still, the InfluxDB side of #321 has to be fixed.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


affo:
Great, I'll merge as soon as Chris reviews #877

PR #880, opened 2019-01-17
feat(control): add query function count metrics

nathanielc:
Closes https://github.com/influxdata/influxdb/issues/11180

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
This differs slightly from the issue design.

1. It does use a `lang` label since in the current design the concept of language is already abstracted away by the time we get to a query spec to execute. Technically we are counting query spec operations kinds, not functions.
2. The namespace of `query` is used in addition to the subsystem `control` since that is where all metrics are currently collected and managed.

PR #886, opened 2019-01-17
feat(parser): track nested blocks in the parser

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The parser now tracks nested blocks so when it sees something like:

    a[b(]

It will now be able to identify that the `]` closes the index expression
and recognize that the parenthesis was never closed. Previously, it
would aggressively search for the end parenthesis and would exhaust the
input. This would cause any part of the AST after the unbalanced
parenthesis to be completely useless even though we can reasonably guess
that `]` isn't a valid token for an expression and the `]` is obviously
closing the index expression.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -36,7 +36,8 @@ func ParseFile(f *token.File, src []byte) *ast.File {
 		s: &scannerSkipComments{
 			Scanner: scanner.New(f, src),
 		},
-		src: src,
+		src:    src,
+		blocks: make(map[token.Token]int),
 	}
 	return p.parseFile(f.Name())
 }
@@ -75,6 +76,10 @@ type parser struct {
 	lit      string
 	buffered bool
 	errs     []ast.Error
+
+	// blocks maintains a count of the end tokens for nested blocks
+	// that we have entered.
+	blocks map[token.Token]int
 }
 
 func (p *parser) parseFile(fname string) *ast.File {
@@ -96,7 +101,7 @@ func (p *parser) parseFile(fname string) *ast.File {
 	if len(file.Imports) > 0 {
 		file.Loc.End = locEnd(file.Imports[len(file.Imports)-1])
 	}
-	file.Body = p.parseStatementList(token.EOF)
+	file.Body = p.parseStatementList()
 	if len(file.Body) > 0 {
 		file.Loc.End = locEnd(file.Body[len(file.Body)-1])
 	}
@@ -145,10 +150,10 @@ func (p *parser) parseImportDeclaration() *ast.ImportDeclaration {
 	}
 }
 
-func (p *parser) parseStatementList(eof token.Token) []ast.Statement {
+func (p *parser) parseStatementList() []ast.Statement {
 	var stmts []ast.Statement
 	for {
-		if _, tok, _ := p.peek(); tok == eof || tok == token.EOF {
+		if ok := p.more(); !ok {
 			return stmts
 		}
 		stmts = append(stmts, p.parseStatement())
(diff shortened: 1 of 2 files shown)


jsternberg:
I'm currently fuzzing this, but I think it's ready for review if fuzzing doesn't produce any errors. I did some fuzzing earlier and the only crashes that popped up were unrelated to this specifically.

internal/parser/parser.go
```diff
@@ -1049,6 +1055,79 @@ func (p *parser) repeat(fn func() bool) {
...
+// more will check if we should continue reading tokens for the
+// current block. This is true when the next token is not EOF and
+// the next token is also not one that would close a block.
+func (p *parser) more() bool {
+	_, tok, _ := p.peek()
+	if tok == token.EOF {
+		return false
+	}
+	_, ok := p.stack[tok]
+	return !ok
```
nathanielc: nit, but it would be clearer to me of this used the count and the zero value for if the token is not in the map (i.e. Go's default behavior)

```
count := p.stack[tok]
return count > 0
```

internal/parser/parser.go
```diff
@@ -1049,6 +1055,79 @@ func (p *parser) repeat(fn func() bool) {
...
+	// checked for this.
+	count, ok := p.stack[end]
+	if !ok {
+		panic("closing a block that was never opened")
+	}
+
+	// Decrement the stack whether or not the next token is the
+	// end token.
+	if count == 1 {
+		delete(p.stack, end)
```
nathanielc: With my comment above you could leave this token in the map and remove this branching logic. I.e always decrement.

internal/parser/parser.go
```diff
@@ -1049,6 +1055,79 @@ func (p *parser) repeat(fn func() bool) {
...
+	if end == token.EOF {
+		// TODO(jsternberg): Check for EOF and panic if it isn't.
+		pos, _, lit := p.scan()
+		return pos, lit
+	}
+
+	// The end token must be in the stack. We should have already
+	// checked for this.
+	count, ok := p.stack[end]
+	if !ok {
```
nathanielc: This sanity check should probably also check that `count > 0`.

internal/parser/parser_test.go
```diff
@@ -3686,6 +3705,78 @@ join(tables:[a,b], on:["t1"], fn: (a,b) => (a["_field"] - b["_field"]) / b["_fie
 				},
 			},
 		},
+		// {
+		// 	name: "unbalanced braces piped into a function",
+		// 	raw:  `map(fn: (r) => { return r._value ) |> count()`,
```
nathanielc: Does this one work?

jsternberg: Yes. I realized I didn't need that test. I accidentally left it in.

jsternberg:
Fuzzing after 25 minutes resulted in no crashes.

I'll add test cases for the ones you specified, but they should work properly. If you close something that hasn't been opened, it shouldn't be present in the stack so it won't be considered an ending token when consulted by `p.more()`.

internal/parser/parser.go
```diff
@@ -1049,6 +1055,70 @@ func (p *parser) repeat(fn func() bool) {
...
+
+// more will check if we should continue reading tokens for the
+// current block. This is true when the next token is not EOF and
+// the next token is also not one that would close a block.
+func (p *parser) more() bool {
+	_, tok, _ := p.peek()
+	if tok == token.EOF {
+		return false
+	}
+	return p.blocks[tok] == 0
```
nathanielc: Should this be `> 0`?

jsternberg: No, the function checks if it should read more. It should only read more if the token is not present in the list of blocks.

nathanielc: I get it. For some reason I was thinking that `more` took a terminating token as input and was checking against that. Obviously it is checking that the next token is not a terminating token, i.e. one in the stack

jsternberg:
The fuzzer had no crashes after 10 minutes so I'm going to merge this.

PR #889, opened 2019-01-18
fix(interpreter): cannot access imports of imports

### LINKED ISSUE #869: Package imports are externally visible
Given a package

    package foo
    import bar
    bar.v // allowed

Another package can import foo and access bar. This should not be allowed.

    import foo
    foo.bar // not allowed

jlapacik:
Fixes #869.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -244,6 +244,10 @@ func (itrp *Interpreter) doExpression(expr semantic.Expression, scope Scope) (va
 		if !ok {
 			return nil, fmt.Errorf("object has no property %q", e.Property)
 		}
+		if pkg, ok := v.(*Package); ok {
+			// If the property of a member expression represents a package, then the object itself must be a package.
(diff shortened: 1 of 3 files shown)


jlapacik:
@nathanielc we could separate the scope and the package object. I believe it was like that originally and will definitely be necessary if we ever decide to introduce private vs public variables.

Having said that, I really like the current implementation of the scope. If the only issue is that the REPL (incorrectly) shows a package's imports as visible when it evaluates it, my vote would be to change that specific behavior since those imports technically aren't visible to the REPL (it can't access them).

For instance, why not have a Package's String method skip over any package objects so the REPL won't print them?

jlapacik:
@nathanielc https://github.com/influxdata/flux/pull/889/commits/9d8c797d29fcb2c94ad51ec7e62d513a2679787d fixes the REPL. I don't consider it a hack since the interpreter will not allow you access to those imported packages anyway. What do you think? Or should we redesign the scope?

PR #892, opened 2019-01-18
feat(parser): construct invalid binary expressions when given multiple expressions

jsternberg:
When the parser is in a context where it is expecting a single expression,
the parser will now check to see if more tokens should be read. If more
tokens can be read before reaching the end of a block, it will parse
another expression and create a binary expression with no operator to
signify that the operator was missing.

The parser has also been tested and works properly with a missing left hand
side or right hand side operator. This will now properly produce an error.

- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/testing/testdata/totime.flux
+++ b/stdlib/testing/testdata/totime.flux
@@ -29,8 +29,7 @@ outData = "
 ,,0,2018-05-22T19:52:00Z,2030-01-01T00:00:00Z,k,m,2018-05-22T19:00:00Z
 "
 
-t_toTime = (table=<-) =>
-	(table
+t_toTime = (table=<-) => table
 		|> range(start: 2018-05-22T19:52:00Z)
 		|> toTime()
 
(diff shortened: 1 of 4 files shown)


stdlib/testing/testdata/totime.flux
```diff
@@ -29,8 +29,7 @@ outData = "
 ,,0,2018-05-22T19:52:00Z,2030-01-01T00:00:00Z,k,m,2018-05-22T19:00:00Z
 "
 
-t_toTime = (table=<-) =>
-	(table
+t_toTime = (table=<-) => table
```
wolffcm: So I guess the parser is caught the unbalanced `(`?

jsternberg: It took the rest of the expression as part of the one in the parenthesis and I noticed that it changed the error structure. Then I wondered why there was an error to begin with and I found this.

PR #896, opened 2019-01-18
Feat/check for option dependencies

### LINKED ISSUE #895: disallow dependencies among options declared in same package block
According to the SPEC, dependencies among options declared in the same package block are not allowed. Currently however there are no checks in place to prevent this from happening.

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

Fixes #895 .

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -928,7 +928,7 @@ A function produces side effects when it is explicitly declared to have side eff
 Packages are initialized in the following order:
 
 1. All imported packages are initialized and assigned to their package identifier. 
-2. All option declarations are evaluated and assigned regardless of order. An option cannot have dependencies on any other options assigned in the same package block.
+2. All option declarations are evaluated and assigned regardless of order. An option cannot have a dependency on itself or any other option assigned in the same package block.
 3. All variable declarations are evaluated and assigned regardless of order. A variable cannot have a direct or indirect dependency on itself.
 4. Any package side effects are evaluated.
 
--- a/semantic/check.go
+++ b/semantic/check.go
@@ -4,22 +4,26 @@ import "fmt"
 
 func runChecks(pkg *Package) error {
 	// check for options declared below package block
-	if err := checkOptionDecs(pkg); err != nil {
+	if err := checkOptionAssignments(pkg); err != nil {
+		return err
+	}
+	// check for dependencies among options
+	if err := checkOptionDependencies(pkg); err != nil {
 		return err
 	}
 	// check for variable reassignments
-	if err := checkVarDecs(pkg); err != nil {
+	if err := checkVarAssignments(pkg); err != nil {
 		return err
 	}
 	return nil
 }
 
-func checkOptionDecs(pkg *Package) error {
+func checkOptionAssignments(pkg *Package) error {
 	var stmt *OptionStatement
 	optionFn := func(opt *OptionStatement) {
 		stmt = opt
 	}
-	visitor := optionDecVisitor{
+	visitor := optionAssignmentVisitor{
 		optionFn:     optionFn,
 		packageBlock: true,
 	}
@@ -50,12 +54,14 @@ func optionName(opt *OptionStatement) (string, error) {
 	}
 }
 
-type optionDecVisitor struct {
+// This visitor finds options declared below the package block.
+// Any such option is passed to optionFn.
+type optionAssignmentVisitor struct {
 	optionFn     func(*OptionStatement)
 	packageBlock bool
 }
 
-func (v optionDecVisitor) Visit(node Node) Visitor {
+func (v optionAssignmentVisitor) Visit(node Node) Visitor {
 	n, ok := node.(*OptionStatement)
 	if ok && !v.packageBlock {
 		v.optionFn(n)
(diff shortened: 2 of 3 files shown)


semantic/check.go
```diff
@@ -118,3 +122,82 @@ func (v varDecVisitor) Nest() NestingVisitor {
 }
 
 func (v varDecVisitor) Done(node Node) {}
+
```
wolffcm: It would be helpful to describe the checks you're doing here with a comment.

docs/SPEC.md
```diff
@@ -928,7 +928,7 @@ A function produces side effects when it is explicitly declared to have side eff
 Packages are initialized in the following order:
 
 1. All imported packages are initialized and assigned to their package identifier. 
-2. All option declarations are evaluated and assigned regardless of order. An option cannot have dependencies on any other options assigned in the same package block.
+2. All option declarations are evaluated and assigned regardless of order. An option cannot have a dependency on itself or any other option assigned in the same package block.
 3. All variable declarations are evaluated and assigned regardless of order. A variable cannot have a direct or indirect dependency on itself.
```
wolffcm: My understanding of the intent here is that the initialization order of variables and options should be defined by the RHSs of assignments, and not anything else (not the order of `import` statements, nor lexicographic order of package names, etc)

semantic/check_test.go
```diff
@@ -1216,3 +1338,960 @@ func TestVarAssignmentChecks(t *testing.T) {
...
+										},
+										Name: "a",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			err: fmt.Errorf(`option dependency: option "b" depends on option "a" defined in the same package at b.flux|2:12-2:13`),
```
wolffcm: Is this possible to do?
```
package p
option o1 = 20
```
```
package q
import "p"
option p.o1 = p.o1 * 2
```
```
package r
import "p"
option p.o1 = p.or + 2
```
```
package s
import "p"
import "r"
import "s"
... p.o1 ... // use of o1
```
(o1 could be either 44 or 42)

wolffcm: Actually I didn't see any package qualifiers on the LHS of option assignments here, is that possible to do?

semantic/check_test.go
```diff
@@ -1216,3 +1338,960 @@ func TestVarAssignmentChecks(t *testing.T) {
...
+		},
+	}
```
wolffcm: Is it possible to have cyclical dependencies?
```
package p
option o1 = r.o2
```
```
package r
option o2 = p.o1
```

wolffcm: I think your walker will catch this, but it would be good to have a test.

semantic/check_test.go
```diff
@@ -1216,3 +1338,960 @@ func TestVarAssignmentChecks(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+		{
+			// package foo
+			// option a = 0
+			// option a = a + 1
+			//
```
wolffcm: It seems strange that this is not legal but that reassigning `a` with a different constant is.

docs/SPEC.md
jlapacik: Yes this is correct.

semantic/check_test.go
jlapacik: Yes reassigning options declared in an imported package is allowed. And in your example `o1` would be 44 because imports are ordered within a file. However there is no order across files. Hence if two files in the same package both imported other packages that mutated the same option, this would be non-deterministic behavior and is not allowed.

semantic/check.go
```diff
@@ -7,6 +7,10 @@ func runChecks(pkg *Package) error {
 	if err := checkOptionDecs(pkg); err != nil {
```
nathanielc: I had to do a double take to figure out that  `checkOptionDecs` and  `checkOptionDeps` are different. Maybe we should rename `Decs` to Assignments? Something like `checkOptionAsgns`?

semantic/check_test.go
jlapacik: This walker will only check option dependencies across a single package. Cyclic dependencies between packages would require a different method.

semantic/check.go
jlapacik: I agree. Thanks for the catch.

semantic/check_test.go
jlapacik: I agree it does seem strange and perhaps this changes in the future if it turns out to be too much of a handicap not allowing any option dependencies. But it is consistent with the definition of options being mutable global variables. And it doesn't introduce any non-determinism so I'd prefer not to restrict reassignments in that way.

jlapacik:
@wolffcm it is possible to have a statement such as `option p.foo = 0`. This is indeed a MemberAssignment and currently MemberAssignments are only allowed in the context of option reassignments.

semantic/check_test.go
wolffcm: I guess my example doesn't have any `import` statements, so my question is, do we check for cyclical imports?

jlapacik: We do not check for cyclic imports at the moment but it is needed per the spec. I don't believe there is an issue for it so I will create it.

Issue #205, opened 2018-10-31
Add ability to update objects as part of `map` and other functions


nathanielc:
Currently the map function can either merge the result with the group key or it can explicitly define all new columns. There is no way to generically say given an object update just these columns.

Other langauges for example OCaml have a`with` keyword like this:

```
|> map(fn:(r) => {r with _value: r._value+2}
```

That function updates `r` with the `_value` value column, all other columns of `r` are preserved whether they are part of the group key or not. Adding a similar syntax to Flux would remove the need to have the `mergeKey` boolean on map.

Here is OCaml's description of the `with` record expression:

> The expression { expr with  field1  [= expr1] ; … ;  fieldn  [= exprn] } builds a fresh record with fields field1 …  fieldn equal to expr1 …  exprn, and all other fields having the same value as in the record expr. In other terms, it returns a shallow copy of the record expr, except for the fields field1 …  fieldn, which are initialized to expr1 …  exprn.

http://caml.inria.fr/pub/docs/manual-ocaml/expr.html#htoc69

chnn:
JavaScript uses the spread operator to do something similar:

```js
rows.map(r => {...r, _value: r._value + 2})
```

pauldix:
What if you want to update multiple columns at the same time or insert a new one and update one? Possible to do?
```
|> map(fn:(r) => {r with {_value: r._value+2, foo:"bar"}}
```

aanthony1243:
this is now supported for row functions #1303 , but type checking is not working generally within the language.

nathanielc:
For the type bits see #1378

russorat:
Seems like this can be closed in favor of https://github.com/influxdata/flux/issues/1378 ?

PR #897, opened 2019-01-18
feat(ast): parse "with" operator for objects

nathanielc:
See #205

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/parser/grammar.md
+++ b/internal/parser/grammar.md
@@ -65,7 +65,7 @@ The parser directly implements the following grammar.
                                | ObjectLiteral
                                | ArrayLiteral
                                | ParenExpression .
-    ObjectLiteral              = "{" PropertyList "}" .
+    ObjectLiteral              = "{" ObjectBody "}" .
     ArrayLiteral               = "[" ExpressionList "]" .
     ParenExpression            = "(" ParenExpressionBody .
     ParenExpressionBody        = ")" FunctionExpressionSuffix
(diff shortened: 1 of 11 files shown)


internal/parser/parser.go
```diff
@@ -892,6 +890,59 @@ func (p *parser) parseParenIdentExpression(lparen token.Pos, key *ast.Identifier
...
+		properties := p.parsePropertyListSuffix(str)
+		return &ast.ObjectExpression{
+			Properties: properties,
+		}
+	default:
+		// TODO(nathanielc): BadExpression.
+		return nil
+	}
+
+}
```
affo: For the same reason as in the grammar, in this code: if you get a string literal, it is a standard object and you can parse the property list. If it is an identifier, then you `peek` the next token, if it is a `with` than you proceed and parse the property list; if it is a `COLON` then it was a property list. Does it sound reasonable?

internal/parser/grammar.md
```diff
@@ -80,8 +80,15 @@ The parser directly implements the following grammar.
...
+    ObjectBody                 = identifier ObjectBodySuffix .
+                               | string_lit PropertyListSuffix .
+    ObjectBodySuffix           = "with" PropertyList
+                               | PropertyListSuffix .
+    PropertyListSuffix         = PropertySuffix { "," PropertyList } .
     PropertyList               = [ Property { "," Property } ] .
-    Property                   = identifier [ ":" Expression ]
+    Property                   = identifier PropertySuffix
+                               | string_lit PropertySuffix .
+    PropertySuffix             = [ ":" Expression ] .
```
affo: Why don't:

```
ObjectBody = identifier "with" PropertyList .
           | PropertyList .
```

?

PR #905, opened 2019-01-22
feat(stdlib): socket source

affo:
Must be merged after #904 (the relevant commit is the second one).

Socket source that takes input from a socket, decodes it (raw or csv) and outputs table.

It is still WIP. For example, the raw decoder should add `_start`, `_stop`, and `_time`.

The broader idea is to create a class of streaming sources, the next step is also to provide a `stdio` package with `from` and `to`.  
An issue that comes to my mind in doing streaming: If we process the same group key twice, most of the functions fail. Should we create multiple caches based on different time frames?

An issue that I can see now: the `csv` decoder uses the `csv.ResultDecoder` which actually streams out tables as they complete. However, as long as `socket.from` is by itself, it works well. As I add another transformation (e.g. `group`, that doesn't fail on the same gk twice), the output is provided once the connection is closed.

I foresee to include a rule to merge `socket.from` and `window` to make `raw` decoder stream out tables using windowing semantics. Of course, this merge should apply to any streaming source (`io` too).

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/socket/from_test.go
```diff
@@ -0,0 +1,241 @@
...
+			c := execute.NewTableBuilderCache(executetest.UnlimitedAllocator)
+			c.SetTriggerSpec(flux.DefaultTrigger)
+			r := ioutil.NopCloser(bytes.NewReader([]byte(tc.input)))
+			ss, err := socket.NewSocketSource(tc.spec, r, id, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// Add `yield` in order to add `from` output tables to cache.
+			ss.AddTransformation(executetest.NewYieldTransformation(d, c))
```
jlapacik: Ah I see the reason for the yield now. Disregard my previous comment.

stdlib/socket/from.go
```diff
@@ -0,0 +1,309 @@
...
+	"csv":  true,
+	"raw":  true,
+}
+
+type FromSocketOpSpec struct {
+	URL     string `json:"url"`
+	Decoder string `json:"decoder"`
+}
+
+func init() {
```
nathanielc: Can you move this init function to the top of the file?  That way its easy to see the function signature being used and follows the convention of the rest of the builtins.

stdlib/socket/from.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+	"github.com/pkg/errors"
+)
+
+const FromSocketKind = "fromSocket"
+
+type RawDecoder struct {
```
nathanielc: What does RawDecoder do? What format is it decoding?

stdlib/socket/from.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	itr := &socketIterator{
+		rc:      rc,
+		decoder: decoder,
+		data:    make(chan flux.Table),
+		err:     make(chan error),
+		done:    make(chan struct{}),
+	}
+	return execute.CreateSourceFromDecoder(itr, dsid, a)
```
nathanielc: It looks like using  the CreateSourceFromDecoder is not ideal for this use case. The flux.ResultDecoder already matches the execute.Source interface well. Maybe we should use the direct source interface here instead?

stdlib/socket/from.go
```diff
@@ -0,0 +1,309 @@
...
+				i.data <- table
+				return nil
+			})
+			if err != nil {
+				i.err <- err
+			}
+			close(i.data)
+			close(i.err)
+			close(i.done)
+		}()
```
nathanielc: You should not need to use  a go routine inside of the SourceDecoder interface. This is a sign that the bare execute.Source interface should be used instead.

affo: Yeah, I see your point. I wanted to implement that interface because conceptually it is what I need, but it doesn't match the behavior of our decoders. Fixing to execute.Source would also fix the race problem.

stdlib/socket/from.go
affo: Sure 👍

stdlib/socket/from.go
affo: I'll put docs above it.

nathanielc: What is the purpose of the decoder? Just a simple way to test the source?

affo: Nope, it takes raw input from the reader and puts it into a `_value` column. I put the doc above it now.
It is a way to get raw messages from the socket, without any specified schema as the CsvDecoder does.

nathanielc: OK, in that case I think we should change a few things. 

* It should be moved to its own package and have unit tests for its decoding methods.
* It should not add the _start and_stop columns, that is the function of the range transformation and it doesn't make sense here

nathanielc: The influxdb.from adds _start and _stop because it is acting as the range transformation since it gets pushed down.

affo: For the first point 👍 

For the second, how `range` work with something that doesn't have timestamps attached?
The same for `window`.

affo: Oh ok, I misunderstood your point. So, _time is ok, but not _start and _stop.

stdlib/socket/from_test.go
nathanielc: OK, if `yieldTransformation` is only for tests then lets move it to a test package.

affo:
@nathanielc `yieldTransformation` is in the `executetest` package. What do you mean?

stdlib/socket/from.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func createFromSocketSource(s plan.ProcedureSpec, dsid execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	spec, ok := s.(*FromSocketProcedureSpec)
+	if !ok {
+		return nil, fmt.Errorf("invalid spec type %T", s)
+	}
+
+	url := spec.URL
+	conn, err := net.Dial("tcp", url)
```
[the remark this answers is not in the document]
affo: I addressed it in the last commit, now the URL is suppossed to include the network:

```
tcp://...
unix://
```

If no protocol is specified, it defaults to `tcp`.

stdlib/socket/from.go
```diff
@@ -0,0 +1,201 @@
...
+
+func createFromSocketSource(s plan.ProcedureSpec, dsid execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	spec, ok := s.(*FromSocketProcedureSpec)
+	if !ok {
+		return nil, fmt.Errorf("invalid spec type %T", s)
+	}
+
+	url := spec.URL
+	network := "tcp"
+	i := strings.LastIndex(url, "://")
```
nathanielc: Can you use `url.Parse` here instead of parsing the scheme yourself?

stdlib/socket/from.go
```diff
@@ -0,0 +1,201 @@
...
+	// if :// is at the beginning or end,
+	// we pass the url as it is, and use "tcp" as default.
+	if i > 0 && i < len(url)-3 {
+		network = url[:i]
+		url = url[i+3:]
+	}
+	conn, err := net.Dial(network, url)
```
nathanielc: What about `file://`? The net.Dial function will not open the file for you.

line/result.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// NewResultDecoder creates a new result decoder from config.
+func NewResultDecoder(config *ResultDecoderConfig) *ResultDecoder {
+	return &ResultDecoder{config: config}
+}
+
+// TimeProvider gives the current time.
+type TimeProvider interface {
+	GetCurrentTime() values.Time
```
nathanielc: nit, this should not need the prefix `Get`, so `CurrentTime` would be a better name for this method.

line/result.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// ResultDecoderConfig is the configuration for a result decoder.
+type ResultDecoderConfig struct {
+	Separator byte
+	Tp        TimeProvider
```
nathanielc: This field is more clear if it is called `TimeProvider`

stdlib/socket/from.go
```diff
@@ -0,0 +1,201 @@
...
+
+type administrationTimeProvider struct {
+	a execute.Administration
+}
+
+func (a *administrationTimeProvider) GetCurrentTime() values.Time {
+	return a.a.ResolveTime(flux.Time{
+		IsRelative: false,
+		Absolute:   time.Now(),
+	})
```
nathanielc: Why are we resolving times through the administration?

stdlib/socket/from.go
affo: 👍

stdlib/socket/from.go
affo: Mmmm, that's ok, but this is a SocketSource, should I account for it or let net.Dial return an error?

stdlib/socket/from.go
affo: I don't know actually. I thought that was the way to go for preserving the `now` specified by the user and so on.
The fact you are asking makes me think there is a much easier way 😅

nathanielc: Using a.a.ResolveTime only matters if IsRelative is true. In this case it makes it look like it doing something useful when in reality its not. Lets just have the code return time.Now() converted to a values.Time. That way its simple and clear that no real resolving is happening.

stdlib/socket/from_test.go
nathanielc: I see its in the executetest package we are good then.

stdlib/socket/from.go
nathanielc: Lets only pass networks into net.Dial that we know will work, that way we can return our own more meaningful error messages.

nathanielc:
OK, this is close, I have answered your questions. Once the small nits are complete this should be good to go. Thanks!

stdlib/socket/from.go
affo: 👍

stdlib/socket/from.go
affo: Agreed. In the future, the source could serve as a base class for every source that needs a `io.ReadCloser`. And so being used by:
 - `csv.from`
 - `file.from`
 - `stdio.from`
 - `net.from`

`csv` could also become `file.from(decoder:"csv", ...)`.

Issue #906, opened 2019-01-22
Derivative, integral and stateDuration should error if the _time is not sorted


nathanielc:
The idea here is that when the planner can understand sorting requirements it can prevent these errors in the future. The currently implementation should avoid these errors since data is  sorted by _time by default.

wolffcm:
Hey @affo I'll update the behavior for `derivative` and I guess you can update your existing PR for `stateDuration` to have the behavior discussed at IPM.  I'm not familiar with `integral`, I guess either one of us could take a look.

affo:
For integral, I can do it

affo:
Fixed by #908 and #913

PR #908, opened 2019-01-22
refactor(stdlib): make derivative return error for null or out-of-order times

wolffcm:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

PR #913, opened 2019-01-23
refactor(stdlib): make integral return err for null or out-of-order t…

affo:
…imes, add support for int and uint

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/integral.go
+++ b/stdlib/universe/integral.go
@@ -8,6 +8,7 @@ import (
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
 )
 
 const IntegralKind = "integral"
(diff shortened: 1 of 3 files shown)


stdlib/universe/integral.go
```diff
@@ -146,37 +147,70 @@ func (t *integralTransformation) Process(id execute.DatasetID, tbl flux.Table) e
...
+		}
+
+		if tbl.Key().HasCol(c) {
+			return fmt.Errorf("cannot aggregate columns that are part of the group key")
+		}
+
+		if typ := cols[idx].Type; typ != flux.TFloat &&
+			typ != flux.TInt &&
+			typ != flux.TUInt {
+			return fmt.Errorf("cannot perform integral over %v", typ)
```
wolffcm: Nice.  When I modified `derivative` I neglected to add checks like this, so I filed a bug for it.

wolffcm: Apropos of nothing, it irks me that we have to do this check at run time.

stdlib/universe/integral.go
```diff
@@ -146,37 +147,70 @@ func (t *integralTransformation) Process(id execute.DatasetID, tbl flux.Table) e
...
+				prevTime = tm
+
+				switch tbl.Cols()[j].Type {
+				case flux.TInt:
+					if vs := cr.Ints(j); vs.IsValid(i) {
+						in.updateFloat(tm, float64(vs.Value(i)))
+					}
+				case flux.TUInt:
+					if vs := cr.UInts(j); vs.IsValid(i) {
+						in.updateFloat(tm, float64(vs.Value(i)))
```
wolffcm: Would this have panicked before for `int` and `uint` columns?   Nice catch.  I guess doing it this way we lose some numeric accuracy for 64-bit integer values that cannot be represented as a float.  (Just making the observation, not suggesting you fix this now)

affo: Oh, good catch!  
Yep, it panicked 😅

stdlib/universe/integral.go
affo: Yeah, you are right. Anyhow, I think that being schemaless would always require runtime checks on the schema.
If we had typed streams we could do any type checking at compile time and also avoid schema collisions. However we would loose flexibility. I am always open to introduce typed streams :)

stdlib/universe/integral_test.go
```diff
@@ -205,9 +296,58 @@ func TestIntegral_Process(t *testing.T) {
...
+					{Label: "_time", Type: flux.TTime},
+					{Label: "_value", Type: flux.TFloat},
+					{Label: "t", Type: flux.TString},
+				},
+				Data: [][]interface{}{
+					{execute.Time(1), execute.Time(3), execute.Time(1), 2.0, "a"},
+					{execute.Time(1), execute.Time(3), execute.Time(2), 1.0, "b"},
 				},
 			}},
+			wantErr: fmt.Errorf("cannot perform integral over string"),
```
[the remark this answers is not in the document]
affo: Bug fixed, thank you :)

affo:
Fixes #906 together with #908

PR #916, opened 2019-01-23
fix(csv/result): fix decoder bug where a default table id is given when none is required.

### LINKED ISSUE #898: Infinite loop in REPL when calling csv.from
I get an infinite loop of flux tables when I enter this from the REPL:

```
inData = "
#datatype,string,long,string,long
#group,false,false,true,false
#default,_result,long,,
,result,table,key,val
,,0,k,0
,,0,k,1
,,0,k,2
"
import "csv"
csv.from(csv: inData)
```
I get this:
```
Table: keys: [key]
            key:string                     val:int
----------------------  --------------------------
Table: keys: [key]
            key:string                     val:int
----------------------  --------------------------
```
Over and over, forever.  I have to `kill -9` to stop it.

aanthony1243:
fixes #898 

related issue: #919 

### Done checklist
- [X] Test cases written

### DIFF
--- a/csv/result.go
+++ b/csv/result.go
@@ -319,6 +319,9 @@ func readMetadata(r *csv.Reader, c ResultDecoderConfig, extraLine []string) (tab
 		case defaultAnnotation:
 			resultID = line[resultIdx]
 			tableID = line[tableIdx]
+			if _, err := strconv.ParseInt(tableID, 10, 64); tableID != "" && err != nil {
+				return tableMetadata{}, fmt.Errorf("default Table ID is not an integer")
+			}
 			defaults = copyLine(line[recordStartIdx:])
 		default:
 			if annotation == "" {
@@ -579,10 +582,10 @@ DONE:
 }
 
 func (d *tableDecoder) init(line []string) error {
-	if d.meta.TableID != "" {
-		d.id = d.meta.TableID
-	} else if len(line) != 0 {
+	if len(line) != 0 {
 		d.id = line[tableIdx]
+	} else if d.meta.TableID != "" {
+		d.id = d.meta.TableID
 	} else {
 		return errors.New("missing table ID")
 	}
--- a/csv/result_test.go
+++ b/csv/result_test.go
@@ -555,6 +555,63 @@ func TestResultDecoder(t *testing.T) {
 			},
 		},
 		{
+			name:          "single table with unnecessary default tableID",
+			encoderConfig: csv.DefaultEncoderConfig(),
+			encoded: toCRLF(`#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,double
+#group,false,false,true,true,false,true,true,false
+#default,_result,111,,,,,,
+,result,table,_start,_stop,_time,_measurement,host,_value
+,,0,2018-04-17T00:00:00Z,2018-04-17T00:05:00Z,2018-04-17T00:00:00Z,cpu,A,42
+,,0,2018-04-17T00:00:00Z,2018-04-17T00:05:00Z,2018-04-17T00:00:01Z,cpu,A,43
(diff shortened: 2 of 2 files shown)


csv/result.go
```diff
@@ -579,10 +579,10 @@ DONE:
 }
 
 func (d *tableDecoder) init(line []string) error {
-	if d.meta.TableID != "" {
-		d.id = d.meta.TableID
-	} else if len(line) != 0 {
+	if len(line) != 0 {
```
wolffcm: I'm unclear on what `line` is in this context...Is it the `#defaults` annotation row, or the first line of the table, or something else?

jlapacik: @wolffcm `line` represents the first record from the table.

wolffcm: It would be great to rename this parameter to something like `firstRecord`.

wolffcm:
@jlapacik Is it an error to have defaults for a non-empty table?  Reading the SPEC, it's not the impression that I get.  One could omit values from the table rows and just use the defaults.  This would be useful for tables with lots of repeated values.

stdlib/testing/testdata/csvloop.flux
```diff
@@ -0,0 +1,27 @@
...
+#datatype,string,long,string,long
+#group,false,false,true,false
+#default,_result,,,
+,result,table,key,val
+,,0,k,0
+,,0,k,1
+,,0,k,2
+"
+
+
```
wolffcm: It would be good to link to an issue here and note that this should actually be an error since `long` will not parse as an int, and the `#datatype` line says the tableID column should be a long.

nathanielc:
how does this solve the loop issue?

jlapacik:
@aanthony1243 informed me that defaults are only used in the case of empty tables and so should probably be an error if given for a non-empty table as right now that behavior is undefined.

stdlib/testing/testdata/csvloop.flux
```diff
@@ -0,0 +1,27 @@
...
+// reading the data should result in a correctly read table that forgets about the default.
+outData = "
+#datatype,string,long,string,long
+#group,false,false,true,false
+#default,_result,,,
+,result,table,key,val
+,,0,k,0
+,,0,k,1
+,,0,k,2
+"
```
nathanielc: Can you add a test case to the `csv` package? Using the unit tests we already have for csv encoding/decoding?

Using a Flux test to ensure the csv parsing works seems odd.

nathanielc:
According to the SPEC defaults apply even for non empty tables. From the SPEC:

> default - a default value to be used for rows whose string value is the empty string.

They are the value that is to be used if the value for a cell is the empty string. We make use of this for non empty tables already in the `result` column, we set the default to `_result` indicating that every row's `result` column value is `_result`.

nathanielc:
> In that case an error should be returned if a default value cannot be parsed as its intended datatype.

yes that is correct

aanthony1243:
@nathanielc the crux of what I fixed is that if a row has a tableID it uses that first, and then checks for the default value.   This seems to be the correct logic to have according to the spec.  

re: why this fixes the infinite loop, as I told Chris earlier:  "b.advance was returning that it was time to start the next table and returns the input line as an extra line.  the extra line gets passed to b.advance where it determines that it's time to start the next table and returns the input...."

finally, as I discussed:  validating the default as being a proper `long` type is a separate issue from this.  A valid value of `1111`  instead of `"long"`  would have also looped infinitely. So I think we should accept this change as-is and then refactor the default table ID processing separately.  

finally:  the tableID being a `long` value seems to be little more than a convention. it's only used to indicate a transition from one table to the next.  the ID is not copied into the `flux` runtime at all.  as long as each table had a unique string ID that would be sufficient.  

will add a unit test, I agree with your point there.

PR #920, opened 2019-01-23
refactor(influxdb): remove from implementation to move it to influxdb

affo:
InfluxDB PR: https://github.com/influxdata/influxdb/pull/11515.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/querytest/compiler.go
+++ b/querytest/compiler.go
@@ -4,57 +4,35 @@ import (
 	"context"
 
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/stdlib/csv"
-	"github.com/influxdata/flux/stdlib/influxdata/influxdb"
-	"github.com/influxdata/flux/stdlib/influxdata/influxdb/v1"
 )
 
-// FromCSVCompiler wraps a compiler and replaces all From operations with FromCSV
-type FromCSVCompiler struct {
+// ReplaceOpInSpecFunction is a function that takes an Operation and returns a new OperationSpec for substitution,
+// or nil, if nothing has to be changed.
+type ReplaceOpInSpecFunction func(op *flux.Operation) flux.OperationSpec
+
+type ReplaceSpecCompiler struct {
 	flux.Compiler
-	InputFile string
+	fn ReplaceOpInSpecFunction
 }
 
-// FromInfluxJSONCompiler wraps a compiler and replaces all From operations with FromJSON
-type FromInfluxJSONCompiler struct {
-	flux.Compiler
-	InputFile string
+func NewReplaceSpecCompiler(compiler flux.Compiler, fn ReplaceOpInSpecFunction) *ReplaceSpecCompiler {
+	return &ReplaceSpecCompiler{Compiler: compiler, fn: fn}
 }
 
-func (c FromCSVCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
+func (c *ReplaceSpecCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
 	spec, err := c.Compiler.Compile(ctx)
 	if err != nil {
 		return nil, err
 	}
-	ReplaceFromSpec(spec, c.InputFile)
+	ReplaceOpInSpec(spec, c.fn)
 	return spec, nil
 }
 
-func (c FromInfluxJSONCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
-	spec, err := c.Compiler.Compile(ctx)
-	if err != nil {
-		return nil, err
-	}
-	ReplaceFromWithFromInfluxJSONSpec(spec, c.InputFile)
-	return spec, nil
-}
-
-func ReplaceFromSpec(q *flux.Spec, csvSrc string) {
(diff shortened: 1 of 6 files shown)


querytest/compiler.go
```diff
@@ -9,52 +9,51 @@ import (
...
+func FromInfluxJSONCompiler(c flux.Compiler, jsonFile string) *ReplaceSpecCompiler {
+	return NewReplaceCompiler(c, func(op *flux.Operation) {
 		if op.Spec.Kind() == influxdb.FromKind {
 			op.Spec = &v1.FromInfluxJSONOpSpec{
-				File: jsonSrc,
+				File: jsonFile,
 			}
 		}
-	}
+	})
```
affo: This change is required in order to let InfluxDB change the operation spec according to its kinds and not Flux's.

stdlib/influxdata/influxdb/from_test.go
```diff
@@ -1,714 +0,0 @@
-package influxdb_test
```
wolffcm: It seems like there should be at least some testing in here, even if `from` doesn't match to an actual source in vanilla Flux, and therefore cannot be actually executed.  How about a marshaling test, similar to the other operations?

stdlib/influxdata/influxdb/from.go
```diff
@@ -123,446 +76,35 @@ func (s *FromProcedureSpec) Kind() plan.ProcedureKind {
...
-	if !s.BoundsSet || (s.Bounds.Start.IsZero() && s.Bounds.Stop.IsZero()) {
-		var bucket string
-		if len(s.Bucket) > 0 {
-			bucket = s.Bucket
-		} else {
-			bucket = s.BucketID
-		}
-		return fmt.Errorf(`%s: results from "%s" must be bounded`, id, bucket)
+	if !s.Bounded {
+		return fmt.Errorf(`%s: results from "%s" must be bounded`, id, s.Bucket)
```
wolffcm: I wonder if this is needed here?  As it is now, if we have a `from` in the vanilla REPL, it will say it needs to be bounded.  Then when the user bounds it with a `range`, execution says, "I can't do that."  

Maybe vanilla `from` doesn't need to be bounded?  It will be an error to execute it no matter what.

stdlib/influxdata/influxdb/from_test.go
affo: I removed marshaling intentionally, because I thought that this kind of from shouldn't ever come in a spec from the API. Should I put in place again?

stdlib/influxdata/influxdb/from.go
affo: I kept a behavior similar to InfluxDB's from because some tests rely on it. I was tempted to remove the need for a bound. Should I do it and change tests as a consequence?

querytest/compiler.go
affo: An interesting thing about this piece of code is that actually it is unused in Flux! Should I move it towards InfluxDB?

stdlib/influxdata/influxdb/from.go
affo: 👍 seems legit. I should change some tests around. Like the controller's ones that expect to fail if no range is applied.

querytest/compiler.go
affo: 👍 Ok for this version? Or the previous one?

stdlib/influxdata/influxdb/from.go
wolffcm: Oh, yeah.  I think I wrote that one.  Maybe you can use your valdiation-breaking rules to cause a planner error in that case.

stdlib/influxdata/influxdb/from_test.go
affo: I mean, this from doesn't do anything 😅 

I can add a test for `NewQuery`, even if not even `Bucket` is required :)

querytest/compiler.go
wolffcm: Not sure what you mean?  We capture 1.x data as JSON, and `fromInfluxJSON` can read it and produce Flux tables.

stdlib/influxdata/influxdb/from_test.go
wolffcm: Yeah, I think just having a super-simple `NewQuery` test would be good, to show the Flux text produces a Spec.

querytest/compiler.go
affo: I wasn't clear, sorry. I mean, should I move this version to InfluxDB or the previous one? This one is more general in that it lets specify generic `ReplaceCompiler`s.

wolffcm: Oh, got it. 😄   I like the refactoring you did for `ReplaceSpecCompiler` and it seems like it belongs in `flux`.  I think you can just move `FromInfluxJSONCompiler` to `influxdb`.

affo:
@wolffcm 
I had to slightly change `querytest/compiler.go` because of idpe dependencies. Is it ok for you?

PR #922, opened 2019-01-23
feat(ast): expose literal parsers used within the parser

### LINKED ISSUE #921: Expose literal parsing methods in the ast package
The literal parsing methods are needed outside of the parser so they need to be exposed.

These methods are located inside of `internal/parser/parser.go`.

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The literal parsers take the raw literal that is tokenized by the
scanner and returns the appropriate value type to be used with the ast
literals.

Fixes #921.

### DIFF
--- a/ast/strconv.go
+++ b/ast/strconv.go
@@ -1,0 +1,165 @@
+package ast
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+	"unicode/utf8"
+)
+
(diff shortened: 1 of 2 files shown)


jsternberg:
I was wondering if these should be in another package. The way the Go standard library does it is most of these are in the `strconv` package.

I don't know if that's appropriate for us or not. I also don't want to get too many conflicting package names with the Go standard library. We can probably review this later when we review the API for flux 1.0.

PR #926, opened 2019-01-24
feat(testing): add assertEmpty method and use it with testing.test

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

This modifies `testing.test` to use `testing.diff` and a new
`testing.assertEmpty` function.

Diff has now been modified so it does not emit a table when the table is
empty and `assertEmpty` will return an error when one of the tables is
not empty. `testing.test` has been modified to include a yield and the
end to end tests have been modified to use the console-friendly
formatter so any output from `diff` will be output with the test
failure.

### DIFF
--- a/stdlib/testing/assert_empty.go
+++ b/stdlib/testing/assert_empty.go
@@ -1,0 +1,111 @@
+package testing
+
+import (
+	"fmt"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+)
+
+const AssertEmptyKind = "assertEmpty"
+
+type AssertEmptyOpSpec struct{}
+
+func (s *AssertEmptyOpSpec) Kind() flux.OperationKind {
+	return AssertEmptyKind
+}
+
+func init() {
+	assertEmptySignature := flux.FunctionSignature(nil, nil)
+
+	flux.RegisterPackageValue("testing", "assertEmpty", flux.FunctionValue(AssertEmptyKind, createAssertEmptyOpSpec, assertEmptySignature))
+	flux.RegisterOpSpec(AssertEmptyKind, newAssertEmptyOp)
+	plan.RegisterProcedureSpec(AssertEmptyKind, newAssertEmptyProcedure, AssertEmptyKind)
+	execute.RegisterTransformation(AssertEmptyKind, createAssertEmptyTransformation)
+}
+
+func createAssertEmptyOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
+	if err := a.AddParentFromArgs(args); err != nil {
+		return nil, err
+	}
+	return &AssertEmptyOpSpec{}, nil
+}
+
+func newAssertEmptyOp() flux.OperationSpec {
+	return new(AssertEmptyOpSpec)
+}
+
+type AssertEmptyProcedureSpec struct {
+	plan.DefaultCost
+}
+
+func (s *AssertEmptyProcedureSpec) Kind() plan.ProcedureKind {
+	return AssertEmptyKind
+}
+
+func (s *AssertEmptyProcedureSpec) Copy() plan.ProcedureSpec {
+	ns := *s
+	return &ns
+}
+
+func newAssertEmptyProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.ProcedureSpec, error) {
+	if _, ok := qs.(*AssertEmptyOpSpec); !ok {
+		return nil, fmt.Errorf("invalid spec type %T", qs)
+	}
+	return &AssertEmptyProcedureSpec{}, nil
+}
+
+type AssertEmptyTransformation struct {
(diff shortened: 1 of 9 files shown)


jsternberg:
@nathanielc @aanthony1243 this doesn't have everything it needs (if we use this), but I want to start a discussion.

Here is sample output. I modified a test file to change the expected data to the wrong thing and then changed it back.

```
jsternberg:[~/g/s/g/i/flux] $ (feat/testing-assert-empty *+$) go test -v -run=Test_QueryEndToEnd/'^'mean.flux ./stdlib/testing/
=== RUN   Test_QueryEndToEnd
=== RUN   Test_QueryEndToEnd/mean.flux
Table: keys: [_start, _stop, _measurement, _field]
                   _start:time                      _stop:time     _measurement:string           _field:string            _diff:string                  _value:float
------------------------------  ------------------------------  ----------------------  ----------------------  ----------------------  ----------------------------
2018-12-01T00:00:00.000000000Z  2030-01-01T00:00:00.000000000Z                     Sgf                 GxUPYq1                       -                             9
2018-12-01T00:00:00.000000000Z  2030-01-01T00:00:00.000000000Z                     Sgf                 GxUPYq1                       +                           9.5
--- FAIL: Test_QueryEndToEnd (0.01s)
    --- FAIL: Test_QueryEndToEnd/mean.flux (0.01s)
        end_to_end_test.go:123: found 1 tables that were not empty
FAIL
FAIL    github.com/influxdata/flux/stdlib/testing       0.382s
jsternberg:[~/g/s/g/i/flux] $ (feat/testing-assert-empty *+$) vim stdlib/testing/testdata/mean.flux
jsternberg:[~/g/s/g/i/flux] $ (feat/testing-assert-empty *+$) go test -v -run=Test_QueryEndToEnd/'^'mean.flux ./stdlib/testing/
=== RUN   Test_QueryEndToEnd
=== RUN   Test_QueryEndToEnd/mean.flux
--- PASS: Test_QueryEndToEnd (0.01s)
    --- PASS: Test_QueryEndToEnd/mean.flux (0.01s)
PASS
ok      github.com/influxdata/flux/stdlib/testing       0.380s
```

stdlib/testing/assert_empty.go
```diff
@@ -0,0 +1,126 @@
...
+
+func init() {
+	assertEmptySignature := semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			flux.TablesParameter: flux.TableObjectType,
+		},
+		Required:     semantic.LabelSet{flux.TablesParameter},
+		Return:       flux.TableObjectType,
+		PipeArgument: flux.TablesParameter,
+	}
```
nathanielc: There is a helper method flux.FunctionSignature that you can use here instead.

stdlib/testing/assert_empty.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (t *AssertEmptyTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey) error {
+	return t.d.RetractTable(key)
+}
+
+func (t *AssertEmptyTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	var failed bool
+	if err := tbl.Do(func(cr flux.ColReader) error {
```
nathanielc: The table interface has an `Empty` method you can call directly

jsternberg: We still need to call `Do` if I remember correctly. I wasn't sure if I should rely on `.Empty()`. I'll change things to use that.

nathanielc: Hmm, I think you are right that we need to read the data or it will deadlock. Can you add a comment to that fact since it is easy to miss?

jsternberg:
Updated the output of this and I made sure that `Finish` was actually called and `FinishedErr` was checked.

```
--- FAIL: TestAssertEmpty_Process (0.00s)
    --- FAIL: TestAssertEmpty_Process/non-empty_table (0.00s)
        assert_empty_test.go:59: expected error found 1 tables that were not empty, got none
--- FAIL: Test_QueryEndToEnd (1.07s)
    --- FAIL: Test_QueryEndToEnd/mean.flux (0.01s)
        end_to_end_test.go:132: found 1 tables that were not empty
        end_to_end_test.go:122: Table: keys: [_start, _stop, _measurement, _field]
        end_to_end_test.go:122:                    _start:time                      _stop:time     _measurement:string           _field:string            _diff:string                  _value:float
        end_to_end_test.go:122: ------------------------------  ------------------------------  ----------------------  ----------------------  ----------------------  ----------------------------
        end_to_end_test.go:122: 2018-12-01T00:00:00.000000000Z  2030-01-01T00:00:00.000000000Z                     Sgf                 GxUPYq1                       -                             9
        end_to_end_test.go:122: 2018-12-01T00:00:00.000000000Z  2030-01-01T00:00:00.000000000Z                     Sgf                 GxUPYq1                       +                           9.5
```

That's an example of a test that fails.

PR #927, opened 2019-01-24
bug: options editor should use ast.Expression

docmerlin:
- [x ] Test cases written

### DIFF
--- a/ast/edit/option_editor.go
+++ b/ast/edit/option_editor.go
@@ -2,11 +2,8 @@ package edit
 
 import (
 	"fmt"
-	"sort"
 
 	"github.com/influxdata/flux/ast"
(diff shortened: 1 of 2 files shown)


ast/edit/option_editor.go
```diff
@@ -66,9 +66,9 @@ func OptionObjectFn(keyMap map[string]values.Value) OptionFn {
 		}
 
 		for _, p := range obj.Properties {
-			value, found := keyMap[p.Key.Key()]
+			exp, found := keyMap[p.Key.Key()]
 			if found {
-				p.Value = CreateLiteral(value)
```
nathanielc: If CreateLiteral is unused can you delete it?

PR #937, opened 2019-01-31
Add compile subcommand that compiles flux to spec

desa:
This PR introduces a `compile` subcommand that translates a flux script into a spec.

### DIFF
--- a/cmd/flux/cmd/compile.go
+++ b/cmd/flux/cmd/compile.go
@@ -1,0 +1,59 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
+	_ "github.com/influxdata/flux/builtin"
+	"github.com/influxdata/flux/lang"
+	"github.com/spf13/cobra"
+)
+
+// compileCmd represents the compile command
+var compileCmd = &cobra.Command{
+	Use:   "compile",
+	Short: "Compile a Flux script into a query specification",
+	Long:  "Compile a Flux script into a query specification (use @ as prefix to the file)",
+	Args:  cobra.ExactArgs(1),
+	RunE:  compile,
+}
+
+var prettyPrint bool
+
+func init() {
+	rootCmd.AddCommand(compileCmd)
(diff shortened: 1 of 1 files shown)


cmd/flux/cmd/compile.go
```diff
@@ -0,0 +1,65 @@
...
+
+var prettyPrint bool
+
+func init() {
+	compileCmd.Flags().BoolVarP(&prettyPrint, "pretty-print", "p", false, "pretty print the compiled query")
+}
+
+func init() {
+	rootCmd.AddCommand(compileCmd)
+}
```
affo: @desa can you merge the two init functions?

cmd/flux/cmd/compile.go
```diff
@@ -0,0 +1,65 @@
...
+
+	_ "github.com/influxdata/flux/builtin"
+	"github.com/influxdata/flux/lang"
+	"github.com/spf13/cobra"
+)
+
+// compileCmd represents the compile command
+var compileCmd = &cobra.Command{
+	Use:   "compile",
+	Short: "Compile a Flux script",
```
nathanielc: Compile has lots of meanings in the Flux eco system right now, can we be more specific in these descriptions to help clarify which kind of compile we mean. Maybe something like `Compile a Flux script into a query specification.`

cmd/flux/cmd/compile.go
```diff
@@ -0,0 +1,65 @@
...
+	spec, err := c.Compile(context.Background())
+	if err != nil {
+		return err
+	}
+
+	enc := json.NewEncoder(os.Stdout)
+	if prettyPrint {
+		enc.SetIndent("", " ")
+	}
+	if err := enc.Encode(spec); err != nil {
```
nathanielc: this can be simplified to `return enc.Encode(spec)`

cmd/flux/cmd/compile.go
desa: @nathanielc Was the `.` intended to be included? I just checked a couple of the other commands and there seems to be inconsistent usage of periods.

nathanielc: It was not intended. I'd like it to be consistent, but I am not sure we have decided on what that means. I have no opinion either way.

PR #947, opened 2019-02-05
feat: make duration conversion public

docmerlin:
This is needed for options handling.


affo:
It would also be cool to have a method that takes a `time.Duration` and returns an `ast.DurationLiteral`.  
It could be done by properly splitting the result of `time.Duration.String()` and creating a slice of `ast.Duration`. I can't remember where, but some tests would have needed it.

PR #951, opened 2019-02-05
Chore/move strconv2parser

docmerlin:


ast/edit/option_editor_test.go
```diff
@@ -75,17 +75,17 @@ option task = {
 	retry: 10,
 }`,
 			edit: func(node ast.Node) (bool, error) {
-				every, err := ast.ParseDuration("2hr3m10s")
+				every, err := parser.ParseDuration("2hr3m10s")
```
jsternberg: This is likely supposed to be `2h` and not `2hr`.

ast/ast.go
```diff
@@ -1130,6 +1131,38 @@ type Duration struct {
 	Unit      string `json:"unit"`
 }
 
+func (l *Duration) Duration() (time.Duration, error) {
```
jsternberg: I'm hesitant to put this in here as an API with an immediate deprecation notice, but the todo implies this function is deprecated without an actual godoc saying it is deprecated. If this is needed for a temporary implementation, would copy/paste be more efficient and less likely to result in a breaking api change?

internal/parser/parser.go
```diff
@@ -787,11 +787,9 @@ func (p *parser) parseTimeLiteral() *ast.DateTimeLiteral {
 func (p *parser) parseDurationLiteral() *ast.DurationLiteral {
 	pos, lit := p.expect(token.DURATION)
 	// todo(jsternberg): handle errors.
-	values, _ := ast.ParseDuration(lit)
-	return &ast.DurationLiteral{
-		Values:   values,
-		BaseNode: p.posRange(pos, len(lit)),
-	}
+	node, _ := ParseDuration(lit)
```
jsternberg: I would prefer at the minimum that the duration parser in the internal api not change. It should return `[]ast.Duration` as that's what it parses. I know this implementation works, but I don't see a reason why this needs  to change to be different from the others and I prefer the old pattern since it makes type safety more useful by not partially creating a value.

parser/strconv.go
```diff
@@ -0,0 +1,22 @@
+package parser
+
+import (
+	"time"
+
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/internal/parser"
+)
+
+func ParseTime(lit string) (time.Time, error) {
```
jsternberg: I don't like the stutter of `parser.ParseTime` from an API usage point of view.

nathanielc: Should we rename all the parse methods in the parser package to not include the stutter? For example `parser.Dir` vs `parser.ParseDir` and `parser.Source` vs `parser.ParseSource`?

internal/parser/parser.go
nathanielc: As I understand it the concern is that returning ast.DurationLiteral leaves the position information incomplete. That makes sense. What about the `Source` field of the ast.SourceLocation, I would expect this method to populate that, or should we break that into two separate methods? One that parser the values and the other that creates the ast.Node with the SourceLocation.Source populated?

parser/strconv.go
jsternberg: The weird part is it doesn't really work in my opinion. I don't think anybody would  say `parser source`, but they would say `parse source`. If we wanted to fully remove the stutter, we could rename the package? None of the parsing logic is in here anyway.

I do think that mixing the parsing of literals into this package would  be confusing. As an example, the `go/parser` package is focused only on parsing go files and doesn't have parsing individual portions such as literals. The reason why I placed these functions into `ast` originally is because they were more about constructing an AST rather than directly related to the parser.

ast/ast.go
nathanielc: The TODO existed in the previous code, wether the code lives here or elsewhere it needs to be refactored in the future. Having only one copy means we can refactor all its uses at once.

internal/parser/parser.go
jsternberg: The `Source` field is part of the position information and gets filled in the same way. That's how it works currently.

ast/ast.go
jsternberg: My main point was that previously it was on an unexposed function. But, I think that it is fine. I think that this is the meat of the PR anyway. I would like this to have a godoc that notes that the `time.Duration` is an approximation and does not handle timezones so that a person knows that this conversion is lossy.

internal/parser/parser.go
nathanielc: But if these functions do not return an ast.Node then the caller has to know how to fill in the source information. I think a godoc stating that only the source is filled in while the rest of the position info is left blank is fine. It meets the current needs and is clear how it behaves.

parser/strconv.go
docmerlin: I think parsing literals should be separate from the AST, as the literal is defined in context of the language not in context of the AST.  Someone can define a different syntax with different literals that creates the same AST.

nathanielc: These functions are parsing related. If we want to create separate packages for parsing different kinds of things (files vs literals ) we can but I do not see the value currently. The action of parsing is to construct ast objects from source text. 

As for the stutter I am not worried about it as the method names are clear.

jsternberg: If you want to move the parsing functions to the package it's fine. I still have my preference, but it doesn't really matter.

I do wish for these methods to return the raw values instead of the literal though. So `ParseString` should return a `string` and not a `StringLiteral`.

internal/parser/parser.go
jsternberg: Ah, I didn't understand what you meant. The source is filled in directly from a byte slice anyway  so the two are tied to each other. In the current implementation, the source attribute is not filled in by the input to this method.

I think it gets confusing to separate these two and that's not how the current implementation does it so I would prefer that remain unchanged.

nathanielc: Yeah, I think the current implementation in this PR is in an odd middle ground. It should either return the raw values or return ast.Node with the Source data filled in.

I am fine with either.

jsternberg:
I'm in agreement on 1 and 2, but I consider the source attribute to be position information and part of the parser's responsibility. I think the functions exposed in parser should just invoke the ones in internal directly.

nathanielc:
@jsternberg If the exposed parser function are just a pass through to the internal ones, then who is responsible for setting the source information? My understanding is that @docmerlin needs to use ast.Format to print out the literal values as they were typed. As such the Source location field needs to be filled. So who should fill it? If the internal methods return raw values they can't fill it, so it made sense to me to have the exposed methods handle that case. Also there is a nice consistency this way in that each method in the public parser package all return ast.Node instances.

jsternberg:
I believe the caller should fill it in personally. Let me think about it a little bit. I don't personally like having the source information partially filled, but maybe it is worth it for this specific scenario.

jsternberg:
I forgot to respond. I think I would still prefer what I wrote above, but I also don't think it matters if the public interface returns a literal so we can just go with that.

parser/strconv.go
```diff
@@ -0,0 +1,34 @@
+package parser
```
nathanielc: All Parse* methods in this package should return ast literal nodes. This remains consistent with all other Parse methods already in this package.

docmerlin: Made the change, just need the thumbs up now.

PR #954, opened 2019-02-06
feat(stdlib/universe): add contains function to check for membership …

aanthony1243:
…in lists

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


docs/SPEC.md
```diff
@@ -3042,6 +3042,19 @@ Example:
...
 
+#### Contains 
+
+Tests whether a value is a member of a set.  
+
+Contains has the following parameters: 
+
+| Name    | Type                                          | Description                  |
+| ----    | ----                                          | -----------                  |
+| value   | bool, int, uint, float, string, time          | The value to search for.     |
```
nathanielc: Does it work only for these types? Or would it work for any type Flux can express that has an Equal behavior? We haven't really specified equality of flux values though so maybe this is fine for now?

stdlib/universe/contains.go
```diff
@@ -0,0 +1,104 @@
...
+type ContainsFunc struct{}
+
+func (c *ContainsFunc) Type() semantic.Type {
+	return semantic.Invalid
+}
+func (c *ContainsFunc) PolyType() semantic.PolyType {
+	return semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"value": semantic.Tvar(1),
+			"set":   semantic.NewArrayPolyType(semantic.Tvar(2)),
```
nathanielc: These should both reference Tvar(1), that way the value will always have to be the same types as the elements in the array.

stdlib/universe/contains.go
```diff
@@ -0,0 +1,104 @@
...
+				}
+			}
+		}
+	}
+
+	return values.NewBool(found), err
+}
+
+func init() {
+	flux.RegisterPackageValue("universe", "contains", &ContainsFunc{})
```
nathanielc: Can you use `values.NewFunction` instead? Technically `values.NewFunction` require monotypes but that is only because it hasn't yet been updated. It should be straightforward to update it to use poly types.

docs/SPEC.md
aanthony1243: I'm not sure that I actually restrict the type in the code...I use val.Equal(Value)  so it might work?  but I only tested for those types. maybe over-deliver and under-promise?

stdlib/universe/contains.go
aanthony1243: oh, that's cool.  I should have noticed that.

PR #959, opened 2019-02-07
feat(ast/edit): ast match

affo:
ast.Match allows to get nodes from an ast from a pattern.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/ast/edit/match.go
+++ b/ast/edit/match.go
@@ -1,0 +1,483 @@
+package edit
+
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/ast"
+)
+
+// Match takes an AST and a pattern and returns the nodes of the AST that
+// match the given pattern.
+// The pattern is an AST in turn, but it is partially specified, i.e. nil nodes are ignored
+// while matching.
+// In the case of slices (e.g. ast.File.Body, ast.CallExpression.Arguments) the matching
+// mode can be "exact" or "fuzzy".
+// Let ps be a slice of nodes in the pattern and ns in node.
+// In "exact" mode, slices match iff len(ps) == len(ns) and every non-nil node ps[i] matches ns[i].
+// In "fuzzy" mode, slices match iff ns is a superset of ps; e.g. if ps is empty, it matches every slice.
+func Match(node ast.Node, pattern ast.Node, matchSlicesFuzzy bool) []ast.Node {
+	var sms sliceMatchingStrategy
+	if matchSlicesFuzzy {
+		sms = &fuzzyMatchingStrategy{}
+	} else {
+		sms = &exactMatchingStrategy{}
+	}
+	mv := &matchVisitor{
+		matched: make([]ast.Node, 0),
+		pattern: pattern,
+		sms:     sms,
+	}
+	ast.Walk(mv, node)
+	return mv.matched
+}
+
+type matchVisitor struct {
+	matched []ast.Node
+	pattern ast.Node
+	sms     sliceMatchingStrategy
+}
+
+func (mv *matchVisitor) Visit(node ast.Node) ast.Visitor {
+	if match(mv.pattern, node, mv.sms) {
+		mv.matched = append(mv.matched, node)
+	}
+	return mv
+}
+
+func (mv *matchVisitor) Done(node ast.Node) {}
+
+type sliceMatchingStrategy interface {
+	matchFiles(p, n []*ast.File) bool
+	matchImportDeclarations(p, n []*ast.ImportDeclaration) bool
+	matchStatements(p, n []ast.Statement) bool
+	matchProperties(p, n []*ast.Property) bool
+	matchExpressions(p, n []ast.Expression) bool
+}
+
+type fuzzyMatchingStrategy struct{}
+
+func (fms *fuzzyMatchingStrategy) matchFiles(p, n []*ast.File) bool {
+	// if the pattern slice is bigger than the node slice it can't be a subset
+	if len(p) > len(n) {
+		return false
+	}
+	// check if every node in p is also in n, but do not use the same
+	// node in n twice
+	used := make(map[int]bool)
+	for i := 0; i < len(p); i++ {
+		var matched bool
+		for ii := 0; ii < len(n) && !matched; ii++ {
+			if !used[ii] && match(p[i], n[ii], fms) {
+				used[ii] = true
+				matched = true
+			}
+		}
+		if !matched {
+			// found at least one node in the p that is not in n
+			return false
+		}
+	}
+	return true
+}
+
+func (fms *fuzzyMatchingStrategy) matchImportDeclarations(p, n []*ast.ImportDeclaration) bool {
+	if len(p) > len(n) {
+		return false
+	}
+	used := make(map[int]bool)
+	for i := 0; i < len(p); i++ {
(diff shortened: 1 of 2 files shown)


ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	if len(p.Files) > len(op.Files) {
+		return false
+	}
+	if len(p.Files) == 0 {
+		return true
+	}
+	var matched bool
+	for i := 0; i < len(op.Files) && !matched; i++ {
+		if len(p.Files) > len(op.Files)-i-1 {
```
wolffcm: I don't understand this code... if `p.Files` and `op.Files` are the same length, won't this function return `false`?

ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+	return true
+}
+
+func matchLogicalOperator(o, oo ast.LogicalOperatorKind) bool {
+	if o >= 0 && o != oo {
+		return false
+	}
+	return true
+}
+
```
wolffcm: Can you add a comment here describing the behavior of this function?  Is it okay for there to be extra files in `op` that do not exist in `p`?

ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+		for ii := 0; i < len(f.Imports) && !matched; ii++ {
+			if !match(f.Imports[ii], of.Imports[i+ii]) {
+				break
+			}
+			if ii == len(f.Imports)-1 {
+				matched = true
+			}
+		}
+	}
+	return matched
```
wolffcm: Do you not recurse into the `Body` statements here?

It seems like the matcher will most often be used at the level of the statements in `File.Body` and below... Do we need to be able to match the structure of a package with files, imports, etc?  It seems maybe of limited use.

ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/ast"
+)
+
+// Match takes an AST and a pattern and returns the nodes of the AST that
+// match the given pattern.
+// The pattern is an AST in turn, but it is partially specified: nil nodes are ignored
+// while matching, and slices are checked for the first matched sub-sequence.
```
wolffcm: Can you give an example of how slice matching works?

ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+
+func matchCallExpression(e *ast.CallExpression, oe *ast.CallExpression) bool {
+	if !match(e.Callee, oe.Callee) {
+		return false
+	}
+	if len(e.Arguments) > len(oe.Arguments) {
+		return false
+	}
+	if len(e.Arguments) == 0 {
+		return true
```
wolffcm: So I guess there's no way to express that you want to match a function that has exactly 0 arguments?  You could check for a `nil` slice?

ast/edit/match.go
```diff
@@ -0,0 +1,436 @@
...
+
+func matchVariableAssignment(a *ast.VariableAssignment, oa *ast.VariableAssignment) bool {
+	return match(a.ID, oa.ID) && match(a.Init, oa.Init)
+}
+
+func matchMemberAssignment(a *ast.MemberAssignment, oa *ast.MemberAssignment) bool {
+	return match(a.Member, oa.Member) && match(a.Init, oa.Init)
+}
+
+func matchCallExpression(e *ast.CallExpression, oe *ast.CallExpression) bool {
```
wolffcm: I like the convention of using `p` to indicate a pattern in this file, would be good to continue that here.

ast/edit/match_test.go
```diff
@@ -0,0 +1,419 @@
...
+								Properties: []*ast.Property{
+									{Key: &ast.Identifier{Name: "a"}, Value: &ast.IntegerLiteral{Value: 2}},
+									{Key: &ast.Identifier{Name: "b"}, Value: &ast.IntegerLiteral{Value: 3}},
+								},
+							}},
+						},
+					},
+				},
+			}}},
+			match: &ast.File{
```
wolffcm: I must be missing something, I don't see how `matchFile` matches anything in the body of the file...

ast/edit/match.go
affo: Yeah, I think you are right. That `-1` should not be there 😅

ast/edit/match.go
affo: Absolutely, I totally missed it.

ast/edit/match.go
affo: It was more `statement`, `other-statement`. I could change everything to `p`attern, `n`ode.

ast/edit/match.go
affo: Yeah, you are right. The same holds for every other array around here.

ast/edit/match.go
affo: I wanted to change slice matching behavior, actually:

```
Two slices match if every node provided in the pattern exists in the slice subject to comparison.
```

I think it is what you really want when thinking of "matching" and the code for the comparison would also be more readable. What do you think?

wolffcm: Maybe.  An empty slice in a pattern would match anything, which could be the opposite of what you want.  I suppose you could use the following:
- pattern is `nil` slice, matches anything
- pattern is empty slice, matches only an empty slice
- pattern is non-empty slice, matches slices that are a superset of pattern

This is still inconsistent though.  I'm not sure the best way to handle this issue,

ast/edit/match.go
wolffcm: I like that.  One of the troubles with a tool like this is that the pattern and the thing you're matching are the same data type so it's easy to get confused.

ast/edit/match.go
```diff
@@ -0,0 +1,483 @@
...
+
+type sliceMatchingStrategy interface {
+	matchFiles(p, n []*ast.File) bool
+	matchImportDeclarations(p, n []*ast.ImportDeclaration) bool
+	matchStatements(p, n []ast.Statement) bool
+	matchProperties(p, n []*ast.Property) bool
+	matchExpressions(p, n []ast.Expression) bool
+}
+
+type fuzzyMatchingStrategy struct{}
```
affo: For slice matching strategies I merely copy pasted the code. I don't want to copy the slice to get a `[]ast.Node`. I thought about code generation, but I didn't want to lose too much time on it. Is this the way to go?

ast/edit/match.go
affo: Fixed, files must be match

affo:
@wolffcm after discussion I introduced some changes:

Two strategies for slice comparison, fuzzy and exact. I also integrated your comments. The body of a File is now matched and I opted for the convention `p`, `n` for `p`attern, `n`ode.

ast/edit/match.go
wolffcm: I think this is fine for now.  You have what you need for the test refactor tool.

Moving forward, maybe there will be a `Matcher` object, and you can have an `Options` class that uses varargs similar to how the Planner handles options:
https://github.com/influxdata/flux/blob/master/plan/logical.go#L24
This way the API doesn't need to change when new options are added (for example, if there is an additional predicate that must pass for matches).

PR #960, opened 2019-02-07
Feat/testgen

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR:
1. Adds the new test key word for identifying a test case.
2. Refactors testing.run
3. Updates existing e2e tests to use new test case format with new testing.run.

Open question: Is the `test` key word redundant? Would it be better to just prefix test cases with `test_`?


stdlib/testing/testdata/aggregate_empty_window.flux
```diff
@@ -41,18 +43,15 @@ outData = "
...
+	(table
+		|> range(start: 2018-05-22T19:53:26Z, stop: 2018-05-22T19:55:00Z)
+		|> aggregateWindow(every: 30s, fn: sum))
+mem = testing.loadMem
+store = testing.loadStorage
+want = mem(csv: outData)
+got = store(csv: inData)
+	|> t_window_generate_empty()
+
+testing.assertEquals(name: "aggregate_empty_window", want: want, got: got)
```
aanthony1243: why aren't these using the new testing framework?

stdlib/testing/testdata/aggregate_window.flux
```diff
@@ -31,14 +33,12 @@ outData = "
...
-
 aggregate_window = (table=<-) =>
-  table
-  |> range(start: 2018-05-22T00:00:00Z, stop:2018-05-22T00:01:00Z)
-  |> aggregateWindow(every:30s,fn:mean)
+	(table
+		|> range(start: 2018-05-22T00:00:00Z, stop: 2018-05-22T00:01:00Z)
+		|> aggregateWindow(every: 30s, fn: mean))
+
+test test_aggregate_window = () =>
```
aanthony1243: I'm ok with this keyword, but not the stutter.  `test aggregate_window = () => `  sounds better.  Many of the existing tests have a `t_` prefix to avoid naming collisions on function calls.

stdlib/testing/testdata/cov.flux
```diff
@@ -32,23 +35,23 @@ outData = "
...
+		|> filter(fn: (r) =>
+			(r.user == "user2"))
+		|> group(columns: ["_measurement"])
+	got = cov(x: left, y: right, on: ["_time", "_measurement"])
+	want = testing.loadStorage(csv: outData)
 
-    got = cov(x:left, y:right, on: ["_time", "_measurement"])
-    want = testing.loadStorage(csv: outData)
-    return testing.assertEquals(name: "cov", want: want, got: got)
+	return testing.assertEquals(name: "cov", want: want, got: got)
```
aanthony1243: we can probably upgrade this to the new framework too.  the old one didn't play well with multiple sources.

stdlib/testing/testing.flux
```diff
@@ -9,7 +9,17 @@ builtin diff
...
+    tc = case()
+    got = tc.input |> tc.fn()
+    dif = got |> diff(want: tc.want)
+    pass = dif |> yield(name: "diff") |> assertEmpty()
+    return {
+        input: tc.input,
+        want: tc.want,
+        fn: tc.fn,
+        got: got,
+        diff: diff,
```
aanthony1243: `diff:dif`?

stdlib/testing/testing.flux
```diff
@@ -9,7 +9,17 @@ builtin diff
...
 loadMem = (csv) => c.from(csv: csv)
 
-test = (name, input, want, testFn) => {
-    got = input |> testFn()
-    return got |> diff(want: want) |> yield(name: name) |> assertEmpty()
+run = (case) => {
+    tc = case()
+    got = tc.input |> tc.fn()
+    dif = got |> diff(want: tc.want)
+    pass = dif |> yield(name: "diff") |> assertEmpty()
```
aanthony1243: do we need to yield here, if we're also returning it in the result?

stdlib/testing/testdata/yield.flux
```diff
@@ -29,20 +32,23 @@ outData = "
...
+
 testing.assertEquals(name: "yield", want: want, got: got)
-testing.test(name: "yield",
-            input: testing.loadStorage(csv: inData),
-            want: testing.loadMem(csv: outData),
-            testFn: t_yield)
+
+test test_yield = () =>
+	({input: testing.loadStorage(csv: inData), want: testing.loadMem(csv: outData), fn: t_yield})
+
```
aanthony1243: this test is disabled because we don't have a good way to inspect yield() text anymore, since we quit doing byte-for-byte text comparison.  any ideas?

stdlib/testing/testdata/aggregate_empty_window.flux
jlapacik: The script skipped these. I'll update.

stdlib/testing/testdata/aggregate_window.flux
jlapacik: I guess I'll have to go 1 by 1 and update the test names so that they're descriptive but do not conflict with any function names.

stdlib/testing/testing.flux
jlapacik: Oops.

stdlib/testing/testing.flux
jlapacik: I think we do need it because if assertEmpty fails(outside of the repl) then yield is the only way to get to the diff.

stdlib/testing/testdata/columns.flux
```diff
@@ -1,6 +1,8 @@
+package main
+ 
 import "testing"
 
-option now = () => 2030-01-01T00:00:00Z
+option now = () => (2030-01-01T00:00:00Z)
```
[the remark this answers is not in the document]
jlapacik: No not necessary. They were added by Lorenzo's script.

PR #964, opened 2019-02-11
Refactor/prepcsvtests

### LINKED ISSUE #942: Update prepcsvtest with new testing standards
With the standard test case format we can create a tool to easily update test cases.

The prepcsvtest utility does this already, we should update it and rename it to work with the new standard. 

See #935

affo:
Fixes #942 

Merged after #959 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


cmd/utils/refactortests/refactortests.go
```diff
@@ -0,0 +1,291 @@
...
+		fmt.Printf("%s\n\n", diffBuf.String())
+		return resultBuf.String(), nil
+	}
+
+	fmt.Println("want/got are equal; nothing to do.")
+	return "", nil
+}
+
+func editDataOrQuery(node *ast.Package, rawResult string) error {
+	fmt.Println("Do you want to edit the expected result (1) or the query (2)? (any other to exit)")
```
wolffcm: I would think the easiest way to handle this would be to allow the user to update either the query or the input data using an editor of their choice, and then to re-run the script and allow the user to accept whatever output it produces as expected.

affo:
@nathanielc 👍 

If the command is in `internal`, should I worry about what @wolffcm pointed out:

> Not sure about letting the user edit the script from inside the tool, seems better to let them use their own editor

?

affo:
@nathanielc @wolffcm 

I addressed Nathaniel's comments and created an issue to make test generation/refactoring available through the REPL: https://github.com/influxdata/flux/issues/968

nathanielc:
> Not sure about letting the user edit the script from inside the tool, seems better to let them use their own editor

I agree with @wolffcm on this point, as a user of this tool myself, I want to use my normal editor to modify the files.

affo:
Got it👍 I am removing that part.

affo:
@wolffcm With the first commit I fix some problems in edit.Match. There where issues with typed nils and I forgot the TestStatement in the switch. I can split the PRs if you prefer.

For test refactoring, it was pretty tough, because it is difficult to differentiate errors in the case that `assertEmpty` failed (and so we have a diff to output), or a manually written `assertEquals` fails (the case in which there is no test statement, but a hand-written test), or the query fails. However I think I handled that (`executeScript` function) 😅 

Another issue was that now tests can't simply run. They need a code generation step. I encoded that logic into the `inPlaceTestGen` method.  
Eventually, I wrote some tests,  so that they will break if the testing framework changes.

internal/cmd/refactortests/cmd/refactortests_test.go
```diff
@@ -0,0 +1,230 @@
...
+,result,table
+"
+
+test t = () => ({
+    input: testing.loadMem(csv: inData),
+    want: testing.loadMem(csv: outData),
+    fn: (table=<-) => table
+})
+
+testing.run(case: t)
```
affo: I think this line shouldn't be here. The objective was to see if a test without that line runs.

internal/cmd/refactortests/cmd/refactortests.go
```diff
@@ -0,0 +1,364 @@
...
+	err = ioutil.WriteFile(fname, []byte(ast.Format(script)), 0644)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s updated\n", fname)
+	return nil
+}
+
+func getOutDataVariable(node ast.Node) (*ast.VariableAssignment, error) {
+	fmt.Println("Name for the variable holding the expected result:")
```
wolffcm: (minor comment) You could offer the user a default of `outData` so they can just press enter here if they want.

affo:
@wolffcm 

As agreed, I did the last modifications to make the tool more usable (`outData` as default, and better behavior when reloading from disk). I also updated the test that was useless and removed `prepcsvtests`, which I forgot to do 😅 .

This should be ready to merge. Waiting for your 👍 @wolffcm

PR #967, opened 2019-02-12
feat(testing): generate go tests from flux tests

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

With this PR the builtin `generate` command now generates flux ASTs from the flux testing packages that make up the end-to-end tests defined within the stdlib. It also generates a go test driver that runs all of the end-to-end tests in the stdlib. The flux ASTs are written to a file named `flux_test_gen.go` in the same directory that the flux tests are located.


internal/cmd/builtin/cmd/generate.go
```diff
@@ -255,18 +402,66 @@ func constructValue(v reflect.Value) (jen.Code, error) {
 		typ := v.Type()
 		s := indirectType(typ)
 		values := make(jen.Dict, v.NumField())
-		for i := 0; i < v.NumField(); i++ {
-			field := v.Field(i)
-			if !field.CanInterface() {
-				// Ignore private fields
-				continue
+		if typ.Name() == "DateTimeLiteral" {
```
nathanielc: For readability I think it would help to use a switch statement here and then to put all of this logic into its own function that way we do not have so many nested scopes.

stdlib/end_to_end_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+	"time"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/querytest"
+)
+
+func runEndToEnd(t *testing.T, querier *querytest.Querier, pkgs []*ast.Package) {
```
nathanielc: Do these functions need to be exported?

stdlib/flux_test.go
```diff
@@ -0,0 +1,24 @@
+// DO NOT EDIT: This file is autogenerated via the builtin command.
```
nathanielc: Why do we need to generate this file? Does it change if any of the test cases change? While possible to generate it I wonder if it adds unnecessary complexity.  As such I think we should just commit this file as is, instead of generating it. Then we can add to it without having to update the generating code.

jlapacik: This file does change if the test cases change. In particular if a new testing package is added, the new import will be added to this file and a new call to `runEndToEnd` added. This way any modifications to the flux tests won't require changes to how they are run. That will automatically be taken care of. But I don't like complexity either, so I'm totally fine having it be a manual process for now too.

stdlib/end_to_end_test.go
jlapacik: At the moment no. Only if they are needed in influxDB.

internal/cmd/builtin/cmd/generate.go
jlapacik: I agree.

stdlib/flux_test.go
nathanielc: What if we generated a file that created a list of test packages. Then we can generate a simple file that is just a bunch of append calls. Then the test file itself is hand written to use that list.

stdlib/end_to_end_test.go
nathanielc: What is the plan for using the tests in InfluxDB? Is it to write their own runEndToEnd equivalent and just grab the list of test cases?

stdlib/flux_test.go
jlapacik: That's a lot better because then influxDB can use the same file for it's test runner.

stdlib/end_to_end_test.go
jlapacik: Yes I think so. We can generate a file with the list of imports in flux. Then influxDB just grabs that list, imports the packages, and then runs all of the tests for each import with its own equivalent of `runEndToEnd`.

stdlib/testing/testdata/aggregate_empty_window.flux
```diff
@@ -55,4 +55,3 @@ test aggregate_window_empty = () => ({
             |> aggregateWindow(every: 30s, fn: sum),
 })
 
-testing.run(case: aggregate_window_empty)
```
nathanielc: My understanding is that now the `test` keyword is an implicit call to testing.run. Where is that happening? I didn't see it in this code.

jlapacik: The call to `testing.run` is added to the end of the ast right here https://github.com/influxdata/flux/blob/017b3bd84fb1bdac2e3e8ae836fd1f1f4900bda7/internal/cmd/builtin/cmd/generate.go#L257 .

stdlib/flux_test.go
```diff
@@ -0,0 +1,95 @@
...
+func TestFluxEndToEnd(t *testing.T) {
+	runEndToEnd(t, querier, stdlib.FluxTestPackages)
+}
+func BenchmarkFluxEndToEnd(b *testing.B) {
+	benchEndToEnd(b, querier, stdlib.FluxTestPackages)
+}
+
+func runEndToEnd(t *testing.T, querier *querytest.Querier, pkgs []*ast.Package) {
+	for _, pkg := range pkgs {
+		name := pkg.Files[0].Name
```
nathanielc: Should this be the package name not the file name?

jlapacik: There can be many test files in the same test package so the file name is probably a better test name (this is the current behavior). Otherwise each test would have the same name (the package name) followed by an index number.

nathanielc: I see, I had it backwards. Because we are artificially making each test file its own package (but all with the same name), each test package only has one file, therefore the file name is the better test name.  Makes sense.

PR #979, opened 2019-02-15
bug(options): Add key instead of erroring when option doesn't have key

### LINKED ISSUE #974: Options Editor doesn't allow inserting of option keys that don't already exist in the option.
Options Editor doesn't allow inserting of option keys that don't already exist in the option.

Instead it errors with: cannot find property 'offset' in object expression

https://github.com/influxdata/influxdb/issues/11324

docmerlin:
This adds a key to options if the key doesn't already exist in the option instead of erroring.
Closes #974 
### Done checklist
- [x ] Test cases written

### DIFF
--- a/ast/edit/option_editor.go
+++ b/ast/edit/option_editor.go
@@ -53,11 +53,11 @@ func OptionObjectFn(keyMap map[string]ast.Expression) OptionFn {
 		// check that every specified property exists in the object
 		found := make(map[string]bool, len(obj.Properties))
 		for _, p := range obj.Properties {
-			found[p.Key.Key()] = false
+			found[p.Key.Key()] = true
 		}
 
 		for k := range keyMap {
-			if _, ok := found[k]; !ok {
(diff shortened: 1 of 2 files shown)


ast/edit/option_editor.go
```diff
@@ -51,14 +51,17 @@ func OptionObjectFn(keyMap map[string]ast.Expression) OptionFn {
...
+		found := make(map[string]bool, len(obj.Properties))
 		for _, p := range obj.Properties {
-			keys[p.Key.Key()] = true
+			found[p.Key.Key()] = false
 		}
 
 		for k := range keyMap {
-			if !keys[k] {
-				return nil, fmt.Errorf("cannot find property '%s' in object expression", k)
+			if _, ok := found[k]; !ok {
```
nathanielc: Nit, this should be simplified to `if !found[k] {`

docmerlin:
Thanks @affo, doh

PR #981, opened 2019-02-15
refactor(testing): use testing.run and testing.inspect as separate phases

nathanielc:
This change allows for ensuring that the reported diff is complete,
because a query that errors is not guaranteed to return all results
before the error is encountered.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/testing/testing.flux
+++ b/stdlib/testing/testing.flux
@@ -9,27 +9,20 @@ builtin diff
 option loadStorage = (csv) => c.from(csv: csv)
 option loadMem = (csv) => c.from(csv: csv)
 
-run = (case) => {
-    tc = case()
-    return tc.input
-        |> tc.fn()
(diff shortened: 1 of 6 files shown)


stdlib/testing/testing.flux
```diff
@@ -9,27 +9,20 @@ builtin diff
...
 option loadMem = (csv) => c.from(csv: csv)
 
-run = (case) => {
-    tc = case()
-    return tc.input
-        |> tc.fn()
-        |> yield(name: "_test_result")
-        |> diff(want: tc.want)
-        |> yield(name: "diff")
-        |> assertEmpty()
```
jlapacik: So previously assertEmpty could fail before all tables were processed in which case the diff will be incomplete?

PR #988, opened 2019-02-20
refactor(lang): add extern to ast compiler

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -68,3 +68,8 @@ func (c ASTCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
 func (ASTCompiler) CompilerType() flux.CompilerType {
 	return ASTCompilerType
 }
+
+// PrependFile prepends a file onto the compiler's list of package files.
+func (c *ASTCompiler) PrependFile(file *ast.File) {
+	c.AST.Files = append([]*ast.File{file}, c.AST.Files...)
+}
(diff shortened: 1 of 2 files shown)


lang/compiler.go
```diff
@@ -54,11 +54,15 @@ func (c SpecCompiler) CompilerType() flux.CompilerType {
...
 // ASTCompiler implements Compiler by producing a Spec from an AST.
 type ASTCompiler struct {
+	Ext *ast.File    `json:"extern"`
 	AST *ast.Package `json:"ast"`
 	Now func() time.Time
 }
 
 func (c ASTCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
+	if c.Ext != nil {
+		c.AST.Files = append([]*ast.File{c.Ext}, c.AST.Files...)
```
nathanielc: This seems like the wrong design. Everytime `Compile` is called the file will be prepended to the package. We should not be modifying the AST from within the compiler.

lang/compiler.go
```diff
@@ -68,3 +68,7 @@ func (c ASTCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
 func (ASTCompiler) CompilerType() flux.CompilerType {
 	return ASTCompilerType
 }
+
+func (c *ASTCompiler) PrependFile(file *ast.File) {
```
nathanielc: Can you add a doc comment here? Otherwise this looks great.

PR #996, opened 2019-02-22
fix(stdlib/testing): add locks to make diff threadsafe

aanthony1243:


aanthony1243:
yeah, and Union too.  I'm not sure how critical it is, but it sometimes will trip up `go test -race`

PR #998, opened 2019-02-22
refactor(testing): remove multiple yields from diff

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

The diff property from `testing.inspect` now only yields a single result `result: "diff"`. Previously it yielded the result of the query as well as the diff between actual and expected result. Now it is much easier to read in the flux repl.

**UPDATE:** `csv.from` produces tables that wrap around a csv reader. As a result, these tables cannot be shared among goroutines. Now `csv.from` creates a deep copy of each table before sending it to one of its downstream transformations. As a side effect, all end-to-end tests that needed to use `assertEquals` have been updated to the standard test case format.

### DIFF
--- a/stdlib/testing/testdata/yield.flux
+++ b/stdlib/testing/testdata/yield.flux
@@ -33,18 +33,13 @@ outData = "
 ,,0,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:54:16Z,87.88598574821853,usage_idle,cpu,cpu-total,host.local
 "
 
-t_yield = (table=<-) =>
-	(table
-		|> sort()
-		|> limit(n: 3)
-		|> yield(name: "1: lowest 3")
-		|> sample(n: 2, pos: 1)
-		|> yield(name: "2: 2nd row"))
-indata = testing.loadStorage(csv: inData)
-got = indata
-	|> t_yield()
-	|> yield(name: "5")
-want = testing.loadStorage(csv: outData)
-	|> yield(name: "6")
+t_yield = (table=<-) => table
+    |> sort()
+    |> limit(n: 3)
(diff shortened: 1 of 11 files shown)


stdlib/testing/testing.flux
```diff
@@ -11,14 +11,14 @@ option loadMem = (csv) => c.from(csv: csv)
...
+    got = tc.input |> tc.fn()
+    dif = got |> diff(want: tc.want)
     return {
         fn:    tc.fn,
         input: tc.input
         want:  tc.want,
-        got:   got,
-        diff:  dif,
+        got:   got |> yield(name: "_test_result"),
+        diff:  dif |> yield(name: "diff"),
```
nathanielc: This change looks like a noop. Why does moving the yields down to these lines change anything?

jlapacik: `_test_result` is no longer returned as part of dif. Only the diff tables are returned.

nathanielc: ah that makes sense

stdlib/testing/testdata/yield.flux
```diff
@@ -33,18 +33,13 @@ outData = "
...
+t_yield = (table=<-) => table
+    |> sort()
+    |> limit(n: 3)
+    |> yield(name: "1: lowest 3")
+    |> sample(n: 2, pos: 1)
+    |> yield(name: "2: 2nd row")
+    |> yield(name: "5")
 
-testing.assertEquals(name: "yield", want: want, got: got)
+test _yield = () =>
```
aanthony1243: this test is disabled, but with our current test framework, we might be able to make this work.  The biggest issue is processing the results when they may be in an arbitrary order.  I think the test framework will currently only be fully guaranteed for single result queries.

PR #1003, opened 2019-02-25
feat(control): support attaching arbitrary query metadata from the executor

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

A source node can expose statistics for its execution and the executor
will attach this metadata to the query returned from the controller.
This metadata can be any arbitrary set of key/value pairs. The metadata
is formatted similar to http headers where one key may have multiple
values. This means that a consumer can decide what to do with multiple
values with the same key such as whether to average or sum the values.

This removes the `flux.Statistics` from arbitrary `flux.Result` values
and the table iterators because the statistics are mostly meaningless
for most iterators.

BREAKING CHANGE: The `Statistics() flux.Statistics` method has been
removed from all interfaces not related to `flux.Query` or
`flux.ResultIterator`. The `flux.ResultIterator` should be considered a
wrapper of `flux.Query` so it exposes the statistics too.

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -162,6 +162,7 @@ func (c *Controller) createQuery(ctx context.Context, ct flux.CompilerType) *Que
 		c:                  c,
 		now:                time.Now().UTC(),
 		ready:              make(chan map[string]flux.Result, 1),
+		metaCh:             noMetadata, // This will be set to a non-closed channel upon successful execution.
 		parentCtx:          parentCtx,
 		parentSpan:         parentSpan,
 		cancel:             cancel,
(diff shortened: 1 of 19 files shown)


control/controller.go
```diff
@@ -632,8 +642,14 @@ TRANSITION:
 	}
 	q.currentSpan, q.currentCtx = nil, nil
 
-	// If we are transitioning to a finished state from a non-finished state, finish the parent span.
 	if isFinishedState(newState) {
+		// Invoke the cancel function to ensure that we have signaled that the query should be done.
+		// The user is supposed to read the entirety of the tables returned before we end up in a finished
+		// state, but user error may have caused this not to happen so there's no harm to canceling multiple
+		// times.
+		q.cancel()
```
jlapacik: Was this a previous bug in the controller?

jsternberg: It might have been? I don't think it could have deadlocked previously. Now, the `Done()` method could get blocked on the `es.do()` invocation since the metadata channel needs to be closed.

PR #1007, opened 2019-02-27
Refactor/schema mutators

jlapacik:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

This PR refactors all of the schema mutation functions (drop, keep, rename, duplicate) to not error when consuming tables with missing columns.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2088,7 +2088,6 @@ Rename renames specified columns in a table.
 There are two variants: one which takes a map of old column names to new column names,
 and one which takes a mapping function.
 If a column is renamed and is part of the group key, the column name in the group key will be updated.
-If a specified column is not present in a table an error will be thrown.
 
 Rename has the following properties: 
 
@@ -2117,9 +2116,9 @@ from(bucket: "telegraf/autogen")
 
 #### Drop 
 
-Drop excludes specified columns from a table. Columns to exclude can be specified either through a list, or a predicate function.
+Drop excludes specified columns from a table.
+Columns to exclude can be specified either through a list, or a predicate function.
 When a dropped column is part of the group key it will also be dropped from the key.
-If a specified column is not present in a table an error will be thrown.
 
 Drop has the following properties:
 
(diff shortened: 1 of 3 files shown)


docs/SPEC.md
```diff
@@ -2181,7 +2179,6 @@ from(bucket: "telegraf/autogen")
 #### Duplicate 
 
 Duplicate duplicates a specified column in a table.
-If the specified column is not present in a table an error will be thrown.
 If the specified column is part of the group key, it will be duplicated, but it will not be part of the group key of the output table.
```
affo: did the spec of rename said anything about non-existing columns?

docs/SPEC.md
```diff
@@ -2181,7 +2179,6 @@ from(bucket: "telegraf/autogen")
 #### Duplicate 
 
 Duplicate duplicates a specified column in a table.
```
aanthony1243: I guess this is consistent but it feels wrong.  the 'duplicate' of nothing is nothing...  and yet I feel like if I'm asking to duplicate something I should probably be told if that thing is not there.  

I would presume that it's OK to quietly move on here, since they'll likely hit the error on some subsequent code in the next step.  Unless they wrote back to storage...

jlapacik: I removed the checks in duplicate and rename with the optimizer in mind. My think was that if one of the schema functions errors on missing columns then that reduces the space of optimizations we can make.

The example I'm thinking of is `schema_function -> filter`. If `schema_function` errors on missing columns then we can't push `filter` below it, because filter could potentially remove the table or tables that would have caused `schema_function` to error. The same holds for the reverse case. However the operations do commute if we remove the column checks.

jlapacik:
I'm closing this PR and leaving the semantics as they are right now. By default keep will error on a missing column. However, if a user would like to skip missing columns, they can use the `fn` parameter which allows specifying a regex for column matching. This leaves the user with the choice to use whichever options suites their needs best.

PR #1025, opened 2019-03-05
fix(execute): group key comparison works regardless of column ordering

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Group keys are considered identical if they have the same set of columns
and these columns have the same values. They are also sorted by a
similar metric where any missing column is considered null and nulls
sort before anything.

Previously, the sort algorithm that was employed assumed that the
columns were already sorted which only worked if the group key was
constructed that way. Otherwise, it would compare them as different.

The group key maintains its ordering for the purpose of a call to
`Cols()` so it is order preserving, but they are considered identical as
if to compare the values in a map so order doesn't matter.

Fixes #1023.

PR #1026, opened 2019-03-06
fix(stdlib): deterministic sorting of input tables in join

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Previously join would sort the input tables according to the join
key (columns on which to join). But the join key was stored as a
map, so it was possible to get different sort orders for the rows
of a table.

Now the join key is stored as a column slice, hence for each table
output by the join, the rows of that table will have a deterministic
sort order.

This combined with a similar bug in the table
sort that caused it to use the first row for sorting if the column could
not be found.

Signed-off-by: Jonathan A. Sternberg <jonathan@influxdata.com>

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -1245,11 +1245,11 @@ func (b *ColListTableBuilder) ClearData() {
 }
 
 func (b *ColListTableBuilder) Sort(cols []string, desc bool) {
-	colIdxs := make([]int, len(cols))
-	for i, label := range cols {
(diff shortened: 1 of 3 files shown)


jsternberg:
This seems to require the group key changes to work? I'll close this and cherry pick it to that commit and we'll have to do both at the same time.

jlapacik:
I'm closing this PR as this commit was added as part of #1025 .

Issue #1021, opened 2019-03-04
Audit and fix transformations that do not track their memory allocations


nathanielc:

jsternberg:
Contrary to my expectations, it does not appear like arrow stuff is untracked in a lot of places. I'll continue looking and auditing what's going on, but I may need access to the real queries that are causing this problem.

nathanielc:
The kinds of queries that use lots of memory are windowing queries. A basic:

```
from()
 |> range()
 |>  filter()
 |> window()
```

query was causing the OOMs. The more windows are created the more memory was used. For example if the series cardinality is 1000, and the time range was 1h and the windows were 1m, then the total groups before the window is 1000 and after the window is 60000. All of the groups needed to be held in memory.  It should be straigtforward to create a windowing query that creates a lot of windows. Just use range -1h and window 1s or something similar.

jsternberg:
Ok. I'll give that type of query a try locally and see if I can reproduce it.

jsternberg:
I think memory allocations are being properly tracked. We will still have to do more work to ensure that allocations within each transformation (for local state) are being tracked, but that may be outside of the scope of this issue or may require this issue to be rescoped because I would rank that as a high difficulty that needs to be broken down further.

jsternberg:
Closing this for now since work was done on this and we can open a new issue or reopen this one if the above has to be addressed.

PR #1029, opened 2019-03-06
fix(stdlib): diff should track memory allocations when it copies the table

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Related to #1021.

### DIFF
--- a/stdlib/testing/diff.go
+++ b/stdlib/testing/diff.go
@@ -109,6 +109,7 @@ type DiffTransformation struct {
 
 	d     execute.Dataset
 	cache execute.TableBuilderCache
+	alloc *memory.Allocator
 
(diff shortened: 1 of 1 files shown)


stdlib/testing/diff.go
```diff
@@ -109,6 +109,7 @@ type DiffTransformation struct {
 
 	d     execute.Dataset
 	cache execute.TableBuilderCache
+	alloc *memory.Allocator
```
nathanielc: Where does the `alloc` field get set?

nathanielc: nevermind I see it now, somehow I missed on first review

jsternberg: It gets set in `NewDiffTransformation`.

PR #1034, opened 2019-03-07
feat(stdlib): added a math package and ported all 64 bit go math libr…

aanthony1243:
…ary functions

### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written


stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+builtin y1
+builtin yn
+
+// Table transformations for trig functions
+sinValue = (tables=<-) => tables |> map(fn:(r) => sin(x:r._value))
+cosValue = (tables=<-) => tables |> map(fn:(r) => cos(x:r._value))
+tanValue = (tables=<-) => tables |> map(fn:(r) => tan(x:r._value))
+asinValue = (tables=<-) => tables |> map(fn:(r) => asin(x:r._value))
+acosValue = (tables=<-) => tables |> map(fn:(r) => acos(x:r._value))
+atanValue = (tables=<-) => tables |> map(fn:(r) => atan(x:r._value))
```
nathanielc: I don't think we want these functions, they provide little value in my opinion. Thoughts?

stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+builtin exp2
+builtin expm1
+builtin float64bits
+builtin float64frombits
+builtin floor
+builtin frexp
+builtin gamma
+builtin hypot
+builtin ilogb
+builtin m_inf
```
nathanielc: is there a name we can use that doesn't use and `_`?

stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+builtin ln2
+builtin log2e
+builtin ln10
+builtin log10e
+builtin maxfloat
+builtin maxint
+builtin minint
+builtin maxuint
+
+//missing builtin values [y1 sqrte roundtoeven nextafter32 sqrtphi sqrtpi copysign ceil], extra builtin values [Ceil sqrtPhi sqrtPi sqrtE roundToEven copySign]
```
nathanielc: can this comment be removed now?

stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+builtin phi
+builtin sqrt2
+builtin sqrte
+builtin sqrtpi
+builtin sqrtphi
+builtin ln2
+builtin log2e
+builtin ln10
+builtin log10e
+builtin maxfloat
```
nathanielc: what about smallest float?

stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+
+
+
+
+
+
+
+
+
+
```
nathanielc: nit, can you remove all these blank lines?

stdlib/math/math.go
```diff
@@ -0,0 +1,507 @@
...
+	flux.RegisterPackageValue("math", "sqrt", generateMathFunctionX("sqrt", math.Sqrt))
+	flux.RegisterPackageValue("math", "tan", generateMathFunctionX("tan", math.Tan))
+	flux.RegisterPackageValue("math", "tanh", generateMathFunctionX("tanh", math.Tanh))
+	flux.RegisterPackageValue("math", "trunc", generateMathFunctionX("trunc", math.Trunc))
+	flux.RegisterPackageValue("math", "y0", generateMathFunctionX("y0", math.Y0))
+	flux.RegisterPackageValue("math", "y1", generateMathFunctionX("y1", math.Y1))
+
+	SpecialFns = map[string]values.Function{
+		// float --> uint
+		"float64bits": values.NewFunction(
```
affo: Why isn't this directly used as a type conversion? Is it different?

In `typeconv` we do:

```
uint64(v.Float())
```

To convert a float64 to uint64.

stdlib/math/math.flux
```diff
@@ -0,0 +1,198 @@
...
+// Table transformations for trig functions
+sinValue = (tables=<-) => tables |> map(fn:(r) => sin(x:r._value))
+cosValue = (tables=<-) => tables |> map(fn:(r) => cos(x:r._value))
+tanValue = (tables=<-) => tables |> map(fn:(r) => tan(x:r._value))
+asinValue = (tables=<-) => tables |> map(fn:(r) => asin(x:r._value))
+acosValue = (tables=<-) => tables |> map(fn:(r) => acos(x:r._value))
+atanValue = (tables=<-) => tables |> map(fn:(r) => atan(x:r._value))
+
+// hack to simulate an imported math package
```
affo: why do we need this?

stdlib/math/math.flux
affo: Yes, I agree. Are these function here because it is supposed that trig functions are commonly used like this?  
Maybe we could provide a function generator in the prelude:

```
valueMapperFor = (func) => {
    return (tables=<-) => tables |> map(fn:(r) => func(r._value))
}
```

To avoid verbose syntax to users when they want to write those functions. However, it would require positional arguments...

stdlib/math/math.flux
affo: I mean, why putting both in the prelude and making it available as an imported package?

stdlib/math/math.flux
aanthony1243: yeah, they should not be permanent.  I have them in the demo to show that it's something you can do, will remove before I merge.

stdlib/math/math.flux
aanthony1243: sure.  it conflicts with the global namespace and we can rename it to `inf` when we resolve the scope correctly for `map` and `filter`

stdlib/math/math.go
aanthony1243: this returns the underlying binary representation of the float.  float64bits(x) != uint64(x)

stdlib/math/math.flux
aanthony1243: when we fix an issue involving visibility of package functions in calls to `map` and `filter`  this can go away.

affo: So you need it in the prelude to make map and filter see math functions?

aanthony1243: yes, and this would apply to any package, e.g., package `strings`  does not have this hack and therefore is currently not visible for `map` or `filter`.

PR #1049, opened 2019-03-11
Fix/memory consumption

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR does several things:

First, it adds a new `TriggerSpec` field to a physical plan node. The trigger spec defines (partly) how a transformation is implemented and therefore should be part of physical query planning.

Next it defines a registration method for registering procedures with specific trigger specifications. In addition a method has been added to the physical planning stage that walks the plan graph and sets the `TriggerSpec` field on each plan node. For procedures that do not have a trigger spec registered, the default `AfterWatermarkTriggerSpec` is used.

It moves all the different trigger spec definitions into the plan package. It also defines a new triggering spec called the `NarrowTranformationTriggerSpec` that fires immediately and finishes immediately.

Finally it changes all narrow transformations to use this new trigger, ensuring that for such transformations, their data caches only ever contain at most one table, thereby reducing the total memory consumption of queries composed of narrow transformations.

### DIFF
--- a/stdlib/socket/from_test.go
+++ b/stdlib/socket/from_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
 	"github.com/influxdata/flux/mock"
+	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/querytest"
 	"github.com/influxdata/flux/stdlib/socket"
 	"github.com/influxdata/flux/stdlib/universe"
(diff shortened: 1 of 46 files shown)


stdlib/socket/from_test.go
```diff
@@ -212,7 +213,7 @@ source
 			id := executetest.RandomDatasetID()
 			d := executetest.NewDataset(id)
 			c := execute.NewTableBuilderCache(executetest.UnlimitedAllocator)
-			c.SetTriggerSpec(flux.DefaultTrigger)
+			c.SetTriggerSpec(plan.DefaultTriggerSpec)
```
wolffcm: I would think that sources don't need a trigger spec, since they always produces tables that are ready to be consumed downstream, right?

I question the need to keep the mapping from `kinds -> trigger spec` in a global map.  I recognize that some global maps are needed to make the language extensible, but in general I think we should avoid them where we can.

In this case, it seems like we could just make types that implement `PhysicalProcedureSpec` able to optionally implement a `TriggerSpecker` interface:
```go
type TriggerSpecker interface {
  TriggerSpec() plan.TriggerSpec
}
```
- transformations that are narrow don't implement `TriggerSpecker` and get the narrow trigger spec by default
- transformations that are not narrow can implement `TriggerSpecker` and choose whatever trigger spec they need

This would result in less boilerplate (no changes needed for narrow transformations).

nathanielc:
One thing to keep in mind is that eventually users will be able to define triggers. So while we can leverage triggers to optimize narrow transformations they will ultimately be user defined. I like this change and it looks good to merge, just wanted to point out the longer term vision with triggers.

PR #1052, opened 2019-03-12
perf(execute): optimize group lookup for non ordered access patterns

### LINKED ISSUE #1035: Optimize GroupLookup
GroupLookup appears to be a bottleneck according to the torch map from this benchmark.

group_lookup_torch_graph

Create a benchmark for common access patterns (like the one used in `window()`) and attempt to optimize it. It seems to spend most of its time in sort. #1032 may help with making it easier to use group keys in a trie-like format which may reduce the amount of time spent sorting, although that may just trade this for hash access time.

It is worth an attempt anyway.

jsternberg:
This optimizes group lookup for the common scenario of ordered and
interweaved additions with some benefits to unordered. It reduces the
number of allocations and the usage of `sort.Search` by creating groups of
entries that are appended to.

The data structure contains a slice of time ranges. When adding a new
value, it will find which of the time ranges it can append to easily. So
when we are adding new values in sorted order, this results in a single
slice that is continuously appended to.

If the new value is sorted before all of the existing groups, it will
create a new group at the beginning and add the value to that group instead
of attempting to shift all of the values in the existing group like you
would have to do for a single slice.

If the new value would get inserted into the middle of a group such as with
unordered access, it will split the existing group into two so the values
after the currently added value would be in a new group and the new value
can be appended to the current group.

This means that if you add values in a sorted order, you will only get a
worst case scenario the first time you append a value rather than on every
insertion.

The worst case scenario for this data structure is adding the values in a
reverse sorted order.

The data structure could further be improved by having the top slice be
sorted backwards rather than sorted in the current direction. This would
likely eliminate the worst case scenario and would prevent the copy when
appending a key that is less than all of the other keys.

- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #1035.

### DIFF
--- a/execute/group_lookup.go
+++ b/execute/group_lookup.go
@@ -7,66 +7,198 @@ import (
 )
 
 type GroupLookup struct {
-	groups groupEntries
+	// groups contains groups of group keys in sorted order.
+	// These are optimized for appending access.
+	groups []*groupKeyList
 
-	//  range state
-	rangeIdx int
+	// lastIndex contains the last group that an entry was
+	// found in or appended to. This is used to optimize appending.
+	lastIndex int
 
-	// Indicates whether the group needs to be sorted.
-	needSort bool
+	// nextID is the next id that will be assigned to a key group.
+	nextID int
 }
 
-type groupEntry struct {
-	key   flux.GroupKey
-	value interface{}
+// groupKeyList is a group of keys in sorted order.
+type groupKeyList struct {
+	id       int // unique id for the key group within the group lookup
+	elements []groupKeyListElement
+	deleted  int
+}
+
+type groupKeyListElement struct {
+	key     flux.GroupKey
+	value   interface{}
+	deleted bool
+}
+
+func (kg *groupKeyList) First() flux.GroupKey {
+	return kg.elements[0].key
+}
+
+func (kg *groupKeyList) Last() flux.GroupKey {
+	return kg.elements[len(kg.elements)-1].key
+}
+
+func (kg *groupKeyList) set(i int, value interface{}) {
+	if kg.elements[i].deleted {
+		kg.elements[i].deleted = false
+		kg.deleted--
+	}
+	kg.elements[i].value = value
+}
+
+func (kg *groupKeyList) delete(i int) {
+	kg.elements[i].value = nil
+	kg.elements[i].deleted = true
+	kg.deleted++
+}
+
+// Index determines the location of this key within the key group.
+// It returns -1 if this key does not exist within the group.
+// It will return -1 if the entry is present, but deleted.
+func (kg *groupKeyList) Index(key flux.GroupKey) int {
+	i := kg.InsertAt(key)
+	if i >= len(kg.elements) || kg.elements[i].deleted || !kg.elements[i].key.Equal(key) {
+		return -1
+	}
+	return i
+}
+
+// InsertAt will return the index where this key should be inserted.
+// If this key would be inserted before the first element, this will
+// return 0. If the element exists, then this will return the index
+// where that element is located. If the key should be inserted at the
+// end of the array, it will return an index the size of the array.
+func (kg *groupKeyList) InsertAt(key flux.GroupKey) int {
+	if kg.Last().Less(key) {
+		return len(kg.elements)
+	}
+	return sort.Search(len(kg.elements), func(i int) bool {
+		return !kg.elements[i].key.Less(key)
+	})
+}
+
+func (kg *groupKeyList) At(i int) interface{} {
+	return kg.elements[i].value
 }
 
 func NewGroupLookup() *GroupLookup {
 	return &GroupLookup{
-		groups: make(groupEntries, 0, 100),
+		lastIndex: -1,
+		nextID:    1,
 	}
-}
-
-func (l *GroupLookup) findIdx(key flux.GroupKey) int {
-	if l.needSort {
-		sort.Sort(l.groups)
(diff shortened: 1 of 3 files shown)


jsternberg:
Here are the performance numbers:

```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         129291        52930         -59.06%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     3113840       85321         -97.26%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       2863064       193151        -93.25%

benchmark                                              old allocs     new allocs     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         2              10             +400.00%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     86             35             -59.30%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       102            152            +49.02%

benchmark                                              old bytes     new bytes     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         9728          10648         +9.46%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     12416         9656          -22.23%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       12928         11784         -8.85%
```

As can be seen, there's some additional overhead for the new data structure in the sorted method, but the CPU usage is universally better across the board. The additional allocations are likely because I am not preallocating array space for the added entries and they don't appear to be too significant.

We will likely have to decide if the memory allocations are worth it. The number of bytes appears to be less for everything except ordered where ordered is likely because the test uses 105 series and the old group lookup had an initial buffer of 100 so it only needed to do one additional allocation.

execute/group_lookup.go
```diff
@@ -7,102 +7,242 @@ import (
...
 }
 
-func (l *GroupLookup) findIdx(key flux.GroupKey) int {
-	if l.needSort {
-		sort.Sort(l.groups)
-		l.needSort = false
+// Index determines the location of this key within the key group.
+// It returns -1 if this key does not exist within the group.
+// It will return -1 if the entry is present, but deleted.
+func (kg *keyGroup) Index(key flux.GroupKey) int {
```
wolffcm: Seems like this could be a non-exported method?

execute/group_lookup.go
```diff
@@ -7,102 +7,242 @@ import (
...
-	// Indicates whether the group needs to be sorted.
-	needSort bool
+	// nextID is the next id that will be assigned to a key group.
+	nextID int
 }
 
-type groupEntry struct {
-	key   flux.GroupKey
-	value interface{}
+// keyGroup is a group of keys in sorted order.
```
wolffcm: It would help to make this code easier to understand if you used a word besides `group` to describe a collection of group keys in sorted order... Not sure what the best word would be though.

wolffcm: `keyList` maybe?

execute/group_lookup.go
jsternberg: It's technically not exported because it's on an unexported type. I like to do this because I find it easier to read and it doesn't really matter.

execute/group_lookup.go
jsternberg: I like this idea. I'm changing it to `groupKeyList`.

jsternberg:
I'll integrate these into the pushed PR, but here's two additional benchmarks. One is if the groups are added in reverse order and one is where I believe there will be many splits. It takes sorted groups of 3 and swaps the 2nd and 3rd elements. This means that the 1st gets appended somewhere, the 3rd gets appended, and then the 2nd forces a split.

```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         130501        52601         -59.69%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     2954610       83837         -97.16%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       2932631       176408        -93.98%
BenchmarkGroupLookup_LookupOrCreate/Reversed-12        3947501       186467        -95.28%
BenchmarkGroupLookup_LookupOrCreate/ManySplits-12      1016825       129218        -87.29%

benchmark                                              old allocs     new allocs     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         2              10             +400.00%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     86             35             -59.30%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       102            152            +49.02%
BenchmarkGroupLookup_LookupOrCreate/Reversed-12        105            218            +107.62%
BenchmarkGroupLookup_LookupOrCreate/ManySplits-12      36             148            +311.11%

benchmark                                              old bytes     new bytes     delta
BenchmarkGroupLookup_LookupOrCreate/Ordered-12         9728          10648         +9.46%
BenchmarkGroupLookup_LookupOrCreate/Interweaved-12     12416         9656          -22.23%
BenchmarkGroupLookup_LookupOrCreate/Unordered-12       12928         11784         -8.85%
BenchmarkGroupLookup_LookupOrCreate/Reversed-12        13024         12120         -6.94%
BenchmarkGroupLookup_LookupOrCreate/ManySplits-12      10816         12712         +17.53%
```

Memory consumption is a bit higher, but CPU usage is still down. I think it might help to preallocate a small array for each new key list, but I'll save that for another PR.

jsternberg:
I refactored the tests and benchmarks so they would share code. If it passes CI, I'll merge this and we can iterate on it further.

PR #1053, opened 2019-03-13
Perf/window trigger

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/plan/triggers.go
+++ b/plan/triggers.go
@@ -35,7 +35,7 @@ func SetTriggerSpec(node PlanNode) error {
 	spec := ppn.Spec
 	if n, ok := spec.(TriggerAwareProcedureSpec); ok {
 		ppn.TriggerSpec = n.TriggerSpec()
-	} else {
+	} else if ppn.TriggerSpec == nil {
 		ppn.TriggerSpec = DefaultTriggerSpec
 	}
 	return nil
--- a/stdlib/universe/window.go
+++ b/stdlib/universe/window.go
@@ -44,6 +44,7 @@ func init() {
 	flux.RegisterOpSpec(WindowKind, newWindowOp)
 	flux.RegisterPackageValue("universe", "inf", infinityVar)
 	plan.RegisterProcedureSpec(WindowKind, newWindowProcedure, WindowKind)
+	plan.RegisterPhysicalRules(WindowTriggerPhysicalRule{})
 	execute.RegisterTransformation(WindowKind, createWindowTransformation)
 }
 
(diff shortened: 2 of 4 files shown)


plan/triggers.go
```diff
@@ -41,6 +45,84 @@ func SetTriggerSpec(node PlanNode) error {
 	return nil
 }
 
+// OptimizeWindows optimizes the triggering mechanism for certain window transformations.
+// In particular, if a windowing operation occurs in a query plan and does not have any
+// dependencies on other window operations, then this particular window operation may send
+// its output tables immediately downstream after processing an input table.
+func OptimizeWindows(spec *PlanSpec) error {
```
wolffcm: Is there any reason this can't fit into the rule framework?  There will be lots of optimizations like this, and I'd prefer to extend the rule framework to allow for this kind of transformation, rather than do many passes like this.

jlapacik: @wolffcm thank you for the suggestion. The window optimization has been changed to be a physical plan rule. It's much more concise. The logic is in the rule itself now, not in the planner.

jlapacik:
@wolffcm here is the next iteration. The rule has been updated to only match a physical window operation. The allowed predecessors in order for the optimization to take place have been explicitly defined in the body of the rewrite rule.

The allowed predecessors include all selectors, schema mutators, and transformations like filter, group, and sort.

stdlib/universe/window.go
```diff
@@ -465,12 +459,8 @@ func hasValidPredecessors(node plan.PlanNode) bool {
 		DuplicateKind,
 		RenameKind,
 		ShiftKind,
-		SkewKind,
 		SortKind,
-		SpreadKind,
 		StateTrackingKind,
-		StddevKind,
-		SumKind,
 		UniqueKind:
```
wolffcm: I got confused by the logic here because `go` does not have implicit fallthrough.

I think if you hit a `range` here, you can just return true and stop recursing.  Range will set the `_start` and `_stop` columns to be in the group key and have the same value in all tables.

Also, not all sources will produce uniform bound for all tables...  It would be possible to produce overlapping bounds using `csv.from` for example.

I think the best way to do this would be to check that the plan node implements `BoundsAwareProcedureSpec`, and that its `TimeBounds` method returns a non-nil value.  This will match both `range` and `influxdb.from` with bounds pushed down.

stdlib/universe/window.go
```diff
@@ -405,42 +405,78 @@ func (WindowTriggerPhysicalRule) Name() string {
...
+		LimitKind,
+		MaxKind,
+		MeanKind,
+		MinKind,
+		PercentileKind,
+		ExactPercentileAggKind,
+		ExactPercentileSelectKind,
+		RangeKind,
+		SampleKind,
+		DropKind,
```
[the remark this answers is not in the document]
jlapacik: Yes the bounds defined by `_start` and `_stop` are already part of the plan nodes that represent those transformations. So yes they are still narrow.

stdlib/universe/window.go
jlapacik: I added this comment here because I was unsure if that was a correct assumption:
https://github.com/influxdata/flux/blob/3dd4a1fedbd8d5c578a7e6c21d89db988aaf9c66/stdlib/universe/window.go#L433

The time bounds check however will ensure it is. Thanks,

stdlib/universe/window.go
```diff
@@ -395,3 +396,84 @@ func (t *fixedWindowTransformation) UpdateProcessingTime(id execute.DatasetID, p
...
+func hasValidPredecessors(node plan.PlanNode) bool {
+	pred := node.Predecessors()
+	// Source nodes might not produce uniform time bounds for all
+	// tables in which case we can't optimize window. However if a
+	// source is a BoundsAwareProcedureSpec then it must produce
+	// bounded data in which case we can perform the optimization.
+	if len(pred) == 0 {
+		s := node.ProcedureSpec()
+		_, ok := s.(plan.BoundsAwareProcedureSpec)
+		return ok
```
wolffcm: This will work just fine, but can you add an additional check here?
```
if bs, ok := s.(plan.BoundsAwareProcedureSpec); bs.TimeBounds(nil) != nil {
  return true
}
```
If the `from` does not have a `range` pushed into it, it's not really bounded and `TimeBounds` will return `nil`.  Currently such a query would fail planning since we require bounds on all InfluxDB sources, but it might not always be the case.  I just want to be sure that this rule is robust to any changes we might make in the future.

PR #1069, opened 2019-03-19
feat(stdlib): row function now includes local identifier scope.  

aanthony1243:
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


stdlib/flux_test.go
```diff
@@ -33,6 +33,7 @@ var skip = map[string]string{
 	"yield":                       "yield requires special test case (https://github.com/influxdata/flux/issues/535)",
 	"task_per_line":               "join produces inconsistent/racy results when table schemas do not match (https://github.com/influxdata/flux/issues/855)",
 	"rowfn_with_import":           "imported libraries are not visible in user-defined functions (https://github.com/influxdata/flux/issues/1000)",
+	"string_trim":                 "cannot reference a package function from within a row function",
```
jsternberg: Is this a future fix? I thought the local identifier scope and the package issue were the same one.

aanthony1243: yeah I wrote that with the intention of getting it to pass.  This PR is a half-solution to get object identifiers visible and working for the front-end team

aanthony1243:
@jsternberg  no rush, but I added support for array index expressions

Issue #1077, opened 2019-03-22
The AST generated for logical operations is wrong


affo:
When parsing `a or b and c`, the parser produces:

![Schermata 2019-03-22 alle 18 26 25](https://user-images.githubusercontent.com/4246582/54842091-35713c00-4cd1-11e9-9631-7f0647119e84.png)

So:

```
            and
            /  \
          or    c
         /  \
        a   b
```

While it should produce:

```
    or
   /  \
 a    and
      /  \
     b    c
```

It is also weird that the SPEC says that `and` and `or` operators have the same precedence. `and` should have higher precedence than `or`.

jsternberg:
We would have to change the spec. The current spec has them at the same precedence which means the parser currently follows the spec.

I'm fine if we decide to change the precedence though. It seems python gives `and` a higher precedence and ruby gives `and/or` the same precedence. Javascript doesn't have the words, but it has a similar precedence to python.

affo:
Mmm 🤔 It seems to me that both Python and Ruby give the same precedence to boolean operators:

Python:

![Schermata 2019-03-25 alle 09 36 29](https://user-images.githubusercontent.com/4246582/54905614-bf8fef00-4ee1-11e9-8f2b-950996b437ff.png)

Ruby:

![Schermata 2019-03-25 alle 09 37 20](https://user-images.githubusercontent.com/4246582/54905619-c454a300-4ee1-11e9-9e0c-2c5e47b2e3f9.png)

Operator precedence in boolean algebra is not defined, but I think that precedence of `and` wrt `or` is a well established convention, so we should follow that.

jsternberg:
I'm not disagreeing with the conclusion, but `and` and `or` have different connotations in programming languages. Ruby has distinct operators for those with different precedence and ours were originally based off of those.

If you notice, the Ruby operators for `&&` and `||` have much higher precedence in the precedence table than `and` and `or` and our precedence rules match this.

Now, I think we should go with the Python precedences and I think it is more common for people to believe what you stated in the top post. I just wanted to give some clarity about the different connotations of the words vs the symbols.

<img width="305" alt="Screen Shot 2019-03-25 at 2 46 08 PM" src="https://user-images.githubusercontent.com/1212129/54949313-c4b66380-4f0c-11e9-86b3-17b720b05052.png">

affo:
Yeah, got your point. Thank you for the clarification 👍

nathanielc:
[...]

PR #1078, opened 2019-03-22
fix(ast): format AST preserving operator precedence

### LINKED ISSUE #1012: Formatting AST loses parentheses and thus operator associativity
An AST that was parsed from `(1.0 * 2.0 + 4.0 / 6.0 - 10.0) * r._value` gets formatted out by `ast.Format()` as `1.0 * 2.0 + 4.0 / 6.0 - 10.0 * r._value`, losing parentheses and thus losing the original associativity.

Either parentheses should be explicitly preserved in the AST or the Formatter needs to add them back according to operator precedence, when needed.

affo:
Closes #1012 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/ast/format.go
+++ b/ast/format.go
@@ -7,15 +7,12 @@ import (
 	"time"
 )
 
-/*
-Returns a valid script for a given AST rooted at node `n`.
-
-Formatting rules:
- - In a list of statements, if two statements are of a different type
-	(e.g. an `OptionStatement` followed by an `ExpressionStatement`), they are separated by a double newline.
- - In a function call (or object definition), if the arguments (or properties) are more than 3,
-	they are split into multiple lines.
-*/
+//Returns a valid script for a given AST rooted at node `n`.
+//Formatting rules:
+// - In a list of statements, if two statements are of a different type
+//	(e.g. an `OptionStatement` followed by an `ExpressionStatement`), they are separated by a double newline.
+// - In a function call (or object definition), if the arguments (or properties) are more than 3,
+//	they are split into multiple lines.
 func Format(n Node) string {
 	f := &formatter{new(strings.Builder), 0}
 	f.formatNode(n)
@@ -31,9 +28,9 @@ func (f *formatter) get() string {
 	return f.String()
 }
 
-// `strings.Builder`'s methods never return a non-nil error,
-// so it is safe to ignore it.
 func (f *formatter) writeString(s string) {
+	// `strings.Builder`'s methods never return a non-nil error,
+	// so it is safe to ignore it.
 	f.WriteString(s)
 }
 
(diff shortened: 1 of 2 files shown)


ast/format.go
```diff
@@ -64,6 +61,113 @@ func (f *formatter) writeComment(comment string) {
...
+	getIntForOp(NotEmptyOperator):         4,
+	getIntForOp(EmptyOperator):            4,
+	getIntForOp(EqualOperator):            4,
+	getIntForOp(NotEqualOperator):         4,
+	getIntForOp(RegexpMatchOperator):      4,
+	getIntForOp(NotRegexpMatchOperator):   4,
+	getIntForOp(NotOperator):              5,
+	// theses are LogicalOperatorKinds:
+	getIntForLOp(AndOperator): 6,
+	getIntForLOp(OrOperator):  7,
```
affo: Here I changed the precedence of `and` wrt `or` according to https://github.com/influxdata/flux/issues/1077.
If we do not agree on that, I can change to `6, 6` and avoid skipping related tests.

ast/format.go
```diff
@@ -73,7 +177,9 @@ func (f *formatter) formatPackage(n *Package) {
 			f.writeRune('\n')
 			f.writeRune('\n')
 		}
-		f.writeComment(file.Name)
+		if len(file.Name) > 0 {
+			f.writeComment(file.Name)
+		}
```
affo: This fixes the issue we had when formatting a generated AST that doesn't have a file provenance.
Otherwise, it would generate:

```
package main
//

...
```

The related test is below

ast/format_test.go
```diff
@@ -283,27 +320,79 @@ highestAverage = (n, columns=["_value"], by, tables=<-) =>
...
+			name:   "unary with pars",
+			script: `not (b and c)`,
+		},
+		{
+			name:   "function call with pars",
+			script: `(a + b * c == 0)(foo: "bar")`,
+		},
+		{
+			name:   "member with pars",
+			script: `((a + b) * c)._value`,
```
jlapacik: Should this be a test case? I guess this parses just fine, but I don't think it would type check. Are there any member expression examples that are valid Flux expressions? I can't really think of any.

ast/format_test.go
```diff
@@ -283,27 +320,79 @@ highestAverage = (n, columns=["_value"], by, tables=<-) =>
...
+		},
+		{
+			name:   "index with pars",
+			script: `((a - b) / (c + d))[3]`,
```
jlapacik: Same here. This parses just fine, but it's not a valid Flux expression (would not pass type checking). Is it possible to come up with a test case for index expressions that is also a valid Flux expression?

ast/format.go
```diff
@@ -64,6 +61,113 @@ func (f *formatter) writeComment(comment string) {
...
+
+func getPrecedenceForLOp(op LogicalOperatorKind) int {
+	return getPrecedence(getIntForLOp(op))
+}
+
+// this matches the SPEC
+var opPrecedence = map[int]int{
+	functionCall: 1,
+	member:       1,
+	index:        1,
```
jlapacik: Do index and member need to be included? See my comments regarding their associated test cases.

affo: I absolutely see your point, but I made it for completeness and in order to directly match the SPEC. I also know that those in the tests wouldn't pass type checking, but in the future they could be (say we support `+` for arrays). So I wanted to check if operator precedence was respected even in that cases, also given that there is actually a formally defined precedence of (say) member VS addition in the SPEC.

ast/format_test.go
affo: Yeah, tests for index and member expressions are in `TestFormat_Nodes`. This test is specific about associativity/precedence. The rest of the answer is below.

ast/format_test.go
affo: Answered below.

ast/format.go
```diff
@@ -64,6 +61,113 @@ func (f *formatter) writeComment(comment string) {
...
+	getIntForOp(NotEqualOperator):         4,
+	getIntForOp(RegexpMatchOperator):      4,
+	getIntForOp(NotRegexpMatchOperator):   4,
+	getIntForOp(NotOperator):              5,
+	// theses are LogicalOperatorKinds:
+	getIntForLOp(AndOperator): 6,
+	getIntForLOp(OrOperator):  7,
+}
+
+func (f *formatter) formatChildWithPars(parent, child Node) {
```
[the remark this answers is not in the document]
affo: 👍

ast/format.go
```diff
@@ -253,27 +359,29 @@ func (f *formatter) formatFunctionExpression(n *FunctionExpression) {
...
-	f.writeString(n.Operator.String())
-	f.writeRune(' ')
-	f.formatNode(n.Right)
+	f.formatBinary(n.Operator.String(), n, n.Left, n.Right)
 }
 
 func (f *formatter) formatLogicalExpression(n *LogicalExpression) {
-	f.formatNode(n.Left)
+	f.formatBinary(n.Operator.String(), n, n.Left, n.Right)
+
```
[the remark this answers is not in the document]
affo: 👍

affo:
In general, I think this PR should be merged after I produce a fix for https://github.com/influxdata/flux/issues/1077 (thanks to @jsternberg 😄  ).
In that case, I could remove skipping tests. I will ping you @jlapacik once it is done

affo:
I rebase on the branch of #1098 .

@jlapacik In the last commit I addressed your comments and removed the skipped tests for bad parenthesis.

ast/format_test.go
```diff
@@ -10,12 +10,9 @@ import (
...
 var skip = map[string]string{
-	"array_expr":            "without pars -> bad syntax, with pars formatting removes them",
-	"conditional":           "how is a conditional expression defined in spec?",
-	"multi_var_decl":        "how is a variable declaration with multiple declarations represented?",
-	"logic":                 "the AST created for logic ops is wrong (https://github.com/influxdata/flux/issues/1077)",
-	"logic with comparison": "the AST created for logic ops is wrong (https://github.com/influxdata/flux/issues/1077)",
-	"logic and math":        "the AST created for logic ops is wrong (https://github.com/influxdata/flux/issues/1077)",
+	"array_expr":     "without pars -> bad syntax, with pars formatting removes them",
+	"conditional":    "how is a conditional expression defined in spec?",
+	"multi_var_decl": "how is a variable declaration with multiple declarations represented?",
```
jlapacik: Flux doesn't have a `var` keyword. Can you remove this test?

affo: absolutely, we had it in the past 😅

PR #1091, opened 2019-03-25
feat: parse-signed-duration

### LINKED ISSUE #1081: Bug: Duration parsing fails for negative durations.
Duration parsing fails for negative durations.
This errors for  `-1m`.
```
func ParseDuration(lit string) ([]ast.Duration, error) {
	var values []ast.Duration
	for len(lit) > 0 {
		n := 0
		for n < len(lit) {
			ch, size := utf8.DecodeRuneInString(lit[n:])
			if size == 0 {
				panic("invalid rune in duration")
			}

			if !unicode.IsDigit(ch) { 
				break
			}
			n += size
		}

		magnitude, err := strconv.ParseInt(lit[:n], 10, 64)
		if err != nil {
			return nil, err
		}
		lit = lit[n:]

		n = 0
		for n < len(lit) {
			ch, size := utf8.DecodeRuneInString(lit[n:])
			if size == 0 {
				panic("invalid rune in duration")
			}

			if !unicode.IsLetter(ch) {
				break
			}
			n += size
		}
		unit := lit[:n]
		if unit == "µs" {
			unit = "us"
		}
		values = append(values, ast.Duration{
			Magnitude: magnitude,
			Unit:      unit,
		})
		lit = lit[n:]
	}
	return values, nil
}
```

I think the problem is at `if !unicode.IsDigit(ch) { `, it needs to check to allow negative signs.

docmerlin:
A public function to parse signed durations.

closes #1081

### DIFF
--- a/parser/strconv.go
+++ b/parser/strconv.go
@@ -1,10 +1,13 @@
 package parser
 
 import (
+	"unicode/utf8"
+
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/internal/parser"
 )
 
+// ParseTime will convert a string into an *ast.DateTimeLiteral.
 func ParseTime(lit string) (*ast.DateTimeLiteral, error) {
 	d, err := parser.ParseTime(lit)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


parser/strconv.go
```diff
@@ -16,6 +17,32 @@ func ParseTime(lit string) (*ast.DateTimeLiteral, error) {
 	}, nil
 }
 
+// ParseSignedDuration will convert a string into a possibly negative DurationLiteral.
+func ParseSignedDuration(lit string) (*ast.DurationLiteral, error) {
+	if len(lit) > 0 && lit[0] == '-' {
```
nathanielc: We should use https://golang.org/pkg/unicode/utf8/#DecodeRuneInString here to ensure that we are utf-8 safe.

parser/strconv.go
```diff
@@ -16,6 +17,32 @@ func ParseTime(lit string) (*ast.DateTimeLiteral, error) {
...
+		d, err := parser.ParseDuration(lit[1:])
```
nathanielc: I would prefer this to consume the local ParseDuration method instead of the one in the parser package. That way the logic to construct the ast node is not duplicated.

parser/strconv.go
docmerlin: lit[0] == '-' is utf8 safe because it is at the start of the string, but I can see how using DecodeRuneInString is more standard, I'll switch over.

parser/strconv.go
```diff
@@ -16,6 +19,28 @@ func ParseTime(lit string) (*ast.DateTimeLiteral, error) {
 	}, nil
 }
 
+// ParseSignedDuration will convert a string into a possibly negative DurationLiteral.
+func ParseSignedDuration(lit string) (*ast.DurationLiteral, error) {
+	r, s := utf8.DecodeRuneInString(lit)
+	if len(lit) > 0 && r == '-' {
```
nathanielc: The len check is no longer needed.

PR #1092, opened 2019-03-25
fix(lang): Add error message to Compile methods for empty Spec

### LINKED ISSUE #1054: Failed to create initial logical plan: query has no root nodes
@davidgs reports that this query produces an error in 2.0 alpha5: `Error: failed to create initial logical plan: query has no root nodes`

```
Tref = 298.15
Pref = 1013.25 

Tmeas = from(bucket: "telegraf")
  |> range(start: v.timeRangeStart)
  |> filter(fn: (r) => r._measurement == "environment" and (r._field == "temp_c"))
  |> aggregateWindow(every: 30s, fn: mean)
  |> keep(columns: ["_value", "_time"])
  
CO2meas = from(bucket: "telegraf")
  |> range(start: v.timeRangeStart)
  |> filter(fn: (r) => r._measurement == "k30_reader" and (r._field == "co2") and (r.sensor == "k30_co2"))
  |> aggregateWindow(every: 30s, fn: mean)
  |> keep(columns: ["_value", "_time"])

Pmeas = from(bucket: "telegraf")
  |> range(start: v.timeRangeStart)
  |> filter(fn: (r) => r._measurement == "environment" and (r._field == "pressure"))
  |> aggregateWindow(every: 30s, fn: mean)
  |> keep(columns: ["_value", "_time"])

 
 first_join = join(tables: {CO2meas: CO2meas, Tmeas: Tmeas}, on: ["_time"])
 

 second_join = join(tables: {first_join: first_join, Pmeas: Pmeas}, on: ["_time"])
    |>map(fn: (r) => ({_time: r._time, _Pmeas: r._value, _CO2meas:r._value_CO2meas, _Tmeas:r._value_Tmeas}))

final = second_join 
    |>map(fn: (r) => ({Pmeas: r._Pmeas, CO2meas:r._CO2meas, Tmeas:r._Tmeas, Pref: Pref, Tref: Tref, _time: r._time,}))
    |> keep(columns: ["_time", "CO2meas", "Pmeas", "Tmeas", "Pref", "Tref"])
CO2corr = final 
     |> map(fn: (r) => ({"_time": r._time, "CO2-Measured": r.CO2meas, "CO2-Adjusted": r.CO2meas * (((r.Tmeas + 273.15) * r.Pref) / (r.Pmeas * r.Tref))}))
```
That error would only occur just as planning was starting, so should not require any data to reproduce.
I tried to reproduce from the CLI, but I could not.  Would be good to try to reproduce from the web UI.

wolffcm:
Fixes #1054.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -41,6 +41,7 @@ type options struct {
 
 // Compile evaluates a Flux script producing a query Spec.
 // now parameter must be non-zero, that is the default now time should be set before compiling.
+// If the given Flux script would produce an empty Spec, an error will be returned.
 func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spec, error) {
 	astPkg, err := Parse(q)
 	if err != nil {
@@ -52,6 +53,7 @@ func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spe
 
 // CompileAST evaluates a Flux AST and produces a query Spec.
 // now parameter must be non-zero, that is the default now time should be set before compiling.
+// If the given AST would produce an empty Spec, an error will be returned.
 func CompileAST(ctx context.Context, astPkg *ast.Package, now time.Time, opts ...Option) (*Spec, error) {
 	o := new(options)
 	for _, opt := range opts {
@@ -82,6 +84,10 @@ func CompileAST(ctx context.Context, astPkg *ast.Package, now time.Time, opts ..
 	spec, err := ToSpec(sideEffects, nowTime.Time().Time())
 	if err != nil {
 		return nil, err
+	}
+
+	if len(spec.Operations) == 0 {
+		return nil, errors.New("this Flux script returns no streaming data. Consider adding a \"yield\" or invoking streaming functions directly, without performing an assignment")
 	}
 
 	if o.verbose {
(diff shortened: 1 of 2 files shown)


compile.go
```diff
@@ -84,6 +86,10 @@ func CompileAST(ctx context.Context, astPkg *ast.Package, now time.Time, opts ..
 		return nil, err
 	}
 
+	if len(spec.Operations) == 0 {
+		return nil, errors.New("Flux script did not invoke any table-streaming functions.  Do you need to add a \"yield\" to your query?")
```
nathanielc: This message is a little misleading as invoking a table streaming function is not enough. For example this produces an empty spec:

```
foo = from() |> ...
```

The assignment to `foo` means that the function calls did not produce a side effect, in order to get invoked.

wolffcm: Yeah.  I'm struggling to find a concise way to express this in a friendly error message that tells the user how to fix the problem.  How about this?
```
this Flux script returns no streaming data.  Consider adding a "yield" or invoking streaming functions directly, without performing an assignment.
```

nathanielc: @wolffcm That is much better. Thanks.

Issue #1085, opened 2019-03-25
Make flux.Spec private


nathanielc:
This is a parent issue for all the tasks needed to make the flux.Spec and internal detail only.

Other issues will be linked to this issue.

wolffcm:
Related issue to make the planner to provide a builder and convenient interface:
https://github.com/influxdata/flux/issues/1097

wolffcm:
I created this issue to update all clients to use `ASTCompiler` instead of `SpecCompiler`:
https://github.com/influxdata/influxdb/issues/12928

wolffcm:
Issue to deal with dependencies with new `Compiler` interface.
https://github.com/influxdata/flux/issues/1126

jsternberg:
This is effectively private and we'll talk more about what should be done as we begin work on the IR language refactor in the future.

PR #1095, opened 2019-03-26
refactor(lang): Update Compiler interface to produce a Program

### LINKED ISSUE #1086: Change flux.Compiler interface to remove Spec
The compiler interface should not expose the Spec.

The interface should look like:

```
type Compiler interface {
    Compile(context.Context, Request) (Program,error)
}
```

Where Program is:

```
type Program interface {
    // Starts execution of the program and returns immediately
    // As results are produced they arrive on the channel, the program is finished once the result channel is closed and all results have been consumed.
    Start(context) (<-chan flux.Result, error)
}
```

Previously the results returned from execution where a `map[string]flux.Result` but that is no longer a valid return value since an undetermined number of Results may be produced. A result knows it own name so the map is not needed. 

Also of note is that multiple results can be read off the channel before the results themselves are fully consumed. So while the retrieval of the results is serial the results can still be consumed concurrently. 

This is part of #1085

wolffcm:
I did as little as possible to enable the vanilla Flux REPL to work.

Unit tests are not yet passing, I thought I would get feedback before I went to the trouble of fixing them to avoid unnecessary work.

I made `Controller.Query()` just trivially compile and run the `Program` produced by its `Compiler` argument.  As a result, `Controller.compileQuery` is totally unneeded now and caused compiler errors, so I just deleted its body.

Fixes #1086.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -6,6 +6,9 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
 )
 
 const (
@@ -38,8 +41,21 @@ type FluxCompiler struct {
 	Query string `json:"query"`
 }
 
-func (c FluxCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
-	return flux.Compile(ctx, c.Query, time.Now())
+func (c FluxCompiler) Compile(ctx context.Context) (flux.Program, error) {
+	spec, err := flux.Compile(ctx, c.Query, time.Now())
+	if err != nil {
+		return nil, err
+	}
+
+	planner := (&plan.PlannerBuilder{}).Build()
+	ps, err := planner.Plan(spec)
+	if err != nil {
+		return nil, err
+	}
+
+	return Program{
+		ps: ps,
+	}, err
 }
 
 func (c FluxCompiler) CompilerType() flux.CompilerType {
@@ -51,9 +67,18 @@ type SpecCompiler struct {
 	Spec *flux.Spec `json:"spec"`
 }
 
-func (c SpecCompiler) Compile(ctx context.Context) (*flux.Spec, error) {
-	return c.Spec, nil
+func (c SpecCompiler) Compile(ctx context.Context) (flux.Program, error) {
+	planner := (&plan.PlannerBuilder{}).Build()
+	ps, err := planner.Plan(c.Spec)
+	if err != nil {
+		return nil, err
+	}
+
+	return Program{
+		ps: ps,
+	}, err
 }
+
 func (c SpecCompiler) CompilerType() flux.CompilerType {
 	return SpecCompilerType
 }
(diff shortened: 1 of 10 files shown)


lang/compiler.go
```diff
@@ -79,3 +114,45 @@ func (ASTCompiler) CompilerType() flux.CompilerType {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &Query{
+		rmap: rmap,
+	}, nil
+}
+
+// TODO(cwolff): There should be an easy method presented by the plan package to do this.
```
nathanielc: Agreed can you create an issue to track fixing this? And link it to the parent #1085 issue so we track it as part of this whole process?

query.go
```diff
@@ -6,10 +6,6 @@ import (
 
 // Query represents an active query.
 type Query interface {
```
nathanielc: I would have expected the Ready method to need to change to be a `<-chan Result` instead of a `<-chan map[string]Result`. The difference being that now the Ready channel can produce more than one set of results. In fact the name Ready is inaccurate now since its not about the query being ready but just a mechanism to deliver results.

If this should be a follow up PR that is fine, lets create an issue for it.

lang/compiler.go
wolffcm: I went ahead and did the work for this:
https://github.com/influxdata/flux/pull/1096
I was planning on merging it to master, and then rebasing this branch when it lands.

wolffcm:
@jsternberg 
> I like most of the PR, but I want to see if we can modify the interface so that the program potentially accepts an executor instead of creating its own.

@nathanielc mentioned that he doesn't think that `Start` should accept an `Executor` but I think that was more with respect to dependencies, rather than resource management.  It does seem like resource management needs to be handled somehow.

jsternberg:
Yea, I don't care too much about the execution itself, but just found executor to be a useful abstraction for that. Maybe we could make the dispatcher and memory allocator into interfaces in the flux package and pass those specifically? Or we can create a new `flux.ResourceManager` interface.

```
type ResourceManager interface {
    Allocator() *memory.Allocator // or the arrow allocator
    Dispatch(ctx context.Context, fn func()) // access to the dispatcher
}
```

wolffcm:
I pushed a commit that addresses feedback:
- `Program.Start` method now takes a `*memory.Allocator` as a parameter.  Nathaniel and I talked about making this more general but for now this will work.  
- I made `Query.Ready` return `chan Result`

query.go
```diff
@@ -6,14 +6,10 @@ import (
...
-	// Spec returns the spec used to execute this query.
-	// Spec must not be modified.
-	Spec() *Spec
-
 	// Ready returns a channel that will deliver the query results.
 	// Its possible that the channel is closed before any results arrive,
 	// in which case the query should be inspected for an error using Err().
-	Ready() <-chan map[string]Result
+	Ready() <-chan Result
```
nathanielc: Can you rename this from `Ready` to `Results` or something similar? Ready implies that it fires once which is no longer the case.

result_iterator.go
```diff
@@ -48,26 +48,32 @@ func (r *queryResultIterator) More() bool {
...
+
+	r.nextResult = nr
+	return true
 }
 
 func (r *queryResultIterator) Next() Result {
-	return r.results.Next()
+	if r.nextResult == nil {
+		r.nextResult = <-r.query.Ready()
+	}
```
nathanielc: Next is allowed to panic if More returns `false` and `Next` is called. That means that this check is not necessary and could possibly lead to odd bugs.

wolffcm: I don't think that would quite get the behavior we want, but you're right there's logic errors here. I added a new commit (with comments) that addresses them.

PR #1098, opened 2019-03-27
fix(parser): fix logical operators precedence

### LINKED ISSUE #1077: The AST generated for logical operations is wrong

affo:
Closes #1077 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/parser/grammar.md
+++ b/internal/parser/grammar.md
@@ -26,11 +26,14 @@ The parser directly implements the following grammar.
     AssignStatement                = "=" Expression .
     ReturnStatement                = "return" Expression .
     ExpressionStatement            = Expression .
-    Expression                     = LogicalExpression .
-    ExpressionSuffix               = { PostfixOperator } { PipeExpressionSuffix } { MultiplicativeExpressionSuffix } { AdditiveExpressionSuffix } { ComparisonExpressionSuffix } { LogicalExpressionSuffix } .
-    LogicalExpression              = UnaryLogicalExpression { LogicalExpressionSuffix } .
-    LogicalExpressionSuffix        = LogicalOperator UnaryLogicalExpression .
-    LogicalOperator                = "and" | "or" .
+    Expression                     = LogicalOrExpression .
+    ExpressionSuffix               = { PostfixOperator } { PipeExpressionSuffix } { MultiplicativeExpressionSuffix } { AdditiveExpressionSuffix } { ComparisonExpressionSuffix } { LogicalAndExpressionSuffix } { LogicalOrExpressionSuffix } .
+    LogicalOrExpression            = LogicalAndExpression { LogicalOrExpressionSuffix } .
+    LogicalOrExpressionSuffix      = LogicalOrOperator LogicalAndExpression .
+    LogicalOrOperator              = "or" .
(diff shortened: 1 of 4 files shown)


affo:
@jsternberg I can remember that you told me about a test you wanted to add, but I can't remember which one.

internal/parser/grammar.md
```diff
@@ -26,11 +26,14 @@ The parser directly implements the following grammar.
...
     ReturnStatement                = "return" Expression .
     ExpressionStatement            = Expression .
-    Expression                     = LogicalExpression .
-    ExpressionSuffix               = { PostfixOperator } { PipeExpressionSuffix } { MultiplicativeExpressionSuffix } { AdditiveExpressionSuffix } { ComparisonExpressionSuffix } { LogicalExpressionSuffix } .
-    LogicalExpression              = UnaryLogicalExpression { LogicalExpressionSuffix } .
-    LogicalExpressionSuffix        = LogicalOperator UnaryLogicalExpression .
-    LogicalOperator                = "and" | "or" .
+    Expression                     = LogicalOrExpression .
+    ExpressionSuffix               = { PostfixOperator } { PipeExpressionSuffix } { MultiplicativeExpressionSuffix } { AdditiveExpressionSuffix } { ComparisonExpressionSuffix } { LogicalAndExpressionSuffix } { LogicalOrExpressionSuffix } .
+    LogicalOrExpression            = LogicalAndExpression { LogicalOrExpressionSuffix } .
```
jsternberg: Can you group these together so that `LogicalOrExpression` is followed by `LogicalOrExpressionSuffix`?

internal/parser/parser_test.go
```diff
@@ -2286,38 +2286,263 @@ k / l < m + n - o or p() <= q() or r >= s and not t =~ /a/ and u !~ /a/`,
...
+			name: "logical operators precedence",
+			raw: `not a or b
+a or not b
+not a and b
+a and not b
+a and b or c
+a or b and c
+
+not (a or b)
+not (a and b)`,
```
affo: If I put more parenthesized stuff it fails:

```
not a or b
a or not b
not a and b
a and not b
a and b or c
a or b and c

not (a or b)
not (a and b)
(a or b) and c
a and (b or c)
```

It seemed like it merged lines 9 and 10

jsternberg: Each of those should be a separate test. It is a bit weird that it merged the lines though since they should be separate statements.

affo: Yep it is so. Do you prefer I add a test for that hat gets skipped?

I'll separate tests

jsternberg: I don't think that's needed. Can you file a separate bug for it though so it can be investigated and fixed?

affo: I filed the bug: https://github.com/influxdata/flux/issues/1102

Anyhow I put and skipped the test in the last commit, just to ease reproduction and avoid regression in the future.

PR #1103, opened 2019-03-28
fix: Allow flux unit tests to pass on refactor/private-spec

### LINKED ISSUE #1099: Update compilers used for testing to use new interface
#1095 addresses implementations of `Compiler` in production code, but there are others that we use for testing:
- the `ReplaceSpec` compiler.  It should be updated so tests that use it can run.  Since the compiler now does planning, maybe the planner can add a rule to transform the generated plan (to swap `influxdb.from` nodes to `csv.from` or whatever is needed)
- there is at least one mock compiler.

Related to this issue to update the compiler interface: https://github.com/influxdata/flux/issues/1086

And this is the parent issue for this work: https://github.com/influxdata/flux/issues/1085

wolffcm:
These changes allow flux unit tests to pass on `refactor/private-spec`.

I made changes to the controller so `controller_test` would compile, but I disabled the tests there.  Clearly controller will need a lot of attention.

I updated callers of the `Query` interface to reflect the change.

I also updated the mock Compiler and the RepaceSpecCompiler to use the new compiler interface.  `influxdb` and `idpe` will still be broken, but this moves things forward.

Fixes #1099.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -156,7 +156,7 @@ func (c *Controller) createQuery(ctx context.Context, ct flux.CompilerType) *Que
 		state:              Created,
 		c:                  c,
 		now:                time.Now().UTC(),
-		ready:              make(chan map[string]flux.Result, 1),
+		ready:              make(chan flux.Result, 1),
 		metaCh:             noMetadata, // This will be set to a non-closed channel upon successful execution.
 		parentCtx:          parentCtx,
 		parentSpan:         parentSpan,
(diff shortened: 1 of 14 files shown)


jsternberg:
Can you create an issue for the controller tests? I'll approve after that.

wolffcm:
@jsternberg Sure, here's the issue:
https://github.com/influxdata/flux/issues/1104

PR #1107, opened 2019-03-28
refactor(stdlib): rename percentile to quantile and implement percent…

### LINKED ISSUE #1072: Rename percentile and quantile
Percentile and quantile are basically the same thing except that percentile carries the context of it being a percentage i.e  a number between 0 - 100. We should rename the current percentile function to quantile and add a percentile function that is written in terms of quantile.

```
percentile = (p,tables=<-) => tables |> quantile(q: p / 100)
```

jlapacik:
…ile in terms of quantile

BREAKING CHANGE: Rename the percentile function to quantile. Also rename the percentile
parameter to q. Also implement a new percentile function that takes a parameter p that
is an integer between 0 and 100 inclusive. The new percentile function is defined in
terms of quantile.

Fixes #1072.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1592,10 +1592,10 @@ Median is defined as:
 
     median = (method="estimate_tdigest", compression=0.0, tables=<-) =>
     	tables
-    		|> percentile(percentile:0.5, method:method, compression:compression)
+            |> quantile(q:0.5, method:method, compression:compression)
 
-Is it simply a `percentile` with the `percentile` paramter always set to `0.5`.
-It therefore shares all the same properties as the percentile function.
+Is it simply a `quantile` with the `q` paramter always set to `0.5`.
+It therefore shares all the same properties as the quantile function.
 
 Example:
 ```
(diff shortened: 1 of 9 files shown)


docs/SPEC.md
```diff
@@ -1592,10 +1592,10 @@ Median is defined as:
 
     median = (method="estimate_tdigest", compression=0.0, tables=<-) =>
     	tables
-    		|> percentile(percentile:0.5, method:method, compression:compression)
+    		|> quantile(quantile:0.5, method:method, compression:compression)
```
nathanielc: Below the argument is defined as `q` not `quantile.`

jlapacik:
@nathanielc I think it's probably unnecessary, even though I put up the PR :) . They're exactly the same function, their arguments are just at different scales. I think I prefer to have just one way to perform a quantile calculation. I don't see it as a huge mental burden to switch between a [0, 1] scale vs [0, 100].

nathanielc:
OK, in that case should we just rename it to `quantile` and not have a `percentile` function? There was some confusion around calling it a percentile while using a value between `0-100`.

jlapacik:
I would prefer it called quantile, especially if we are using values between 0 and 1. Currently it's a little inconsistent. But we don't need both functions.

PR #1114, opened 2019-04-02
refactor(flux): move Compile functions to lang package and make them return flux.Program

### LINKED ISSUE #1111: [pvt-spec] Resolve cyclic dependency flux -> lang
TableObject can now be compiled into a `flux.Program` using a dedicated compiler in package `lang` (see #1110).

Package `flux` (`/compile.go`) contains a function for compiling an AST into a `flux.Spec`, that uses the `ToSpec` method to generate a spec using TableObjects: https://github.com/influxdata/flux/blob/e31c1d96e08b5fbf506684488284d0c427d29c02/compile.go#L57

That function should return a `flux.Program instead` and take leverage of the `TableObjectCompiler` (in `lang`) to compile root TableObjects and return results.

However it is impossible to do so, because `lang` depends on `flux` and the compiler lies in `lang`.

Should `lang` be merged into `flux`?

This is part of #1085

affo:
Depends on #1110 

Closes #1111 

This PR moves Compile functions in `flux` to `lang`.

This solves:
 - the cyclic dependency issue between the packages, because `flux` doesn't need to produce a Spec from TableObjects, but only `lang` does.
 - `/compile.go` doesn't contain any reference to Spec anymore.

The result is that, from `flux`, one can `Parse` a string into an AST and `Eval` it, and knows that `Compiler`s exist (objects that produce `Program`s).

In `lang` we have various compilers that still relies on the Spec to carry out their job.

Problems:
 - the plan is hidden in the compilation step and so it not testable, thus commenting out some tests.
 - the Spec is no more inspectable, so every test for new queries fail.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


lang/compiler.go
```diff
@@ -36,26 +42,111 @@ func AddCompilerMappings(mappings flux.CompilerMappings) error {
...
+	return func(o *options) {
+		o.lopts = lopts
+	}
+}
+
+func WithPhysicalPlanningOptions(popts ...plan.PhysicalOption) Option {
+	return func(o *options) {
+		o.popts = popts
+	}
+}
```
affo: Should we set planning rules (options) at this level?
This is unused for now.

lang/compiler.go
```diff
@@ -64,12 +155,16 @@ func (c FluxCompiler) CompilerType() flux.CompilerType {
...
-	Spec *flux.Spec `json:"spec"`
+	Spec  *flux.Spec `json:"spec"`
+	lopts []plan.LogicalOption
+	popts []plan.PhysicalOption
 }
 
 func (c SpecCompiler) Compile(ctx context.Context) (flux.Program, error) {
-	planner := plan.PlannerBuilder{}.Build()
-	ps, err := planner.Plan(c.Spec)
+	planner := plan.PlannerBuilder{}
```
wolffcm: nit, the variable being assigned here is not a `Planner`, but a `PlannerBuilder`.  Maybe rename it to `pb`?

lang/compiler.go
```diff
@@ -118,6 +201,133 @@ func (c *ASTCompiler) PrependFile(file *ast.File) {
...
+	buildSpec(c.Tables, ider, spec, visited)
+
+	specCompiler := &SpecCompiler{
+		Spec: spec,
+	}
+	return specCompiler.Compile(ctx)
+}
+
+func (*TableObjectCompiler) CompilerType() flux.CompilerType {
+	panic("TableObjectCompiler is not associated with a CompilerType")
```
wolffcm: This method is needed when a `Compiler` is sent over the network, so the receiving end knows how to deserialize it from JSON.  For a REPL in `influxdb` of `idpe`, seems like we'll want to be able to serialize `TableObjectCompiler`?  It's unclear to me how this will work.

plan/logical_test.go
```diff
@@ -1,5 +1,8 @@
 package plan_test
 
+// TODO(affo): what to do now that compilation produces a Program?
+//  Should the logical and physical plans be accessible?
+/*
```
wolffcm: I don't want to lose these tests. I have an example of how I did similar testing:
https://github.com/influxdata/flux/blob/55fb97f40907408a3e2ccaf2f3d2c0e11a10c2f0/lang/compiler_test.go#L20

You can access the plan in the `lang.Program`'s `Spec` member.  I think you'll want to do something similar here.

querytest/compile.go
```diff
@@ -33,8 +34,12 @@ var opts = append(
 func NewQueryTestHelper(t *testing.T, tc NewQueryTestCase) {
 	t.Helper()
 
+	// TODO(affo): understand what to do with new queries tests, given that the Spec
+	//  will not be inspectable.
```
wolffcm: We can look at the `plan.Spec`, right?

lang/compiler.go
affo: I think we discussed about this yesterday. Is it ok if I leave like this for now, and when we decide how to serialize TableObjects we change this bit?

wolffcm: I think it's fine for now.

It's unclear to me how having a `TableObjectCompiler` works with the changes that @jsternberg has in mind for the interpreter.  Jonathan, do you want to have a look at these changes?

affo:
__UPDATE__

@wolffcm The planning is now controllable by means of CompilerOptions and it can be stopped at some stages for testing.

Tests use this feature to inspect the plan.Spec at different planning Phases.

affo:
> Don't merge the `flux` binary!!

Aaaaargh! Removed! Victim of `add -A`

plan/physical.go
```diff
@@ -127,6 +127,9 @@ func WithDefaultMemoryLimit(memBytes int64) PhysicalOption {
 func OnlyPhysicalRules(rules ...Rule) PhysicalOption {
 	return physicalOption(func(pp *physicalPlanner) {
 		pp.clearRules()
+		// Always add physicalConverterRule. It doesn't change the plan but only convert nodes to physical.
+		// This is required for some pieces to work on the physical plan (e.g. SetTriggerSpec).
+		pp.addRules(physicalConverterRule{})
```
affo: I am sorry, I haven't seen that those changes remained the same from my previous push.
I added them because of how i managed tests before. @wolffcm do you want me to keep them?

plan/plantest/cmp.go
```diff
@@ -6,11 +6,25 @@ import (
...
+var CmpOptions = append(
+	semantictest.CmpOptions,
+	cmp.AllowUnexported(flux.Spec{}),
+	cmp.AllowUnexported(universe.JoinOpSpec{}),
+	cmpopts.IgnoreUnexported(flux.Spec{}),
+	cmpopts.IgnoreUnexported(universe.JoinOpSpec{}),
+	cmp.AllowUnexported(kafka.ToKafkaProcedureSpec{}),
+	cmpopts.IgnoreUnexported(kafka.ToKafkaProcedureSpec{}),
+)
+
```
affo: The same here

plan/physical.go
affo: This one seems pretty logical, because `physicalConverterRule` can't be added manually once removed.

plan/plantest/cmp.go
affo: Created a PR for it to merge to master #1139 

I am sorry for the mess 😅

PR #1115, opened 2019-04-02
refactor(control): refactor the controller to remove references to the spec and use program

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


control/controller.go
```diff
@@ -377,32 +323,37 @@ func (q *Query) Cancel() {
...
-	return q.ready
+	return q.results
 }
 
 // Done signals to the Controller that this query is no longer
 // being used and resources related to the query may be freed.
 //
+// The Results method must have been completely read before calling
+// this method either by the query executing, being canceled, or
+// an error occurring.
```
affo: I think you wanted to remove the comment below.  

Anyhow, why should the result be consumed before calling Done()?
Is it because `pump` would block?

Couldn't `Done` call `cancel` on `currentCtx`, so that pump stops consuming `exec.Results()`. Maybe, `pump` would also block on sending the result on `q.results`...

Could you provide a `ConsumeResults(q Query, fn func(flux.Result) error) error` as a helper to summarize the logic of consuming a result by applying a function and check for `q.Err()` and calling `q.Done()`?

control/controller.go
```diff
@@ -580,37 +516,59 @@ func (q *Query) setErr(err error) {
...
+			q.results <- res
+		case <-done:
+			// Signal to the underlying executor that the query
+			// has been canceled. Usually, the signal on the context
+			// is likely enough, but this explicitly signals just in case.
+			exec.Cancel()
+
+			// Set the done channel to nil so we don't do this again
+			// and we continue to drain the results.
+			done = nil
```
affo: If the query has been canceled, shouldn't we break out of the loop?

control/controller.go
```diff
@@ -84,228 +69,202 @@ func New(c Config) *Controller {
...
-		// Update resource gauges
-		c.consume(q)
+func (c *Controller) enqueueQuery(q *Query) error {
+	if !q.tryQueue() {
+		return errors.New("failed to transition query to queueing state")
+	}
 
-		// Remove the query from the queue
-		pop = true
+	// TODO(jsternberg): We should have a real queue! It should only
```
affo: Why don't we keep the code for the real queue and we use it when switching to execute?
This would require minimum execution overhead, but would prevent us to reimplement the queue logic in a second time.

control/controller_test.go
```diff
@@ -1,682 +1,225 @@
...
-	// The state should be canceled.
-	if want, got := Canceled, q.(*Query).State(); want != got {
-		t.Errorf("unexpected state: want=%s got=%s", want, got)
-	}
-
-	// Now finish the query by using Done.
-	q.Done()
+func TestController_ShutdownWithRunningQuery(t *testing.T) {
+	ctrl := control.New(control.Config{})
+	defer shutdown(t, ctrl)
```
affo: do you really need this?

control/controller.go
jsternberg: I think I messed this up. I'll make it so done will cancel the pumping of results.

control/controller.go
jsternberg: The reason for this is because cancel signals that the loop should be broken out of and since we are wrapping the query from the program that is being executed, we need to rely on when that query says we are done and not ourselves. So this signals to the inner query that the results have been canceled, but continues pumping the results.

control/controller.go
jsternberg: The code for the real queue is something we probably want to redesign so I thought it would be too complicated to keep it. Most of this logical is about resource checking which we aren't keeping.

The new queue is probably going to be "check if we have the minimum resources for a task, pop off the first element in the queue, then start it. block if there aren't enough resources." We might decide to do more, but that's even further in the future as there's no designed way for us to request resource allocations or limits yet.

control/controller_test.go
jsternberg: The defer? It's a failsafe in case it fails somewhere else. Using shutdown multiple times should be a noop.

control/controller.go
```diff
@@ -580,37 +509,68 @@ func (q *Query) setErr(err error) {
...
+			}
+
+			// It is possible for the underlying query to misbehave.
+			// We have to continue pumping results even if this is the
+			// case, but if the query has been canceled or finished with
+			// done, nobody is going to read these values so we need
+			// to avoid blocking.
+			select {
+			case <-q.currentCtx.Done():
+			case q.results <- res:
```
affo: great

control/controller.go
```diff
@@ -84,228 +69,202 @@ func New(c Config) *Controller {
...
+		id:                 id,
+		labelValues:        labelValues,
+		compileLabelValues: compileLabelValues,
+		state:              Created,
+		c:                  c,
+		results:            make(chan flux.Result),
+		parentCtx:          parentCtx,
+		parentSpan:         parentSpan,
+		cancel:             cancel,
+	}
```
affo: Wouldn't it be necessary to pass `cctx` as `currentCtx` here?

jsternberg: The `currentCtx` gets set by the state transitions. The `cctx` is passed as it is the parent of `parentCtx`.

affo: Oh great, I haven't noticed that! Thank you

PR #1125, opened 2019-04-03
update runtime and parser to recognize a nil type

aanthony1243:
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -107,6 +107,7 @@ func (*Property) node()   {}
 func (*Identifier) node() {}
 
 func (*BooleanLiteral) node()         {}
+func (*NilLiteral) node()             {}
 func (*DateTimeLiteral) node()        {}
 func (*DurationLiteral) node()        {}
 func (*FloatLiteral) node()           {}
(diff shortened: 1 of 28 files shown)


wolffcm:
I'd advocate against using `==` or `!=` to detect null values, at least in the final product.  If `null` truly represents missing data, the proper result if either operand is `null` would be a boolean-valued `null`.

SQL implementations use a three-valued logic and provides an `is null` operator; we might do the same.  There is a good discussion of it here:
https://modern-sql.com/concept/three-valued-logic

aanthony1243:
@wolffcm yeah, I had a discussion with  nathaniel about this same issue.  fwiw this was a useful exploration and parts of it needed to be done anyway becuase panics in row functions were happening not becuase we were operating on null values, but becuase of how we were decoding them at compile time.  This code includes that fix (in runtime.go) and that code should probably be isolated and merged back.  there may be some other bits that are similarly useful.  

I suppose nobody doubted that we could handle nulls in Flux, but I feel like a concrete proof of concept will move us closer to a final solution.

Issue #1126, opened 2019-04-04
[pvt-spec] Make new Compiler assign dependencies to Programs it generates


wolffcm:
On master, currently the controller gets its `execute.Dependencies` assigned from its config object and propagates them down to the `Executor`.

In the new architecture, the `Compiler` produces a `Program` that can be `Start`ed.  The call to `Start` instantiates a new `Executor`, which needs dependencies to know how to access the services that implement `from`, `to`, etc.

jsternberg:
I think this is the same or related to #1140. The compiler design makes it difficult to inject stuff into the chain so, for example, these dependencies are needed when running the AST compiler, but shouldn't need to be passed down from the compilers that use AST compiler. We also need a way to be able to send in options to the compiler.

jsternberg:
We're thinking that this already has a workaround and the proper solution to this can be done later so I'm throwing this back to planned. We are likely not going to do this before merging the private spec changes.

PR #1128, opened 2019-04-04
refactor(lang): run the executor for a program in a goroutine

### LINKED ISSUE #1119: Program should execute in a new goroutine
Currently on the `refactor/private-spec` branch, `lang.Program.Start` does not execute in a new goroutine:
https://github.com/influxdata/flux/blob/refactor/private-spec/lang/compiler.go#L134-L150

We should change this code:
- Make the call to `e.Execute` happen inside the goroutine
- Return any error via a channel in `lang.Query`
- Add tests that verify this works right

jsternberg:
- [x] docs/SPEC.md updated
- [ ] Test cases written

The execution of a program now happens in a goroutine and correctly
responds to the cancel signal.

The program and query implementations are also now hidden since they are
not used by anyone outside of lang.

Fixes #1119.

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -55,7 +55,7 @@ func (c FluxCompiler) Compile(ctx context.Context) (flux.Program, error) {
 		return nil, err
 	}
 
-	return Program{
+	return &Program{
 		PlanSpec: ps,
 	}, err
 }
@@ -76,7 +76,7 @@ func (c SpecCompiler) Compile(ctx context.Context) (flux.Program, error) {
 		return nil, err
 	}
 
-	return Program{
+	return &Program{
 		PlanSpec: ps,
 	}, err
 }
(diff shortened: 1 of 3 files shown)


lang/compiler.go
```diff
@@ -206,34 +206,49 @@ func buildSpec(t *flux.TableObject, ider flux.IDer, spec *flux.Spec, visited map
 	spec.Operations = append(spec.Operations, t.Operation(ider))
 }
 
-// Program implements the flux.Program interface
-type Program struct {
+// program implements the flux.Program interface
+type program struct {
 	deps     execute.Dependencies
```
wolffcm: One thing that I have been dealing with in getting tests to pass in `influxdb` is that the dependencies need to be set so that the executor knows how to execute `from`, `to`, etc.  I filed this issue for it: #1126

In the meantime I have been working around it by making the controller assign the dependencies.  But my workaround won't work if you make `lang.Program` unexported.  Would you mind keeping it public for now, until #1126 is fixed?

PR #1137, opened 2019-04-08
refactor(lang): rename SpecCompiler to ReplCompiler

jlapacik:
Closes #13220 .

Once https://github.com/influxdata/influxdb/pull/13222 is merged, the flux version can be updated on the `refactor/private-spec` branch.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -17,7 +17,6 @@ import (
 
 const (
 	FluxCompilerType = "flux"
-	SpecCompilerType = "spec"
 	ASTCompilerType  = "ast"
 )
 
@@ -35,9 +34,7 @@ func AddCompilerMappings(mappings flux.CompilerMappings) error {
 	}); err != nil {
 		return err
 	}
-	return mappings.Add(SpecCompilerType, func() flux.Compiler {
-		return new(SpecCompiler)
-	})
+	return nil
 }
 
 // CompileOption represents an option for compilation.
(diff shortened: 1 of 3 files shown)


lang/compiler.go
```diff
@@ -64,14 +64,27 @@ func (c FluxCompiler) CompilerType() flux.CompilerType {
 	return FluxCompilerType
 }
 
-// SpecCompiler implements Compiler by returning a known spec.
-type SpecCompiler struct {
+// ReplCompiler is what was formerly known to be a SpecCompiler.
+// It was explicitly renamed to denote its use in the Flux REPL and only the Flux REPL.
```
wolffcm: _nit_ Coding convention in Go is to use call caps for abbreviations, so this would be `REPLCompiler`.

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

lang/compiler.go
```diff
@@ -81,10 +94,6 @@ func (c SpecCompiler) Compile(ctx context.Context) (flux.Program, error) {
 	}, err
 }
 
-func (c SpecCompiler) CompilerType() flux.CompilerType {
```
wolffcm: Is this really dead code?  Will the `influxdb` REPL work if you remove this?

wolffcm: Oops,  I see that you just moved it.

jlapacik:
@wolffcm I removed the repl compiler into the repl package and made it private.

PR #1142, opened 2019-04-09
refactor(stdlib/aggregates): aggregates now accept only a 'column' pa…

### LINKED ISSUE #336: Should be able to use selector functions with the aggregateWindow function
Selectors are different from aggregators and as such the aggregateWindow function does not allow selectors functions to be passed in.

Maybe we should update the function signatures to make it possible to do so.

This way a user can easily switch from `aggregateWindow(fn:mean)` to `aggregateWindow(fn:max)`.

aanthony1243:
…rameter, columns not used.

BREAKING CHANGE: aggregates no longer accept multiple columns to aggregate, but rather a single 'column' parameter.

spec updated with new paramter set for aggregates
add test demonstrating use of max in aggregateWindow

closes #336.

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1447,9 +1447,9 @@ Any output table will have the following properties:
 
 All aggregate operations have the following properties:
 
-| Name    | Type     | Description                                       |
-| ----    | ----     | -----------                                       |
-| columns | []string | Columns specifies a list of columns to aggregate. |
+| Name   | Type   | Description                             |
+| ----   | ----   | -----------                             |
+| column | string | Column specifies a column to aggregate. |
 
 ##### AggregateWindow
 
@@ -1460,14 +1460,14 @@ an output table for every input table.
 AggregateWindow has the following properties:
 
 
-| Name        | Type                                            | Description                                                                                                           |
-| ----        | ----                                            | -----------                                                                                                           |
-| every       | duration                                        | Every specifies the window size to aggregate.                                                                         |
-| fn          | (tables: <-stream, columns: []string) -> stream | Fn specifies the aggregate operation to perform. Any of the functions in this Aggregate section can be provided.      |
-| columns     | []string                                        | Columns specifies a list of columns to aggregate. Defaults to ["_value"].                                             |
-| timeSrc     | string                                          | TimeSrc is the name of a column from the group key to use as the source for the aggregated time. Defaults to "_stop". |
-| timeDst     | string                                          | TimeDst is the name of a new column in which the aggregated time is placed. Defaults to "_time".                      |
-| createEmpty | bool                                            | CreateEmpty, if true, will create empty windows and fill them with a null aggregate value.  Defaults to true.         |
+| Name        | Type                                            | Description                                                                                                                                                     |
+| ----        | ----                                            | -----------                                                                                                                                                     |
+| every       | duration                                        | Every specifies the window size to aggregate.                                                                                                                   |
+| fn          | (tables: <-stream, columns: []string) -> stream | Fn specifies the aggregate operation to perform. Any of the functions in this Aggregate section that accept a singular `column` parameter can be provided.      |
+| column      | string                                          | Columns specifies column to aggregate. Defaults to "_value".                                                                                                    |
+| timeSrc     | string                                          | TimeSrc is the name of a column from the group key to use as the source for the aggregated time. Defaults to "_stop".                                           |
+| timeDst     | string                                          | TimeDst is the name of a new column in which the aggregated time is placed. Defaults to "_time".                                                                |
+| createEmpty | bool                                            | CreateEmpty, if true, will create empty windows and fill them with a null aggregate value.  Defaults to true.                                                   |
 Example:
 
 ```
@@ -1485,11 +1485,11 @@ Covariance computes the covariance between two columns.
 
 Covariance has the following properties:
 
-| Name     | Type     | Description                                                                                 |
-| ----     | ----     | -----------                                                                                 |
-| columns  | []string | Columns specifies a list of columns to aggregate. Defaults to `["_value"]`.                 |
-| pearsonr | bool     | Pearsonr indicates whether the result should be normalized to be the Pearson R coefficient. |
-| valueDst | string   | ValueDst is the column into which the result will be placed. Defaults to `_value`.          |
+| Name     | Type     | Description                                                                                            |
+| ----     | ----     | -----------                                                                                            |
+| columns  | []string | Columns specifies a list of the two columns to aggregate. This property is required and has no default.|
+| pearsonr | bool     | Pearsonr indicates whether the result should be normalized to be the Pearson R coefficient.            |
+| valueDst | string   | ValueDst is the column into which the result will be placed. Defaults to `_value`.                     |
 
 Additionally exactly two columns must be provided to the `columns` property.
 
@@ -1530,9 +1530,9 @@ For each aggregated column, it outputs the number of records as an integer. It w
 
 Count has the following property:
 
-| Name    | Type     | Description                                                                 |
-| ----    | ----     | -----------                                                                 |
-| columns | []string | Columns specifies a list of columns to aggregate. Defaults to `["_value"]`. |
+| Name    | Type     | Description                                                      |
+| ----    | ----     | -----------                                                      |
+| column  | string   | Columns specifies a column to aggregate. Defaults to `"_value"`. |
 
 Example:
 ```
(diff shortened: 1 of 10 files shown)


stdlib/flux_test.go
```diff
@@ -34,6 +34,7 @@ var skip = map[string]string{
 	"task_per_line":               "join produces inconsistent/racy results when table schemas do not match (https://github.com/influxdata/flux/issues/855)",
 	"rowfn_with_import":           "imported libraries are not visible in user-defined functions (https://github.com/influxdata/flux/issues/1000)",
 	"string_trim":                 "cannot reference a package function from within a row function",
+	"integral_columns":            "aggregates changed to operate on just a single columnm.",
```
nathanielc: Should we update integral to operate on a single column as well?

stdlib/universe/quantile_test.go
```diff
@@ -246,7 +203,7 @@ func TestQuantile_NewQuery(t *testing.T) {
 		{
 			Name:    "aggregate with column",
 			Raw:     `from(bucket:"testdb") |> range(start: -1h) |> quantile(q: 0.99, method: "estimate_tdigest", column: "1")`,
-			WantErr: true,
+			WantErr: false,
```
affo: This should be like the test you deleted above, so something like `with custom column` and an expected result.

stdlib/universe/quantile.go
```diff
@@ -38,8 +38,7 @@ type QuantileOpSpec struct {
 func init() {
 	quantileSignature := flux.FunctionSignature(
 		map[string]semantic.PolyType{
-			"column":      semantic.String,                            // selector
-			"columns":     semantic.NewArrayPolyType(semantic.String), // aggregate
+			"column":      semantic.String, // selector
```
nathanielc: The `// selector` is no longer relevant.

execute/aggregate.go
```diff
@@ -22,7 +21,7 @@ type aggregateTransformation struct {
 
 type AggregateConfig struct {
 	plan.DefaultCost
-	Columns []string `json:"columns"`
+	Columns []string `json:"column"`
```
affo: Why is this a slice?

docs/SPEC.md
```diff
@@ -1487,7 +1487,7 @@ Covariance has the following properties:
 
 | Name     | Type     | Description                                                                                 |
 | ----     | ----     | -----------                                                                                 |
-| columns  | []string | Columns specifies a list of columns to aggregate. Defaults to `["_value"]`.                 |
+| columns  | []string | Columns specifies a list of the two columns to aggregate. This property is required and has no default.                 |
```
affo: I would say that, as such, `covariance` is no more an aggregate? Because it cannot be used in `aggregateWindow`. Should we provide a `covarianceBuilder` that given two columns, provides an aggregate? Like a function generator, idk 🤔

affo:
```
covGen(col1, col2) = (column: col1) => covariance(columns: [col1, col2])
```

affo:
Looks great 😎 I put some comments around.

Another thing: I know it is boring, but can you give better formatting to tables for params in SPEC? Idk if there is some tool around that does it. I usually do it bu hand 😅

stdlib/flux_test.go
aanthony1243: that's what I did.  this is a skip-test list.  I thought we could bring this back later if we deliver on a multi-aggregate function.

execute/aggregate.go
aanthony1243: I had an idea to use this code to implement a multi-aggregate result (e.g. to computer count + sum in one pass).  so I left it as a slice for now to make that work easier to complete.

docs/SPEC.md
aanthony1243: I agree.  but I didn't know how else to categorize it.  in one way, it is an aggregator because it is combining points into a single value, but in the other way it is not because its function signature is different.  I thought about using a function generator pattern to make it fit, but I couldn't think of a good reason to hard-code one parameter.  

something like  `covGen(column: col1) = (column: col2) => covariance(columns: [col1, col2])` could work but the user is free to create this function if they need it.

execute/aggregate.go
affo: 👍

docs/SPEC.md
```diff
@@ -1464,7 +1464,7 @@ AggregateWindow has the following properties:
 | ----        | ----                                            | -----------                                                                                                           |
 | every       | duration                                        | Every specifies the window size to aggregate.                                                                         |
 | fn          | (tables: <-stream, columns: []string) -> stream | Fn specifies the aggregate operation to perform. Any of the functions in this Aggregate section can be provided.      |
-| columns     | []string                                        | Columns specifies a list of columns to aggregate. Defaults to ["_value"].                                             |
```
affo: Nit: in the line above, we say:

> Any of the functions in this Aggregate section can be provided.

This is not true. Covariance is here, but cannot be provided. It could be changed to:

```
Any aggregate function that accepts a `column` property can be provided.
```

aanthony1243: good point.  will fix.

zarbis:
@aanthony1243 I've recently upgraded InfluxDB 1.7.7 => 1.7.8 (Flux 0.24 => 0.32) and faced this change. I get it's noble purpose but two moments are unclear, at least from docs, without experimenting on data:
1) how would it interact with tables with several value column? Put sum of `column` values into single first/last cell of that column and leave the rest as-is?
```

+-----+-----+------+
| foo | bar | baz  |
+-----+-----+------+
|   1 |   1 |    1 |
|   2 |   2 |    2 |
|   3 |   3 |    3 |
+-----+-----+------+
```
`|> sum(column: "bar)`
```
+-----+-----+------+
| foo | bar | baz  |
+-----+-----+------+
|   1 |   6 |    1 |
|   2 |     |    2 |
|   3 |     |    3 |
+-----+-----+------+
```
2) what is idiomatic way to modify my legacy `|> sum(columns: ["foo", "bar", "baz"])` statements? Chain several `sum()`s?

sanderson:
@zarbis, aggregate functions drop any columns that are not part of the group key or defined as the aggregate column. In the case above, the `foo` and `baz` columns would get dropped.

To sum multiple columns at once, you could use [`reduce()`](https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/reduce/) to create a custom aggregate.

```js
// ...
|> reduce(
  identity: {foo_sum: 0, bar_sum: 0, baz_sum: 0},
  fn:  (r, accumulator) => ({
    foo_sum: r.foo + accumulator.foo_sum,
    bar_sum: r.bar + accumulator.bar_sum,
    baz_sum: r.baz + accumulator.baz_sum
  }))
|> rename(columns: {foo_sum: "foo", bar_sum: "bar", baz_sum: "baz"})
```

zarbis:
@sanderson hm, may be it's stepping into separate issue territory, but I've replaced `sum()` with `reduce()` and I'm getting `panic: unknown type invalid`

```js
...
result = union(tables: [real_expense, profit])
|> group(columns: ["colname", "utm_campaign"])
|> pivot(rowKey: ["utm_campaign"], columnKey: ["colname"], valueColumn: "_value")
|> map(fn: (r) => ({
  utm_campaign: r.utm_campaign,
  profit: r.profit,
  expense: r.real_expense,
  profitability: (r.profit - r.real_expense) / r.real_expense,
}))

total = result
|> group()
|> reduce(
  identity: {profit_sum: 0.0, expense_sum: 0.0},
  fn: (r, accumulator) => ({
    profit_sum: r.profit + accumulator.profit_sum,
    expense_sum: r.expense + accumulator.expense_sum,
  }))
|> rename(columns: {profit_sum: "profit", expense_sum: "expense"})
|> set(key: "utm_campaign", value: "TOTAL")
```
And it is super slow, takes about 1 minute. `result` table is about 10-20 rows total and is produced in about second or two.

sanderson:
@zarbis The error system is in the process of an upgrade so hopefully soon it'll point to the specific line causing the error. I'm guessing that something in the calculations used in `map()` and/or `reduce()` is not a float. Try casting mathematical operands to floats:

```js
// ...

|> map(fn: (r) => ({
  utm_campaign: r.utm_campaign,
  profit: r.profit,
  expense: r.real_expense,
  profitability: (float(v: r.profit) - float(v: r.real_expense)) / float(v: r.real_expense),
}))

// ...

|> reduce(
  identity: {profit_sum: 0.0, expense_sum: 0.0},
  fn: (r, accumulator) => ({
    profit_sum: float(v: r.profit) + v: accumulator.profit_sum,
    expense_sum: float(v: r.expense) + accumulator.expense_sum,
  }))

//...

```

affo:
@zarbis thanks for signaling performance problems on union. I'll create a separate issue for that.

PR #1143, opened 2019-04-09
fix(lang): make Program.Start start execution synchronously

wolffcm:
If an error occurred when setting up execution state, it would not be reported.  This
was causing a test to fail for a query like `from(bucket:"non-existent")`.
The call the `Execute` just walks the plan, creates the execution state, and returns, allowing
query execution to continue asynchronously, so it's okay to start execution synchronously to allow
errors to be reported.

Fixes influxdata/idpe#3073.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jsternberg:
I'm rereviewing. I misread the pull request.

lang/compiler.go
```diff
@@ -215,37 +215,30 @@ type Program struct {
...
-	}
-	q.wg.Add(1)
-	go p.run(ctx, q)
-	return q, nil
-}
-
-func (p *Program) run(ctx context.Context, q *query) {
-	defer q.wg.Done()
-	defer close(q.results)
+	q := NewQuery(alloc, cancel)
```
jsternberg: I would like it if we didn't publicly expose the implementation of query here so can we revert this section back to what it was?

wolffcm: You're right.  I'll revert it.

PR #1152, opened 2019-04-11
Merge master into refactor/private-spec

wolffcm:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/examples/library/README.md
+++ b/examples/library/README.md
@@ -1,0 +1,39 @@
+# Flux as a Library Example
+
+This folder contains a [main.go](main.go) file that shows how flux can be used as a library in your programs.
+
+## Components
+
+The main components you need are:
+
+- The interpreter;
+- The scope (aka Prelude);
+- The builtin library and your additional functions, if you want to define them;
+- The language specification compiler;
+- A querier.
+
+## What does this example do?
+
+This example takes all the components described above from the Flux repo and uses them to
+compile, execute and print the results of this query.
+
+```flux
+import g "generate"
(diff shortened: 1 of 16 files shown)


examples/library/library_example_test.go
```diff
@@ -0,0 +1,62 @@
...
+	// #datatype,string,long,dateTime:RFC3339,long
+	// #group,false,false,false,false
+	// #default,_result,,,
+	// ,result,table,_time,_value
+	// ,,0,1993-02-16T00:00:00Z,1
+	// ,,0,1993-02-16T00:00:36Z,1
+	// ,,0,1993-02-16T00:01:12Z,1
+	// ,,0,1993-02-16T00:01:48Z,1
+	// ,,0,1993-02-16T00:02:24Z,1
+}
```
affo: Oh I see, you cannot get back an AST from the TableObject, so you bypassed the interpreter and directly used the compiler. This changes the nature of the example, but it's fine and thanks for doing it!

However the README is now inconsistent with the example. I'll comment on that

examples/library/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Components
+
+The main components you need are:
+
+- The interpreter;
+- The scope (aka Prelude);
+- The builtin library and your additional functions, if you want to define them;
+- The language specification compiler;
+- A querier.
+
```
affo: I think it's:

```
- A Flux Compiler to parse and generate the AST;
- A Querier to send the query expressed in the AST;
- A ResultEncoder to display results to humans;
```

affo:
Also note that I merged to master a change in the example that didn't pass `make vet`. We should also raise awareness on that.

PR #1153, opened 2019-04-11
test(lang): add tests for query

### LINKED ISSUE #1134: [pvt-spec] Write unit tests for new Query interface

affo:
Closes #1134 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/query_test.go
+++ b/lang/query_test.go
@@ -1,0 +1,168 @@
+package lang_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	_ "github.com/influxdata/flux/builtin"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/memory"
+	"github.com/pkg/errors"
+)
+
+func runQuery(script string) (flux.Query, error) {
+	program, err := lang.Compile(script, time.Unix(0, 0))
+	if err != nil {
+		return nil, err
+	}
+	q, err := program.Start(context.Background(), &memory.Allocator{})
+	if err != nil {
+		return nil, err
+	}
+	return q, nil
+}
+
+var validScript = `
+import "csv"
+
+data = "
+#datatype,string,long,long,string
+#group,false,false,false,true
+#default,_result,,,
+,result,table,value,tag
+,,0,10,a
+,,0,10,a
+,,1,20,b
+,,1,20,b
+,,2,30,c
+,,2,30,c
+,,3,40,d
+,,3,40,d
+"
+
+csv.from(csv: data) |> yield(name: "res")`
+
+func TestQuery_Results(t *testing.T) {
+	q, err := runQuery(validScript)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "unexpected error while creating query"))
+	}
+
+	// gather counts
+	var resCount, tableCount, totRows int
+	for res := range q.Results() {
+		resCount++
+		if err := res.Tables().Do(func(tbl flux.Table) error {
+			tableCount++
+			return tbl.Do(func(cr flux.ColReader) error {
+				tags := cr.Strings(1)
+				for i := 0; i < tags.Len(); i++ {
+					totRows++
+				}
+				return nil
+			})
+		}); err != nil {
+			t.Fatal(errors.Wrap(err, "unexpected error while iterating over tables"))
+		}
+	}
+
+	// com,pare expected counts
+	if resCount != 1 {
+		t.Errorf("got %d results instead of %d", resCount, 1)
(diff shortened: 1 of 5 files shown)


affo:
@wolffcm maybe I should add something for `Cancel()`.

I pushed what I add just to see if it meets you expectations when you opened the issue.

lang/query_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// release query resources
+	q.Done()
+
+	if q.Err() != nil {
+		t.Fatal(errors.Wrap(q.Err(), "unexpected error from query execution"))
+	}
+}
+
+func TestQuery_Stats(t *testing.T) {
+	t.Skip("stats are updated by the controller, running a standalone query won't update them")
```
wolffcm: Hmm, it occurs to me that because `lang.Query` needs to use an `execute.Executor`, this is hard to test.  One solution might be to add a method that lets clients decide which executor to use when running the program, in this interface:
https://github.com/influxdata/flux/blob/master/lang/compiler.go#L208
```go
type DependenciesAwareProgram interface {
	SetExecutorDependencies(execute.Dependencies)
	SetLogger(logger *zap.Logger)
        SetExecutor(e execute.Executor)
}
```
Then you could pass in a `mock.Executor` that updates the stats.

wolffcm: I am a little reluctant to add code that is only used for tests, but it seems like the benefits outweigh the cost here.

mock/query.go
```diff
@@ -0,0 +1,76 @@
...
+func (q *query) Cancel() {
+	q.Done()
+}
+
+func (q *query) Err() error {
+	return q.err
+}
+
+func (q *query) Statistics() flux.Statistics {
+	return flux.Statistics{}
```
wolffcm: It would be good to have a way for clients to specify what statistics they expect to be returned.  The logic for this is a little subtle, since statistics are only valid if `Done()` has been called.  What they are before that is undefined.

result_iterator_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+func TestQueryResultIterator_Error(t *testing.T) {
+	expectedErr := errors.New("hello, I am an error")
+	q := mock.QueryFromTablesWithError(expectedErr)
+	ri := flux.NewResultIteratorFromQuery(q)
+	defer ri.Release()
+
+	for ri.More() {
```
wolffcm: With result iterator, the code is a little subtle since the iterator is fed by a channel and you can't peek at the next value without consuming it.  So it would be good to verify that the following assumptions hold in this test:
- `More()` is idempotent
- Calling `Next()` will panic if there is no more data
- You can call `Next()` without calling `More()` first
- Calling `Statistics()` returns the zero value for `flux.Statistics` before `Release()` is called.

It would be nice to have a test that verifies that all the implementations if `ResultIterator` behave correctly with these assumptions, but that could be done as a separate issue.

wolffcm:
Oh, also, `refactor/private-spec` is a dead branch, so you can just rebase this commit on master.

lang/query_test.go
affo: I see your point. But, if I do it for this test, isn't it the same if the query itself returns the stats it wants to return?

I say this, because, for "real" queries, it is the controller that sets statistics.  
Moreover, when it comes to the executor, the query has already been compiled and planned, so some stats can't be updated (or they should be update with random values).

mock/query.go
affo: I think I'll unify the 2 implementations of query in package `mock` (https://github.com/influxdata/flux/blob/f94914ae407b2aa124e6378178f07e399d2dad1e/mock/program.go#L40), and make it more tunable.

lang/query_test.go
wolffcm: Hmm.  This is becoming a design discussion, that should involve our teammates.

Both the `Query` interface and the `ResultIterator` that wraps it are tricky, especially with regard to canceling and statistics, and so I wanted to ensure that we have coverage against breaking things, as well as a good example for clients to follow.

I think you're right that since the controller is the thing that updates the stats, maybe it's really the controller we want to test (or refactor like you mention).

wolffcm: I guess what I'm saying is if you want to merge this as-is, and then we can file other issues where we are missing test coverage, that would be fine.

affo: Yeah, I think this point is a bit tricky and worth a new issue and relative PR.

affo:
@wolffcm I think I addressed your comments (except for the controller/stats one).

affo:
Wah, I forgot to change the commit message 😓

PR #1154, opened 2019-04-11
fix(control): the controller pump could reference a nil pointer

jsternberg:
- [x] docs/SPEC.md updated
- [ ] Test cases written

This modifies the controller in two ways. First, it ensures the current
context is always set to a valid value. The second and more important is
that it treats the current context as something that has to be
synchronized and, instead of reading it from the query, each state
returns the context it will run in so there isn't a potential race
condition involved with its usage.

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -180,11 +180,12 @@ func (c *Controller) countQueryRequest(q *Query, result requestsLabel) {
 }
 
 func (c *Controller) compileQuery(q *Query, compiler flux.Compiler) error {
-	if !q.tryCompile() {
+	ctx, ok := q.tryCompile()
+	if !ok {
 		return errors.New("failed to transition query to compiling state")
 	}
 
-	prog, err := compiler.Compile(q.currentCtx)
+	prog, err := compiler.Compile(ctx)
 	if err != nil {
 		return errors.Wrap(err, "compilation failed")
 	}
@@ -199,7 +200,7 @@ func (c *Controller) compileQuery(q *Query, compiler flux.Compiler) error {
 }
 
 func (c *Controller) enqueueQuery(q *Query) error {
-	if !q.tryQueue() {
+	if _, ok := q.tryQueue(); !ok {
 		return errors.New("failed to transition query to queueing state")
 	}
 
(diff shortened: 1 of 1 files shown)


control/controller.go
```diff
@@ -540,40 +549,40 @@ func (q *Query) pump(exec flux.Query) {
...
-			done = nil
+			signalCh = nil
 		}
 	}
 }
 
 // tryCompile attempts to transition the query into the Compiling state.
-func (q *Query) tryCompile() bool {
+func (q *Query) tryCompile() (context.Context, bool) {
 	q.stateMu.Lock()
```
wolffcm: Should `stateMu` just be locked inside the `transitionTo` function instead?

control/controller.go
```diff
@@ -399,7 +405,7 @@ func (q *Query) State() State {
 // transitionTo will transition from one state to another. If a list of current states
 // is given, then the query must be in one of those states for the transition to succeed.
 // This method must be called with a lock and it must be called from within the run loop.
-func (q *Query) transitionTo(newState State, currentState ...State) bool {
+func (q *Query) transitionTo(newState State, currentState ...State) (context.Context, bool) {
```
wolffcm: It's not intuitive why you would both set `q.currentCtx` and return it here, so it would be good to add a comment to this effect.

control/controller.go
```diff
@@ -523,10 +529,13 @@ func (q *Query) setErr(err error) {
...
 // results channel.
-func (q *Query) pump(exec flux.Query) {
+func (q *Query) pump(exec flux.Query, done <-chan struct{}) {
 	defer close(q.results)
 
-	done := q.currentCtx.Done()
+	// When the done channel is closed, we should send a signal
+	// to the underlying query to tell it to stop. We copy this
+	// to a separate channel so this signal can be done once.
+	signalCh := done
```
wolffcm: I don't understand this comment. I must be missing something.  `signalCh` and `done` are always the same, so why have two different variables?

control/controller.go
jsternberg: Probably. It wasn't originally because `transitionTo` was sometimes called within a function that locked the mutex, but it no longer appears to do that. It's possible this should be modified to return an `error` too.

control/controller.go
jsternberg: I guess you're right that's not really needed anymore other than winking at the developer to create a race condition.

control/controller.go
jsternberg: It is modified below. `done` never changes but `signalCh` is modified to be `nil` after it receives the signal for the first time.

control/controller.go
jsternberg: When I went to modify this, I found that `transitionTo` invokes itself so it would be hard to put the lock in there without doing a larger change than I'm comfortable with so I'm going to leave this as-is.

control/controller.go
wolffcm: Ok, now I got it.  I wasn't clear on the semantics of reading from a `nil` channel.

jsternberg: Ah, ok. A nil channel will never return while a closed channel will instantly return.

PR #1164, opened 2019-04-13
feat(stdlib/universe): add functionality to test records for missing values.

### LINKED ISSUE #965: Implement an "is null" function or operator in Flux
We should have a way of testing to see if a value is null.  This could be done either with a new operator (like SQL does it) or with a function in the standard library.

aanthony1243:
added exists function to test row records for missing values

updated spec

closes #965 
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -28,17 +28,21 @@ type Func interface {
 
 type Evaluator interface {
 	Type() semantic.Type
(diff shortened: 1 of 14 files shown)


aanthony1243:
replaced by #1278

PR #1166, opened 2019-04-15
fix(ast): format with good parens

affo:
Closes https://github.com/influxdata/flux/issues/1012

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/ast/format.go
+++ b/ast/format.go
@@ -122,7 +122,32 @@ var opPrecedence = map[int]int{
 	getIntForLOp(OrOperator):  7,
 }
 
+// formatChildWithParens applies the generic rule for parenthesis (not for binary expressions).
 func (f *formatter) formatChildWithParens(parent, child Node) {
+	f.formatLeftChildWithParens(parent, child)
+}
+
+// formatLeftChildWithParens applies the generic rule for parenthesis to the left child of a binary expression.
+func (f *formatter) formatLeftChildWithParens(parent, child Node) {
+	pvp, pvc := getPrecedences(parent, child)
+	if needsParenthesis(pvp, pvc, false) {
+		f.formatNodeWithParens(child)
+	} else {
+		f.formatNode(child)
+	}
+}
+
+// formatRightChildWithParens applies the generic rule for parenthesis to the right child of a binary expression.
+func (f *formatter) formatRightChildWithParens(parent, child Node) {
+	pvp, pvc := getPrecedences(parent, child)
+	if needsParenthesis(pvp, pvc, true) {
+		f.formatNodeWithParens(child)
+	} else {
+		f.formatNode(child)
+	}
+}
+
+func getPrecedences(parent, child Node) (int, int) {
 	var pvp, pvc int
 	switch parent := parent.(type) {
 	case *BinaryExpression:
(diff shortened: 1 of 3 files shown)


affo:
@juliusv You are right and thanks for pointing it out.

`1 - (2 - 3)` is short for `1 - 1 * (2 - 3)`.

My algorithm worked for the second one but not for the first one. It required custom logic, because that is the only case in which you need to check AST composition to put parenthesis.

internal/parser/parser_test.go
```diff
@@ -2096,6 +2096,134 @@ a = 5.0
...
+								Right: &ast.IntegerLiteral{
+									BaseNode: base("1:10", "1:11"),
+									Value:    3,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
affo: I added these tests in the parser while studying AST generation and I thought about leaving them there.

Surely, they can't hurt.

ast/format_test.go
```diff
@@ -325,6 +328,29 @@ func TestFormat_Associativity(t *testing.T) {
...
+			shouldFail: true,
+		},
+		{
+			name:       "plus with unintended parens 2",
+			script:     `1 + (2 + 3)`,
+			shouldFail: true,
+		},
+		{
+			name:       "minus with unintended parens",
+			script:     `(1 - 2) - 3`,
```
jlapacik: Could you add a test case for `1 - 2 - 3`?

internal/parser/parser_test.go
jlapacik: Definitely

ast/format.go
```diff
@@ -122,7 +122,59 @@ var opPrecedence = map[int]int{
 	getIntForLOp(OrOperator):  7,
 }
 
+// About parenthesis:
+// As a golden rule, we need parenthesis if a child node has lower precedence (bigger value) then its parent node.
```
jlapacik: nit: then -> than

ast/format.go
```diff
@@ -122,7 +122,59 @@ var opPrecedence = map[int]int{
...
+	f.formatLeftChildWithParens(parent, child)
+}
+
+// formatLeftChildWithParens applies the golden rule for parenthesis.
+func (f *formatter) formatLeftChildWithParens(parent, child Node) {
+	pvp, pvc := getPrecedences(parent, child)
+	// If one of parent or child has not matched any case, then we shouldn't apply any parenthesis.
+	par := !(pvc == 0 || pvp == 0)
+	// We need a parenthesis if parent and child have matched, and if the child node
+	// has lower precedence (bigger value) then the current node.
```
jlapacik: nit: then -> than

ast/format.go
```diff
@@ -154,18 +179,37 @@ func (f *formatter) formatChildWithParens(parent, child Node) {
...
+// (C) - When there are parenthesis, instead, the parser recursively generates a AST for the expression contained
+// in the parenthesis, and makes it the right child.
+// So, when formatting:
+//  - if we encounter a child with lower precedence on the left, this means it requires parenthesis, because, for sure,
+//    the parser detected parenthesis to break (A);
+//  - if we encounter a child with higher or equal precedence on the left, it doesn't need parenthesis, because
+//    that was the natural parsing order of elements (see (B));
+//  - if we encounter a child with lower or equal precedence on the right, it requires parenthesis, otherwise, it
+//    would have been at root (see (C)).
+func needsPar(pvp, pvc int, isRight bool) bool {
```
jsternberg: Can you rename this something that's more descriptive? Maybe just spell out the entire word.

PR #1174, opened 2019-04-16
feat(universe): add stream table index functions

### LINKED ISSUE #1089: Create stream/table index functions
We need a class of functions that allows a user to index into a stream of tables and the tables themselves to select either a scalar value, a row record or a column slice.

These functions will depend on #1088.

We likely need at least these three functions:

```
selectTable // selects a specific table from a stream
sliceColumn // given a single table produce an array of column values
sliceRow // given a single table produce a record for the row
scalarValue // given a table produce the scalar value at a specific cell in the table
```

We need to design out these functions, using numeric indexes is likely prone to creating brittle code as it will depend on table, column and row order being stable. We should use map lookup like interfaces where possible, for example the sliceColumn function should accept a column label name, not a numerical index. For tables inside a stream we should use the group key concept. For rows I can't think of anything but a numeric index or a predicate search function.

The above functions names are not final, when we have a design that works with the expected types we can determine better names.

The way I expect these function to work is once #1088 is complete then these function can consume a table object as an argument, execute the table object and use that as the input value. Maybe on the selectTable needs to do that part while the rest of the functions operate on pure values.Value types. Should we create easy to use implementation of a `table` type that represents a table as a list of records?

Also note that we will eventually want syntax sugar for these function but for now that work is out of scope of this issue.

This is part of #1085

### LINKED ISSUE #1168: Implement record function
See https://github.com/influxdata/flux/issues/1089

### LINKED ISSUE #1169: Implement columnSlice function
See https://github.com/influxdata/flux/issues/1089

affo:
Closes #1089 
Closes #1168 
Closes #1169 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -3147,8 +3147,102 @@ Contains has the following parameters:
 | value   | bool, int, uint, float, string, time          | The value to search for.     |
 | set     | array of bool, int, uint, float, string, time | The set of values to search. |
 
+Example: 
+    `contains(value:1, set:[1,2,3])` will return `true`.
+
+#### Stream/table functions
+
+These functions allow to extract a table from a stream of tables (`tableFind`) and access its
+columns (`getColumn`) and records (`getRecord`).  
+The example below provides an overview of these functions, further information can be found in the paragraphs below.
+
+```
+data = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+
+// extract the first available table for which "_field" is equal to "usage_idle"
+t = data |> tableFind(fn: (key) => key._field == "usage_idle")
+// this is the same as 'tableFind(tables: data, fn: (key) => key._field == "usage_idle")'
+
+// extract the "_value" column from the table
+values = t |> getColumn(column: "_value")
+// the same as 'getColumn(table: t, column: "_value")'
+
+// extract the first record from the table
+r0 = t |> getRecord(idx: 0)
+// the same as 'getRecord(table: t, idx: 0)'
+```
+
+##### TableFind
+
+TableFind extracts the first table in a stream of table whose group key values match a given predicate. If no table
+is found, the function errors.
+
+It has the following parameters:
+
+| Name | Type                  | Description                                                                     |
+| ---- | ----                  | -----------                                                                     |
+| fn   | (key: object) -> bool | Fn is a predicate function. The result is the first table for which fn is true. |
+
 Example:
-    `contains(value:1, set:[1,2,3])` will return `true`.   
+
+```
+t = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> tableFind(fn: (key) => key._field == "usage_idle")
+
+// `t` contains the first table whose group key value for `_field` is "usage_idle" after filtering.
+// `t` can be used as input to `getColumn` and `getRecord`.
+```
+
+##### GetColumn
+
+GetColumn extracts a column from a table given its label. If the label is not present in the set of columns,
+the function errors.
+
+It has the following parameters:
+
+| Name   | Type   | Description                                  |
+| ----   | ----   | -----------                                  |
+| column | string | Column is the name of the column to extract. |
+
+Example:
+
+```
+vs = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> tableFind(fn: (key) => key._field == "usage_idle")
+    |> getColumn(column: "_value")
+
+// use values
+x = vs[0] + vs[1]
+```
+
+##### GetRecord
+
+GetRecord extracts a record from a table given its index. If the index is out of bounds, the function errors.
+
+It has the following parameters:
+
+| Name | Type | Description                                |
+| ---- | ---- | -----------                                |
+| idx  | int  | Idx is the index of the record to extract. |
+
+Example:
+
+```
+r0 = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> tableFind(fn: (key) => key._field == "usage_idle")
+    |> getRecord(idx: 0)
+
+// use values
+x = r0._field + "--" + r0._measurement
+```
 
 #### Type conversion operations
 
--- a/stdlib/universe/table_fns.go
+++ b/stdlib/universe/table_fns.go
@@ -1,0 +1,309 @@
+package universe
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/semantic"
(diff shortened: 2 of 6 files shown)


values/table.go
```diff
@@ -0,0 +1,44 @@
...
+	TableMonoType semantic.Type
+)
+
+func init() {
+	TableMonoType, _ = TableType.MonoType()
+}
+
+// Table is a values.Value that represent a table.
+// (Unlike flux.TableObject which represents a stream of tables.)
+type Table struct {
```
jsternberg: It's probably a bad idea to copy this. If we're going to do things this way, I think it would be better to have this act as utility functions that operate on a `flux.Table` and implements the `values.Object` interface.

Also, this specific implementation is row-based, but all of our representations of a table are column-based.

affo: Yeah, totally agree with you. I am implementing with a wrapper for flux.Table right now.

affo:
Closes #1168 
Closes #1169

affo:
I still have to write down the SPEC for this.

stdlib/universe/table_fns.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	fn, err := arguments.GetRequiredFunction(tableFindFunctionArg)
+	if err != nil {
+		return nil, fmt.Errorf("missing argument: %s", tableFindFunctionArg)
+	}
+
+	c := lang.TableObjectCompiler{
+		Tables: to,
+		Now:    time.Now(),
```
affo: Which time this should be?

stdlib/universe/table_fns.go
```diff
@@ -0,0 +1,445 @@
...
+func arrayFromColumn(idx int, cr flux.ColReader) values.Array {
+	typ := cr.Cols()[idx].Type
+	vsSlice := make([]values.Value, 0, cr.Len())
+	for i := 0; i < cr.Len(); i++ {
+		switch typ {
+		case flux.TString:
+			if vs := cr.Strings(idx); vs.IsValid(i) {
+				vsSlice = append(vsSlice, values.New(vs.ValueString(i)))
+			} else {
+				vsSlice = append(vsSlice, values.NewNull(semantic.String))
```
aanthony1243: @nathanielc should we be skipping nulls in this case?

stdlib/universe/table_fns.go
```diff
@@ -0,0 +1,445 @@
...
+		}
+		r = objectFromRow(int(rowIdx), cr)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return r, nil
+}
+
+func objectFromRow(idx int, cr flux.ColReader) values.Object {
```
aanthony1243: replace this with `execute.GetRow`

aanthony1243:
Overall, I think there's a few points worth discussing here, that I just didn't have a good feel for until I read the spec and had questions.  it's worth taking the time to get this right.

docs/SPEC.md
```diff
@@ -3124,7 +3124,101 @@ Contains has the following parameters:
...
+Example:
+
+```
+t = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> tableFind(fn: (key) => key._field == "usage_idle")
+
+// explore the schema
+firstColLabel = t.schema[0].label
```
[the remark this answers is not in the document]
affo: I see your point. I considered that too!  

This implementation is coherent with how we represent the schema internally though.

One thing I can see in the future is iterating over objects/arrays. One should be able to explore the schema of a table programmatically in Flux. If we decided for objects, one should be able to iterate over objects

```
for label, colMeta in t.schema {
   stdout.Print(label + " - " + colMeta.type)
}
```

stdlib/universe/table_fns.go
affo: Thank you, I suppose there was an util for this, but I couldn't find that 😅

stdlib/universe/index.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/influxdata/flux/values"
+	"github.com/pkg/errors"
+)
+
+const (
+	streamArg   = "tables"
+	functionArg = "fn"
+)
+
+func init() {
```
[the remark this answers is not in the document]
affo: I don't know if this problem is still present in the new version of the code. I resolve the conversation for now. If you see the same problem again, please tell me :)

docs/SPEC.md
```diff
@@ -3124,7 +3124,101 @@ Contains has the following parameters:
...
+| ---- | ----                  | -----------                                                                     |
+| fn   | (key: object) -> bool | Fn is a predicate function. The result is the first table for which fn is true. |
+
+Example:
+
+```
+t = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> tableFind(fn: (key) => key._field == "usage_idle")
```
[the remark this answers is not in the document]
affo: `headTable()` seems great to have! And it should be also easy to express in Flux:

```
headTable = (tables=<-) => tables |> tableFind(fn: (key) => true)
```

How does the planner behave when encountering a function wrt a transformation?  
Are you sure that would be its behavior?  
Wouldn't the interpreter provide a `TableObject` representing `from ... filter` and then pass it to the `tableFind` function?  
I don't think `tableFind` would be taken into consideration by the planner. It is not a transformation.

stdlib/universe/table_fns.go
affo: I can't find `GetRow`. We have `ReadRow` but that doesn't return an `values.Object` actually. I could move this function as an util in `execute` or create a `helpers.go` in package `values` that contains helpers that convert from `ColReader` to `values.Value` instances.

stdlib/universe/index.go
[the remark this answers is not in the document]
affo: I pushed new changes that use values.NewFunction

affo:
Updated to use `values.NewFunction`.

docs/SPEC.md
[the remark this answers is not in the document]
aanthony1243: The planner is extensible and we would have to provide a matching pattern and a rewrite rule to make it work.  But there's no limitation in the planner against doing this.   

We don't have to do this right away.  probably we don't have to change the spec here either.

stdlib/universe/table_fns.go
aanthony1243: hm, yeah. nevermind.

docs/SPEC.md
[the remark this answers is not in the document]
aanthony1243: I think we'll likely cover iteration with a combination of `map` `reduce` and recursion.  furthermore, it shouldn't be hard to define a schema object as an iterable.  

it goes further than this, though.  If we do it this way, we need to define the order for the user.  That order needs to be consistent (sorted by key?) so that they can reliably index the schema to get what they need.  I think we should strongly consider refactoring this to be a map/object instead of an array.

affo:
> That order needs to be consistent (sorted by key?) so that they can reliably index the schema to get what they need. I think we should strongly consider refactoring this to be a map/object instead of an array.

I do agree with you @aanthony1243 . It seems also more natural to me. I don't know if we can merge as it is and later apply the change, or directly make this PR include the schema-as-an-object version.

docs/SPEC.md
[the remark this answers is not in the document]
affo: I do agree with you @aanthony1243 . It seems also more natural to me. I don't know if we can merge as it is and later apply the change, or directly make this PR include the schema-as-an-object version

PR #1175, opened 2019-04-16
fix(parser): add error handling for property list

### LINKED ISSUE #1102: Parser merges two separate statements
When parsing:

```
not (a and b)
(a or b) and c
```

The parser parses a single statement:

```
unexpected statement -want/+got
            λ({*ast.File}.Body[0].BaseNode.Loc.End):
            	-: "1:13"
            	+: "2:15"
            {*ast.File}.Body[0].BaseNode.Loc.Source:
            	-: "not (a and b"
            	+: "not (a and b)\n(a or b) and c"
            {*ast.File}.Body[0].Expression:
            	-: &ast.UnaryExpression{BaseNode: ast.BaseNode{Loc: s"1:1-1:13"}, Operator: s"not", Argument: &ast.LogicalExpression{BaseNode: ast.BaseNode{Loc: s"1:6-1:13"}, Operator: s"and", Left: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"1:6-1:7"}, Name: "a"}, Right: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"1:12-1:13"}, Name: "b"}}}
            	+: &ast.LogicalExpression{BaseNode: ast.BaseNode{Loc: s"1:1-2:15"}, Operator: s"and", Left: &ast.UnaryExpression{BaseNode: ast.BaseNode{Loc: s"1:1-2:9"}, Operator: s"not", Argument: &ast.CallExpression{BaseNode: ast.BaseNode{Loc: s"1:6-2:9"}, Callee: &ast.LogicalExpression{BaseNode: ast.BaseNode{Loc: s"1:6-1:13"}, Operator: s"and", Left: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"1:6-1:7"}, Name: "a"}, Right: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"1:12-1:13"}, Name: "b"}}, Arguments: []ast.Expression{&ast.ObjectExpression{BaseNode: ast.BaseNode{Loc: s"2:2-2:8"}, Properties: []*ast.Property{&{BaseNode: ast.BaseNode{Loc: s"2:2-2:3"}, Key: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:2-2:3"}, Name: "a"}}, &{BaseNode: ast.BaseNode{Loc: s"2:7-2:8"}, Key: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:7-2:8"}, Name: "b"}}}}}}}, Right: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:14-2:15"}, Name: "c"}}
            {*ast.File}.Body[1->?]:
            	-: &ast.ExpressionStatement{BaseNode: ast.BaseNode{Loc: s"2:2-2:15"}, Expression: &ast.LogicalExpression{BaseNode: ast.BaseNode{Loc: s"2:2-2:15"}, Operator: s"and", Left: &ast.LogicalExpression{BaseNode: ast.BaseNode{Loc: s"2:2-2:8"}, Operator: s"or", Left: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:2-2:3"}, Name: "a"}, Right: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:7-2:8"}, Name: "b"}}, Right: &ast.Identifier{BaseNode: ast.BaseNode{Loc: s"2:14-2:15"}, Name: "c"}}}
            	+: <non-existent>
```

wolffcm:
Fixes #1102.
BREAKING CHANGE: adds additional errors to parser.

### Done checklist
- [x] Test cases written

### DIFF
--- a/internal/parser/grammar.md
+++ b/internal/parser/grammar.md
@@ -61,7 +61,7 @@ The parser directly implements the following grammar.
     MemberExpression               = DotExpression  | MemberBracketExpression
     DotExpression                  = "." identifer
     MemberBracketExpression        = "[" string "]" .
-    CallExpression                 = "(" ParameterList ")" .
+    CallExpression                 = "(" PropertyList ")" .
     IndexExpression                = "[" Expression "]" .
     PrimaryExpression              = identifer
                                    | int_lit
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -1053,6 +1053,7 @@ func (p *parser) parseParenIdentExpression(lparen token.Pos, key *ast.Identifier
 
 func (p *parser) parsePropertyList() []*ast.Property {
 	var params []*ast.Property
+	perrs := make([]ast.Error, 0)
 	for p.more() {
 		var param *ast.Property
 		switch _, tok, _ := p.peek(); tok {
@@ -1061,22 +1062,28 @@ func (p *parser) parsePropertyList() []*ast.Property {
 		case token.STRING:
 			param = p.parseStringProperty()
 		default:
-			// TODO(jsternberg): BadExpression.
-			p.consume()
-			continue
+			param = p.parseInvalidProperty()
 		}
 		params = append(params, param)
-		if _, tok, _ := p.peek(); tok == token.COMMA {
-			p.consume()
+
+		if p.more() {
+			if _, tok, lit := p.peek(); tok != token.COMMA {
+				perrs = append(perrs, ast.Error{
+					Msg: fmt.Sprintf("expected comma in property list, got %s (%q)", tok, lit),
+				})
+			} else {
+				p.consume()
+			}
 		}
 	}
+	p.errs = append(p.errs, perrs...)
 	return params
 }
 
 func (p *parser) parseStringProperty() *ast.Property {
 	key := p.parseStringLiteral()
 	p.expect(token.COLON)
-	val := p.parseExpression()
+	val := p.parsePropertyValue()
 	return &ast.Property{
 		Key:   key,
 		Value: val,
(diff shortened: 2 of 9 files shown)


internal/parser/grammar.md
```diff
@@ -59,7 +59,7 @@ The parser directly implements the following grammar.
     MemberExpression               = DotExpression  | MemberBracketExpression
     DotExpression                  = "." identifer
     MemberBracketExpression        = "[" string "]" .
-    CallExpression                 = "(" ParameterList ")" .
```
jsternberg: Even if they share the same implementation, parameter list and property list are different and so this should stay the same.

internal/parser/parser.go
```diff
@@ -972,15 +972,21 @@ func (p *parser) parsePropertyList() []*ast.Property {
...
-		switch _, tok, _ := p.peek(); tok {
+		switch pos, tok, lit := p.peek(); tok {
 		case token.IDENT:
 			param = p.parseIdentProperty()
 		case token.STRING:
 			param = p.parseStringProperty()
 		default:
 			// TODO(jsternberg): BadExpression.
+			p.errs = append(p.errs, ast.Error{
+				Msg: fmt.Sprintf("in property list, expected %s or %s, got %s (%q) at %s",
```
jsternberg: Can you modify the message for this?

```
fmt.Sprintf("unexpected token for property key: %s (%q)", tok, lit)
```

It might also be good to encapsulate this in its own function where you consume the bad token and then consume any colon and expression that follows so it will catch things like `{5: a}`. This is with a special exception that if it sees a comma that it construct a `missing property` message (or something like that) and if it sees a colon it will construct a `missing property key` message. That should cover our bases.

internal/parser/grammar.md
wolffcm: Hmm, this doesn't seem right to me.  The parameter list is part of a function expression like `f = (x, y) => x * y`, but the property list is for calls like `f(x:5, y:10)`.  There are both `parsePropertyList` and `parseParameterList` methods in the parser, since they have different syntaxes.

affo:
Closes https://github.com/influxdata/flux/issues/1235

internal/parser/parser.go
```diff
@@ -1061,50 +1062,122 @@ func (p *parser) parsePropertyList() []*ast.Property {
...
 	return params
 }
 
 func (p *parser) parseStringProperty() *ast.Property {
 	key := p.parseStringLiteral()
 	p.expect(token.COLON)
-	val := p.parseExpression()
+	val := p.parsePropertyValue()
+	var endPos ast.Position
+	if val != nil {
```
jsternberg: I understand why this is here, but can you use the current pattern of calling `locEnd` on only the value so this is consistent with the other parts of the code?

internal/parser/parser.go
```diff
@@ -1061,50 +1062,122 @@ func (p *parser) parsePropertyList() []*ast.Property {
...
 		p.consume()
-		property.Value = p.parseExpression()
-		property.Loc = p.sourceLocation(
-			locStart(key),
-			locEnd(property.Value),
-		)
+		val := p.parsePropertyValue()
+		if val != nil {
+			property.Value = val
+			endPos = locEnd(val)
```
jsternberg: Here's another place where it's ok to just call `locEnd` on the null value.

I'm thinking I want to refactor how locations are created entirely so I would like the problem to be consistent so it is easy to find.

internal/parser/parser.go
wolffcm: Okay, I'll do that.  The location info will be the zero value for properties with no value set for now.  I filed an issue for  your idea of `loc(nodes ...ast.Node)`:
https://github.com/influxdata/flux/issues/1237

PR #1176, opened 2019-04-16
fix(execute): rename json tag so go vet passes

wolffcm:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -21,7 +21,7 @@ type aggregateTransformation struct {
 
 type AggregateConfig struct {
 	plan.DefaultCost
-	Columns []string `json:"column"`
+	Columns []string `json:"columns"`
 }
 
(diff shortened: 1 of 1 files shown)


execute/aggregate.go
```diff
@@ -21,7 +21,7 @@ type aggregateTransformation struct {
 
 type AggregateConfig struct {
 	plan.DefaultCost
-	Columns []string `json:"column"`
```
jsternberg: We can't do this because it makes it so `aggregateWindow` won't work correctly.

wolffcm: I'm missing something.  This changes the serialization format, but not much else, right?  All the go tests passed, are we missing coverage?

jsternberg: Ah, I think I misunderstood the aggregate config change. It seems it has a special case for having both `column` and `columns`.

PR #1178, opened 2019-04-16
docs(SPEC): Add description of conditional expressions to SPEC

### LINKED ISSUE #1170: Design conditional syntax
See https://github.com/influxdata/flux/issues/923

wolffcm:
Fixes #1170.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -729,29 +729,45 @@ It is not possible to access an object's property using an arbitrary expression.
     DotExpression           = "." identifer
     MemberBracketExpression = "[" string_lit "]" .
 
+#### Conditional Expressions
+
+Conditional expressions evaluate a boolean-valued condition and if the result it _true_,
+the expression following the `then` keyword is evaluated and returned.
+Otherwise the condition evaluates to _null_ or _false_,
+and the expression following the `else` keyword is evaluated and returned.
+
+    ConditionalExpression   = "if" Expression "then" Expression "else" Expression .
+
+Example:
+
+    color = if code == 0 then "green" else if code == 1 then "yellow" else "red"
+
 #### Operators
 
 Operators combine operands into expressions.
 The precedence of the operators is given in the table below. Operators with a lower number have higher precedence.
 
-|Precedence| Operator |        Description        |
-|----------|----------|---------------------------|
-|     1    |  `a()`   |       Function call       |
-|          |  `a[]`   |  Member or index access   |
-|          |   `.`    |       Member access       |
-|     2    | `*` `/`  |Multiplication and division|
-|     3    | `+` `-`  | Addition and subtraction  |
-|     4    |`==` `!=` |   Comparison operators    |
-|          | `<` `<=` |                           |
-|          | `>` `>=` |                           |
-|          |`=~` `!~` |                           |
-|     5    |  `not`   | Unary logical expression  |
-|     6    |  `and`   |        Logical AND        |
-|     7    |  `or`    |        Logical OR         |
+|Precedence| Operator       |        Description        |
+|----------|----------------|---------------------------|
+|     1    |  `a()`         |       Function call       |
+|          |  `a[]`         |  Member or index access   |
+|          |   `.`          |       Member access       |
+|     2    | `*` `/`        |Multiplication and division|
+|     3    | `+` `-`        | Addition and subtraction  |
+|     4    |`==` `!=`       |   Comparison operators    |
+|          | `<` `<=`       |                           |
+|          | `>` `>=`       |                           |
+|          |`=~` `!~`       |                           |
+|     5    |  `not`         | Unary logical expression  |
+|     6    |  `and`         |        Logical AND        |
+|     7    |  `or`          |        Logical OR         |
+|     8    | `if/then/else` |        Conditional        |
 
 The operator precedence is encoded directly into the grammar as the following.
 
-    Expression               = LogicalExpression .
+    Expression               = ConditionalExpression .
+    ConditionalExpression    = LogicalExpression
+                             | "if" Expression "then" Expression "else" Expression .
     LogicalExpression        = UnaryLogicalExpression
                              | LogicalExpression LogicalOperator UnaryLogicalExpression .
     LogicalOperator          = "and" | "or" .
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -731,9 +731,9 @@ It is not possible to access an object's property using an arbitrary expression.
 
 #### Conditional Expressions
 
-Conditional expressions evaluate a boolean-valued condition and if the result it _true_,
+Conditional expressions evaluate a boolean-valued condition and if the result is _true_,
 the expression following the `then` keyword is evaluated and returned.
-Otherwise the condition evaluates to _null_ or _false_,
+Otherwise the condition evaluates to _false_,
 and the expression following the `else` keyword is evaluated and returned.
 
     ConditionalExpression   = "if" Expression "then" Expression "else" Expression .
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -735,6 +735,8 @@ Conditional expressions evaluate a boolean-valued condition and if the result is
 the expression following the `then` keyword is evaluated and returned.
 Otherwise the condition evaluates to _false_,
 and the expression following the `else` keyword is evaluated and returned.
+In either case, the branch not taken is not evaluated;
+only side effects associated with the branch that is taken will occur.
 
     ConditionalExpression   = "if" Expression "then" Expression "else" Expression .
 


docs/SPEC.md
```diff
@@ -729,29 +729,45 @@ It is not possible to access an object's property using an arbitrary expression.
...
+|     2    | `*` `/`        |Multiplication and division|
+|     3    | `+` `-`        | Addition and subtraction  |
+|     4    |`==` `!=`       |   Comparison operators    |
+|          | `<` `<=`       |                           |
+|          | `>` `>=`       |                           |
+|          |`=~` `!~`       |                           |
+|     5    |  `not`         | Unary logical expression  |
+|     6    |  `and`         |        Logical AND        |
+|     7    |  `or`          |        Logical OR         |
+|     8    | `if/then/else` |        Conditional        |
```
jsternberg: Can you explain the reason for this decision with an example? It doesn't have to be in the document, I'm just curious. I think this is correct, but I'd like to see what the prior art is before I just go with my feeling on the matter.

docs/SPEC.md
```diff
@@ -729,29 +729,45 @@ It is not possible to access an object's property using an arbitrary expression.
     DotExpression           = "." identifer
     MemberBracketExpression = "[" string_lit "]" .
 
+#### Conditional Expressions
+
+Conditional expressions evaluate a boolean-valued condition and if the result it _true_,
+the expression following the `then` keyword is evaluated and returned.
+Otherwise the condition evaluates to _null_ or _false_,
```
jlapacik: Not sure why the condition should ever evaluate to _null_. What's the use case?

wolffcm: In general I think null values can appear almost anywhere that a value appears. In the case of `map`, for example:
```
   ... |> map(fn: (r) => {code: if r.code == 0 then "green" else "red"})
```

jsternberg: I think as long as we assert that if statements have to type check to a boolean this is fine, but every value is nullable in flux at the moment so the boolean could be null.

If we want to take the type system further, we can differentiate between nullable columns and non-nullable ones. Anything coming from the storage engine is non-nullable, but some methods like join or pivot will add a nullable property onto everything and it might be too "pure" for many users.

pauldix: Curious to hear about this one too. Otherwise this LGTM

docs/SPEC.md
wolffcm: I was basing the keywords on SML syntax.  I chose the precedence intuitively, but you're right to ask about prior art.  The definition of SML is here:
http://sml-family.org/sml97-defn.pdf
On page 75 there is a discussion of grammar and precedence.  In their notation, forms are listed in decreasing precedence.  They give infix operator expressions higher precedence than others, including `if/then/else`.  Only `while`, `case` and `fn` have lower precedence. (This last one is interesting because it seems different from the precedence of function expressions in Flux)

Another example would be C, in which the `?:` conditional expression has lower precedence than everything except the compound assignments (`+=`, etc) and the `,` operator.

docs/SPEC.md
wolffcm: Ok, I'll remove the `or null` bit.

So I guess `exists code r` would evaluate to true if either the column `code` exists and is null, or if the column `code` does not exist?

Will we be doing any type checking at runtime once the schema of tables is known, or is type checking only a compile-time thing?

This whole discussion also seems relevant to the stream index functions, since we'll want to figure out how to avoid nulls leaking out there.

jsternberg:
@nathanielc I think we need to have a larger conversation about nulls then. With influxql, I ran into a lot of trouble because we kept repeating, "Nulls don't exist," but they very obviously did exist and we ran into issues when the null values ended up showing up within the query engine.

I don't think we can just say "flux does not have null values." Just because the user cannot specify that the null values exist through a keyword doesn't mean they aren't present. They were present in influxql and the storage engine didn't support null values.

If we want to build that into the type system so that `if` will not operate on null values, that's good too. Many functional systems do that. But, they don't ignore the presence of null values. Haskell uses monads to represent a null value and others use a special option syntax to represent the difference between a boolean and a nullable boolean along with a way to unbox those values.

This seems like it would be a bit overengineered for our purpose and saying "null means the value is unknown so it always returns false for everything" seems like a reasonable answer to me. I just think we have to address null in the spec because, whether or not we put it in there, null values exist in the system. Otherwise, we would be able to remove the `IsNull` method from everywhere.

docs/SPEC.md
nathanielc: > So I guess exists code r would evaluate to true if either the column code exists and is null, or if the column code does not exist?

Yes, the semantics of null are the value is missing, whether all rows or just some of the rows are missing the value is not an important distinction.

> Will we be doing any type checking at runtime once the schema of tables is known, or is type checking only a compile-time thing?

Type checking will happen at compile time and runtime, this is what I mean when I say we want Flux to be gradually typed. If the compiler can do the type checking then it will insert the runtime type checks instead.

> This whole discussion also seems relevant to the stream index functions, since we'll want to figure out how to avoid nulls leaking out there.

I agree, the `record` function needs to not create properties on the returned object if it is null.

nathanielc:
@jsternberg Sorry for the consfution, to be clear I am not saying that `nulls` do not exist, I am saying that the value `null` does not exist , but the semantics of `nulls` do exist.

Let me state it a different way.

In the current implementation we have the concept of `null` values, that is not changing. The implementation of columns as arrow buffers will always use the concept of `null` to represent data that is missing. 

What I am saying is that the semantics of `null` mean that a value does not exist. There are at least two ways to represent those semantics in the Flux language.

1. Values can be nullable i.e. meaning that you do not actually have a value. 
2. Explicitly represent the missing data in the type system. 

I am proposing the second method. The first method allows for null checks to happen at runtime since nulls are values. I am proposing that with the type system we are building we can safely model nulls in the type system while still being able to defer null checks to runtime when needed.

Specifically objects are the only place that data can be missing. Meaning that code might expect a property to exist on data but in reality it does not in fact exist. So if we protect that specific access pattern we can keep the semantics of `nulls` to the type system and not have to allow for all Flux values to be nullable. 

This design combined with gradual typing means that Flux can defer the type checks to runtime when needed, so null check just become runtime type checks as needed.

docs/SPEC.md
wolffcm: Ok, I think I get it.

We might have an expression like this, say as the lambda in a `map`:
```
if exists _value r then r._value else 0
```
At compile time, we don't know what the fields of `r` are, so we instrument the output of the compiler to be something like:
```
if 
    exists _value r 
then 
    assert has_field _value r
    typeassert r._value int
    r._value
else
    0
```
So at compile time, the type checking is a pass over the IR.  At runtime, type checks are evaluated as part of the program.  Is this close to how you're thinking it will work?

nathanielc: What is the difference between `assert has_field` and `exist _value r` checks?

I was thinking something slightly different:

Case `r`'s type is unknown:

```
if
    exists _value r 
then 
    typeassert r._value int
    r._value
else
    0
```


Case `r`'s type is known to have the `_value` property with type `int`:

```
r._value
```

wolffcm: > What is the difference between assert has_field and exist _value r checks?

Yes, you're right, that first assert is unnecessary.  I think we're on the same page.

jsternberg:
I'm not really sure I understand the difference. If I were to do `if r.my_boolean ...` then there is the possibility that `my_boolean` does not exist and is "null". This can either be because there is no column with that name in the object or that the data itself is missing within that specific row.

I'm going to use the naming of Haskell since that's the functional language I'm most familiar with. If we capture this information within the type system, we're going to have a lot of `Maybe` types and every single access of an object property is going to have to be an assertion that the value is not null and then returning an error if it is. I think that's a pretty big change to make. If we don't do that, then there's no way to tell the difference between a nullable boolean and a non-nullable boolean so we have to assume that any value could be null.

nathanielc:
>  If I were to do if r.my_boolean ... then there is the possibility that my_boolean does not exist and is "null". 

I am saying that it is possible that the above is not a possibility.

>  If we don't do that, then there's no way to tell the difference between a nullable boolean and a non-nullable boolean so we have to assume that any value could be null.

Specifically I am proposing that we do not need the concept of a nullable boolean value. Or any nullable values for any types.

I am going to try and explain how this works using the simple example of `if r.my_boolean ...`.

This is an expression: `r.my_boolean`  that says get the value of the property `my_boolean` from the object `r`. 

Let's first examine the simple case where the type of `r` is known statically (i.e. at compile time).

If the property exists then the value is returned. If the property `my_boolean` does not exist then it is a type error. In no case does the expression evaluate to `null`. 

Simple Flux REPL output demonstrating that behavior:

```
$ flux repl
> a = {x: 1, y: 2}
<nil>
> a.z
Error: missing object properties (z)
> a.x
1
```

OK now lets examine the case where `r`'s type is not known statically (i.e. it is known only at runtime).

In this case for consistency we would like the behavior of accessing the property `my_boolean` to behave similar to the static case. So there are still two possible results from evaluating the expression `r.my_boolean`, either the value is returned or a type error occurs. A type error will kill the query. Killing the query is not good, so we provide a type safe way to access the `my_boolean` property if the users knows or otherwise expects that some `my_boolean` values may be missing.

The type safe way to access `my_boolean` looks something like this (again ignore the specific syntax, its the semantics that matter):

```
if exists my_boolean r 
then 
   r.my_boolean
else 
   // user gets to define the behavior
   // in this case we return false
   false
```

We now have a way to access properties of dynamic objects in a type safe way. The above expression will never result in a type error, nor will it ever evaluate to the null value. 

This workflow allows the users to write safe code to handle missing data without having to worry about if every value in a Flux script is possible `null` because `null` isn't actually a value.

What if the user wants to preserve the information that in fact `r.my_boolean` did not exist?

In that case this is where the concept of variants and a Maybe type are introduced. The user could write this code instead(borrowing the Just and Nothing constructors from Haskell):

```
if exists my_boolean r 
then 
   Just r.my_boolean
else 
   // user gets to define the behavior
   // in this case we return the Nothing value of the Maybe variant
   Nothing
```

Alternatively like you mentioned we could model all objects as being key values pairs where every value must be a Maybe value and a lookup of a non existent key returns a Nothing value. But that requires that users always unwrap the Maybe value. This approach of having a type safe `exists` operator seems like a good middle ground, where users can write safe code or they can ignore the problem entirely.

This is a bit nuanced specially in the face of how it is implemented using `null` values in the arrow buffers. I hope I am being clear and I am happy to keep discussing till we understand one another.

docs/SPEC.md
wolffcm: @nathanielc 

So I guess for filters a common idiom would be something like:
```
filter(fn: (r) => if exists id r and r.id == 100 then true else false)
```
I wonder if we should provide some syntactic sugar for this common case?

I think I like this approach for InfluxDB sources, but what about SQL?  I have a couple concerns.

Suppose a user writes a Flux script that accesses SQL data, and maybe joins it to data in their InfluxDB source.  Their test data has no nulls, and the script works fine.  In production (potentially a long while later) nulls appear, and the script now breaks.  This could be painful for users.

My second concern regards how we optimize Flux queries for SQL sources.  I'd imagine we want to allow the user to write Flux in a natural way and let the planner push things down into the SQL source.  How should we translate something like `filter(fn: (r) => r.id == 100)` into SQL?  If `r.id` being null is a type error, then I guess we'd want SQL to throw an exception, and I'm not sure how to do that.  With syntactic sugar for nulls+predicates, maybe it gets easier.

jsternberg:
I think I have few questions.

1. Are row functions (the ones that would need to be type checked at runtime) compiled per row or per table? I think per table would be better, especially for performance.
2. If compiled per row and there are null values, are they removed from the input so that there is no way to have a null value?

The second question isn't relevant if the first is per table. I thought that the function would be compiled per table using the schema information for the table and we would then compile it using the types we knew based on the schema. This means that every property would be capable of being a nullable value since all columns can have nulls in them unless we add a specific syntax to tell the schema "this column will have no null values". I actually like that idea so I'll explain what I mean at the end.

At the current moment, I don't see this solution as consistent. For example, this looks like Ruby:

```
if exists my_boolean r 
then 
   r.my_boolean
else 
   // user gets to define the behavior
   // in this case we return false
   false
```

There's nothing wrong with that. Ruby is one of my favorite languages. But, specifically, the `exists` syntax can't exist in a statically typed language to my knowledge. Whether or not the point exists, it has to go through type checking and so there needs to be a type that it resolves to. And type checking usually has to resolve all of the branches. If there is no `my_boolean` property, I don't think this would resolve. I also anticipate that this will have to be done a lot when there are joins involved.

But I also have to wonder what happens if I were to write this query:

```
filter(fn: (r) => r._measurement == "cpu" and r.host == "server01")
```

Imagine that I have been careful and every cpu measurement has the host tag. But, this query would technically run on every series that exists and we wouldn't short circuit type checking. If we validate the property at compile time, then this compilation should fail whenever presented with a record that doesn't have `host`. On the other hand, it would be difficult to write every single query with a check to see if every property we ever access exists.

When I say having the columns denote if they have a nullable value, I'm talking about monads (or specifically option in rust, maybe in haskell). We could have the column type include the properties such as `boolean;nullable` somewhere in there and then the type inference would see it as a different type and act accordingly. So accessing the nullable type would have to be done in a different way.

I do think that this is too complicated though. I don't think it's reasonably possible for us to implement that and the other features while trying to get a GA version ready within the timeframe we've given ourselves. I think it would lead to scope creep.

The reason I'm so cautious is I remember that this was something that permanently crippled influxql. I know that dealing with null may create a lot of edge cases, but I also think that declaring that nulls don't exist in the language just means we aren't dealing with nulls at all because they very plainly do exist in the language. Data will sometimes be missing and certain transformations even generate null values. I think if we want to say "null values don't exist" then we need to look at how nulls are getting into the system.

docs/SPEC.md
```diff
@@ -729,29 +729,45 @@ It is not possible to access an object's property using an arbitrary expression.
     DotExpression           = "." identifer
     MemberBracketExpression = "[" string_lit "]" .
 
+#### Conditional Expressions
```
nathanielc: I think we should explicitly state that only one branch is evaluated. This only matters for side effects but we should be explicit about it.

PR #1184, opened 2019-04-17
feat(semantic): add handling for conditional expressions to type inference

### LINKED ISSUE #1173: Implement conditionals in the semantic graph
See https://github.com/influxdata/flux/issues/923

wolffcm:
Fixes #1173.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/semantic/inference_test.go
+++ b/semantic/inference_test.go
@@ -1606,6 +1606,71 @@ foo.b
 				},
 			},
 		},
+		{
+			name:   "conditional expression",
+			script: `if true then 3 else 30`,
+			solution: &solutionVisitor{
+				f: func(node semantic.Node) semantic.PolyType {
+					switch node.(type) {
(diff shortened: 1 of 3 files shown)


semantic/inference_test.go
```diff
@@ -1606,6 +1606,71 @@ foo.b
...
+							return tv
+						}
+					}
+					return nil
+				},
+			},
+		},
+		{
+			name:    "conditional branches must agree",
+			script:  `if true then 0 else 0.0`,
```
jsternberg: Can you change this test so one of them is a string instead? I think we want to fix this specific type constraint by making it so an integer literal could be an integer or a float which would make this test case wrong.

wolffcm: Sure, will do.

PR #1185, opened 2019-04-17
fix(stdlib): tagKeys should include a call to distinct

jsternberg:
- [x] docs/SPEC.md updated
- [ ] Test cases written

The full chain is to look only at distinct keys. This seems to have not
been included. This changes `v1.tagKeys()` to include `distinct()` at the
end.

### DIFF
--- a/stdlib/influxdata/influxdb/v1/v1.flux
+++ b/stdlib/influxdata/influxdb/v1/v1.flux
@@ -34,6 +34,7 @@ tagKeys = (bucket, predicate=(r) => true, start=-30d) =>
         |> filter(fn: predicate)
         |> keys()
         |> keep(columns: ["_value"])
+        |> distinct()
 
 // MeasurementTagKeys returns the list of tag keys for a specific measurement.
 measurementTagKeys = (bucket, measurement) =>
--- a/stdlib/influxdata/influxdb/v1/flux_gen.go
+++ b/stdlib/influxdata/influxdb/v1/flux_gen.go
@@ -22,10 +22,10 @@ var pkgAST = &ast.Package{
 			Loc: &ast.SourceLocation{
 				End: ast.Position{
 					Column: 51,
-					Line:   44,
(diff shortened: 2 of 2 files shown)


stdlib/influxdata/influxdb/v1/v1.flux
```diff
@@ -34,6 +34,7 @@ tagKeys = (bucket, predicate=(r) => true, start=-30d) =>
         |> filter(fn: predicate)
         |> keys()
         |> keep(columns: ["_value"])
+        |> unique()
```
nathanielc: This should probably be `distinct` and should be after the call to `keys`. If I remember right that is what the planner was expecting.

jsternberg: What's the difference between unique and distinct?

nathanielc: Unique is a selector, so it returns the first row for each new value. Distinct is more like an aggregate where all other data is lost except for the set of distinct values

jsternberg: Ah. Do we need to have `keep` in there then? It seems to me like `distinct()` is the same as `keep(columns: ["_value"]) |> unique()`?

nathanielc: I still think that this needs to be after the `keys` call and before the `keep` call, otherwise the planner won't know how to optimize. I could be remembering incorrectly 

I think we need the keep to remove some extraneous columns but I don't remember exactly.

jsternberg: It doesn't work in the opposite order because keep also ends up grouping everything into the same table do distinct works.

This will be used by the new rpc call so we don't need to really worry about the old one.

jsternberg: Another one that works group with no arguments and then distinct.

PR #1202, opened 2019-04-22
fix(stdlib/universe): fix reducer so that it resets the reduce value to the neutral element value for each new group key, and report an error when two reducers write to the same destination group key.

### LINKED ISSUE #1200: reduce does not reset reducer on group key boundaries.

aanthony1243:
fixes #1200 

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


mark-rushakoff:
Confirmed this fixes the repro case I posted: https://github.com/influxdata/flux/issues/1200#issuecomment-485456136

PR #1213, opened 2019-04-25
fix(execute): make executor respect memory limit from caller

wolffcm:
Fixes influxdata/idpe#3115.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -89,8 +89,6 @@ func (e *executor) createExecutionState(ctx context.Context, p *plan.Spec, a *me
 	if err := validatePlan(p); err != nil {
 		return nil, errors.Wrap(err, "invalid plan")
 	}
-	// Set allocation limit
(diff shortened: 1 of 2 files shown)


execute/executor.go
```diff
@@ -266,10 +264,15 @@ func (es *executionState) do(ctx context.Context) {
 					default:
 						err = fmt.Errorf("%v", e)
 					}
-					es.abort(fmt.Errorf("panic: %v\n%s", err, debug.Stack()))
-					if entry := es.logger.Check(zapcore.InfoLevel, "Execute source panic"); entry != nil {
-						entry.Stack = string(debug.Stack())
-						entry.Write(zap.Error(err))
+
+					if _, ok := err.(memory.LimitExceededError); ok {
+						es.abort(err)
```
jsternberg: Instead of the hanging else, can you do a `return` and then remove the else part?

PR #1216, opened 2019-04-25
fix(execute): properly use RefCount to reference count tables

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The consecutive transport and the result both use goroutines to process
the tables and so need to have their own references to the tables. But,
previously, the dataset was the only one that increased the reference
count even though there was no guarantee it was using the consecutive
transport and no guarantee that the reference count would be
decremented.

This modifies these two areas so that the reference count is increased
on the call to `Process()` to signal that it will keep a reference to
the table and then it decrements the reference count after it discards
its usage of the table.

### DIFF
--- a/execute/result.go
+++ b/execute/result.go
@@ -42,6 +42,7 @@ func (s *result) RetractTable(DatasetID, flux.GroupKey) error {
 }
 
 func (s *result) Process(id DatasetID, tbl flux.Table) error {
+	tbl.RefCount(1)
 	select {
 	case s.tables <- resultMessage{
 		table: tbl,
@@ -68,8 +69,10 @@ func (s *result) Do(f func(flux.Table) error) error {
 				return msg.err
 			}
 			if err := f(msg.table); err != nil {
+				msg.table.RefCount(-1)
 				return err
 			}
+			msg.table.RefCount(-1)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


execute/result.go
```diff
@@ -68,8 +69,10 @@ func (s *result) Do(f func(flux.Table) error) error {
 				return msg.err
 			}
 			if err := f(msg.table); err != nil {
+				msg.table.RefCount(-1)
```
nathanielc: Why does only this error branch decrement the counter?

execute/transport.go
```diff
@@ -68,6 +68,7 @@ func (t *consecutiveTransport) RetractTable(id DatasetID, key flux.GroupKey) err
 }
 
 func (t *consecutiveTransport) Process(id DatasetID, tbl flux.Table) error {
+	tbl.RefCount(1)
```
nathanielc: Where is the companion decrement to the increment?

execute/result.go
jsternberg: There's another line under this one for decrementing it in the normal situation. Let me refactor this so it uses a defer and anonymous function.

execute/transport.go
jsternberg: It's in process message. It was already there previously, but the increment was in unrelated code so I thought it would be easier to track if the increment and decrement were in the same structure.

That's why I removed the increment in the dataset because it relied on an implementation detail.

PR #1218, opened 2019-04-26
test(execute): add test for memory limits

wolffcm:
This adds a unit test for the fix that was added here:
https://github.com/influxdata/flux/pull/1213

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	uuid "github.com/satori/go.uuid"
 )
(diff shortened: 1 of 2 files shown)


execute/executetest/source.go
```diff
@@ -63,3 +64,46 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
...
+}
+
+func CreateAllocatingFromSource(spec plan.ProcedureSpec, id execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	s := spec.(*AllocatingFromProcedureSpec)
+	s.alloc = a.Allocator()
+	return s, nil
+}
+
+func (s *AllocatingFromProcedureSpec) Run(ctx context.Context) {
+	err := s.alloc.Allocate(s.ByteCount)
```
jsternberg: Can you do this syntax?

```
if err := s.alloc.Allocate(s.ByteCount); err != nil {
    panic(err)
}
```

PR #1226, opened 2019-04-29
fix(control): fix request results labels to count runtime errors

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -146,7 +146,6 @@ func (c *Controller) Query(ctx context.Context, compiler flux.Compiler) (flux.Qu
 		c.countQueryRequest(q, labelQueueError)
 		return nil, q.Err()
 	}
-	c.countQueryRequest(q, labelSuccess)
 	return q, nil
 }
 
@@ -430,6 +429,13 @@ func (q *Query) Done() {
 
 		q.transitionTo(Finished)
 		q.c.finish(q)
+
+		// count query request
+		if q.err != nil || len(q.runtimeErrs) > 0 {
+			q.c.countQueryRequest(q, labelRuntimeError)
+		} else {
+			q.c.countQueryRequest(q, labelSuccess)
+		}
 	})
 }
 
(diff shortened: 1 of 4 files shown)


control/controller.go
```diff
@@ -422,6 +421,13 @@ func (q *Query) Done() {
 
 		q.transitionTo(Finished)
 		q.c.finish(q)
+
+		// count query request
+		if q.err != nil {
+			q.c.countQueryRequest(q, labelRuntimeError)
```
wolffcm: I don't think this will catch all runtime errors.  It will only catch those errors that occur during execution initialization.  Errors that occur afterwards are just passed as results to the client, here:
https://github.com/influxdata/flux/blob/master/execute/result.go#L68
These errors are only reported to the client when `TableIterator` finally gets them.

There can be multiple errors that occur during runtime, maybe we need to change the `Query` interface to reflect this.  For now you could rebase on my branch and pull the errors from `flux.Statistics.RuntimeErrors` for the metric.

Maybe we should talk about how to handle error reporting moving forward, there is a deceptive amount of surface area for our client API.

affo: Shouldn't `Query.Done()` be called once the results have been consumed by the client?

At that point the `err` field should be properly populated.

wolffcm: That seems intuitive, but it's not currently the case.

We just pass the `Result` from the `exec.Query` to the `control.Query`'s output channel here:
https://github.com/influxdata/flux/blob/master/control/controller.go#L617
When the client receives it, they call `result.Tables().Do(..)` to get at the tables in the result.  `Do` will return an error code if there is an error during execution, like a non-existent field reference or something like that.

So there's really three different error paths:
- Compile time error: returned from call to `controller.Query(..)`
- Execution initialization time error: returned from `query.Err()`
- Run time error: returned from call to `result.Tables().Do(...)`

I don't think we should be reporting the same error in multiple places, so it seems unclear to me if we want to return errors produced by `Tables().Do(...)` in `query.Err()` as well.

For reference, my PR for logging runtime errors is here:
https://github.com/influxdata/flux/pull/1227
https://github.com/influxdata/idpe/pull/3336

affo: Thank you Chris, I took a look at your PRs.
I was forgetting the `r.Tables().Do()` path. 👍

PR #1227, opened 2019-04-30
refactor: add runtime errors to flux.Statistics

wolffcm:
Also make the controller append any runtime errors to stats.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


wolffcm:
I'm re-requesting review based on our slack discussion, which I'll summarize here:
- It's possible for more than one runtime error to occur
- We will combine both compile and run time errors to the same tag key in the query log
- In the future we should think about how to simplify or client interface because if its confusing for us, it will be confusing to users.

PR #1228, opened 2019-05-01
fix(execute): return the error from the context in the executor

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

When a query has had its context canceled, the query should abort with
the error contained within the context instead of swallowing the error
and returning `context done` as the error.

### DIFF
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -297,7 +297,7 @@ func (es *executionState) do(ctx context.Context) {
 			select {
 			case <-t.Finished():
 			case <-ctx.Done():
-				es.abort(errors.New("context done"))
+				es.abort(ctx.Err())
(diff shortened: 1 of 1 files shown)


execute/executor.go
```diff
@@ -297,7 +297,9 @@ func (es *executionState) do(ctx context.Context) {
 			select {
 			case <-t.Finished():
 			case <-ctx.Done():
-				es.abort(errors.New("context done"))
+				if err := ctx.Err(); err != nil {
```
jlapacik: Is an error guaranteed not to be nil here?

jsternberg: The documentation says that if done is closed then this will always return non-nil so I'll remove the check.

PR #1229, opened 2019-05-01
fix(stdlib): properly check if the right table is null when joining

jsternberg:
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -556,7 +556,7 @@ func (c *MergeJoinCache) Table(key flux.GroupKey) (flux.Table, error) {
 		}
 
 		right := c.buffers[c.rightID].table(preJoinGroupKeys.right)
-		if left == nil {
+		if right == nil {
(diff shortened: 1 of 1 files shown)


jlapacik:
Do you think it would be easy to write a test for this, so we don't regress?

jsternberg:
I tried to make a unit test and realized that this solution is entirely incorrect. The reason why it panics is fixed by the above, but I didn't ask the question "why is right null?" And that's because the tables are stored in the buffer by their group key which is actually a reference to the pointer, not the actual group key.

jsternberg:
I may have rushed to judgment on this. There are further issues. I resolved that specific question of the panic, but I didn't answer, "Why is right null?" It appears as if join is using `flux.GroupKey` as a key in maps and this uses the memory address and not the content of the group key which is incorrect.

I'll have another PR for reproducing this in a test and then also changing join to use group lookup instead of a map.

PR #1239, opened 2019-05-03
perf(stdlib): avoid using ValueForRow when appending values in window

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

It was found on a flamegraph that using `ValueForRow` and then appending
the value immediately generated a lot of allocations because it would
allocate memory for an interface and then discard it immediately.

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -267,6 +267,47 @@ func AppendMappedRecordExplicit(i int, cr flux.ColReader, builder TableBuilder, 
 	return nil
 }
 
+// AppendValueFromRow will retrieve a value from an arrow column reader at the
+// given index and append it to a TableBuilder in the specified row.
+func AppendValueFromRow(builder TableBuilder, dst int, cr flux.ColReader, i, j int) error {
+	t := cr.Cols()[j].Type
+	switch t {
+	case flux.TString:
+		if cr.Strings(j).IsNull(i) {
+			return builder.AppendNil(dst)
+		}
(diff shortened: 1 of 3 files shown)


jsternberg:
Yea, it's kind of funky. I'm not sure why the number of allocations doesn't change.

jsternberg:
It might be the time period. When I set the benchmark to generate points for 24 hours, I get these:

```
# before
jsternberg:[~/g/s/g/i/flux] $ (perf/append-row-value *+$=) go test -run=NONE -bench=FixedWindow_Process ./stdlib/universe/
goos: darwin
goarch: amd64
BenchmarkFixedWindow_Process-12                1        1848452786 ns/op        851720648 B/op  16008028 allocs/op
PASS
ok      github.com/influxdata/flux/stdlib/universe      3.208s
# after
jsternberg:[~/g/s/g/i/flux] $ (perf/append-row-value *$=) go test -run=NONE -bench=FixedWindow_Process ./stdlib/universe/
goos: darwin
goarch: amd64
BenchmarkFixedWindow_Process-12                1        1476167638 ns/op        663714152 B/op   7713627 allocs/op
PASS
ok      github.com/influxdata/flux/stdlib/universe      2.860s
```

When I do the benchmark in this commit, it's an 8 byte difference for allocations that seems to go out of control when dealing with larger tables. I'm going to see if I can improve the benchmark...

Issue #1224, opened 2019-04-29
Return better error message when range not specified


jlapacik:
All Flux queries must be time bounded using a call to `range`. Currently `range` must be called directly following a call to `from`. See #1215 . If a Flux query is not time bounded or if a `range` cannot be pushed down, the following error will be thrown:

    Error: <invalid> compilation failed: cannot set trigger spec on plan node of type *plan.LogicalNode

This is a very confusing error message for a user. It should be replaced with something like:

    Error: query planning failed: query must be time bounded with a call to range

PR #1242, opened 2019-05-03
refactor(plan): do not error for logical node when setting trigger specs

jlapacik:
This is needed in the fix for #1224.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/plan/physical.go
+++ b/plan/physical.go
@@ -3,8 +3,6 @@ package plan
 import (
 	"fmt"
 	"math"
-
-	"github.com/pkg/errors"
 )
 
 // PhysicalPlanner performs transforms a logical plan to a physical plan,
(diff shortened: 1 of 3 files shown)


plan/physical.go
```diff
@@ -70,6 +58,16 @@ func (pp *physicalPlanner) Plan(spec *Spec) (*Spec, error) {
 		}
 	}
 
+	// Compute time bounds for nodes in the plan
+	if err := transformedSpec.BottomUpWalk(ComputeBounds); err != nil {
```
wolffcm: I'm not sure about setting the trigger specs, but wouldn't you want to propagate time ranges before doing validation?

I think this is fine for now, but the complexity of how the different passes interact will increase over time... Intuitively it seems like validation should be last, since you wouldn't want to validate and then do something that invalidates the plan; in that case we wouldn't be able to catch it.

jlapacik: You're right. Perhaps the mechanism for setting trigger specs needs to change?

Basically trigger specs are a property of physical plan nodes. However, if a range is not pushed down, then the output of physical planning is a plan with a logical `FromProcedureSpec`. So because we set the trigger specs first, we try to set a trigger spec on the logical `From` node, but we error because it should be a physical plan node.

wolffcm: Maybe make the trigger-setting pass be more robust and do nothing for logical nodes?  We know that validation will catch them and produce a reasonable error.

jlapacik:
@wolffcm I have updated the trigger spec method to pass over logical plan nodes without producing an error as validation will always catch such malformed physical query plans. Now with https://github.com/influxdata/influxdb/pull/13786, any query that doesn't specify or cannot push down a range will fail with a meaningful error message.

plan/physical.go
```diff
@@ -88,11 +86,9 @@ func validatePhysicalPlan(plan *Spec) error {
 		if validator, ok := pn.ProcedureSpec().(PostPhysicalValidator); ok {
 			return validator.PostPhysicalValidate(pn.ID())
 		}
-
 		if _, ok := pn.(*PhysicalPlanNode); !ok {
-			return errors.Errorf("logical node \"%v\" could not be converted to a physical node", pn.ID())
+			return fmt.Errorf("invalid physical query plan; found logical operation \"%v\"", pn.ID())
 		}
-
```
wolffcm: I guess you could also add a check in here to check that the trigger is not `nil`.

PR #1245, opened 2019-05-06
chore: update ragel version used by CI

### LINKED ISSUE #1183: CI requires particular version of ragel to pass checkgenerate
If you modify `internal/scanner/scanner.rl` and regenerate `scanner.gen.go` your changes may fail in CI if your local version of ragel is different than in CI.

I have ragel 6.10 installed via `brew`.  CI has 6.9, and this causes some small differences in the generated code between local development and CI.

The workaround is to build and install an earlier version, which is not that hard to do, as older versions are here:
http://www.colm.net/files/ragel/

wolffcm:
This was causing headaches because most users are using ragel 6.10 on
local machines, but the Docker image used 6.9 (several years old now).

Fixes #1183.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/Dockerfile_build
+++ b/Dockerfile_build
@@ -1,8 +1,13 @@
 # This file describes an image that is capable of building Flux.
 
-FROM golang:1.12
+FROM golang:1.12-alpine
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
-		ruby \
-		ragel=6.9-1.1+b1 \
-	&& rm -rf /var/lib/apt/lists/*
+# Note: gcc and libc-dev are only needed since go test -race requires CGO.
+RUN apk add --no-cache \
+        bash \
+        git \
+        make \
+        gcc \
+        libc-dev \
+        ruby \
+        ragel=6.10-r0
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,13 @@ GO_ARGS=-tags '$(GO_TAGS)'
 
 # Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
+export GO_BUILD=env GO111MODULE=on CGO_ENABLED=0 go build $(GO_ARGS)
+export GO_TEST=env GO111MODULE=on CGO_ENABLED=0 go test $(GO_ARGS)
+# go test -race requires CGO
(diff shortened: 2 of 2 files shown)


wolffcm:
@nathanielc I was able to build and test this new `Dockerfile_build` locally and it seems okay, but CI is not picking it up.  Do I need to update the image ` nathanielc/flux-build:latest` in a Docker registry somewhere?

nathanielc:
@wolffcm Yes, its a bit backwards, but the source of the image is the `Dockerfile_build` from the Flux repo, but the docker image CI uses will not be updated till the changes are merged to master.  So its a bit of chicken and egg issue. 

One way to solve would be to split this PR, on PR to change the docker image  and one for the changes to the generated scanner. The docker image PR will go green, since it doesn't really change anything in CI. Then the second PR can be merged. The only catch is that no other PRs will go green until the changes to the scanner are merged.

wolffcm:
@nathanielc Ok, makes sense.  I made this PR just update the dockerfile, and created a separate PR for updating the generated code:
https://github.com/influxdata/flux/pull/1246

Dockerfile_build
```diff
@@ -1,8 +1,12 @@
...
-RUN apt-get update && apt-get install -y --no-install-recommends \
-		ruby \
-		ragel=6.9-1.1+b1 \
-	&& rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache \
+        bash \
+        git \
+        make \
+        gcc \
+        libc-dev \
```
nathanielc: Why are gcc and libc-dev needed?

wolffcm: I found that `go vet` seems to use CGO, so it needs them.  We wouldn't need them if we set `CGO_ENABLED=0` in the environment.  Do you have a preference?

wolffcm: I did a little more digging on this issue, I think that `go vet` needs to build something for some reason, and through transitive dependencies something that needs CGO (I seem to recall that the `net` package uses CGO) gets compiled.

This is a recent change in Go 1.11 that seems not to have been fixed in Go 1.12:
https://github.com/golang/go/issues/26988

Maybe the best way forward would be to add CGO_ENABLED=0 to the environment, and if/when that Go bug gets fixed, we won't have any unneeded dependencies in the docker build.

wolffcm:
As it turns out, it's not possible to use `go test -race` with an Alpine-based image:
https://github.com/golang/go/issues/14481
Incidentally, `go test -race` also requires CGO, so we would need to include `gcc` and `libc-dev` dependencies anyway, but even with them there I get the link errors described in the golang issue.

I'll update the issue to reflect this, a different solution is needed.

PR #1249, opened 2019-05-07
refactor(stdlib/universe): convert last to index selector

mark-rushakoff:
I was looking at the implementation of first and last, and I noticed
that first was an index selector while last was a row selector. It
seemed strange that they weren't a nearly identical implementation. Now
they are both index selectors, which improved the benchmarks for last:

```
name    old time/op    new time/op    delta
Last-4     395ns ± 6%     262ns ±13%  -33.61%  (p=0.000 n=8+8)

name    old alloc/op   new alloc/op   delta
Last-4      224B ± 0%      185B ± 0%  -17.41%  (p=0.000 n=8+8)

name    old allocs/op  new allocs/op  delta
Last-4      9.00 ± 0%      2.00 ± 0%  -77.78%  (p=0.000 n=8+8)
```

### DIFF
--- a/stdlib/universe/last.go
+++ b/stdlib/universe/last.go
@@ -74,7 +74,7 @@ func (s *LastProcedureSpec) TriggerSpec() plan.TriggerSpec {
 }
 
 type LastSelector struct {
-	rows []execute.Row
+	selected bool
 }
 
 func createLastTransformation(id execute.DatasetID, mode execute.AccumulationMode, spec plan.ProcedureSpec, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
Sure thing -- I changed the base from master to dev.

nathanielc:
@jlapacik @mark-rushakoff I think this breaks the implementation of last. With the change it selects the last row per flux.ColReader instead of the last row of the flux.Table. It looks like the test cases only have a single flux.ColReader per table so it wasn't caught. 

Can either of you look into adding a test case that uses multiple flux.ColReaders to ensure that this works? I could be remembering the internals wrong, but we need the test case anyways.

mark-rushakoff:
@nathanielc @jlapacik I don't have the bandwidth for that ask right now -- do you want to make an issue so this doesn't fall off the radar?

jlapacik:
I created the following issue #1260 and am looking into it.

jlapacik:
#1264 updates the tests for `last` as well as other selectors (`max` and `min`) that only used a single column reader per table. It also reverts this commit which unfortunately breaks the implementation of `last` when multiple column readers are needed/used to read a flux table.

PR #1264, opened 2019-05-08
test(universe): use multiple column readers to iterate over tables

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR reverts a previous change (#1249) that converted the `last` function from a row selector to an index selector for performance reasons. Unfortunately the implementation was not correct when using multiple column readers to iterate over the rows of a table.

This PR refactors the tests for `last` to use multiple column readers when iterating over a table. It also does the same for the index selector functions `min` and `max` which previously only used a single column reader when reading a table.

### DIFF
--- a/stdlib/universe/last_test.go
+++ b/stdlib/universe/last_test.go
@@ -91,7 +91,9 @@ func TestLast_Process(t *testing.T) {
 			executetest.RowSelectorFuncTestHelper(
 				t,
 				new(universe.LastSelector),
-				tc.data,
+				&executetest.RowWiseTable{
+					Table: tc.data,
+				},
 				tc.want,
 			)
 		})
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
You said this reverts #1249, but it looks like this only updates test files?

jlapacik:
@mark-rushakoff I meant to say reset. It resets HEAD to the previous commit and then updates the test files. I figured since this is the dev branch it's ok to rewrite its history. Although if you'd rather not lose this commit, I can do a proper revert instead.

mark-rushakoff:
The files changed view still only shows the test files being changed, so I don't think GitHub is doing what you want it to do here.

Either way, I have no preference if the commit is explicitly reverted or if the original change is rebased out of the dev branch.

It might be good to add a comment somewhere in last.go explaining why it's a different implementation from `first`; that's what I noticed and my motivation for changing the implementation of `last`.

jlapacik:
Closing this PR in favor of #1266 .

PR #1270, opened 2019-05-10
refactor: modify the release script to push the latest tag to a maint branch

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The `maint` branch will hold the last release. The script will also
ensure that the current branch has `maint` as an ancestor so any
commits made to the `maint` branch are ensured to be in the next
release.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -11,9 +11,18 @@ if ! env | grep GITHUB_TOKEN= > /dev/null; then
   exit 1
 fi
 
+# Run git fetch to ensure that the origin is updated.
+git fetch
+
+# Ensure that the maint branch is included in this release.
+if ! git merge-base --is-ancestor origin/maint HEAD; then
+  echo "maint branch has not been merged into $(git rev-parse --abbrev-ref HEAD)." 2>&1
+  exit 1
+fi
+
 export GO111MODULE=on
 
 version=$(go run ./internal/cmd/changelog nextver)
 git tag -s -m "Release $version" $version
-git push origin $version
+git push origin "$version" "HEAD:maint"
 go run github.com/goreleaser/goreleaser release --rm-dist --release-notes <(go run ./internal/cmd/changelog generate --version $version --commit-url https://github.com/influxdata/flux/commit)


release.sh
```diff
@@ -11,9 +11,18 @@ if ! env | grep GITHUB_TOKEN= > /dev/null; then
...
+  echo "maint branch has not been merged into $(git rev-parse --abbrev-ref HEAD)." 2>&1
+  exit 1
+fi
+
 export GO111MODULE=on
 
 version=$(go run ./internal/cmd/changelog nextver)
 git tag -s -m "Release $version" $version
-git push origin $version
+git push origin "$version" "HEAD:maint"
```
nathanielc: What does this line do? Is it just moving the `main` branch to the latest commit?

jsternberg: Yea. It's moving whatever the `HEAD` ref is (which is what you just tagged) and moving `maint` to point to that, but without having to modify anything locally.

Unfortunately, you can't put the tag in there directly so `$version:maint` doesn't work. I tried it out in a test repository before committing this.

release.sh
```diff
@@ -11,9 +11,18 @@ if ! env | grep GITHUB_TOKEN= > /dev/null; then
   exit 1
 fi
 
+# Run git fetch to ensure that the origin is updated.
+git fetch
```
nathanielc: I think we should revert this change. I typically do a fetch and then do some testing etc before running the `make release` script. I would rather the script not auto update, if something lands in master maint during that local testing, I am fine if it waits till next release.

jsternberg: This is to ensure that `git merge-base --is-ancestor` works correctly. It won't change your local branch. If you don't do this fetch, there's a larger chance that the below push will fail after you have already tagged it.

It's minimal since this script shouldn't be run very frequently anyway, but it exists as a potential problem.

It's not a pull so it won't affect anything local on your machine.

nathanielc: OK yeah that makes sense.

PR #1273, opened 2019-05-13
ci: add codecov.io support to flux

### LINKED ISSUE #1094: Integrate code coverage in the CI process
Adding code coverage to the CI process can be a good way of ensuring that new code doesn't decrease the coverage of tests. There are tools that integrate with GitHub that allow you to upload your coverage report and it will provide a summary of how the coverage changed between two commits.

[CodeCov](https://codecov.io/) is one example that I saw on another repository and it would likely suit our needs perfectly. We would just need to update the tests to run with coverage enabled and to upload the results and we would get the benefits of code coverage tracking.

It is free for open source to use so we wouldn't have to pay anything and can always remove it if it doesn't work well.

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

Fixes #1094.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,11 @@ jobs:
       - run: make vet
       - run: make staticcheck
       - run: make checkgenerate
-      - run: make test
+      - run: make test GO_TEST_FLAGS='-coverprofile=coverage.txt -covermode=atomic'
+      - run:
+          name: Uploading coverage report
+          command: |
+            bash <(curl -s https://codecov.io/bash)
       - save_cache:
           name: Saving GOCACHE
           key: flux-gocache-{{ .Branch }}-{{ .Revision }}
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor
 .vscode
 .vs
 .tern-project
+/coverage.txt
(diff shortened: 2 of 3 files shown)


jsternberg:
Here's the output showing this works: https://codecov.io/gh/influxdata/flux/branch/refactor%2Fcodecov-ci

It won't paste anything into the issue until after it is merged to master. Once it is merged to master, it'll show a comparison for every pull request.

nathanielc:
@jsternberg How do we use this report when reviewing PRs? I get that the report is hard to read right now because there is no baseline to compare against, but once we do, how do you intend for use to consume this information?

Is it that we won't merge PRs that reduce code coverage? Is it just informational to help us think about test cases? What are the goals of adding code coverage?

jsternberg:
I want it so we don't merge PRs unless they increase code coverage, but it's obviously flexible. I think the goal of adding code coverage is to give us a signal that we didn't add tests or, if we did add tests, we didn't add tests in places that we thought we might have.

At the moment our coverage seems to be at 54%. If we can get that up to above 80% and have it stay there, I think that will increase our test quality.

nathanielc:
I see that there are two checks associated with codecov `codecov/patch` and `codecov/project` what do those do?

Is it possible to have the check be green on increased code coverage and yellow on decreased coverage? If so is it possible to have the code coverage report linked from those health check details instead of as a comment?

jsternberg:
I had never seen that, but apparently yes. Here is the configuration file for it: https://docs.codecov.io/docs/codecov-yaml

I think this might be what you're specifically looking for: https://docs.codecov.io/docs/commit-status

nathanielc:
Yes, lets try this out, what I would like to see is a github bot integration that acts a little like the semantic commit message bot. It never goes red, but will be yellow if the coverage goes down. That way it becomes a good reminder and positive feedback, but it doesn't prevent merged. That said, this will likely all change once we switch to the only bots merge to master approach. So lets go forward with this as is.

Issue #1251, opened 2019-05-07
Move controller to InfluxDB


affo:
Why is the controller in the Flux repo?

Flux is a language and an engine.

The controller should be part of InfluxDB that hosts the engine and offers it as a service in a multi-tenant fashion.

nathanielc:
The current reason is because the REPL uses it. Once we refactor the REPL that will no longer be the case and we can move the controller to influxdb.

PR #1277, opened 2019-05-14
fix(control): remove the race condition within the (*Query).Done method

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

If cancel and done were called in quick succession, it could result in a
race condition because there was no lock over the state transition
inside of done. It was assumed that done was called after results were
read, but that's not part of the interface and you are allowed to call
done without having read the results. This became most obvious when
cancel was called and then done immediately afterwards.

This modifies it so that cancel is asynchronous and signals that the
execution should be halted, which is how it was previously. It modifies
done to wait for the task to actually be completed before the call
returns.

Fixes influxdata/influxdb#13867.

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -191,6 +191,7 @@ func (c *Controller) createQuery(ctx context.Context, ct flux.CompilerType) (*Qu
 		parentCtx:          parentCtx,
 		parentSpan:         parentSpan,
 		cancel:             cancel,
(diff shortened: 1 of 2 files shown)


jsternberg:
This is probably more proof that we need to move this to #1251 so that we can have a discussion of what the influxdb controller needs and implement it there. It doesn't really do anything inside of flux and just causes issues currently.

PR #1278, opened 2019-05-15
feat(stdlib/universe): add functionality to test records for missing values.

aanthony1243:
added exists function to test row records for missing values

updated spec

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


docs/SPEC.md
```diff
@@ -1903,6 +1903,11 @@ from(bucket:"telegraf/autogen")
                 r.service == "app-server")
 ```
 
+##### Helper Predicate Functions
+Two predicate functions that return boolean values and are useful in filters are: 
+* `contains(value:T0, set: [T0...])`: tests whether the `set` contains the given value.  The type of value and set can vary, but must match.
+* `exists(r:R0, key: "key")`: tests whether the record `r` has a field `key`.  
+
```
nathanielc: Contains was already documented in the SPEC. Can you move its doc to this section of the SPEC? And give `exists` its own heading as well? This should make them easier to find.

nathanielc: It looks like we are inconsistent in referring to object keys are either keys or properties. The grammar referres to them as a `property` and the text of the SPEC occasionally uses `key`. Lets standardize on one, I vote for `property` as that is what the grammar uses.

stdlib/universe/filter.go
```diff
@@ -146,7 +146,10 @@ func (t *filterTransformation) Process(id execute.DatasetID, tbl flux.Table) err
 	// Prepare the function for the column types.
 	cols := tbl.Cols()
 	if err := t.fn.Prepare(cols); err != nil {
-		// TODO(nathanielc): Should we not fail the query for failed compilation?
+		if _, ok := err.(*execute.UnknownColumnError); ok {
```
nathanielc: Referencing an unknown column in other contexts is an error. Why in this case do we ignore it and just assume the predicate evals to false?

stdlib/testing/testdata/filter_nulls.flux
```diff
@@ -0,0 +1,57 @@
+package testdata_test
```
nathanielc: Does it make sense to rename the files for these tests to `_exists` instead of `_nulls` as the function under test is the `exists` function?

docs/SPEC.md
wolffcm: I've always thought of it like "a property is composed of a key/value pair."

aanthony1243: yes, I must have  missed the other reference to contains.  will fix.

stdlib/testing/testdata/filter_nulls.flux
aanthony1243: yeah, good point. will change.

stdlib/universe/filter.go
aanthony1243: hm.  maybe I ought to pull this into another PR.  and get more discussion going on it.  

You are right, this is wrong, not in the sense that there should sometimes be an error.  In the context of filters, I don't think a missing column should ever be an error.  I'll make a new issue with more details and remove this change from this PR.

docs/SPEC.md
nathanielc: That makes sense, perhaps we should write that phrase at least once in the SPEC to make the association?

jlapacik:
Hey @aanthony1243 so ignore my previous comment. I agree with you that there should just be a single function for asserting existence. Where I differ is on the arguments to that function. Instead of taking in an object and a property name, this function should be used to test any value for existence.

So far we haven't had any real support for null values in Flux. However that is changing and so this function should support any nullable type being passed in.

Also since this will be called mainly from within filter's lambda function and will be involved in query planning optimizations for influxdb storage, I would prefer for it to be a unary operator so that we don't have any complications from identifier scoping later on down the road. But that's just my opinion.

compiler/runtime.go
```diff
@@ -28,19 +28,23 @@ type Func interface {
 
 type Evaluator interface {
 	Type() semantic.Type
-	EvalString(scope Scope) (string, error)
-	EvalInt(scope Scope) (int64, error)
-	EvalUInt(scope Scope) (uint64, error)
-	EvalFloat(scope Scope) (float64, error)
-	EvalBool(scope Scope) (bool, error)
-	EvalTime(scope Scope) (values.Time, error)
+	EvalString(scope Scope) (values.Value, error)
```
aanthony1243: summarize this change:  the old `Eval<Type>`  functions were retrieving the value, converting it to a `go` type and then returning it.  Then this value was taken and converted back into a `values.Value`
the problem is that converting a null value causes a panic
so this sends that null further up the chain (edited) 
hidden inside a `values.value`
so it's actually not a big change, but just a large number of tiny refactor changes

PR #1286, opened 2019-05-16
refactor: mandate that the table do method can only be called once

jsternberg:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/executetest/table.go
+++ b/execute/executetest/table.go
@@ -2,6 +2,7 @@ package executetest
 
 import (
 	"fmt"
+	"testing"
 
 	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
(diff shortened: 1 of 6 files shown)


jsternberg:
I had to make some small changes to get it compiling and testable, but here's the output of `executetest.Table` with minimal changes from the test case that was written.

```
--- FAIL: TestTable (0.00s)
    --- FAIL: TestTable/Normal (0.00s)
        table.go:505: table is not done after calling Do
    --- FAIL: TestTable/MultipleDoCalls (0.00s)
        table.go:519: expected error when calling Do twice
```

In the second commit, I fix `executetest.Table` to follow the new pattern and it then passes the tests.

jsternberg:
This was done as part of a separate PR.

PR #1293, opened 2019-05-20
fix(lang): inject context aware values in scope for each new program execution

affo:
This is required for dynamic query support in InfluxDB

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -34,6 +34,7 @@ func Parse(flux string) (*ast.Package, error) {
 }
 
 // Eval accepts a Flux script and evaluates it to produce a set of side effects (as a slice of values) and a scope.
+// ScopeMutators passed are applied to the prelude.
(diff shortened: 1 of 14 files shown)


affo:
@wolffcm @jlapacik @jsternberg 
Hey guys, can you please review this PR as it is?

I made a change  to the `flux.Program` interface so that it receives a `flux.ExecxutionContext` that contains everything that a program needs in order to run.  Once (and if) this gets approved, I will create a PR in `flux-staging` that applies the changes to the controller and fixes the usage of the `Program` interface.

I didn't use `execute.Administration`, because that is intended to be used by the executor on a per-node basis. And that doesn't actually fit this case.

nathanielc:
The implementation of this change depends on #1321, closing until we have a better understanding.

PR #1295, opened 2019-05-21
fix(http): reuse tcp sockets

### LINKED ISSUE #1294: http.to doesn't reuse sockets.
The body buffer in http.to is never read through or flushed, just thrown away which means that the TCP socket won't be reused.

this would fix it.
```
res, _ := client.Do(req)
io.Copy(ioutil.Discard, res.Body)
res.Body.Close()
```

https://github.com/influxdata/flux/blob/a7247c91fb2787b761ad72396bddf49ec9976d6e/stdlib/http/to.go#L442

docmerlin:
closes #1294

### DIFF
--- a/stdlib/http/to.go
+++ b/stdlib/http/to.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


docmerlin:
This no longer applies as golang flushes a fair bit on its own on close now

jsternberg:
I ended up getting done in #1835.

PR #1301, opened 2019-05-28
fix(execute): copy the table when a table is used multiple times

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

This changes the code that sends a table to multiple downstream
transformations to copy the table into a buffered table and then to use
a new copy of the table for each transformation.

### DIFF
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -3,6 +3,7 @@ package executetest
 import (
 	"context"
 
+	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
@@ -45,10 +46,44 @@ func (src *FromProcedureSpec) AddTransformation(t execute.Transformation) {
 
 func (src *FromProcedureSpec) Run(ctx context.Context) {
 	id := execute.DatasetID(uuid.NewV4())
-	for _, t := range src.ts {
+
+	if len(src.ts) == 0 {
+		return
+	} else if len(src.ts) == 1 {
+		t := src.ts[0]
+
 		var max execute.Time
 		for _, tbl := range src.data {
 			t.Process(id, tbl)
+			stopIdx := execute.ColIdx(execute.DefaultStopColLabel, tbl.Cols())
+			if stopIdx >= 0 {
+				if s := tbl.Key().ValueTime(stopIdx); s > max {
+					max = s
+				}
+			}
+		}
+		t.UpdateWatermark(id, max)
+		t.Finish(id, nil)
+		return
+	}
+
+	buffers := make([]flux.BufferedTable, 0, len(src.data))
+	for _, tbl := range src.data {
+		bufTable, _ := execute.CopyTable(tbl)
+		buffers = append(buffers, bufTable.(flux.BufferedTable))
+	}
+
+	// Ensure that the buffers are released after the source has finished.
+	defer func() {
+		for _, tbl := range buffers {
+			tbl.Done()
+		}
+	}()
+
+	for _, t := range src.ts {
+		var max execute.Time
+		for _, tbl := range buffers {
+			t.Process(id, tbl.Copy())
 			stopIdx := execute.ColIdx(execute.DefaultStopColLabel, tbl.Cols())
 			if stopIdx >= 0 {
 				if s := tbl.Key().ValueTime(stopIdx); s > max {
--- a/execute/table.go
+++ b/execute/table.go
@@ -36,43 +36,90 @@ func GroupKeyForRowOn(i int, cr flux.ColReader, on map[string]bool) flux.GroupKe
 	return NewGroupKey(cols, vs)
 }
 
-// OneTimeTable is a Table that permits reading data only once.
-// Specifically the ValueIterator may only be consumed once from any of the columns.
-type OneTimeTable interface {
-	flux.Table
-	onetime()
+// tableBuffer maintains a buffer of the data within a table.
+// It is created by reading a table and using Retain to retain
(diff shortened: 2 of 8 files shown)


execute/table.go
```diff
@@ -36,43 +36,80 @@ func GroupKeyForRowOn(i int, cr flux.ColReader, on map[string]bool) flux.GroupKe
...
-	for j, c := range cols {
-		colMap[j] = j
-		if _, err := builder.AddCol(c); err != nil {
-			return nil, err
+func (tb *tableBuffer) Cols() []flux.ColMeta {
+	return tb.colMeta
+}
+
+func (tb *tableBuffer) Do(f func(flux.ColReader) error) error {
+	defer tb.Done()
```
affo: By deferring Done, don't you Release twice?

execute/table.go
```diff
@@ -36,43 +36,80 @@ func GroupKeyForRowOn(i int, cr flux.ColReader, on map[string]bool) flux.GroupKe
...
+
+func (tb *tableBuffer) Empty() bool {
+	return len(tb.buffers) == 0
+}
+
+func (tb *tableBuffer) Copy() flux.BufferedTable {
+	for i := tb.i; i < len(tb.buffers); i++ {
+		tb.buffers[i].Retain()
+	}
+	return &tableBuffer{
```
affo: Couldn't you simply return tb? 🤔

execute/table.go
jsternberg: `Done` is idempotent. The iteration calls `Release` as it processes tables, but it keeps the current index so it won't iterate over that section again.

execute/table.go
jsternberg: No since it would break the contract and then I'd have to update this when I start getting things to follow the interface correctly. This effectively makes it so that the two tables have the same contents, but will have different `i` values if you iterate over them separately.

stdlib/universe/table_fns_test.go
```diff
@@ -346,6 +348,8 @@ t = inj |> tableFind(fn: (key) => key.user == "user1")`
 // We have to write this test as a non-standard e2e test, because
 // our framework doesn't allow comparison between "simple" values, but only streams of tables.
 func TestIndexFns_EndToEnd(t *testing.T) {
+	t.Skip("table cannot be used more than once")
```
affo: I should file an issue for this because it breaks tableFind/getColumn/GetRecord use.

jsternberg: Yes. I can help you with this. I think we will have to have `objects.NewTable` call `CopyTable` and then we will have to find a way to have a scope call `Done` on tables after they go out of scope so we can release the values.

execute/table_test.go
```diff
@@ -253,3 +255,70 @@ func TestColListTable_SetNil(t *testing.T) {
...
+			return nil
+		}); err != nil {
+			t.Fatalf("unexpected error: %s", err)
+		}
+	}
+
+	// Ensure we can copy the table and read a point from the
+	// column reader without panicking.
+	for _, buf := range buffers {
+		cpy := buf.Copy()
```
wolffcm: So would it be an error to call `Do(..)` on `buf` directly here without the call to `Copy()`?  If so, would be good to verify that...

jsternberg: Yes it would. This is going to be part of the common table tests.

execute/executetest/source.go
```diff
@@ -58,6 +64,32 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
...
+				if s := tbl.Key().ValueTime(stopIdx); s > max {
+					max = s
+				}
+			}
+		}
+		t.UpdateWatermark(id, max)
+		t.Finish(id, nil)
+	}
+
+	for _, tbl := range buffers {
```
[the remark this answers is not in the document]
jsternberg: Sure. That seems like a good idea.

execute/table.go
```diff
@@ -36,43 +36,80 @@ func GroupKeyForRowOn(i int, cr flux.ColReader, on map[string]bool) flux.GroupKe
...
-// CacheOneTimeTable returns a table that can be read multiple times.
-// If the table is not a OneTimeTable it is returned directly.
-// Otherwise its contents are read into a new table.
-func CacheOneTimeTable(t flux.Table, a *memory.Allocator) (flux.Table, error) {
-	_, ok := t.(OneTimeTable)
-	if !ok {
-		return t, nil
-	}
-	return CopyTable(t, a)
+type tableBuffer struct {
```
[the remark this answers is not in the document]
jsternberg: Will do.

execute/table.go
```diff
@@ -36,43 +36,79 @@ func GroupKeyForRowOn(i int, cr flux.ColReader, on map[string]bool) flux.GroupKe
...
+	key     flux.GroupKey
+	colMeta []flux.ColMeta
+	i       int
+	buffers []flux.ColReader
```
aanthony1243: I get what's happening, and why this works but I'm also concerned.  The purpose (though rarely realized) for the `flux.ColReader`  is to chunk data that is too large to fit into memory.  

Given that most of our streams end up holding all of the data in memory anyway, this is perhaps not a big issue.  But you may want to add a TODO about how this tableBuffer may not be sufficient on massive data sets (if that's even a design goal still, idk).  

Another suggestion would be to make it an `memoryTableBuffer` to make it clearer that the data will be buffered wholly in memory.

jsternberg: I agree with this, but I figure that this will be a general consequence of certain operations. I think we included the allocator to catch when the user tries to allocate too much memory and we're going to have to be prepared for the times when a user asks us to buffer a large dataset in memory. I'll add a comment describing this though.

PR #1303, opened 2019-05-28
Add "with" operator support to semantic  package

nathanielc:
This is an old branch I had that adds "with" support to type inference for most common use cases.

See #205 
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -880,6 +880,7 @@ func (e *ArrayExpression) Copy() Node {
 // ObjectExpression allows the declaration of an anonymous object within a declaration.
 type ObjectExpression struct {
 	BaseNode
+	With       *Identifier `json:"with,omitempty"`
 	Properties []*Property `json:"properties"`
 }
 
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -676,10 +676,19 @@ Literals construct a value.
 
 Object literals construct a value with the object type.
 
-    ObjectLiteral = "{" PropertyList "}" .
-    PropertyList  = [ Property { "," Property } ] .
-    Property      = identifier [ ":" Expression ]
-                  | string_lit ":" Expression .
+    ObjectLiteral  = "{" ObjectBody "}" .
+    ObjectBody     = WithProperties | PropertyList .
+    WithProperties = identifier "with"  PropertyList .
+    PropertyList   = [ Property { "," Property } ] .
+    Property       = identifier [ ":" Expression ]
+                   | string_lit ":" Expression .
+
+Examples:
+
+    {a: 1, b: 2, c: 3}
+    {a, b, c}
+    {o with x: 5, y: 5}
+    {o with a, b}
 
 ##### Array literals
 
--- a/execute/row_fn.go
+++ b/execute/row_fn.go
@@ -41,17 +41,13 @@ func newRowFn(fn *semantic.FunctionExpression) (rowFn, error) {
 func (f *rowFn) prepare(cols []flux.ColMeta, extraTypes map[string]semantic.Type) error {
 	// Prepare types and recordCols
 	propertyTypes := make(map[string]semantic.Type, len(f.references))
+	for j, c := range cols {
+		propertyTypes[c.Label] = ConvertToKind(c.Type)
+		f.recordCols[c.Label] = j
+	}
+
 	for _, r := range f.references {
-		found := false
-		for j, c := range cols {
-			if r == c.Label {
-				f.recordCols[r] = j
-				found = true
-				propertyTypes[r] = ConvertToKind(c.Type)
-				break
-			}
-		}
-		if !found {
+		if _, ok := f.recordCols[r]; !ok {
 			return fmt.Errorf("function references unknown column %q", r)
 		}
 	}
@@ -126,8 +122,8 @@ func (f *rowFn) eval(row int, cr flux.ColReader, extraParams map[string]values.V
 		return nil, errors.New("null reference used in row function: skipping evaluation until null support is provided")
 	}
 
-	for _, r := range f.references {
-		f.record.Set(r, ValueForRow(cr, row, f.recordCols[r]))
+	for r, col := range f.recordCols {
+		f.record.Set(r, ValueForRow(cr, row, col))
 	}
 	f.inRecord.Set(f.recordName, f.record)
 	for k, v := range extraParams {
--- a/internal/parser/grammar.md
+++ b/internal/parser/grammar.md
@@ -61,7 +61,7 @@ The parser directly implements the following grammar.
     MemberExpression               = DotExpression  | MemberBracketExpression
     DotExpression                  = "." identifer
     MemberBracketExpression        = "[" string "]" .
-    CallExpression                 = "(" PropertyList ")" .
+    CallExpression                 = "(" ParameterList ")" .
     IndexExpression                = "[" Expression "]" .
     PrimaryExpression              = identifer
                                    | int_lit
@@ -73,7 +73,8 @@ The parser directly implements the following grammar.
                                    | ObjectLiteral
                                    | ArrayLiteral
                                    | ParenExpression .
-    ObjectLiteral                  = "{" PropertyList "}" .
+    ObjectLiteral                  = "{" ObjectLiteralBody "}"
+    ObjectLiteralBody              = [ ObjectBody ]
     ArrayLiteral                   = "[" ExpressionList "]" .
     ParenExpression                = "(" ParenExpressionBody .
     ParenExpressionBody            = ")" FunctionExpressionSuffix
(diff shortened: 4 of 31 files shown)


docs/SPEC.md
```diff
@@ -626,10 +626,19 @@ Literals construct a value.
...
+    PropertyList   = [ Property { "," Property } ] .
+    Property       = identifier [ ":" Expression ]
+                   | string_lit ":" Expression .
+
+Examples:
+
+    {a: 1, b: 2, c: 3}
+    {a, b, c}
+    {o with x: 5, y: 5}
+    {o with a, b}
```
jlapacik: What if `with` were just a binary operator (operating on objects)? So instead of

    {o with a, b}

it was

    o with {a, b}

semantic/constraints.go
```diff
@@ -338,11 +341,37 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
+					continue
+				}
+				for _, p := range obj.upper {
+					properties[p] = obj.properties[p]
+				}
+				upper = upper.union(obj.upper)
+			}
+		}
 		v.cs.AddKindConst(nodeVar, ObjectKind{
+			with:       with,
```
jlapacik: This is I think where it would be useful for `with` to be an operator. The constraint phase for object expressions does not need to change. Instead we could add a case for `WithExpression`, and all that would need to be done I believe is to create a kind constraint for the new object expression. And this would be achieved by just taking the union of the lower labels and the upper labels.

Something like this:
```go
case *WithExpression:
    ...
    leftLowerLabels := ...
    rightLowerLabels := ...

    leftUpperLabels := ...
    rightUpperLabels := ...

    v.cs.AddKindConst(nodeVar, ObjectKind{
        properties: properties,
        lower: leftLowerLables.union(rightLowerLabels),
        upper: leftUpperLabels.union(rightUpperLabels),
    }
```

semantic/constraints.go
```diff
@@ -466,14 +495,17 @@ type Constraints struct {
...
 }
+type Kvar int
 
 func (c *Constraints) Copy() *Constraints {
 	n := &Constraints{
 		f:           new(fresher),
 		annotations: make(map[Node]annotation, len(c.annotations)),
 		typeConst:   make([]TypeConstraint, len(c.typeConst)),
 		kindConst:   make(map[Tvar][]Kind, len(c.kindConst)),
+		typeKinds:   make(map[Kvar]Tvar, len(c.typeKinds)),
```
jlapacik: I'm not clear on what this is used for?

semantic/constraints.go
aanthony1243: this looks so much easier to maintain.  the `with` as part of the definition gets messy really fast.

wolffcm: I like it.

What happens if the field already exists?  I guess it would make sense for the RHS of `with` override what's already there.  E.g., what should these do?
```
// (1)
{foo: 1.0} with {foo: 2.0}
// (2)
{foo: 1.0} with {foo: "two dot oh"}
```

jlapacik: So it turns out it's not quite as easy as I'm making it out to be. While I definitely prefer the syntax, my suggestion above won't work with functions that take records as input. So this won't type check:

    f = (r) => r with {foo: "bar"}

And we sort of need this to type check

Now we can work around this for row functions since row functions in transformations like map and filter aren't really type checked as strictly as functions defined outside of such transformations. But I'd rather have all functions, including row functions, be treated the same way by the type system.

So how to support the `with` syntax is still an open question for our type system right now.

To your question @wolffcm it makes sense to override to me as well.

execute/row_fn.go
```diff
@@ -41,17 +41,14 @@ func newRowFn(fn *semantic.FunctionExpression) (rowFn, error) {
...
-		for j, c := range cols {
-			if r == c.Label {
-				f.recordCols[r] = j
-				found = true
-				propertyTypes[r] = ConvertToKind(c.Type)
-				break
-			}
-		}
-		if !found {
+		_, ok := f.recordCols[r]
```
jsternberg: Nitpick: Can you put this declaration and if statement on the same line?

```
if _, ok := f.recordCols[r]; !ok {
```

internal/parser/grammar.md
```diff
@@ -88,9 +88,16 @@ The parser directly implements the following grammar.
     FunctionBodyExpression         = Block | Expression .
     Block                          = "{" StatementList "}" .
     ExpressionList                 = [ Expression { "," Expression } ] .
+    ObjectBody                     = identifier ObjectBodySuffix .
```
jsternberg: The `.` indicates the end of an EBNF expression so it is incorrect to use one on this line since it is followed by an alternate operator `|`. Also, the empty expression needs to be available too so you probably need:

```
ObjectLiteral = "{" [ ObjectBody ] "}"
```

I don't generally like putting so many things on the same line so you may want to do something like this:

```
ObjectLiteral = "{" ObjectLiteralBody "}"
ObjectLiteralBody = [ ObjectBody ]
```

The only thing `ObjectLiteralBody` would do is check if there is more to parse.

internal/parser/grammar.md
```diff
@@ -88,9 +88,16 @@ The parser directly implements the following grammar.
...
+                                   | string_lit PropertyListSuffix .
+    ObjectBodySuffix               = "with" PropertyList
+                                   | PropertyListSuffix .
+    PropertyListSuffix             = PropertySuffix { "," PropertyList } .
     PropertyList                   = [ Property { "," Property } ] .
-    Property                       = identifier [ ":" Expression ]
-                                   | string_lit ":" Expression .
+    Property                       = identifier PropertySuffix
+                                   | string_lit PropertySuffix .
+    PropertySuffix                 = [ ":" Expression ]
```
jsternberg: I think the alternate operator is unnecessary here since `[ a ]` means "this expression is optional" so I think `[ a ]` already means what you put here.

internal/parser/parser.go
```diff
@@ -1054,6 +1052,68 @@ func (p *parser) parseParenIdentExpression(lparen token.Pos, key *ast.Identifier
...
+		return &ast.ObjectExpression{
+			Properties: p.parsePropertyList(),
+		}
+	}
+
+}
+func (p *parser) parseObjectBodySuffix(ident *ast.Identifier) *ast.ObjectExpression {
+	switch _, tok, lit := p.peek(); tok {
+	case token.IDENT:
+		if lit != "with" {
```
jsternberg: Should we be adding `with` as a keyword?

If we don't add `with` as a keyboard, can you somehow fall through to the default so it produces an invalid AST? I think if we see `{a b}` we can assume that this was a mistake for `{a: b}` which the default case handles.

internal/parser/parser.go
```diff
@@ -1085,35 +1145,32 @@ func (p *parser) parsePropertyList() []*ast.Property {
...
 }
 
 func (p *parser) parseIdentProperty() *ast.Property {
 	key := p.parseIdentifier()
+	return p.parsePropertySuffix(key)
+}
 
-	var val ast.Expression
+func (p *parser) parsePropertySuffix(key ast.PropertyKey) *ast.Property {
+
```
jsternberg: Can you remove these random extra lines that are added in various parts of the code? It's a nitpick, but they annoy me.

semantic/constraints.go
```diff
@@ -339,11 +339,37 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
+				return nil, err
+			}
+			tv, ok := t.(Tvar)
+			if !ok {
+				return nil, errors.New("object 'with' identifier must be a type variable")
+			}
+			for _, k := range v.cs.kindConst[tv] {
+				obj, ok := k.(ObjectKind)
+				if !ok {
+					return nil, errors.New("object 'with' identifier must be have only object kind constraints")
```
jsternberg: Looks like a bad sentence?

stdlib/testing/testdata/drop_non_existent.flux
```diff
@@ -42,8 +42,8 @@ t_drop = (table=<-) =>
 		|> range(start: 2018-05-22T19:53:26Z)
 		|> drop(columns: ["non_existent"]))
 
-test _drop_non_existent = () =>
-	({input, want, fn: drop_fn})
+//test _drop_non_existent = () =>
```
jsternberg: What's going on here? Should this line just be removed?

internal/parser/parser.go
aanthony1243: added as a keyword.

ast/ast.go
```diff
@@ -879,6 +879,7 @@ func (e *ArrayExpression) Copy() Node {
 // ObjectExpression allows the declaration of an anonymous object within a declaration.
 type ObjectExpression struct {
 	BaseNode
+	With       *Identifier `json:"with,omitempty"`
```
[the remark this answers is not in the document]
nathanielc: We talked about this offline, but for the record:

The `with` operation must be more restricted than a binary operator on two expressions. Otherwise type inference is undecidable. Spefically the right hand side must always be an object literal. While not strictly required it is much easier to implement if the left hand side is always an identifier. So the type would look more like this:

```
type WithExpression struct {
    Left *Identifier
    Right *ObjectExpression
}
```

As such I think the current implementation of adding the details directly to the ObjectLiteral type is fine.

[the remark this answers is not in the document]
jsternberg: I don't really care too much about the specific syntax, but I'm not sure that it is true the right hand side needs to be an object literal. I think the type system should be able to decide what attributes are inside of the object even if it is an arbitrary identifier and it will also be able to assert that it is an object itself.

I don't want to hold up this PR with that though as I would prefer we merge something and change the syntax later.

PR #1305, opened 2019-05-28
fix(control): recover from panics

wolffcm:
If a panic happens outside of the `execute` package, we would not be able to recover from it.  This change makes it so that we can.

Fixes influxdata/idpe#3409.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/control/controller.go
+++ b/control/controller.go
@@ -20,6 +20,7 @@ package control
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


wolffcm:
Closing this ancient PR whose corresponding issue seems to have already been closed by other means.

PR #1312, opened 2019-05-30
fix(ast): return count of errors when checking AST

### LINKED ISSUE #1311: ast.Check() does not return correct error count
Some errors produced by the parser are silently swallowed.  There is a method `ast.Check()` that is for finding errors in the AST, but it does not always return the correct error count.

For example, `foo(]` is obviously invalid syntax but `ast.Check()` returns zero for the AST generated by the parser.

wolffcm:
Fixes #1311.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/internal/parser/parser_test.go
+++ b/internal/parser/parser_test.go
@@ -51,9 +51,10 @@ func BenchmarkParser(b *testing.B) {
 
 func testParser(runFn func(name string, fn func(t testing.TB))) {
 	for _, tt := range []struct {
-		name string
-		raw  string
-		want *ast.File
+		name  string
+		raw   string
+		want  *ast.File
+		nerrs int
 	}{
 		{
 			name: "package clause",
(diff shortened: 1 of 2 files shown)


internal/parser/parser_test.go
```diff
@@ -51,9 +51,10 @@ func BenchmarkParser(b *testing.B) {
 
 func testParser(runFn func(name string, fn func(t testing.TB))) {
 	for _, tt := range []struct {
-		name string
-		raw  string
-		want *ast.File
+		name      string
+		raw       string
+		want      *ast.File
+		nerrsWant int
```
jsternberg: Can you rename this to just `nerrs`? I don't think the `want` is necessary.

wolffcm: Sure, done.

PR #1314, opened 2019-05-30
docs: null specification

### LINKED ISSUE #1302: Update SPEC with nulls behavior
Nulls are values, fill  in proposal and update SPEC.

jlapacik:
Closes #1302.

This PR specifies how nulls behave in Flux.

I saw two ways of doing this. The first was to define null as a predeclared identifier without a default type but which could take on any basic type (similar to go). This had the advantage of not being able to assign `null` directly to a variable as all Flux variables must be typed. However accessing a non-existent property of an object would have to return typed `null` which is just not feasible.

Therefore the second way the, and the way I ultimately chose, was to define null as both a value and a type. The `null` value being the only value that comprises the _null type_. By introducing a null type to the language it becomes possible to assign `null` directly to a variable which is not ideal. However it also allows us to return `null` when accessing a non-existent property on an object which is what we would like.

In order to still support type inference as well as the type constraints of our current operators (`+` cannot operate on both a _null_ and an _int_ type) I have redefined the basic flux types to be nullable union types. What this means is that the basic flux types are now defined as follows:

    int = int | null
    float = float | null
    string = string | null
    time = time | null

So an integer value for instance can be of type _int_ or of type _null_. A float value can be of type _float_ or of type _null_. And so on and so forth. It might seem a bit convoluted, but I believe this allows us to keep our current type inference while at the same time having consistent null behavior across the language.

### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -402,31 +402,63 @@ Types are never explicitly declared as part of the syntax.
 Types are always inferred from the usage of the value.
 Type inference follows a Hindley-Milner style inference system.
 
-#### Boolean types
+#### Union types
+
+A union type defines a set of types.
+In the rest of this section a union type will be specified as follows:
+
+    T = t1 | t2 | ... | tn
+
+where `t1`, `t2`, ..., and `tn` are types.
+In the example above a value of type `T` is either of type `t1`, type `t2`, ..., or type `tn`.
+
+#### Basic types
+
+These are the types from which all other Flux data types are constructed.
+
+##### Null type
+
+The _null type_ represents a missing or unknown value.
+The _null type_ name is `null`.
+There is only one value that comprises the _null type_ and that is the _null_ value.
+
+A type `t` is nullable if it can be expressed as follows:
+
+    t = {s} | null
+
+where `{s}` defines a set of values.
+
+##### Boolean types
 
 A _boolean type_ represents a truth value, corresponding to the preassigned variables `true` and `false`.
 The boolean type name is `bool`.
+The boolean type is nullable and can be formally specified as follows:
 
-#### Numeric types
+    bool = {true, false} | null
+
+##### Numeric types
 
 A _numeric type_ represents sets of integer or floating-point values.
 
 The following numeric types exist:
 
-    uint    the set of all unsigned 64-bit integers
-    int     the set of all signed 64-bit integers
-    float   the set of all IEEE-754 64-bit floating-point numbers
+    uint    = {the set of all unsigned 64-bit integers} | null
+    int     = {the set of all signed 64-bit integers} | null
+    float   = {the set of all IEEE-754 64-bit floating-point numbers} | null
 
-#### Time types
+Note all numeric types are nullable.
+
+##### Time types
 
 A _time type_ represents a single point in time with nanosecond precision.
 The time type name is `time`.
+The time type is nullable.
 
-
-#### Duration types
+##### Duration types
 
 A _duration type_ represents a length of time with nanosecond precision.
 The duration type name is `duration`.
+The duration type is nullable.
 
 Durations can be added to times to produce a new time.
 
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -1286,13 +1329,25 @@ Within a stream each table's group key value is unique.
...
-A record may be missing a value for a specific column.
-Missing values are represented with a special _null_ value.
-The _null_ value can be of any data type.
+* not _null_ = _null_
+* _null_ or false = _null_
+* _null_ or true = true
+* _null_ or _null_ = _null_
+* _null_ and false = false
+* _null_ and true = _null_
+* _null_ and _null_ = _null_
```
wolffcm: It would be good to include in here the behavior of `if b then ... else ...` where `b` is the `null` value.

It would also be good to call out explicitly that `==` will return `null` if either side is `null`, and even if both sides are `null`.  This can be a bit surprising the first time its encountered.

jlapacik: Ah yes I knew I was forgetting something.

affo: Just analyzing the thing: `not null = null` implies that `null` is both true and false, this implies that `if null then ... else ...` should result in an error? 🤔 What do you think?

It is weird to take a decision when a value has no value...

aanthony1243: this is a bit what I'd been worried about, that we'll see lots of code like   `if r.b != null and r.b then ... else ...`

jsternberg: I'm a bit concerned by this section too. I think this conflates two different solutions for null.

1. SQL method: Null is unknown and is the same as NaN and matches with nothing.
2. Most common programming languages method: Null is a distinct value that is usually falsey.

I don't think you can mix these. In SQL, null can't be evaluated as a boolean so you don't end up with this weird problem. But SQL also has explicit null checks with `IS NULL` and `IS NOT NULL`. Flux acts more like a programming language and uses that paradigm than SQL. I like SQL's method so it pains me a bit to say this, but I think we need to drop the idea that anything with null is null and instead just steal whatever Ruby does since it is the most similar to what we want to do. Python is kind of similar, but they're different because Python consider the zero integer a falsey value.

(Falsey means that the expression evaluates to false in a boolean context, but it is not the same thing as being the false value)

For those unfamiliar, here are the differences:

1. Python considers `False`, `None`, `0`, and empty sequences as falsey values.
2. Ruby considers `false` and `nil` as falsey values.

I think the second is closer to what we're going with. I typed in the above to Ruby and here's what I got:

```
not nil = true
nil or false = false
nil or true = true
nil or nil = nil
nil and false = nil
nil and true = nil
nil and nil = nil
```

The `not` predicate always returns a boolean value. The conditional operators are different though. They return the value that it short circuited on so in `nil or false` it saw that `nil` was falsey so it moved onto the second value which was `false` and returned that as the false value.

This might be a bit hard for our type system to encapsulate though so I would modify the above so that conditionals always return a boolean value.

```
not nil = true
nil or false = false
nil or true = true
nil or nil = false
nil and false = false
nil and true = false
nil and nil = false
```

docs/SPEC.md
```diff
@@ -1286,13 +1329,25 @@ Within a stream each table's group key value is unique.
...
 
-[IMPL#300](https://github.com/influxdata/platform/issues/300) Design how nulls behave
+And finally, because records are represented using object types, attempting to access a column whose value is unknown or missing from a record will also return _null_.
```
jsternberg: I'd like to talk about this one a bit more as cited above.

docs/SPEC.md
jlapacik: I don't believe this solution conflates the two solutions you mention @jsternberg . This solution is basically the SQL method which I find the most intuitive. I find it incredibly easy and natural to think of nulls as unknown values. In that context, operating on an unknown (non-boolean) value will always produce something that is unknown.

The only case where this is not true is in the case of boolean logic. But I still find the above truth table very intuitive.

jsternberg: I think I can live with that since the result for most of these ends up being the same. We also had a conversation in Slack where `null == null` needs to return false for SQL interop and I think that makes sense because `5 + null == 3 + null` should probably be false and it would be false under this proposal.

The only one I do want to talk about is `not null`. It seems very unintuitive to me for `not null` to be `null` which would evaluate as false. Is this another sql-ism that will just have to exist?

wolffcm: SQL's `null` is not quite the same as `NaN`.  If you do `count(floatCol)` in SQL, you will get a count of the non-null values in the column, including the NaNs.  Moreover, in SQL, a predicate of the form `foo = 3.0` where `foo` is `null` will return `unknown` (boolean null).  On the other hand if `foo` is `NaN`, then the predicate will return `false`.

jlapacik: @jsternberg yes it is. However I always go back to thinking about nulls as unknown. If you don't know if `x` is _true_ or _false_, then you also don't know if the negation of `x` is _true_ or _false_. And so if you want to compare things against `null`. You just have to use the built-in `exists()` function. Or maybe it should just be called `isNull()`.

Either way we don't have to describe a bunch of edge cases in the SPEC. It's a very concise description of how nulls behave, so that for those who find it unintuitive at first, they can easily reason about it because everything is consistent. There are no weird unexpected edge cases.

At least that's my opinion.

jsternberg: We talked about this in Slack. If you create a new issue to discuss how the null behavior should work on comparisons, I'll approve the current proposal and we can reevaluate the correct behavior after implementing one of the two by seeing it work in practice.

docs/SPEC.md
```diff
@@ -721,10 +760,14 @@ Index expressions access a value from an array based on a numeric index.
 #### Member expressions
 
 Member expressions access a property of an object.
+They are specified via an expression of the form `obj.k` or `obj["k"]`.
 The property being accessed must be either an identifer or a string literal.
 In either case the literal value is the name of the property being accessed, the identifer is not evaluated.
 It is not possible to access an object's property using an arbitrary expression.
 
+If `obj` contains an entry with property `k`, both `obj.k` and `obj["k"]` return the value associated with `k`.
+If `obj` does **not** contain an entry with property `k`, both `obj.k` and `obj["k"]` return _null_.
```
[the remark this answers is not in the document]
jlapacik: Yes that is what I am proposing. However compile time errors on object literals is something that is easy to reintroduce if we feel that it is beneficial. The reason for doing it this way is that we do not have a separate record type right now, just an object type, and so I think object types should exhibit the same behavior whether they are specified as literals or if they are retrieved from tables.

docs/SPEC.md
```diff
@@ -436,39 +468,46 @@ Examples:
...
+The string type is also nullable.
 
 The length of a string is its size in bytes, not the number of characters, since a single character may be multiple bytes.
 
-#### Regular expression types
+##### Regular expression types
 
 A _regular expression type_ represents the set of all patterns for regular expressions.
 The regular expression type name is `regexp`.
+The regular expression type is also nullable.
```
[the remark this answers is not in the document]
jlapacik: This is incorrect and I will change this. Regex types should not be nullable. Regex match operators will be able to return nulls however.

PR #1319, opened 2019-06-02
fix(changelog): changelog generator now handles merge commits better

### LINKED ISSUE #1304: Changelog generator does not work with the release workflow
The changelog generator has a problem with the current release workflow.

```
* f18ea549 (HEAD -> master, tag: v0.31.0, origin/master, origin/maint, origin/HEAD) fix(execute): copy the table when a table is used multiple times (#1301)
* 979cf176 refactor: change the table interface (#1300)
* 3c742a58 (refactor/sql) test(stdlib): skip tests while filter produces empty tables
* 025e7f63 test(stdlib):chronograf aggregate window flux tests
* 4c61c181 test(influxdb/v1): chronograf meta query flux tests
* 3204dd2d ci: add codecov.io support to flux (#1273)
*   f0b68507 Merge branch 'maint'
|\
| * ec371174 (tag: v0.30.0) feat(universe): support for dynamic queries (#1284)
* | 2cce6fe4 feat(universe): support for dynamic queries (#1284)
|/
• 32fb2fb7 (tag: v0.29.0) fix(execute): properly use RefCount to reference count tables (#1216)
```

The changelog currently defaults to using a depth first search when processing log entries. This works fine for a continuous chain, but in the above, it checked the entire history before finally checking the second branch.

This means that it generates the wrong changelog. But the solution is a bit more complicated than that as seen by the above. If a breadth-first search was used, it could still find changes that are unrelated. If the parents had been reversed, then the change `2cce6fe4` would have potentially been missed since the version 0.30.0 branch would have been checked first and iteration would have been finished.

It seems to me like we need to have the changelog generator walk parents and finish at either the common parent between the previous revision and the current one and have it finish at the previous revision in order to get all new commits.

jsternberg:
- [ ] docs/SPEC.md updated
- [ ] Test cases written

When the git log contained merges, the changelog generator didn't
necessarily work correctly. First, it would attempt to parse the message
in merge commits which was always wrong and always printed an error
message. Now, the changelog generator is told to just skip those commits
and move on since merge commits won't contain the message we're looking
for anyway.

The second happened with patch releases and feature branches that were
under work for more than a release. If a release was performed on the
maint branch and another commit was done on master before merging
maint back to master, the tree could be constructed in a way where the
previous version commit could not be clearly found such as in this
example from the `v0.31.0` release.

    * 3204dd2d ci: add codecov.io support to flux (#1273)
    *   f0b68507 Merge branch 'maint'
    |\
    | * ec371174 (tag: v0.30.0) feat(universe): support for dynamic queries (#1284)
    * | 2cce6fe4 feat(universe): support for dynamic queries (#1284)
    |/
    * 32fb2fb7 (tag: v0.29.0) fix(execute): properly use RefCount to reference count tables (#1216)

If the changelog generator encountered this pattern, it was undefined
exactly how it would behave. In this specific scenario, it parsed every
commit on the left branch going back to the root commit and then found
the previous commit on the second side of the branch. The opposite could
happen too and then it would have missed adding the commit from the
other branch.

The changelog generator has now been modified to find the merge base
before it attempts to read the log to prevent searching back through the
entire history. In the above situations, `v0.29.0` would be found and
act as a final constraint for parent commits.

The changelog parser has also been modified to use a breadth-first
search instead of depth-first search and it does not use the `Log()`
method from the go-git library. This is because the go-git library does
not have a way to stop at a specific commit and stop looking at parent
commits. We now limit the breadth-first search of commits to either the
merge base or the previous version commit which should be the same
commit if the release process was followed correctly. A breadth-first
search is used to ensure we get all commit messages from any merges and
don't skip any commits because we reached the terminating commit first.

Fixes #1304.

### DIFF
--- a/internal/cmd/changelog/cmd/generate.go
+++ b/internal/cmd/changelog/cmd/generate.go
@@ -147,34 +147,139 @@ func findPreviousRelease(r *git.Repository, v *semver.Version) (*semver.Version,
 
 // findNewCommits queries the git repository for new commits since the last tagged release.
 func findNewCommits(r *git.Repository, prevHash, currHash plumbing.Hash) ([]Commit, error) {
-	logs, err := r.Log(&git.LogOptions{
+	// Find a common merge base which will tell us all of the commits that have been
+	// merged in since the previous release that may not have been branched off of the
+	// previous release. Usually, the result of this will be the same as prevHash.
+	mergeBase, err := findMergeBase(r, prevHash, currHash)
+	if err != nil {
+		return nil, err
+	}
+
+	// Keep a queue of which commits we will process and which commits we have already
+	// seen. If we have processed a commit before, we do not want to process it again.
+	queue := []plumbing.Hash{currHash}
+	seen := make(map[plumbing.Hash]bool)
+
+	var commits []Commit
+	for len(queue) > 0 {
+		h := queue[0]
+		queue = queue[1:]
+
+		if seen[h] {
+			continue
+		}
+		seen[h] = true
+
+		if h == prevHash || h == mergeBase {
+			// Do not process this commit or its
+			// parents. We are done processing commits on this
+			// branch.
+			continue
+		}
+
+		commit, err := r.CommitObject(h)
+		if err != nil {
+			return nil, err
+		}
+
+		// Ignore merge commits when parsing commit messages.
+		if len(commit.ParentHashes) <= 1 {
+			c, err := parseCommit(commit)
+			if err == nil {
+				commits = append(commits, c)
+			} else {
+				log.Printf("invalid commit message %v: %v", commit.Hash, err)
+			}
+		}
+
+		// Iterate over the parents and add them to the queue.
+		queue = append(queue, commit.ParentHashes...)
+	}
+	return commits, nil
+}
+
+// findMergeBase finds the common parent commit the two commits have.
+//
+//
+func findMergeBase(r *git.Repository, prevHash, currHash plumbing.Hash) (plumbing.Hash, error) {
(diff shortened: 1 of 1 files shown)


internal/cmd/changelog/cmd/generate.go
```diff
@@ -147,34 +147,118 @@ func findPreviousRelease(r *git.Repository, v *semver.Version) (*semver.Version,
...
+			log    object.CommitIter
+		)
+
+		if prevCommit == nil {
+			commit, log = &currCommit, currLog
+		} else if currCommit == nil {
+			commit, log = &prevCommit, prevLog
 		} else {
-			log.Printf("invalid commit message %v: %v", commit.Hash, err)
+			if currCommit.Committer.When.Before(prevCommit.Committer.When) {
```
wolffcm: Is this `When` field that actual time that the commit was added relative to other commits?

I have seen dates in the git log that are definitely from way before the commit was added, which must be due to `git commit --amend` or force push.

jsternberg: It is. Git is... not very smart about this. It's what the actual `merge-base` command does. Here's the code that sets the comparison:

https://github.com/git/git/blob/master/commit-reach.c#L38

The generation that's referenced comes from a commit graph that isn't written to disk unless you explicitly write it to disk so it's almost always empty and isn't exposed by the go-git library.

The generation number would be much better and exactly what would be needed, but it doesn't exist for practical purposes. Git code just seems to use commit dates as a way of trying to reduce the amount of time for practical purposes, but it's obviously not very reliable.

jsternberg:
I updated the code with comments to explain what I was doing.

PR #1332, opened 2019-06-05
docs(overview): new vision for compilation and execution of queries

### LINKED ISSUE #1321: Create Design Doc for Interpreter as Executor
This should be a detailed design doc with explicit Go interfaces on how the interpreter + compiler + executor types play together. 

The end goal is to have a framework capable of executing dynamic queries while having decoupled components.

affo:
Closes #1321 

I updated the old `Overview.md` because it looked obsolete.

I added some bits from it to this new document.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stormasm:
This overview is very nice and a great start to developers having a better understanding to the internals and details of Flux.  Thank you very much for updating this document.  My only request would be to have links in this document back to the underlying source code in Flux where the interfaces are defined in the code.  I know they might not be exactly the same --- but it would be great to be able to tie the "theory" of what you are trying to present to the real world which is the Flux code.  Without the links, it is hard for a developer to connect the dots.  Just my 2 cents.
Thanks,
Michael

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+A language is used to describe a query. It consists of words whose letters are taken from an alphabet.
+Letters can be of any type and complexity, from characters, to nodes in an Abstract Syntax Tree (AST). Various languages
+can be used to represent a query.
 
 ```go
-
-type Parser interface {
-    Parse(flux string) Query
+type Language interface {
+    Name() string
```
jsternberg: Name is very generic and it makes it so a lot of things can accidentally implement this interface. Can we change this to `LanguageName() string` to avoid accidental implementations?

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+type CreateTransformation = fn (ExecutionContext) Transformation
+type CompilerType interface {
+    From() Language
+    To() Language
+}
 
-Both a query and a plan are represented by a DAG and describe an operation that needs to be performed.
-The difference is that a plan in addition to describing what the operation is, also describes how that operation will be performed.
-In short, a query describes what the operation is and a plan describes how that operation will be carried out.
+type Engine interface {
```
jsternberg: Every interface in here needs godoc comments.

I am going to wait to comment on the individual methods until there are the godoc comments. I would also prefer if this were reorganized so that the types like `Representation` were put higher up so that I could read it in the order of usage. It will make it easier to comment.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+```go
 type Query interface {
-    Nodes() []QueryNode
-    Edges() []Edge
+    Results() Results
+    Stats() map[string]interface{}
+    Representation() Representation
+}
+
+type Representation interface {
```
jsternberg: I think this implementation could be simplified a lot. It's unclear to me why it needs to be a `Visitor` as that seems to be very dependent on the internal structure and related to the compiler itself. I can't find a definition of the `Visitor` interface with a quick glance though.

I also think that `Representation` and `Language` are essentially the same thing and the name is very vague. I think `Language` should be the struct that defines the language content and we should just use strings for the language names. I don't see how an extra `Language` interface adds much.

I also do not think we should add a `String()` method to the representation. When I think of `String()` I think of a small thing I can use with logging and printing. This is guaranteed to not be the case with the representation and I'd like it if we played cautiously and did not include this at first. I think there's a reasonable belief that we will need it eventually, but I'd like to add something rather than remove it at the current moment.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+execution context that embeds everything that is necessary for the execution of the query, included the registered values,
+sources and transformations. 
 
-// Edge establishes a parent child relationship between two nodes.
-type Edge interface {
-    Parent() NodeID
-    Child() NodeID
+```go
+type Executor interface {
+    Execute(Query, ExecutionContext) error // fills query results.
```
jsternberg: This interface seems weird to me and it seems more like an implementation rather than a usable interface.

A first problem is that it's unclear how it can use the `Query` interface. There is no way in the public query interface to send results, but only iterate. That would force the interface to use an internal implementation by a typecast to get access at the write components and that will likely result in a bad interface.

I'd like to propose a separate interface that I wrote in some notes when we were discussing this. I had different interface names so I've translated them to their equivalent in this document.

```go
type Executor interface {
    Execute(ExecutionContext, Representation) (Query, error)
    Type() Language
}
```

I did not have the `Type` method in my original proposal, but I like it so I would say to keep it.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+    MemoryAllocator() Allocator
+    Logger() Logger
 }
 
-// Plan is a DAG of the specific steps to execute.
-type Plan interface  {
-    Nodes() []PlanNode
-    Edges() []Edge
+// Scope is a set of bindings from identifiers to values.
+type Scope = map[string]Value
```
jsternberg: I'd like to see `Scope` as an interface so nested scopes do not have to copy the entire map.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+```
 
-// PlanNode is a single step in the plan DAG.
-type PlanNode interface {
-    ID() NodeID
-    Predicates() []Predicate
+As an example, let `e` be an engine that supports `String`, `AST`, and `SemanticGraph` queries:
+
+```go
+func (e *Engine) Query(r Representation) (Query, error) {
```
jsternberg: I think it would be useful to also include the compilers in here too since they transform representations to other representations. I think it would be helpful to add a mechanism to the engine to register compilers and executors.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
 }
+```
 
-// Predicate filters data.
-type Predicate interface {}
+The `StringExecutor` does not actually know how to execute a script, but it uses a compiler to translate the script into its
+AST representation and delegate to an `ASTExecutor` in turn:
+
+```go
+func (e *StringExecutor) Execute(q Query, ec ExecutionContext) error {
```
jsternberg: I think this works badly because it tightly couples the compiler and the executor. I think separating them makes it possible for us to translate between different representations and execute them separately. If we do this, then we tightly couple them and there is only one way to execute an AST. This makes it impossible to do things like proxies.

affo:
@stormasm Thank you for the comment and for reading the document!

This is still a design on how we want to change the internals of the engine. Once we really implement its content, we can link to the actual implementation. Thanks for the advice :)

docs/Overview.md
affo: I understand your point. I hadn't put godoc because every interface is introduced with its description on top. Anyways, I'll add it to the snippets.

The order in which the types are introduced is for better interweaving with the explanation outside of the snippet. If I change the order of the interfaces I would need to turn the explanation the other way around. It's something I'll do if the flow of the explanation is not clear as it is.

docs/Overview.md
affo: Agreed on removing `Visitor` (it was there only for explicitly stating that the Representation can be consulted, but you are right, it makes no sense 😅 ).

Agreed on removing `String`.

I think that `Language` and `Representation` are pretty different concept and should stay separated, they can be considered as different as "type" and "object". Indeed `Language` is used for the types of compilers and executors.

docs/Overview.md
affo: Absolutely need to add those method to the `Query` interface. Thank you.

I understand your point for `Execute` to accept a `Representation` instead of a `Query`. I took that decision to make executors delegate each other. We should further discuss this point. I see you have a comment on that below.

docs/Overview.md
affo: Yep, understood. I put `map[string]interface` only for a simplistic definition of `Scope`. Actually, that would be `interpreter.Scope`.

docs/Overview.md
affo: At first I did it, but, by making executors compile and delegate to others, it seemed useless because they would always need access to the engine for getting a compiler. We'll discuss on this.

docs/Overview.md
affo: We absolutely need to talk more about this. I am not against my design (obviously) neither against yours. Indeed, I know we talked about compilation and execution as 2 separated phases.

However, when designing  this, I understood that I couldn't force an executor not to compile to another representation and delegate to another executor. I think that I need to talk to better understand this:

> If we do this, then we tightly couple them and there is only one way to execute an AST.

The `ASTExecutor` knows the path to reach another executor which it can delegate execution. I understand that they are coupled now, but, how can you deal with having multiple executors for the same language? How should the engine decide on which to use?

An alternative approach could be:
 - register compilers and executors through `Create*` functions.
 - the engine would be the only one able to instantiate executors and compilers.
 - it compiles until it finds an executable representation (if it doesn't find one, it could proxy to another engine that has it).
 - the executor cannot instantiate compilers neither other executors, the only thing it can do is to execute.

I don't know if this is possible:

> the executor cannot instantiate compilers neither other executors

Or should be enforced as a good practice.

docs/Overview.md
affo: No ok, I was wrong. I agree with changing to `Execute` take a `Representation`.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+    From() Language
+    To() Language
+}
 
-Both a query and a plan are represented by a DAG and describe an operation that needs to be performed.
-The difference is that a plan in addition to describing what the operation is, also describes how that operation will be performed.
-In short, a query describes what the operation is and a plan describes how that operation will be carried out.
+type Engine interface {
+    BindValue(id string, Value)
+    Prelude() Scope
```
jlapacik: I don't think these methods should be part of the `Engine` interface. I'm actually not sure this should even be an interface. The engine is just an executor that executes a representation in the form of a physical query plan. In essence, it is the target platform for a compiler that produces an optimized query plan. Variable context and binding are not relevant concepts to the engine, just to the compiler that targets the engine (in my opinion).

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+    // ... methods to iterate other the contents of the result
 }
+```
 
-// NodeID uniquely identifies a node.
-type NodeID
+The executor executes a query expressed in some language and fills its results, provided an execution context.
+There can be multiple executors, one for each language supported by the engine.
+The engine is responsible for providing the executor with an
+execution context that embeds everything that is necessary for the execution of the query, included the registered values,
```
jlapacik: With all due respect, I would like to disagree. I believe the engine is a concrete instance of an executor that takes as input a very specific representation that is the physical query plan.

docs/Overview.md
jlapacik: Unfortunately I think I disagree with both. I don't believe an executor should take an `ExecutionContext` as input. Why should it? It is an executor after all. Shouldn't it have all the context that it needs? I had envisioned a much simpler interface like so:
```go
type Executor interface {
    Execute(ir Representation) (Query, error)
}
```

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+// The ExecutionContext contains any information necessary for properly executing a query.
+type ExecutionContext interface {
+    Bindings() Scope
+    GetSource(id string) Source
+    GetTransformation(id string) Transformation
+    
+    // ... examples of other components of the execution context.
+    Context() context.Context
+    MemoryAllocator() Allocator
+    Logger() Logger
```
jlapacik: As I mentioned above, I think this interface is unnecessary. An executor should have all of the context it needs. It should not need any extra, in my opinion.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
+        return err
+    }
+    // ... add results the query (see below).
+    return nil
+}
+```
+
+While evaluating, the interpreter updates the scope of execution by visiting the semantic graph. It performs
+additions, subtractions, comparisons, etc., variable assignments, imports, and data manipulation functions. 
+The interpreter can further compile the semantic graph to an IR and to an
```
jlapacik: Just to drive the point home. The interpreter is an executor. And not just any executor, it is the Flux Virtual Machine. It executes a typed semantic graph and produces a query result. Because it executes queries it must embed a query engine. And because the query engine executes query plans, the interpreter must also embed a compiler that targets the query engine. That's how I see it at least.

affo:
@jlapacik I see your point and I have no points against it. The Engine was made to be an instance against which one can register values, transformations, source, etc. in order to reduce the number of global variables we have around. We can rename it to something else, such as `Runtime` (idk, please suggest). I like your view for executors and it seems to me that you agree with how executors and compilers work together ---i.e. an executor can compile and delegate.

Correct me if I am wrong: it seems to me that, in your vision, what you call `Engine` is our current `Executor`, and so, something that is able to run queries. The interpreter is able to run everything, both queries and "standard" instructions, but it delegates queries to the `Engine`. The modality  in which this happens coincides with my design, and so, I like it 😄 .

I also like your point to avoid the execution context being a parameter to pass in, but to be an attribute. This would mean that, every time that an executor delegates, it creates a new executor passing in its execution context. Cool.

I think that once everybody has reviewed this PR. We should plan a meeting to discuss reviews, so that I can change this doc and we can start decomposing the implementation in well-encapsulated issues.

aanthony1243:
I read the whole doc, and perhaps I don't have enough experience with the current engine to have a strong opinion.   I was hoping this document would make it very obvious how we might improve on runtime-compiled functions to include local scope, and to also enable those functions to (safely) modify the local scope.

If I can speculate, is the thinking that a runtime function may be executed by an executor that imports the scope from the calling executor?   Or could we ask the existing executor to execute new code?  I think it's a cool idea that the executor has a "stable" state where it has all of its scope defined, and is waiting for new pipelines/calls to execute.

affo:
@aanthony1243 this document doesn't touch runtime compilation of row functions. However, thank  you so much for mentioning it, if we need to touch the `values.Function` interface, we would need to take that into account too.

jlapacik:
@affo just to make a small correction, what I am calling engine is not what you are calling `Executor`. At least I don't think so. I am saying that the query engine (of which there is only one) should implement the `Executor` interface. That is, the query engine is an executor. But the query engine can only execute query plans. That's the only "language" that it understands.

The interpreter should also implement the `Executor` interface. But the interpreter should only execute semantic graphs. That's the only "language" that it understands. Internally as an implementation detail, it embeds a query engine and a compiler that produces optimized physical query plans, because any path that produces query results must go through the query engine.

In my mind there really only needs to be two executors - the query engine and the interpreter. And then we need compilers that target the languages that these executors understand.

affo:
@jlapacik I like that too.

That's another point @jsternberg and I where discussing.

Should it be better to clearly separate compilation and execution? So that you compile until you get something executable and then only execute.

This way you end up having multiple compilers from script --> semantic graph and 1 executor: the interpreter. But, at that point, the interpreter has to translate the semantic graph into IR (or optimized IR) to delegate execution to something that understands query plans. That's why I was thinking of that  generalized approach.

jsternberg:
I like that and I think it ends up making it questionable whether or not a `flux.Engine` type makes sense. It seems to me that if we split the compilers and executors we get a more robust infrastructure and, in addition, what is currently being called the engine becomes the executor. The executor would keep all of that global state of which functions are registered. The transformations and function implementations that are registered with the `flux.RegisterPackageValue` and various other functions are _very_ specific to the spec implementation so it makes sense they would not be an interface and instead be a direct component of the execution implementation.

I still like the `ExecutionContext` though and I think we should keep that. In this proposal, there's `Executor`, `ExecutionContext`, and `Engine`. We can probably just drop `Engine` and, in return, we also drop a lot of circular dependencies in the `flux` package because all of the registration functions would be part of the executor implementations instead. I like the name runtime. The user would instantiate a runtime and the runtime implements the `Executor` interface. The stdlib packages have a function that takes a runtime and registers their implementations to the runtime so they can be executed.

docs/Overview.md
```diff
@@ -1,96 +1,246 @@
...
-// Query defines work to be performed on time series data and a result.
-// A query is represented as a directed acyclic graph (DAG).
+A query defines a unit of work for the engine. It comprises one or more pipelines.
+It contains its representation, its results, and execution statistics.
+
+```go
 type Query interface {
-    Nodes() []QueryNode
-    Edges() []Edge
+    Results() Results
```
wolffcm: The current `Query` interface produces a `chan Results` so that the caller can block until the results are actually ready.  This way the call to `Engine.Query()` can return immediately upon successful compilation.
Was it your intent to change this behavior. or just something you've omitted for simplicity?

affo: Nope, it was just omitted for simplicity. I put `Results` only to represent them somehow. This doc purpose is only to change how compilers and executors interact, nothing else 😄 .

affo:
@nathanielc I see your point.

The idea is that, once the interpreter becomes an executor, it has both the concrete implementations of transformations (passed by the `Engine`) and the context needed to execute a particular query, such as memory allocator, passed by call.

At that point it has everything it needs to run the query properly.

The interface for `Executor` states explicitly that every executor knows the particular `ExecutionContext` for a query. The `Engine` is a new way for registering values/transformations. InfluxDB can replace the implementation for a source with its one when instantiating the `Engine`. No need to pass dependencies around now that one can instantiate a `storage.Reader`, create the `FromSource` that uses it, and register that source in the new `Engine`. Before, one had to specify the function to create the source on `init`. This precluded the possibility to instantiate a `Reader` and use it in the source. Does this sound ok?

Anyways, at the moment I am refactoring this to a new design that addresses the comments. Some pieces will change, but the idea to make the Interpreter work is still the same in my mind.

To address your comment (@nathanielc) I will also add an explicit example on how InfluxDB can replace the `from` source. Does it sound good?

nathanielc:
@affo I don't quite follow, the `Engine` is a lot more than just a collection of transformation implementations. There is a lot of logic in how the tables flow between transformations. Are you proposing that logic is moved inside the interpreter? 

Also I don't understand how changing how transformations are registered solves the problems at hand. To me the issue is solely focused around how to model the behavior of the interpreter correctly. If making clear interfaces for compilers and executors achieves that then great, but there are likely other ways to solve this. 

 I'd like to focus the solution more narrowly on how to resolve the issues with the interpreter, things like changing the registration pattern do not seem related or necessary at this time.

affo:
@nathanielc got it, thank you 👍 

You are right, I lost the focus of the issue and tried to put more things in derived from brainstorming sessions with the team. I will narrow down to the scope of interpreter and compilers.

affo:
@nathanielc @jsternberg @aanthony1243 @jlapacik @wolffcm 

Thank you for the feedback!

I preferred to create a new file, and we can care about docs consistency later.

This is scoped to interpreters and executors only.

docs/CompilationExecution.md
```diff
@@ -0,0 +1,121 @@
...
+These are the interfaces:
+
+```go
+// Language is a language used to express a Flux query.
+type Language interface {
+    LanguageName() String
+}
+
+// Representation contains the actual content of a Flux query expressed in some Language.
+type Representation interface {
```
jsternberg: Can we find a better name for this? I understand what it is, but the name does not seem like it will be good in the code. It might be long, but we may need to make this `LanguageRepresentation` to make it clear exactly what this is.

docs/CompilationExecution.md
```diff
@@ -0,0 +1,121 @@
...
+ So, it takes their representation ---i.e., the `Spec`-- and optimizes it by applying subsequent _transformations_,
+ in order to delegates execution to the `Engine`.
+ Those transformations are logical and physical planning. We consider these transformations as further compilation
+ steps, in that they turn a query representation into another.
+ - Finally, the `Engine` executes the optimized representation and outputs the results.
+
+Wrapping up, the actors that play in this process are `Compiler`s ---that turn a representation in some language into another one---
+and `Executor`s ---that execute a representation and produce results.
+Both the `Interpreter` and the `Engine` are `Executor`s:
+ - The `Engine` actually executes data manipulation pipelines and produces results;
```
jsternberg: From a previous comment from someone else, I like runtime as a better name. I do like that the engine/runtime do not have an interface description. To my mind, when we register functions, we'll register AST to op spec transformations with the interpreter and the engine/runtime will be an abstract separation.

I think this will also make things like the repl possible and use the same proxy behavior we have currently. So the repl will know how to transform AST to op spec and then will use an interface to send that spec to a server. We'll expose as a repl endpoint some version constraint API and generally require the repl to be the same version as the server to avoid errors similar to how postgres' client has requirements for the versioning.

docs/CompilationExecution.md
affo: It could be `QueryRepresentation`.

docs/CompilationExecution.md
affo: Yeah, that was me actually 😅 

Would you like to change `Engine` into `Runtime` 🤔 ?

To me `Runtime` sounds more passive than `Engine` which sounds like something that does things. I don't know, only matter of taste. I am not a native speaker too 😅 

I proposed `Runtime` for the object that gets things registered and provides them to `Executors`.

docs/CompilationExecution.md
```diff
@@ -0,0 +1,121 @@
...
+// Statistics are the statistics of a Query execution.
+type Statistics interface {}
+
+// Query is an executed Flux query.
+// It provides its representation, results, error, and statistics.
+type Query interface {
+    Results() Results
+    Error() error
+    Stats() Statistics
+    Representation() Representation
```
nathanielc: Why does a query need to know its representation?

docs/CompilationExecution.md
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// The ExecutionContext contains the information necessary for properly executing a query.
+type ExecutionContext interface { 
+    // ... examples of components for the execution context.
+    Context() context.Context
```
nathanielc: What is this context for? Is this just the context of the query request to listen for cancel/timeout events? Or is there more to it?

docs/CompilationExecution.md
affo: I thought about that for debugging. If you want you can turn a `Representation` into a string (somehow) and log it. I can just remove that and thought about it later.

docs/CompilationExecution.md
affo: This is intended to be Go's `context.Context`. Should I be more explicit about that?

affo:
@nathanielc 

> How does this solve the interpreter being both an executor and a compiler? My understanding is that the issue with the interpreter being both is it was not clear how to manage dependencies. Meaning the interpreter didn't now how to get access to the Engine executor needed in order to leverage it.

The short answer is that it is somewhat implicit in the "delegation pattern": the `Interpreter` holds an instance of an `Engine` in order to make delegation possible. The `Engine` contains the dependencies to execute the data manipulation. 

Either the `Engine` is passed to the `Interpreter` upon creation, or the `Interpreter` is given the dependencies and it can instantiate an `Engine` accordingly.

I left this unspecified, because I think that both are valid alternatives and we can decide when we implement the concrete implementations.

The core idea is that, if the `Interpreter` holds an instance of an `Engine`, it can pass that instance to `tableFind` to trigger execution, or take  the result of `tableFind` (maybe that could return a function value that consumes an `Engine` upon `Call`) and delegate the execution to the `Engine`.

affo:
Last changes:
 - `svg` image;
 - `Representation  -> QueryRepresentation`
 - sentences  on one line

docs/CompilationExecution.md
```diff
@@ -0,0 +1,115 @@
...
+    MemoryAllocator() Allocator // (Allocator definition is not relevant here)
+    Logger() Logger // (Logger definition is not relevant here)
+}
+
+// Executor executes a query QueryRepresentation and returns a Query given an ExecutionContext.
+type Executor interface {
+    Execute(ExecutionContext, QueryRepresentation) (Query, error)
+    // each Executor can execute query representations in a target Language.
+    ExecutorType() Language
```
jlapacik: Just a nitpick but this method doesn't seem like a necessary part of the interface. I would remove it.

docs/CompilationExecution.md
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Compiler turns a QueryRepresentation in a Language into another one.
+type Compiler interface {
+    Compile(QueryRepresentation) (QueryRepresentation, error)
+    // each Compiler has a source and a target Language.
+    CompilerType() (Language, Language)
```
jlapacik: Again another nitpick. This method doesn't seem that useful. I would remove it from the interface.

docs/CompilationExecution.md
```diff
@@ -0,0 +1,115 @@
...
+import "context"
+
+// Language is a language used to express a Flux query.
+type Language interface {
+    LanguageName() String
+}
+
+// QueryRepresentation contains the actual content of a Flux query expressed in some Language.
+type QueryRepresentation interface {
+    Lang() Language
```
jlapacik: Another nitpick. I would remove the Language interface and just have `Lang()` return a string which is the name of the query representation.

docs/CompilationExecution.md
affo: I don't know if I agree 😅 I think we can decide upon implementation.

docs/CompilationExecution.md
affo: I don't know if I agree 😅 I think we can decide upon implementation.

docs/CompilationExecution.md
affo: You are right, I'll make `type Language string`.

affo:
> I see that you cover using functions like `tableFind` which must evaluate a pipeline and return the result back to the interpreter. This is nice!
> 
> There is also another special case, where we must evaluate a lambda in a stream function like `filter` or `map`, and we do not know the schema until we actually have the tables. I guess this implies that we must compile and interpret the lambdas from the `Engine`, during execution. Do I have that right?
> 
> If I recall, one of the challenges of handling this case was that it created a circular dependency. I think if the top-level interfaces `QueryRepresentation`, `Compiler` and `Executor` are in the top-level `flux` package, if each concrete representation/compiler/executor lives in its own package, maybe you can get around the circular dependency issue. Each compiler or executor can talk to interfaces instead of a concrete implementation.
> 
> Maybe you should add some language discussing this; is that in the scope of this document?

@wolffcm I see your point. 
I think that that case could be embedded in the new implementation by making the `Engine` aware of the `Interpreter`. In this way, it can use it to interpret lambdas. This should be cool, because the lambdas would have the scoping needed (imports, etc.), and this would solve a number of issues.

I'll try to cover this point in the new commit.

affo:
@wolffcm and @jlapacik I addressed  your points in the last commit.

If @nathanielc can give another look it would be great :)

PR #1339, opened 2019-06-06
feat(stdlib): added MySQL type support

roshie548:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/rowreader.go
+++ b/execute/rowreader.go
@@ -1,0 +1,17 @@
+package execute
+
+import (
+	"database/sql"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/values"
+)
+
+type RowReader interface {
+	Next() bool
+	GetNextRow() ([]values.Value, error)
+	InitColumnNames([]string)
+	InitColumnTypes([]*sql.ColumnType)
+	ColumnNames() []string
+	ColumnTypes() []flux.ColType
+	SetColumns([]interface{})
+}
--- a/stdlib/sql/from.go
+++ b/stdlib/sql/from.go
@@ -3,20 +3,18 @@ package sql
 import (
 	"database/sql"
 	"fmt"
-
-	"reflect"
-	"time"
-
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
-	"github.com/influxdata/flux/values"
 	_ "github.com/lib/pq"
 )
 
 const FromSQLKind = "fromSQL"
+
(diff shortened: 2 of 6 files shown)


stdlib/sql/from.go
```diff
@@ -110,7 +114,8 @@ func createFromSQLSource(prSpec plan.ProcedureSpec, dsid execute.DatasetID, a ex
 		return nil, fmt.Errorf("invalid spec type %T", prSpec)
 	}
 
-	if spec.DriverName != "postgres" && spec.DriverName != "mysql" {
+	// Alow for "sqlmock" for testing purposes in "sql_test.go"
```
jsternberg: There's a typo in this comment.

jsternberg:
It looks like there's no test coverage for any of the new mysql stuff, but I thought I saw added tests. Can you check on that?

stdlib/sql/sql_test.go
```diff
@@ -0,0 +1,355 @@
...
+				Parameters: map[string]semantic.PolyType{
+					"driverName":     semantic.String,
+					"dataSourceName": semantic.String,
+					"query":          semantic.String,
+				},
+				Required: semantic.LabelSet{"driverName", "dataSourceName", "query"},
+				Return:   flux.TableObjectType,
+			}
+
+			fluxFn := flux.FunctionValue(FromSQLKind, createFromSQLOpSpec, fromSQLSignature).Function()
```
aanthony1243: instead of calling the function from this level,  let's instantiate a  SQLIterator type directly, call `connect` `fetch` and `decode` ,  then compare the table returned by `decode` to the expected result.

execute/rowreader.go
```diff
@@ -0,0 +1,227 @@
...
+	columns     []interface{}
+	columnTypes []flux.ColType
+	columnNames []string
+}
+
+func (m *RowReaderImpl) SetColumns(i []interface{}) {
+	m.columns = i
+}
+
+type MySQLRowReader struct {
```
aanthony1243: i apologize for all the bouncing around and code re-org i'm having you do.  let's put the MySQLRowReader  and PostgresRowReader   back in stdlib/sql  and import the rest of this code from there.   All the code looks good, just need to organize it.

stdlib/sql/from.go
```diff
@@ -244,3 +196,40 @@ func (c *SQLIterator) Decode() (flux.Table, error) {
 func (c *SQLIterator) Close() error {
 	return c.db.Close()
 }
+
+func NewMySQLRowReader(r *sql.Rows) (execute.RowReader, error) {
```
aanthony1243: when you move the above code from execute back to this package,   create 2 files  `mysql.go`  and `postgres.go`  and put these helper functions inside of there.

PR #1341, opened 2019-06-07
fix(lang): return compile time errors before program start

### LINKED ISSUE #1292: Compile time Flux errors not reported until run time
AstProgram does not compile nor plan its Flux query until Start is invoked:
https://github.com/influxdata/flux/blob/master/lang/compiler.go#L291-L311

This means that any errors during semantic analysis or planning will not be received until after the query is queued, dequeued and execution begins, here:
https://github.com/influxdata/flux/blob/master/control/controller.go#L283

As a result, errors that could be returned synchronously (possibly as 5xx errors over the http interface) are deferred and returned as 400s.

This code should be refactored so that we compile and plan the query before queueing it, so as to provide a prompt response and better insight into the issues with queries.  This means that compilation and planning should happen during the `AstCompiler.Compile()` step.

affo:
Close #1292 

I caught the moment to refactor how spans are used in the compile -> eval -> execute path.  
I also updated the test for checking errors in the different phases.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -72,6 +72,9 @@ func Eval(semPkg *semantic.Package, opts ...ScopeMutator) ([]interpreter.SideEff
 // Apart from the side effects and the scope used by the interpreter, it returns the actual now time used during the
 // evaluation (in the case that a now option overrode the now passed in).
 func EvalWithNow(semPkg *semantic.Package, now time.Time) ([]interpreter.SideEffect, interpreter.Scope, time.Time, error) {
+	if now.IsZero() {
+		now = time.Now()
+	}
 	sideEffects, scope, err := Eval(semPkg, SetNow(now))
 	if err != nil {
 		return nil, nil, time.Time{}, err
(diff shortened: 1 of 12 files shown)


affo:
@wolffcm I pushed a new commit to move type inference before `start`. What do you think?

jlapacik:
@affo is this urgent? It seems like this code will be refactored once #1332 is approved. I agree with @wolffcm that the semantic graph should be decorated with types and that the interpreter should not be type inferring a flux script. Unless this is urgent I think we should wait till #1332 is approved before we make these changes as that design will influence how we report compile time errors. And I think it will make it much simpler. What do you think?

affo:
@wolffcm @jlapacik Yep, I see your point. If you think that moving type inference before `Start` is too much, the I can switch back to the previous implementation and leave the test cases there. If you want me to, I can do a "partial revert" :)

I would prefer to merge it before the changes introduced by #1332 because I think that could happen in a long time (and this issue was put into the weekly plan).

The more we move errors before starting the query, the more observability we have on query errors.

nathanielc:
The implementation of this change depends on #1321, closing until we have a better understanding.

PR #1345, opened 2019-06-07
feat(flux): add exists unary operator

### LINKED ISSUE #1327: Implement exists
The SPEC states:

https://github.com/influxdata/flux/blob/15e15bd756e290d797de67afd90e44769db52404/docs/SPEC.md#L1366-L1370

This issue is to track the implementation of `exists`.

Note that the SPEC defines `exists` to be a function, however it could be implemented as a unary operator that returns `false` if its operand evaluates to _null_ and `true` otherwise.

I prefer the latter implementation as Flux does not support operator overloading and so incorporating null checks into query planning rules would not require variable context information as it would in the case of an `exists` function.

affo:
Closes #1327 

Still need to update `grammar.md` though

By having a glance at type inference, it seemed to me that I shouldn't update anything.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


internal/parser/grammar.md
```diff
@@ -38,7 +38,7 @@ The parser directly implements the following grammar.
     LogicalAndOperator             = "and" .
     UnaryLogicalExpression         = ComparisonExpression
                                    | UnaryLogicalOperator UnaryLogicalExpression .
-    UnaryLogicalOperator           = "not" .
+    UnaryLogicalOperator           = "not" | "exists" .
```
jlapacik: Not sure if exists should be considered a logical operator as it can operate on more than just booleans.

nathanielc:
@jlapacik and @affo can you make sure we are on the same page here https://github.com/influxdata/flux/issues/1327#issuecomment-500026609 before merging this PR?

internal/parser/grammar.md
affo: Yeah, I see your point. However, it always returns a boolean, so I would say it is used in logical expression 🤔 

I can change that, however, this grammar should exactly match the parser implementation. So, I should also update the parser accordingly.

jlapacik: I agree, let's not modify the parser too much with this addition. We'll keep it as you have specified here. I'm resolving this thread.

compiler/runtime.go
```diff
@@ -855,8 +856,16 @@ func (e *unaryEvaluator) EvalInt(scope Scope) (int64, error) {
...
 		return 0, err
 	}
-	// There is only one integer unary operator
-	return -v, nil
+	var sign int64
+	switch e.op {
+	case ast.AdditionOperator:
+		sign = 1
+	case ast.SubtractionOperator:
+		sign = -1
```
affo: this was a bug, added a test case for that

compiler/runtime.go
```diff
@@ -866,16 +875,29 @@ func (e *unaryEvaluator) EvalFloat(scope Scope) (float64, error) {
...
-	// There is only one float unary operator
-	return -v, nil
+	var sign float64
+	switch e.op {
+	case ast.AdditionOperator:
+		sign = 1.0
+	case ast.SubtractionOperator:
+		sign = -1.0
+	default:
+		return 0.0, fmt.Errorf("unknown unary operator: %s", e.op)
```
affo: same here

compiler/runtime.go
```diff
@@ -885,8 +907,16 @@ func (e *unaryEvaluator) EvalDuration(scope Scope) (values.Duration, error) {
...
 		return 0, err
 	}
-	// There is only one duration unary operator
-	return -v, nil
+	var sign int64
+	switch e.op {
+	case ast.AdditionOperator:
+		sign = 1
+	case ast.SubtractionOperator:
+		sign = -1
```
affo: same here

compiler/runtime.go
jlapacik: `+a` is valid according to the grammar?

jlapacik: Ah yes it is. Thanks for this.

PR #1346, opened 2019-06-07
feat(values): add nullable value types

jsternberg:
The `values` package has been modified to include a separate struct for
each type that has a `Value` and `Null` attribute so the typed
evaluators can use these instead of the values interface.

The inspiration for these types is taken from the `database/sql`
package.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/values/bool.go
+++ b/values/bool.go
@@ -1,0 +1,57 @@
+package values
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/influxdata/flux/semantic"
+)
+
+// Bool represents a float value.
+type Bool struct {
+	Value bool
+	Null  bool
+}
+
+func (Bool) Type() semantic.Type {
+	return semantic.Bool
+}
+
+func (Bool) PolyType() semantic.PolyType {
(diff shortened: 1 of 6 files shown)


jsternberg:
The intention was to make it possible for parts of the code that use real types like `float64` to instead use this. It would be a similar copy profile and wouldn't incur an allocation for the interface, but would allow null to be passed around.

This seems to mostly be relevant to the compiler. It seems like the binary functions do not use these types like I thought they did, but the `EvalFloat` and related functions in the compiler do use these and they don't have a way to recognize that a null value has happened without switching to using the values interface.

We can switch the compiler to just use a single `Eval` which would give it parity with the binary evaluation which already uses interfaces. I know we want to have things be typed if we can, but I'm not sure we're actually doing that.

jsternberg:
We changed all of the compilation functions to return `Eval` for now until we actually have a reason to add this. I'm going to close it for now as this was a good experiment, but it's not needed and not worth finishing the work.

PR #1347, opened 2019-06-07
feat(execute/row_fn): filter row functions now allow nil references (nils must be safely checked by the user)

aanthony1243:


aanthony1243:
likely to be covered by other work.  will save branch for now.

PR #1348, opened 2019-06-07
feat(compiler): implement nulls in the compiler runtime

### LINKED ISSUE #1329: Update filter to only return records that evaluate to true
The SPEC states that filter only returns records for which the predicate evaluates to `true`.

https://github.com/influxdata/flux/blob/15e15bd756e290d797de67afd90e44769db52404/docs/SPEC.md#L1995

Filter must be updated to understand _null_ values. Specifically it must not produce records when its predicate function returns _null_.

jsternberg:
BREAKING CHANGE: This updates the compiler function so there is only one
`Eval` function instead of one for each type. This may break code that
called these other methods.

- [ ] docs/SPEC.md updated
- [ ] Test cases written

Fixes #1329.

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -20,6 +20,10 @@ func init() {
 }
 
 func ValueEqual(x, y values.Value) bool {
+	if x == values.Null && y == values.Null {
+		return true
+	}
+
 	switch k := x.Type().Nature(); k {
 	case semantic.Object:
 		if x.Type() != y.Type() {
@@ -139,6 +143,7 @@ func TestCompileAndEval(t *testing.T) {
 	}{
 		{
 			name: "simple ident return",
+			// f = (r) => r
 			fn: &semantic.FunctionExpression{
 				Block: &semantic.FunctionBlock{
 					Parameters: &semantic.FunctionParameters{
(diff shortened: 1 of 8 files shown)


compiler/compiler_test.go
```diff
@@ -484,19 +490,363 @@ func TestCompileAndEval(t *testing.T) {
...
+			input: values.NewObjectWithValues(map[string]values.Value{
+				"a": values.NewBool(true),
+				"b": values.NewBool(false),
+			}),
+			want:    values.NewBool(true),
+			wantErr: false,
+		},
+		{
+			name: "call with nonexistant value",
+			// f = (r) => r.a + r.b
```
jlapacik: I would not have expected this test case to type check because `r.b` has to type - it's null. If you replicate this test case in the semantic package (type inference test), would it pass there too?

jsternberg: I'll try writing the tests there tomorrow, but the cool part is it doesn't actually have a type because the value isn't present in the record value. Since type checking happens before the `values.Value` stuff, null isn't technically present anywhere and the type checking seems to pass around the `Tvar` everywhere so weird things like this work.

```
r = {a: 1}
a = r.b
c = a + 2
```

This ends up correctly mapping `a` to be an integer because `+` sets the type constraint and there's nothing else to contradict it.

The problem seems to be when an operator doesn't set a type constraint like the comparison operators. They fail because they can't infer the type from something that doesn't exist so we'll have to find some way to resolve `Tvar`'s with a default value possibly. I think constraints may work for that, but I need to do more investigation.

jlapacik: Oh that's very interesting. I did not think about that. So basically the worst that can happen is that a type variable just stays around and never gets assigned to a concrete type. But there's still no contradiction or anything, so it passes type checking. That's really cool.

I don't think it should work any differently for comparison operators. In fact, I just confirmed that those operators don't constrain their operands to be of equal types, which is definitely a bug. You should not be able to compare values of different types.

jlapacik: Here's the issue describing the type checking bug you found for comparison operators #1356. Feel free to close it out with these changes.

semantic/poly_types.go
```diff
@@ -822,6 +837,40 @@ func (k ObjectKind) resolvePolyType(kinds map[Tvar]Kind) (PolyType, error) {
...
+	return n
+}
+func (n NullableKind) unifyKind(kinds map[Tvar]Kind, k Kind) (Kind, Substitution, error) {
+	switch k := k.(type) {
+	case ObjectKind:
+		return nil, nil, errors.New("cannot unify null with object kind")
+	case ArrayKind:
+		return nil, nil, errors.New("cannot unify null with array kind")
+	case NullableKind:
+		return k, nil, nil
```
jlapacik: Null kinds must unify with other null kinds, because otherwise `a + b` wouldn't type check if both are null correct? Also if we are unifying two nullable kinds, shouldn't we also unify the types they describe like so:
```go
case NullableKind:
    sub, err := unifyTypes(kinds, n.T, k.T)
    if err != nil {
        return nil, nil, err
    }
    return k, sub, nil
```

semantic/poly_types.go
```diff
@@ -822,6 +837,40 @@ func (k ObjectKind) resolvePolyType(kinds map[Tvar]Kind) (PolyType, error) {
...
+	T PolyType
+}
+
+func (NullableKind) MonoType() (Type, bool)        { return Nil, true }
+func (NullableKind) freeVars(*Constraints) TvarSet { return nil }
+func (n NullableKind) resolveType(kinds map[Tvar]Kind) (Type, error) {
+	return Nil, nil
+}
+func (n NullableKind) resolvePolyType(kinds map[Tvar]Kind) (PolyType, error) {
+	return n.T, nil
```
jlapacik: Why is this not `return n.T.resolvePolyType(kinds)`?

semantic/poly_types.go
```diff
@@ -822,6 +837,40 @@ func (k ObjectKind) resolvePolyType(kinds map[Tvar]Kind) (PolyType, error) {
...
+// variable to be the null type if no other type is
+// more appropriate.
+type NullableKind struct {
+	T PolyType
+}
+
+func (NullableKind) MonoType() (Type, bool)        { return Nil, true }
+func (NullableKind) freeVars(*Constraints) TvarSet { return nil }
+func (n NullableKind) resolveType(kinds map[Tvar]Kind) (Type, error) {
+	return Nil, nil
```
jlapacik: Is it possible for resolveType to be called on a NullableKind whose T value points to an integer or some concrete type?

semantic/solve.go
```diff
@@ -189,9 +190,14 @@ func unifyKindsByType(kinds map[Tvar]Kind, tv Tvar, t PolyType) (Substitution, e
...
 	case ObjectKind, ArrayKind:
-		_, ok := t.(Tvar)
-		if !ok {
-			return nil, errors.New("invalid type for kind")
+		// TODO(jsternberg): This appears to be unused and to cause an error
+		// when the error is actually handled correctly.
+		// if _, ok := t.(Tvar); !ok {
+		// 	return nil, errors.New("invalid type for kind")
+		// }
+	case NullableKind:
```
jlapacik: I guess there's no way to unify a nullable kind with a regexp type directly since kinds unify with kinds and types unify with types. I'm a little concerned that this might not catch all cases, since this method wasn't even being used.

semantic/poly_types.go
```diff
@@ -785,13 +785,28 @@ func (l ObjectKind) unifyKind(kinds map[Tvar]Kind, k Kind) (Kind, Substitution,
...
-				return nil, err
+		if _, ok := ft.(invalid); ok {
+			continue
+		}
+
+		t, err := func() (Type, error) {
+			switch ft := ft.(type) {
+			case Tvar:
+				// Check if this can be resolved.
+				if _, ok := kinds[ft]; !ok {
```
jlapacik: If this property has not been assigned a type, shouldn't it be associated with a NullableKind in the `kinds` map? When does `!ok` evaluate to true?

semantic/constraints.go
```diff
@@ -358,7 +364,7 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
 		}
 		v.cs.AddKindConst(tv, ObjectKind{
 			properties: map[string]PolyType{n.Property: ptv},
-			lower:      LabelSet{n.Property},
+			lower:      LabelSet{},
```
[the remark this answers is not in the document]
jsternberg: I was going to keep it because it would be useful if we ever wanted to make object literals require certain properties. If we end up deciding against it, we can remove this.

semantic/poly_types.go
jsternberg: This is where the implementation gets really weird. The kind only applies to `Tvar` and so these methods only get called if the variable is a `Tvar`. If I call `unifyTypes`, then `unifyTypes` consults the kind and this becomes an infinite loop. Objects and arrays don't have to deal with this because they unify their properties rather than themselves.

It seems like this works because when we unify `Tvar` values, it unifies their kinds if they exist so ensuring that both of them have the same kind and therefore a reference to the same `Tvar` guarantees they'll be the same type.

semantic/poly_types.go
jsternberg: It causes an infinite loop is mostly the reason. But the idea behind it is that when you are consulting `NullableKind`, you have an unresolved `PolyType` otherwise you wouldn't be consulting the `Kind` to determine what the type is.

semantic/poly_types.go
jsternberg: I don't think it is but I spent a bunch of time trying to figure that out. I believe that if this resolves to a concrete type, then you get that type directly instead of the `Tvar` and the only time kinds are consulted is when you have a `Tvar`.

semantic/poly_types.go
jsternberg: Whoops. This was added as my first attempt before using constraints. It didn't work. I'll revert this section.

semantic/solve.go
jsternberg: It might not. This function was used, but the error was never checked and the method has no side effects so I changed it to check for errors. It might be good to make it so this function actually returns a substitution and that substitution gets used so it is more than just a check.

semantic/poly_types.go
jlapacik: Yes you are definitely correct. Resolving this conversation.

semantic/poly_types.go
jlapacik: This is again completely correct. Resolving this conversation.

semantic/poly_types.go
jlapacik: Ok I agree. We can't unify the underlying types because it would be an infinite loop. And it's not necessary since if a null kind is being unified with another kind, then it must be true that the underlying types of each are constrained to be equal and will be unified at some point. And so a null kind should only unify with another null kind and there is no need for further recursive unification.

semantic/solve.go
jlapacik: I don't think `unifyKindsByType` does anything useful and I think it should go away. In order to ensure that a regexp type is never nullable, I think we should define a `NonNullableKind`. Regexp types as well as composite types (functions, objects, etc...) should then be constrained to be non-nullable kinds. This would ensure that a regexp type and a nullable type would never pass unification.

Thoughts?

jsternberg:
The discussion on type inference changes will now be in #1363. I split it off into a separate pull request and I will begin writing tests there.

values/binary.go
```diff
@@ -22,9 +22,33 @@ func LookupBinaryFunction(sig BinaryFuncSignature) (BinaryFunction, error) {
...
+// binaryFuncLookup contains a mapping of BinaryFuncSignature's to
+// the BinaryFunction that implements them.
+//
+// The values passed into these functions will be non-nil so a null
+// check is unnecessary inside of them.
+//
+// Even though nulls will never be passed to these functions,
+// the left or right type can be defined as nil. This is used to
+// mark that it is valid to use the operator between those two types,
+// but the function will never be invoked so it can be nil.
```
jlapacik: Can we just remove these elements from the map since they'll never be called?

jsternberg: It's to keep the binary lookup error message working. So if you try to do `int + null` it won't lookup and it will fail. Honestly, the ones for the comparison operators are going away as soon as we start working on implementing some form of typeclasses.

compiler/runtime.go
```diff
@@ -54,17 +33,8 @@ func (c compiledFn) validate(input values.Object) error {
...
-			if v.Nature() == semantic.Object && properties[k].Nature() == semantic.Object {
-				for f := range v.Properties() {
-					if _, ok := properties[k].Properties()[f]; !ok {
-						return fmt.Errorf("parameter %q has the wrong type, expected %v got %v", k, v, properties[k])
-					}
-				}
-			} else {
-				return fmt.Errorf("parameter %q has the wrong type, expected %v got %v", k, v, properties[k])
-			}
+		if !values.AssignableTo(properties[k], v) {
```
jlapacik: This is just a precaution right? If this were to return false, it would never have compiled in the first place, right?

jsternberg: That's been my impression. This code works fine when you comment out this entire section, but I don't know if that's a good type system or a lack of test coverage.

PR #1360, opened 2019-06-11
Feat/regexp compile

### LINKED ISSUE #1127: Feature request: ability to build a regex from a string
It would be terrific to be able to create a regular expression from a string—essentially to have [`regexp.Compile`](https://golang.org/pkg/regexp/#Compile) available in Flux.

It would make the “variables” feature in the InfluxDB UI more powerful. We programmatically insert variable assignments into dashboard queries before executing them. The value of the assignment is based on what a user selects from a dropdown in the UI, and the query runs every time a new value is selected. We want to enable writing dashboard queries like this:

```
import "regexp"

data |> filter(fn: (r) => r.region =~ regexp.Compile(v.region + ".*"))
```

which after the substitution has taken place, might look something like this:

```
import "regexp"

option v = {
  region: "us-west-"
}

data |> filter(fn: (r) => r.region =~ regexp.Compile(v.region + ".*"))
```

ewendai:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

closes #1127


stdlib/testing/testdata/filter_by_regex.flux
```diff
@@ -39,7 +40,7 @@ outData = "
 t_filter_by_regex = (table=<-) =>
 table
   |> range(start: 2018-05-20T19:53:26Z)
-  |> filter(fn: (r) => r["name"] =~ /.*0/)
+  |> filter(fn: (r) => (r["name"] =~ /.*0/))
```
aanthony1243: r.name =~ regexp.regexpCompile(v:".*0")

stdlib/regexp/regexp.flux
```diff
@@ -0,0 +1,7 @@
+package regexp
+
+builtin compileRegexp
```
aanthony1243: let's just call this compile,   so `regex.compile`  which has less name stutter

stdlib/regexp/regexp.go
```diff
@@ -0,0 +1,42 @@
...
+			}),
+			func(args values.Object) (values.Value, error) {
+				v, ok := args.Get("v")
+				if !ok {
+					return nil, errors.New("missing argument v")
+				}
+
+				if v.Type().Nature() == semantic.String {
+					re, ok := regexp.Compile(v.Str())
+					return values.NewRegexp(re), ok
```
aanthony1243: for errors, we usually use a variable `err`  unless it's already been used.   `ok` is used as a name when the function returns a boolean check instead of an error type.  

more important:   check the error before you call `values.NewRegexp`  otherwise, you may get bad things happening if `re` is null.

stdlib/regexp/regexp.go
```diff
@@ -0,0 +1,45 @@
...
+			}),
+			func(args values.Object) (values.Value, error) {
+				v, ok := args.Get("v")
+				if !ok {
+					return nil, errors.New("missing argument v")
+				}
+
+				if v.Type().Nature() == semantic.String {
+					re, err := regexp.Compile(v.Str())
+					if err == nil {
```
aanthony1243: canonical go expects you to ask  `if err != nil {...}`

PR #1362, opened 2019-06-12
feat(internal/rust): ast marshalling

### LINKED ISSUE #1350: Make Rust AST produce correct JSON serialization
Implement `serde::ser::Serialize` for all AST struct in the Rust parser. 

The JSON must match the JSON produced by the Go struct/swagger spec.

affo:
Closes #1350 

Tests pass locally.

What I did here is to port the tests we had in Go to Rust and make them succeed.  
It seems to me that our swagger API isn't explicit in which missing fields we tolerate.

It seems that our Go implementation would be fine with empty JSONs, such as:

```
{"type": "Package"}
```

Because our Go implementation would deserialize a `Package` and fill the fields to their zero value (included `nil`).  In other words, in Go this:

```
{"type": "Package"}
```

is the same of this:

```
{"type": "Package", "path": "", "package":"", "files": null}
```

once deserialized.

This comes in handy, because, in Rust, objects are not nullable and so we must give default values to every field for an AST node. Those default values would be deserialized as expected, so there is no problem (hold on for slices).

In spite of this, I skipped the serialization of some fields if they are "empty" to match our implementation in Go (see`BaseNode` in every AST node, and `path` in `Package`).

I followed the go implementation to understand what could really be optional and what not. In these cases we use `Option`, and the serialization of `Option::None` would be `null`. I only added the optional value for `Property`.
  
I had to adapt a pair of tests at the beginning for `Package`, because some things shouldn't be optional (e.g. `imports` and `files`) and the zero value for a slice in Go is `nil`, while in Rust is `[]`.  
However, once deserialized, iterating over a `nil` slice or an empty one doesn't make any difference in Go.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/rust/ast/Cargo.lock
+++ b/internal/rust/ast/Cargo.lock
@@ -4,15 +4,54 @@
 name = "ast"
 version = "0.1.0"
 dependencies = [
+ "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
+name = "autocfg"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "chrono"
+version = "0.4.6"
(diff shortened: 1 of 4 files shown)


affo:
I am merging this despite `go vet` is failing.

This way @jlapacik can work on the last version of AST for the parser.

If you decide to change the behavior for zero values, we can make those changes once @jlapacik has finished with the parser and according to the changes he will introduce in the AST.

PR #1363, opened 2019-06-12
feat(semantic): add nullable kind

### LINKED ISSUE #1353: Create kind constraint for Nullable types
We want to model the basic types that can be nullable using the kind constraints.

The reason is that it allows us to treat types and nullability as orthogonal concepts in the type system. Thus allowing for a type that can expression non-nullable basic types and nullable basic types.

For example 

```
a = true
```

Type inference can correctly infer that `a` is a non-nullable boolean value. This allows for better type checking and error messages to the user. 

This design is not final, we can explore related designs.

jsternberg:
The nullable kind is used for marking that a type variable may be null
in the semantic graph. This kind constraint is added to any values
retrieved using a `MemberExpression` because these are retrieved from
records and records can return null values if the key isn't present.

The nullable kind will unify with anything and it will discard itself if
it is unified with any other kind constraint. At that point, it is
determined that the type cannot be nullable anymore because the other
kind constraint says it cannot be.

If a type variable that is nullable never resolves itself to a concrete
type, it will determine itself to be the null type when used.

Fixes #1353.

- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/semantic/inference_test.go
+++ b/semantic/inference_test.go
@@ -1729,19 +1729,6 @@ name(p: j)
 			},
 		},
 		{
-			name: "structural polymorphism error",
-			script: `
-john = {name: "John", age: 30, weight: 100.0}
-jane = {name: "Jane", lastName: "Smith"}
-
-fullName = (p) => p.name + " " + p.lastName
-
-fullName(p:jane)
-fullName(p:john)
-`,
-			wantErr: errors.New(`type error 8:1-8:17: missing object properties (lastName)`),
-		},
-		{
 			name: "function with polymorphic object parameter",
 			script: `
 foo = (r) => ({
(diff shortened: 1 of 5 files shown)


semantic/poly_types.go
```diff
@@ -761,8 +761,17 @@ func (l ObjectKind) unifyKind(kinds map[Tvar]Kind, k Kind) (Kind, Substitution,
...
 
+	// Ensure that all of the values that are missing are allowed to be missing.
 	diff := lower.diff(upper)
-	if len(diff) > 0 {
+	for _, lbl := range diff {
+		if ptv, ok := properties[lbl].(Tvar); ok {
+			// If this tvar is nullable, then it is allowed
+			// to be missing.
+			kind := kinds[ptv]
+			if _, ok := kind.(NullableKind); ok {
```
jlapacik: When will this not be the case? All member expressions induce a nullable kind constraint on the property type. Are we certain that kind will be removed from the `kinds` map at this point?

semantic/solve.go
```diff
@@ -107,6 +93,89 @@ func (sol *Solution) solve() error {
...
+			// We may want to visit this constraint.
+			// Check if any other unvisited tvar
+			// occurs in this type.
+			if canVisit := func() bool {
+				for tvr := range unvisited {
+					if tvl == tvr {
+						continue
+					}
+
+					ks := sol.cs.kindConst[tvr]
```
jlapacik: Did you mean `ks := sol.cs.kindConst[tvl]`?

semantic/poly_types.go
jsternberg: Yes because `unifyKinds` has been modified to require all `Tvar` values it depends on to have already been resolved. So if a `NullableKind` constraint is added to the same `Tvar` that is also an `ObjectKind`, it will be resolved as an `ObjectKind`. This is the part that causes the added test to succeed.

jsternberg: You can see the line getting hit in code coverage: https://codecov.io/gh/influxdata/flux/pull/1363/src/semantic/poly_types.go?before=semantic/poly_types.go#L775

semantic/solve.go
jsternberg: No. The `unvisited` map contains all of the `Tvar` values that I still need to resolve. So here I am checking that every `Tvar` that hasn't been visited is not contained within the current kind constraints.

semantic/inference_test.go
```diff
@@ -1988,6 +2002,13 @@ r.a < r.b`,
 				},
 			},
 		},
+		{
+			name: "access nonexistant object",
+			script: `r = {a: 1}
+r.b.c + r.a
+`,
+			wantErr: errors.New("missing object properties (b)"),
```
jlapacik: We discussed in slack but it would be nice to have a similar test case for properties that resolve to regexp or function types. We might have to skip them for right now and create a specific issue to unskip them later.

semantic/solve.go
jlapacik: We discussed in slack that this should be `ks := sol.cs.kindConst[tvl]` and so I agree this algorithm is correct.

PR #1370, opened 2019-06-13
Add Random Package with uint64

rawkode:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


mark-rushakoff:
Just a couple high-level notes without reading the code too closely:

- Top-level functions in the math/rand package like `rand.Intn` share a common mutex, so this may not scale well with many concurrent queries in a single Flux runtime calling `rand`.
- If a user needs a _predictable_ but _arbitrary_ sequence of numbers -- he may want to provide a seed so he gets the same sequence every time. IMO any rand API we introduce should support seeding.
- If we were to use https://www.godoc.org/golang.org/x/exp/rand#PCGSource -- i.e. a known "standard" PRNG algorithm -- then it is more straightforward to port identical rand behavior to any other Flux implementation.

See also https://github.com/golang/go/issues/21835 for discussion on the flaws in Go's math/rand.

rawkode:
Great feedback, @mark-rushakoff; thank you.

Happy to switch to the other implementation

aanthony1243:
a few things...

1.  I agree with @mark-rushakoff that we should provide the flexibility to add a seed value as a second parameter
2.  avoiding the shared mutex will be tricky.  we don't want to instantiate a new source for each function call, that'd be worse than locking.  Do we create one shared source for each query instance?  Do we expose the idea of a source at the language level?  These seem like after-the-fact optimizations so I don't know if they should be a blocker to merging this.  
3.  if that portable source is do-able, it's not a bad idea. 
4.  I, for one, would rather have a rand( ) that returns a float on [0,1].   I can always simulate a random int by multiplying that by math.maxint

overall, I think that establishing a "what does random number generation look like in flux?" is a good thing.  

Finally, could you add an integration test?   you can find examples in influxdb/v1/fluxtests that show how to write pure-flux tests for a package

mark-rushakoff:
Maybe the right compromise here would look like Go's model:

- Package level functions like `rand.int` and `rand.float`. Perhaps the package-level functions  do not contain a `seed` -- the seed, and whether the backing RNG is shared across queries, is implementation-determined.
- Ability to generate an `rng` object that can optionally be instantiated with a `seed`, defaulting to an implementation-defined choice for seed. The methods on `rng` mirror the package-level methods.

Package-level functions keep it simple when the stakes are low, but instantiating an `rng` value gives you control and predictability when you need it.

aanthony1243:
@rawkode this is so far behind the current flux, I don't know if it's going to make it to production.   It's a good idea, but maybe it's just not the right time for it.   It's also likely that we'll need to build out RNG in a custom way, and not port-in the Go calls.

rawkode:
No problem 👍

PR #1371, opened 2019-06-14
refactor(lang): add extern and now fields to FluxCompiler

wolffcm:
Related influxdb PR:
https://github.com/influxdata/influxdb/pull/14135

### Done checklist
- [x] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -43,6 +43,8 @@ type CompileOption func(*compileOptions)
 type compileOptions struct {
 	verbose bool
 
+	extern *ast.File
+
 	planOptions struct {
 		logical  []plan.LogicalOption
 		physical []plan.PhysicalOption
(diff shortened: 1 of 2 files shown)


lang/compiler.go
```diff
@@ -84,11 +84,14 @@ func Verbose(v bool) CompileOption {
 
 // Compile evaluates a Flux script producing a flux.Program.
 // now parameter must be non-zero, that is the default now time should be set before compiling.
-func Compile(q string, now time.Time, opts ...CompileOption) (*AstProgram, error) {
+func Compile(extern *ast.File, q string, now time.Time, opts ...CompileOption) (*AstProgram, error) {
```
jsternberg: Can the extern values get added as a compile option? It seems weird to make it the first parameter and required when it is optional.

wolffcm: Yes, good idea.  I repushed.

PR #1381, opened 2019-06-17
feat(interpreter): nulls work in table and row functions

jsternberg:
- [x] docs/SPEC.md updated
- [x] Test cases written

The interpreter has been updated for the work on nulls. The row and
table functions now work with nulls.

Unfortunately, some potential issues were found. It seems type inference
doesn't resolve some types for the interpreter correctly specifically
related to functions so certain type constraints don't get identified.
This causes them to erroneously be marked as null types rather than
their appropriate type and so there's currently a hack included to make
it work while still allowing us to evaluate null types in the interim.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -254,10 +254,7 @@ func (itrp *Interpreter) doExpression(expr semantic.Expression, scope Scope) (va
 		if typ := obj.Type().Nature(); typ != semantic.Object {
 			return nil, fmt.Errorf("cannot access property %q on value of type %s", e.Property, typ)
 		}
-		v, ok := obj.Object().Get(e.Property)
-		if !ok {
-			return nil, fmt.Errorf("object has no property %q", e.Property)
-		}
+		v, _ := obj.Object().Get(e.Property)
 		if pkg, ok := v.(*Package); ok {
 			// If the property of a member expression represents a package, then the object itself must be a package.
 			return nil, fmt.Errorf("cannot access imported package %q of imported package %q", pkg.Name(), obj.(*Package).Name())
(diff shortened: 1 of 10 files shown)


values/binary.go
```diff
@@ -22,15 +22,15 @@ func LookupBinaryFunction(sig BinaryFuncSignature) (BinaryFunction, error) {
...
+	return binaryFuncNullCheck(f, sig), nil
 }
 
 // binaryFuncNullCheck will wrap any BinaryFunction and
 // check that both of the arguments are non-nil.
 //
 // If either value is null, then it will return null.
 // Otherwise, it will invoke the function to retrieve the result.
-func binaryFuncNullCheck(fn BinaryFunction) BinaryFunction {
+func binaryFuncNullCheck(fn BinaryFunction, sig BinaryFuncSignature) BinaryFunction {
```
jlapacik: Why is this added to the function signature?

interpreter/interpreter.go
```diff
@@ -658,6 +676,13 @@ func (f function) Call(argsObj values.Object) (values.Value, error) {
 	return v, nil
 }
 func (f function) doCall(args Arguments) (values.Value, error) {
+	if f.itrp == nil {
```
jlapacik: Why was this check needed? Does a test case need to be added?

interpreter/interpreter.go
```diff
@@ -313,10 +310,22 @@ func (itrp *Interpreter) doExpression(expr semantic.Expression, scope Scope) (va
 			return nil, err
 		}
 
+		ltyp := itrp.typeof(e.Left, l.Type())
+		rtyp := itrp.typeof(e.Right, r.Type())
+		// TODO(jsternberg): This next section needs to be removed
+		// since type inference should give the correct type.
+		if ltyp == semantic.Nil && l.Type() != semantic.Nil {
+			// There's a weird bug in type inference where it
+			// determines the type is null even when it's not.
```
jlapacik: We talked in Slack, but I'd like to see a test case added to the semantic package highlighting this bug. This appears to be an existing issue, but I want to make sure that the recent null work hasn't expanded the bug.

interpreter/interpreter.go
jsternberg: It is possible for a function to be called with no interpreter. The old implementation didn't use the interpreter for type information, but the new one does. Basically, all of the calls to `itrp` inside of function could be null for a decently long time and we just got lucky.

values/binary.go
jsternberg: I added it originally. This is a mistake that was kept when I modified things.

PR #1386, opened 2019-06-17
fix(stdlib/strings): fix joinStr, including adding an EndToEnd Test

ewendai:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/strings/strings.go
+++ b/stdlib/strings/strings.go
@@ -453,7 +453,7 @@ func init() {
 			"joinStr",
 			semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
 				Parameters: map[string]semantic.PolyType{
-					"arr": semantic.Array,
+					"arr": semantic.NewArrayPolyType(semantic.String),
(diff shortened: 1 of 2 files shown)


stdlib/strings/strings.go
```diff
@@ -466,8 +466,11 @@ func init() {
 				if !ok {
 					return nil, fmt.Errorf("missing argument %q", "arr")
 				}
-				if val.Type().Nature() != semantic.Array {
-					return nil, fmt.Errorf("expected argument %q to be of type %v, got type %v", "arr", semantic.Array, val.Type().Nature())
+				arr := val.Array()
+				if arr.Len() > 0 {
+					if arr.Get(0).Type().Nature() != semantic.String {
```
wolffcm: I don't think this will work right if `arr` has zero elements.  You can still peek at the type though, maybe try this?
```suggestion
					if arr.Type().ElementType().Nature() != semantic.String {
```

PR #1397, opened 2019-06-20
Docs/unicode

ewendai:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/Unicode.md
+++ b/docs/Unicode.md
@@ -32,19 +32,51 @@ and length of strings differently?
 
 ### Python
 
+Has `bytes.decode()` and `str.encode()` methods that translate unicode to string and vice versa. 
+One-character unicode strings can also be created with the chr() built-in function.
+
+`unicodedata.normalize()` converts strings to one of several normal forms, where letters followed
+ by a combining character are replaced with single characters. normalize() can be used to perform
+ string comparisons that won’t falsely report inequality if two strings use combining characters
+ differently.
+
+```
+import unicodedata
+
+def compare_strs(s1, s2):
+    def NFD(s):
+        return unicodedata.normalize('NFD', s)
+
+    return NFD(s1) == NFD(s2)
+
+single_char = 'ê'
+multiple_chars = '\N{LATIN SMALL LETTER E}\N{COMBINING CIRCUMFLEX ACCENT}'
+print('length of first string=', len(single_char))
+print('length of second string=', len(multiple_chars))
+print(compare_strs(single_char, multiple_chars))
+```
+Returns:
+```
+length of first string= 1
+length of second string= 2
+True
+```
+
 Documentation: https://docs.python.org/3/howto/unicode.html
 
 ### Go
+
+Go encodes string as byte array and thus the `length` is determined by the number of bytes 
+instead of the number of characters.
 
 Documentation/Source:
 - https://golang.org/pkg/unicode/
 - https://coderwall.com/p/k7zvyg/dealing-with-unicode-in-go
 
-### Rust 
-
-??
-
 ### Ruby
+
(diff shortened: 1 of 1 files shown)


docs/Unicode.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Oracle
+
+Does support unicode.
+
+The `NCHAR` datatype stores data encoded as Unicode. The column length specified for the `NCHAR` and 
+`NVARCHAR2` columns is always the number of characters instead of the number of bytes.
+
+"The lengths of the SQL `NCHAR` datatypes are defined as number of characters. This is the same as 
+the way they are treated when using wchar_t strings in Windows C/C++ programs. This reduces programming
```
aanthony1243: interesting!   so C++ and other languages just make a 'character' take up more space.  This is generally non-compact and probably not the solution we will offer, since this would require a custom string type internally.

docs/Unicode.md
```diff
@@ -0,0 +1,129 @@
...
+complexity."
+
+Documentation: https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch6unicode.htm
+
+### MongoDB
+
+Has functions to deal by both bytes and by code points. For those functions that work through bytes, returns
+an error if user requests something that does not split a character evenly according to its number of bytes.
+
+Example: `{ $substrBytes: [ "cafétéria", 7, 3 ] }` \
```
aanthony1243: hmm...this is interesting.  gives the user some feedback but it's up to the user to know character sizes?  are there functions to help them around this?

docs/Unicode.md
```diff
@@ -0,0 +1,129 @@
...
+Documentation: https://docs.python.org/3/howto/unicode.html
+
+### Go
+
+Go encodes string as byte array and thus the `length` is determined by the number of bytes 
+instead of the number of characters.
+
+Documentation/Source:
+- https://golang.org/pkg/unicode/
+- https://coderwall.com/p/k7zvyg/dealing-with-unicode-in-go
```
aanthony1243: can you summarize this a bit?  I know go has something called a 'glyph'  is that what they call a unicode character? what functions are available for detecting and processing unicode characters in a string?   Is there a way to iterate over characters or isolate a single unicode char?

docs/Unicode.md
```diff
@@ -0,0 +1,129 @@
...
+### Length
+
+Should we define length as the number of characters or the number of bytes?
+(Will be the same for the 128 US-ASCII characters that we are used to, but will be different for other
+languages that Flux may want to support.)
+
+### Substring/Slicing/Indexing
+
+Will probably be more convenient to define length as number of values in the resulting array 
+when `strings.split(v: "...", t: "")` is applied. Alternatively, can define length of array
```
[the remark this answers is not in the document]
ewendai: I was thinking that length of a string `"influx"` can be defined as `len(strings.split(v: "influx", t: "")`, which would be `6`. And length of a string `"汉字"` will be `2` since `len(strings.split(v: "汉字", t: "")` will be `2`.

docs/Unicode.md
ewendai: There is a `stringLenBytes` that user can utilize to find the number of bytes in a string? And they can find the number of bytes in 1 code point by entering a 1 character string. 

Other than that I'm not sure. Maybe they just expected more info to be easy to Google?

docs/Unicode.md
ewendai: Only glyphs I can find are Hieroglyphs that are part of the *Rangetables. 

There's the functions in the Unicode package and [utf8 pacakge](https://golang.org/pkg/unicode/utf8/). utf8 does provide ways to see how many bytes each character is. 

I'll add this to the doc.

PR #1398, opened 2019-06-21
test(universe/typeconv): wrote test cases for type conversions

### LINKED ISSUE #1157: bool() function returns "true" for 0 and "false" for 1
See https://github.com/influxdata/flux/blob/90ce58c56b50fc000abf8770290fbda9d1108a38/stdlib/universe/typeconv.go#L456-L482

Converting numeric values to boolean using `bool()` yields `true` for `0` and `false` for `1`. It should be the other way around. Fix is easy enough, but should probably be accompanied by extensive tests for the whole file.

sauren-khosla:
### Done checklist
- [X] Test cases written
fixes #1157

### DIFF
--- a/stdlib/universe/typeconv_test.go
+++ b/stdlib/universe/typeconv_test.go
@@ -1,15 +1,239 @@
 package universe
 
 import (
-	"github.com/influxdata/flux/values"
 	"testing"
+	"errors"
+	"github.com/influxdata/flux/values"
+	"math"
 )
+
+
+func TestTypeconv_String(t *testing.T) {
+	testCases := []struct {
+		name string
+		v    interface{}
+		want string
+		expectErr error
+	}{
+		{
+			name: "string(v:1)",
+			v:    int64(541),
+			want: "541",
+		},
+		{
+			name: "string(v:2)",
+			v:    uint64(501),
+			want: "501",
+		},
+		{
+			name: "string(v:3)",
+			v:    float64(653.28),
+			want: "653.28",
+		},
+		{
+			name: "string(v:4)",
+			v:    bool(true),
+			want: "true",
+		},
+		{
+			name: "string(v:5)",
+			v:    bool(false),
+			want: "false",
+		},
+		{
+			name: "string(v:6)",
+			v:    values.Time(1136239445999999999),
+			want: "2006-01-02T22:04:05.999999999Z",
+		},
+		{
+			name: "string(v:7)",
+			v:    values.Duration(184000000000),
+			want: "3m4s",
+		},
+		{
+			name: "string(v:8)",
+			v:    int64(-541),
+			want: "-541",
+		},
+	}
+	for _, tc := range testCases {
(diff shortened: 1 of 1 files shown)


stdlib/universe/typeconv_test.go
```diff
@@ -5,32 +5,239 @@ import (
...
+func TestTypeconv_String(t *testing.T) {
+	testCases := []struct {
+		name  	  string
+		v     	  interface{}
+		want  	  string
+		expectErr error
+	} { 
+		{
+			name : "string(v:1)",
+			v : int64(541),
```
aanthony1243: add a test case with a negative number?

stdlib/universe/typeconv_test.go
```diff
@@ -5,32 +5,239 @@ import (
...
+			v : false,
+			want : int64(0),
+		},
+		{
+			name : "int64(v:5)",
+			v : values.Time(1136239445999999999),
+			want : int64(1136239445999999999),
+		},
+		{
+			name : "int64(v:6)",
```
aanthony1243: we do in fact want to verify that functions return errors as expected, so we want a case like this: 
```
{
  name: "cannotConvertInt", 
  v: "notanumber"
  want: nil
  wantErr: errors.new("cannot parse notanumber as an integer")
}
```

and then in your test case driver, 

```
result, err := c.Call(args)
if tc.wantErr != nil {
	if got, want := err.Error(), tc.wantErr.Error(); got != want {
		t.Errorf("unexpected error want: %s got: %s", want, got)
 	}
	return
}
```

stdlib/universe/typeconv_test.go
```diff
@@ -2,51 +2,505 @@ package universe
...
-				"v" : values.New(tc.v),
+		t.Run(tc.name, func(t *testing.T) {
+			myMap := map[string]values.Value{
+				"v": values.New(tc.v),
+			}
+			args := values.NewObjectWithValues(myMap)
+			c := &intConv{}
+			got, err := c.Call(args)
+			if err != nil {
+				if want, got := tc.expectErr.Error(), err.Error(); got != want {
```
aanthony1243: around this add a check, `if tc.expectErr != nil  { ... }  else {t.Errorf("unexpected error %s", err.Error)}

Also in all test cases if you are not expecting an error in that case, then you should not have anything in there.

stdlib/universe/typeconv_test.go
```diff
@@ -2,51 +2,487 @@ package universe
...
+			name: "float64(v:6)",
+			v: int64(-753),
+			want: float64(-753),
+		},
+		{
+			name: "float(error)",
+			v: "NotANumber",
+			want: float64(0),
+			expectErr: errors.New("strconv.ParseFloat: parsing \"NotANumber\": invalid syntax"),
+		},
```
nathanielc: Can you add cases for the special float values `+Inf, -Inf, NaN`?

PR #1405, opened 2019-06-21
feat(stdlib/strings): add length and substring functionality includin…

ewendai:
…g unit and integration tests

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/strings/strings.go
```diff
@@ -382,7 +384,69 @@ func generateUnicodeIsFunction(name string, Fn func(rune) bool) values.Function
...
+				return values.NewInt(int64(l)), nil
+			}
+
+			return nil, fmt.Errorf("procedure cannot be executed")
+		}, false,
+	)
+}
+
+func generateSubstring(name string) values.Function {
+	return values.NewFunction(
```
nathanielc: As a small suggestion I would change these to be `var`s instead of functions as the function is only called once to produce a single value. 

```
var substring = values.NewFunction(...)
```

Then in the init function it becomes 
```

	flux.RegisterPackageValue("strings", "substring", substring)
```

stdlib/strings/strings.go
```diff
@@ -382,7 +384,69 @@ func generateUnicodeIsFunction(name string, Fn func(rune) bool) values.Function
...
+			Return:     semantic.Int,
+		}),
+		func(args values.Object) (values.Value, error) {
+			v, ok := args.Get(stringArgV)
+			if !ok {
+				return nil, fmt.Errorf("missing argument %q", stringArgV)
+			}
+
+			if v.Type().Nature() == semantic.String {
+				arr := strings.Split(v.Str(), "")
```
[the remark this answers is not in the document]
nathanielc: This method is going to be the fastest https://golang.org/pkg/unicode/utf8/#RuneCountInString we should use it.

stdlib/strings/strings_test.go
```diff
@@ -1250,3 +1251,154 @@ func TestJoinStr(t *testing.T) {
...
+				if got, want := err.Error(), tc.expectErr.Error(); got != want {
+					t.Errorf("unexpected error - want: %s, got: %s", want, got)
+				}
+				return
+			}
+
+			res := result.Str()
+
+			if res != (tc.want) {
+				t.Errorf("string function result %s expected: %v, got: %s", tc.name, tc.want, result)
```
aanthony1243: great tests!

stdlib/strings/strings.go
```diff
@@ -382,7 +385,77 @@ func generateUnicodeIsFunction(name string, Fn func(rune) bool) values.Function
...
+		}
+
+		vStr := v.Str()
+		aInt := int(a.Int())
+		bInt := int(b.Int())
+		if aInt < 0 || bInt > len(vStr) {
+			return nil, fmt.Errorf("indices out of bounds")
+		}
+
+		if (v.Type().Nature() == semantic.String) && (a.Type().Nature() == semantic.Int) && (b.Type().Nature() == semantic.Int) {
```
nathanielc: This type check needs to happen before lines 427-429. Otherwise those methods will panic.

stdlib/strings/strings_test.go
```diff
@@ -1250,3 +1251,163 @@ func TestJoinStr(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func BenchmarkSubstring(b *testing.B) {
```
nathanielc: Did you compare this new implementation to the old one with this benchmark? How did it compare?

stdlib/strings/strings.go
```diff
@@ -382,7 +385,77 @@ func generateUnicodeIsFunction(name string, Fn func(rune) bool) values.Function
...
+			count := 0
+			for _, c := range vStr {
+				if count < aInt {
+					count++
+					continue
+				}
+				if count >= bInt {
+					break
+				}
+				newStr = strings.Join([]string{newStr, string(c)}, "")
```
nathanielc: This line is still constructing a new string with each loop which is allocation heavy. You can use Go's substring here to help.

I am thinking something along the lines of finding the byteStart and byteEnd indexes and then doing this after the loop.

```
return values.NewString(vStr[byteStart:byteEnd]), nil
```
Can you try that and use [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare the two implementations to see which is better?

stdlib/strings/strings_test.go
ewendai: I did, the one that I kept was consistently faster.

stdlib/strings/strings.go
ewendai: using `byteStart` and `byteEnd` makes it run about the same or faster, depending on which end of the string I'm getting the substring from. will be changing it to the faster version

stdlib/strings/strings.go
```diff
@@ -382,7 +385,77 @@ func generateUnicodeIsFunction(name string, Fn func(rune) bool) values.Function
...
+				if count == aInt {
+					byteStart = i
+				}
+				if count >= bInt-1 {
+					byteEnd = i + len(string(c))
+					break
+				}
+				count++
+			}
+			return values.NewString(vStr[byteStart:byteEnd]), nil
```
nathanielc: This looks great, can you share the benchmark results? I am interested in seeing how this change performs compared to the previous implementation.

ewendai: How do I share the benchmark results? Just write a line with the previous implementation and time compared to current?

ewendai:
**Benchmark Test Summary**
- Slowest Implementation: Split given string using `strings.split`, retrieve the desired parts of the string, return after joining using `strings.joinStr`
- Consistently better: Loop over the given string and add the desired parts of the string together 
- About the same if desired substring is at the beginning but better if desired substring is near the end of the given string: Loop over the given string to find `byteStart` and `byteEnd` and return `givenString[byteStart:byteEnd]`

PR #1414, opened 2019-06-24
fix(stdlib): read the table in assert empty to verify it is not empty

jsternberg:
Some table implementations don't know they are empty until `Do` has been
called. `assertEmpty` has been fixed so that it will not rely on
`Empty()` to return the correct value.

The table copy implementation has also been updated so it won't attempt
to retain buffers that have a length of zero for the same reason.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/testing/assert_empty.go
+++ b/stdlib/testing/assert_empty.go
@@ -86,13 +86,30 @@ func (t *AssertEmptyTransformation) RetractTable(id execute.DatasetID, key flux.
 }
 
 func (t *AssertEmptyTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
-	if !tbl.Empty() {
+	if tbl.Empty() {
+		// Table is empty so we succeed.
+		// TODO: The Do method must be called at the moment.
+		return tbl.Do(func(cr flux.ColReader) error {
+			return nil
+		})
+	}
+
+	// Confirm that the table is, in fact, not empty.
+	// Some implementations don't know until Do is called.
+	failed := false
+	if err := tbl.Do(func(cr flux.ColReader) error {
+		if cr.Len() > 0 {
+			failed = true
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	if failed {
 		t.failures++
 	}
-	// TODO: The Do method must be called at the moment.
-	return tbl.Do(func(cr flux.ColReader) error {
-		return nil
-	})
+	return nil
 }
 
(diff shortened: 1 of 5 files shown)


stdlib/testing/assert_empty.go
```diff
@@ -86,13 +86,30 @@ func (t *AssertEmptyTransformation) RetractTable(id execute.DatasetID, key flux.
...
+		// TODO: The Do method must be called at the moment.
+		return tbl.Do(func(cr flux.ColReader) error {
+			return nil
+		})
 	}
-	// TODO: The Do method must be called at the moment.
-	return tbl.Do(func(cr flux.ColReader) error {
+
+	// Confirm that the table is, in fact, not empty.
+	// Some implementations don't know until Do is called.
```
wolffcm: Isn't this a problem?  Seems really unintuitive.  Which implementations don't know they are empty until `Do` is called?

jsternberg: I think the intention was to fix it at some point, but the influxdb one doesn't for simplicity.

wolffcm: Is there an issue for it?  If not, can you file one?

jsternberg: I wasn't going to. I didn't really see it as an issue. I thought of the `Empty()` as more of a hint. In general, requiring that a table know it is empty before `Do` is called means that whatever resource it is currently utilizing must be used before `Do` is called.

Although if we choose to view it that way, it should probably be noted in the interface docs.

jsternberg: I think that it's fine if `Empty()` returns false in situations where the table is empty, but I created influxdata/influxdb#14193 anyway so we can investigate it and see if it's easy enough to implement.

I changed the godoc comment on empty so that it is true for what is currently true at the moment.

nathanielc: The docs state:

```
Empty returns whether the table contains no records.
```

It is not a hint and a table implementation that returns `false` where there are no rows in the table would be an incorrect implementation of the interface.

Yes it requires that the implementation know before Do is called whether there will be at least one record produced. The CSV decoders/encoders implement this correctly and take advantage of it to encode empty tables appropriately.

nathanielc: FWIW I believe the InfluxDB source does implement this correctly simply because the storage RPC calls never produce an empty table.

jsternberg:
We have decided that this isn't a bug in assert empty, but a bug in the influxdb table implementation so I'm going to close this.

PR #1415, opened 2019-06-25
feat(stdlib): add custom PostgreSQL type support

roshie548:
- custom types in PostgreSQL (received as byte arrays) are now parsed and appended

### DIFF
--- a/stdlib/sql/postgres.go
+++ b/stdlib/sql/postgres.go
@@ -2,6 +2,7 @@ package sql
 
 import (
 	"database/sql"
+	"fmt"
 	"github.com/influxdata/flux"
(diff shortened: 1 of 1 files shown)


stdlib/sql/postgres.go
```diff
@@ -42,6 +43,33 @@ func (m *PostgresRowReader) GetNextRow() ([]values.Value, error) {
 			row[i] = values.New(col)
 		case time.Time:
 			row[i] = values.NewTime(values.ConvertTime(col))
+		case []uint8:
+			// Hack for MySQL, might need to work with charset?
```
aanthony1243: these comments should be removed or made specific to postgres.

Issue #1412, opened 2019-06-24
Update type converstion functions to use "with" operator 


nathanielc:
The current implementations of the `toFloat` etc functions drops all columns but the `_value` column being converted. Now that we have `with` support we should update the functions to use `with` so the are usable in a more general sense.

Issue #816, opened 2019-01-11
Remove mergeKey parameter from map


nathanielc:
When #205 is complete we will have a more expressive way to to update records when mapping then the mergeKey boolean logic.

wolffcm:
I'm re-opening this since the parameter was re-added:
https://github.com/influxdata/flux/commit/ea30f953927244b50e67c99bfef94c4c27bf961a

PR #1416, opened 2019-06-25
fix(stdlib/map): remove mergeKey parameter from the map function

nathanielc:
BREAKING CHANGE: The removal of the mergeKey parameter is a breaking
change. Users will either need to use the `with` operator to preserve
columns or explicitly list all columns they wish to preserve through the
map function.
Several other test cases that used map have been simplified to remove
their use of the map function as we want the test cases to focus on
testing the function they are designed to test.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2141,17 +2141,13 @@ When the output record drops a column that was part of the group key that column
 
 Map has the following properties:
 
-| Name     | Type                  | Description                                                                                               |
-| ----     | ----                  | -----------                                                                                               |
-| fn       | (r: record) -> record | Function to apply to each record.  The return value must be an object.                                    |
-| mergeKey | bool                  | MergeKey indicates if the record returned from fn should be merged with the group key.  Defaults to true. |
+| Name | Type                  | Description                                                            |
+| ---- | ----                  | -----------                                                            |
+| fn   | (r: record) -> record | Function to apply to each record.  The return value must be an object. |
 
 
-When merging, all columns on the group key will be added to the record giving precedence to any columns already present on the record.
-When not merging, only columns defined on the returned record will be present on the output records.
-
-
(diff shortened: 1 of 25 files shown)


nathanielc:
Closes #1412  
Closes #816

docs/SPEC.md
```diff
@@ -2161,7 +2157,8 @@ from(bucket:"telegraf/autogen")
                 r.service == "app-server")
     |> range(start:-12h)
     // Square the value
-    |> map(fn: (r) => r._value * r._value)
+    // The output table has all column from the input table because of the use of `with`.
+    |> map(fn: (r) => {r with _value: r._value * r._value})
```
jsternberg: I think we still require parenthesis around the object literal syntax when using a function.

stdlib/universe/map.go
```diff
@@ -174,23 +161,15 @@ func (t *mapTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 				// Determine on which cols to group
 				on = make(map[string]bool, len(tbl.Key().Cols()))
 				for _, c := range tbl.Key().Cols() {
-					on[c.Label] = t.mergeKey || execute.ContainsStr(keys, c.Label)
+					on[c.Label] = execute.ContainsStr(keys, c.Label)
```
affo: Why are we still checking for the columns to group given that `map` cannot remove columns anymore?  
The group key should remain the same.  

I think that `on` is useless now ad you can directly use the group key of the original table in the util function `groupKeyForObject`. 
You still need to check the values because it can cause regrouping.

nathanielc: It is still possible to remove columns if the function does not use `with`. 

I have added  a test for that case.

affo: Definitely right.

Issue #1364, opened 2019-06-12
Flux should provide a way to distinguish between user and internal error


wolffcm:
Some errors happen because a user did something wrong:
- Invalid syntax
- Function call with wrong number/type/name of parameters
- Divide by zero (maybe)
- etc

When Flux is run through an http connection, for these errors it should return 4xx to indicate that the user should change their request before trying again.

Other errors happen because there was an internal error:
- A bug in our code (e.g, a recovered panic)
- Some service that flux depends on is down
- etc

In these cases, the http status should be 5xx.

The flux repo can't depend on `influxdb` and has no http interface of its own, but it should provide someway for clients to make the distinction between a bad request and an internal error.

Consider also that the type of error returned is _not_ necessarily dependent on when the error occurs; it's possible that we hit a user error when evaluating a lambda in a filter for example.

Related to influxdata/idpe#3519.

jsternberg:
It might be worth either copying the same pattern that influxdb is now using for errors and including it in flux, as the pattern is generic, or possibly separating out the error code from influxdb so that flux can use it as a library.

@goller do you have any thoughts on this?

nathanielc:
@jsternberg are you referring to https://github.com/influxdata/influxdb/blob/master/kit/errors/errors.go ?

The code is small enough that I am fine with copying the code directly into Flux. A little copypasta is better than a little dependency.

jsternberg:
Yes, that's what I was referring to.

The potential reason for a dependency is then influxdb and flux could share the code and also the types would evaluate to the same for type switches. I don't think they would currently so parts of influxdb that do this:

```go
if err, ok := err.(*influxdb.Error); ok {
}
```

Those wouldn't function.

Since the code is fairly small, it could likely be a fairly reliable library so that's why I think that might be an option. I'll start playing around with the idea of copying that code.

nathanielc:
@jsternberg Sounds good, an alternative is to have both copies of the code implement a common interface, so the type switch is on an interface instead of a concrete type, thus avoiding the need to share the type themselves.

jsternberg:
[...]

PR #1423, opened 2019-06-26
feat: add enriched error interface to flux

jsternberg:
We need a common way of categorizing and understanding errors returned
from InfluxDB. We model the errors based on the same concepts that
InfluxDB uses, but we create our own error codes that are modeled on the
grpc error codes.

Each native flux error will have a code that represents the category or
type of error and a message that gives detailed and human-readable
output about the error.

The errors library is included as an internal library and the error
struct itself is exposed using a type alias in the top level package.
This allows the errors to be used by packages that cannot depend on the
top level package (like semantic), but still have the error type be
exposed.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

Related to #1364.

### DIFF
--- a/codes/codes.go
+++ b/codes/codes.go
@@ -1,0 +1,229 @@
+/*
+ * This file has been modified for use in flux, but the original
+ * is part of the grpc-go project.
+ *
+ * The original can be located here: https://github.com/grpc/grpc-go/blob/master/codes/codes.go
+ *
+ * Copyright 2014 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package codes defines the error codes used by flux.
+package codes
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type Code uint32
+
+const (
+	// OK is returned on success.
+	OK Code = 0
+
+	// Canceled indicates the operation was canceled (typically by the caller).
+	Canceled Code = 1
+
+	// Unknown error. An example of where this error may be returned is
+	// if a Status value received from another address space belongs to
+	// an error-space that is not known in this address space. Also
+	// errors raised by APIs that do not return enough error information
+	// may be converted to this error.
+	Unknown Code = 2
+
+	// Invalid indicates the client made an invalid request.
+	// Note that this differs from FailedPrecondition. It indicates arguments
+	// that are missing or problematic regardless of the state of the system
+	// (e.g., a missing or invalid argument, a script error, etc).
+	Invalid Code = 3
+
+	// DeadlineExceeded means operation expired before completion.
+	// For operations that change the state of the system, this error may be
+	// returned even if the operation has completed successfully. For
+	// example, a successful response from a server could have been delayed
+	// long enough for the deadline to expire.
+	DeadlineExceeded Code = 4
+
+	// NotFound means some requested entity (e.g., bucket or database) was
+	// not found.
+	NotFound Code = 5
+
+	// AlreadyExists means an attempt to create an entity failed because one
+	// already exists.
+	AlreadyExists Code = 6
+
+	// PermissionDenied indicates the caller does not have permission to
+	// execute the specified operation. It must not be used for rejections
+	// caused by exhausting some resource (use ResourceExhausted
+	// instead for those errors). It must not be
+	// used if the caller cannot be identified (use Unauthenticated
+	// instead for those errors).
+	PermissionDenied Code = 7
+
+	// ResourceExhausted indicates some resource has been exhausted, perhaps
+	// a per-user quota, or perhaps the the memory allocation limit has been
+	// reached.
+	ResourceExhausted Code = 8
+
+	// FailedPrecondition indicates operation was rejected because the
+	// system is not in a state required for the operation's execution.
+	// For example, directory to be deleted may be non-empty, an rmdir
+	// operation is applied to a non-directory, etc.
+	//
+	// A litmus test that may help a service implementor in deciding
+	// between FailedPrecondition, Aborted, and Unavailable:
+	//  (a) Use Unavailable if the client can retry just the failing call.
+	//  (b) Use Aborted if the client should retry at a higher-level
+	//      (e.g., restarting a read-modify-write sequence).
+	//  (c) Use FailedPrecondition if the client should not retry until
+	//      the system state has been explicitly fixed. E.g., if an "rmdir"
+	//      fails because the directory is non-empty, FailedPrecondition
+	//      should be returned since the client should not retry unless
+	//      they have first fixed up the directory by deleting files from it.
+	//  (d) Use FailedPrecondition if the client performs conditional
+	//      REST Get/Update/Delete on a resource and the resource on the
+	//      server does not match the condition. E.g., conflicting
+	//      read-modify-write on the same resource.
+	FailedPrecondition Code = 9
+
+	// Aborted indicates the operation was aborted, typically due to a
+	// concurrency issue like sequencer check failures, transaction aborts,
+	// etc.
+	//
+	// See litmus test above for deciding between FailedPrecondition,
+	// Aborted, and Unavailable.
+	Aborted Code = 10
+
+	// OutOfRange means operation was attempted past the valid range.
+	// E.g., accessing a row or column outside of the table dimensions.
+	//
+	// Unlike Invalid, this error indicates a problem that may
+	// be fixed if the system state changes. For example, a 32-bit file
+	// system will generate Invalid if asked to read at an
+	// offset that is not in the range [0,2^32-1], but it will generate
+	// OutOfRange if asked to read from an offset past the current
+	// file size.
+	OutOfRange Code = 11
+
+	// Unimplemented indicates operation is not implemented or not
+	// supported/enabled in this service.
+	Unimplemented Code = 12
+
+	// Internal errors. Means some invariants expected by underlying
+	// system has been broken. If you see one of these errors,
+	// something is very broken.
+	//
+	// This should not be used when wrapping unidentified errors from
+	// other services. Unknown should be used instead when wrapping
+	// external errors.
+	Internal Code = 13
+
+	// Unavailable indicates the service is currently unavailable.
+	// This is a most likely a transient condition and may be corrected
+	// by retrying with a backoff. Note that it is not always safe to retry
+	// non-idempotent operations.
+	//
+	// See litmus test above for deciding between FailedPrecondition,
+	// Aborted, and Unavailable.
+	Unavailable Code = 14
+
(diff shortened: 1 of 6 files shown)


nathanielc:
It seems there are two error mechanisms in InfluxDB, the one you have copied here and this one  https://github.com/influxdata/influxdb/tree/master/kit/errors 

What is the difference?

errors.go
```diff
@@ -0,0 +1,244 @@
...
+	ENotFound            = "not found"
+	EConflict            = "conflict"             // action cannot be performed
+	EInvalid             = "invalid"              // validation failed
+	EUnprocessableEntity = "unprocessable entity" // data type is correct, but out of range
+	EEmptyValue          = "empty value"
+	EUnavailable         = "unavailable"
+	EForbidden           = "forbidden"
+	ETooManyRequests     = "too many requests"
+	EUnauthorized        = "unauthorized"
+	EMethodNotAllowed    = "method not allowed"
```
nathanielc: These error code seem very specific to HTTP and that does not apply well to Flux as the repo itself does not interact with HTTP APIs.

jsternberg:
@nathanielc the error library in `kit/errors` appears to be a first draft and is completely unused beyond accidental usage from `goimports`. I'll confirm that with @goller and then remove it from the repository.

errors.go
jsternberg: I agree on some of them. `EMethodNotAllowed` is a bit _too_ http specific. Forbidden and unauthorized are decently generic even if they map directly to HTTP. Too many requests is technically generic even if it is directly mapped to http.

One possible issue though is, a little by necessity, these need to be the same as influxdb. I guess we can remove the ones that aren't used or are more egregious like method not allowed, but I think that's a judgment call.

nathanielc: Why do we need to the same codes as influxdb? That is not clear to me.

Also it looks like the methods below have been implemented by using type assertions on anonymous interfaces so copying the types should also be safe. Not translation should be needed from errors originating in Flux vs errors originating in Influxdb.

jsternberg: It's because of serialization. If we don't have the same codes as influxdb, then they won't be able to read our error codes properly.

They'll do something like:

```go
if influxdb.ErrorCode(err) == influxdb.EForbidden {
    // ...
}
```

And even if we returned `EForbidden`, it won't recognize it. Additionally, JSON serialization, which is very possible that this will go through, will cause a `flux.Error` to get reconstructed as an `influxdb.Error` so it wouldn't be possible to differentiate the two by what type they are. We cannot make the assumption a `flux.Error` will stay a `flux.Error`.

nathanielc: I see, I had assumed that these error were meant to be opaque to the code, but looking through the influxdb code I see that is not the case. I think we should add a comment explicitly stating that since common Go error handling is that errors are opaque.

Is there a different approach we can take that doesn't tightly couple error handling in Flux with how InfluxDB does error handling?


As an aside:

> We cannot make the assumption a flux.Error will stay a flux.Error.

This is a good thing in my mind. It means the errors are defined by their behavior and not their representation.

jsternberg: I think it would be best to take the core of the errors library from influxdb and put it into its own separate and versioned package. The utility methods that influxdb has don't have to come along with it for simplicity and to ensure the interface doesn't change. That way we have no dependency on influxdb and we can type alias all of the functions we want or any new ones.

The error codes are supposed to be generic similar to HTTP (although the codes are not supposed to be HTTP) so they aren't supposed to be influxdb specific. There's a note about adding them to the swagger doc, but that could be changed to a go generate command that reads a file from a dependency in influxdb.

jsternberg: For clarity, I don't think using the same error handling would tightly couple the two. Once it is its own library, it stands on its own and it just happens to be that the two use the same method of reporting errors.

nathanielc: I see two ways this could work:

1. Couple the error handling in which case it makes sense to create a separate package that flux and influxdb both depend on. Otherwise the divergence would create a nightmare of bugs.
2. Decouple the error handling, meaning do not use the same error codes or types, but just take the same basic strategy. Flux can export its error codes and InfluxDB can make an explicit translation between them if needed. I would rather Flux errors be opaque to InfluxDB with a few explicit exceptions. This way packages consuming Flux (i.e. InfluxDB) can make decisions about what Flux error codes mean and if they will take action on them, otherwise they can treat them as opaque. The only current use case will be for Flux to report errors as either system errors or user errors, which InfluxDB will inspect to update appropriate statistics.

I prefer option 2 as it reduces the surface area of the Flux Go API. 

I have a few concerns over how option 1 could play out in confusing and negative ways. 

* Since InfluxDB and Flux will use the same error codes, the set of error codes will be the union of the two needs. In addition to the error code being part of the  InfluxDB public API this means that changes need to be strictly controlled and may not be possible to change in some cases. These two combined leads to a tight coupling that makes it hard to make changes to either project.
* All consumers of Flux as a library will need to make a decision about how to handle the error codes Flux can produce. As such that set of error codes should be as small as possible, including influxdb specific codes is confusing and error prone to consumers of Flux.
* This tight coupling can cause very hard to debug and unexpected issues. For example given that InfluxDB has logic that conditions on the error codes of various errors. And given that Flux produces errors with error codes that InfluxDB consumes. A change to errors either in Flux code or anywhere in the call stack between Flux and InfluxDB can trigger different logic in InfluxDB. These kinds of changes will have very indirect and hard to predict behaviors as the logic being affected is far away from the changes being made. 

Said another way, in general the practice of inspecting errors creates the possibility that changes can effect code that is loosely related (i.e. way up the call stack). As such I want to tightly control that surface area so it is more clear how said changes effect the code base as a whole.

jsternberg: Ok. I think this means the interfaces aren't really needed then because we would be saying the API is completely different so there's no expectation the values are the same. I think that means we'll need to add a method to influxdb so if it sees a `*flux.Error` it will translate it to an `*influxdb.Error`. I took another look at the source blog post for our system ([here](https://middlemost.com/failure-is-your-domain/) and that's what it said to do so I think that's a good solution.

jsternberg:
I have removed the extraneous error codes and removed the interfaces so this remains its own distinct error type and we will translate this error to an `*influxdb.Error` inside of influxdb.

I have removed all but three of the error types. I don't think this will be enough, but I'm not going to try and get the correct level of what we need versus generic error codes in this pull request.

errors.go
goller: > Since InfluxDB and Flux will use the same error codes, the set of error codes will be the union of the two needs. In addition to the error code being part of the InfluxDB public API this means that changes need to be strictly controlled and may not be possible to change in some cases. These two combined leads to a tight coupling that makes it hard to make changes to either project.

I don't think this will be a problem. Error codes are for groups of problems that inform consumers of the library what to do. Except the `method not available` I think each of the error codes in errors.go today are good groupings and each are applicable for flux. 

I think method not available should be removed and I'll make a PR to do so.

What groupings of errors do you imagine are not covered?  Flux is integral to influxdb and if influxdb does not support an error group flux needs, then, it absolutely should be added.

>    All consumers of Flux as a library will need to make a decision about how to handle the error codes Flux can produce. As such that set of error codes should be as small as possible, including influxdb specific codes is confusing and error prone to consumers of Flux.

I don't really see any error groupings that would not be appropriate for flux except method not allowed.

>    This tight coupling can cause very hard to debug and unexpected issues. For example given that InfluxDB has logic that conditions on the error codes of various errors. And given that Flux produces errors with error codes that InfluxDB consumes. A change to errors either in Flux code or anywhere in the call stack between Flux and InfluxDB can trigger different logic in InfluxDB. These kinds of changes will have very indirect and hard to predict behaviors as the logic being affected is far away from the changes being made.

Ultimately, the behavior that would change in influxdb from error codes would be the HTTP status code from /api/v2/query.  I think it is quite reasonable for flux to govern /api/v2/query

jsternberg:
We're going to discuss this in a meeting sometime next week to hash out the details and also how it integrates with go 1.13's handling of errors.

jsternberg:
@nathanielc I copied over the error codes from grpc. I filtered out the ones I didn't think made sense inside of flux and modified the style to match how we display error messages. I removed op and the extra code that dealt with recursive behavior and instead opted to just implement `Unwrap`. I removed the json serialization, but I plan to add that back.

codes/codes.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Unknown error. An example of where this error may be returned is
+	// if a Status value received from another address space belongs to
+	// an error-space that is not known in this address space. Also
+	// errors raised by APIs that do not return enough error information
+	// may be converted to this error.
+	Unknown Code = 2
+
+	// InvalidArgument indicates client specified an invalid argument.
+	// Note that this differs from FailedPrecondition. It indicates arguments
```
nathanielc: Several of these comments refer to `FailedPrecondition` but that is not one of the errors brought over. Can you update the comments or add `FailedPrecondition` back in?

codes/codes.go
```diff
@@ -0,0 +1,172 @@
...
+	Internal Code = 11
+
+	// Unavailable indicates the service is currently unavailable.
+	// This is a most likely a transient condition and may be corrected
+	// by retrying with a backoff. Note that it is not always safe to retry
+	// non-idempotent operations.
+	//
+	// See litmus test above for deciding between FailedPrecondition,
+	// Aborted, and Unavailable.
+	Unavailable Code = 12
```
nathanielc: I really like the distinction between Internal and Unavailable. It allows us to communicate that one of our dependencies is failing but otherwise our system is working as expected.

errors.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"strings"
+
+	"github.com/influxdata/flux/codes"
+)
+
+// Error is the error struct of flux.
+type Error struct {
+	Code codes.Code
+	Msg  string
```
nathanielc: Should we add a comment that we expect Code and Msg to always be set and that Err is optional and only needed to wrap an existing error?

Also we should state that only the `Code` field should be inspected by calling code.

jsternberg: Yea, I think so. Although message is also going to be optional. There may be situations where we can't add any additional context.

codes/codes.go
jsternberg: I'm going to look at `FailedPrecondition` again. It seems to be mentioned a lot and I didn't understand it at first. I'll update these comments regardless, but I want a better reason why I think `FailedPrecondition` shouldn't be there other than, "I didn't understand what it meant and it didn't seem to apply to any situations I could think of."

jsternberg:
I was trying this out on Wednesday to see how it would be used and I found that the errors don't seem to cover the generic categories nearly as well as I thought. In particular, I found that including the name of the error in the error text was mostly ineffective and created bad error messages. Here was how a type error code changed:

```
invalid argument: type error 3:1-3:23: invalid record access "_value": int != float
```

The `invalid argument` seemed the best choice based on our conversations, but the error message seemed like complete nonsense to me. I think it might be best not to include the code in the error message unless there is no message.

codes/codes.go
```diff
@@ -0,0 +1,229 @@
...
+	Canceled Code = 1
+
+	// Unknown error. An example of where this error may be returned is
+	// if a Status value received from another address space belongs to
+	// an error-space that is not known in this address space. Also
+	// errors raised by APIs that do not return enough error information
+	// may be converted to this error.
+	Unknown Code = 2
+
+	// InvalidArgument indicates client specified an invalid argument.
```
jsternberg: This name has made no sense while I have been playing around with reporting errors. I'm renaming it to `Invalid` to indicate that the client gave us something that was not valid rather than the specific "invalid argument" which doesn't make sense in contexts where there is no argument or the argument is missing.

jsternberg:
I added tests for all of the errors stuff and also added a go 1.13 specific test which passes when using go 1.13, but doesn't run with go 1.12.

Issue #1317, opened 2019-05-31
Construct tests for the Table interface


jsternberg:
The table interface has been defined in #1281 and we need to construct tests that use the table to ensure that all of the tables are compliant. It is not required that every Table interface use the test, but all of the interfaces used by the core flux should to ensure that users of the interface are operating correctly.

This is because we had an issue where the influxdb version of a Table interface only allowed `Do` to be called once, but not a single implementation of this was included in flux so this constraint was never tested. If flux had defined and followed the interface of calling `Do` only once, this issue would have never happened in the dependent repository.

These tests should contain tests for the following:

* Call `Do` once.
* Call `Do` followed by `Done` once.
* Multiple `Do` calls should have the second call onwards fail.
* Calling `Done` alone should work.
* Multiple calls to `Done` should not cause a problem.
* Call `Retain` on the column reader from `Do` and test that it is still valid.
* Copy table should work when used on the table.
* The retained column reader should still be valid even after multiple calls to `Done` on the original table (ensures that `Done` is idempotent).
* Ensure all memory is freed for all of the above.
* When `Do` is called, memory should be consumed as a test to make sure the above isn't a false positive.
* Test that the column reader is not modified after it is retained (the underlying table implementation recognizes that the column reader is still being used and doesn't reuse it).
* The returned `ColReader` returns the correct value for `Len()`.

jsternberg:
Fixed by #1428.

PR #1428, opened 2019-06-28
tests(execute): add an initial version of table tests

jsternberg:
The table tests are a common set of tests that should test the basic
functionality of the table interface because the interface itself can be
tricky to get right. Any valid implementation of table should pass these
tests and the tests also act as an example of how to use the table
properly.

Implements part of #1317.

### DIFF
--- a/execute/executetest/table.go
+++ b/execute/executetest/table.go
@@ -1,14 +1,21 @@
 package executetest
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"testing"
+	"time"
 
 	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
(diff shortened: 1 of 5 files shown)


execute/executetest/table.go
```diff
@@ -470,3 +488,287 @@ func MustCopyTable(tbl flux.Table) flux.Table {
...
+		// The same buffer contents with different memory addresses
+		// are considered different.
+		switch a.Cols()[i].Type {
+		case flux.TFloat:
+			if a.Floats(i) != b.Floats(i) {
+				return false
+			}
+		case flux.TInt:
+			if a.Ints(i) != b.Ints(i) {
+				return false
```
affo: Don't you want to be sure that they don't share any buffer? In this  way, it is enough not to share a single buffer and share the others and the readers are not equal.

jsternberg: I don't think that's necessary. I check earlier to see if the metadata is the same so we're certain that we're comparing the same columns.

This is mostly a check to make sure that any buffering mechanism for column readers, like the one influxdb employs, doesn't end up modifying the old column reader when it is retained since there's specific code exercising that.

affo: Mmmm, I still think that it should be better to check for every col reader to be different. Maybe I am missing something 😅

PR #1430, opened 2019-06-28
feat(stdlib/universe): moving average function

roshie548:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1723,6 +1723,33 @@ from(bucket: "telegraf/autogen")
 	|> median()
 ```
 
+##### Moving Average
+
+Moving Average is an aggregate operation.
+For each aggregated column, it means the values of the following records for a defined window range.
+Moving Average is defined as
+```
+movingAverage = (every, period, column="_value", tables=<-) =>
+    tables
+        |> window(every: every, period: period)
+        |> mean(column:column)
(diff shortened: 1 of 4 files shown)


docs/SPEC.md
```diff
@@ -1723,6 +1723,34 @@ from(bucket: "telegraf/autogen")
 	|> median()
 ```
 
+##### Moving Average
+
+Moving Average is an aggregate operation.
+For each aggregated column, it means the values of the following records for a defined window range.
+Moving Average is defined as
+```
+movingAverage = (every, period, column="_value", timeSrc="_stop",timeDst="_time", tables=<-) =>
```
jlapacik: Are the `timeSrc` and `timeDst` parameters necessary? By default the output time column is `_stop` for all aggregates, and since the aggregate, in this case mean, is encapsulated within the function declaration, maybe these parameters shouldn't be configurable.

Also the `column` parameter is not used in the `mean()` call.

roshie548:
> Should this be:
> 
> ```
> |> movingAverage(every: 1y, period: 5y)
> ```

yes, thanks for catching that!

docs/SPEC.md
jlapacik: Can you update the SPEC with the new signature?

Issue #1408, opened 2019-06-24
mode transformation for flux.


aanthony1243:
similar to influxql mode function

nathanielc:
See #195

PR #1431, opened 2019-06-28
feat(stdlib/universe): added mode() function

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/mode.go
+++ b/stdlib/universe/mode.go
@@ -1,0 +1,630 @@
+package universe
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+const ModeKind = "mode"
+
+type ModeOpSpec struct {
+	Column string `json:"column"`
+}
+
+func init() {
+	modeSignature := flux.FunctionSignature(
+		map[string]semantic.PolyType{
+			"column": semantic.String,
+		},
+		nil,
+	)
+
+	flux.RegisterPackageValue("universe", ModeKind, flux.FunctionValue(ModeKind, createModeOpSpec, modeSignature))
+	flux.RegisterOpSpec(ModeKind, newModeOp)
+	plan.RegisterProcedureSpec(ModeKind, newModeProcedure, ModeKind)
+	execute.RegisterTransformation(ModeKind, createModeTransformation)
+}
+
+func createModeOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
+	if err := a.AddParentFromArgs(args); err != nil {
+		return nil, err
+	}
+
+	spec := new(ModeOpSpec)
+
+	if col, ok, err := args.GetString("column"); err != nil {
+		return nil, err
+	} else if ok {
+		spec.Column = col
+	} else {
+		spec.Column = execute.DefaultValueColLabel
+	}
+	return spec, nil
+}
+
+func newModeOp() flux.OperationSpec {
+	return new(ModeOpSpec)
+}
+
+func (s *ModeOpSpec) Kind() flux.OperationKind {
+	return ModeKind
+}
+
+type ModeProcedureSpec struct {
+	plan.DefaultCost
+	Column string
+}
+
+func newModeProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.ProcedureSpec, error) {
+	spec, ok := qs.(*ModeOpSpec)
+	if !ok {
+		return nil, fmt.Errorf("invalid spec type %T", qs)
+	}
+
+	return &ModeProcedureSpec{
+		Column: spec.Column,
+	}, nil
+}
+
+func (s *ModeProcedureSpec) Kind() plan.ProcedureKind {
+	return ModeKind
+}
+func (s *ModeProcedureSpec) Copy() plan.ProcedureSpec {
+	ns := new(ModeProcedureSpec)
+
+	*ns = *s
+
+	return ns
+}
+
+// TriggerSpec implements plan.TriggerAwareProcedureSpec
+func (s *ModeProcedureSpec) TriggerSpec() plan.TriggerSpec {
+	return plan.NarrowTransformationTriggerSpec{}
+}
+
+func createModeTransformation(id execute.DatasetID, mode execute.AccumulationMode, spec plan.ProcedureSpec, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
+	s, ok := spec.(*ModeProcedureSpec)
+	if !ok {
+		return nil, nil, fmt.Errorf("invalid spec type %T", spec)
+	}
+	cache := execute.NewTableBuilderCache(a.Allocator())
+	d := execute.NewDataset(id, mode, cache)
+	t := NewModeTransformation(d, cache, s)
+	return t, d, nil
+}
+
+type modeTransformation struct {
+	d     execute.Dataset
+	cache execute.TableBuilderCache
+
+	column string
+}
+
+func NewModeTransformation(d execute.Dataset, cache execute.TableBuilderCache, spec *ModeProcedureSpec) *modeTransformation {
+	return &modeTransformation{
+		d:      d,
+		cache:  cache,
+		column: spec.Column,
+	}
+}
+
+func (t *modeTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey) error {
+	return t.d.RetractTable(key)
+}
+
+func (t *modeTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	builder, created := t.cache.TableBuilder(tbl.Key())
+	if !created {
+		return fmt.Errorf("mode found duplicate table with key: %v", tbl.Key())
+	}
+
+	colIdx := execute.ColIdx(t.column, tbl.Cols())
+	if colIdx < 0 {
+		// doesn't exist in this table, so add an empty value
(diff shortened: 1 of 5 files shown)


sauren-khosla:
absent benchmark tests

stdlib/universe/mode.go
```diff
@@ -0,0 +1,519 @@
+package universe
+
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/values"
```
jsternberg: Can you ensure the imports are grouped correctly? Standard library imports are grouped together and all other imports are grouped together.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,519 @@
...
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+)
+
+const ModeKind = "mode"
+
+type ModeOpSpec struct {
+	Column string `json:"column"`
```
jsternberg: If I understand correctly, mode is supposed to be an aggregate which means it should match with the same behavior as the other aggregates. We have `execute.AggregateConfig` and other related functions in `execute/aggregate.go` that handle some of the extra portions of this scenario.

I think that pattern would be better for this scenario. Can you take a look at things like `stdlib/universe/count.go` and try to match that pattern? I also think there's a reasonable chance it would resolve some of my other comments.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,519 @@
...
+		if err := execute.AppendKeyValues(tbl.Key(), builder); err != nil {
+			return err
+		}
+		// TODO: hack required to ensure data flows downstream
+		return tbl.Do(func(flux.ColReader) error {
+			return nil
+		})
+	}
+
+	var (
```
jsternberg: Instead of defining all of the various types in the outermost portion, I think it would be good if we just determined the column type and then went to a function to deal with that column type instead of having to type switch on the same type repeatedly. I could go into more detail, but the aggregate interface spells it out better than I think I would be able to in a short paragraph. Could you take a look at that?

stdlib/universe/mode_test.go
```diff
@@ -0,0 +1,309 @@
...
+import (
+	"testing"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/stdlib/universe"
+)
+
+func TestMode_Process(t *testing.T) {
```
jsternberg: Can you ensure there's a test case for each of the types that mode can process? It looks like the test case only covers strings right now.

stdlib/universe/mode.go
sauren-khosla: followed distinct.go here because we ended up deciding on a custom OpSpec vs. aggregate

stdlib/universe/mode.go
sauren-khosla: see below conversation

stdlib/universe/mode.go
jsternberg: I think regardless of whether or not this uses the aggregate functions, it should still be using a similar pattern. The method with distinct is very difficult to read so I would like to have a struct definition for each of the types that takes the same signature as aggregate and uses the same method, regardless of whether it uses the exact code because of returning multiple values.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,584 @@
...
+	colIdx, err := builder.AddCol(flux.ColMeta{
+		Label: execute.DefaultValueColLabel,
+		Type:  col.Type,
+	})
+	if err != nil {
+		return err
+	}
+
+	if tbl.Key().HasCol(t.column) {
+		j := execute.ColIdx(t.column, tbl.Key().Cols())
```
jsternberg: I think this entire section can be replaced with using `Value(j)` and then using `AppendValue` on the builder.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,584 @@
...
+			return err
+		}
+		// TODO: hack required to ensure data flows downstream
+		return tbl.Do(func(flux.ColReader) error {
+			return nil
+		})
+	}
+
+	var (
+		numNil     int64
```
jsternberg: This section should no longer be necessary. You should be able to now just create these references in the switch statement.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,584 @@
...
+		uintMode   map[uint64]int64
+		floatMode  map[float64]int64
+		stringMode map[string]int64
+		timeMode   map[execute.Time]int64
+	)
+
+	j := execute.ColIdx(t.column, tbl.Cols())
+	switch col.Type {
+	case flux.TBool:
+		boolMode = make(map[bool]int64)
```
jsternberg: Can you refactor the logic in this to a function and then invoke the function?

stdlib/universe/mode.go
```diff
@@ -0,0 +1,584 @@
...
+			l := cr.Len()
+			for i := 0; i < l; i++ {
+				if cr.Strings(j).IsNull(i) {
+					numNil++
+					continue
+				}
+				v := cr.Strings(j).ValueString(i)
+				stringMode[v]++
+			}
+			storedVals := []string{}
```
jsternberg: I think we can simplify the logic below since it seems to be fairly complex. Instead of playing weird tricks with a byte slice, we can feasibly just iterate over the map twice. The first time we iterate we are looking for the number of times a value has been seen and the number of values that match that frequency and the second time we construct the stored values. It would also be possible to optimize that second pass in certain circumstances.

The logical steps:

1. Iterate over the mode frequencies.
2. If a key has been seen more frequently than another key, then store the new key.
3. If a key has been seen with the same frequency as another key, increment the number of total values.

Here is some pseudocode.

```go
curr, n, total := "", int64(0)
for k := range stringMode {
    if stringMode[k] > n {
        curr, n, total = k, stringMode[k], 1
    } else if stringMode[k] == n {
        total++
    }
}

storedVals := make([]string, 0, len(total))
for k := range stringMode {
    if stringMode[k] == n {
        storedVals = append(storedVals, k)
    }
}
```

The variable names are bad so it might be a good idea to rename those.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,668 @@
...
+		})
+	}
+
+	var (
+		numNil int64
+	)
+
+	switch col.Type {
+	case flux.TBool:
+		boolMode := make(map[bool]int64)
```
jsternberg: It should be possible to move `boolMode` and `numNil` into the `doBool` function so you can pass it as a table instead.

```go
func (t *modeTransformation) doBool(tbl flux.Table, builder execute.TableBuilder, idx int) {
    var nils int64
    mode := make(map[bool]int64)
    if err := tbl.Do(func(cr flux.ColReader) error {
        // implementation
    }); err != nil {
        return err
    }
    // rest of the implementation
}

switch col.Type {
case flux.TBool:
    if err := t.doBool(tbl, builder, colIdx); err != nil {
        return err
    }
}
```

stdlib/universe/mode.go
```diff
@@ -0,0 +1,668 @@
...
+		return tbl.Do(func(cr flux.ColReader) error {
+			l := cr.Len()
+			err := t.doBool(cr, tbl, boolMode, l, numNil, builder, colIdx)
```
jsternberg: In Go, errors should typically be done like this:

```go
if err := myfunc(); err != nil {
    // handle error
}
```

Or they can be directly returned. Rarely, we need the error outside of the temporary scope created by if or there's another return parameter that we need.

In this specific scenario, you can just return directly. The if condition on line 193 doesn't change anything.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,668 @@
...
+
+	if err := execute.AppendKeyValues(tbl.Key(), builder); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (t *modeTransformation) doBool(cr flux.ColReader, tbl flux.Table, boolMode map[bool]int64, l int, numNil int64, builder execute.TableBuilder, colIdx int) error {
+	j := execute.ColIdx(t.column, tbl.Cols())
```
jsternberg: I think you can just use `colIdx` because the column has already been determined.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,668 @@
...
+
+	storedVals := make([]bool, 0, total)
+	for val := range boolMode {
+		if boolMode[val] == max {
+			storedVals = append(storedVals, val)
+		}
+	}
+
+	// if more nils than the max occurrences of another value, mode is nil
+	if numNil > max {
```
jsternberg: It looks like this should be above the creation of `storedVals`.

stdlib/universe/mode.go
sauren-khosla: I get a type conversion panic if I set j to be colIdx.

jsternberg: It looks like the reason is because you use `colIdx` twice above so you lose the initial assignment. I think you should resolve that so that you can pass both the source column and the output column rather than using `ColIdx` to determine the column again here.

stdlib/universe/mode.go
```diff
@@ -0,0 +1,643 @@
...
+	}
+
+	if err := execute.AppendKeyValues(tbl.Key(), builder); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (t *modeTransformation) doString(cr flux.ColReader, tbl flux.Table, builder execute.TableBuilder, oldIdx int, colIdx int) error {
```
jsternberg: In Go, if you have repeated arguments with the same type, you can do `oldIdx, colIdx int`. Can you also rename these to be `srcColIdx` and `destColIdx`?

timhallinflux:
closes #1408

PR #1440, opened 2019-07-02
tests(stdlib): fix the e2e benchmarks

jsternberg:
It needs to copy the package before adding the testing calls because
otherwise it adds the tests multiple times.

Ideally, this would be done before the benchmark, but a bit of
refactoring has to be done to separate that.

- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -305,11 +305,14 @@ func (p *AstProgram) Start(ctx context.Context, alloc *memory.Allocator) (flux.Q
 		p.Now = time.Now()
 	}
 
+	astPkg := p.Ast
 	if p.opts.extern != nil {
-		p.Ast.Files = append([]*ast.File{p.opts.extern}, p.Ast.Files...)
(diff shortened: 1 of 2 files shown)


lang/compiler.go
```diff
@@ -106,7 +106,7 @@ func CompileAST(astPkg *ast.Package, now time.Time, opts ...CompileOption) *AstP
 		Program: &Program{
 			opts: applyOptions(opts...),
 		},
-		Ast: astPkg,
+		Ast: astPkg.Copy().(*ast.Package),
```
wolffcm: Why do we need to copy here?  I would have thought that the AST is read-only from here.

jsternberg: Your changes to the `AstProgram` means it can modify the package struct now.

jsternberg: I modified this back so the copy happens lazily in the `Start` method only when it is modifying the package that way we aren't getting hit with a potentially unnecessary copy.

PR #1442, opened 2019-07-02
fix(stdlib): readd mergeKey parameter to map in deprecated state

nathanielc:
The transpiler relies heavily on the mergeKey behavior. As such we are
adding it back. It will remain undocumented and will default to false.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
@jsternberg Done, rebasing into maint.

PR #1443, opened 2019-07-02
feat(stdlib/universe): added go movingAverage

roshie548:
Added the go version of `movingAverage`, which follows the spec used by InfluxQL. It means over `n` points.
The pure flux implementation is changed to `timedMovingAverage` for the time being, since it means over a time window.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1723,34 +1723,6 @@ from(bucket: "telegraf/autogen")
 	|> median()
 ```
 
-##### Moving Average
-
-Moving Average is an aggregate operation.
-For each aggregated column, it means the values of the following records for a defined window range.
-Moving Average is defined as
-```
-movingAverage = (every, period, column="_value", tables=<-) =>
-    tables
-        |> window(every: every, period: period)
-        |> mean(column:column)
-        |> duplicate(column: "_stop", as: "_time")
-        |> window(every: inf)
-```
-Moving Average has the following properties:
-| Name        | Type     | Description
-| ----        | ----     | -----------
-| every       | duration | Every specifies the frequency of windows.
-| period      | duration | Period specifies the window size to mean.       
-| column      | string   | Column specifies a column to aggregate. Defaults to `"_value"`            
-
-Example:
-```
-// A 5 year moving average would be called as such:
-from(bucket: "telegraf/autogen"):
-    |> range(start: -7y)
-    |> movingAverage(every: 1y, period: 5y)
-```
-
 ##### Mode
 
 Mode produces the mode for a given column. Null is considered as a potential mode if it is present. If there are multiple modes, all of them are returned in a table in sorted order. 
@@ -3117,6 +3089,75 @@ Given the following input table.
     | 00002 | 4      |
     | 00003 | 7      |
     | 00004 | 8      |
+    
+    
+####  Moving Average
+
+Moving Average computes the moving averages of a series of records.
+It means the values of a user-defined period for a defined number of points,
+
+Moving Average has the following properties:
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| n           | duration | N specifies the number of points to mean.       
+| columns     | []string | Columns is list of all columns that `movingAverage` should be performed on. Defaults to `["_value"]`. |
+
+Rules for taking the moving average for numeric types:
+ - the average over a period populated by `n` values is equal to their algebraic mean
+ - the average over a period populated by only null values is null
+ - moving averages that include null values skip over those values
+ - if `n` is less than the number of records in a table, `movingAverage` returns the average of the available values
+ 
+Example of moving average (`N` = 2):
+
+| _time |   A  |   B  |   C  |   D  | tag |
+|:-----:|:----:|:----:|:----:|:----:|:---:|
+|  0001 | null |   1  |   2  | null |  tv |
+|  0002 |   6  |   2  | null | null |  tv |
+|  0003 |   4  | null |   4  |   4  |  tv |
+
+Result:
+
+| _time |   A  |   B  |   C  |   D  | tag |
+|:-----:|:----:|:----:|:----:|:----:|:---:|
+|  0002 |   6  |  1.5 |   2  | null |  tv |
+|  0003 |   5  |   2  |   4  |   4  |  tv |
+
+Example of script:
+```
+// A 5 point moving average would be called as such:
+from(bucket: "telegraf/autogen"):
+    |> range(start: -7y)
+    |> movingAverage(n: 5, columns: ["_value"])
+```
+
+##### Timed Moving Average
+
+Timed Moving Average means the values of the following records for a defined number of points, for the specified column.
+```
+timedMovingAverage = (every, period, column="_value", tables=<-) =>
+    tables
+        |> window(every: every, period: period)
+        |> mean(column:column)
+        |> duplicate(column: "_stop", as: "_time")
+        |> window(every: inf)
+```
+Timed Moving Average has the following properties:
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| every       | duration | Every specifies the frequency of windows.
+| period      | duration | Period specifies the window size to mean.       
+| column      | string   | Column specifies a column to aggregate. Defaults to `"_value"`            
+
+Example:
+```
+// A 5 year moving average would be called as such:
+from(bucket: "telegraf/autogen"):
+    |> range(start: -7y)
+    |> timedMovingAverage(every: 1y, period: 5y)
+```
 
 #### Distinct
 
--- a/stdlib/testing/testdata/timed_moving_average.flux
+++ b/stdlib/testing/testdata/timed_moving_average.flux
@@ -1,0 +1,41 @@
+package testdata_test
+
+import "testing"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,double,string,string,string,string,string,string
+#group,false,false,false,false,true,true,true,true,true,true
+#default,_result,,,,,,,,,
+,result,table,_time,_value,_field,_measurement,device,fstype,host,path
+,,0,2018-05-22T00:00:00Z,30,used_percent,disk,disk1s1,apfs,host.local,/
+,,0,2018-05-22T00:00:10Z,30,used_percent,disk,disk1s1,apfs,host.local,/
+,,0,2018-05-22T00:00:20Z,30,used_percent,disk,disk1s1,apfs,host.local,/
+,,0,2018-05-22T00:00:30Z,40,used_percent,disk,disk1s1,apfs,host.local,/
+,,0,2018-05-22T00:00:40Z,40,used_percent,disk,disk1s1,apfs,host.local,/
+,,0,2018-05-22T00:00:50Z,40,used_percent,disk,disk1s1,apfs,host.local,/
+"
+
+outData = "
+#group,false,false,true,true,true,true,true,true,true,true,false,false
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,string,string,string,string,string,string,double,dateTime:RFC3339
+#default,_result,,,,,,,,,,,
+,result,table,_start,_stop,_field,_measurement,device,fstype,host,path,_value,_time
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:10Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:20Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:30Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,33.333333333333336,2018-05-22T00:00:40Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,36.666666666666664,2018-05-22T00:00:50Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
+"
+timed_moving_average = (table=<-) =>
+	(table
+		|> range(start: 2018-05-22T00:00:00Z, stop: 2018-05-22T00:01:00Z)
+		|> timedMovingAverage(every: 10s, period: 30s))
+
+test _timed_moving_average = () =>
+	({input: testing.loadStorage(csv: inData), want: testing.loadMem(csv: outData), fn: timed_moving_average})
+
--- a/stdlib/universe/moving_average.go
+++ b/stdlib/universe/moving_average.go
@@ -1,0 +1,557 @@
+package universe
+
+import (
+	"fmt"
+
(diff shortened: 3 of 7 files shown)


stdlib/universe/typeconv_test.go
```diff
@@ -342,29 +341,29 @@ func TestTypeconv_Float(t *testing.T) {
...
 			name: "float64(v:8)",
-			v: "-Inf",
+			v:    "-Inf",
 			want: float64(math.Inf(-1)),
 		},
 		{
-			name: "float64(v:8)",
-			v: "NaN",
-			want: float64(math.NaN()),
+			name:      "float64(v:8)",
```
affo: Are these gofmt results?

stdlib/sql/to_test.go
```diff
@@ -59,10 +59,10 @@ func TestToSQL_Process(t *testing.T) {
 	dsn := "root@/db"
 	_, _, _ = sqlmock.NewWithDSN(dsn)
 	type wanted struct {
-		Table  []*executetest.Table
-		ColumnNames []string
+		Table        []*executetest.Table
+		ColumnNames  []string
 		ValueStrings [][]string
-		ValueArgs [][]interface{}
```
affo: Are these gofmt results?

docs/SPEC.md
```diff
@@ -1723,20 +1723,40 @@ from(bucket: "telegraf/autogen")
 	|> median()
 ```
 
-##### Moving Average
+#####  Moving Average
 
 Moving Average is an aggregate operation.
-For each aggregated column, it means the values of the following records for a defined window range.
+For each aggregated column, it means the values of the following records for a defined number of points.
```
affo: Can you change this to 

```
Moving Average is an aggregate operation.
It means the values of the following records for a defined number of points, for the specified aggregated column.
```

There is  only one column on which it operates, right?

stdlib/testing/testdata/timed_moving_average.flux
```diff
@@ -0,0 +1,41 @@
...
+#default,_result,,,,,,,,,,,
+,result,table,_start,_stop,_field,_measurement,device,fstype,host,path,_value,_time
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:10Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:20Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,30,2018-05-22T00:00:30Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,33.333333333333336,2018-05-22T00:00:40Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,36.666666666666664,2018-05-22T00:00:50Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,40,2018-05-22T00:01:00Z
```
affo: This is tricky, I would expect `_time` to be:

```
2018-05-22T00:00:10Z
2018-05-22T00:00:20Z
2018-05-22T00:00:30Z
2018-05-22T00:00:40Z
2018-05-22T00:00:50Z
2018-05-22T00:01:00Z
2018-05-22T00:01:10Z
2018-05-22T00:01:20Z
```

Can you comment on this?

I imagine no record can have `_time` > `_stop`, and this is due to `window` implementation.

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,585 @@
+package universe
+
+import (
+	"container/list"
+	"fmt"
```
affo: Can you separate `github.com` import from others with a newline?

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,585 @@
...
+
+func (s *MovingAverageOpSpec) Kind() flux.OperationKind {
+	return MovingAverageKind
+}
+
+type MovingAverageProcedureSpec struct {
+	plan.DefaultCost
+	N          int64    `json:"n"`
+	Columns    []string `json:"columns"`
+	TimeColumn string   `json:"timeColumn"`
```
affo: Is this useful?

stdlib/testing/testdata/timed_moving_average.flux
roshie548: Nathaniel, Adam, Jonathan and I had a discussion about this yesterday. The issue is that windows are clipped to match the given bounds, so we end up with incomplete windows. It's more of a problem with the windows function itself.

docs/SPEC.md
```diff
@@ -3099,6 +3052,74 @@ Given the following input table.
...
+Moving Average has the following properties:
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| n           | duration | N specifies the number of points to mean.       
+| column      | string   | Column specifies a column to aggregate. Defaults to `"_value"`
+
+Rules for taking the moving average for numeric types:
+ - the average over a period populated by non-null types is equal to their algebraic mean;
+ - the moving average at a null point is null;
```
affo: This is a bit counter intuitive. Why?

stdlib/universe/moving_average.go
```diff
@@ -72,9 +70,8 @@ func (s *MovingAverageOpSpec) Kind() flux.OperationKind {
 
 type MovingAverageProcedureSpec struct {
 	plan.DefaultCost
-	N          int64    `json:"n"`
-	Columns    []string `json:"columns"`
-	TimeColumn string   `json:"timeColumn"`
+	N       int64    `json:"n"`
+	Columns []string `json:"columns"`
```
affo: Did we decide for columns or column?

docs/SPEC.md
```diff
@@ -3099,6 +3052,74 @@ Given the following input table.
...
+ - moving averages at non-null points that have nulls in their scope skip over those nulls
+ 
+Example of moving average:
```
affo: On how many numbers? What is `N`?

stdlib/universe/moving_average.go
```diff
@@ -237,9 +231,6 @@ func (t *movingAverageTransformation) passThroughBool(ts *array.Int64, vs *array
 	l := vs.Len()
 	for ; i < l; i++ {
 		cTime := execute.Time(ts.Value(i))
-		if cTime < pTime {
```
affo: Could you please remove every check on time? Also below.

stdlib/universe/moving_average.go
```diff
@@ -225,8 +221,6 @@ func (t *movingAverageTransformation) Finish(id execute.DatasetID, err error) {
 	t.d.Finish(err)
 }
 
-const movingAverageUnsortedTimeErr = "moving average found out-of-order times in time column"
-
 func (t *movingAverageTransformation) passThroughBool(ts *array.Int64, vs *array.Boolean, b execute.TableBuilder, bj int) error {
```
affo: I would say  that you don't  need this function anymore right? At this point you can simply remove the first `N - 1` records and append the rest of the column.

affo: (Also for string and time)

stdlib/universe/moving_average.go
affo: Please, address this.

stdlib/universe/moving_average.go
affo: I think this should contrast with the `timed` version.
Why should that have `column` and this one `column`?
We will bring this into discussion with the team.

stdlib/universe/moving_average.go
roshie548: I think this is an outdated file

docs/SPEC.md
```diff
@@ -3079,6 +3052,73 @@ Given the following input table.
     | 00002 | 4      |
     | 00003 | 7      |
     | 00004 | 8      |
+    
+    
+####  Moving Average
+
+Moving Average computes the moving averages of a series of records
```
affo: `.` missing at the end of period

docs/SPEC.md
affo: You should specify what is `N` for the example. I suppose it's 2.

stdlib/universe/moving_average_test.go
```diff
@@ -0,0 +1,521 @@
...
+					{execute.Time(6 * time.Second), 28.0 / 3},
+					{execute.Time(7 * time.Second), 34.0 / 3},
+					{execute.Time(8 * time.Second), 38.0 / 3},
+					{execute.Time(9 * time.Second), 32.0 / 3},
+					{execute.Time(10 * time.Second), 8.0},
+				},
+			}},
+		},
+		{
+			name: "int",
```
affo: this seems a repetition

stdlib/universe/moving_average_test.go
```diff
@@ -0,0 +1,521 @@
...
+					{execute.Time(6), 9.5},
+					{execute.Time(7), 13.0},
+					{execute.Time(8), 13.5},
+					{execute.Time(9), 8.5},
+					{execute.Time(10), 6.0},
+				},
+			}},
+		},
+		{
+			name: "int with 3",
```
affo: this one too

stdlib/universe/moving_average_test.go
```diff
@@ -0,0 +1,521 @@
...
+			}},
+			want: []*executetest.Table{{
+				ColMeta: []flux.ColMeta{
+					{Label: "_time", Type: flux.TTime},
+					{Label: "x", Type: flux.TFloat},
+					{Label: "y", Type: flux.TFloat},
+					{Label: "z", Type: flux.TFloat},
+				},
+				Data: [][]interface{}{
+					{execute.Time(2), 0.0, 1.5, 2.0},
```
affo: The first `0.0` is still surprising. I would expect that the avg of `nil nil` is `nil`.

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,444 @@
...
+			}
+			doMovingAverage[j] = true
+		} else {
+			_, err := builder.AddCol(c)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	if timeIdx < 0 {
```
affo: You don't need any time column. This check is useless.

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,444 @@
...
+	return nil
+}
+
+func (t *movingAverageTransformation) doInt(ts *array.Int64, vs *array.Int64, b execute.TableBuilder, bj int, doMovingAverage bool) error {
+	i := 0
+	window := make([]interface{}, t.n)
+	sum := int64(0)
+	count := 0
+
+	if vs.IsValid(i) {
```
affo: What if the table is empty? This could result in an out of bounds exception.

Could you add a test for it?

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,444 @@
...
+					return err
+				}
+			} else {
+				if err := b.AppendNil(bj); err != nil {
+					return err
+				}
+			}
+		} else {
+			average := 0.0
+			if count != 0 {
```
affo: If count == 0, you should append a nil.

This means the whole window is nil.

affo: Default to 0.0 seems a wrong option to me.

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,444 @@
...
+				return err
+			}
+		} else {
+			if err := b.AppendNil(bj); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
affo: This can be done like this:

```go
func (t *movingAverageTransformation) passThroughBool(vs *array.Boolean, b execute.TableBuilder, bj int) error {
	s := arrow.BoolSlice(vs, int(t.n-1), vs.Len())
	defer s.Release()
	return b.AppendBools(bj, s)
}
```

Note that `ts` is useless in the method signature.

stdlib/universe/moving_average.go
roshie548: This works. An empty table will return in the `Process()` method without these functions every being called

stdlib/universe/moving_average.go
aanthony1243: did we add the newline here?

stdlib/universe/moving_average.go
aanthony1243: I agree, why columns instead of column?

stdlib/universe/moving_average.go
```diff
@@ -338,49 +378,60 @@ func (t *movingAverageTransformation) doUInt(vs *array.Uint64, b execute.TableBu
 					return err
 				}
 			}
-			next := window[int64(i+1)%t.n]
+
+			next := t.window[bj][int64(t.i[bj]+1)%t.n]
 			if next != nil {
-				sum -= next.(uint64)
-				count--
+				*sum -= next.(uint64)
```
aanthony1243: i'm a little surprised that this works since `next` is technically a values.Value  type, right?

roshie548: `next` is either int64, uint64. or float64:
`func (a *Int64) Value(i int) int64    { return a.values[i] }`

stdlib/universe/moving_average.go
```diff
@@ -217,44 +227,59 @@ func (t *movingAverageTransformation) Finish(id execute.DatasetID, err error) {
 }
 
 func (t *movingAverageTransformation) passThroughBool(vs *array.Boolean, b execute.TableBuilder, bj int) error {
-	s := arrow.BoolSlice(vs, int(t.n-1), vs.Len())
-	defer s.Release()
-	return b.AppendBools(bj, s)
+	if int64(vs.Len()) < t.n {
+		if err := b.AppendBool(bj, vs.Value(vs.Len()-1)); err != nil {
```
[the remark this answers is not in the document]
affo: yep, that's the corner case in which he has less values than the "period". When passing through he appends only the last value.

stdlib/universe/moving_average.go
```diff
@@ -124,6 +124,11 @@ type movingAverageTransformation struct {
 
 	n       int64
 	columns []string
+
+	i      []int
+	sum    []interface{}
```
affo: Can you comment on why you are using global slices?

docs/SPEC.md
```diff
@@ -3079,6 +3052,75 @@ Given the following input table.
...
+
+| _time |   A  |   B  |   C  |   D  | tag |
+|:-----:|:----:|:----:|:----:|:----:|:---:|
+|  0001 | null |   1  |   2  | null |  tv |
+|  0002 |   6  |   2  | null | null |  tv |
+|  0003 |   4  | null |   4  |   4  |  tv |
+
+Result:
+
+| _time |   A  |   B  |   C  |   C  | tag |
```
affo: C -> D

stdlib/universe/moving_average.go
roshie548: Since tables may be split into chunks, we want to be able to pick up where we left off when moving to a new table.

docs/SPEC.md
```diff
@@ -3117,6 +3089,75 @@ Given the following input table.
...
+####  Moving Average
+
+Moving Average computes the moving averages of a series of records.
+It means the values of a user-defined period for a defined number of points,
+
+Moving Average has the following properties:
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| n           | duration | N specifies the number of points to mean.
```
sanderson: `n`'s type should be `int`.

affo: we should open an issue for that

affo: Opened: #1493 

Thanks @sanderson

stdlib/universe/moving_average.go
```diff
@@ -0,0 +1,557 @@
...
+			next := t.window[bj][int64(t.i[bj]+1)%t.n]
+			if next != nil {
+				*sum -= next.(int64)
+				t.count[bj]--
+			}
+
+		}
+		t.i[bj]++
+	}
+	if int64(t.i[bj]) < t.n-1 {
```
roshie548: This produces unwanted behavior; it should be called in Process after the RowReader. #1496

PR #1452, opened 2019-07-03
test(stdlib/range): added test for range stop parameter correctness

aanthony1243:
### Done checklist

- [X] Test cases written


wolffcm:
My comment was for a different PR but this one looks good too. 😁

PR #1455, opened 2019-07-03
Add Rust to CircleCI config

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


.circleci/config.yml
```diff
@@ -2,12 +2,13 @@ version: 2
 jobs:
   test:
     docker:
-      - image: nathanielc/flux-build:latest
+      - image: nathanielc/flux-build:rust
```
affo: Is this your image id?

Did you push a new one built from `Dockerfile_build` with that tag?

Dockerfile_build
```diff
@@ -2,7 +2,69 @@
...
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+    ruby \
+    ca-certificates curl file \
+    build-essential \
+    autoconf automake autotools-dev libtool xutils-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install ragel
+ENV RAGEL_VERSION=6.10
```
affo: Yeah, so now we  can `brew install ragel`, right?

internal/rust/scanner/src/lib.rs
```diff
@@ -515,6 +515,7 @@ a
 
     // TODO(affo): this fails.
     #[test]
+    #[ignore] // See https://github.com/influxdata/flux/issues/1448
```
affo: I think there is another test that fails in the parser.

The duration one

.circleci/config.yml
nathanielc: The image is under my account since we don't have a company account that I know of. The image is auto built from the `flux` repo, but inorder to CI to pass before the changes are merged to master I built  the image locally and pushed it with the `rust` tag. Once this all merges to master we can  edit that line again to say `latest`.

Dockerfile_build
nathanielc: That is the intent, yes.

internal/rust/scanner/src/lib.rs
nathanielc: Correct, it has also been ignored https://github.com/influxdata/flux/pull/1455/files#diff-d998de313c2d6e6df4ca4916fc1d4da0R5029

.circleci/config.yml
```diff
@@ -24,6 +25,13 @@ jobs:
             - flux-gomod-{{ .Branch }}-{{ .Revision }} # Matches when retrying a single run.
             - flux-gomod-{{ .Branch }}-                # Matches a new commit on an existing branch.
             - flux-gomod-                              # Matches a new branch.
+      # Populate Rust cache
```
wolffcm: I guess this is like the GOCACHE but for Rust compilation artifacts?

Dockerfile_build
wolffcm: This will fix:
https://github.com/influxdata/flux/issues/1183
Nice!!

.circleci/config.yml
wolffcm: I was confused by this until I realized you are not merging into master.

Dockerfile_build
wolffcm: Maybe this part should get merged to master?

.circleci/config.yml
nathanielc: Yes, and we use a tool called `sccache` which allows you to specify a size limit on the cache as an LRU, otherwise Rust will grow the cache indefinitely.

internal/rust/scanner/src/lib.rs
affo: Wops, didn't see that!

PR #1457, opened 2019-07-05
fix(stdlib/universe): Added end to end tests for conversion funcitons

aanthony1243:
* tests for toBool, toFloat, toInt, toString, toTime, toUint added, minor issues fixed.
* toDuration removed, since flux tables do not support duration as a column type.

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


docs/SPEC.md
```diff
@@ -3400,15 +3400,6 @@ Example: `from(bucket: "telegraf") |> filter(fn:(r) => r._measurement == "mem" a
...
 
-##### toDuration
-
-Convert a value to a duration.
-
-Example: `from(bucket: "telegraf") |> filter(fn:(r) => r._measurement == "mem" and r._field == "used") |> toDuration()`
-
-The function `toDuration` is defined as `toDuration = (tables=<-) => tables |> map(fn:(r) => duration(v:r._value))`.
-If you need to convert other columns use the `map` function directly with the `duration` function.
-
```
nathanielc: Why is this being removed?

aanthony1243: tables don't have a duration type.  we can add it back in if we wanted to extend tables to include duration columns.

Issue #1037, opened 2019-03-08
Remove the implicit _value column behavior of map


nathanielc:
Currently this is possible `map(fn: (r) => r._value + 2)` and it equivalent to this:

```
map(fn: (r) => ({_value: r._value + 2}))
```

Allowing for the implicit column name is confusing and inconsistent. We should remove this special case.

As an added bonus this simplification will make mapping over non table types easier.

affo:
Do we want the same behavior for `reduce`? I am touching a portion of code which is very similar.

PR #1460, opened 2019-07-08
fix(execute): make row map fn error if return type is not object

affo:
#1037 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/testing/testdata/show_all_tag_keys.flux
+++ b/stdlib/testing/testdata/show_all_tag_keys.flux
@@ -1137,8 +1137,7 @@ t_show_all_tag_keys = (table=<-) =>
 		|> keys()
 		|> group()
 		|> distinct()
-		|> map(fn: (r) =>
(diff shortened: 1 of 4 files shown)


stdlib/testing/testdata/show_all_tag_keys.flux
```diff
@@ -1138,7 +1138,7 @@ t_show_all_tag_keys = (table=<-) =>
 		|> group()
 		|> distinct()
 		|> map(fn: (r) =>
-			(r._value)))
+			({_value: r._value})))
```
nathanielc: We should convert this map call to a `keep` call since its purpose is to drop all columns but the `_value` column.

affo: Yeah, I guess you are right!  I made a no-look change 😅

Issue #155, opened 2018-10-19
Implement Time and date functions


nathanielc:
From the SPEC:

These are builtin functions that all take a single `time` argument and return an integer.

* `second` - integer
    Second returns the second of the minute for the provided time in the range `[0-59]`.
* `minute` - integer
    Minute returns the minute of the hour for the provided time in the range `[0-59]`.
* `hour` - integer
    Hour returns the hour of the day for the provided time in the range `[0-59]`.
* `weekDay` - integer
    WeekDay returns the day of the week for the provided time in the range `[0-6]`.
* `monthDay` - integer
    MonthDay returns the day of the month for the provided time in the range `[1-31]`.
* `yearDay` - integer
    YearDay returns the day of the year for the provided time in the range `[1-366]`.
* `month` - integer
    Month returns the month of the year for the provided time in the range `[1-12]`.

PR #1472, opened 2019-07-09
feat(stdlib/universe): days of the week constants

roshie548:
Added days of the week as constants as defined in the spec

### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written


stdlib/universe/universe.flux
```diff
@@ -257,3 +257,11 @@ toUInt     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: uint(v:r._v
...
 toBool     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: bool(v:r._value)}))
 toTime     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: time(v:r._value)}))
+
+Sunday    = 0
+Monday    = 1
+Tuesday   = 2
+Wednesday = 3
+Thursday  = 4
+Friday    = 5
+Saturday  = 6
```
nathanielc: What do you think? Should we put this constants in a `time` package?

roshie548: If there are other functions that are implemented/will be implemented for `time`, then I think so

roshie548: eg: For [these](https://github.com/influxdata/flux/issues/155), it would be nice to have everything in a `time` package

roshie548:
Moved to [#1479](https://github.com/influxdata/flux/pull/1479)

PR #1477, opened 2019-07-10
Add Rust Parser

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


.circleci/config.yml
```diff
@@ -2,12 +2,13 @@ version: 2
 jobs:
   test:
     docker:
-      - image: nathanielc/flux-build:latest
+      - image: nathanielc/flux-build:rust
```
affo: Should you change this to `latest` given that we are merging to master?

Or should that be done after this gets merged with a separate PR 🤔

nathanielc: It needs to be done after we merge to master. Its kind of a chicken and the egg. I'll create a PR to do that now so we don't forget.

internal/parser/parser.go
```diff
@@ -1336,26 +1336,6 @@ func (p *parser) consume() {
...
-		p.errs = append(p.errs, ast.Error{
-			Msg: fmt.Sprintf("expected %s, got %s (%q) at %s",
-				exp,
-				p.tok,
-				p.lit,
-				p.s.File().Position(p.pos),
-			),
-		})
-	}
-
```
affo: Why has this been removed?

nathanielc: It was redundant code, in porting the Go code to Rust I saw that it was not necessary. This was confirmed after removing it and all tests still pass.

In short the logic the check for `buffered` is already inside the call to `p.scan()` and the error logic is identical to the logic below after calling scan.

Issue #1413, opened 2019-06-24
row functions can only refer to globally defined elements


aanthony1243:
row functions (e.g. for map, filter, reduce) are all limited to referring to elements in the global scope.  The elements they should be able to refer to include: 
1.  local values to the current script, but outside the function's scope
2.  package-level values for imported packages
3.  local variables inside the function body. 

the last one surprised me: 
```
intables |> map(fn: (r) => {  myVal = strings.split(v:r.host, t: "") return {host: myVal[0]}})
Error: error calling function "map": name "myVal" does not exist in scope
```

andig:
Shouldn't the example actually be

    intables 
    |> map(fn: (r) => (
        myVal = strings.split(v:r.host, t: "") 
        return {host: myVal[0]}
    ))

Notice the different brackets. But I still feel I might be understanding flux totally wrong...

PR #1479, opened 2019-07-10
feat(stdlib): date package

roshie548:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


compile.go
```diff
@@ -89,6 +89,7 @@ var (
 		"math",
 		"strings",
 		"regexp",
+		"time",
```
nathanielc: Lets not add the `time` package to the prelude. People can import it if they need it. See https://github.com/influxdata/flux/issues/1480

nathanielc: Fixing #1480 is blocked on #1413, as such lets not block this work. We can add `time` to the prelude for now as doing so does not make fixing #1480 any harder.

stdlib/calendar/time_fns.flux
```diff
@@ -0,0 +1,60 @@
+package calendar
```
nathanielc: Can you rename this file to `calendar.flux` to be consistent?

stdlib/testing/testdata/time_hour.flux
```diff
@@ -0,0 +1,71 @@
...
+,,1,QQ,_m,2018-05-22T19:54:00Z,19
+,,1,QQ,_m,2018-05-22T19:54:10Z,19
+,,1,QQ,_m,2018-05-22T19:54:20Z,19
+,,2,RR,_m,2018-05-22T19:53:00Z,19
+,,2,RR,_m,2018-05-22T19:53:10Z,19
+,,2,RR,_m,2018-05-22T19:53:20Z,19
+,,2,RR,_m,2018-05-22T19:53:30Z,19
+,,3,SR,_m,2018-05-22T19:53:40Z,19
+,,3,SR,_m,2018-05-22T19:53:50Z,19
+,,3,SR,_m,2018-05-22T19:54:00Z,19
```
nathanielc: I'd like to see these tests cases have varying output, so instead of using a time stamp with the hour 19 for all values, instead use varying hours. What do you think, should we add a 24 rows, one for each hour?

nathanielc: We could also do the same thing for all the other functions as well. The largest would be day of the year which shouldn't be too large.

jsternberg:
Random nitpick, it makes more sense for these to be part of a `date` package rather than `calendar`. @nathanielc do you have a specific preference for the name?

docs/StandardLibrary.md
```diff
@@ -157,7 +155,6 @@ Saturday  = 6
 
 Time and date functions.  Each of these accept a `time` value and return an integer.
 From the SPEC:
-[IMPL#155](https://github.com/influxdata/flux/issues/155)
 - `second` - integer Second returns the second of the minute for the provided time in the range [0-59].
```
jsternberg: The spec says these functions are supposed to be in the `time` package. Can you communicate with @nathanielc to figure out which package they're supposed to be in? If they're supposed to be in `time`, move them there. If they're supposed to be in `date`, let's modify the spec to show that.

stdlib/testing/testdata/date_month.flux
```diff
@@ -0,0 +1,71 @@
...
+import "date"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,string,string,double
+#group,false,false,false,true,true,false
+#default,_result,,,,,
+,result,table,_time,_measurement,_field,_value
+,,0,2018-05-22T19:53:00Z,_m,FF,1
```
jsternberg: I agree with @nathanielc on this. Can you vary these so the output has different things?

stdlib/testing/testdata/date_second.flux
```diff
@@ -0,0 +1,71 @@
...
+import "testing"
+import "date"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,string,string,double
+#group,false,false,false,true,true,false
+#default,_result,,,,,
+,result,table,_time,_measurement,_field,_value
```
jsternberg: This one is good.

docs/StandardLibrary.md
roshie548: I discussed this with Nathaniel. A `time` package was causing errors because it the name was already assigned to the `time` function, which is why we changed the package name to `calendar` (now `date`). I'll update the spec

stdlib/date/time_fns.flux
```diff
@@ -0,0 +1,60 @@
+package date
```
nathanielc: Can you rename this file to `date.flux` to be consistent with the other packages?

Issue #1366, opened 2019-06-13
Bug: inferring type of call expression


jpacik:
Given the following function definition:

    f = (r) => r.a + r.b

the type of the call `f( {a: 1} )` should be inferred as `int`, however currently an unbound type variable is the result. This issue is to track the fix for this bug.

PR #1486, opened 2019-07-10
fix(semantic): update type inference

jlapacik:
BREAKING CHANGE: removed exported methods.

This commit updates type inference by removing most of the methods associated
with a substitution. In particular, substitutions are no longer merged.
Instead unifyType and unifyKind have been updated to take a substitution as input.

This allows type variables to be unified correctly as previously the Merge method
assumed that each substitution was distict from the other, where distinct means
no type variables in common. This assumption was not correct.

#1366.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/semantic/constraints.go
+++ b/semantic/constraints.go
@@ -472,11 +472,7 @@ func (v ConstraintGenerator) freshType(typ PolyType) PolyType {
 		}
 		sub[tv] = f
 	}
-	tp, ok := typ.apply(sub)
-	for ok {
-		tp, ok = tp.apply(sub)
-	}
-	return tp
+	return sub.applyToType(typ)
 }
 
 func (v ConstraintGenerator) applyKindConstraints(typ PolyType) PolyType {
@@ -561,19 +557,9 @@ func (c *Constraints) Instantiate(s Scheme, loc ast.SourceLocation) PolyType {
 
 	// Add any new kind constraints
 	for _, tv := range s.Free {
-		ks, ok := c.kindConst[tv]
-		if ok {
-			ntv, ok1 := tv.apply(sub)
-			for tvar, ok2 := ntv.(Tvar); ok2 && ok1; {
-				tv = tvar
-				ntv, ok1 = ntv.apply(sub)
-			}
+		if ks, ok := c.kindConst[tv]; ok {
 			for _, k := range ks {
-				nk, ok := k.apply(sub)
-				for ok {
-					nk, ok = nk.apply(sub)
-				}
-				c.AddKindConst(tv, nk)
+				c.AddKindConst(sub.applyToTvar(tv), sub.applyToKind(k))
 			}
 		}
 	}
@@ -583,23 +569,10 @@ func (c *Constraints) Instantiate(s Scheme, loc ast.SourceLocation) PolyType {
 		fvs := tc.l.freeVars(c)
 		// Only add new constraints that constrain the left hand free vars
 		if fvs.hasIntersect(s.Free) {
-			l, ok := tc.l.apply(sub)
-			for ok {
-				l, ok = l.apply(sub)
-			}
-			r, ok := tc.r.apply(sub)
-			for ok {
-				r, ok = r.apply(sub)
-			}
-			c.AddTypeConst(l, r, loc)
+			c.AddTypeConst(sub.applyToType(tc.l), sub.applyToType(tc.r), loc)
 		}
 	}
-
-	tp, ok := s.T.apply(sub)
-	for ok {
-		tp, ok = tp.apply(sub)
-	}
-	return tp
+	return sub.applyToType(s.T)
 }
 
 func (c *Constraints) String() string {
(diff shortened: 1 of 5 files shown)


semantic/poly_types.go
```diff
@@ -85,22 +93,36 @@ func (tv Tvar) freeVars(c *Constraints) TvarSet {
...
+	// Use the substitution to reduce the left and right hand sides.
+	// This ensures we never add a transitive cycle in the substitution.
+	tl, ok := l.apply(s)
+	for ok {
+		tl, ok = tl.apply(s)
+	}
+	tr, ok := r.apply(s)
+	for ok {
+		tr, ok = tr.apply(s)
+	}
```
nathanielc: I have some questions and I am going to use this section of code to demonstrate them.

As I understand it the overall change is to no longer merge substitutions together but to grow the substitution and never remove from it. This means that a substitution can contains chains of subs.

For example give a  substitution  like this:

```
t0 -> t1 
```

And another sub like this:

```
t1 -> int
```

Previously we would have merged the subs to become this:

```
t0 -> int
```

And lost the information about `t1`.
Previously the assumption as that it was safe to loose information about `t1` because we would have substituted all instances of `t1` with `int` already? 

This change simply grows the substitution instead:

```
t0 -> t1
t1 -> int
```

Therefore in order to completely apply the substitution we have to re-apply it until nothing changes. I.e a fixed-point algorithm.


Some of my questions:

1. How does this change actually solve the issue? Was it because the assumption the it was safe to loose information in the substitution was violated? If so can we track down were that happened?
2. Why is the loop part of this code and not part of the substitution type itself? In other words, how do I know i am supposed to apply in a loop here but not in other places?
3. If `s`, `l`, and `r` are all inputs to this function why was the substitution not applied before passing them into this function? In other words, how do I know its my responsibility to apply the substitution here in addition to performing the unification.

semantic/constraints.go
```diff
@@ -340,12 +340,13 @@ func (v ConstraintGenerator) typeof(n Node) (PolyType, error) {
...
 			upper = append(upper, field.Key.Key())
 		}
-		var with *Tvar
+		var with PolyType
 		if n.With != nil {
 			t, err := v.lookup(n.With)
 			if err != nil {
 				return nil, err
 			}
+			with = t
```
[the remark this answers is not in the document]
jlapacik: 'with' isn't tested at all yet. But previously this attribute was a pointer to type Tvar which isn't correct. The `with` parameter should be a polytype. And no it wasn't being set before.

semantic/poly_types.go
jlapacik: The example you give is not quite related to this issue. The issue was that you could have the same Tvar bound in both substitutions. `t0 -> t1` in one, and `t0 -> int` in the other. In this case, you must unify `t1` and `int`, but that never happened, so you never got `t1 -> int` in the returned substitution.

Having said that and going back to your example, it is not safe to lose information about a Tvar in a substitution unless you also perform the same substitution over the remaining constraints. 

Addressing point 2, yes I can add a method to the substitution so that the loop is hidden and that applying a substitution to a type automatically reduces the type completely. Because you should always apply a substitution that way.

As for point 3, only in Tvar unification can a transitive cycle be added to the substitution. This is why you **must** use the substitution to reduce both sides of the current constraint before adding it to the substitution. This is the **only** place where this occurs.

semantic/poly_types.go
```diff
@@ -132,23 +148,23 @@ func (tv Tvar) Equal(t PolyType) bool {
...
 			return nil, fmt.Errorf("%v != %v", n, t)
 		}
 	case Tvar:
-		return t.unifyType(kinds, n)
+		return t.unifyType(kinds, s, n)
 	default:
 		return nil, fmt.Errorf("cannot unify %v with %T", n, t)
 	}
-	return nil, nil
+	return s, nil
```
wolffcm: Why return `s` here?  If you get here, you have two equal `Nature` types, so the substitution should be `nil`?

jlapacik: The `unifyTypes` method now takes the substitution for the whole program as input, and returns a possibly updated program substitution. In this case if the two types unify, we just return the substitution that was passed in.

Only when we unify a type variable with another type is there a possibility that we return an expanded substitution.

wolffcm: I see... the comment on the interface method should be updated to reflect that.

semantic/poly_types.go
```diff
@@ -23,12 +23,16 @@ type TypeExpression interface {
...
 	// occurs reports whether tv is a free variable in the type.
 	occurs(tv Tvar) bool
-	// substituteType replaces tv for t producing a new type.
-	substituteType(tv Tvar, t PolyType) PolyType
+
+	// apply applies a substitution to a type
+	// apply returns false if there were no substitutions made
+	apply(sub Substitution) (PolyType, bool)
+
 	// unifyType unifies the two types given the kind constraints and produces a substitution.
```
wolffcm: This comment should be updated to reflect that it will return the possibly unchanged, or possible expanded set of substitutions.

jlapacik: Resolved.

jlapacik:
Closing this. See #1613 .

PR #1488, opened 2019-07-10
fix(docs/covarianceSpec): fix the example for covariance

### LINKED ISSUE #177: Covariance SPEC example is wrong
The covariance function will need two columns to operate on, the from function only returns a single _value column, I think the example should show a join or something so its clear how to use covariance.

ewendai:
### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

Closes #177

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1665,7 +1665,12 @@ Covariance has the following properties:
 Additionally exactly two columns must be provided to the `columns` property.
 
 Example:
-`from(bucket: "telegraf/autogen") |> range(start:-5m) |> covariance(columns: ["x", "y"])`
+```
+from(bucket: "telegraf/autogen") 
+    |> range(start: -5m) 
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -1665,7 +1665,10 @@ Covariance has the following properties:
 Additionally exactly two columns must be provided to the `columns` property.
 
 Example:
-`from(bucket: "telegraf/autogen") |> range(start:-5m) |> covariance(columns: ["x", "y"])`
+`from(bucket: "telegraf/autogen") 
+|> range(start:-5m) 
+|> map(fn: (r) => ({r with x: r._value and y: r._value*r._value/2})) 
+|> covariance(columns: ["x", "y"])`
```
affo: Should be:
``````
```
    |> range(start: -5m) 
    |> map(fn: (r) => ({r with x: r._value, y: r._value * r._value / 2})) 
    |> covariance(columns: ["x", "y"])
```
``````

PR #1490, opened 2019-07-11
fix(values): added power operator

sauren-khosla:
### Done checklist
- [x] docs/SPEC.md updated
- [X] Test cases written


nathanielc:
@sauren-khosla Can you also add the operator to the Rust parser? Now that we have both we get to do double the work for a while. If you need help writing the Rust bits reach out we can help. It should be pretty straightforward to copy the existing patterns.

PR #1494, opened 2019-07-11
fix(stdlib): map will not panic when a record is null

jsternberg:
Map will panic in two situations:

1. The entire column is null.
2. The first record is null.

This is because it attempts to create a column with the nil type which cannot
be represented on a record.

Ideally, the properties would be determined after compiling the function
since we should know the complete return type. Unfortunately, type
inference isn't currently capable of getting all of the types that will
be part of the return value when those values aren't directly referenced
such as in the case of merge key or using the `with` keyword.

In order to fix this, we use a hybrid approach where we determine the
types which are referenced from type inference and add or update these
types based on the first record that is returned. This should get us a
correct type for each of the values in the record.

### DIFF
--- a/stdlib/universe/map.go
+++ b/stdlib/universe/map.go
@@ -1,16 +1,16 @@
 package universe
 
 import (
-	"fmt"
 	"sort"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
-	"github.com/pkg/errors"
 )
 
 const MapKind = "map"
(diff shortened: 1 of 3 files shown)


stdlib/universe/map.go
```diff
@@ -147,35 +147,52 @@ func (t *mapTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey
...
 			if i == 0 {
-				properties = m.Type().Properties()
-				keys = make([]string, 0, len(properties))
-				for k := range properties {
-					keys = append(keys, k)
+				// Merge in the valid properties from the type.
+				// This will catch all of the non-referenced keys that are
+				// merged in with merge key or with.
+				for k, typ := range m.Type().Properties() {
+					if execute.ConvertFromKind(typ.Nature()) != flux.TInvalid {
```
jlapacik: Why isn't this an error?

jsternberg: It's because the value could be null and type inference reported a different type. I've modified this so it merges only the keys that weren't found with type inference. I'm not sure if this will catch all instances since type inference can sometimes return a null type when it's not supposed to, but I think it should end up working.

So I also added a todo and told it to merge it regardless if type inference returned null.

PR #1498, opened 2019-07-12
feat(universe): holt winters transformation

affo:
I tried to use arrow arrays as much as I could in the algorithm. Holt-Winters `forecast` uses arrow arrays, but the Nelder-Mead optimizer still uses slices.

Should I turn HW  to slice or should I turn NM to arrow?
Is the approach I choose (allocate arrow array and use the underlying slice for manipulation) sensed and a good practice in your opinion?

NOTE: the e2e test has a TODO. There is an inconsistency in windowing from Flux and InfluxQL (at least, it seems so).

NOTE: I trusted HW and NM implementation as it is a porting of InfluxDB's one.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


affo:
More context on the e2e test.

If I try to change time before `holtWinters`, then it messes up time buckets and I get wrong results.

If I leave as so, I get correct values, but shifted timestamps.

I tried to run the query both in InfluxQL and Flux without `holt_winters` and the `first` for each `window` has a different timestamp.

The `inData` is huge, but the test is cool because it uses the exact same starting dataset that I used in the InfluxQL query.

stdlib/universe/holt_winters.go
```diff
@@ -0,0 +1,463 @@
...
+		interval := int64(hwt.interval)
+		remainder := t % interval
+		if remainder > interval/2 {
+			// round up
+			return (t/interval + 1) * interval
+		}
+		// round down
+		return (t / interval) * interval
+	}
+	nextBucket := func() {
```
aanthony1243: i really like this helper function technique.

stdlib/testing/testdata/holt_winters.flux
```diff
@@ -0,0 +1,1292 @@
...
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T02:48:00.000000000Z,4.662
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T02:54:00.000000000Z,4.551
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:00:00.000000000Z,4.495
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:06:00.000000000Z,4.432
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:12:00.000000000Z,4.367
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:18:00.000000000Z,4.275
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:24:00.000000000Z,4.265
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:30:00.000000000Z,4.111
+,,0,2015-08-22T22:12:00.000000000Z,2015-08-28T03:01:00.000000000Z,water_level,water,2015-08-25T03:36:00.000000000Z,3.986
+,,0,2015-08-22T2
```
aanthony1243: I'd really like to find out why this is the case.  I haven't yet, but will keep looking,

stdlib/universe/holt_winters.go
```diff
@@ -0,0 +1,463 @@
...
+
+	// Appending columns.
+	if err := builder.AppendTimes(newTimeIdx, newTs); err != nil {
+		return err
+	}
+	if err := builder.AppendFloats(newValueIdx, newVs); err != nil {
+		return err
+	}
+	for j, c := range tbl.Key().Cols() {
+		v := tbl.Key().Value(j)
```
aanthony1243: did you consider using  ` execute.AppendKeyValues(tbl.Key(), builder)`  for `newVS.Len()` iterations?   it's inside-out of how you append, perhaps yours is a bit faster?

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -0,0 +1,289 @@
...
+	hwGuessLower = 0.3
+	//  The upper bound on the grid
+	hwGuessUpper = 1.0
+	// The step between guesses
+	hwGuessStep = 0.4
+)
+
+// New creates a new HoltWinters.
+func New(n, s int, withFit bool) *HoltWinters {
+	seasonal := true
```
aanthony1243: nit: `seasonal := s>=2`

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// We both want to track memory usage and use slices.
+	// The algorithm changes values of params and this isn't possible with arrow arrays.
+	// So, we build a new arrow array, and we use its underlying slice for manipulation.
+	pb := newBuilder()
+	size := 6
+	if r.seasonal {
+		size += m
+	}
+	pb.Reserve(size)
```
aanthony1243: it seems like even if `r.seasonal` is true, that `size` is trivially small, so that it's sufficiently constant relative to the rest of the algorithm's memory usage.   Is an arrow array backing worth the overhead involved?

stdlib/universe/holt_winters/neldermead.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Optimize applies the Nelder-Mead simplex method with the Optimizer's settings.
+func (o *Optimizer) Optimize(
+	objfunc func([]float64) float64,
+	start []float64,
+	epsilon,
+	scale float64,
+) (float64, []float64) {
+	n := len(start)
```
aanthony1243: n is very small, right?  I think slices are much more efficient in this case.

stdlib/universe/holt_winters.go
```diff
@@ -0,0 +1,463 @@
...
+	vs := array.NewFloat64Builder(fluxarrow.NewAllocator(hwt.alloc))
+	var start, stop int64
+	bucketEnd := int64(-1)
+	bucketFilled := false
+	roundTime := func(t int64) int64 {
+		interval := int64(hwt.interval)
+		remainder := t % interval
+		if remainder > interval/2 {
+			// round up
+			return (t/interval + 1) * interval
```
aanthony1243: return t + interval?

stdlib/universe/holt_winters.go
```diff
@@ -0,0 +1,463 @@
...
+		}
+		// round down
+		return (t / interval) * interval
```
aanthony1243: return t?

stdlib/universe/holt_winters.go
```diff
@@ -0,0 +1,463 @@
...
+		// we work row-wise
+		for i := 0; i < cr.Len(); i++ {
+			// drop values with invalid timestamp
+			if cts := cr.Times(timeIdx); cts.IsValid(i) {
+				// the first value must be valid, skip it if it isn't so
+				if isFirst() && isNull(cr, i) {
+					continue
+				}
+				trueT := cts.Value(i)
+				roundT := roundTime(trueT)
```
aanthony1243: is roundTime supposed to return the lower interval boundary?  it's always equal to either trueT  or trueT+interval.   

I think you can simplify this to  
1.  isFirst()  -->   record point and  bucketBegin = trueT; bucketEnd = bucketBegin + interval
2.  if  trueT < bucketEnd then continue
3.  else  record point and bucketBegin = bucketEnd ;  bucketEnd += interval

I can't prove it, but my gut feeling is that the 5 min. offset bug is hiding in this function.

stdlib/universe/holt_winters/holt_winters.go
affo: Yeah, I was wondering that too. I think you've got the point. `n` should be reasonably small (max 1k values? I think it is a correct assumption to say no-one will need to predict millions of values). I'll turn the implementation back to slices 👍 

I'll update with using `Allocate` for the allocator, just to remove the overhead of arrow, but tracking memory usage. Then we can discuss if that's a good approach or not :)

stdlib/universe/holt_winters/neldermead.go
affo: See above

stdlib/universe/holt_winters.go
affo: Got it, I'll further examine the behavior of the function.

I am supposing that this implementation is correct given that tests pass, and only the e2e test  that involves windowing fails.

The unit tests use the same dataset after `window -> first` run in InfluxQL.

stdlib/universe/holt_winters.go
affo: Sorry, copy/paste from InfluxQL. Definitely  `return t`!

affo: Mmmmm 🤔 No, it's different `/`  is the integer division. The code is correct.

stdlib/universe/holt_winters.go
affo: See below

stdlib/universe/holt_winters.go
affo: I was definitely looking for that function, thank you!

affo:
@aanthony1243 I addressed your comments.

I also fixed the e2e test.

When InfluxQL executes `SELECT  FIRST(...) ... GROUP BY time(...)`, it associates the timestamp of the beginning of the window to every value extracted by `FIRST`. In my Flux script, instead, I was using the timestamp of the record selected (in general, different from the beginning of the window).

In order to fix that, it was as easy as overwriting `_time` with `_start`.

affo:
Thank you! 🙏

stdlib/universe/holt_winters/neldermead.go
```diff
@@ -0,0 +1,253 @@
...
+
+	//holds vertices of simplex
+	o.tracker.track(n*n + 1)
+	defer o.tracker.free(n*n + 1)
```
nathanielc: This should be `o.tracker.track(n*(n + 1))`

stdlib/universe/holt_winters/neldermead.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// Optimize applies the Nelder-Mead simplex method with the Optimizer's settings.
+func (o *Optimizer) Optimize(
+	objfunc func([]float64) float64,
+	start []float64,
+	epsilon,
+	scale float64,
+) (float64, []float64) {
+	n := len(start)
```
nathanielc: I can see why it is hard to use Arrow here and why it makes sense to just track the allocations out of band from the slices.

Since we need mutability here but do not need the interoperability that Arrow gives us here should we just create a package to encapsulate this behavior? As seen above it is easy to have a typo and track a different value from what is actually allocated. Instead if you just asked for an allocated slice of floats and the tracking was internal to that call do errors could be made:

Maybe a simple API like this:

```
type SliceAllocator interface {
    AllocateFloats(n int) (data []float, free func() )
}
```

And used like:

```
v, free := o.allocator.Floats(n)
defer free()
```

Thoughts?

nathanielc: Also like @aanthony1243  says `n` is small here i.e. likely always less than `100` so maybe we just don't care about tracking this smaller bit of memory usage? But it is `n^2` memory usage...

docs/SPEC.md
```diff
@@ -3264,6 +3264,55 @@ from(bucket: "telegraf/autogen"):
...
+
+```
+from(bucket: "waterhouse/autogen")
+    |> range(start: -7y)
+    |> filter(fn: (r) => r._field == "water_level")
+    |> window(every: 379m)
+    |> first()
+    // We need to "un-window", otherwise the Holt-Winters algorithm would run
+    // for each window separately.
+    |> window(every: inf)
```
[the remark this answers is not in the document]
affo: I am not using `aggregateWindow` because `first` is not an aggregator.

That shouldn't work because it doesn't have the `column` parameter.

Anyways, I'll add an example with `mean`.

docs/SPEC.md
```diff
@@ -3264,6 +3264,55 @@ from(bucket: "telegraf/autogen"):
...
+If the data presents no seasonality, the user can specify `s = 0`.
+The user can request to include the fitted data (on the given dataset) in the result by setting `withFit = true`.
+The `n` predicted values occur at the `interval` specified by the user.
+If your `interval` is `6m` and `n` is `3`, you will receive three predicted values that are each six minutes apart.
+`s` is the seasonal pattern parameter and delimits the length of a seasonal pattern according to `interval`.
+If your `interval` is `2m` and `s` is `3`, then the seasonal pattern occurs every six minutes, that is, every three data points.
+
+Before processing the data, Holt Winters will cleanse it according to these rules:
+ - rows that have a null timestamp get discarded;
+ - rows that have a null value are considered invalid, but used by the algorithm;
```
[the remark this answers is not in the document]
affo: This means that they are not discarded, but the algorithm uses them. I updated the text.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+	}
+
+	size := 6
+	if r.seasonal {
+		size += m
+	}
+	r.tracker.track(size)
+	defer r.tracker.free(size)
+	params := make([]float64, 6, size)
```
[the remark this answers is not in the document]
affo: Mmmmm, you could have 1000 points per season and that makes that slice as big as 1006.

affo:
@nathanielc as a general comment, I think that if I am releasing more than once the same resource, that should panic.

If you look at the last lines in tests, I am checking that the allocator has no memory allocated.

Now I am responding to your questions.

docs/SPEC.md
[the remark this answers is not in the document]
affo: I see, I missed that, being a selector, `first` has the `column` parameter 👍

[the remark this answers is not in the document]
affo: Still  remains that you cannot specify  the `offset` in `aggregateWindow`. So maybe, it is better to show a complete example as the one above. What do you think?

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -123,47 +99,55 @@ func (r *HoltWinters) Do(vs *array.Float64) *array.Float64 {
...
 					if sse < minSSE || bestParams == nil {
+						if bestParams != nil {
+							bestParams.Release()
+						}
 						minSSE = sse
 						bestParams = params
 					}
+					if bestParams != params {
+						params.Release()
+					}
```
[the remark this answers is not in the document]
affo: `Optimize` creates new params from the given ones every time it is invoked (I'll point out in a comment where). Those params need to be released in case they are not the best ones (because we don't need them anymore.

In case that they are the new best, we can release the elder bests.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -123,47 +99,55 @@ func (r *HoltWinters) Do(vs *array.Float64) *array.Float64 {
...
 	}
 
 	// Final forecast
-	fcast := r.forecast(bestParams, false)
-	defer r.tracker.free(len(fcast))
-	builder.AppendValues(fcast, nil)
-	return builder.NewFloat64Array()
+	fcast := func() *mutable.Float64Array {
+		fcast := r.forecast(bestParams, false)
+		defer bestParams.Release()
```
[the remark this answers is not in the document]
affo: `forecast` generates new params from `bestParams`, so we don't need them anymore and we can `Release`. That wasn't  actually been done in the for loop above because of the reason above.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -215,25 +198,25 @@ func (r *HoltWinters) forecast(params []float64, onlyFit bool) []float64 {
 	if onlyFit || r.includeFitData {
 		size += l
 	}
-	r.tracker.track(size)
-	fcast := make([]float64, 0, size)
+	fcast := mutable.NewFloat64Array(r.alloc)
```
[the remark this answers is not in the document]
affo: Because what `forecast` forecasts is used outside of its scope. The caller should release the new array. If `forecast` releases `fcast`, then it couldn't be used outside of its scope.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -245,65 +228,65 @@ func (r *HoltWinters) forecast(params []float64, onlyFit bool) []float64 {
...
 }
 
 // Compute sum squared error for the given parameters.
-func (r *HoltWinters) sse(params []float64) float64 {
+func (r *HoltWinters) sse(params *mutable.Float64Array) float64 {
 	sse := 0.0
 	fcast := r.forecast(params, true)
-	defer r.tracker.free(len(fcast))
-	for i := 0; i < len(fcast); i++ {
+	defer fcast.Release()
```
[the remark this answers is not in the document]
affo: This is fine, that array is a new one every time that `forecast` gets called and should be released every time. It would go out of scope anyways at the end of `sse`.

stdlib/universe/holt_winters_test.go
```diff
@@ -0,0 +1,892 @@
...
+				tc.data,
+				tc.want,
+				nil,
+				func(d execute.Dataset, c execute.TableBuilderCache) execute.Transformation {
+					return universe.NewHoltWintersTransformation(d, c, alloc, tc.spec)
+				},
+			)
+
+			if m := alloc.Allocated(); m != 0 {
+				t.Errorf("HoltWinters is using memory after finishing: %d", m)
```
affo: Here is the memory check.

affo:
@nathanielc I have answered to your detailed questions.

In case that you think that the implementation is correct, your questions prove that it is difficult  to follow the `Allocate`/`Release` logic. Should I add comment for future maintenance?

Thanks for the detailed review!

stdlib/universe/holt_winters/neldermead.go
```diff
@@ -237,14 +246,15 @@ func (o *Optimizer) Optimize(
...
 	vs := 0
 	for i := 0; i <= n; i++ {
-		if f[i] < f[vs] {
+		if f.Value(i) < f.Value(vs) {
 			vs = i
 		}
 	}
 
-	parameters := make([]float64, n)
+	parameters := mutable.NewFloat64Array(o.alloc)
```
affo: Here it is, `Optimize` creates new params from the given ones. It cannot `Release` them for the very same reason of `forecast`, because otherwise they couldn't be used outside of its scope.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -123,47 +99,55 @@ func (r *HoltWinters) Do(vs *array.Float64) *array.Float64 {
...
 				for phi := hwGuessLower; phi < hwGuessUpper; phi += hwGuessStep {
-					params[0] = alpha
-					params[1] = beta
-					params[2] = gamma
-					params[3] = phi
+					params.Set(0, alpha)
+					params.Set(1, beta)
+					params.Set(2, gamma)
+					params.Set(3, phi)
 					sse, params := r.optim.Optimize(r.sse, params, r.epsilon, 1)
```
nathanielc: This is the line that confused me. In general I'd like to write the code such that the calls to `New` and `Release` are in easy to spot pairs.

This line hides the fact that the Optimizer creates a new array and reassigns that on top of the params var and thus requiring a call to Release. At first read it looks like params is being updated inplace but not reallocated. 

I see two ways we can clarify this code.

1. Rewrite Optimizer so that it doesn't reallocate params rather does inplace update
2. Move this inner loop into a function so `defer params.Release()` can be used. This is one is likely harder as we don't want to release the bestParams ...

The first option should be a small change to update the input parameters and not return them as a new array.

affo: I see your point 👍 

I have already tried the second option and that was pretty hard and I couldn't get to a working solution.

The problem with the first one is that the Optimizer shouldn't have side effects on `params`. That first array should be left untouched, as it is used as it is but with `a, b, c, d` changed every time.

I think that code is a bit tricky, because when you `params :=` you are not resetting params, but shadowing it.

affo: In other terms, the _outer_ params must leave for the whole `for` loop, while the `params` created by `Optimize` should be released conditionally.

nathanielc: @affo Why can't the Optimizer method mutate the input params? I looked over the code and it seemed fine to change that at first glance.

affo: I think that the initial `params` must stay the same at every iteration in the `for` loop (they get only changes on the first 4 values). The optimizer only reads from them to generate a new solution that can be the best or not. Anyways, it must start from those initial params every time, it cannot change them. Otherwise it would produce incorrect results

nathanielc: ok, then lets add some comments to make this clear what is going on, i.e. Optimize creates a new params and the initial params is used several times. Perhaps we could rename the outer params to initParams so it is clear that it is being reused instead of shadowing?

affo: Definitely agree 👍

affo:
@nathanielc in the last commit I added comments and fixed the SPEC.

If that's fine for you, I can rebase and solve the conflict

PR #1511, opened 2019-07-12
fix(stdlib): correct moving average behavior with chunking

roshie548:
The previous implementation incorrectly accounted for the case where a chunk ends but the specified period has not beed reached. This was being called at the end of every chunk, which produced the unwanted behavior of appending even if there are more rows in different chunks ([see here](https://github.com/influxdata/flux/pull/1443/files#r302745293)). The solution is to check if the specified period has been reached after iterating through all the chunks, at the end of the `Process` method.

In addition, an arrayContainer interface is created to reduce the number of functions needed for different column types.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/moving_average.go
+++ b/stdlib/universe/moving_average.go
@@ -5,11 +5,11 @@ import (
 
 	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
 )
 
 const MovingAverageKind = "movingAverage"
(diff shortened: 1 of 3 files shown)


stdlib/universe/moving_average_test.go
```diff
@@ -290,6 +508,49 @@ func TestMovingAverage_Process(t *testing.T) {
 				},
 			}},
 		},
+		{
+			name: "uint with 3 with chunking",
+			spec: &universe.MovingAverageProcedureSpec{
+				Columns: []string{execute.DefaultValueColLabel},
+				N:       3,
+			},
+			data: []flux.Table{&executetest.RowWiseTable{
```
affo: This is great!

We should add tests like this for code coverage 👍

stdlib/universe/moving_average.go
```diff
@@ -2,10 +2,10 @@ package universe
 
 import (
 	"fmt"
+	"github.com/influxdata/flux/values"
```
affo: this should go with the second group of imports separated by space.

(You'll get used  to always check imports of  the files in  a commit before pushing :))

PR #1517, opened 2019-07-12
fix(stdlib/universe): fixed mode to display multiple modes if necessary

sauren-khosla:
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


nathanielc:
I am a little concerned that the default behavior of mode can return 0-N rows. Mode will likely be used as part of `aggregateWindow` which will expect exactly one row of output. It won't break anything in that function but it may break a users assumptions about how it works. 

I think we should make the default behavior to return the first mode found and an explicit opt in to handle multi mode or no mode. Thoughts?

PR #1518, opened 2019-07-12
fix(stdlib/universe): mode displays multiple modes for multi-modal data

sauren-khosla:
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


sauren-khosla:
> I see there were tests for multiple modes.
> 
> What have you fixed, and what is now possible to do now that wasn't possible before?
> 
> Can you add a test demonstrating that?

mode should properly return all the modes instead of just one. the new test file demonstrates that functionality.

PR #1519, opened 2019-07-12
feat(TICKscript): TICKscript/Flux equivalency tests added

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/testing/kapacitor/combine_join.flux
```diff
@@ -0,0 +1,58 @@
...
+		|> group(columns: ["user"])
+
+    right = table
+		|> range(start: 2018-12-19T22:13:30Z, stop: 2018-12-19T22:14:30Z)
+		|> drop(columns: ["_start", "_stop"])
+		|> filter(fn: (r) =>
+			(r.user == "user2"))
+		|> group(columns: ["_measurement"])
+
+    return join(tables: {left: left, right: right}, on: ["_time", "_measurement"])
```
aanthony1243: was just looking at this, I think I oversimplified this somewhat.   `combine`  operates on a single stream, it doesn't join 2 streams.  It also has an interesting feature,  `tolerance`  that we need to see if we can support (maybe/probably not).    

I think that `pivot`  using `_time, _measurement`  as a rowKey is closer in nature to combine.

stdlib/testing/kapacitor/noop_yield.flux
```diff
@@ -0,0 +1,58 @@
...
+,,3,2018-05-22T19:54:06Z,2,RAM,user2
+,,3,2018-05-22T19:54:16Z,10,RAM,user2
+"
+
+outData = "
+#datatype,string,long,dateTime:RFC3339,double,string,string
+#group,false,false,false,false,true,true
+#default,1,,,,,
+,result,table,_time,_value,_measurement,user
+,,0,2018-05-22T19:53:26Z,0,CPU,user1
```
aanthony1243: our e2e doesn't check the `result` column but theoretically `yield(name:"1")`  would put `1` all the way down the second column in the CSV under `result`.

PR #1524, opened 2019-07-15
feat(stdlib/universe): double exponential moving average

roshie548:
implemented `doubleExponentialMovingAverage`/`doubleEMA` function
Returns `math.NaN` if there aren't enough points to construct a DEMA

I also added a `universe/moving_average` folder with an `exponential_moving_average.go` file. The `ExponentialMovingAverage` struct simplifies the moving averages, and should make Relative Strength Index a bit cleaner as well for the future.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/double_exponential_moving_average.go
```diff
@@ -0,0 +1,406 @@
...
+				err = t.doFirstEMA(&moving_average.ArrayContainer{Array: cr.Ints(j)}, builder, j, doDEMA[j])
+			case flux.TUInt:
+				err = t.doFirstEMA(&moving_average.ArrayContainer{Array: cr.UInts(j)}, builder, j, doDEMA[j])
+			case flux.TFloat:
+				err = t.doFirstEMA(&moving_average.ArrayContainer{Array: cr.Floats(j)}, builder, j, doDEMA[j])
+			case flux.TString:
+				err = t.passThrough(&moving_average.ArrayContainer{Array: cr.Strings(j)}, builder, j)
+			case flux.TTime:
+				err = t.passThroughTime(cr.Times(j), builder, j)
+			}
```
roshie548: Once we calculate the first EMA, we don't deal with chunks anymore. Is this a problem? Should data be kept in chunks?

stdlib/universe/double_exponential_moving_average.go
```diff
@@ -0,0 +1,406 @@
...
+		} else {
+			vals[i] = 0
+			valid[i] = false
+		}
+	}
+
+	bld := arrow.NewFloatBuilder(t.alloc)
+	defer bld.Release()
+
+	bld.AppendValues(vals, valid)
```
aanthony1243: did you consider building up the arrow array incrementally instead of filling up a slice then calling this?

stdlib/universe/double_exponential_moving_average.go
aanthony1243: I think it's OK to not do the chunking.  it's not trivial to do that in this context.

stdlib/universe/double_exponential_moving_average.go
roshie548: I looked at Jonathan's arrow builder, but as it is right now it doesn't let you append nulls.

roshie548: I'll fix that up and recommit

Issue #1513, opened 2019-07-12
CSV result encoder does not return errors from the query


jsternberg:
The csv encoder does not return an error if an error occurred when attempting to read the results. It will write them out as CSV. This stops us from being able to set the http header in influxdb and then write the error.

The csv result encoder should return an error if the query itself failed for some reason and it had to abort. It will then be up to the caller if they wish to encode the error or do something different.

PR #1525, opened 2019-07-15
feat: return query and result errors in the multi result encoder

jsternberg:
The `DelimitedMultiResultEncoder` now returns errors from the query and
the results if nothing has been written to the writer yet instead of
writing them to the stream. This allows the caller to determine what to
do with the error rather than forcing the output to be written to the
stream.

#1513

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/result.go
+++ b/result.go
@@ -232,7 +232,8 @@ type MultiResultDecoder interface {
 type MultiResultEncoder interface {
 	// Encode writes multiple results from r into w.
 	// Returns the number of bytes written to w and any error resulting from the encoding process.
-	// Errors obtained from the results object should be encoded to w and then discarded.
+	// It is up to the specific implementation for whether it will encode any errors that occur
+	// from the ResultIterator.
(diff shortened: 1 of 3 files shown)


result.go
```diff
@@ -273,33 +272,44 @@ type flusher interface {
 	Flush()
 }
 
+// Encode will encode the results into the writer using the Encoder and separating each entry
```
wolffcm: This comment or a similar one should also precede the interface method for `MultiResultEncoder` in this file.

jsternberg: My thoughts were that it would be up to the implementation for what they did so I left the interface ambiguous. Is that alright?

wolffcm: I guess it's okay if it's left up to the implementation, but in that case it would be good to note that in the interface comment.

jsternberg: Done.

Issue #1118, opened 2019-04-03
promql: Support non-existent column names in keep()


juliusv:
PromQL binary expressions that do an explicit join on a number of labels restrict the output labels to the explicitly listed set of labels (which is actually a bug in PromQL which has always been there, see https://github.com/prometheus/prometheus/issues/5326).

In Flux I try to emulate this by adding a `keep()` after a `join()`, but `keep()` returns an error when a column to keep does not exist: `keep error: column "non_existent" doesn't exist`. Since I do not know the actually existing columns at PromQL->Flux transpilation time, would it be possible to (optionally?) just ignore non-existent column names in `keep()`?

juliusv:
@nathanielc What do you think about adding an `allowNonexistent` option for both `drop()` and `keep()` that defaults to `true` for `drop()` and `false` for `keep()` to mirror the current behavior by default? Or is there a nicer way for me to achieve what I want?

aanthony1243:
I propose quietly continuing past non-existing columns in both cases.  Here's my thinking: 

1.  They may have a set of potential columns they want to keep, but can't predict which ones will come up at any time (e.g. changing series id tags).   

2.  If they keep something that they REALLY NEED to exist, the error will manifest later.  That is, if they expect to always see that column, one way or another they will find out.   I don't think we need to kill their whole query over it.

juliusv:
That sounds fine to me.

nathanielc:
@aanthony1243 I agree skipping missing columns is fine.

wolffcm:
If you say `keep(columns: ["non_existent_column"])`, do we end up with tables with no columns?  Should that be a runtime error?

PR #1526, opened 2019-07-16
feat(universe): allow keep to run regardless of nonexistent columns given. if all columns given are nonexistent, return empty table

ewendai:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

#1118

### DIFF
--- a/stdlib/testing/testdata/drop_non_existent.flux
+++ b/stdlib/testing/testdata/drop_non_existent.flux
@@ -30,17 +30,34 @@ inData = "
 "
 
 outData = "
-#datatype,string,string
-#group,true,true
-#default,,
-,error,reference
-,\"drop error: column \"\"non_existent\"\" doesn't exist\",
+#datatype,string,long,dateTime:RFC3339,double,string,string
+#group,false,false,false,false,true,true
+#default,_result,,,,,
+,result,table,_time,_value,_field,_measurement
+,,0,2018-05-22T19:53:26Z,0,usage_guest,cpu
+,,0,2018-05-22T19:53:36Z,0,usage_guest,cpu
+,,0,2018-05-22T19:53:46Z,0,usage_guest,cpu
+,,0,2018-05-22T19:53:56Z,0,usage_guest,cpu
+,,0,2018-05-22T19:54:06Z,0,usage_guest,cpu
+,,0,2018-05-22T19:54:16Z,0,usage_guest,cpu
+,,1,2018-05-22T19:53:26Z,0,usage_guest_nice,cpu
+,,1,2018-05-22T19:53:36Z,0,usage_guest_nice,cpu
+,,1,2018-05-22T19:53:46Z,0,usage_guest_nice,cpu
+,,1,2018-05-22T19:53:56Z,0,usage_guest_nice,cpu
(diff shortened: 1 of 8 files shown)


stdlib/universe/schema_mutators.go
```diff
@@ -206,13 +206,15 @@ func NewDropKeepMutator(qs flux.OperationSpec) (*DropKeepMutator, error) {
 }
 
 func (m *DropKeepMutator) checkKeepColumns(tableCols []flux.ColMeta) error {
-	if m.KeepCols != nil {
-		for c := range m.KeepCols {
-			if err := checkCol(c, tableCols); err != nil {
-				return errors.Wrap(err, "keep error")
+	// Commenting out to allow keep() to run regardless of existential state of column given
```
wolffcm: Can you just delete these lines instead of commenting them out?  Also, is the method `checkCol` just dead code now, not called from anywhere?  If so, it can be deleted too.

stdlib/testing/testdata/drop_non_existent.flux
```diff
@@ -30,17 +30,33 @@ inData = "
...
+,,2,2018-05-22T19:53:56Z,91.02836436336374,usage_idle,cpu
+,,2,2018-05-22T19:54:06Z,68.304576144036,usage_idle,cpu
+,,2,2018-05-22T19:54:16Z,87.88598574821853,usage_idle,cpu
 "
 
 t_drop = (table=<-) =>
 	(table
-		|> range(start: 2018-05-22T19:53:26Z)
-		|> drop(columns: ["non_existent"]))
+		|> drop(columns: ["non_existent", "dropme1", "dropme2"]))
```
wolffcm: I think without a call to `range` this test will fail when it's run in the `influxdb` repo.

stdlib/testing/testdata/keep_non_existent_only.flux
```diff
@@ -0,0 +1,47 @@
...
+,,2,2018-05-22T19:54:06Z,68.304576144036,usage_idle,cpu,cpu-total,host.local
+,,2,2018-05-22T19:54:16Z,87.88598574821853,usage_idle,cpu,cpu-total,host.local
+"
+
+outData = "
+#datatype,string,long
+#group,false,false
+#default,_result,
+,result,table
+,,0
```
wolffcm: I was surprised by this result.  If we "keep" a non-existent column, I would have expected the same number of rows as in the input tables, all grouped into one table, since there can be no columns in the group key.  In general, I don't expect `keep` to change the number of rows.

Maybe zero rows would also be okay, but a single row seems strange to me.

nathanielc: I agree that a single row is incorrect, it should be either the same number of rows as the input table or it should just be an empty table.

An empty table is probably best at this point since none of our code will be able to consume a table with N rows but 0 columns. For example count will not be able to count the rows without at least one column present.

aanthony1243: my guess is that that "row" is being added by the encoder.  if you look at the code, it's not possible that `drop` or `keep` add rows in this case

if it's the encoder, then I would say that is a separate issue of how to encode a table with no columns.

PR #1528, opened 2019-07-16
Flux unit type conversion tests

vlastahajek:
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written

Added type conversion tests between string, double, long, time, unsignedlong.


vlastahajek:
I planned to add type conversion test for `duration` as well, but it doesn't look it is supported. 
First I got  weird errors when trying to use `toDuration()`, but after recent update from  master, this function seems to be removed.
When I define it manually, [as written in the doc](https://docs.influxdata.com/flux/v0.24/functions/built-in/transformations/type-conversions/toduration/),  I get error:   `flux_test.go:143: map function must return an object, got duration`.

And, there is even no column type for duration. 

Does it mean, it is not supported (yet)?

affo:
@vlastahajek Hello.

A recent update to `map` forbids you to write:

```
...
 map(fn: (r) => r._value)
// r._value is not an object
...
```

You should write:

```
...
 map(fn: (r) => ({_value: r._value}))
// fn now returns an object
...
```

or

```
...
 map(fn: (r) => ({r with _value: r._value}))
...
```

instead.

stdlib/testing/testdata/conv_bool_to_float.flux
```diff
@@ -0,0 +1,33 @@
...
+inData = "
+#datatype,string,long,dateTime:RFC3339,boolean,string,string
+#group,false,false,false,false,true,true
+#default,_result,,,,,
+,result,table,_time,_value,_field,_measurement
+,,0,2018-05-22T19:53:30Z,true,k,m
+,,0,2018-05-22T19:53:31Z,false,k,m
+"
+
+// NOTE: This test will fail with differences in the last two rows when time zone support arrives.
```
affo: What do you mean?

vlastahajek: This was a relic of copy&paste from  original totime.flux i took as an example. I totally overlooked it, sorry.

affo:
@vlastahajek I recently discovered that `toDuration` is  not  there because we do not support duration columns.

So you cannot test duration conversions with e2e tests as you are doing here.

I repost the list above without duration:

```
str str
uint str

int float
uint float
float float

uint bool
bool bool

uint time
time time

int int
uint int

uint uint
float uint
bool uint
time uint
```

stdlib/testing/testdata/conv_bool_to_float.flux
affo: 👍  No worries

vlastahajek:
This is the basic set of tests we agreed on with my boss what makes most sense ;). So, if it possible, please, merge it. 
For sure if needed, I can complete the matrix in the second phase..

vlastahajek:
I've added tests to complete the tests matrix.

affo:
@vlastahajek wow that's great.

I'll review tomorrow as I wake up :)

vlastahajek:
I also created issue for limited values for string to bool  conversion: #1530

vlastahajek:
I squashed only those before merge of master (last 16). For some reason rebasing changes in master from last week creates conflicts.

I've also noticed, that just in the middle of my writing of tests, when I was on vacation last week, someone on your side also added E2E conversion tests. From them I learned that I can set different tables into input data 👍.

Do you want me to to merge all those tests (my and existing ones)?
This would result in having files as there are now in master (to_*), but with more data (from my tests).

affo:
For the rebase, that's totally normal. E2e tests pass through a code generation step and that always generates conflicts. you can simply keep the conflict, remove the file, and run `make` to solve that.

It would be great if you could merge your tests with the new ones from the codebase! Can you do that?

vlastahajek:
I merged those tests. 
However, after various rebasing and merging master, the branch got to into a weird state, so I reset it and commited just summary changes.

affo:
@vlastahajek you see the Semantic Pull  Request indicator yellow because of  your message commit. To fix that, please `--amend` your commit and change the message to something like:

```
test(stdlib): e2e tests for type conversion
```

Then force push. Thank you!

vlastahajek:
Ok, done.

affo:
@vlastahajek Thanks for your work!

vlastahajek:
Thanks, @affo , for your help with finishing this

Issue #1532, opened 2019-07-16
Construct a mutable array builder for arrow arrays


jsternberg:
In #1523, we investigated if it was possible to construct an arrow array in a mutable way. We found that the arrow allocator interfaces were capable of doing it, but the arrays were not mutable as arrow arrays are not supposed to be mutable and the builders did not expose the mutable portions.

This issue is to construct a similar builder interface for arrow arrays that allows mutable transformations for Go slices. An example of how to allocate a slice and create an Arrow array is included in #1523.

A few requirements:

* The builder should expose direct access to the slices.
* Null support is not required.
* The memory should be allocated with an arrow allocator.
* Arrow arrays should be created from the slice.

jsternberg:
After consulting with @affo, we're mostly just going to expose the same interface as the normal builders, but we'll have a `Value()` and `Set()` method on the builder and remove nulls from the builder.

We then may try to integrate this into the official builders with null support.

jsternberg:
After a discussion with @affo, the first requirement is not required. While we need the ability to set and inspect values, direct access is not required.

PR #1537, opened 2019-07-17
feat: add mutable arrow array builders

jsternberg:
The mutable arrow array builders use the arrow library's allocator
interface and the memory buffer, but it exposes the ability to set
values as part of the builder such as the ones that would be exposed if
the values were in a slice.

This is only included for the numeric targets at the moment because
those are the most simple to implement. String builders would likely not
be possible and boolean values can be included in the future when we
need them.

These builders do not support null values. We may support that in the
future if it is needed.

These builders are being added as an internal package while we
experiment with this idea in order to avoid adding it as a publicly
exposed API.

#1532 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/mutable/numericarray.go
+++ b/internal/mutable/numericarray.go
@@ -1,0 +1,397 @@
+package mutable
+
+import (
+	"sync/atomic"
+
+	"github.com/apache/arrow/go/arrow"
+	"github.com/apache/arrow/go/arrow/array"
+	"github.com/apache/arrow/go/arrow/memory"
+)
+
+// Int64Array is an array of int64 values.
+type Int64Array struct {
+	refCount int64
+	mem      memory.Allocator
+	data     *memory.Buffer
+	rawData  []int64
+	length   int
+}
+
+// NewInt64Array constructs a new Int64Array.
+func NewInt64Array(mem memory.Allocator) *Int64Array {
+	return &Int64Array{
+		refCount: 1,
+		mem:      mem,
+	}
+}
+
+// Append will append a value to the array. This will increase
+// the length by 1 and may trigger a reallocation if the length
+// would go over the current capacity.
+func (b *Int64Array) Append(v int64) {
+	b.Reserve(1)
+	b.rawData = append(b.rawData, v)
+	b.length = len(b.rawData)
+}
+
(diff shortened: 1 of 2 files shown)


internal/mutable/numericarray.go
```diff
@@ -0,0 +1,399 @@
...
+	atomic.AddInt64(&b.refCount, 1)
+}
+
+// Release will release any reference to data buffers.
+func (b *Int64Array) Release() {
+	if atomic.AddInt64(&b.refCount, -1) == 0 {
+		if b.data != nil {
+			b.data.Release()
+			b.data = nil
+			b.rawData = nil
```
affo: why not `b.reset()`?

jsternberg: Fixed. It's mostly because it didn't exist when I wrote this code. I've updated it to use reset and also wrote a test that `NewArray` causes the array to be reset.

jsternberg:
I was thinking about adding generation, but it would add a dependency on `tmpl` that wasn't presently available so I didn't add it.

internal/mutable/numericarray_test.go
```diff
@@ -0,0 +1,632 @@
...
+	// Construct an array. This should reset the builder.
+	b.NewArray().Release()
+
+	// Append a value to the builder again.
+	b.Append(2)
+	if got, want := b.Len(), 1; got != want {
+		t.Fatalf("unexpected length -want/+got:\n\t- %d\n\t+ %d", want, got)
+	}
+
+	// Constructing the array creates an arrow array of length 2 with the same values.
```
affo: "length 1"

internal/mutable/numericarray.go
```diff
@@ -0,0 +1,397 @@
...
+// Retain will retain a reference to the builder.
+func (b *Float64Array) Retain() {
+	atomic.AddInt64(&b.refCount, 1)
+}
+
+// Release will release any reference to data buffers.
+func (b *Float64Array) Release() {
+	if atomic.AddInt64(&b.refCount, -1) == 0 {
+		if b.data != nil {
+			b.data.Release()
```
affo: I think you forgot to do it for floats and uints.

PR #1547, opened 2019-07-17
feat(rust/scanner): scanner returns positioning

affo:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/internal/rust/scanner/src/lib.rs
+++ b/internal/rust/scanner/src/lib.rs
@@ -15,12 +15,20 @@ pub struct Scanner {
     token: T,
     ts: u32,
     te: u32,
+    lines: Vec<u32>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Position {
+    pub line: u32,
+    pub column: u32,
 }
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct Token {
     pub tok: T,
     pub lit: String,
+    pub pos: u32,
 }
 
 impl Scanner {
@@ -37,6 +45,7 @@ impl Scanner {
             eof: end,
             ts: 0,
             te: 0,
+            lines: vec![0],
             token: T_ILLEGAL,
             checkpoint: ptr,
         };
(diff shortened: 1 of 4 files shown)


internal/rust/scanner/src/scanner.rl
```diff
@@ -128,14 +161,30 @@ int scan(int with_regex, const char **pp, const char *data, const char *pe, cons
...
     *token = tok;
     *token_start = ts - data;
     *token_end = te - data;
+    *newlines_len = no_nls;
+
+    // Now that the state machine has created a dynamic list of newline offsets (in reverse order),
+    // we can take that list and copy to an array for later use in Rust.
+    if (no_nls > 0) {
+        // We can now allocate contiguous memory given that we know the number of newlines found.
+        unsigned int *arr_nls = (unsigned int *) calloc(no_nls, sizeof(unsigned int));
```
affo: This call to `calloc` is of paramount importance instead of `int arr_nls[no_nls`. I should add a comment on that.

internal/rust/scanner/src/scanner.rl
```diff
@@ -7,7 +8,8 @@
 
     include WChar "unicode.rl";
 
-    newline = '\n';
+    # For comments on newlines tracking, see below.
+    newline = '\n' @{ push(&nls, fpc - data + 1); no_nls++; };
```
affo: I should also comment on `+1`. I want to include the `\n` character.

internal/rust/scanner/src/lib.rs
```diff
@@ -15,12 +15,20 @@ pub struct Scanner {
...
+pub struct Position {
+    pub line: u32,
+    pub column: u32,
 }
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct Token {
     pub tok: T,
     pub lit: String,
+    pub pos: u32,
```
wolffcm: It seems like instead of storing the file offset inside the token, you could just keep the start `Postition` (including the line and column) and the end `Position` directly in the token returned from `scan()`.

Then you wouldn't need to do the mapping from file offset to `Position` which does a binary search for each call to `pos()` and has (log n) performance.

We discussed how this might increase the memory footprint since tokens may be cloned frequently.  It seems like if tokens are immutable, this could be avoided.

Can you file an issue for this?

affo: done https://github.com/influxdata/flux/issues/1608.

Why do you think that if tokens are immutable this could be avoided?

Maybe we should just change the implementation to work with references (ownership problems) in order to avoid cloning, given that tokens are immutable.

wolffcm: Yeah, I just mean that if an instance of token never changes, then only one instance needs to be created (no cloning needed), since all the readers can share the same instance.  Then you could just use references everywhere.

Issue #1533, opened 2019-07-16
date.truncate & universe.truncateTimeColumn


ewendai:
`date.truncate` takes in `time Time`, `unit Duration` and outputs a time truncated to that unit

`universe.truncateTimeColumn` takes in `column Time`, `unit Duration` and truncates the given column

Issue #1122, opened 2019-04-03
Create a "label" type


nathanielc:
I propose that we add a `label` type to the Flux type system. A label type is a statically known string value. Label types are used to access keys of objects. In order for the type system to be sound and support inference we must statically know all object properties that are accessed statically. 

Using a label type allows for defining polymorphic functions that operate with different labels while still remaining sound.

Many transformations functions accept a `column` argument that defines which column they operate on. This is only possible if the value of `column` is known statically. Using a `label` type explicitly codifies this requirement.

For example:

```
builtin mean : (tables : <-stream, column : label) -> stream
```

The above means that `column` must be known statically, so this is valid:

```
from(bucket:"foo") 
    |> ...
    |> mean(column: "usage_idle")
```

but the following would not be allowed.

```

builtin bar : () -> string

from(bucket:"foo") 
    |> ...
    |> mean(column: bar() )
```

As string is not a label as a string value is not known until runtime.

This is a rough draft a proposal at this point and needs to be worked out before we know if it is possible.

aanthony1243:
if label is a valid type, and can be returned by a function, then we run into the same issue.  

but if you are proposing that labels may only exist as literals, then this could work OK.  If we did this, I would prefer a different syntax for labels, maybe `<label>` ?

nathanielc:
@aanthony1243 Are you suggesting that object keys always have `<>` surrounding them? Like:

```
foo =  {
   <mykey>: "some value",
   <another key>: 42
}
```

Or where do you expect the `<label>` syntax to come up? 


> if label is a valid type, and can be returned by a function, then we run into the same issue.

Label will be a valid type but it is required that its value be known statically. So we can decide how far the ocmpiler looks for static values but it is not unreasonable to inspect the body of a function and see that its value is static.

```
defaultValueColumn = () => "_value"

timeShift(column: defaultValueColumn()) // this could work as its value is known statically

```

I am not saying that we should do it this way but that it is possible.

aanthony1243:
I was thinking something like `mean(column: <usage_idle>)`   I don't know that I really like this idea, but I do think that if they look like strings, and feel like strings, users will treat them like strings.

pauldix:
What's the underlying reason for doing this? Why must the type be known statically? This feels like more syntax and harder for the user. What about the case where a `pivot` is done and string values become column names (like on `_field`)?

mcassaniti:
[...]

PR #1554, opened 2019-07-17
Feat/truncate

ewendai:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

#1533

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1132,6 +1132,15 @@ These are builtin functions that all take a single `time` argument and return an
 * `month` int
     Month returns the month of the year for the provided time in the range `[1-12]`.
 
+#### truncate
+
+`date.truncate` takes in a time t and a Duration unit and returns the given time 
+truncated to the given unit.
+
+Examples: 
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1s)` returns time `2019-06-03T13:59:01.000000000Z`
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1m)` returns time `2019-06-03T13:59:00.000000000Z`
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1h)` returns time `2019-06-03T13:00:00.000000000Z`
 
 ### System Time
 
@@ -3566,6 +3575,18 @@ r0 = from(bucket:"telegraf/autogen")
 // use values
 x = r0._field + "--" + r0._measurement
 ```
+
+##### truncateTimeColumn 
+
+Truncates all entries in a given time column to a specified unit.
+
+Example: 
+```
+from(bucket:"telegraph/autogen")
+    |> truncateTimeColumn(unit: 1s)
+```
+
+Currently, `truncateTimeColumn` only works with the default time column `_time`.
 
 #### Type conversion operations
 
--- a/stdlib/date/date.go
+++ b/stdlib/date/date.go
@@ -155,10 +155,11 @@ func init() {
 			}),
 			func(args values.Object) (values.Value, error) {
 				v, ok := args.Get("t")
-				u, unitOk := args.Get("unit")
 				if !ok {
 					return nil, errors.New(codes.Invalid, "missing argument t")
 				}
+
+				u, unitOk := args.Get("unit")
 				if !unitOk {
 					return nil, errors.New(codes.Invalid, "missing argument unit")
 				}
--- a/stdlib/date/date.go
+++ b/stdlib/date/date.go
@@ -2,7 +2,6 @@ package date
 
 import (
 	"fmt"
-
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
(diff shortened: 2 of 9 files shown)


docs/SPEC.md
```diff
@@ -1132,6 +1132,12 @@ These are builtin functions that all take a single `time` argument and return an
 * `month` int
     Month returns the month of the year for the provided time in the range `[1-12]`.
 
+#### truncate
+
+`date.truncate` takes in a time t and a Duration unit and returns the given time 
+truncated to the given unit.
+
+Example: `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1s)` returns time `2019-06-03T13:59:01.000000000Z`
```
affo: Could you give a more insightful example? For  example with minute or hour?

If you add examples as in your unit tests, that would be great!

stdlib/universe/universe.flux
```diff
@@ -81,7 +82,7 @@ builtin inf
 builtin length // length function for arrays
 builtin linearBins
 builtin logarithmicBins
-builtin sleep // sleep is the identity function with the side effect of delaying execution by a specified duration.
+builtin sleep // sleep is the identity function with the side effect of delaying execution by a specified duration
```
affo: Why is this here? A bad rebase or something else?

docs/SPEC.md
```diff
@@ -3567,6 +3573,18 @@ r0 = from(bucket:"telegraf/autogen")
...
+
+Truncates all entries in a given time column to a specified unit.
+
+Example: 
+```
+from(bucket:"telegraph/autogen")
+    |> truncateTimeColumn(timeColumn: "_time", unit: 1s)
+```
+
+Currently, `truncateTimeColumn` only works properly if the given `timeColumn` is "_time".
```
affo: Yeah, I see your point. I would rephrase into:

```
Currently, `truncateTimeColumn` only works with the default time column `_time`.
```

and I would change the example to:

```
from(bucket:"telegraph/autogen")
    |> truncateTimeColumn(unit: 1s)
```

If we omit `timeColumn` from the doc, it is not a big deal, because, for now, nobody can use it.

You should also add a table with the parameters as the other functions, don't you think?

As a final note, I actually see a way to make it work with any column:

```
truncateTimeColumn = (timeColumn="_time", unit, tables=<-) => tables
        |> map(fn:(r) => ({r with __truncated_time: date.truncate(t: r._time, unit: unit)}))
        |> duplicate(column: "__truncated_time", as: timeColumn) // overwrites timeColumn
        |> drop(columns: ["__truncated_time"])
```

If the  user already has a column called `__truncated_time` that could be an issue though 😓 . What do you think?

FYI @aanthony1243 .

stdlib/date/date.go
```diff
@@ -147,6 +146,29 @@ func init() {
...
+			"truncate",
+			semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+				Parameters: map[string]semantic.PolyType{"t": semantic.Time, "unit": semantic.Duration},
+				Required:   semantic.LabelSet{"t", "unit"},
+				Return:     semantic.Time,
+			}),
+			func(args values.Object) (values.Value, error) {
+				v, ok := args.Get("t")
+				u, unitOk := args.Get("unit")
+				if !ok {
```
affo: Can you move this directly below line 157 and directly use `GetRequiredTime`/`Duration`. That would simplify logic.

stdlib/universe/universe.flux
```diff
@@ -1,6 +1,7 @@
 package universe
 
 import "system"
+import "date"
```
affo: this is great. Makes me think why don't we `import "math"` but we add it programmatically to the prelude @aanthony1243 .

stdlib/universe/universe.flux
ewendai: I deleted the period

stdlib/date/date.go
ewendai: `args.GetRequiredTime` and `args.GetDuration` don't pop up when I type it in?

affo: It's here: https://github.com/influxdata/flux/blob/c7e454395264dcf1cb23f9889ef7bbbaa947fc4c/compile.go#L682

docs/SPEC.md
```diff
@@ -1137,7 +1137,10 @@ These are builtin functions that all take a single `time` argument and return an
 `date.truncate` takes in a time t and a Duration unit and returns the given time 
 truncated to the given unit.
 
-Example: `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1s)` returns time `2019-06-03T13:59:01.000000000Z`
+Examples: 
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1s)` returns time `2019-06-03T13:59:01.000000000Z`
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1m)` returns time `2019-06-03T13:59:00.000000000Z`
+- `truncate(t: "2019-06-03T13:59:01.000000000Z", unit: 1h)` returns time `2019-06-03T13:00:00.000000000Z`
```
affo: Cool, thank you :)

stdlib/universe/universe.flux
aanthony1243: never thought about that.  I'm going to test it

docs/SPEC.md
ewendai: `date.truncate` still takes in only `_time`

this just changes `timeColumn` from the column you want to truncate to what the resulting column is named

stdlib/date/date.go
ewendai: I think `args` is `*TableObject` not `Arguments`, so how would that work?

affo: Oh I see, sorry about that! This code is ok though :)

stdlib/universe/universe.flux
affo: This is  mistery. In `master` is there. I  think that if  you rebase that would go away. Anyways this shouldn't be a problem.

affo: Oooooh now I got it. The period `.` 😅 

I haven't noticed that. I am using the split view in Github and it's harder to catch that.

docs/SPEC.md
affo: Yeah, I am sorry, I meant:

```
truncateTimeColumn = (timeColumn="_time", unit, tables=<-) => tables
        |> duplicate(column: timeColumn, as: "__truncated_time")
        |> map(fn:(r) => ({r with __truncated_time: date.truncate(t: r.__truncated_time, unit: unit)}))
        |> duplicate(column: "__truncated_time", as: timeColumn) // overwrites timeColumn
        |> drop(columns: ["__truncated_time"])
```

Anyways, this is a workaround that could overwrite a column with the same name by the user. We need to support labels for this #1122 .

Only brainstorming 😄

PR #1559, opened 2019-07-17
fix(values): throw appropriate error for division by zero

sauren-khosla:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/values/binary.go
+++ b/values/binary.go
@@ -8,7 +8,7 @@ import (
 	"github.com/influxdata/flux/semantic"
 )
 
-type BinaryFunction func(l, r Value) Value
+type BinaryFunction func(l, r Value) (Value, error)
 
 type BinaryFuncSignature struct {
 	Operator    ast.OperatorKind
@@ -31,9 +31,9 @@ func LookupBinaryFunction(sig BinaryFuncSignature) (BinaryFunction, error) {
 // If either value is null, then it will return null.
 // Otherwise, it will invoke the function to retrieve the result.
 func binaryFuncNullCheck(fn BinaryFunction) BinaryFunction {
-	return func(lv, rv Value) Value {
+	return func(lv, rv Value) (Value, error) {
 		if lv.IsNull() || rv.IsNull() {
-			return Null
+			return Null, nil
 		}
 		return fn(lv, rv)
 	}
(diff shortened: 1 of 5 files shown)


values/binary_test.go
```diff
@@ -552,8 +556,11 @@ func TestBinaryOperator(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-
-			if want, got := Value(tt.want), fn(left, right); !ValueEqual(want, got) {
+			want := Value(tt.want)
+			got, err := fn(left, right)
+			if err != nil {
+				fmt.Printf("Dividing by zero")
```
sauren-khosla: I'm not positive if this will accurately account for the new test cases on lines 75 and 88. Without this if statement, I get a panic (seg fault).

values/binary.go
```diff
@@ -53,128 +53,123 @@ var binaryFuncLookup = map[BinaryFuncSignature]BinaryFunction{
...
+		return NewUInt(l / r), nil
 	},
-	{Operator: ast.DivisionOperator, Left: semantic.Float, Right: semantic.Float}: func(lv, rv Value) Value {
+	{Operator: ast.DivisionOperator, Left: semantic.Float, Right: semantic.Float}: func(lv, rv Value) (Value, error) {
 		l := lv.Float()
 		r := rv.Float()
 		if r == 0 {
-			// TODO(#38): reject divisions with a constant 0 divisor.
-			return NewFloat(math.NaN())
+			return nil, fmt.Errorf("cannot divide by zero")
```
wolffcm: Is this what we want to do for floating point division?  It seems like `NaN` is a reasonable thing to return.  Do any of the issues related to division make a distinction between float and integer division?

values/binary.go
```diff
@@ -53,128 +53,123 @@ var binaryFuncLookup = map[BinaryFuncSignature]BinaryFunction{
...
-		return NewUInt(l % r)
+		return NewUInt(l % r), nil
 	},
-	{Operator: ast.ModuloOperator, Left: semantic.Float, Right: semantic.Float}: func(lv, rv Value) Value {
+	{Operator: ast.ModuloOperator, Left: semantic.Float, Right: semantic.Float}: func(lv, rv Value) (Value, error) {
 		l := lv.Float()
 		r := rv.Float()
 		if r == 0 {
-			return NewFloat(math.NaN())
+			return nil, fmt.Errorf("cannot divide by zero")
```
wolffcm: If we're going to return an error then this error should say "modulo" instead of "divide."

values/binary_test.go
wolffcm: It would be good to account for a possible error produced by the test case by adding an `err` field to the struct slice being looped over at line 26.  Then you can change the logic here to look for an error coming from `fn` if `tc.err` is not nil.

An example of a test written this way is here: `execute/executor_test.go`, but there's several like this.

values/binary.go
sauren-khosla: #Flux concluded that throwing an error is the best option to avoid the confusion of IEEE754. Marking as resolved.

values/binary_test.go
```diff
@@ -552,8 +557,13 @@ func TestBinaryOperator(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-
-			if want, got := Value(tt.want), fn(left, right); !ValueEqual(want, got) {
+			want := Value(tt.want)
+			got, err := fn(left, right)
+			if tt.wantErr != nil {
+				if err == nil {
```
wolffcm: This logic doesn't look quite right yet.  If `fn` returns an error and one wasn't expected, then the test should fail with a message like `unexpected error: <errormsg>`.

PR #1561, opened 2019-07-18
test(internal/mutable): add test for mixed reserve and resize

affo:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/internal/mutable/numericarray.go
+++ b/internal/mutable/numericarray.go
@@ -81,6 +81,7 @@ func (b *Int64Array) reset() {
 	b.data.Release()
 	b.data = nil
 	b.rawData = nil
+	b.length = 0
 }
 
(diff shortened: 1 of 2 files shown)


internal/mutable/numericarray.go
```diff
@@ -106,9 +107,10 @@ func (b *Int64Array) Reserve(n int) {
 		if b.data == nil {
 			b.init()
 		}
-		capacity := arrow.Int64Traits.BytesRequired(n)
+		length := len(b.rawData) + n
+		capacity := arrow.Int64Traits.BytesRequired(length)
```
affo: I see, `data.Reserve(...)` is not relative. It's builders that make it relative 👍

PR #1564, opened 2019-07-18
chore(rust): define Rust to npm workflow

chnn:
This PR adds a publish script and some documentation so that myself or others on the applications team can publish new versions of the Rust/WASM parser to npm. I expect we will iterate further on this workflow in the future.

I also added some optimization settings to the parser `Cargo.toml` manifest that shrink the size of the built `.wasm` file. If you're curious about the result, I ran some experiements:

- default cargo release profile, gzipped: 318k
- with `lto = true`, gzipped: 312k
- with `lto = true`, `opt=level = "s"`, gzipped: 270k
- with `lto = true`, `opt=level = "z"`, gzipped: 254k
- with `lto = true`, `opt=level = "s"`, `wasm-opt`, gzipped: 228k

In addition, I made some minor tweaks to the `site` directory. I want to use it as an example of integrating the WASM parser into webpack for some others on my team.


chnn:
> Is the workflow that someone will check out the repo and run the publish script and manually increment the version string?

Yes, unless we want to automate it as part of CI.

> Do we want to make the version string mirror the Flux repo version?

> Do we want to auto publish the module as part of the CI process when we cut new releases of Flux?

Sounds convenient, but perhaps not worth the effort at the moment. I'm happy with keeping the manual workflow for now, and revisiting it later if we feel it would be worth it to automate.

PR #1565, opened 2019-07-19
feat(internal/mutable): tracked slice

affo:
This is a tracked slices that uses Flux `memory.Allocator` and defers arrow memory tracking until the user requests for an arrow array.

This is a good use case for slices that remain slices. This happens, for example, in Holt Winters.

Benchmarks results:

```
pkg: github.com/influxdata/flux/internal/mutable
BenchmarkInt64SizedArray-12                     	   50000	     36218 ns/op	  102432 B/op	     301 allocs/op
BenchmarkInt64SizedSlice-12                     	   50000	     23274 ns/op	   94432 B/op	     201 allocs/op
BenchmarkInt64ArrayWCapacity-12                 	   50000	     35736 ns/op	  102432 B/op	     301 allocs/op
BenchmarkInt64SliceWCapacity-12                 	   50000	     29949 ns/op	  184032 B/op	     201 allocs/op
BenchmarkInt64ArrayAddValues-12                 	  100000	     13249 ns/op	   24736 B/op	       5 allocs/op
BenchmarkInt64SliceAddValues-12                 	  100000	     14074 ns/op	   32848 B/op	       5 allocs/op
BenchmarkInt64ArrayAddValuesKnownCapacity-12    	  100000	     13823 ns/op	   24736 B/op	       5 allocs/op
BenchmarkInt64SliceAddValuesKnownCapacity-12    	  100000	     10947 ns/op	   16464 B/op	       3 allocs/op
BenchmarkInt64ArrayArrowArray-12                	  300000	      5877 ns/op	    8528 B/op	       7 allocs/op
BenchmarkInt64SliceArrowArray-12                	  200000	      9547 ns/op	   26560 B/op	      13 allocs/op
```

This implementation performs better than the other implementation with the exception of reserving (requires growing the slice with -> copying) and creating the arrow array. Anyways, I would expect that the common usage for this implementation would be instantiating with a known size and capacity and not reserving any additional space.

I think that no implementation should survive the other. They should both exist. I see that yours is the best way of crafting an arrow array with a bit of overhead while manipulating, but significant less overhead when creating the final arrow array (perfect for creating something for the `TableBuilder`). Mine is for internal usage in algorithms that do not need a final arrow array and don't need overhead while manipulating (maybe this implementation doesn't even need to implement the very same interface of yours).

I won't proceed any further until I don't hear your opinion @jsternberg .

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/arrow/allocator.go
+++ b/arrow/allocator.go
@@ -10,6 +10,17 @@ type allocator struct {
 	alloc *memory.Allocator
 }
 
+func NewAllocator(a *memory.Allocator) arrowmemory.Allocator {
+	var alloc arrowmemory.Allocator = arrowmemory.NewGoAllocator()
+	if a != nil {
+		alloc = &allocator{
+			Allocator: alloc,
+			alloc:     a,
+		}
+	}
+	return alloc
+}
+
 func (a *allocator) Allocate(size int) []byte {
 	if err := a.alloc.Allocate(size); err != nil {
 		panic(err)
--- a/internal/mutable/benchmarks_test.go
+++ b/internal/mutable/benchmarks_test.go
@@ -1,0 +1,189 @@
+package mutable_test
+
+import (
+	"testing"
+
+	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux/internal/mutable"
+	fluxmemory "github.com/influxdata/flux/memory"
+)
+
+func createArraysWithSize(n, s int) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	for i := 0; i < n; i++ {
+		b := mutable.NewInt64Array(mem)
+		b.Resize(s)
+		b.Release()
+	}
+}
+
+func createSlicesWithSize(n, s int) {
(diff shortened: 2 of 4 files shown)


jsternberg:
I think my main concern is that I want to know why the arrow versions perform slower because they should be the exact same. There shouldn't be a difference between the two. There's no special encoding step. The `memory.Buffer` is just an abstraction for managing a byte array. The `NewArray()` method that's included is mostly just because it can be done easily and it isn't the primary point of these methods so I don't see a difference between these two implementations.

I have some other reservations too, but I would like to put some effort into figuring out why there's a performance difference. I noticed a few problems with the implementation, but I'd like to first figure out what is causing the slowdown in the `memory.Buffer` versions as it may just be a bug.

affo:
After rebasing of the fix to `Reserve` in mutable arrays (https://github.com/influxdata/flux/pull/1561) Holt Winters runs faster (the transformation that generated all this trouble).

@jsternberg and I agree that it is better to have a single implementation. Having multiple ones could be confusing and improperly used.

Still we have to explain benchmarks.

PR #1572, opened 2019-07-22
feat(rust): complete source location for rust parser

affo:
Steps:
- [x] make scanner return token positions #1547 ;
- [x] add source location to AST nodes;
- [x] port parser logic for source location in Go to Rust;
- [x] make parser tests pass;


affo:
@jsternberg had to rebase and fix some problems with the new string interpolation bits. Now it's ok

PR #1579, opened 2019-07-22
fix(stdlib/testing): add range calls to some PromQL e2e tests

wolffcm:


wolffcm:
Just disabling these tests outside of flux.

PR #1588, opened 2019-07-23
docs: calculate deadman alerts in Flux

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/DeadmanAlert.md
+++ b/docs/DeadmanAlert.md
@@ -6,15 +6,16 @@ For example, say you have the following task definition:
 ```
 option task = {
     ...
-    every: 15m,
+    period: 1h,
+    every:  15m,
     ...
 }
 from(bucket: "telegraf/autogen")
-    |> range(start: -1h)
+    |> range(start: -period)
     |> filter(fn: (r) => r._measurement == "cpu")
 ```
 
-In order to be alerted when the above task stops reporting data for a group, you would define a deadman alert like so:
+In order to be alerted when the above task stops reporting data for a group, you would define a deadman alert using the `deadman` function like so:
 ```
 option task = {
     ...
@@ -23,77 +24,33 @@ option task = {
     ...
 }
 from(bucket: "telegraf/autogen")
-    |> range(start: -(task.period + task.every))
+    |> range(start: -perod)
     |> filter(fn: (r) => r._measurement == "cpu")
-    |> deadman(off: task.every)
+    |> deadman(d: task.every)
+    |> alert(crit: (r) => true)
 ```
 
-where the deadman function is defined as follows:
+where the `deadman` function is defined as follows:
 ```
-deadman(off, tables=<-) => {
-
-    r = tables
-        |> group_keys()
-
-    s = tables
-        |> filter(fn: (r) => r._time > now() - off)
-        |> group_keys()
-
-    return diff(r:r, s:s) |> alert(crit: (r) => true)
-}
+deadman = (d, tables=<-) => tables
+    |> sort(columns: ["_time"])
+    |> last()
+    |> filter(fn: (r) => r._time < now() - d)
 ```
 
-Note `group_keys` takes in a stream of tables and returns the group keys of those tables in the output:
-```
-group_keys = (tables=<-) =>
-    |> keys()
-    |> limit(n:1)
-    |> drop(columns: ["_value"])
-```
+Note the `deadman` function takes a stream of tables and a duration and returns all groups **not** observed within the interval defined by `[now() - d, now()]`.
+For example, given a stream called `tables`, grouped by (`_measurement`, `host`):
 
-Lets walk through an example of a query that returns tables grouped by (`_measurement`, `host`).
-Assume that query produces the following tables over the first period:
+| _time      | _measurement | host | _value |
+| ---------- | ------------ | ---- | ------ |
+| now() - 5s | cpu          | A    | 56     |
+| now() - 3s | cpu          | B    | 17     |
+| now() - 1s | cpu          | C    | 18     |
 
-| _measurement | host | _value |
-| ------------ | ---- | ------ |
-| cpu          | A    | 56     |
-| cpu          | B    | 17     |
+`tables |> deadman(d: 4s)` produces a non-empty result:
 
-But over the second period it doesn't receive a value for `host=A`:
+| _time      | _measurement | host | _value |
+| ---------- | ------------ | ---- | ------ |
+| now() - 5s | cpu          | A    | 56     |
 
-| _measurement | host | _value |
-| ------------ | ---- | ------ |
-| cpu          | B    | 22     |
-
-`r` is equal to:
-
-| _measurement | host |
-| ------------ | ---- |
-| cpu          | A    |
-| cpu          | B    |
-
-`s` is equal to:
-
-| _measurement | host |
-| ------------ | ---- |
-| cpu          | B    |
-
-And `diff(r:r, s:s)` is equal to:
-
-| _measurement | host |
-| ------------ | ---- |
-| cpu          | A    |
-
-As a result, a non-empty stream is passed to the alert function and an alert is triggered.
-
-Intuitively the deadman alert computes the set difference between the groups present in each of the two most recent intervals of data.
-If the difference is non-empty, this means there is at least one group that stopped reporting in the most recent interval and an alert is fired.
-
-The `diff` function takes two streams, `r` and `s`, and returns the rows of `r` that are not in `s`.
-`diff` is equivalent to set difference and is defined as follows:
-```
-diff = (r, s) => join.leftAnti(left: r, right: s)
-```
-
-where `join.leftAnti` performs a left anti-join of its input tables.
-That is, it returns all of the rows in `left` that do not join with any of the rows in `right`.
+And therefore `tables |> deadman(d: 4s) |> alert(crit: (r) => true)` triggers an alert that the group defined by `_measurement=cpu,host=A` stopped reporting data.
--- a/docs/DeadmanAlert.md
+++ b/docs/DeadmanAlert.md
@@ -1,0 +1,58 @@
+### Deadman Alert
+
+The `deadman` function is used in the context of a task to alert when a group stops reporting data.
+For example, say you have the following task definition:
+
+```
+option task = {
+    ...
+    every: 15m,
+    ...
+}
+from(bucket: "telegraf/autogen")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._measurement == "cpu")
+```
+
+In order to be alerted when the above task stops reporting data for a group, you would define a deadman alert like so:
+```
+option task = {
+    ...
+    every: 15m,
+    ...
+}
+f = (start, stop) =>
+    from(bucket: "telegraf/autogen")
+        |> range(start: start, stop: stop)
+        |> filter(fn: (r) => r._measurement == "cpu")
+
+deadman(f: f, period: 1h, offset: task.every)
+```
+
+where the deadman function is defined as follows:
+```
+deadman = (f, period, offset) => {
+
+    r = f(start: now() - offset - period, stop: now() - offset)
+        |> keys()
+        |> drop(columns: ["_value"])
+
+    s = f(start: now() - period, stop: now())
+        |> keys()
+        |> drop(columns: ["_value"])
+
+    return diff(r: r, s: s) |> alert(crit: (r) => true)
+}
+```
+
+Intuitively the deadman alert computes the set difference between the groups present in each of the two most recent intervals of data.
+If the difference is non-empty, this means there is at least one group that stopped reporting in the most recent interval and an alert is fired.
+
+The `diff` function takes two streams, `r` and `s`, and returns the rows of `r` that are not in `s`.
+`diff` is equivalent to set difference and is defined as follows:
+```
+diff = (r, s) => join.leftAnti(left: r, right: s)
+```
+
+where `join.leftAnti` performs a left anti-join of its input tables.
+That is, it returns all of the rows in `left` that do not join with any of the rows in `right`.
--- a/docs/DeadmanAlert.md
+++ b/docs/DeadmanAlert.md
@@ -45,6 +45,41 @@ deadman = (f, period, offset) => {
 }
 ```
 
+Lets walk through an example where `f` returns tables grouped by (`_measurement`, `host`).
+Assume `f` produces the following tables over the first period:
+
+| _measurement | host | _value |
+| ------------ | ---- | ------ |
+| cpu          | A    | 56     |
+| cpu          | B    | 17     |
+
+But over the second period it doesn't receive a value for `host=A`:
+
+| _measurement | host | _value |
+| ------------ | ---- | ------ |
+| cpu          | B    | 22     |
+
+`r` is equal to:
+
+| _measurement | host |
+| ------------ | ---- |
+| cpu          | A    |
+| cpu          | B    |
+
+`s` is equal to:
+
+| _measurement | host |
+| ------------ | ---- |
+| cpu          | B    |
+
+And `diff(r:r, s:s)` is equal to:
+
+| _measurement | host |
+| ------------ | ---- |
+| cpu          | A    |
+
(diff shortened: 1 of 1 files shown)


docs/DeadmanAlert.md
```diff
@@ -0,0 +1,58 @@
...
+        |> drop(columns: ["_value"])
+
+    return diff(r: r, s: s) |> alert(crit: (r) => true)
+}
+```
+
+Intuitively the deadman alert computes the set difference between the groups present in each of the two most recent intervals of data.
+If the difference is non-empty, this means there is at least one group that stopped reporting in the most recent interval and an alert is fired.
+
+The `diff` function takes two streams, `r` and `s`, and returns the rows of `r` that are not in `s`.
```
aanthony1243: we have a `diff` function, this seems simpler, but maybe what we have would work?  currently it returns both sides.  It's interesting to consider it being a 1-way diff.

jlapacik: Yeah this is a set difference operation. SQL calls it `MINUS` so maybe that would be a better name. I don't think we'd be able to use `testing.diff` here though just because we need it to be 1-way.

jlapacik:
@aanthony1243 is your idea to basically create a materialized view of the missing series keys?

aanthony1243:
>  is your idea to basically create a materialized view of the missing series keys?

I guess so?  your version is reproducible for all time.   mine is actually the ephemeral one now that I think about it.  

I'm probably getting into implementation details now, so feel free to just 👍  or 👎  and move on without discussing.   But you can do the round/dedupe trick in your version too which would make that `join` cheaper.   

Also, if it's a set-difference then let's call it that, we already have `union`  so it makes sense to add that.

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,58 @@
...
+
+deadman(f: f, period: 1h, offset: task.every)
+```
+
+where the deadman function is defined as follows:
+```
+deadman = (f, period, offset) => {
+
+    r = f(start: now() - offset - period, stop: now() - offset)
+        |> keys()
```
nathanielc: What is the `keys()` function doing here? Can you provide trivial example tables for how this will detect missing series?

jlapacik: I can provide an example, but the pattern:
```
... |> keys() |> drop(columns: ["_value"])
```
just reduces a stream down to its group key values (a table per group key). And then `diff` outputs the tables (group keys) that are in the first set, but not the second. And that gives you the series(groups) that stopped reporting.

Note I'm giving `diff` set semantics instead of bag semantics. I.e it performs an implicit deduplication of the rows. However, as part of this work, I think it would be best to add a transformation that deduplicates an entire stream. That way we can give `diff` bag semantics and perform deduplication before calling `diff`.

jlapacik: I've added an example that hopefully clarifies things.

docs/DeadmanAlert.md
aanthony1243: fwiw `diff` has a column containing `+` and `-` for which table it came from.  so you could run `diff` and then filter out either the `+`  or the `-`.    it's basically doing  `(X Union Y) - (X  intersect Y)`

however, this is also a good opportunity to add transformations for `intersect`  and `difference` .

docs/DeadmanAlert.md
aanthony1243: @jlapacik I think I follow @nathanielc  here.   `keys` gets you the set of column names on a table, not the values.  

If you just want one row per series, then just do `limit(n:1)`

as mentioned earlier,  maybe: 
```
s = f(start: now() - period, stop: now())
        |> limit(n:1)
        |> map(fn: (r)  => ({r with _time  =  date.truncate(t: r._time, unit: 1h)}))
```
i wouldn't drop anything because the user may do that already.  or they want `_value` as part of the alert.    perhaps we need to add a parameter for what identifies a series?  and do  `keep(columns: seriesID)`

affo: I think you are looking for `keyValues`.

jlapacik: Sorry there appears to be a lot of confusion over my use of `keys`. For each table that `keys` consumes, it outputs a single table consisting of the group key of the input table plus a `_value` column. By dropping the `_value` column you're just left with the group key (possibly duplicated over several rows).

@aanthony1243 thank you for the `limit` suggestion. I should use limit to reduce each output table to a single row representing the group key.

@affo you're correct in that `keyValues` would work as well. It functions similarly to `keys`. It just adds a `_key` and a `_value` column to the output table. These columns aren't really important. What we're after is the group key(series key).

@aanthony1243 to your point about dropping `_value`. `_value` is added by the `keys` function which is encapsulated within the deadman function. This is completely hidden from the user. What's important is the group/series keys. Deadman only cares when series stop reporting. In this context, only the group/series key is important. All other columns can and should be dropped so that we do a direct comparison of the series present in the first period that are not present in the second period.

Does this clarify why I'm using the `keys` function?

On a side note, I'd like to have a physical operation that retrieves just the series keys from storage.

affo: Oh, I see! Now I got it.

affo: I'd say that:

```
keep_group_cols = (tables=<-) => tables
    |> keys() // could even use `columns`, the point is to have a schema with only group key cols
    |> limit(n: 1)
    |> drop(columns: ["_value"])

r = f(start: now() - offset - period, stop: now() - offset) |> keep_group_cols()
s = f(start: now() - period, stop: now()) |> keep_group_cols()
```

Would help a lot

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,99 @@
...
+    return diff(r: r, s: s) |> alert(crit: (r) => true)
+}
+```
+
+Note `group_keys` takes in a stream of tables and returns the group keys of those tables in the output:
+```
+group_keys = (tables=<-) =>
+    |> keys()
+    |> limit(n:1)
+    |> drop(columns: ["_value"])
```
nathanielc: Should this function end with `group()` to put all rows into a single table? Otherwise the diff computation below is per table and it will have to cache all tables in order to compute the diff.

nathanielc:
Some thoughts, this design is simpler that one based on counts, meaning that it can't tell you how many points or the rate of points a series is used to getting and it can't alert on a drop in that rate. It can only alert on the complete absence of a series. I think this is a good design. Kapacitor allowed alerting on a change in rate of points as well as a complete absence. I think coupling those made things difficult.

docs/DeadmanAlert.md
jlapacik: As each table will only have a single row, my assumption is that the overhead incurred by keeping each series key in its own table will be minimal. As cardinality increases that assumption might become tenuous. However by not performing a destructive operation like `group()`, we ensure the `deadman` function works for all heterogeneous data.

This of course assumes that `diff` or left anti-join works on heterogeneous data.

docs/DeadmanAlert.md
aanthony1243: if we generalize this concept, it's really saying "given this query, show me the signals that I heard in the previous time period that i haven't heard in the current time period".  

You are making influx-specific assumptions about what a `signal` is.  It may be safe to say, "signals are identified by the group key"  but this requires that the user has the data pre-grouped. Do we want to have the user tell the deadman what the signal-identifying columns are?

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,99 @@
...
+    from(bucket: "telegraf/autogen")
+        |> range(start: start, stop: stop)
+        |> filter(fn: (r) => r._measurement == "cpu")
+
+deadman(f: f, period: 1h, offset: task.every)
+```
+
+where the deadman function is defined as follows:
+```
+deadman = (f, period, offset) => {
```
jsternberg: I'm not sure I like this deadman function syntax. It seems to require too many parameters and it seems to be a bit too tied to `from() |> range()`.

I'm curious if there's a way to make it so deadman is chainable. It seems that the way deadman works is it takes two different streams and checks to see if one of the group keys does not exist in the second stream. It seems to me this is possible with just a single call to from. Imagine if instead of using `period` and `offset`, we instead did:

```
deadman = (for) => ...

    |> range(start: -period)
    |> deadman(for: offset)
```

The deadman could be configured so that it uses the `_time` column for a table and it alerts if there are no points within the last `offset` period. It seems to me that this would be the same thing as the current implementation, but it wouldn't require a specific function to accommodate `range()`.

This wouldn't require `group_keys()` or `diff()` to exist since these are already done by the execution engine.

jlapacik: @jsternberg I definitely like this signature for deadman. But I'm not sure how you would eliminate `group_keys()` and `diff()`. My goal is to have deadman defined in pure Flux. With this new signature, deadman would be defined as follows:

```
deadman(off, tables=<-) => {
    r = tables
        |> group_keys()

    s = tables
        |> filter(fn: (r) => r._time > now() - off)
        |> group_keys()

    return diff(r:r, s:s) |> alert(crit: (r) => true)
}
```

I think this is a much better definition. However we still need `group_keys()` and `diff()`. Unless you mean for it to be implemented in Go, but I would much rather have it implemented in Flux.

jsternberg: I think it could be implemented by adding a new feature to `elapsed()` since `elapsed()` is basically what we want anyway. If we added a new `stop:` parameter to elapsed to indicate an optional stop time, we could do something like this:

```
deadman(tables=<-, for) => elapsed(stop: now()) |> alert(crit: (r) => duration(v: r._value) > for)
```

The `stop` parameter would make it so that elapsed emitted a final row that was the elapsed time since the last entry and the time specified.

Alternatively, maybe this would be better?

```
deadman(tables=<-, for) => limit(n: 1) |> map(fn: (r) => ({r with _value: int(v: now()) - r._time})) |> alert(crit: (r) => duration(v: r._value) > for)
```

This is defined in pure flux and it will take the last row in the table, figure out the difference between the current time and the time of the row, and alert when that difference is greater than the `for` time.

affo: @jsternberg definitely like your idea, and I think a good intermediate step would be to change the signature to what you are suggesting. Your solution has a major pro of not triggering the source twice.

Once we have that signature and we add the capability to `elapsed` we can change `deadman`'s implementation.

jlapacik:
@pauldix I'm unclear on how your query would alert on a series that is present inside of the `stalenessDuration` but not inside of the `deadmanDuration`.

jlapacik:
@pauldix it is clear now and I agree, I like your query without having to do a blocking set difference. I would propose however that instead of a `from` query, `deadman` be defined as a function that can be called on any stream having a `_time` column like so:

```
deadman = (fn, d, tables=<-) => tables
    |> filter(fn: fn)
    |> sort(columns: ["_time"])
    |> last()
    |> filter(fn: (r) => r._time > now() - d)
```

Non-empty results can then be used to alert and subsequently written using `to`. Also  as mentioned by @nathanielc in a slack conversation, the planner can remove `sort` if it determined that the data is already sorted by _time.

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,56 @@
...
+    ...
+    period: 1h,
+    every:  15m,
+    ...
+}
+from(bucket: "telegraf/autogen")
+    |> range(start: -perod)
+    |> filter(fn: (r) => r._measurement == "cpu")
+    |> deadman(d: task.every)
+    |> alert(crit: (r) => true)
```
jsternberg: Should `deadman` include `alert()`? If you have it include `alert()`, you don't need the hacky `(r) => true` to trigger an alert and you can also get output when the deadman returns false (meaning everything is ok) so you can get `ok` values from the alert.

Such as:

```
deadman = (d, tables=<-) => tables
    |> sort(columns: ["_time"])
    |> last()
    |> alert(crit: (r) => r._time < now() - d)
```

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,56 @@
...
+from(bucket: "telegraf/autogen")
+    |> range(start: -period)
+    |> filter(fn: (r) => r._measurement == "cpu")
+```
+
+In order to be alerted when the above task stops reporting data for a group, you would define a deadman alert using the `deadman` function like so:
+```
+option task = {
+    ...
+    period: 1h,
```
nathanielc: Tasks don't have periods? What is this meant to be?

docs/DeadmanAlert.md
nathanielc: Deadman should not include an alert, we do not want to couple those concepts. 

But perhaps deadman should not filter the dasta rather it should add a column `dead` to the data that indicates if it is dead or not and then the alert check can use the `dead` signal to know when a check has recovered.

```
deadman = (d, tables=<-) => 
    |> sort(columns:["_time"])
    |> last()
    |> map(fn: (r) ({r with dead: r._time < now()  -d ))
```

Then when used with a check it becomes:

```
from()
   |> range()
   |> filter()
   |> deadman()
   |> alert(crit: (r) => r.dead)
```

jsternberg: To clarify, I more meant `alert()` in the way it is being used in the checks/notifications document, but that is still in revision.

So I meant more like this:

deadman = (d, tables=<-) => tables
    |> sort(columns: ["_time"])
    |> last()
    |> check(crit: (r) => r._time < now() - d)

nathanielc: What I propose above is effectively the same thing as your example here except that you can still use `deadman` in other contexts besides alerting.

docs/DeadmanAlert.md
jlapacik: My mistake, I'll remove. It's just the range of the query.

jlapacik:
@nathanielc in my opinion it's very clear how dead is defined. `deadman` takes in a stream of tables and a duration value and produces the series that don't have any values within that duration.

I would rather keep `deadman` defined that way instead of as a `from` query, because otherwise the deadman function will need to take two durations, a bucket, and a predicate function as input. I think it's much more powerful (and user friendly) to have the user define their stream using flux and then pipe the results into the `deadman` function.

nathanielc:
@jlapacik Sounds good to me, lets leave it as a separate function from `from`. 

Thoughts on making it a `map` function instead of a `filter`?

docs/DeadmanAlert.md
jlapacik: Is there an example of when you do **not** want to filter out the series that are still reporting?

nathanielc: Yes, in order for an alert check to recover it needs to be given the list of series that are alive as well. If we only ever send the negative signal (i.e. the series is dead), then we cannot ever determine if the series is in an OK state.

jlapacik:
Based on the discussion above, it should be a `map`.

docs/DeadmanAlert.md
```diff
@@ -0,0 +1,60 @@
...
+    ...
+    every: 15m,
+    ...
+}
+from(bucket: "telegraf/autogen")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._measurement == "cpu")
+    |> deadman(d: task.every)
+    |> alert(crit: (r) => r.dead)
+```
```
affo: @jlapacik I know I came late and this got merged, but I have a concern about this.

If you specify `deadman` with `d` equal to `task.every` you will never get what you want, because every series will  contain the data for the specified time interval. If  the series doesn't contain data it won't be reported and you wouldn't be able to trigger an alert.

You should always  specify a `d < task.every` in order for the query to work.

Is this concern vane?

affo: We should state out clealry (at least in the doc) that, if using `deadman` within a task, it should be that `task.every > d`. So, you should always make your task sample data for an interval greater than the deadman duration.

I know this is only an example, just to make sure we are on the same page :).

jlapacik: @affo it's a valid concern however the above alert is valid because the data sampled is for `-1h` and `d < 1h`. `task.every` is just how frequently the task runs. Every time the task runs it reads the last `1h` worth of data.

affo: @jlapacik Absolutely, I haven't noticed that. Thank you.

Issue #761, opened 2018-07-03
Time literal arithmetic 


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#216  

https://github.com/influxdata/influxdb/issues/6881
https://github.com/influxdata/influxdb/issues/8153

Should be able to convert between times, durations and ints.

aanthony1243:
317 above is likely the wrong repo.

roshie548:
Some questions about time literal arithmetic:
When dividing `Durations`, do we want to return an Int or a Float?
```
> 1m/1d
0.0006944444444444445 or 0
```
Should we be able to subtract times and receive a duration (in ns) as a result?
Should we be able to add/subtract durations?

nathanielc:
Dividing durations should be a float and `1mo / 1d` should be `30` for now. When we implement duration vectors it will change.

>Should we be able to subtract times and receive a duration (in ns) as a result?

Yes the different of two times is a duration

> Should we be able to add/subtract durations?

Yes, add/subtracting durations will produce durations
Adding/subtracting durations with times will produce times.

jlapacik:
Adding/subtracting durations with times will require the type system to support polymorphic literals. See #476 for a discussion. But duration or time arithmetic is already supported by the type system. All that's needed is to update the interpreter to support it. I.e adding new binary function signatures to `values/binary.go`.

jlapacik:
Clarification. Adding and subtracting duration values is supported by the by the type signature of `+` and `-`. Every binary arithmetic operator has the following type signature:

    forall ['a] ('a, 'a) -> 'a

Given polymorphic literals, the above type signature supports adding or subtracting time and duration values together. However, it does not support adding two values of the same type together and producing a new different type as is the case with duration division producing a float.

We will need specific operators in order to support duration division returning a float, as well as time subtraction returning a duration. In other words we cannot overload the (/) and (-) operators to support these use cases.

russorat:
was this solved by https://github.com/influxdata/flux/pull/1592 ?

or also https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/experimental/addduration/ ?

PR #1592, opened 2019-07-24
feat(values): same-type time literal arithmetic

roshie548:
- added support for adding durations, subtracting durations and times, and dividing durations

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/values/binary.go
+++ b/values/binary.go
@@ -73,6 +73,11 @@ var binaryFuncLookup = map[BinaryFuncSignature]BinaryFunction{
 		r := rv.Str()
 		return NewString(l + r), nil
 	},
+	{Operator: ast.AdditionOperator, Left: semantic.Duration, Right: semantic.Duration}: func(lv, rv Value) (Value, error) {
+		l := lv.Duration()
+		r := rv.Duration()
+		return NewDuration(l + r), nil
+	},
 	{Operator: ast.AdditionOperator, Left: semantic.Nil, Right: semantic.Nil}: nil,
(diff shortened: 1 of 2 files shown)


values/binary.go
```diff
@@ -89,6 +94,16 @@ var binaryFuncLookup = map[BinaryFuncSignature]BinaryFunction{
 		r := rv.Float()
 		return NewFloat(l - r), nil
 	},
+	{Operator: ast.SubtractionOperator, Left: semantic.Time, Right: semantic.Time}: func(lv, rv Value) (Value, error) {
+		l := lv.Time()
+		r := rv.Time()
+		return NewDuration(Duration(l - r)), nil
```
jlapacik: We will actually need an new binary operator to specifically handle subtraction of time values. This is because subtracting times produces a duration, but the subtraction operator, along with all of our binary arithmetic operators, is defined to return the same type as that of its operands. See the following https://github.com/influxdata/flux/issues/761#issuecomment-514718381.

My suggestion would be to remove this operation for right now until the type system can support it.

values/binary.go
```diff
@@ -130,6 +145,11 @@ var binaryFuncLookup = map[BinaryFuncSignature]BinaryFunction{
 		}
 		return NewFloat(l / r), nil
 	},
+	{Operator: ast.DivisionOperator, Left: semantic.Duration, Right: semantic.Duration}: func(lv, rv Value) (Value, error) {
+		l := lv.Duration()
+		r := rv.Duration()
+		return NewFloat(float64(l) / float64(r)), nil
```
jlapacik: I would also remove this operation for the same reason as above.

PR #1598, opened 2019-07-24
feat(stdlib/universe): double and triple exponential average in flux

roshie548:
- double EMA has changed to be  defined in pure flux
- triple EMA defined in pure flux
- exponentialMovingAverage and movingAverage now only act on the _value column

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/universe.flux
```diff
@@ -255,6 +254,33 @@ timedMovingAverage = (every, period, column="_value", tables=<-) =>
...
+//    |> doubleExponentialMovingAverage(n: 5)
+doubleExponentialMovingAverage = (n, tables=<-) =>
+    tables
+          |> exponentialMovingAverage(n:n)
+          |> duplicate(column:"_value", as:"ema")
+          |> exponentialMovingAverage(n:n)
+          |> map(fn: (r) => ({r with _value: 2.0*r.ema - r._value}))
+          |> drop(columns: ["ema"])
+
+doubleEMA = doubleExponentialMovingAverage
```
nathanielc: Lets not have two names for the same thing, I am fine if we use the shorter names as their only names.

docs/SPEC.md
```diff
@@ -3172,23 +3172,22 @@ Given the following input table.
     
 ####  Moving Average
 
-Moving Average computes the moving averages of a series of records.
-It means the values of a user-defined period for a defined number of points,
+Moving Average computes the moving average the `_value` column.
```
wolffcm: Nit: typo, I think you want to write "moving average **of** the `_value` column."

wolffcm: Also, I'm curious why this was change to only work on the `_value` column, is it to be consistent with our other functions?

docs/SPEC.md
```diff
@@ -3338,7 +3346,82 @@ Example of script:
...
+- `EMA_2` is the exponential moving average of `EMA_1`,
+- `EMA_3` is the exponential moving average of `EMA_2`, and
+- `N = n` is the look-back period.
+
+In flux, `tripleEMA` is defined as
+```
+tripleEMA = (n, tables=<-) =>
+	tables
+		|> exponentialMovingAverage(n:n)
+		|> duplicate(column:"_value", as:"ema1")
```
wolffcm: Defining temporary columns like this seems potentially dangerous, since `duplicate` will silently overwrite data if the new column already exists.  It's unlikely, but a user might already have a column called `ema1`, etc.

It seems like we need a convention for column names that are only used internally.  That would require some more extensive language/library design, but for the time being, maybe change these temporary columns to `__ema1`, `__ema2`, and so on?

stdlib/universe/moving_average_test.go
```diff
@@ -840,8 +795,7 @@ func TestMovingAverage_Process(t *testing.T) {
 		{
 			name: "empty table",
 			spec: &universe.MovingAverageProcedureSpec{
-				Columns: []string{execute.DefaultValueColLabel},
-				N:       2,
+				N: 2,
 			},
```
wolffcm: Is there an error if `N` is less than 2, including 1, 0, or negative?  Would be good to verify that a reasonable error is produced.

docs/SPEC.md
roshie548: Correct, this is to be consistent with `doubleEMA`, and `tripleEMA` since those have `_value` hard-coded into them.

PR #1600, opened 2019-07-24
feat(stdlib/universe): added hourSelection function

sauren-khosla:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/hour_selection_test.go
+++ b/stdlib/universe/hour_selection_test.go
@@ -1,0 +1,453 @@
+package universe_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/stdlib/universe"
+)
+
+const hour int64 = 3600000000000
+
+func TestHourSelection_Process(t *testing.T) {
+	testCases := []struct {
+		name    string
+		spec    *universe.HourSelectionProcedureSpec
+		data    []flux.Table
+		want    []*executetest.Table
+		wantErr error
+	}{
+		{
+			name: "no group key",
+			spec: &universe.HourSelectionProcedureSpec{Start: 17, Stop: 19, TimeColumn: execute.DefaultTimeColLabel},
+			data: []flux.Table{
+				&executetest.Table{
+					ColMeta: []flux.ColMeta{
+						{Label: "_time", Type: flux.TTime},
+						{Label: "tag0", Type: flux.TString},
+						{Label: "tag1", Type: flux.TString},
+					},
+					Data: [][]interface{}{
+						{execute.Time(60000000000000), "a", "b"},
+						{execute.Time(70000000000000), "a", "c"},
+						{execute.Time(80000000000000), "a", "b"},
+						{execute.Time(90000000000000), "a", "d"},
+					},
+				},
+			},
+			want: []*executetest.Table{{
+				ColMeta: []flux.ColMeta{
+					{Label: "_time", Type: flux.TTime},
+					{Label: "tag0", Type: flux.TString},
+					{Label: "tag1", Type: flux.TString},
+				},
+				Data: [][]interface{}{
+					{execute.Time(70000000000000), "a", "c"},
+				},
+			}},
+		},
+		{
+			name: "no group key 2",
+			spec: &universe.HourSelectionProcedureSpec{Start: 6, Stop: 18, TimeColumn: execute.DefaultTimeColLabel},
+			data: []flux.Table{
+				&executetest.Table{
+					ColMeta: []flux.ColMeta{
+						{Label: "_time", Type: flux.TTime},
+						{Label: "tag0", Type: flux.TString},
+						{Label: "tag1", Type: flux.TString},
+					},
+					Data: [][]interface{}{
+						{execute.Time(hour), "a", "d"}, // 1 hour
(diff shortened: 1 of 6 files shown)


stdlib/universe/hourSelection_test.go
```diff
@@ -0,0 +1,424 @@
...
+					},
+				},
+			},
+			want: []*executetest.Table{{
+				ColMeta: []flux.ColMeta{
+					{Label: "_time", Type: flux.TTime},
+					{Label: "tag0", Type: flux.TString},
+					{Label: "tag1", Type: flux.TString},
+				},
+				Data: [][]interface{}(nil),
```
affo: mmmmm this is curious.

Wouldn't it be better to error if one provide `start` or `stop` `> 23` or `< 0`. I wouldn't expect  this result given the intrinsic periodicity of time....

Otherwise, you could say that `start` and `stop` are implicitly `%24` and this test  would be the same as:

```
&universe.HourSelectionProcedureSpec{Start: 1, Stop: 16},
```

stdlib/universe/hourSelection_test.go
```diff
@@ -0,0 +1,424 @@
...
+			spec: &universe.HourSelectionProcedureSpec{Start: 17, Stop: 19},
+			data: []flux.Table{
+				&executetest.Table{
+					ColMeta: []flux.ColMeta{
+						{Label: "_time", Type: flux.TTime},
+						{Label: "tag0", Type: flux.TString},
+						{Label: "tag1", Type: flux.TString},
+					},
+					Data: [][]interface{}{
+						{execute.Time(60000000000001), "a", "b"},
```
affo: I really like the format `3600000000000 * 2` because it  is  clearer and makes me understand the test cases. Could you change these tests to use it?

I also think that you can use `time.Hour` instead of `360....`. That is a duration, but it should be easy to change it to execute.Time.

Otherwise you can declare a `const hour int64 = 36000000...` at the top level to increase readability.

stdlib/universe/hourSelection.go
```diff
@@ -0,0 +1,173 @@
...
+func (t *hourSelectionTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	builder, created := t.cache.TableBuilder(tbl.Key())
+	if !created {
+		return fmt.Errorf("hour selection found duplicate table with key: %v", tbl.Key())
+	}
+	if err := execute.AddTableCols(tbl, builder); err != nil {
+		return err
+	}
+
+	colIdx := execute.ColIdx("_time", builder.Cols())
```
affo: I think you can run in this check on `tbl` instead of builder and directly use that index below (for `j`) without  re-searching the column index.

stdlib/universe/hourSelection.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func createHourSelectionOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
+	if err := a.AddParentFromArgs(args); err != nil {
+		return nil, err
+	}
+
+	spec := new(HourSelectionOpSpec)
+
+	if start, ok, err := args.GetInt("start"); err != nil {
```
affo: Shouldn't `start` and `stop` be required?

stdlib/universe/hourSelection_test.go
```diff
@@ -0,0 +1,424 @@
...
+						{Label: "_value", Type: flux.TFloat},
+						{Label: "tag0", Type: flux.TString},
+						{Label: "tag1", Type: flux.TString},
+					},
+					Data: [][]interface{}{
+						{execute.Time(3600000000000 * 4), 2.0, "a", "b"},
+						{execute.Time(3600000000000 * 28), 2.0, "a", "c"},
+						{execute.Time(3600000000000 * 6), 2.0, "a", "b"},
+						{execute.Time(3600000000000 * 30), 2.0, "a", "d"},
+						{execute.Time(3600000000000 * 32), 2.0, "a", "d"},
```
affo: When you have different group key values, you should separate the input in multiple tables.

stdlib/universe/hourSelection.go
```diff
@@ -0,0 +1,173 @@
...
+	colIdx := execute.ColIdx("_time", builder.Cols())
+	if colIdx < 0 {
+		return fmt.Errorf("no time column exists")
+	}
+
+	j := execute.ColIdx("_time", tbl.Cols())
+	return tbl.Do(func(cr flux.ColReader) error {
+		l := cr.Len()
+		for i := 0; i < l; i++ {
+			curr := execute.Time(cr.Times(j).Value(i)).Time().Hour()
```
affo: You should check for time validity. It could be `null`.

stdlib/universe/hour_selection_test.go
```diff
@@ -277,10 +306,11 @@ func TestHourSelection_Process(t *testing.T) {
 				},
 				Data: [][]interface{}(nil),
 			}},
+			wantErr: fmt.Errorf("start must be between 0 and 23"),
```
affo: you can delete the `want` field when you are expecting an error

stdlib/universe/hour_selection_test.go
```diff
@@ -404,7 +435,35 @@ func TestHourSelection_Process(t *testing.T) {
 					{2.0, "a", "d"},
 				},
 			}},
-			wantErr: fmt.Errorf("no time column exists"),
+			wantErr: fmt.Errorf("invalid time column"),
```
affo: same here, delete `want`

Issue #1575, opened 2019-07-22
Create Alert/Notification Package


nathanielc:
We need a package for alerts and notifications. It needs to be able to do a few things:

1. Evaluate a `check`: Consume data from an arbitrary query with at least `_value` and `_time` columns and evaluate a level function on each table. The level result along with the grouped columns from the input data need to be stored into a system bucket called a `status bucket`.
2.  Evaluate a `notification rule`: Execute a query against the `status bucket` and evaluate a rule function against the statuses. Send messages to third party notification endpoints when the rule indicates.
3. Log the notification events into a `notification bucket`.

Here is a glossary of terms used: 
```
Query: A Flux script that returns time series data (*Tags* + *Timestamp*)

Check: Analyzes the results of a *Query* to determine the current *Status* and writes it to a *Status Bucket*

Status: The *Level* and *Tags* of a *Check* written to the *Status Bucket*

Level: One of OK, INFO, WARN, CRIT, or UNKNOWN

Status Bucket: A location where a *Check* stores the current *Status*

Status Graph: A graph of the *Statuses* of a *Check* over time.

Notification Rule: A *Query* of a *Status Bucket* that returns the *Status*. When warranted by the rules, sends a *Message* to a 3rd Party using the *Notification Endpoint* and stores a receipt in the *Notifications Bucket*.

Notifications Bucket: A location where a *Notification Rule* records the name of the *Notification Endpoint*, *Message*, and *Tags* used in the *Query*.

Message: A templatized payload sent to *Notification Endpoint*.

Notification Endpoint: The configuration describing how to call a 3rd party service. E.g. Slack, Pagerduty

Alert: When a *Check* results in a *Status* that triggers a *Notification Rule*. 

``` 
Internal source: https://github.com/influxdata/idpe/blob/master/docs/content/rfcs/alertandnotify/alerts_and_notifications.md#methods

nathanielc:
DOD:

We have written out a design of these package(s). The design should include what functions the package exposes and the type signature of the functions with docs. Enough detail needs to be provided that someone could write Flux queries to consume this package. The implementations of the functions does not need to be completed as part of this issue.

jsternberg:
How do we plan for the flux queries to run? Do we think it's going to be similar to Kapacitor where the flux scripts are constantly running and getting new data or is the script going to be executed as part of an interval with tasks?

nathanielc:
@jsternberg The queries will be evaluated on an interval using tasks.

PR #1601, opened 2019-07-24
docs: initial document for an alerts package

jsternberg:
#1575 

## Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/Alerts.md
+++ b/docs/Alerts.md
@@ -1,0 +1,328 @@
+# Alert Handling
+
+## Overview
+
+### Requirements
+
+The following are considered requirements for the alert package to both support and/or implement.
+
+* It must be able to consume data from an arbitrary query and evaluate a level function on each row in each table.
+* It must be possible to store this result in a system bucket called a `status bucket`.
+* It must be able to evaluate a `notification rule` based on the output of the previous or from the data stored in the status bucket.
+* The notification must send messages to third party endpoints when the rule indicates.
+* The notification events must be capable of being logged into a `notification bucket`.
+
+### Query Lifecycle
+
+The query lifecycle is tangentially related to alerts.
+At the moment, the flux engine is currently only used for running batch requests.
+This means that there is only one table for each group key that goes through the pipeline.
+
+The alert package as below does not assume this.
+It is designed so the functions will work regardless of whether run in batch mode or if the query were being run with each new row triggering a new table.
+For ease of reading, the example tables will be shown as if only one table were shown.
+
+### Considerations
+
+While designing the alert package, we have looked at the previous work done in this field that was included in [Kapacitor](https://docs.influxdata.com/kapacitor/v1.5/nodes/alert_node/).
+
+It is important to support the same workflows that Kapacitor allows.
+
+## Alerts and Notifications
+
+Alerts and notifications is comprised of two parts: checking the input and notifying a third party about the output of the check.
+
+For this reason, the Flux library will treat these as separate concepts.
+When evaluating an input stream to determine the alert level, we will need to evaluate each row to determine if that row matches into one of 4 possible alert levels: `ok`, `info`, `warn`, and `crit`.
+
+Each of these levels will be represented by an integer to make it easier to compare the levels.
+There will also be an additional level to represent that a check was not able to determine the status or the status was not yet known: `unknown`.
+
+They will be mapped as the following:
+
+|level|value|
+|-----|-----|
+|ok|0|
+|info|1|
+|warn|2|
+|crit|3|
+|unknown|-1|
+
+### Check
+
+The `check` function will take an input and evaluate each row to determine the level.
+The function will have a predicate for each specified level and will output a level for the highest matching check.
+One output table will be produced for each input table.
+The output tables will have the same schema, but with two additional columns: `_level:int` and `_name:string`.
+
+`check` has the following properties:
+
+| Name | Type                | Description
+| ---- | ----                | -----------
+| name | string              | The check name. It will be added as a group key to every table that passes through. Required.
+| crit | (r: record) -> bool | The function that determines if at the `crit` level. If unset, it is ignored.
+| warn | (r: record) -> bool | The function that determines if at the `warn` level. If unset, it is ignored.
+| info | (r: record) -> bool | The function that determines if at the `info` level. If unset, it is ignored.
+| ok   | (r: record) -> bool | The function that determines if at the `ok` level. If unset, this will always return true.
+
+If multiple of the above are true, then the highest level wins.
+If none of the above are true, then `-1 (unknown)` is the result.
+The methods above will short-circuit so there is no expectation that all of them will be invoked for every row.
+The functions should not have side effects.
+
+The `check` function will also use the option `checks.write` if it is defined.
+The function `checks.write` should take in the output tables as a single parameter and output the same tables.
+
+Example 1:
+
+```
+import "influxdata/influxdb/alerts"
+
+from(bucket: "telegraf")
+    |> range(start: -5m)
+    |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_system")
+    |> alerts.check(name: "cpu_usage",
+    	crit: (r) => r._value > 90,
+    	warn: (r) => r._value > 80,
+    )
+```
+
+|Input:|_time|_value|Output:|_time|_value|_level|_name|
+|------|-----|------|-------|-----|------|------|-----|
+| |0|45| |0|45|ok|cpu_usage|
+| |10|55| |10|55|ok|cpu_usage|
+| |20|95| |20|95|crit|cpu_usage|
+| |30|92| |30|92|crit|cpu_usage|
+| |40|89| |40|89|warn|cpu_usage|
+| |50|91| |50|91|crit|cpu_usage|
+| |60|87| |60|87|warn|cpu_usage|
+| |70|81| |70|81|warn|cpu_usage|
+| |80|50| |80|50|ok|cpu_usage|
+| |90|60| |90|60|ok|cpu_usage|
+| |100|30| |100|30|ok|cpu_usage|
+| |110|25| |110|25|ok|cpu_usage|
+| |120|10| |120|10|ok|cpu_usage|
+
+Example 2:
+
+```
+import "influxdata/influxdb/alerts"
+
+from(bucket: "telegraf")
+    |> range(start: -5m)
+    |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_system")
+    |> alerts.check(name: "cpu_usage",
+    	crit: (r) => r._value > 90,
+    	warn: (r) => r._value > 80,
+    	// Do not reset to the ok level until the system usage drops enough.
+    	ok:   (r) => r._value <= 20,
+    )
+```
+
+|Input:|_time|_value|Output:|_time|_value|_level|_name|
+|------|-----|------|-------|-----|------|------|-----|
+| |0|45| |0|45|ok|cpu_usage|
+| |10|55| |10|55|ok|cpu_usage|
+| |20|95| |20|95|crit|cpu_usage|
+| |30|92| |30|92|crit|cpu_usage|
+| |40|89| |40|89|warn|cpu_usage|
+| |50|91| |50|91|crit|cpu_usage|
+| |60|87| |60|87|warn|cpu_usage|
+| |70|81| |70|81|warn|cpu_usage|
+| |80|50| |80|50|unknown|cpu_usage|
+| |90|60| |90|60|unknown|cpu_usage|
+| |100|30| |100|30|unknown|cpu_usage|
+| |110|25| |110|25|unknown|cpu_usage|
+| |120|10| |120|10|ok|cpu_usage|
+
+In this example, the level does not get reset to ok until it passes the `ok` handler.
+This not only requires the `crit` and `warn` checks to return false, but also requires the `ok` check to affirmatively say that things are ok.
+In this way, it is possible to delay the ok signal to the notification handler.
+
+Example 3:
+
+```
+import "influxdata/influxdb/alerts"
+
+from(bucket: "telegraf")
+    |> range(start: -5m)
+    |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_system")
+    |> alerts.check(name: "cpu_usage",
+		crit: (r) => r._value > 90,
+		warn: (r) => r._value > 80,
+    )
+```
+
+(Note): The above drops miscellaneous columns that do not affect the result.
+
+(Note): The above levels would be integers, but are represented as strings for readability.
+
+(Note): `_name` is part of the group key.
+
+### Notify
+
+The `notify` function will send notifications to the specified endpoint.
+The endpoint definition is defined below.
+
+`notify` has the following properties:
+
+| Name     | Type                           | Description
+| ----     | ----                           | -----------
+| endpoint | (tables: <-tables) -> <-tables | The endpoint to notify. Required.
+
+`notify` will call the endpoint and write the results to a bucket.
+
+Example:
+
+```
+import "influxdata/influxdb/alerts"
+import "slack"
+
+endpoint = slack.endpoint(name: "slack", channel: "#flux")
+
+from(bucket: "system")
+	|> range(start: -5m)
+	|> alerts.notify(endpoint: endpoint)
+```
+
+### Notification Endpoints
+
+Notification endpoints are used to notify a third party endpoint about the state of a check.
+Each third party package implements two functions that are common between all notification packages.
+
+The `endpoint()` function is used to construct the endpoint that can be used to notify the third party endpoint.
+All `endpoint()` functions will have at least a `name` parameter and will return a function that will accept a stream of tables as the input.
+
+	endpoint = (tables=<-, name, ...) => tables |> notify(name: name, ...)
+	
+The `notify()` function will be used to perform the notification.
+The `notify` function will take an input with at least `_time` and `_level` as columns.
+The output of the notification endpoint will be the actions that were performed for the group key.
+The output of `notify()` will add a `_endpoint:string` column with the name passed to `notify()` that will be part of the group key.
+Notification handlers will ignore the `unknown` level (or non-positive level values).
+
+Example:
+
+```
+// An example of a notifications package.
+package pagerduty
+
+endpoint = (name, routingKey) => (tables=<-) => tables
+	|> notify(name: name, routingKey: routingKey)
+
+builtin notify
+
+// An example using the above.
+package main
+
+import "influxdata/influxdb/alerts"
+import "pagerduty"
+
+endpoint = pagerduty.endpoint(
+	name: "pagerduty",
+	routingKey: myRoutingKey,
+)
+
+from(bucket: "system")
+	|> range(start: -5m)
+	|> alerts.notify(endpoint: endpoint)
+```
+	
+Example output based on the input from example 2 from above:
+
+|_time|_at|_action|_status|_changed|_name|_endpoint|
+|-----|---|-------|-------|--------|-----|--------|
+|20|30|notify|crit|true|cpu_usage|pagerduty|
+|30|40|notify|crit|false|cpu_usage|pagerduty|
+|40|50|notify|warn|true|cpu_usage|pagerduty|
+|50|60|notify|crit|true|cpu_usage|pagerduty|
+|60|70|notify|warn|true|cpu_usage|pagerduty|
+|70|80|notify|warn|false|cpu_usage|pagerduty|
+|120|130|resolve|ok|true|cpu_usage|pagerduty|
+
+As can be noticed, the handler did not do anything with the `unknown` levels and so there was nothing to log.
+The `_at` time is also 10 seconds after the time the event occured.
+This is because it is the time the action was _performed_ rather than when the check happened.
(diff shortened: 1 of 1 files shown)


docs/alerts.md
```diff
@@ -0,0 +1,183 @@
...
+This assumes that the table is triggered upon receiving the next point in the sequence so it is triggered
+on receiving the next point after 10 seconds.
+
+If we were in a situation where the handler were to skip an action because a future notification pre-empted it (such as if only a single table were received and the handler noticed that the alert was already resolved), the `_at` time would be null since the action never occurred, but the row would still be present.
+
+### Writing to Buckets
+
+Each of the above assumes it would be possible to write the output to a bucket and to have the bucket be the source rather than chained together in the same script.
+
+I am not sure if we should build that functionality directly into the checks or how to integrate this writing.
```
nathanielc: We should create helper functions for this behavior. Maybe we have a `_check` which is what you have described and `check` which wraps it with calls to `to`. Similar for `notify` and `from`.

docs/alerts.md
```diff
@@ -0,0 +1,183 @@
...
+|warn|2|
+|crit|3|
+|unknown|-1|
+
+### Check
+
+The `check` function will take an input and evaluate each row to determine the level.
+The function will have a predicate for each specified level and will output a level for the highest matching check.
+One output table will be produced for each input table.
+The output tables will have the same schema, but with one additional column that is `_level:int`.
```
nathanielc: Check will need to output a `_check` column that contains the name/id of the check. 

Additionally users need to be able to add arbitrary tags to the check output. This can easily be done with a map function, but we should make it easy to do. Perhaps with a `tags` parameter to the `check` function.

jsternberg: I'll add a `name` parameter which will add a value to the group key for the name of the check.

What do you mean by arbitrary tags? To clarify, the current design would keep the existing tags from the input. So it might be better to have a `set()` function that allows setting a value in a column or setting a value and adding it to the group key.

I'm not sure this needs to be part of the behavior of the `check()` function though because check will pass through the existing data and the existing group key.

docs/alerts.md
jsternberg: I agree. I've been thinking that either we should have a parameter that adds the behavior and we would have the influxdb version of flux set the behavior for the default writer or just making sure that there's a `to()` function that will ensure the output is written since I know `to()` can get confusing to use.

I'd like to know a bit more about how it would get used though. If this were to be used by tasks, then tasks would be able to generate the appropriate flux for writing the result.

jsternberg:
The idea for notifications is that they would be their own functions. So you might have something like:

```
import "slack"
... |> check(...) |> slack.notify(channel: "#flux")
```

This would be if you were to retrieve the values and just notify immediatly. Or if you wanted to have two different tasks or two different streams, you might have something like:

```
import "slack"

from(bucket: "telegraf")
    |> stuff()
    |> check(...)
    |> to(bucket: "checks", ...)

from(bucket: "checks")
    |> slack.notify(channel: "#flux")
```

I think this makes it very composable and easy to build tooling around without creating too much feature creep for each individual transformation. The `slack.notify()` transformation would understand the input format and it would send messages to slack possibly only on state changes. It would then output a log of what it does using the above format. Depending on the transformation, it might add an extra column or so to describe the actions.

It would then be easy to build topic handlers into this because the topic handler would also consume the same format and would just be the equivalent of:

```
my_query => from(...) |> ... |> check(...)
my_query |> slack.notify(channel: "#flux")
my_query |> pagerduty.notify(routingKey: routingKey)
```

(I don't actually know what the parameters for pagerduty would be)

docs/alerts.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Query Lifecycle
+
+The query lifecycle is tangentially related to alerts.
+At the moment, the flux engine is currently only used for running batch requests.
+This means that there is only one table for each group key that goes through the pipeline.
+
+The alert package as below does not assume this.
+It is designed so the functions will work regardless of whether run in batch mode or if the query were being run with each new row triggering a new table.
+For ease of reading, the example tables will be shown as if only one table were shown.
```
affo: I suppose that the most common use case for alerting should be to embed it into a task.

The problem with tasks is that you cannot share data between many invocations of the batch job. That would require real streaming. I don't know if we need that capability at the moment or in the future.

jsternberg: @nathanielc mentioned in the connected issue that this was how we were going to do alerting. I don't know if that's permanent or just at the moment until we figure out a way to run queries continuously, but I think the model outlined here will work regardless of the mode.

docs/alerts.md
```diff
@@ -0,0 +1,183 @@
...
+|------|-----|------|-------|-----|------|------|
+| |0|45| |0|45|ok|
+| |10|55| |10|55|ok|
+| |20|95| |20|95|crit|
+| |30|92| |30|92|crit|
+| |40|89| |40|89|warn|
+| |50|91| |50|91|crit|
+| |60|87| |60|87|warn|
+| |70|81| |70|81|warn|
+| |80|50| |80|50|unknown|
```
affo: don't you think it would be better to keep the last valid state instead of `unknown`?

docs/alerts.md
affo: I'd like to leverage the composability of Flux and let the user chain a `to` transformation.

docs/alerts.md
jsternberg: There's a few reasons why not:

1. We may not have access to that. This doesn't assume that it knows the value of the last row so it is stateless and relies on the handler to determine what it should do with the input instead.
2. It would be a lie to have it give the last valid state. It's not in the warning or critical state anymore, so why should we report that it is?

docs/alerts.md
jlapacik: If `to()` is confusing, let's simplify it. I would prefer the writing be handled by `to()` and not encapsulated inside of `check` and `notify`.

nathanielc:
## Notification Rules

@jsternberg The RFC calls for an abstraction of a notification rule which allows for the users to select various check statuses and decide whether or not to notify on them. Once the decision to notify has been made then the check status is passed on to the notification endpoints (i.e. notify.slack).

We need to represent that notification rule logic somewhere. 

And the logic needs to be able to express at the very least  [these](https://github.com/influxdata/flux/pull/1601#pullrequestreview-266294889) rules. It is also OK if the first pass is a set of built-in functions one for each of those rules.

This notification rule also needs to be able to specify which checks it wishes to consume. The RFC says this is defined by the user via a `match` expression. I.e. a Flux filter that matches only the check statuses the notification rule applies to.



## Writing to System Buckets

On the topic of of writing to the system bucket. Writing the check statuses to the system bucket is fundamental to how alerts are implemented. There are a few different kinds of user stories here.

1. A normal user - This user will write very limited Flux to create an alert if any at all. They will not be using the `check` and `notify` functions directly. The UI/API tiers will be responsible for generating Flux code based on the values they provide when creating an alert. 
2. A power user - This user will extend the alerting/notifications API that we provide for custom logic. This users builds on top of the existing alerting/notification functions and features.
3. A platform builder - This user will use the InfluxDB 2.0 platform to build their own alerting and notification framework that is some how different from ours.

The power user should not have to manage writing check status etc back to a system bucket, that should be done for them. For example a power user will create custom notification rules that consume the standard check statuses from the system bucket. Alternatively they may write custom checks that write to the system bucket but that normal notification rules can consume.

The platform builder may not want to write the check status back to a system bucket, in fact they may use a completely different check status mechanism, but they should be able to use the primitives that we have built as desired.



In the context of these different classes of users I want the default to be to write to system buckets. Only the platform builder doesn't need that feature.  As a result what ever the top level API we expose from the alerts package should write to system buckets and not require users to do that explicitly. This doesn't mean we can't expose a more low level API for platform builders, but we need to make sure the top level API suites the needs of the power users.

docs/alerts.md
affo: I agree with you, only wanted to see the cons :)

docs/alerts.md
nathanielc: Using tasks as the backbone of alerting is the long term plan, its not temporary.

docs/alerts.md
nathanielc: @jsternberg Yes using the `set` function is what I had in mind. Basically users will be able to inject new tags on their data as part of the check. This allows them to categorize their checks. This is similar to Kapacitor's topic system but less rigid.

docs/alerts.md
jsternberg: I more meant if tasks would be repeatedly executed or we would have tasks be something that is continuously run. If we don't plan to have any streaming functionality, I think we may need to have a conversation about triggers as the removal of triggers could simplify the query engine a decent amount.

nathanielc: We do not plan to have a true streaming query any time soon. Lets have the discussion on triggers separate from the alert/notification design. We will revisit the alerts design when we go to introduce true streaming later along with many other designs.

jsternberg:
@nathanielc responses are quoted inline below:

> @jsternberg The RFC calls for an abstraction of a notification rule which allows for the users to select various check statuses and decide whether or not to notify on them. Once the decision to notify has been made then the check status is passed on to the notification endpoints (i.e. notify.slack).
>
> We need to represent that notification rule logic somewhere.

This is in the document, but I can do a better job of fleshing it out so I'm going to do that. We'll revisit this once I flesh it out a bit more and put some syntax and examples of it being used behind it.

> And the logic needs to be able to express at the very least these rules. It is also OK if the first pass is a set of built-in functions one for each of those rules.

I think this has confused me for a bit. It seems to me like the logic expressed in those rules should be handled by the notification rule itself rather than the intent to notify. As an example, I don't think `stateChangesOnly` from Kapacitor means something to all of the notification handlers. I don't think it means anything to PagerDuty which operates the same if I remember correctly. It makes sense for the Slack handler. I'll flesh this out and then we can talk about this on the specific section. Maybe it would be good to have some default parameters for notifications so I'll write a section specifically addressing this.

> This notification rule also needs to be able to specify which checks it wishes to consume. The RFC says this is defined by the user via a match expression. I.e. a Flux filter that matches only the check statuses the notification rule applies to.

This can likely be a helper function that uses `filter`. I'll specifically add this.

I do have one question. How much are we allowed to rely on tasks being able to inject or modify code? I figure as long as it is easy, it should be something we are able to do. It might help if I write in the document how tasks is supposed to be using the features that are present and ensure that it is compatible.

jsternberg:
I tried to show how this might be used with tasks and how it would be possible to let casual users use the code generation and how power users could overwrite the default generated behavior.

docs/alerts.md
```diff
@@ -0,0 +1,264 @@
...
+	
+// The function is easy for the user to write and it is easy to
+// inject into a generated script.
+handler1 = (tables=<-) => pagerduty.notify(routingKey: myRoutingKey)
+handler2 = (tables=<-) => slack.notify(channel: "#flux")
+	
+// Send to multiple notification handlers.
+// This is easy to generate.
+tables |> handler1() |> to(bucket: "notifications")
+tables |> handler2() |> to(bucket: "notifications")
```
nathanielc: This example is a bit confusing, where it the `check` in this example? Or is this just and example that is reading from the `check status` system bucket? Maybe it would help to write out the separate tasks completely and show how they interact?

jsternberg: This is an example that reads from the `check status` system bucket. Or `tables` is defined in some other way.

I do think it might be helpful to write separate tasks entirely. I'll work on that.

jsternberg:
> More details one how I understand that use case. There are N number of checks producing statuses. I want to send a notification to slack at most once during a 5m period for those N checks.

I think that's the part I'm cautious about. I think that options like that are specific to the handler itself rather than something more general so I would like that feature to be built in as an option on the slack handler.

When it comes to the actual implementation, we would have mixins that would implement the behavior so that we aren't duplicating code for each handler. But, this functionality doesn't seem to make sense for PagerDuty.

I meant to include an example of filtering levels in the handler code so I am going to add that example.

nathanielc:
> When it comes to the actual implementation, we would have mixins that would implement the behavior so that we aren't duplicating code for each handler. But, this functionality doesn't seem to make sense for PagerDuty.

Can you provide an example of how this would work? I like the sound of the idea. I;d like to see two example tasks (the check task and the notification task) for the use case I specified above.

jsternberg:
I'll add the tasks examples.

When it comes to the mixins, I was thinking it would be similar to how `execute.Aggregate` works. I don't know the specifics, but it would be a common implementation that would add additional argument parameters similar to how `execute.Aggregate` works.

docs/alerts.md
```diff
@@ -262,3 +262,51 @@ handler2 = (tables=<-) => slack.notify(channel: "#flux")
...
+This might be the check script.
+
+```
+// Injected.
+option task = {
+	every: 1m,
+}
+
+// Possibly an option like now?
+notify = (tables=<-) => tables |> to(bucket: "system")
```
nathanielc: Why is this function called `notify`? All it does is write to the system bucket.

docs/alerts.md
```diff
@@ -262,3 +262,51 @@ handler2 = (tables=<-) => slack.notify(channel: "#flux")
...
+
+// User provided.
+from(bucket: "telegraf")
+    |> range(start: -task.every)
+    |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_system")
+    |> check(
+	crit: (r) => r._value > 90,
+    )
+    |> notify()
```
nathanielc: This design doesn't satisfy the requirement that by default check statuses are written to a bucket. A power user using the `check` method would need to remember to write to the system bucket in every alert script.

docs/alerts.md
```diff
@@ -0,0 +1,312 @@
...
+
+```
+// Injected.
+option task = {
+	every: 1m,
+}
+
+tables = from(bucket: "system")
+	|> range(start: -task.every)
+	|> notify()
```
nathanielc: What is the definition of the `notify` function here?

docs/alerts.md
```diff
@@ -0,0 +1,312 @@
...
+
+// The user could customize the function body for each handler,
+// but the function signature would probably be generated.
+handler1 = (tables=<-) => pagerduty.notify(routingKey: myRoutingKey)
+handler2 = (tables=<-) => slack.notify(channel: "#flux")
+
+// Send to multiple notification handlers.
+// This is easy to generate.
+tables |> handler1() |> to(bucket: "notifications")
+tables |> handler2() |> to(bucket: "notifications")
```
nathanielc: Same as above, the API exposed by this package needs to write to the system bucket by default.

docs/alerts.md
```diff
@@ -0,0 +1,312 @@
...
+
+This might be the notification script.
+
+```
+// Injected.
+option task = {
+	every: 1m,
+}
+
+tables = from(bucket: "system")
```
nathanielc: Reading from the system bucket to get a set of check statuses should be a helper method on the package API. This way the reading and writing to the system bucket is abstracted away from the power user.

docs/alerts.md
```diff
@@ -0,0 +1,312 @@
...
+	|> range(start: -task.every)
+	|> notify()
+
+// The user could customize the function body for each handler,
+// but the function signature would probably be generated.
+handler1 = (tables=<-) => pagerduty.notify(routingKey: myRoutingKey)
+handler2 = (tables=<-) => slack.notify(channel: "#flux")
```
nathanielc: >When it comes to the mixins, I was thinking it would be similar to how execute.Aggregate works. I don't know the specifics, but it would be a common implementation that would add additional argument parameters similar to how execute.Aggregate works.

Independent of how the mixins are implemented the Flux script has to declare the  notification rule behavior somewhere. Where is that logic represented in this Flux script? Meaning reading this script how can I tell it is the `notify at most once during a specific interval` behavior vs another behavior.

docs/alerts.md
jsternberg: Because it abstracts away how the notifications are implemented. So you would put `notify()` in your script to say "I want to send a notification" and we would insert the notification behavior. Notify can also be implemented as an option too.

docs/alerts.md
jsternberg: Yes, it does. We would insert the implementation of `notify()` to write to the system bucket. There's no necessity for them to write to the system bucket because that's what `notify()` does.

docs/alerts.md
jsternberg: I am confused. This is the notification log portion. The system bucket seems to only be involved for evaluating the checks. Did I misunderstand the spec?

docs/alerts.md
jsternberg: It is intended that this is generated by tasks as mentioned in the comment.

docs/alerts.md
jsternberg: I'll update it with an example.

docs/alerts.md
jsternberg: This is a mistake. `notify()` should not be here.

docs/alerts.md
nathanielc: The design in the SPEC is that checks and notifications are loosely coupled. This design pattern forces checks and notifications to be tightly coupled because a check's status is not persisted unless a `notify` function is called.

We need to allow for checks to be persisted to a bucket without any call to notify functions.

docs/alerts.md
nathanielc: See my comment above, the check function or whatever wrapper around it needs to persist the status to a bucket without the need to call a notify function.

docs/alerts.md
nathanielc: Both check statuses and notification logs need to be written to a system bucket. But which bucket is not too important for this design as we can easily make that parameterized. 

My initial comment was that this design requires that users explicitly write to a bucket, where we need that to be implicit (i.e. default) from the API.

jsternberg: Ah. The idea here is that the user provided part goes into the implementation of `handler1` or `handler2`. With notifications, I am less clear on how that would be customized by power users. Do you have an example of how a user would want to do something more than just configure the handler using parameters?

It might make it easier for me to develop how this would work and write an example. Otherwise, it seems to me the handler portion is basically completely generated.

docs/alerts.md
jsternberg: I think you and I might just disagree on this. I do not see how they are tightly coupled here. The notify just signals that a notification should happen based on the check result and the implementation of notify is completely abstracted away. It could either be the actual implementation so the notification happens in the same task or it could be part of a notification task as shown below.

Is this a problem with the syntax that is used? The reason why I like this pattern is because it is easier to understand the intention of the user.

docs/alerts.md
nathanielc: The generated task should be able to call a parameterized helper function that encapsulates the basic concepts. The reason for this is we want to enable power users to leverage the same basic constructs of using checks and notifications without having to manage all the details. Specificaly in this case power users should not need to know that check statuses are stored in a system bucket, rather that they are stored somewhere and can be retrieved using the provided API methods. Who knows, maybe we decide that we should store the statuses in a relational database in the future. Abstracting the details of the storage enables us to change that without breaking the API. 

Something like this is what I have in mind:


```
findChecks = ( predicate,interval) =>
    from(bucket : "system") 
        |> range(start: -interval)
        |> filter(fn: predicate)

```

Then the generated task is simply:

```
option task = {
    every: 5m,
}

checks = findChecks(interval: task.every, predicate: <insert predicate function here from notification rule match field>)

checks
    |> // notification rule logic goes here
    |> slack.notifty()
```

docs/alerts.md
nathanielc: What I mean by tightly coupled is that in this design in order for a check's status to be persisted it must be passed to an implementation of a notify function. We do not want users to think that every check must have a notification. Users should be able to create checks and not notify on them. An example workflow is  developing a new check. The users does the following: write the check, leave it running for a while, manually inspect the statuses of the check, repeat until satisfied by its behavior, and finally define a notification that matches the check. 

From the spec,

>Checks and notifications are separate and distinct. Checks can be delivered first and then notifications can be introduced next

> Notification Rules are loosely coupled to checks by the matching check tags. One notification rule may match several checks.

jsternberg: I'll integrate this in. Thanks for the explanation. I think I understand it better now.

jsternberg: This is now integrated. I think something in the notifications part needs to change, but I want you to take a look at the checks part first before I try to integrate a similar solution into notifications or come up with something different there.

docs/alerts.md
```diff
@@ -275,17 +288,14 @@ option task = {
 	every: 1m,
 }
 
-// Possibly an option like now?
-notify = (tables=<-) => tables |> to(bucket: "system")
+// Default value for the flux embedded in influxdb.
+option handleCheckResult = (tables=<-) => tables |> to(bucket: "system")
```
nathanielc: I like the use of the option, I think it works well here. I agree the default should be to write to the bucket.

docs/alerts.md
nathanielc: Check parts look good

docs/alerts.md
jsternberg: One key aspect is I think it should only be the default in influxdb. I think the option should default to doing nothing in flux itself.

I hope I made that clear. I have some apprehensions with making any of the default behavior rely on code from influxdb. I hope that is understood and is ok.

nathanielc: Yes, its clear, I don't have a strong opinion on the default inside vanilla Flux ATM. 

Implementation-wise I am not sure it is easy to have different defaults but we can figure that out later.

jsternberg:
New revision for notification handlers.

PR #1602, opened 2019-07-24
Newly built mqtt to output

davidgs:
mqtt.to() output from flux

### Done checklist
- [ ] docs/SPEC.md updated
- [ x] Test cases written

### DIFF
--- a/stdlib/mqtt/to.go
+++ b/stdlib/mqtt/to.go
@@ -1,0 +1,649 @@
+package mqtt
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/url"
+	"sort"
+	"time"
+	"strings"
+	"strconv"
+	MQTT "github.com/eclipse/paho.mqtt.golang"
+
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/internal/pkg/syncutil"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+	protocol "github.com/influxdata/line-protocol"
+	"github.com/pkg/errors"
+)
+
+const (
+	ToMQTTKind           = "toMQTT"
+	DefaultToMQTTTimeout = 1 * time.Second
+)
+
+func init() {
+	toMQTTSignature := flux.FunctionSignature(
+		map[string]semantic.PolyType{
+			"broker":       semantic.String,
+			"topic":        semantic.String,
+			"message":      semantic.String,
+			"format":       semantic.String,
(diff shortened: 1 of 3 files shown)


stdlib/mqtt/to.go
```diff
@@ -0,0 +1,649 @@
...
+	opts := MQTT.NewClientOptions().AddBroker(t.spec.Spec.Broker)
+	if len(t.spec.Spec.ClientID) > 0 {
+		opts.SetClientID(t.spec.Spec.ClientID)
+	} else {
+		opts.SetClientID("flux-mqtt")
+	}
+	if t.spec.Spec.Timeout > 0 {
+		opts.SetConnectTimeout(t.spec.Spec.Timeout)
+	}
+	if len(t.spec.Spec.Username) > 0 {
```
aanthony1243: if these are strings, it might be clearer to say `if t.spec.Spec.Username != ""`

stdlib/mqtt/to.go
```diff
@@ -0,0 +1,649 @@
...
+			l := er.Len()
+			for i := 0; i < l; i++ {
+				m.truncateTagsAndFields()
+				for j, col := range er.Cols() {
+					switch {
+					case col.Label == timeColLabel:
+						m.t = values.Time(er.Times(j).Value(i)).Time()
+					case measurementNameCol != "" && measurementNameCol == col.Label:
+						if col.Type != flux.TString {
+							return errors.New("invalid type for measurement column")
```
aanthony1243: these types of errors should be checked outside of tbl.Do  using `tbl.Cols()`.  Doing this check over and over for every row will slow you down quite a bit.

stdlib/mqtt/to.go
```diff
@@ -0,0 +1,649 @@
...
+	if err := wg.Wait(); err != nil {
+		client.Disconnect(250)
+		return err
+	}
+	client.Disconnect(250)
+	return nil
+
+}
+// format the message as json
+func (t *toMqttMetric) formatJSON(message string) strings.Builder {
```
aanthony1243: this is sort of an antipattern in Go.  You can use struct annotations to marshal a simple struct as JSON.   so you have something like   `type metric struct{...}` that can be encoded properly as JSON, and then you can have a `func (m metric) toMessage() string {...}`   to encode the message as a simple string.

stdlib/mqtt/to.go
davidgs: Done!

stdlib/mqtt/to.go
davidgs: Sounds like we should leave this as-is.

davidgs:
Pulling out JSON support, re-working the tests, etc. Will re-submit as a new PR.

PR #1603, opened 2019-07-24
feat(stdlib/universe): add keepFirst to difference

roshie548:
- keepFirst is an optional parameter for whether to keep the first row or not
- default behavior is to drop the first row

This parameter would allow for `tripleExponentialDerivative` to be written in pure flux. Currently, this can't be done because `join` after using `difference` causes the first row to be lost.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


values/binary_test.go
```diff
@@ -11,11 +11,11 @@ import (
...
-	intNullValue    = (*int64)(nil)
-	uintNullValue   = (*uint64)(nil)
-	floatNullValue  = (*float64)(nil)
-	stringNullValue = (*string)(nil)
-	timeNullValue   = (*values.Time)(nil)
+	intNullValue      = (*int64)(nil)
+	uintNullValue     = (*uint64)(nil)
+	floatNullValue    = (*float64)(nil)
+	stringNullValue   = (*string)(nil)
+	timeNullValue     = (*values.Time)(nil)
```
roshie548: go fmt result

Issue #1540, opened 2019-07-17
implement CHANDE_MOMENTUM_OSCILLATOR


aanthony1243:
should provide parity with same function from influxQL

PR #1612, opened 2019-07-25
feat(universe/chande): add cmo from influxql to flux

ewendai:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

#1540

### DIFF
--- a/stdlib/universe/chande_momentum_oscillator.go
+++ b/stdlib/universe/chande_momentum_oscillator.go
@@ -172,63 +172,22 @@ func (t *chandeMomentumOscillatorTransformation) Process(id execute.DatasetID, t
 	}
 
 	return tbl.Do(func(cr flux.ColReader) error {
-		if cr.Len() == 0 {
+		if cr.Len() == 0 || len(cr.Cols()) == 0 {
 			return nil
 		}
 
 		for j, c := range cr.Cols() {
 			if !doChandeMomentumOscillator[j] {
-				switch c.Type {
-				case flux.TTime:
-					arr := cr.Times(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendTime(j, execute.Time(arr.Value(i))); err != nil {
-							return err
-						}
-					}
-				case flux.TBool:
-					arr := cr.Bools(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendBool(j, arr.Value(i)); err != nil {
-							return err
-						}
-					}
-				case flux.TString:
-					arr := cr.Strings(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendString(j, arr.ValueString(i)); err != nil {
-							return err
-						}
-					}
-				case flux.TInt:
-					arr := cr.Ints(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendInt(j, arr.Value(i)); err != nil {
-							return err
-						}
-					}
-				case flux.TUInt:
-					arr := cr.UInts(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendUInt(j, arr.Value(i)); err != nil {
-							return err
-						}
-					}
-				case flux.TFloat:
-					arr := cr.Floats(j)
-					for i := int(t.n); i < arr.Len(); i++ {
-						if err := builder.AppendFloat(j, arr.Value(i)); err != nil {
-							return err
-						}
+				for i := int(t.n); i < cr.Len(); i++ {
+					if err := builder.AppendValue(j, execute.ValueForRow(cr, i, j)); err != nil {
+						return err
 					}
 				}
-			} else {
-				if err := t.do(int(t.n), cr, c, builder, j); err != nil {
-					return err
-				}
+				continue
 			}
-
-			fmt.Println(builder.NCols(), builder.NRows())
+			if err := t.do(int(t.n), cr, c, builder, j); err != nil {
+				return err
+			}
 		}
 		return nil
 	})
--- a/stdlib/universe/chande_momentum_oscillator.go
+++ b/stdlib/universe/chande_momentum_oscillator.go
@@ -1,0 +1,297 @@
+package universe
+
+import (
+	"fmt"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+)
+
+const ChandeMomentumOscillatorKind = "chandeMomentumOscillator"
+
+type ChandeMomentumOscillatorOpSpec struct {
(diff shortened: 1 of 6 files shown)


ewendai:
There's some (a lot of) repetitive code. If there are suggestion on how to condense, that'd be great.

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -0,0 +1,348 @@
...
+						}
+					}
+				}
+			} else {
+				if err := t.do(int(t.n), cr, c, builder, j); err != nil {
+					return err
+				}
+			}
+
+			fmt.Println(builder.NCols(), builder.NRows())
```
jsternberg: This looks like an extra println.

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	return tbl.Do(func(cr flux.ColReader) error {
+		if cr.Len() == 0 {
+			return nil
+		}
+
+		for j, c := range cr.Cols() {
+			if !doChandeMomentumOscillator[j] {
+				switch c.Type {
```
jsternberg: I think there's an `AppendValues` function that should be better at this instead of including a type switch for copying over values.

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -0,0 +1,348 @@
...
+				case flux.TFloat:
+					arr := cr.Floats(j)
+					for i := int(t.n); i < arr.Len(); i++ {
+						if err := builder.AppendFloat(j, arr.Value(i)); err != nil {
+							return err
+						}
+					}
+				}
+			} else {
+				if err := t.do(int(t.n), cr, c, builder, j); err != nil {
```
jsternberg: Instead of `else`, you should put a `continue` at the end of the above if block and then have this second part just stay here unindented. The `else` is unnecessary.

ewendai: Is there a reason why this is better?

jsternberg: It reduces indentation which makes it easier to read. But the main reason is just because it's part of the Go style guide.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

stdlib/universe/chande_momentum_oscillator.go
ewendai: Is there an example code snippet? I don't know how to use `AppendValue`. Or how to make type `*array._____` into `[]values.Value`.

ewendai: Nevermind, I finally figured it out.

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -0,0 +1,307 @@
...
+
+func (t *chandeMomentumOscillatorTransformation) do(n int, cr flux.ColReader, c flux.ColMeta, builder execute.TableBuilder, j int) error {
+	var sumUp float64
+	var sumDown float64
+	sumUp = 0
+	sumDown = 0
+
+	switch c.Type {
+	case flux.TInt:
+		arrValues := cr.Ints(j)
```
aanthony1243: it looks to me like all of the values in this case besides `curr` and `prev` are float types.  Since  `float(curr) - float(prev) == float(curr - prev)`   it seems like you can move most of this logic into a function   `func nextCMO(curr, prev float) (float,float)`   that you call like,   `val, diffNAgo = nextCMO(float64(curr), float64(prev))`  

Then all of these cases become much shorter, easier to read.

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -198,36 +198,19 @@ func (t *chandeMomentumOscillatorTransformation) do(n int, cr flux.ColReader, c
...
 	sumUp = 0
 	sumDown = 0
+	diffNAgo := 0.0
 
 	switch c.Type {
 	case flux.TInt:
-		arrValues := cr.Ints(j)
 		c.Type = flux.TFloat
-		prev := arrValues.Value(0)
+		prev := execute.ValueForRow(cr, 0, j).Int()
```
aanthony1243: original code:  
```
              arrValues := cr.Ints(j)	
		c.Type = flux.TFloat			
		prev := arrValues.Value(0)			
		curr := int64(0)			
		for i := 0; i < cr.Len(); i++ {		
			curr = arrValues.Value(i)				
			diff := float64(curr - prev)	
			if diff > 0 {	
				sumUp += diff	
			} else if diff < 0 {	
				sumDown -= diff	
			}	

 			if i >= n {				
				val := 100 * (sumUp - sumDown) / (sumUp + sumDown)					
				if err := builder.AppendFloat(j, val); err != nil {	
					return err	
				}	

 				diffNAgo := float64(arrValues.Value(i-n+1) - arrValues.Value(i-n))	
				if diffNAgo > 0 {	
					sumUp -= diffNAgo	
				} else if diffNAgo < 0 {	
					sumDown += diffNAgo	
				}	
			}				

 			prev = curr
```

new helper: 
```
func nextCMO(sumUp, sumDown, diff, diffNAgo float64) (float64, flo6t64, float64) {
			if diff > 0 {	
				sumUp += diff	
			} else < 0 {	
				sumDown -= diff	
			}	
                      val := 100 * (sumUp - sumDown) / (sumUp + sumDown)	
  			if diffNAgo > 0 {	
				sumUp -= diffNAgo	
			} else  {	
				sumDown += diffNAgo	
			}	
                      return val, sumUp, sumDown
}
```

new version: 
```
              arrValues := cr.Ints(j)	
		c.Type = flux.TFloat			
		prev := arrValues.Value(0)			
		curr := int64(0)			
		for i := 0; i < cr.Len(); i++ {		
			curr = arrValues.Value(i)				
			diff := float64(curr - prev)	
			if i >= n {		
                             diffNAgo := float64(arrValues.Value(i-n+1) - arrValues.Value(i-n))			
				val, sumUp, sumDown := nextCMO(sumUp, sumDown, diff, diffNAgo)		
				if err := builder.AppendFloat(j, val); err != nil {	
					return err	
				}	
			}  else {		
                           _, sumUp, sumDown = nextCMO(sumUp, sumDown, diff, 0)
                      }
 			prev = curr
```

stdlib/universe/chande_momentum_oscillator.go
```diff
@@ -0,0 +1,295 @@
...
+		arrValues := cr.UInts(j)
+		c.Type = flux.TFloat
+		prev := arrValues.Value(0)
+		curr := uint64(0)
+		for i := 0; i < cr.Len(); i++ {
+			curr = arrValues.Value(i)
+			diff := float64(curr - prev)
+			if i >= n {
+				diffNAgo := float64(arrValues.Value(i-n+1) - arrValues.Value(i-n))
+				val, su, sd := nextCMO(sumUp, sumDown, diff, diffNAgo)
```
aanthony1243: nit, you can just write this as val, sumUp, sumDown = nextCMO

probably you got an error about val.   Just declare it along with sumUp and sumDown at the beginning of this function.

PR #1613, opened 2019-07-26
docs: 2nd iteration of type inference

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/Type_Inference.md
+++ b/docs/Type_Inference.md
@@ -1,79 +1,280 @@
-## Flux and Type Inference
+#  ?????
+
+Have you seen this when using the Flux code editor?
+
+<Insert GIF of sweet autocompletetion>
+
+Pretty neat to be able to get that much help from an editor while writing code.
+
+Have you ever wondered how that worked? Today I'll take us on a bit of a deep dive on the behind the scenes that enable these autocompletion features in the editor.
+You may also be surprised to learn that the same features that make autocompletion work in the editor also make Flux safer and faster when working with data.
 
 
-Recently Flux saw an update that introduced a language feature called _type inference_.
-What is type inference and how does it help me as a user of InfluxDB?
+Let's jump right in, there are two main properties of Flux that make the above possible:
 
-What is type inference?
+1. Flux is strongly typed
+2. Flux is statically typed
+
+What do these terms mean?
+
+First let's define what a _type_ is in the context of programming languages.
+A _type_ in a programming language defines what kinds of values are allowed for a given variable.
+For example if I have the variable `threshold` in my Flux script and its type is `integer` then I know that the value of `threshold` can only be a whole number, it cannot be a string, a duration, or even a table of data.
+
+Flux is _strongly_ typed means that all variables in a Flux program have a known type and they do not change automatically.
+Flux is _statically_ typed means that Flux knows the type of the variables without having to run the program. This is sometimes described as knowing the types as compile time instead of at runtime.
+The opposite of a strongly statically typed language is a weakly dynamically typed language where variables do not have a known type until runtime and their type can change automatically.
+The languages Go and C are strong and static while Javascript is weak and dynamic.
+
+Here is a quick example to illustrate these points:
+
+    a = "1"
+    b = 1
+    a + b
+
+What happens if you execute that code in Flux?
+You will get this error:
+
+    Error: type error 3:1-3:6: string != int
+
+This is because Flux knows the types of `a` and `b` and you cannot add a string to an integer.
+
+What happens if you execute the same code in Javascript?
+You will get the string `'11'`. This is because Javascript automatically converted the type of `b` from a number to a string and then performed string concatenation.
+
+> As an aside, this is not a judgement on strong static languages versus weak dynamic languages. Each has their place and I hope this reading makes it clear why Flux chooses to be strongly and statically typed.
+
+Now armed with some new concepts let's break down how this helps us build rich editor features.
+Take for example this feature where we can make autocompletion suggestions based on the name of a function.
+
+<Insert Function Autocompletion GIF>
+
+This works by inspecting the code determining the type of the function by its name and then crafting some suggestions from the type.
+A functions type contains the information about which parameters it needs, what type those parameters have as well as the return type of the function.
+Using that information we construct detailed suggestions for what to insert next because we know exactly what parameters the function needs.
+
+This is only possible because we can know the function's type based on its name without having to execute the Flux program, i.e. static typing.
+
+As result of these two properties of Flux we get many benefits.
+We have seen the benefits it provides to editor features like the above.
+These editor features make developing Flux code a more productive experience because you get quick feedback as you write the code.
+Things like autocompletion make it quicker to write code and reduces typos.
+Helpful error messages that can be precise about the mistake make it quicker to fix typos.
+All of these editor features make you more productive.
+
+Additionally Flux can execute faster and safer because of its strong and static typing.
+The code Flux executes is faster because we know the type of the data ahead of time so when we process the data we do not have to constantly check the type of the data.
+
+Flux code also safer because data in Flux tables is known so it is consistent.
+A type is not restricted to just simple types like `float` or `string`.
+A type can also be composed of other types, for example the Flux table type is described as a set of columns where the type of each column is known.
+This means that you do not have to deal with the class of data cleaning problems where some of your data is of a different type to the rest of the data.
+Flux can ensure all records have the same type.
+
+At this point I hope its clear why we have designed Flux to be a strong and static language.
+If you would like to go a layer deeper we can explore how we implemented Flux such that it is both statically and strongly typed.
 
 
-Natural data types assigned to the data.
-1 + 1
+## Extra Credit
+
+For extra credit what follows is an explanation of the process of how Flux determines the types of values within a program.
+This process is known as _type inference_ and we use a specific implementation known as Algorithm W.
+
+Type inference is the idea that Flux can determine the type of the values within the program without requiring that the user to be explicit about types in the code.
+For example Flux knowns that the type of this line of code is a string:
+
+    message = "The system is down"
+
+To a reader of the code its obvious that message is a string because they can easily see its value is a string.
+Other strongly typed programming languages require that the author of the code tell the program that `message` is a string.
+For example here is the equivalent Go code:
+
+    var message string = "The system is down"
+
+The `string` keyword tells the Go compiler that the variable `message` has the string type.
+
+> If you are familiar with Go, you may be thinking that you can use the `:=` syntax instead. That is correct and is in fact a limited case of type inference within Go as well.
+
+In contrast Flux does not require those type annotations in the code because it can infer the type of any value without them.
+
+Remember algebra and solving a system of equations?
+It's those problems where they gave two equations that had two variables and you had to solve for both variables.
+Those problems are solved in a very similar manner to how Flux solves for the types in a program.
+For example given these two equations with variables `x` and `y` you can solve for both variables.
+
+    2x + y = 5
+    y = 2 + x
 
 
+The process of solving these variable involves _substituting_ one variable equation into the other equations.
+We have an equation what `y` is equal to, so we can substitute all values of `y` in the first equation with the right hand side of `y = ` equation.
 
-How does it help
+    2x + (2 + x) = 5
 
+Now that we have only a single variable `x` in the equation we _solve_ for `x` using the rules of algebra.
+Here are the steps explicitly:
 
-catch errors earlier
+    2x + (2 + x) = 5
+    3x + 2 = 5
+    3x = 3
+    x = 1
 
-errors that make more sense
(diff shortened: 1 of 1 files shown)


docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+
+* Nulls
+
+    Extensible records do not support accessing non-existent fields on an record.
+    Such operations will result in a unification error.
+    The Flux language specification states that accessing a non-existent field on a record will return an untyped null value.
+    The spec will need to be updated accordingly as this will no longer be the case.
+
+    Null values will still be possible at runtime but they must be typed nulls.
+    The concept of an untyped null will need to be removed from the spec.
```
nathanielc: I don't follow this bit, if we know that that field doesn't exist on the record then we know it will be a null when we access it. So we know it would have any other type but null. So where does the untyped vs typed nulls part come in?

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+    ```
+
+    This would result in a unification error in the case of extensible records.
+    What this means is that we lose some of our ability to represent polymorphic container types.
+
+    This might be concerning since the 2nd goal of this iteration is to represent table streams as a list of **heterogeneous** record types.
+    However such concern is unwarranted as streams cannot be constructed from atomic types the way list literals can.
+    Instead they are obtained entirely via a side effect operation - a call to the database.
+    This means that from the point of view of the type system, a table stream is just a list of **homogeneous** records.
+    In other words, record fields or columns that are never accessed never need to be typed.
```
nathanielc: Is this the same as saying that we now only track the lower bound of records? And the upper bound is now implicitly the set of all labels? (Using the lower, upper bound concept of kinded records)

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+3. `{a: 't | 'r} = {b: 'u | 'r} => error`
+
+Note here `=` is synonymous for unify.
+
+### Implications
+
+The Flux type system extends Hindley-Milner with a _kind_ system.
+Currently records are modeled as _kinds_ in this system.
+Extensible records however fall entirely within the _type_ system, not the _kind_ system.
+Describing records this way allows us to break the mutual recursion between type unification and kind unification.
```
nathanielc: What about the function kind? It does a similar thing were it models its parameters as  a lower/upper bound of set of parameters. Does the mutual recursion not occur because parameter type vars are never free?

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+    As a result `filter` expects to receive the same list type that `from` returns and so it expects `r.host` to be of type _string_.
+    This contradicts the usage of `r.host` as an _int_.
+
+* range
+
+    ```
+    forall ['r] where 'r:Rec (start: time, stop: time, tables: [ {_start: time | _stop: time | 'r} ]) -> [ {_start: time | _stop: time | 'r} ]
+    ```
+    Let's focus on the `tables` parameter.
+    `range` takes a `tables` parameter that is a list of records that must have fields `_start` and `_stop` and returns the same list type.
```
nathanielc: Technically `range` can accept data without `_start` and `_stop` columns and will add them if they do not exist. If my memory serves it is still possible to describe this using record extensions since our flavor will not allow duplicate fields. So it's implied they are update if they exist and added otherwise.

Taking a crack at describing this in the type language:

```
forall ['r] where 'r:Rec (start: time, stop: time, tables: [ { 'r} ]) -> [ {_start: time | _stop: time | 'r} ]
```

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+
+* filter
+
+    ```
+    forall ['r] where 'r:Rec (fn: (r: 'r) -> bool, tables: ['r]) -> ['r]
+    ```
+    `filter` takes a list of records and returns that same list of records.
+
+    From the type signatures of `range` and `filter` it should be clear why the above example program fails unification.
```
nathanielc: Will the usage information we can infer from the filter predicate function propagate both up and down the chain?


For example if the user does this `from() |> filter(fn: (r) => r.host == "a")` since `'r` is unmodified in `filter` then we infer that the record produced by `from` must have a `host` of type string?

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+
+* map
+
+    ```
+    forall ['r, 's] where 'r:Rec, 's:Rec (fn: ('r) => 's, tables: [ 'r ]) -> [ 's ]
```
nathanielc: Similar question here, will the constraints propagate back up the chain?

For example:

```
from(bucket:"foo")
   |> range(start: -1h)
   |> map(fn: (r) => {r with _value: r._value + r.host}) // assume `_value` is a string
```

In this case `r.host` inside the map function will constrain the input type of the `map` function which will propagate back up correct? There is no let-polymorphism here so the record type should unify all the way back to the `from` function record output type?

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+    `keep` and `drop` operations can be implemented using `map`:
+    ```
+    map(fn: (r) => ( {a: r.a | b: r.b | {} } )) // keep columns a and b
+    map(fn: (r) => ( {r - a, b} ))              // drop columns a and b
+    ```
+    Assigning a type signature to the actual `keep` and `drop` functions unfortunatly is not so simple.
+    These functions take the names of the columns they operate on via a `columns` parameter.
+    Assigning an accurate type signature to these functions requires the notion of compile time constants or label types [#1122](https://github.com/influxdata/flux/issues/1122).
+
+    Another option is a preprocessing step before type inference that rewrites the Flux AST by replacing calls to `keep` and `drop` with their `map` equivalents.
```
nathanielc: I see `label` types as a sort of preprocessing step anyways, so this fits. 

Tangential but I think we should actually implement `keep` and `drop` in pure Flux and add a planner rule to rewrite them to their specific transformations. 

Using psuedo syntax  from label types here are the `keep` and `drop` type signatures, these signatures assume only a single column is applied at once. In the case of multiple columns it can be seen as multiple calls to the function.

```
keep : forall ['r] where 'r:Rec  (column: label, tables: [ 'r ]) -> [ {[label] : r[label] } ]
drop : forall ['r] where 'r:Rec  (column: label, tables: [ 'r ]) -> [ { r - [label] } ]
```

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+
+* aggregates
+
+    Assigning accurate types to aggregates is also a bit challenging.
+    This is due to the fact the aggregates will implicitly drop columns that are not part of the group key.
+    However if aggregates didn't drop columns dynamically, we could type them like so:
+    ```
+    forall ['r] (column: str, tables: [ {_time: time | 'r} ]) -> [ 'r ]
+    ```
+    In words, aggregates take a list of records with a `_time` fields and return that same list minus the `_time` field.
```
nathanielc: Should we look into what it would take to model the group key in the type system? Maybe just a kinding constraint on the fields? Even if we did how would we model the dynamic behavior of dropping all non group keys fields?

Also aggregates always produce a `_value` column. So its really:

```
 forall ['r] (column: str, tables: [ {_time: time | 'r} ]) -> [ {_value: TYPE | 'r} ]
```

Where `TYPE` depends on the specific aggregate function.

docs/TypeInference.md
```diff
@@ -0,0 +1,218 @@
...
+This would give us partial information on the type of pivot's output.
+However this too might be too restrictive.
+Do users want to be able to group and pivot heterogeneous data?
+
+#### What about other data sources?
+
+Unlike InfluxDB, sources like `csv` and `sql` cannot have such a strict type signature.
+Instead sources like these need to be defined to return lists of generic records `forall ['r] (...) -> [ 'r ]`.
+However based on usage, type inference will be able to determine a more specific type for `'r`.
+Each source can then perform their own type checking to ensure the data on disk matches with the type of `'r` that is inferred.
```
nathanielc: Two thoughts here:

### Filter vs Type Check

 If we use type inference as a filter back to sources it solves a lot of problems and creates a few of its own.

 First, you are guaranteed to get semi-homogenous data out of the source, by semi-homogenous I mean that any field that we know about will have the expected type, but other fields may still exist. This solves many of the issues we were having with heterogeneous data. 

Second, it creates a problem where users may inadvertently filter their data instead of getting an error. In your previous example of using: `r.host == 1` in a filter predicate you would get an error because `from` knows that the other columns are of type string so a type error is produced and the user can fix their error. But as you point out for other sources this would result is data that does not have an integer `host` field to be filtered, resulting in no data being returned and the user being confused about why no data is read from the `csv` file. 


### Query by Example

The idea of query by example is the user provides a query that demonstrates the entire structure of the data they expect. This is what @aanthony1243 was discussing at the last all-hands. The idea is that the syntax and arguments of the `from` function (or whatever wrapper we put on it) provides enough information to get homogeneous records from sources. 

This seems to have the advantage of disambiguating a users intention. Meaning in the above case if the user has a typo related to types in such a query it would not be as surprising to the user that the data was filtered since the expectation was the query types were filters. 

In short, maybe these two approaches are the same thing and we just need to correctly educate users that types are filters?

docs/TypeInference.md
jlapacik: Unifying extensible records can only be well-defined if we error when a non-existent field is accessed. An example might help clarify.

Say I have a function that takes in a record that needs an `a` field of type _int_.
```
typeof f = forall ['t] where 't:Rec (r: {a:int | 't}) -> int
```
Then we call `f` with a record that does not have `a`:
```
f(r: {b: 3})
```
This results in the following unification implications:
```
{a:int | 't} = {b:int | {} }
    => 't = {b:int | 's} and {} = {a:int | 's}
```
How do we unify `{}` with `{a: int | 's}`?

I guess we could say that `'s = {}`, but what do we do with `a: int`? We've already inferred that `a` is an _int_. _int_ is a concrete monotype and we can't just switch it to be _null_. This has to be a unification error.

At runtime however, in the case of row functions, this won't happen. This is because if `f` is a row function defined in `map` or `filter`, it's never called in the language itself. It's execution is implied by the internal implementation of `map` or `filter`. Hence null values can exist as part of an integer column, the type system just sees them as integers and the interpreter/execution engine operates on them according to their semantics as defined in the spec.

docs/TypeInference.md
jlapacik: Yes. For example, take a record type like `{a:int | 'r }`. In the literature `'r` is referred to as a row variable. Technically `'r` is a type variable representing some unknown record type. Intuitively we can think of `'r` as a variable that ranges over every possible field or label.

docs/TypeInference.md
jlapacik: In the current system functions are actually _types_ not _kinds_. Hence there's no mutual recursion introduced and we can keep the current definition and not lose parity.

docs/TypeInference.md
jlapacik: Unfortunately record operations which either extend or update based on the presence of the field cannot be correctly type inferred in the proposed system. In particular we are unable to infer a principal type in certain situations. 

An example that I did not come up with:
```
f = (r) => if true then {r with x:1} else {x: 2}
```

If record extension includes update, then the type of `r` could be either `{}` or possibly `{x:int | {}}`.

As a result, our operations should either modify an already present field, or extend a record with a new field.

Fortunately I don't think this is too much of a restriction. In the case of `range` all it's doing is performing a filter on time. Hence it makes sense for `_start` and `_stop` to always be required.

docs/TypeInference.md
jlapacik: Yes, Hindley-Milner ensures this.

docs/TypeInference.md
jlapacik: Correct

docs/TypeInference.md
jlapacik: As I mentioned above, instead of conflating record extension and update, perhaps it would be best for aggregates to take in a record with `_time` and `_value` fields and return a list of the same record type?

As for modeling the group key, I'm not totally sure yet. A kind constraint might work. Another option is instead of treating a stream like a list of records, treat is as a record with a `group_key` property (that is again another record) along with a `data` property which is a list of records. This I think would force us into schema homogeneity though as you'd somehow have to specify what the group key from `from` should be. So still a lot of unknowns here. At least for me anyway.

docs/TypeInference.md
jlapacik: Having `from` implicitly filter rows based on the inferred types is heavily influenced by @aanthony1243's query by example proposal. And in the case of the type signature I propose for `from`, this would only apply to the `_value` columns as that is the only column for which types are unknown.

Note however that this really only applies to the influxDB source. Others like csv and sql I don't think should do any implicit filtering. Instead the sources themselves should implement type checking and if a type error occurs it happens at the source instead of halfway through the pipeline.

Also keep in mind that having `from` implicitly filter on types and even updating the current transformations with accurate type signatures does **not** fall under this proposal. It is something that we can do easily and iteratively after simplifying the _kind_ system and introducing extensible record types. But only the latter two are required as part of this 2nd iteration.

PR #1625, opened 2019-07-31
feat: MQTT output from flux query

davidgs:
### Done checklist
- [X] docs/SPEC.md updated
- [ X] Test cases written


stdlib/packages.go
```diff
@@ -11,6 +11,8 @@ import (
 	_ "github.com/influxdata/flux/stdlib/influxdata/influxdb/v1"
 	_ "github.com/influxdata/flux/stdlib/kafka"
 	_ "github.com/influxdata/flux/stdlib/math"
+	_ "github.com/influxdata/flux/stdlib/mqtt"
+	_ "github.com/influxdata/flux/stdlib/promql"
```
aanthony1243: we had some situations where we merged some promql bits then took them out.  i'm not sure what's going on here, but unless your code actually needs promql (probably not?)  let's remove this line.

davidgs: I thought this file was auto--generated? It says not to edit it directly as it is auto-generated.

davidgs:
Had been merging master instead of rebasing, so I have fixed all the commit messages, etc. and will resubmit.

PR #1628, opened 2019-08-02
test(e2e): add usage e2e tests for billing queries

affo:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


affo:
I was trying to run this test cases in InfluxDB, but they run too slow. Need to workout a solution for this.

stdlib/testing/usage/api.flux
```diff
@@ -0,0 +1,1688 @@
...
+
+#group,false,false,true,true,false,false,true,true,true,true,true,true
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,long,string,string,string,string,string,string
+#default,_result,,,,,,,,,,,
+,result,table,_start,_stop,_time,_value,_field,_measurement,endpoint,hostname,org_id,status
+,,36,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T11:00:01.017450939Z,379321,req_bytes,http_request,/api/v2/write,gateway-77ff9ccb7d-vcbpg,03cbe13cce931000,204
+,,36,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T11:05:00.600950432Z,327507,req_bytes,http_request,/api/v2/write,gateway-77ff9ccb7d-vcbpg,03cbe13cce931000,204
+,,36,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T11:10:00.609220694Z,372594,req_bytes,http_request,/api/v2/write,gateway-77ff9ccb7d-vcbpg,03cbe13cce931000,204
+,,36,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T11:10:01.09095546Z,287155,req_bytes,http_request,/api/v2/write,gateway-77ff9ccb7d-vcbpg,03cbe13cce931000,204
+,,36,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08
```
wolffcm: Is this query right?  If it's not filtering on `_field` isn't it getting too many points for the count?

stdlib/testing/usage/reads.flux
```diff
@@ -0,0 +1,3219 @@
...
+import "testing"
+
+// This dataset has been generated with this query:
+// from(bucket: "system_usage")
+//     |> range(start: 2019-08-01T12:00:00Z, stop: 2019-08-01T14:00:00Z)
+//     |> filter(fn: (r) =>
+//         r._measurement == "http_request"
+//         and r._field == "resp_bytes"
+//         and r.endpoint == "/api/v2/query"
+//         and (r.status == "200" or r.status == "500" or r.status == "400")
```
wolffcm: I wonder why you needed to filter on `r.status` like this?  Seems like those are the only status codes that queryd can generate.

stdlib/testing/usage/storage.flux
```diff
@@ -0,0 +1,2210 @@
...
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:00Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:10Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:20Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:30Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:40Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:23:50Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:24:00Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:24:10Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a85ca099000
+,,0,2019-08-01T11:00:00Z,2019-08-01T14:00:00Z,2019-08-01T12:24:20Z,34,gauge,storage_usage_org_bytes,storage-3,038b7a
```
wolffcm: This query also does not filter on `_field`.  Seems like if another field was ever added, it would either produce a wrong answer or errors.

nathanielc:
@lukevmorris Can you take a look at this PR as well? Maybe you can answer some of @wolffcm 's questions above about the queries themselves?

stdlib/testing/usage/reads.flux
affo: 👍 I guess you are right. I could retry and comment out that line. Anyways, the result would be the same (in the case they are the only error codes that queryd can generate).

For the sake of curiosity, `401` and `403` are not possible?

affo:
@wolffcm First, thank you for the precise and review!

Queries on flux run super fast.

Yeah, they could be downsampled. My idea was to create a dataset as close to the data we usually deal with as possible to reproduce a close-to-reality behavior. I don't have a strong opinion on this and I could craft smaller datasets from the ones I used now.

One thing I could do is to only use the data that matches the first filter. I used more data in order to further test pushing down filters.

The cool thing of these datasets is that those results are reproducible using Chronograf on our buckets (as long as the retention policy keeps the data 😅 ).

@wolffcm @nathanielc @lukevmorris Any advice/opinion on how could I generate those datasets?

stdlib/testing/usage/reads.flux
wolffcm: No, actually you're right. I think you can get 401 and 403 from `queryd` now, since we now have a `flux.Error` type that can produce them.  Eg., if you try to read from a bucket you don't have read permissions for, you would get 401 or 403.

affo:
@lukevmorris could you provide some feedback on this?

@wolffcm has some points on the queries (they are an exact copy of the ones we use).

Apart from that, it seems that this PR accomplishes the DOD of its issue.

affo:
@lukevmorris @wolffcm 

__UPDATE__

The issues with `to` being slow on OSS does not bother us anymore.

> Does the data set for this need to be so large? 

I think @lukevmorris could have an answer to that :)

At this point, @wolffcm are we worried to commit input files that are so big?

I think we can solve this in the the future by generating the input with a custom flux source. For now, I guess it can be fine. Thoughts?

For the problems on the queries, I think we can create separate issues for them and change those queries and these ones too.

wolffcm:
@affo My concern isn't so much with the size of the files, but the time it takes to run the test.  I just don't want to add a lot of time to running `go test ./...`.  How long does it take to run?

affo:
@wolffcm Oh, I see. I answered above! That's what I get on my laptop (sorry for not running benchmarks):

image

PR #1640, opened 2019-08-05
docs(outputs): design for external services in Flux

affo:
### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/Outputs.md
+++ b/docs/Outputs.md
@@ -1,0 +1,202 @@
+# Flux Outputs
+
+This documentation contains the design of how Flux will produce outputs to external services.
+This doc has the aim of defining the tiny bits of Go functions that must be implemented to allow a pure Flux implementation for external service communication.
+ 
+## Endpoints, Encoders, and Transports
+
+There are three key components for an external service:
+ - an _endpoint_: it is the component that takes a table and pushes it to the external service.
+ - an _encoder_: it is the component that translates tables into something understandable by the external service.
+ - the _transport_: it is the communication protocol used to communicate with the external service.
+
+In a nutshell, endpoints use encoders to encode data and a transport to push data to external services.
+
+Encoders and transports are written in Go as `values.Function`s.
+For example there will be `json.encode`, as well as `csv.encode` for encoders, and `http.post` and `tcp.send` for transports.
+
+The minimal set that must be implemented is `http.post` and `json.encode`.
+
+Endpoints are implemented in pure Flux.
+
+## Flux
+
+Encoders are provided in Flux as `encode` functions.
+For example the JSON encoder is part of the `json` package.
+Encoders take an object `data` and return its encoding as a string:
+
+```flux
+encode: (data:object) -> bytes
+```
+
+Communication protocols are also implemented (in Go), so that one can, for example, send an HTTP request with JSON data in pure Flux:
+
+```flux
+import "http"
+import "json"
+
+data = {a: 1, b: 2}
+json_enc = json.encode(data)
+http.post(header: {"Content Type": "application-json"}, url: "http://some/url/", data: json_enc)
+```
+
+External services in Flux provide an `endpoint` function for building an endpoint.
+The `endpoint` function returns a factory function for building a specialized transformation for sending tables to the external service.
+The factory function is used to create multiple ways of sending data to the service, as such it accepts a `mapFn` to map each record in a table to a proper object.
+The fields in that object are used by the endpoint to build the message for the external service.
+Every external service must define how that object must be shaped for the communication to happen.
+
+For example:
+
+```flux
+import ext "external_service"
+
+someFilter = (r) => ...
+someOtherFilter = (r) => ...
+
+// Obtain the factory function with this configuration for the endpoint.
+endpoint = ext.endpoint(url: "https://url/to/external/service")
+
+someMapFn = (r) => {
+   // Generate payload from record.
+   return {title: r.someTag, body: string(v: r._value)}
+})
+someOtherMapFn = (r) => {
+   // Generate payload from record.
+   return {title: r.someOtherTag, body: string(v: r._value)}
+})
+
+// Invoke the factory to get custom transformations.
+to1 = endpoint(mapFn: someMapFn)
+to2 = endpoint(mapFn: someOtherMapFn)
+
+from(...)
+    |> range(...)
+    |> filter(fn: someFilter)
+    |> to1()
+
+from(...)
+    |> range(...)
+    |> filter(fn: someOtherFilter)
+    |> to2()
+```
+
+The output of `to` functions contains the result of the communication with the external service and varies on a per-service base.
+
+
+### An Example: Slack
+
+We provide below an example implementation for a Slack endpoint.
+The package provides an `endpoint` function that return the Slack endpoint.
+It also provides helpers to send a single message in a channel and build the request header.
+
+```flux
+package slack
+
+import "http"
+import "json"
+
+// This can be globally overridden in another package with:
+// slack.defaultURL = "https://slack.com/api/chat.anotherURL"
+option defaultURL = "https://slack.com/api/chat.postMessage"
+
+// `message` sends a single message to a Slack channel.
+message = (url=defaultURL, token, channel, text) => {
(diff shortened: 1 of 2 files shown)


nathanielc:
@affo I agree with @aanthony1243 here, I think we need to work out how to represent the various encodings as part of Flux. Encodings should be composable and definable from Flux.

affo:
@nathanielc @aanthony1243 got it guys! Working on that 👍

affo:
@nathanielc @aanthony1243 

Just changed so that starting from `encode` and a protocol, one can implement an endpoint in pure Flux.

docs/Outputs.md
```diff
@@ -0,0 +1,184 @@
...
+import "http"
+import "json"
+
+builtin endpoint
+
+// Util functions to create the header and the url from the endpoint configuration.
+header = (token, ...) => ...
+url = (...) => ...
+
+to = (header, url) => (tables=<-) => tables
```
aanthony1243: I think we should separate the concept of tables from the concept of message creation/sending as much as possible.  

Kapacitor has had so many feature requests/EAR's because it tries to have built-in message packing for all of these services.   

unless i'm misunderstanding the requirements, we are not trying to send bulk data to these services, but rather a single message to the service.  We should provide helpers to create such messages.  

for example, the slack POST API:   https://api.slack.com/methods/chat.postMessage

we should have a helper that generates the JSON perfectly for the given parameters.  

how does a table fit into a slack message?  I think that's less well-defined and up to the user.    Separating these tasks will help us with maintainability and re-usability.  Packaging up a large table into a compact CSV-string version, for example, can be used by probably most alert formats.

docs/Outputs.md
```diff
@@ -0,0 +1,184 @@
...
+conn = tcp.dial(url: "tcp://some/url/")
+tcp.send(connection: conn, data: csv_enc)
+tcp.close(connection: conn)
+```
+
+External services in Flux provide an `endpoint` function for building an endpoint.
+All `endpoint` functions return an object for the endpoint.
+Every `endpoint` function has at least these parameters:
+ - `name` for the user-defined name of the endpoint.
+ - `async` for sending results asynchronously.
```
nathanielc: I think we should leave the `async` bits out for now. It complicates the design, async is hard especially in functional languages. I think we should strive to have the API ambivalent to to wether the io operation is synchronous or asynchronous.

docs/Outputs.md
```diff
@@ -0,0 +1,184 @@
...
+data = {a: 1, b: 2}
+
+json_enc = json.encode(data)
+csv_enc = csv.encode(data)
+
+http.post(header: {"Content Type": "application-json"}, url: "http://some/url/", data: json_enc)
+
+conn = tcp.dial(url: "tcp://some/url/")
+tcp.send(connection: conn, data: csv_enc)
+tcp.close(connection: conn)
```
nathanielc: I like the general pattern here of passing raw data to transports. I think this will work well.

The pattern of opening a connection and then closing it is very imperical and not very functional friendly, we should probably rethink that pattern. That said `tcp` is not part of the current requirements so we can punt on that design if it proves to be very challenging.

docs/Outputs.md
affo: @aanthony1243 I see you point! I can expose the method used in `map` to the user. I'll do that right away.

docs/Outputs.md
```diff
@@ -0,0 +1,225 @@
...
+
+import "http"
+import "json"
+
+// This can be globally overridden in another package with:
+// slack.defaultURL = "https://slack.com/api/chat.anotherURL"
+option defaultURL = "https://slack.com/api/chat.postMessage"
+
+// Util function to create the header.
+header = (token, ctype="application/json") => ({
```
nathanielc: The functions in the package make up the public API of the package, since `header` is not needed publicly and is only called in one location, I would prefer to see this logic moved into the `message` function instead of exposing it directly.

docs/Outputs.md
```diff
@@ -0,0 +1,225 @@
...
+// The `mapFn` must return an object with `channel` and `text` fields.
+endpoint = (url=defaultURL, token) =>
+    // The factory function.
+    (mapFn) =>
+        (tables=<-) => tables
+            |> map(fn: mapFn)
+            |> map(fn: (r) => {
+                resp = message(url: url, header: header(token), channel: r.channel, text: r.text)
+                return {r with status: resp.status}
+            })
```
nathanielc: Hmm, I really like the map patern but it means that the data returned from the endpoint is not a pass through rather its the result of the mapFn function. We need the original data to be returned with the status.

Perhaps we should do it like this:

```
endpoint = (url=defaultURL, token) =>
    // The factory function.
    (mapFn) =>
        (tables=<-) => tables
            |> map(fn: (r) => {
                data = mapFn(r)
                resp = message(url: url, header: header(token), channel: data.channel, text: data.text)
                return {r with status: resp.status}
            })
```

docs/Outputs.md
```diff
@@ -0,0 +1,225 @@
...
+    	crit: (r) => r._value > 90,
+        warn: (r) => r._value > 80,
+    )
+    
+checks |> alerts.notify(name: "slack", endpoint: slack_ep(mapFn: (r) => ({channel: ..., text: ...})))
+checks |> alerts.notify(name: "fileLog", endpoint: file_ep(mapFn: (r) => ...))
+checks |> alerts.notify(name: "pagerDuty", endpoint: pgr_ep(mapFn: (r) => ...))
+```
+
+### Replaying Notifications on Error
```
nathanielc: Lets remove this bit, its not currently required and we can add it on later.

docs/Outputs.md
affo: I see your point 👍

docs/Outputs.md
```diff
@@ -0,0 +1,226 @@
...
+Endpoints are implemented in pure Flux.
+
+## Flux
+
+Encoders are provided in Flux as `encode` functions.
+For example the JSON encoder is part of the `json` package.
+Encoders take an object `data` and return its encoding as a string:
+
+```flux
+encode: (data:object) -> string
```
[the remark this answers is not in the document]
affo: Are you sure we have this?

From the SPEC

```
bool     // boolean
int      // integer
uint     // unsigned integer
float    // floating point number
duration // duration of time
time     // time
string   // utf-8 encoded string
regexp   // regular expression
type     // a type that itself describes a type
```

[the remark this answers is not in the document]
nathanielc: Hmm I thought we did, well we need it, we can't enforce all encoders to produce utf-8 encoded data and we need to keep strings utf-8 encoded.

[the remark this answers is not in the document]
nathanielc: We have a reference to it here, but its not complete https://github.com/influxdata/flux/blob/master/docs/SPEC.md#annotations

docs/SPEC.md
```diff
@@ -468,6 +468,12 @@ Examples:
     2018-07-01T00:00:00Z + 2y  // 2020-07-01T00:00:00Z
     2018-07-01T00:00:00Z + 5h  // 2018-07-01T05:00:00Z
 
+#### Binary types
+
+A _binary type_ represents a base64 encoded sequence of bytes as defined in RFC 4648.
+The binary type name is `bytes`.
+The binary type is nullable.
```
nathanielc: This binary doesn't need to be base64 encoded. That was just how the csv format would represent the bytes type. Can you change this so say that it is a sequence of byte values? 

Also can you add it to the list of named types?

docs/SPEC.md
```diff
@@ -466,6 +466,12 @@ Examples:
     2018-07-01T00:00:00Z + 2y  // 2020-07-01T00:00:00Z
     2018-07-01T00:00:00Z + 5h  // 2018-07-01T05:00:00Z
 
+#### Binary types
+
+A _bytes type_ represents a sequence of byte values.
+The bytes type name is `bytes`.
+The bytes type is nullable.
```
jlapacik: Why is a `bytes` type _nullable_? Is `bytes` just an alias for `[]byte`? Arrays are not _nullable_.

Issue #1535, opened 2019-07-16
Map types cant have variable as a key


roshie548:
Currently, map can't use variables from the scope of the function in which they are used.

In my case, I'd like to define double exponential moving average completely in flux, which should be possible because it's computed using exponential moving average. The flux definition would look like:

```
doubleExponentialMovingAverage = (n, column="_value", tables=<-) =>
    tables
          |> exponentialMovingAverage(n:n)
          |> duplicate(column:column, as:"ema")
          |> exponentialMovingAverage(n:n)
          |> map(fn: (r) => ({r with column: 2.0*r.ema - r.column}))
          |> drop(columns: ["ema"])
```
Rather than replacing `column` with the `column` parameter value, `map` treats `column` as a string literal.

nathanielc:
@roshie548 Have a look at #1122 that should fix this issue for this use case.

ewendai:
`timeColumn` in the following example has the same issue:

```
truncateTimeColumn = (timeColumn, unit, tables=<-) =>
    tables
        |> map(fn:(r) => ({r with timeColumn: date.truncate(t: r.timeColumn, unit: unit)}))
```

Issue #701, opened 2018-06-15
first and last functions should have n argument


pauldix:
The `first` and `last` functions should take an argument `n` which specifies the number of results to return. It should default to 1

jlapacik:
After updating the type signature of `first` I would like to see the go implementation removed and for `first` to be defined in terms of `limit` like what was mentioned by @odd (just without the sort):
```
first = (tables=<-, n=1) => tables |> limit(n: n)
```

jsternberg:
I don't think we can use `sort()` for either for either of them because the behavior is different. I would be interested in seeing some kind of `tail()` function or to add that behavior as part of limit so that both `first` and `last` can be implemented in the same way.

jsternberg:
@jlapacik the `first` function needs to have a `column` parameter even if it is unused.

roshie548:
With `tail()` implemented:
```
first = (table=<-, n=1, column) => table |> filter(fn: (r) => exists(r.column)) |> limit(n: n)
last = (table=<-, n=1, column) => table |> filter(fn: (r) => exists(r.column)) |> tail(n: n)
```
However, we currently can't use the `column` parameter in the filter (see #1535)

PR #1641, opened 2019-08-05
feat(stdlib/universe): Tail function

roshie548:
- for each input table, tail returns the last `n` records before the last `offset` records

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/tail.go
+++ b/stdlib/universe/tail.go
@@ -1,0 +1,206 @@
+package universe
+
+import (
+	"fmt"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/pkg/errors"
+)
+
+const TailKind = "tail"
+
+// TailOpSpec tails the number of rows returned per table.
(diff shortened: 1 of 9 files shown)


roshie548:
Ideally, this could be used to rewrite `first` and `last` as described in #701 . According to the spec, though, `first` and `last` should select non-null records with respect to a provided `column` parameter. In pure flux, this is being blocked (see #1535  and #1122)

jlapacik:
Is `tail` sort of like the opposite of `limit`?

stdlib/universe/tail.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	}
+
+	cols := tbl.Cols()
+	isEmpty := false
+
+	n := t.n
+	offset := t.offset
+	index := 0
+	window := make([][]values.Value, n+offset)
```
jlapacik: Instead of allocating a table of values, would it be possible to calculate the beginning and ending row indices, and then use [appendSlicedCols](https://github.com/influxdata/flux/blob/11215b685510974f3742a649449ec5a1bf9d03e9/stdlib/universe/limit.go#L177-L232)? This way we don't have to allocate any untracked memory.

PR #1643, opened 2019-08-05
feat(streams): add the basic streams type and functions

ewendai:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -441,6 +441,12 @@ func (t *TableObject) Array() values.Array {
 func (t *TableObject) Object() values.Object {
 	return t
 }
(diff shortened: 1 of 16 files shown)


aanthony1243:
@nathanielc  this is good code, but we're holding off on it due to security concerns, right?

nathanielc:
@aanthony1243 I don't remember raising a security concern with this code, but its been a while.

The main question for me is do we want `stream` to be a fundamental data type of Flux? Or should we do something different(i.e. introduce traits/interfaces and make a stream just an iterable)

PR #1648, opened 2019-08-06
docs(secrets.md): add documentation for design around the secrets API

aanthony1243:


docs/secrets.md
```diff
@@ -0,0 +1,44 @@
+Notification endpoints will require credentials in order to communicate with the various external services.
+
+We do not want users to include those credentials in plain text into Flux scripts. We need to provide a mechanism for users load secrets from a safe storage mechanism and provide those secrets to the endpoints.
+
+**Backing Store:** : There are many backing store possibilities, which need not be discussed here.  Any store that provides secure access for key/value lookup may be used by providing an implementation of the KeyLookup  interface in the flux repository.  
+
+**Fine Grained Authorization**:  keys will generally not be accessible in the flux language.  A secret-enabled function will accept meta-data about the secret (a key) and use it to look up the stored value in memory. The lookup will error if the key is not found or if the user is not authorized to use that key.
```
nathanielc: What does the Flux code look like for retrieving a secret? Can the authorization check also be performed there so that the error occurs closer to accessing the secret?

docs/secrets.md
```diff
@@ -0,0 +1,44 @@
...
+- Not return secrets under any circumstances (e.g. in error messages) 
+
+The external interface for secret lookup is: 
+```Go
+func LookupKey(keyObject values.Object) (string, error)
+```
+The keyObject is a values.Object type because: 
+
+- The secret parameter can be passed to the service without being modified
+- The parameter may be validated by the service, minimizing the dependency of the calling function to have prior knowledge of the key lookup protocol.
```
nathanielc: I am not sure I like the values.Object as part of the external interface, as it means that each implementation needs to understand how to extract the key information from the object. 

What about a helper function that knows how to extract the key information from a values.Object consistently, then neither endpoints or implementations of a secret lookup need to know how it works?

docs/secrets.md
```diff
@@ -0,0 +1,44 @@
...
+slackID = {secretKey: "FLUX_SLACK_ID"}
+
+but the object may include supplementary information for consumption by the calling function.  They may also be embedded + replaced in another values.Value type (e.g. an encapsulating Object or Array) by the calling function as in:
+
+http.post(header: {"Content Type": "application-json", "ApplicationSecret": {secretKey: "FLUX_HTTP_POST_KEY"}}, url: "http://some/url/", data: json_enc)
+
+
+A final implementation may look like: 
+```Go
+// calling flux:  slack.message(message: {token: {secretKey: "FLUX_SLACK_ID"}, text: "hello slack"})
```
affo: I think it could be more user friendly to  have this resulting flux:

```
import "vault"
token = vault.get("mytoken")
slack.message(token: token, text: "hello slack")
```

We could achieve this by defining a `secret` type that is basically a promise for a `string`, so that in Flux you won't be able of calling it, accessing its parts, print it, etc.

In Go, you can use its internal function and retrieve the token.

docs/secrets.md
aanthony1243: yeah, same page.  there will need to be a middle layer in this design.  i'm thinking of "service" in abstract terms of something that implements an API.

docs/secrets.md
aanthony1243: I spent some time discussing with Nathaniel about this. we determined not to implement a new type.    However, my design is compatible with this look/feel.  

`vault.Get(key: "SLACK_ID")`  can be provided to create the object so that users don't have to memorize/risk mistyping the _required_  object type schema.  This would also allow us to extend the API in a backward compatible way.  

I do kind of like the idea of a token being a promise for a string, but it would open up security risks:  what functions are privileged to evaluate the `token` as a string?  Can it be evaluated via string interpolation (a forthcoming feature)?  When is it evaluated?  do we delay string interpolation until runtime? If they do evaluate the token in a non-privileged setting, do we error or provide a description of the promise?  So then the identifier has a different evaluation depending on context, which is unusual.   it's reasons like these that we ruled out "it's a string, but not really"  type of approaches.  

Also considered:  encrypting the keyvalue and just storing it as a regular string.  this also has complexity in terms of key management and proper security protocols to prevent a breach.

affo: I see your point.

Anyways, will that object be "special". I mean nobody should get anything out from it. At least not in  Flux.

The Go code can extract the secret. Wasn't this the assumption?

I am  quite not following how that object masks the token from the doc 🤔

docs/secrets.md
```diff
@@ -0,0 +1,65 @@
...
+
+Finally, a nice-to-have, but not required behavior would be to enforce that the token-lookup object be created by the vault.Get call.  That is, the user should get an error if they construct a request object by hand.  
+
+A final implementation may look like: 
+```Go
+// calling flux:  
+//     import "vault"
+//     import "slack"
+//
+//     tok = vault.get(key: "FLUX_SLACK_ID")
```
jlapacik: If I follow what you're saying, `tok` is just an ordinary object. It has properties and they can be accessed just like any other object. And it doesn't contain any privileged information. Functions that consume this object however must be implemented in Go because they will use the information in that object to actually obtain the privileged information and use it.

Is this correct?

aanthony1243: that's correct

docs/secrets.md
aanthony1243: it's just   

`vault.Get = (key) = ({secretKey: key})`

or if we determine that the API needs more information (userid, orgid, ...)  we can add it there too.

affo: Yeah, I got it. I think we are on the same page. It's not technically a promise for the token, but everything  that Go needs to build a token.

I wonder if we could obtain the token and keep it isolated in the Flux context.

What I mean is that I expect that:

```
token = vault.get("SLACK_ID")
```

would  error if there is some connection problem or else.

I am not expecting lazy evaluation here 🤔 .

Anyways, I agree that this is a valid  design :)

docs/secrets.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Be built-in to the language
+- Register an ExternalDependencies struct  (note: it will be straightforward and advantageous to refactor the `InjectFromDependencies` and similar functions into this pattern as well).   
+- Not return secrets under any circumstances (e.g. in error messages) 
+
+The external interface for secret lookup is: 
+```Go
+func LookupRequest(keyObject values.Object) SecretRequest
+func PreAuthLookup(keyObject values.Object) 
+func LookupKey(req SecretRequest) (string, error)
```
nathanielc: Can you change this to be written as an actual Go interface so its clear what implementations need to implement?

docs/secrets.md
```diff
@@ -0,0 +1,66 @@
...
+A function that requires secret lookup must: 
+
+- Be built-in to the language
+- Register an ExternalDependencies struct  (note: it will be straightforward and advantageous to refactor the `InjectFromDependencies` and similar functions into this pattern as well).   
+- Not return secrets under any circumstances (e.g. in error messages) 
+
+The external interface for secret lookup is: 
+```Go
+func LookupRequest(keyObject values.Object) SecretRequest
+func PreAuthLookup(keyObject values.Object)
```
nathanielc: What does PreAuthDo?

docs/secrets.md
```diff
@@ -0,0 +1,66 @@
...
+## Internal Functions
+A function that requires secret lookup must: 
+
+- Be built-in to the language
+- Register an ExternalDependencies struct  (note: it will be straightforward and advantageous to refactor the `InjectFromDependencies` and similar functions into this pattern as well).   
+- Not return secrets under any circumstances (e.g. in error messages) 
+
+The external interface for secret lookup is: 
+```Go
+func LookupRequest(keyObject values.Object) SecretRequest
```
nathanielc: Can you remove the `values.Object` type from this interface? Looking up secrets should not be coupled to how we represent values inside of the Flux interpreter. Can you change this function to accept the raw arguments it needs instead of extracting them from a `values.Object`?

docs/secrets.md
```diff
@@ -0,0 +1,66 @@
...
+
+The keyObject should minimally have a `secretKey` property: 
+slackID = {secretKey: "FLUX_SLACK_ID"}
+
+but the object may include supplementary information for consumption by the calling function.  They may also be embedded + replaced in another values.Value type (e.g. an encapsulating Object or Array) by the calling function as in:
+
+http.post(header: {"Content Type": "application-json", "ApplicationSecret": {secretKey: "FLUX_HTTP_POST_KEY"}}, url: "http://some/url/", data: json_enc)
+
+## Flux Package
+We will provide a `vault` package that will help the user create an object with the required properties for secret lookup.
```
nathanielc: The package should be called `secrets` or something generic, the fact we are using `vault` should be abstracted away.

dependencies/dependencies.go
```diff
@@ -0,0 +1,23 @@
...
+	Lookup(ctx context.Context, name string) (platform.ID, bool)
+	LookupName(ctx context.Context, id platform.ID) string
+}
+
+type Dependencies struct {
+	Reader             Reader
+	BucketLookup       BucketLookup
+	OrganizationLookup OrganizationLookup
+	Metrics            *metrics
+}
```
nathanielc: Does this need to be part of this PR?If so can you add comments to describe why/how it changing?

aanthony1243: no, just a dirty repo mistake.  this isn't even a proper design, just something I laid out for discussion

docs/secrets.md
aanthony1243: pre-auth can be executed by `secrets.Get(key: k)`  to make sure that the user is authorized to retrieve that key.  we could also just call `Lookup`  and see what it returns but I thought this was a clearer API.

nisley:
Just a comment that OSS will use BoltDB as the default secret store.  At this time, we have not settled on whether we'll provide additional store connectors in OSS.

docs/secrets.md
```diff
@@ -0,0 +1,62 @@
...
+for example, a http.post function may reference a token in the header object by having a sub-object with the form  `{secretKey: <string>}`
+http.post(header: {"Content Type": "application-json", "ApplicationSecret": {secretKey: "FLUX_HTTP_POST_KEY"}}, url: "http://some/url/", data: json_enc)
+
+## Flux Package
+We will provide a `secrets` package that will help the user create an object with the required properties for secret lookup.  
+A helper function is useful because we may wish to change the properties of a secret request over time, and we don't want users to have to update their code unless strictly needed.  
+
+The vault package contains: 
+
+- `secrets.get(key: "FLUX_SLACK_ID")`: returns an object with all properties needed to make a request, e.g.,  `{userID: "11918", orgID: "19119384", secretKey: "my-secret"}`
```
jsternberg: This is a bit confusing. Below here, you say it will be `vault.get` but up here it appears to be `vault.secrets.get`. Can this be clarified?

It also seems to me like the secrets package itself may work more as something that is influxdb specific. May I propose that we have this be `influxdata/influxdb/secrets` similar to how we're doing monitoring/alerts and we eschew any direct references to vault as anything other than an implementation detail?

stdlib/influxdata/influxdb/from.go
```diff
@@ -25,6 +26,8 @@ func init() {
 		Return:   flux.TableObjectType,
 	}
 
+	flux.RegisterDependency(FromKind, "bucketlookup", dependencies.BucketLookup())
```
jsternberg: Instead of named keys, should we have dependencies registered by the interface they implement? It would help to prevent name conflicts and would be decently easy. There's some precedent to doing this with `context.Context` and the `Set` method.

docs/secrets.md
nathanielc: Agreed, we should not reference vault in the Flux code.

stdlib/influxdata/influxdb/from.go
aanthony1243: truth: not sure how this got here.  was just playing with ideas.  will revert.

docs/secrets.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Internal Functions
+A function that requires secret lookup must: 
+
+- Be built-in to the language
+- Be secrets-aware, meaning that it can acquire the injected secret service, scan its input for secret references, and look up the values for the given secret keys.  
+- Not return secrets under any circumstances (e.g. in error messages) 
+
+The external interface for secret lookup is: 
+```Go
```
jsternberg: Can you fix the formatting here?

docs/secrets.md
aanthony1243: this is  mentioned below.

docs/secrets.md
```diff
@@ -0,0 +1,69 @@
...
+the internal function must be "secret aware" which means that it knows how to search its own input parameters for references to secrets, so that it can make the proper key-for-value substitution at the appropriate time (as close to when the key is used as possible)
+
+for example, a http.post function may reference a token in the header object by having a sub-object with the form  `{secretKey: <string>}`
+http.post(header: {"Content Type": "application-json", "ApplicationSecret": {secretKey: "FLUX_HTTP_POST_KEY"}}, url: "http://some/url/", data: json_enc)
+
+## Flux Package
+We will provide a `secrets` package that will help the user create an object with the required properties for secret lookup.  
+A helper function is useful because we may wish to change the properties of a secret request over time, and we don't want users to have to update their code unless strictly needed.  
+
+The vault package contains:
```
jsternberg: This references a `vault` package that doesn't exist.

docs/secrets.md
```diff
@@ -0,0 +1,69 @@
...
+We will provide a `secrets` package that will help the user create an object with the required properties for secret lookup.  
+A helper function is useful because we may wish to change the properties of a secret request over time, and we don't want users to have to update their code unless strictly needed.  
+
+The vault package contains: 
+
+- `secrets.get(key: "FLUX_SLACK_ID")`: returns an object with a single property: `{secretKey: "my-secret"}`
+
+The function will provide all that's needed to do the lookup.  It may also be implemented to pre-authorize that the user can access the given key, so that errors occur earlier.  
+
+Finally, a nice-to-have, but not required behavior would be to enforce that the token-lookup object be created by the vault.Get call.  That is, the user should get an error if they construct a request object by hand.
```
jsternberg: I don't think that this is necessary personally. I do think we should try to make it as hard as possible to accidentally create a secret key when not intending to such as by making it so the object property is something that very few if any people would ever create manually, but I think we should refrain from putting any nice to haves in the specification document.

I think this might work better if it were in a section "further improvements" or "other considerations".

docs/secrets.md
```diff
@@ -0,0 +1,69 @@
...
+
+The function will provide all that's needed to do the lookup.  It may also be implemented to pre-authorize that the user can access the given key, so that errors occur earlier.  
+
+Finally, a nice-to-have, but not required behavior would be to enforce that the token-lookup object be created by the vault.Get call.  That is, the user should get an error if they construct a request object by hand.  
+
+A final implementation may look like: 
+Go code: 
+```Go
+// calling flux:  
+//     import "secret"
```
jsternberg: I don't think this section is needed. I think it is getting too far into implementation for a specification document.

docs/secrets.md
aanthony1243: agreed.

docs/secrets.md
```diff
@@ -0,0 +1,52 @@
...
+// LoadSecret retrieves the secret value v found at key k 
+// ctx is the current query context
+    LoadSecret(ctx context.Context,k string) (string, error)
+}
+```
+
+the internal function must be "secret aware" which means that it knows how to search its own input parameters for references to secrets, so that it can make the proper key-for-value substitution at the appropriate time (as close to when the key is used as possible)
+
+for example, a http.post function may reference a token in the header object by having a sub-object with the form  `{secretKey: <string>}`
+http.post(header: {"Content Type": "application-json", "ApplicationSecret": {secretKey: "FLUX_HTTP_POST_KEY"}}, url: "http://some/url/", data: json_enc)
```
jsternberg: Something that occurs to me while I was looking at `http.post` is how would this work when it requires appending or prepending another string to the input?

The following will likely be common.

```
import "influxdata/influxdb/secrets"
token = secrets.get("influxdb_auth_token")
http.post(header: {"Authorization": "Token ${token}"})
```

It makes sense that string interpolation could be used here, but how would this end up working? There has to be some way to evaluate the string expression with secret substitution within the `http.post` function because of the requirements in the specification and I don't see this scenario covered in the document.

My original implementation was going to be just to check if the value was a `values.String` or `values.Object` and perform the secret substitution in that way, but I don't think it will be enough because my method doesn't cover the above.

I think this is possible but it also depends on how string interpolation is implemented as we might need to make string expressions lazily evaluated and the ability to evaluate a string expression with the secret service in certain contexts.

jsternberg: We talked on slack about potentially making it so `secrets.get` returned a function and that we could have something that evaluated functions within a secret-aware context. We'll have to flesh it out, but I'm going to approve the current method and we'll experiment with the secret aware stuff since string interpolation will very likely be needed in some capacity.

GeorgeMac: Could this be achieved by creating a `SecretStore` implementation that is namespaced / scoped by some configured secret? This would be before wiring it into flux. So in OSS it could have wide scope. Then in cloud the SecretStore could be user / org scoped?

PR #1651, opened 2019-08-06
feat(stdlib/universe): added aggregateMonths function

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/stdlib/universe/universe.flux
+++ b/stdlib/universe/universe.flux
@@ -2,6 +2,7 @@ package universe
 
 import "system"
 import "date"
+import "strings"
 
 // now is a function option whose default behaviour is to return the current system time
 option now = system.time
--- a/stdlib/universe/universe.flux
+++ b/stdlib/universe/universe.flux
@@ -291,6 +291,13 @@ truncateTimeColumn = (timeColumn="_time", unit, tables=<-) =>
     tables
         |> map(fn:(r) => ({r with _time: date.truncate(t: r._time, unit: unit)}))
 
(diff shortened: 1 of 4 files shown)


stdlib/universe/universe.flux
```diff
@@ -291,6 +292,13 @@ truncateTimeColumn = (timeColumn="_time", unit, tables=<-) =>
     tables
         |> map(fn:(r) => ({r with _time: date.truncate(t: r._time, unit: unit)}))
 
+// AggregateMonths applies an aggregate function to calendar months.
+aggregateMonths = (fn, column="_value", tables=<-) =>
+    tables
+        |> map(fn: (r) => ({r with monthYear: strings.joinStr(arr: [strings.substring(v: string(v: date.year(t: r._time)*100 + date.month(t: r._time)), start: 0, end: 4), strings.substring(v: string(v: date.year(t: r._time)*100 + date.month(t: r._time)), start: 4, end: 6)], v: "-")}))
```
aanthony1243: I was thinking about this later....what if we made `monthYear` type time instead of type string?  it should still have the same outcome.

also makes me think of `windowMonth`  that sets the _start as midnight on the first of that month, and _stop as 11:59 on the last day of that month.

jsternberg:
We just merged calendar time operation support (`1mo`) to flux natively so this can be done using the `window()` function. I'm going to close it for now because there's nothing really different from using `aggregateWindow`, but it might be nice to have a specific function that says "this is by months" to call out that we have support for this type of aggregation.

Issue #1639, opened 2019-08-05
Implement check function


nathanielc:
The check function has been defined in #1601. It might be able to implement in pure Flux in terms of a map function with conditionals. Prefer the pure Flux implementation.

wolffcm:
I am implementing this with a slight difference from the doc that was checked into Flux:
https://github.com/influxdata/flux/blob/11215b685510974f3742a649449ec5a1bf9d03e9/docs/Alerts.md#check

Instead of adding `_name` to the group key, the `_name` column will just be added to the table without being in the group key.  This will allow the implementation of `check` to be in pure Flux.

The `_name` field should be part of the group key when the output of `check` is stored via the `write` option.  In order to achieve this in `influxdb`, changes to `to()` need to be prioritized instead.  That's this issue: #213

wolffcm:
New proposed implementation of `check` looks like this:
```
check = (tables=<-, name, crit, warn, info, ok, messageFn, extraTags) =>
    tables
    |> experimental.set(o: extraTags)
    |> map(fn: (r) => ({r with
        _name:  name,
        _level: if crit(r: r) then levelCrit else ...,
    }))
    |> map(fn: (r) => ({r with _message: messageFn(r)})
    |> experimental.group(mode: "extend", columns: ["_level", "_name"])
    |> experimental.group(mode: "extend", columns: objectKeys(o: extraTags))
    |> experimental.to()
```
I made issues for `experimental.group`, `experimental.to`, `objectKeys` and `experimental.set`, and this issue depends on all of them.

PR #1652, opened 2019-08-06
refactor(stdlib): change parameters to flux to() function

wolffcm:
The changes to SPEC.md in this PR describe updated parameters and behavior of the `to()` function.

Implementations of `to()` are in `influxdb` and `idpe`. I'll hold off on merging this until I get some agreement on this design and have the corresponding implementations ready.

The motivation for making this breaking change is to allow the `alerts.check()` function to be written in pure Flux:
https://github.com/influxdata/flux/issues/1639

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
 
 #### To
 
-The To operation takes data from a stream and writes it to a bucket.
-To has the following properties:
+The `to` operation takes data from a stream and writes it to a bucket.
+`to` has the following properties:
 
-| Name       | Type                  | Description                                                                                                                                                                                                                        |
-| ----       | ----                  | -----------                                                                                                                                                                                                                        |
-| bucket     | string                | Bucket is the bucket name into which data will be written.                                                                                                                                                                         |
-| bucketID   | string                | BucketID is the bucket ID into which data will be written.                                                                                                                                                                         |
-| org        | string                | Org is the organization name of the bucket.                                                                                                                                                                                        |
-| orgID      | string                | OrgID is the organization ID of the bucket.                                                                                                                                                                                        |
-| host       | string                | Host is the location of a remote host to write to. Defaults to `""`.                                                                                                                                                               |
-| token      | string                | Token is the authorization token to use when writing to a remote host. Defaults to `""`.                                                                                                                                           |
-| timeColumn | string                | TimeColumn is the name of the time column of the output.  Defaults to `"_time"`.                                                                                                                                                   |
-| tagColumns | []string              | TagColumns is a list of columns to be used as tags in the output. Defaults to all columns of type string, excluding all value columns and the `_field` column if present.                                                          |
-| fieldFn    | (r: record) -> record | Function that takes a record from the input table and returns an object. For each record from the input table `fieldFn` returns on object that maps output field key to output value. Default: `(r) => ({ [r._field]: r._value })` |
-
-TODO(nathanielc): The fieldFn is not valid and needs to change. It uses dynamic object keys which is not allowed.
+| Name              | Type           | Description                                                                                                                                                                                                                        |
+| ----              | ----           | -----------                                                                                                                                                                                                                        |
+| bucket            | string         | Bucket is the bucket name into which data will be written.                                                                                                                                                                         |
+| bucketID          | string         | BucketID is the bucket ID into which data will be written.                                                                                                                                                                         |
+| org               | string         | Org is the organization name of the bucket.                                                                                                                                                                                        |
+| orgID             | string         | OrgID is the organization ID of the bucket.                                                                                                                                                                                        |
+| host              | string         | Host is the location of a remote host to write to. Defaults to `""`.                                                                                                                                                               |
+| token             | string         | Token is the authorization token to use when writing to a remote host. Defaults to `""`.                                                                                                                                           |
+| measurementColumn | string         | MeasurementColumn is the measurement into which data will be written. Defaults to `"_measurement"`.                                                                                                                                |
+| fieldColumn       | string         | FieldColumn defines the column from which the field keys and values will be taken.  Defaults to `"_field"`.                                                                                                                                   |
+| timeColumn        | string         | TimeColumn is the name of the time column of the output.  Defaults to `"_time"`.                                                                                                                                                   |
+| valueColumn       | string         | ValueColumn is the name of the value column of the output.  Defaults to `"_value"`.                                                                                                                                                 |
 
 Either `bucket` or `bucketID` is required.
 Both are mutually exclusive.
 Similarly `org` and `orgID` are mutually exclusive and only required when writing to a remote host.
 Both `host` and `token` are optional parameters, however if `host` is specified, `token` is required.
 
-_NOTE_: make sure that `fieldFn`'s parameter names match the ones specified above (see [why](#Transformations)).
+By default `to` allows data produced by an InfluxDB source to be easily rewritten to a new bucket, like so:
+```
+from(bucket: "my-bucket") |> range(start: -1h) |> to(org: "my-org", bucket: "my-other-bucket")
+```
 
-For example, given the following table:
+The `_measurement`, `_field`, `_time` and `_value` columns will be used to describe the data that 
+is written to the destination bucket.
+The special columns `_start` and `_stop` will be ignored by `to`.
+Any other columns will be used as tag keys and tag values in the data written to the destination bucket.
+If a column that would be used for tag keys and values does not have type string, then an error will be returned.
+
+For example, given the following input table:
 
 | _time | _start | _stop | _measurement | _field | _value |
 | ----- | ------ | ----- | ------------ | ------ | ------ |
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
...
+| measurementColumn | string         | MeasurementColumn is the measurement into which data will be written. Defaults to `"_measurement"`.                                                                                                                                |
+| fieldColumn       | string         | FieldColumn defines the column from which the field keys and values will be taken.  Defaults to `"_field"`.                                                                                                                                   |
+| timeColumn        | string         | TimeColumn is the name of the time column of the output.  Defaults to `"_time"`.                                                                                                                                                   |
+| valueColumn       | string         | ValueColumn is the name of the value column of the output.  Defaults to `"_value"`.                                                                                                                                                 |
```
aanthony1243: so you expect the user to drop incompatible/unwanted columns before writing back?  e.g. we are losing the tagColumns parameter which seems like it could still be useful.

wolffcm:
@nathanielc Not yet, I just wanted to change the SPEC until everything is ready, I'll remove the changes to `to.go` from this PR.

One thing about this design is that it will no longer be possible to write to multiple series at the same time.  That used to be possible with the `fieldFn` parameter, and is a capability of line protocol, e.g, `m,tag1=foo field1=0,field2=10 0001`.

I guess that we're okay with that though?

docs/SPEC.md
```diff
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
...
+| host              | string         | Host is the location of a remote host to write to. Defaults to `""`.                                                                                                                                                               |
+| token             | string         | Token is the authorization token to use when writing to a remote host. Defaults to `""`.                                                                                                                                           |
+| measurementColumn | string         | MeasurementColumn is the measurement into which data will be written. Defaults to `"_measurement"`.                                                                                                                                |
+| fieldColumn       | string         | FieldColumn defines the column from which the field keys and values will be taken.  Defaults to `"_field"`.                                                                                                                                   |
```
aanthony1243: in general, the fieldFn is not a useful construct anymore since we can compute a new field using `with` in a map call.  So it makes sense to drop it.   However, should we allow the possibility for the field name to be the column name?  because that's how it would come out if we used `map`.  there's not a convenient way to 'unpivot' back to a _field column.

docs/SPEC.md
```diff
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
...
-
-TODO(nathanielc): The fieldFn is not valid and needs to change. It uses dynamic object keys which is not allowed.
+The `to` operation takes data from a stream and writes it to a bucket.
+`to` has the following properties:
+
+| Name              | Type           | Description                                                                                                                                                                                                                        |
+| ----              | ----           | -----------                                                                                                                                                                                                                        |
+| bucket            | string         | Bucket is the bucket name into which data will be written.                                                                                                                                                                         |
```
jlapacik: `bucket` and `bucketID` are specified as mutually exclusive. The type system does not allow for this. Instead I propose we only have a single `bucket` parameter that can take either a bucket name or ID as input (they're both strings after all).

I say we do the same for `org` and `orgID`. Let's just have a single `org` parameter.

docs/SPEC.md
nathanielc: Unpivoting is simple using `map`,

```
|> map(fn:(r) => ({_field : "max", _value: r.max}))
```

This only works for a single value, if you have multiple values we would need a `flatMap` function. 


```
|> flatMap(fn:(r) => [{_field : "max", _value: r.max}, {_field: "min", _value: r.min}])
```

docs/SPEC.md
```diff
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
...
+| orgID             | string         | OrgID is the organization ID of the bucket.                                                                                                                                                                                        |
+| host              | string         | Host is the location of a remote host to write to. Defaults to `""`.                                                                                                                                                               |
+| token             | string         | Token is the authorization token to use when writing to a remote host. Defaults to `""`.                                                                                                                                           |
+| measurementColumn | string         | MeasurementColumn is the measurement into which data will be written. Defaults to `"_measurement"`.                                                                                                                                |
```
jlapacik: I would like to see these parameters removed (measurementColumn, fieldColumn, etc.) and instead have the input schema be enforced by the type system (eventually). In other words a user must provide `to` with tables having `_time`, `_field`, `_value`, `_start`, and `_stop` columns. The rest of the column types being string.

docs/SPEC.md
nathanielc: How would you distinguish bucketIDs from bucket names? A bucket name that looks like a bucketID is a valid bucket name. 

I agree the mutual exclusiveness of the arguments is odd, but we can't just merge them, maybe there is another way?

docs/SPEC.md
nathanielc: @jlapacik is this because it simplifies the type system to have the column names static?

If so what about the pattern of most other Flux functions to take a `column` parameter? 

Does the `label type` (aka const string type) solve this problem?

docs/SPEC.md
jlapacik: Perhaps you give precedence to bucketIDs. I.e you check if the provided argument is a valid bucketID and if not you check if it is a valid bucket name?

nathanielc: That would mean that bucket names cannot be valid bucketIDs, which is not true.

jlapacik: But is it possible for a user to have a bucket with name `mybucket` and another bucket whose ID is `mybucket`?  (I know `mybucket` is not a valid bucketID, but you get my point)

Another option is for `to` to only take bucket names exclusively (or bucketIDs exclusively).

docs/SPEC.md
jlapacik: @nathanielc yes as it stands right now (that is, without label types) in order to assign a valid type signature to `to`, the names and types of the columns must be known at compile time. This precludes us from passing the column names as input to the function itself.

I _think_ we could get around this with label types, but that would come with the added complexity of an AST preprocessor (or something similar). Although we'll probably need that anyway since like you mention, functions like `keep` and `drop` have the same issue.

docs/SPEC.md
nathanielc: Yes it is possible to have a bucket with ID `03cbbf8eb3a9e000` and another bucket with name `03cbbf8eb3a9e000`. It would be a bad idea as a user to create such a named bucket but its not illegal. 

I just tried it out on my prod account to make sure, and yes it is possible.

jlapacik: We could make it clear in the spec that we try bucketIDs first and only if that fails do we try bucket names?

docs/SPEC.md
wolffcm: Nice, `flatMap` would be a great addition.

@adam I think that within a single function it's tough to satisfy both the trivial case of `from |> range |> to` and to also allow more intuitively structured data (e.g., value column label becomes field key in influxdb) to work right.

Currently no one is using `fieldFn` and `tagColumns` in production, so the former case is all that we need to support.

Also, it's worth noting that when we have a working `flatMap` function, we could wrap this simple `to` with another function to handle data that has been pivoted.

docs/SPEC.md
```diff
@@ -3685,31 +3685,39 @@ from(bucket: "telegraf/autogen")
...
+| org               | string         | Org is the organization name of the bucket.                                                                                                                                                                                        |
+| orgID             | string         | OrgID is the organization ID of the bucket.                                                                                                                                                                                        |
+| host              | string         | Host is the location of a remote host to write to. Defaults to `""`.                                                                                                                                                               |
+| token             | string         | Token is the authorization token to use when writing to a remote host. Defaults to `""`.                                                                                                                                           |
```
affo: would this be of type `secret`?

docs/SPEC.md
aanthony1243: i'm not convinced this is the right move.  we set those as defaults to get this behavior, and I don't want to see the type system becoming an impediment to writing data to influxdb.  Furthermore, if we were being smart with our planner, any renaming should be rolled into this function anyway.  This idea also falls under the category of being "too helpful"  where the easy path feels great but doing anything advanced or unorthodox requires a herculean effort.  

Also, currently, a measurement name is required for writing data, I don't think we can work around that at this time.  One of the tags has to be specially earmarked as the measurement name.  

I would, however, prefer to eliminate the _field column.  But that's a different discussion -- I want it to be eliminated from the top of the chain to the bottom.  Field names should be column labels, period, and things like a default _value should be used more sparingly.  In theory, this would be a good place to start but it's awkward to require writing data back to influxdb in a format that's different than how it comes out.

docs/SPEC.md
aanthony1243: nobody is using fieldFn and tagColumns in production because it has an automagic default routine that happens to be inexpressible in the flux language syntax.  The current behavior that I expect most users will be surprised if it's lost is: 

_measurement == measurement name
field names:  all field names returned by fieldFn, which like you said is pretty much always whatever is in _field
tags: everything that's not a field column.  

With flatmap, you're looking at a headache of lookups from a flatmap:  _fieldNames1, _fieldNames2, ..., _value1, _value2, ....

I just want to make sure we like how that would work out.  It might be more future-proof to take that kind of data as input and pivot it.  Regardless, if we want to be prepared for this case then the field identifiers should be lists:  fieldColumns: ["_fieldNames1", "_fieldNames2", ...] and similarly for valueColumn(s)

aanthony1243: also, consider you are proposing this to perform a pivot in-line with the writes.  this is not a deal breaker, but it also is only half true to say "we don't have to pivot the data"

docs/SPEC.md
jlapacik: After discussing with the team, I will create a separate issue to address this.

wolffcm:
This PR is now obsolete.

PR #1653, opened 2019-08-07
feat: mqtt to() function

davidgs:
### Done checklist
- [ X] docs/SPEC.md updated
- [ X] Test cases written


onyxet:
Hi guys! Thanks for this great feature!
Any updates when it could be merged?

davidgs:
It's actually been merged a while ago, but it is in the `experimental` pkg, so all you need to do in a flux task is:

`import` experimental/mqtt
|> mqtt.to(
		broker: "tcp://yourbroker.com:port,
		topic: "topicname",
		clientid: "optional-client-id",
		valueColumns: ["_value"],
		tagColumns: ["cpu", "host"],
	)`
And you should be good to go!

onyxet:
@davidgs And we have to use InfluxDB v2 as I understood ?

davidgs:
I believe so, as there are no tasks in InfluxDB v1 -- I think it's still in the version of flux shipped with 1.8 so in theory you could I guess.

onyxet:
@davidgs Thank you for support

mike006322:
Could there be an option for the MQTT message to be in JSON format instead of line protocol?

davidgs:
You're welcomed to implement such a feature.

PR #1654, opened 2019-08-07
feat(stdlib): alerts package - pure Flux

affo:
This is WIP.

Just gathering opinion on:
 - how should I test this?
 - should I add something to the SPEC?

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -0,0 +1,9 @@
+package alerts
+
+builtin check
```
jlapacik: Let's add the builtins later.

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -0,0 +1,9 @@
...
+
+write = (tables=<-) => tables |> to(bucket: "system")
```
jlapacik: I think this will be implemented in Go now.

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -0,0 +1,9 @@
...
+
+from = () => from(bucket: "system")
```
jlapacik: I think `alerts.from` will automatically pivot the data now if I understand things correctly.

affo:
@jlapacik I guess you can create the  `alerts` package with deadman and we keep  this alive until we have the Go bits up (`check` and `write`).

Once we  have those, one can address this issue by adding `from`, `notify` and `log` in pure Flux and address this issue.

What do you think?

stdlib/influxdata/influxdb/alerts/alerts.flux
affo: It will be `experimental`, right?

PR #1655, opened 2019-08-07
fix(stdlib/universe): mode() now properly considers nulls when calcul…

sauren-khosla:
…ating the mode

### Done checklist
- [x] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1812,8 +1812,9 @@ from(bucket: "telegraf/autogen")
 
 ##### Mode
 
-Mode produces the mode for a given column. Null is considered as a potential mode if it is present. If there are multiple modes, all of them are returned in a table in sorted order. 
+Mode produces the mode for a given column. If there are multiple modes, all of them are returned in a table in sorted order. 
 If there is no mode, null is returned. The following data types are supported: string, float64, int64, uint64, bool, time.
+Mode only considers non-null values.
 
 Mode has the following properties: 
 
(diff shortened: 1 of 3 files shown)


docs/SPEC.md
```diff
@@ -1807,8 +1807,9 @@ from(bucket: "telegraf/autogen")
 
 ##### Mode
 
-Mode produces the mode for a given column. Null is considered as a potential mode if it is present. If there are multiple modes, all of them are returned in a table in sorted order. 
+Mode produces the mode for a given column. If there are multiple modes, all of them are returned in a table in sorted order. 
 If there is no mode, null is returned. The following data types are supported: string, float64, int64, uint64, bool, time.
+`Null` is not considered as a potential mode.
```
jsternberg: What does this mean?

stdlib/universe/mode.go
```diff
@@ -240,44 +240,37 @@ func (t *modeTransformation) doString(cr flux.ColReader, tbl flux.Table, builder
 		}
 	}
 
-	if numNil > max {
+	if numNil == total {
```
jsternberg: I guess this confuses me. If we're not considering null values, why do we need to count the null values? Is the intent here that if there are no values that it will emit null?

I'd like a comment on this and I think this could be potentially better done by appending nil if there were no values considered.

stdlib/universe/mode.go
```diff
@@ -240,44 +242,35 @@ func (t *modeTransformation) doString(cr flux.ColReader, tbl flux.Table, builder
 		}
 	}
 
-	if numNil > max {
+	// if every value occurs the most number of times or all values are null, there is no mode
```
jsternberg: Is there a reason for `every value occurs the most number of times ... there is no mode`?

If all values are null, then shouldn't `len(stringMode) == 0`?

sauren-khosla: Definitionally, if every value occurs the same/most number of times, there is no mode. 

len(stringMode) == 0 is equivalent to total == 0 since total is initialized to 0, and if the len(stringMode) is 0, the loop on line 237 never executes, so total stays 0.

PR #1656, opened 2019-08-07
fix(stdlib/sql): Close connection after sql.to() & Postgres datetime fix

roshie548:
- Connection to database is closed after Process() is finished

### DIFF
--- a/stdlib/sql/to.go
+++ b/stdlib/sql/to.go
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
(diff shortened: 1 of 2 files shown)


stdlib/sql/to.go
```diff
@@ -174,7 +174,8 @@ func (t *ToSQLTransformation) Process(id execute.DatasetID, tbl flux.Table) (err
 			return nil
 		}
 	}
-	return err
+
+	return t.db.Close()
```
aanthony1243: we probably want to keep this open for the duration of the transformation.  `Finish` is the right place.

roshie548:
> have you thought of any ways we can improve the tests on this?

@aanthony1243 
Other than spinning up docker instances or having dedicated test servers, not really

PR #1658, opened 2019-08-07
feat(universe): add ker and kama to flux

ewendai:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

#1541 
#1542


stdlib/universe/universe.flux
```diff
@@ -2,6 +2,9 @@ package universe
 
 import "system"
 import "date"
+import "math"
+import "strings"
+import "regexp"
```
wolffcm: I see you took these packages out of the prelude in `compile.go`, curious about why?

stdlib/universe/kaufmansAMA_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/influxdata/flux/stdlib/universe"
+	"testing"
+)
+
+func TestKamaOperation_Marshaling(t *testing.T) {
+	data := []byte(`{"id":"kaufmansAMA","kind":"kaufmansAMA","spec":{"n":1}}`)
+	op := &flux.Operation{
+		ID: "kaufmansAMA",
+		Spec: &universe.KamaOpSpec{
+			N: 1,
```
wolffcm: You could verify that the `columns` parameter is marshaled correctly here too.

stdlib/universe/kaufmansAMA_test.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	querytest.OperationMarshalingTestHelper(t, data, op)
+}
+
+func TestKama_PassThrough(t *testing.T) {
+	executetest.TransformationPassThroughTestHelper(t, func(d execute.Dataset, c execute.TableBuilderCache) execute.Transformation {
+		s := universe.NewkamaTransformation(
+			d,
+			c,
+			&universe.KamaProcedureSpec{},
```
wolffcm: So I guess that kama will just forward its input tables to its output if `n` is zero and `columns` is empty... Do both of these conditions have to be true for kama to be a passthrough?

Do we have other functions that behave this way, i.e., they become pass through functions for certain parameter values.  I could imagine returning an error for cases like this as well.

stdlib/universe/kaufmansAMA_test.go
```diff
@@ -0,0 +1,209 @@
...
+					{Label: "_time", Type: flux.TTime},
+					{Label: "_value", Type: flux.TFloat},
+				},
+				Data: [][]interface{}{
+					{execute.Time(14), 1.4444444444444446},
+					{execute.Time(14), 2.5802469135802473},
+				},
+			}},
+		},
+	}
```
wolffcm: You might want to add test coverage for some of the error cases too (like if a column is not a type that kama will accept).

stdlib/universe/kaufmansAMA_test.go
wolffcm: Looking over code in the `Process` method, it seems like `columns:[]` should be a pass through, but there should be an error returned if `n` is not positive.

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,259 @@
...
+						values = append(values, float64(arr.Value(i)))
+					}
+				case flux.TFloat:
+					arr := cr.Floats(j)
+					values = arr.Float64Values()
+				}
+				kers := t.makeKERS(values, int(t.n))
+				var kama float64
+				sc := 0.0
+				kama = values[t.n-1]
```
wolffcm: if `t.n` is greater than the number of rows in the table, then this will `panic`, right?  What would the best behavior for this be?  It could either return an error, or just return an empty table.  What does EMA do?

stdlib/universe/kaufmansAMA_test.go
```diff
@@ -0,0 +1,209 @@
...
+		data []flux.Table
+		want []*executetest.Table
+	}{
+		{
+			name: "flux.TInt input",
+			spec: &universe.KamaProcedureSpec{
+				N:       10,
+				Columns: []string{"_value"},
+			},
+			data: []flux.Table{&executetest.Table{
```
wolffcm: Some of the tests for EMA use row-wise tables:
https://github.com/influxdata/flux/blob/master/stdlib/universe/exponential_moving_average_test.go#L78

Could you add at least one test case here that does as well?

wolffcm: This will force the table block size to be exactly one row, so you can verify that KAMA works right for tables that span multiple blocks.

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,259 @@
...
+				for i := int(t.n); i < cr.Len(); i++ {
+					if err := builder.AppendValue(j, execute.ValueForRow(cr, i, j)); err != nil {
+						return err
+					}
+				}
+			} else {
+				var values []float64
+				switch c.Type {
+				case flux.TInt:
+					arr := cr.Ints(j)
```
wolffcm: I don't think this will work quite right for large tables.  Tables are partitioned into multiple "blocks" that by default can have up to 1000 rows.  The issue is that `table.Do` will be called multiple times for each block.  So at the 1001st row, it will start over assuming that it's a new table.

EMA uses data structures in a package called `moving_average`

wolffcm: EMA uses data structures from the pacakge `flux/internal/moving_average` to make this work right.  Since this is also a moving average, I wonder if you could reuse this code?
https://github.com/influxdata/flux/blob/master/stdlib/universe/exponential_moving_average.go#L154
https://github.com/influxdata/flux/tree/master/internal/moving_average

docs/SPEC.md
```diff
@@ -3550,7 +3550,107 @@ from(bucket: "telegraf/autogen"):
     |> relativeSearchIndex(n: 14)
 ```
 
+#### Kaufman's Efficiency Ratio
+
+Kaufman's Efficiency Ratio indicator is created by taking the absolute value of CMO and dividing by 100 to give the 0 to 1 range.
```
wolffcm: Can you expand CMO to its full name?  I think it will make this easier to understand.  Also, maybe add a sentence that tells what KER does, like `Kaufman's Efficiency Ratio provides a way of measuring the volatility of a series.`

docs/SPEC.md
```diff
@@ -3550,7 +3550,107 @@ from(bucket: "telegraf/autogen"):
...
+
+The default column `KaufmansER` takes in is "_value".
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| n           | int      | N specifies the period.
+
```
wolffcm: You want to include the `columns` parameter here, right?

wolffcm: Ah, there's no `columns` parameter for this one.

docs/SPEC.md
```diff
@@ -3550,7 +3550,107 @@ from(bucket: "telegraf/autogen"):
...
+ Example of script:
+ ```
+ from(bucket: "telegraf/autogen"):
+     |> range(start: -7d)
+     |> kaufmansER(n: 10)
+ ```
+
+#### Kaufman's Adaptive Moving Average
+
+Kaufman's Adaptive Moving Average is created by taking the absolute value of CMO and dividing by 100 to give the 0 to 1 range.
```
wolffcm: Same comment here regarding CMO.

compile.go
```diff
@@ -88,10 +88,6 @@ var (
 	prelude = []string{
 		"universe",
 		"influxdata/influxdb",
-		"math",
```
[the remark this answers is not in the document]
ewendai: @aanthony1243 removed those because otherwise "math.abs" would not perform as expected in `universe.flux`. Has to do with "math" should be imported first but sometimes it would be imported after because maps are processed randomly.

stdlib/universe/universe.flux
ewendai: re: above

stdlib/testing/testdata/kama_v2.flux
```diff
@@ -0,0 +1,78 @@
...
+import "testing"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,double
+#group,false,false,false,false
+#default,_result,,,
+,result,table,_time,_value
+,,0,2018-05-22T00:00:00Z,110.46
```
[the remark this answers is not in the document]
ewendai: done

compile.go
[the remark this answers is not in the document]
aanthony1243: there was a data race wrt the order the packages were loaded.   moving them to `universe.flux` has the same outcome

stdlib/universe/kaufmansAMA.go
ewendai: EMA doesn't consider this case.

compile.go
```diff
@@ -429,9 +425,6 @@ func (t *TableObject) PolyType() semantic.PolyType {
 func (t *TableObject) Str() string {
 	panic(values.UnexpectedKind(semantic.Object, semantic.String))
 }
-func (t *TableObject) Bytes() []byte {
```
wolffcm: What are these changes?  Seems unrelated to the KAMA work?

ewendai: i think i broke it when rebasing. fixing rn

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,313 @@
+package universe
+
+import (
+	"fmt"
+	"github.com/influxdata/flux/interpreter"
```
wolffcm: This import should be below with the other ones.

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,313 @@
...
+	t.sumDown = 0
+	t.diffNAgo = make([]float64, t.n)
+	t.count = 0
+
+	err := tbl.Do(func(cr flux.ColReader) error {
+		if cr.Len() == 0 || len(cr.Cols()) == 0 {
+			return nil
+		}
+
+		if cr.Len() < int(t.n) {
```
wolffcm: Hmm, I don't think this is quite right.  The value returned by `cr.Len()` only reflects the number of rows in this particular chunk of the table.  But the entire table could be composed of many chunks.  So it's okay if `cr.Len()` is less than N.  Flux should get the same answer no matter what the chunk size is.

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,301 @@
...
+			doKAMA[j] = true
+		} else {
+			_, err := builder.AddCol(c)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	t.prevValue = 0
```
wolffcm: Are these variables only used inside the call to `tbl.Do()` below?  If so, can you make them local variables to the Process method?

docs/SPEC.md
```diff
@@ -3613,7 +3613,107 @@ from(bucket: "telegraf/autogen"):
...
+ ```
+
+#### Kaufman's Adaptive Moving Average
+
+Kaufman's Adaptive Moving Average is designed to account for market noise or volatility. This trend-following indicator can be used to identify the overall trend, time turning points and filter data movements.
+
+| Name        | Type     | Description
+| ----        | ----     | -----------
+| n           | int      | N specifies the period.
+| columns     | []string | Columns is the list of all columns that `KaufmansAMA` should be performed on and is assumed to be ["_value"] if left unspecified.
```
wolffcm: I guess this needs to just be `column`.

stdlib/universe/kaufmansAMA_test.go
```diff
@@ -0,0 +1,309 @@
...
+			name: "flux.TInt input rowWise",
+			spec: &universe.KamaProcedureSpec{
+				N:      10,
+				Column: "_value",
+			},
+			data: []flux.Table{&executetest.RowWiseTable{
+				Table: &executetest.Table{
+					ColMeta: []flux.ColMeta{
+						{Label: "_value", Type: flux.TInt},
+					},
```
wolffcm: Can you add a test case that uses rowwise table *and* also has columns that besides just `_value`, like maybe a `_time` column?

stdlib/universe/kaufmansAMA.go
```diff
@@ -0,0 +1,301 @@
...
+	t.count = 0
+
+	err := tbl.Do(func(cr flux.ColReader) error {
+		if cr.Len() == 0 || len(cr.Cols()) == 0 {
+			return nil
+		}
+
+		for j, c := range cr.Cols() {
+			if !doKAMA[j] {
+				for i := int(t.n); i < cr.Len(); i++ {
```
wolffcm: I'm not sure this will work quite right if there is more than one chunk in the table...  The first chunk will only have some rows copied to the output, but the latter chunks would need to have all their values copied.  You can find out if your on the first chunk by lookng at `t.count` I guess.

Issue #1620, opened 2019-07-29
Design how scanning string interpolation will work. 


nathanielc:
Look at other languages like Ruby and PHP for how their scanners treat tokens inside string interpolations.

jsternberg:
Something to consider is it will likely be necessary to intersperse the parser and the scanner and this should be fine and should not cause a big problem.

The way you would do this is decently simple, but it involves a lot of tokens.

You would have one token for a quote. This would signify both the beginning and the end of the string. The parser would begin reading a string expression when it sees a quote. After you see the quote, you would enter into a new lexer whose sole job is to continue scanning until it sees an end quote, the expression start symbol, or end of file. It would emit a string literal token for that.

When it sees a string expression begin token, the parser would go back to using the normal lexer and it would parse an expression. It would keep going until it saw the `}` bracket.

I don't think this should be too difficult. @jlapacik have you found anything that might indicate a better way of doing this? I can start trying to put together an example of how you would modify the lexer and a change to the EBNF for what this would look like.

PR #1673, opened 2019-08-09
docs(parser): design how string expressions will work

jsternberg:
The grammar now includes a `StringExpression`. The scanner will be
updated so that a string literal cannot contain a left bracket in the
string. The scanner will also be updated to include a double quote as a
possible token so that the parser can recognize that the quote begins a
string expression.

A string expression is then processed with an alternate scanner
implementation which recognizes a text token, which is the literal
component of the string expression, and an ability to recognize when a
left bracket occurs.

#1620

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -270,8 +270,7 @@ String literals support several escape sequences.
     \t   U+0009 horizontal tab
     \"   U+0022 double quote
     \\   U+005C backslash
-    \{   U+007B open curly bracket
-    \}   U+007D close curly bracket
+    \${  U+0024 U+007B dollar sign and opening curly bracket
 
 Additionally any byte value may be specified via a hex encoding using `\x` as the prefix.
 
@@ -281,7 +280,7 @@ Additionally any byte value may be specified via a hex encoding using `\x` as th
     hex_digit        = "0" … "9" | "A" … "F" | "a" … "f" .
     unicode_value    = unicode_char | escaped_char .
     escaped_char     = `\` ( "n" | "r" | "t" | `\` | `"` ) .
-    StringExpression = "{" Expression "}" .
+    StringExpression = "${" Expression "}" .
 
 TODO(nathanielc): With string interpolation string_lit is not longer a lexical token as part of a literal, but an entire expression in and of itself.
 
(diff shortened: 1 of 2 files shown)


docs/SPEC.md
```diff
@@ -311,9 +310,9 @@ To include the literal curly brackets within a string they must be escaped.
 Interpolation example:
 
     n = 42
-    "the answer is {n}" // the answer is 42
-    "the answer is not {n+1}" // the answer is not 43
-    "openinng curly bracket \{" // openinng curly bracket {
+    "the answer is ${n}" // the answer is 42
+    "the answer is not ${n+1}" // the answer is not 43
+    "opening curly bracket \${" // opening curly bracket {
```
nathanielc: nit, typo at the end missing `$` line should be `"opening curly bracket \${" // opening curly bracket ${`

internal/parser/grammar.md
```diff
@@ -63,10 +63,14 @@ The parser directly implements the following grammar.
...
     CallExpression                 = "(" ParameterList ")" .
     IndexExpression                = "[" Expression "]" .
+    StringExpression               = "\"" [ StringExpressionPart ] "\"" .
+    StringExpressionPart           = text | StringExpressionBlock .
+    StringExpressionBlock          = "${" Expression "}" .
     PrimaryExpression              = identifer
                                    | int_lit
                                    | float_lit
                                    | string_lit
+                                   | StringLiteralExpression
```
nathanielc: Should this say `StringExpression`?

jsternberg: Probably. I went back and forth on the name a few times.

docs/SPEC.md
```diff
@@ -311,9 +310,9 @@ To include the literal curly brackets within a string they must be escaped.
 Interpolation example:
 
     n = 42
-    "the answer is {n}" // the answer is 42
-    "the answer is not {n+1}" // the answer is not 43
-    "openinng curly bracket \{" // openinng curly bracket {
+    "the answer is ${n}" // the answer is 42
+    "the answer is not ${n+1}" // the answer is not 43
+    "opening curly bracket \${" // opening curly bracket ${
     "closing curly bracket \}" // closing curly bracket }
```
nathanielc: This line should be removed now, I missed that on the first pass.

internal/parser/grammar.md
```diff
@@ -63,10 +63,14 @@ The parser directly implements the following grammar.
...
     CallExpression                 = "(" ParameterList ")" .
     IndexExpression                = "[" Expression "]" .
+    StringExpression               = "\"" [ StringExpressionPart ] "\"" .
+    StringExpressionPart           = text | StringExpressionBlock .
+    StringExpressionBlock          = "${" Expression "}" .
     PrimaryExpression              = identifer
                                    | int_lit
                                    | float_lit
                                    | string_lit
+                                   | StringExpression
```
nathanielc: great, FWIW I like this name.

docs/SPEC.md
jsternberg: I kept it mostly because it's a demonstration of the escape sequence, but I'll update the text to make clear the dollar sign is there too.

jsternberg: Never mind you meant the closing bracket.

Issue #1399, opened 2019-06-21
parser recognizes string interpolation


aanthony1243:
syntax is already defined in spec.md: 

n = 10
r = "this will be eleven: {n +1}"
// this will be eleven: 11

wolffcm:
I'm finding it difficult to understand what the design for this will be after reading the SPEC.

For a string like `"foo {fn(n: 3)}"`, what will the scanner return when it reaches this text?  It might:
- return a string literal token like `foo {fn(n: 3)}` that the parser will deal with.  Note that it would have to pass the text `fn(n:3)` back to the (another..?) scanner in order to tokenize the text.
- the scanner might instead have a separate method called `scanStringPart` that is called after the parser encounters a `"`.  This seems doable, but would add another kind of scanning method in addition to the "normal" and "regex" tokenization that the scanner does now.

Can you nest interpolating strings?  E.g., is this valid? 
```
"here is a string: {fn(str: \"foo: {fn(n: 3)}\")}"`
```

jsternberg:
I think you have summarized the plan. It is to my impression that you can nest string expressions so I'm operating under that assumption. You can in Ruby and it would likely be more effort to avoid allowing that.

I'll be working on the scanner portion of this and then handing it off to @jlapacik who will do the parsing portion. We were going to pair on this, but he's feeling under the weather today and I am going to be out most of next week.

jsternberg:
The scanning component of this is done so I'm removing myself from this issue as @jlapacik is going to take the parsing section.

PR #1674, opened 2019-08-09
feat(scanner): implement the scanning components for string expressions

jsternberg:
The scanner adds an additional machine that will process a string
expression.

The existing string literal token has not been removed so the scanner is
backwards compatible. The only addition is that if it sees the sequence
`${` in the string, it will not parse it as a string literal since it
isn't one.

The scanner will now emit a `QUOTE` token which can be used to begin a
string expression. It is then up to the parser to start using
`ScanStringExpr` which will scan the string for three possible tokens:
`QUOTE`, `STRINGEXPR`, and `TEXT`. The `QUOTE` token signifies that an
unescaped quote was encountered and indicates the string expression is
finished. The `TEXT` token is a new token that indicates raw text. The
`STRINGEXPR` token indicates that the string expression sequence (`${`)
has been encountered and signifies that the parser should parse an
expression. The parser will then parse an expression until it reaches a
right brace (`}`) which is not included in the string expression scanner
because an unescaped right brace is allowed in a string expression
unescaped.

This also modifies the parser code that parses the output of the scanner
because the scanner will now allow an escaped dollar sign in a normal
string literal.

BREAKING CHANGE:
The sequence `${` must be escaped if it is in a string literal.

#1399

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


internal/parser/strconv.go
```diff
@@ -120,6 +120,8 @@ func writeNextUnescapedRune(s string, builder *strings.Builder) (width int, err
 			r = '\\'
 		case '"':
 			r = '"'
+		case '$':
+			r = '$'
```
nathanielc: Does this needs to check that the next rune is `{`? Otherwise it will incorrectly accept `\$` as a valid escape sequence? Or does the scanner ensure its always `${` together?

jsternberg: The scanner should ensure that it is always `\${` all together so it would be difficult to make this one multiple characters.

PR #1692, opened 2019-08-13
feat(dependencies): added a SecretService interface and a new dependencies package and a basic test of functionality

aanthony1243:
service takes a key and a context which are known at the flux level.  In influxdb or other services, we have utilities that can process the context for proper authentication.   This design prevents a dependency on an external service.  

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/dependencies/secret_service.go
+++ b/dependencies/secret_service.go
@@ -1,0 +1,11 @@
+package dependencies
+
+import (
+	"context"
+)
+
+// SecretService generalizes the process of looking up secrets based on a key.
+type SecretService interface {
+	// LoadSecret retrieves the secret value v found at key k given the calling context ctx.
+	LoadSecret(ctx context.Context, k string) (string, error)
+}
--- a/dependencies/secret_service_test.go
+++ b/dependencies/secret_service_test.go
@@ -1,0 +1,23 @@
+package dependencies_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/influxdata/flux/mock"
+)
+
+func TestSecret_Service(t *testing.T) {
+	ss := mock.SecretService{"key": "val"}
+	val, err := ss.LoadSecret(context.Background(), "key")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if val != "val" {
+		t.Error("secret service returned wrong value")
+	}
+
+	if _, err = ss.LoadSecret(context.Background(), "k"); err == nil {
+		t.Error("secret service should have errored on key lookup")
+	}
+}
--- a/mock/secret_service.go
+++ b/mock/secret_service.go
@@ -1,0 +1,17 @@
+package mock
+
+import (
+	"context"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+)
+
+type SecretService map[string]string
+
+func (s SecretService) LoadSecret(ctx context.Context, k string) (string, error) {
+	v, ok := s[k]
+	if ok {
+		return v, nil
+	}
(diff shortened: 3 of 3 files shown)


dependencies/secret_service.go
```diff
@@ -0,0 +1,10 @@
+package dependencies
+
+import (
+	"context"
+)
+
+type SecretService interface {
```
jsternberg: Can this get a godoc comment?

dependencies/secret_service.go
```diff
@@ -0,0 +1,10 @@
...
+	// LoadSecret retrieves the secret value v found at key k for organization orgID.
+	LoadSecret(ctx context.Context, k string) (string, error)
```
jsternberg: Should the organization id be included in this or are we going to rely on the context?

I think the latter is ok (to rely on the context) as we have pretty few references to organizations within flux itself and I kind of like that. I feel it's a good abstraction and the organization implementation is "context" to the query. But, if that is the case, then we shouldn't reference organizations in the doc comment as there is no way to get or set the organization from flux code.

One further comment, should this return `values.Value`? I asked about this in standup and was under the impression that secrets would be an object type. It seems to me like it would be useful if this returned a `values.Value` so it was up to the implementation what type it returned.

dependencies/secret_service_test.go
```diff
@@ -0,0 +1,24 @@
...
+	ss := mock.SecretService{}
+	val, err := ss.LoadSecret(context.Background(), "key")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if val != "val" {
+		t.Error("secret service returned wrong value")
+	}
+
+	_, err = ss.LoadSecret(context.Background(), "k")
```
jsternberg: Nitpick: When you have something like this where the value doesn't have to be used outside of the if statement, it's preferred to do this:

```
if _, err := ss.LoadSecret(context.Background(), "k"); err == nil {
    ...
}
```

dependencies/secret_service.go
nathanielc: > Should the organization id be included in this or are we going to rely on the context?

It is part of the context, the Flux side of things will not know about organizations. Agreed we should update the comments that reference them.

> One further comment, should this return values.Value? I

No, this is the lower level API for retrieving the secret value itself. The `secrets.get` Flux code will not use this API directly, rather that will be the responsibility of the `http.post` function for example to use this service.

@aanthony1243 
That brings up a question for me though: How does `secrets.get` uses this service to check for authorization? Does it just load the secret and then discard the value and check the error?

jsternberg:
Another random thought. It might be better to have this be the `dependency` package so then you write `dependency.SecretService` or `dependency.Register` or whatever functions we add to this package.

Just something to consider.

dependencies/secret_service.go
aanthony1243: @nathanielc I believe that  we can use the permissions API on the influxdb side to pre-auth without loading the secret.  But in this design I had presumed we would just pull the secret and discard it.

dependencies/secret_service.go
jsternberg: I don't think this needs to specifically mention dependency injection. I think it just needs to say "it is a service for looking up a secret based on the key."

mock/secret_service.go
```diff
@@ -0,0 +1,15 @@
+package mock
+
+import (
+	"context"
+	"errors"
+)
+
+type SecretService struct{}
```
jsternberg: This is simple enough that you could just have the mock secret service be a `map[string]string` and implement it.

PR #1693, opened 2019-08-13
feat(stdlib): add objectKeys function

nathanielc:
Adds objectKeys function to extract the keys as a list.
Fixes two bugs found when writing a test case:
    1. Binary functions are now indexed by Nature not Type.
    2. Object Equal function has been fixed.

Tests have been added.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -302,8 +302,8 @@ func compile(n semantic.Node, typeSol semantic.TypeSolution, builtIns Scope, fun
 		rt := r.Type()
 		f, err := values.LookupBinaryFunction(values.BinaryFuncSignature{
 			Operator: n.Operator,
-			Left:     lt,
-			Right:    rt,
+			Left:     lt.Nature(),
+			Right:    rt.Nature(),
 		})
 		if err != nil {
 			return nil, err
--- a/stdlib/experimental/object_keys.go
+++ b/stdlib/experimental/object_keys.go
@@ -1,0 +1,38 @@
+package experimental
+
+import (
+	flux "github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/semantic"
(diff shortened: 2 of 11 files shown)


compiler/compiler.go
```diff
@@ -302,8 +302,8 @@ func compile(n semantic.Node, typeSol semantic.TypeSolution, builtIns Scope, fun
 		rt := r.Type()
 		f, err := values.LookupBinaryFunction(values.BinaryFuncSignature{
 			Operator: n.Operator,
-			Left:     lt,
-			Right:    rt,
+			Left:     lt.Nature(),
```
wolffcm: Is the purpose of `Nature` to make it so that identical types compare as equal, even if they maybe represented as different objects?

stdlib/experimental/object_keys.go
```diff
@@ -0,0 +1,38 @@
...
+			if !ok {
+				return nil, errors.New(codes.Invalid, "missing parameter \"o\"")
+			}
+			if o.Type().Nature() != semantic.Object {
+				return nil, errors.New(codes.Invalid, "parameter \"o\" is not an object")
+			}
+			obj := o.Object()
+			keys := make([]values.Value, 0, obj.Len())
+			obj.Range(func(name string, _ values.Value) {
+				keys = append(keys, values.NewString(name))
```
wolffcm: I wonder if this will return the keys in any particular order, or if determinism might be a concern here?

nathanielc: The implementation is deterministic and uses the order they are defined. I thought that was part of the SPEC, but a second look doesn't mention it.

compiler/compiler.go
nathanielc: Yes, if it was using a `Type` then I would have to enumerate all array types in the map, there are an infinite number of them, so that is not possible.  Using `Nature` allows for enumerating a finite set of possible comparisons.

PR #1694, opened 2019-08-13
feat(parser): parse string interpolation

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


ast/ast.go
```diff
@@ -532,6 +535,56 @@ func (*StringLiteral) expression()          {}
...
+		ne.Parts = make([]*StringExpressionPart, len(e.Parts))
+		for i, p := range e.Parts {
+			ne.Parts[i] = p.Copy().(*StringExpressionPart)
+		}
+	}
+
+	return ne
+}
+
+type StringExpressionPart struct {
```
nathanielc: Should `StringExpressionPart` be an interface instead of a struct? It seems to behave like a union type.

jlapacik: I think a struct representing a string part makes more sense vs separate `Text` and `${ ... }` ast nodes. In my opinion.

jlapacik:
@nathanielc yes I agree we should update the Rust parser too. I remembered the rust parts after I put up the PR. I'll have those updates soon.

jlapacik:
Ported over string interpolation to the Rust parser as well.

ast/ast.go
jlapacik: To keep parity with the Rust implementation where `StringExpressionPart` is implemented as an `enum` type, I did indeed have to make it an interface type in the Go implementation.

PR #1698, opened 2019-08-14
feat(stdlib/json): add json.encode function

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written
- [x] Update to use bytes type once that is merged See #1696


nathanielc:
@wolffcm It has been updated to use the `bytes` type.

wolffcm:
Oops, I don't see my comment in there so I'll repeat it here.  By allowing conversion from `bytes` to `string` in Flux, it's possible to create strings that are invalid UTF8.  I wonder if we've considered guarding against this somehow?

nathanielc:
Hmm, good point, I remove conversions from bytes to string.

PR #1702, opened 2019-08-14
feat(stdlib): add experimental set function to update entire object

nathanielc:
Adds a copy of the set function that uses an object instead of just a
single key value pair.

Fixes a small ergonomic issue with generating ColMaps, by not tying them
to ColReaders and just using a bare cols list. As the Cols don't change
for a ColReaders in a single table.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/experimental/set.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func (t *setTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	key := tbl.Key()
+	updateKey := false
+	t.object.Range(func(k string, _ values.Value) {
+		updateKey = updateKey || execute.HasCol(k, key.Cols())
+	})
+	if updateKey {
```
jlapacik: Is this necessary? Would it possible for `set` to just extend and not to update silently?

stdlib/experimental/set_test.go
```diff
@@ -0,0 +1,378 @@
...
+					{Label: "t1", Type: flux.TString},
+				},
+				Data: [][]interface{}{
+					{execute.Time(1), nil, "bob"},
+					{execute.Time(2), 1.0, "bob"},
+				},
+			}},
+		},
+		{
+			name: "replace col",
```
jlapacik: Perhaps it is a requirement that `set` has both update and extend semantics? But if not, I would like for it to just have extend semantics. That is, I would like for `set` to just extend its input with new columns and not update existing columns.

The reason for this is that we can't infer a principal type for an object when a single operation might update an existing field on an object or extend that object with a new field. These operations have to be distinct if we want to be able to perform inference on them.

I know the current (non-experimental) `set` function does not do this either, but my understanding is that at some point this `set` function will replace our current one, and so I would like the new version to be able to assign the new one a valid type.

stdlib/experimental/set.go
nathanielc: I am not sure I follow, if the set object is going to overwrite existing columns in the group key then it  needs to update the key.

nathanielc:
In the extend only version of set does it error of the column already exists?



> Also I was thinking it would be nice to put end to end tests for experimental functions inside the experimental package itself. What do you think?

I think we should move all tests out of the `testing/testdata` package and place them in their respective packages. 

I'll move this one to start the trend.

jlapacik:
Yes it would be an error to extend tables with a column that already exists. Ultimately though it would/should be a compile error.

stdlib/experimental/set.go
jlapacik: I'm saying that we should have a separate function, call it `update` maybe, for explicitly updating columns that already exist on the input tables. Then `set` would only be used to add columns that don't exist on the input tables.

nathanielc:
@jlapacik  Ok tests have been moved using the new conventions.

Issue #1668, opened 2019-08-08
Create an experimental to() function


wolffcm:
Create an `experimental.to()` function.
It should:
- Expect "pivoted" data. I.e., expect a column called `usage_idle` instead of two columns `_field` (set to `usage_idle`) and `_value`
- Take any column that is in the group key and treat it as a tag in storage. Non-string columns in group key will produce an error.
- Any other column should be treated as a field.  Column types unsupported by InfluxDB (like `time`) should produce an error.

DOD:
- A new `to` in the package `experimental` exists
- It has tests to verify its behavior

jsternberg:
Should this drop `_start` and `_stop` by default or rely on the user to do that?

wolffcm:
My vote would be for it to drop `_start` and `_stop` automatically.

PR #1703, opened 2019-08-14
feat(stdlib): add an experimental "to" function

wolffcm:
#1668


wolffcm:
This is showing code coverage issues, but much of it is in changes I made in the `executetest` package that are used by a pipeline test in idpe.  Not sure why it's saying `mode.go` coverage went down.

Issue #1699, opened 2019-08-14
Implement notify function


nathanielc:
Implement `notify` as defined here https://github.com/influxdata/flux/blob/master/docs/Alerts.md#notify

jsternberg:
Are we going to create a separate issue for implementing an endpoint such as a log endpoint that emits the results to a logger? Primarily for testing since we likely don't really care for an endpoint that writes to the log for the actual server.

PR #1705, opened 2019-08-15
feat(stdlib): add the notify function to the influxdb alerts library

jsternberg:
#1699

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/alerts/alerts.flux
+++ b/stdlib/influxdata/influxdb/alerts/alerts.flux
@@ -1,5 +1,6 @@
 package alerts
 
+import "experimental"
 import "influxdata/influxdb/v1"
 import "influxdata/influxdb"
 
@@ -16,7 +17,16 @@ from = (start, stop=now(), fn) =>
         |> v1.fieldsAsCols()
 
 // Log records notification events
-log = (tables=<-) => tables |> to(bucket: bucket)
+log = (tables=<-) => tables |> experimental.to(bucket: bucket)
+
+// Notify will call the endpoint and log the results.
+notify = (tables=<-, endpoint, data={}) =>
+    tables
+        |> experimental.set(o: data)
+        |> experimental.group(mode: "extend", columns: experimental.objectKeys(o: data))
+        |> duplicate(column: "_time", as: "_status_timestamp")
+        |> endpoint()
+        |> log()
 
 // Logs retrieves notification events that have been logged.
 logs = (start, stop=now(), fn) =>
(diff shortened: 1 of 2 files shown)


stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -18,6 +19,14 @@ from = (start, stop=now(), fn) =>
 // Log records notification events
 log = (tables=<-) => tables |> to(bucket: bucket)
 
+// Notify will call the endpoint and log the results.
+notify = (tables=<-, endpoint, data={}) =>
+    tables
+        |> experimental.set(o: data)
+        |> experimental.group(mode: "extend", columns: experimental.objectKeys(o: data))
+        |> endpoint()
+        |> log()
```
nathanielc: We need to add some columns here:

* `_sent` - this one of `true` or `false` as a string (it needs to be a tag). Maybe the endpoint needs to do this part? 
* `_status_timestamp` - this should be the `_time` of the original data
* `_time` - this should be updated to be `now()`. I.e. record the time the notification ran.
* `_error` - this should be the error message if any. must be set when `_sent` is false. Same question here maybe the endpoint needs to do this part.

jsternberg: There's a similar value for all of those that show up in the original spec and they are all properties of the endpoint. While the spec needs to be updated to the rfc, I think it makes sense that those properties would all be part of the endpoint.

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -18,6 +19,14 @@ from = (start, stop=now(), fn) =>
 // Log records notification events
 log = (tables=<-) => tables |> to(bucket: bucket)
```
nathanielc: This should use `experimental.to` correct? Since the data is in pivoted form?

stdlib/influxdata/influxdb/alerts/alerts.flux
nathanielc: The timestamps should be handled as part of `notify` though as that is not something the endpoint needs to be aware of. So that means `_sent` and `_error` should be handled by the endpoint and `_time` and `_status_timestamp` should be handled in `notifty`. Does that sound good?

stdlib/influxdata/influxdb/alerts/alerts.flux
jsternberg: Yes, that is likely the case. I'll update this.

stdlib/influxdata/influxdb/alerts/alerts.flux
jsternberg: `_time` should definitely be handled by the endpoint as it is the time that the endpoint sent the message. It's not really related to `notify()` IMO because notify isn't actually sending anything.

We could potentially duplicate the `_status_timestamp` column though in notify so that `endpoint()` only has to worry about adding `_time`. I think that would make sense.

PR #1706, opened 2019-08-15
refactor(stdlib): move test files alongside package files

nathanielc:
This change makes it so that test files can be placed inside the same
Go package as the code they are testing.
Test files must use the `_test.flux` suffix and declare a `*_test`
package name.
A README has been added explaining the convention.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
Eventually I would like to see multiple Flux test be specified in a single file instead of having a file per test case. Then use the name of the test case itself (i.e. `test <name> = ... `) as the Go test name instead of the file name.

I would also like the Flux tests to be part of the Go tests for the same package. As it stands now the Flux tests all belong to the `stdlib` Go package. 

Both of these changes can come later, but should help organize our Flux stdlib in a more encapsulated way.

nathanielc:
Another note, I decided to leave the `testing/testdata` package as a standalone package. Some but not all of the test cases apply to the universe package and so could be relocated. That refactor can be separate and come later when its more valuable.

PR #1718, opened 2019-08-16
feat(stdlib): add an alerts.check function

wolffcm:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/alerts/alerts.flux
+++ b/stdlib/influxdata/influxdb/alerts/alerts.flux
@@ -6,8 +6,8 @@ import "influxdata/influxdb"
 
 bucket = "_monitoring"
 
-// Write persists the check statuses
-write = (tables=<-) => tables |> influxdb.to(bucket: bucket)
+// write optionally persists the check statuses
+option write = (tables=<-) => tables |> drop(columns: ["_start", "_stop"])
 
 // From retrieves the check statuses that have been stored.
 from = (start, stop=now(), fn) =>
@@ -41,3 +41,45 @@ logs = (start, stop=now(), fn) =>
 deadman = (t, tables=<-) => tables
     |> max(column: "_time")
     |> map(fn: (r) => ( {r with dead: r._time < t} ))
+
+// levels describing the result of a check
+levelOK = "ok"
+levelInfo = "info"
+levelWarn = "warn"
+levelCrit = "crit"
+levelUnknown = "unknown"
+
+// Check performs a check against its input using the given ok, info, warn and crit functions
+// and writes the result to a system bucket.
+check = (
+    tables=<-,
+    data={},
+    messageFn,
+    crit=(r) => false,
+    warn=(r) => false,
+    info=(r) => false,
+    ok=(r) => true
+) =>
+    tables
+        |> experimental.set(o: data.tags)
+        |> experimental.group(mode: "extend", columns: experimental.objectKeys(o: data.tags))
+        |> map(fn: (r) => ({r with
+            _measurement: "statuses",
+            _source_measurement: r._measurement,
+            _type: data._type,
+            _check_id:  data._check_id,
+            _check_name: data._check_name,
+            _level:
+                if crit(r: r) then levelCrit
+                else if warn(r: r) then levelWarn
+                else if info(r: r) then levelInfo
+                else if ok(r: r) then levelOK
+                else levelUnknown,
+            _source_timestamp: r._time,
+            _time: now(),
+        }))
+        |> map(fn: (r) => ({r with
+            _message: messageFn(r: r),
+        }))
+        |> experimental.group(mode: "extend", columns: ["_source_measurement", "_type", "_check_id", "_check_name", "_level"])
+        |> write()
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -255,10 +255,10 @@ func (e *conditionalEvaluator) Eval(scope Scope) (values.Value, error) {
 		return nil, err
 	}
 
-	if t.Bool() {
-		return eval(e.consequent, scope)
-	} else {
+	if t.IsNull() || !t.Bool() {
 		return eval(e.alternate, scope)
+	} else {
+		return eval(e.consequent, scope)
 	}
 }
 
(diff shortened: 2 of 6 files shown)


stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,28 @@ logs = (start, stop=now(), fn) =>
...
+levelInfo = 1
+levelWarn = 2
+levelCrit = 3
+levelUnknown = -1
+
+// Check performs a check against its input using the given ok, info, warn and crit functions
+// and writes the result to a system bucket.
+check = (tables=<-, check, messageFn, crit, warn, info, ok) =>
+    tables
+        |> experimental.set(o: check.tags)
```
jsternberg: Is there a reason we don't just add an extra option to `set` so that it sets them as tags? Either of these two:

```
// an optional parameter to say that the data is going to be added to the group key
experimental.set(data: check.tags, key: true)

// use data to set normal data and tags to set tags
experimental.set(tags: check.tags)
```

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,29 @@ logs = (start, stop=now(), fn) =>
...
+// levels describing the result of a check
+levelOK = 0
+levelInfo = 1
+levelWarn = 2
+levelCrit = 3
+levelUnknown = -1
+
+// Check performs a check against its input using the given ok, info, warn and crit functions
+// and writes the result to a system bucket.
+check = (tables=<-, data, messageFn, crit, warn, info, ok) =>
```
jsternberg: `data` should be optional I think. So it would default to `data={}` if that's possible.

stdlib/influxdata/influxdb/alerts/alerts.flux
wolffcm: Personally I like having smaller composable operations, but since adding such an option would never make it so that we need to regroup, I'd be okay with it.

stdlib/influxdata/influxdb/alerts/alerts.flux
wolffcm: If it's optional then the `_check_id` would be null, and then writing it to storage would... just omit the tag?  actually I'm not sure what would happen.

stdlib/influxdata/influxdb/alerts/alerts.flux
jsternberg: That's mostly my perspective on it. The primary reason is performance, but also just that it requires less code to get the desired effect.

stdlib/influxdata/influxdb/alerts/alerts.flux
jsternberg: I figured these were more generic functions and that the actual influxdb implementation would invoke the function correctly.

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,29 @@ logs = (start, stop=now(), fn) =>
 deadman = (t, tables=<-) => tables
     |> max(column: "_time")
     |> map(fn: (r) => ( {r with dead: r._time < t} ))
+
+// levels describing the result of a check
+levelOK = 0
+levelInfo = 1
+levelWarn = 2
+levelCrit = 3
+levelUnknown = -1
```
nathanielc: These need to be strings, it was decided we want users to be able to search for status using the string name of the level.

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,29 @@ logs = (start, stop=now(), fn) =>
...
+        |> experimental.set(o: data.tags)
+        |> experimental.group(mode: "extend", columns: experimental.objectKeys(o: data.tags))
+        |> map(fn: (r) => ({r with
+            _check_id:  data._check_id,
+            _level:
+                if crit(r: r) then levelCrit
+                else if warn(r: r) then levelWarn
+                else if info(r: r) then levelInfo
+                else if ok(r: r) then levelOK
+                else levelUnknown
```
nathanielc: There are a few missing columns:

* `_check_name` - this should come from the check data object
* `_type` -  this should come from the check data object as well
* `_source_timestamp` - this should be the original `_time` value.
* `_time` - this should be the `now` time of the current execution. i.e. record the status for the current time.
* `_source_measurement` - this needs to be the original `_measurement` tag
* `_measurement` - needs to be redefined statically as `statuses`

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,29 @@ logs = (start, stop=now(), fn) =>
...
+        }))
+        |> map(fn: (r) => ({r with _message: messageFn(r: r, data: data)}))
+        |> experimental.group(mode: "extend", columns: ["_check_id", "_level"])
```
nathanielc: This looks good, of the above we need to add `_check_name` `_type` and `_source_timestamp`

stdlib/influxdata/influxdb/alerts/alerts.flux
```diff
@@ -31,3 +32,29 @@ logs = (start, stop=now(), fn) =>
...
+        |> map(fn: (r) => ({r with
+            _check_id:  data._check_id,
+            _level:
+                if crit(r: r) then levelCrit
+                else if warn(r: r) then levelWarn
+                else if info(r: r) then levelInfo
+                else if ok(r: r) then levelOK
+                else levelUnknown
+        }))
+        |> map(fn: (r) => ({r with _message: messageFn(r: r, data: data)}))
```
nathanielc: Is all of the `data` added to the table? If so then I think we should just pass the record `r`. Otherwise there are two ways to get at the same data.

stdlib/influxdata/influxdb/alerts/alerts.flux
wolffcm: Won't `_source_timestamp` create high cardinality?  Maybe you mean `_source_measurement`?

stdlib/influxdata/influxdb/alerts/alerts.flux
wolffcm: Ok, it makes sense.  I made a second call to `map` to create the message so that the `messageFn` can look at `r._level` and create an appropriate message.

There could potentially be other fields on `data` that the Flux library doesn't need to know about, but the user might use in `messageFn` though?  I could go either way on this one.

wolffcm:
I've updated this to reflect the comments.

 I had to skip over the test because of the option/scoping issue: https://github.com/influxdata/flux/issues/1720

PR #1719, opened 2019-08-16
fix(compiler): calling polymorphic values

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -97,7 +97,7 @@ func eval(e Evaluator, scope Scope) (values.Value, error) {
 	if err != nil {
 		return nil, err
 	}
-	//values.CheckKind(v.Type().Nature(), e.Type().Nature())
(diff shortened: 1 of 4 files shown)


compiler/runtime.go
```diff
@@ -471,8 +479,7 @@ func (e *memberEvaluator) Eval(scope Scope) (values.Value, error) {
 		return nil, err
 	}
 	v, _ := o.Object().Get(e.property)
-	// TODO(nathanielc): Fix this poly type issue
-	//values.CheckKind(v.PolyType().Nature(), e.t.Nature())
+	values.CheckKind(v.Type().Nature(), e.t.Nature())
```
nathanielc: Does member expression need to use the poly Type as well? If not why? The issue I ran into was `http.post` which was a polymorphic member expression.

PR #1725, opened 2019-08-19
tests(stdlib): ensure that package functions can be used in the compiler

jsternberg:
This changes the compiler so that it will not fail if it cannot resolve
a function since it is not an error for a function to not be
resolvable in the compiler.

#1509

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -827,7 +827,12 @@ func (f *function) resolveIdentifiers(n semantic.Node) (semantic.Node, error) {
 
 		v, ok := f.scope.Lookup(n.Name)
 		if ok {
-			return resolveValue(v)
+			// Attempt to resolve the value if it is possible to inline.
+			node, ok, err := resolveValue(v)
+			if !ok {
+				return n, nil
+			}
+			return node, err
 		}
 		return nil, fmt.Errorf("name %q does not exist in scope", n.Name)
 	case *semantic.Block:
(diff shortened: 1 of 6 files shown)


interpreter/interpreter.go
```diff
@@ -1011,16 +1017,17 @@ func resolveValue(v values.Value) (semantic.Node, error) {
 				return
 			}
 			var n semantic.Node
-			n, err = resolveValue(el)
+			n, ok, err := resolveValue(el)
 			if err != nil {
 				return
+			} else if ok {
+				node.Elements[i] = n.(semantic.Expression)
 			}
```
nathanielc: I believe you need an else case here where `ok` is false and `err == nil`. The case should set a flag to false that the array can be in-lined. If any value of the array cannot be in-lined then the entire array cannot be in-lined.

interpreter/interpreter.go
```diff
@@ -1031,21 +1038,22 @@ func resolveValue(v values.Value) (semantic.Node, error) {
...
-			n, err = resolveValue(v)
+			n, ok, err := resolveValue(v)
 			if err != nil {
 				return
+			} else if ok {
+				node.Properties = append(node.Properties, &semantic.Property{
+					Key:   &semantic.Identifier{Name: k},
+					Value: n.(semantic.Expression),
+				})
 			}
```
nathanielc: Similar case is needed here, if any property cannot be in-lined then the entire object cannot be in-lined.

interpreter/interpreter.go
jsternberg: I'm not sure this is correct. If we see an array, then to my knowledge, it is being inlined. It's just one of the elements that can't be. So if you have something like:

```
msg = "message"
a = () => "foobar"
b = [msg, a()]
```

(Pretend that `a` cannot be resolved, even though in this case it is I think)

This would cause the `b` to be rewritten as `["foobar", a()]` which is still valid.

nathanielc: The above code doesn't do that, it never adds anything to the array for values that cannot be resolved so you get `["foobar", nil]` with the current code.

jsternberg: My mistake. I'll fix it.

PR #1729, opened 2019-08-19
feat(stdlib/http): add http endpoint

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/http/http.flux
```diff
@@ -1,5 +1,19 @@
...
 // Post submits an HTTP post request to the specified URL with headers and data.
 // The HTTP status code is returned.
 builtin post
+
+endpoint =  (url) =>
+    (mapFn) =>
+        (tables=<-) =>
+            tables
+                |> map(fn: (r) => {
+                    obj = mapFn(r)
```
jlapacik: should be `mapFn(r: r)`.

nathanielc: Technically its valid since its the shortcut syntax of calling a function with name args if the identifier is the name of the arg. But I agree, I'd like the stdlib to be more explicit. I'll change it.

nathanielc:
FWIW, I tested the `http_endpoint` test case locally using an http server on localhost and it works. So I know the code all works, its just we don't have the ability to test as part of the e2e test cases.

stdlib/http/http.flux
jlapacik: Oh I didn't realize functions supported the shorthand syntax. In that case no need to change.

PR #1735, opened 2019-08-20
feat: simple heat index

davidgs:
Co-authored with @anaisdg

### Done checklist
- [ X] docs/SPEC.md updated
- [ X] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -270,7 +270,8 @@ String literals support several escape sequences.
     \t   U+0009 horizontal tab
     \"   U+0022 double quote
     \\   U+005C backslash
-    \${  U+0024 U+007B dollar sign and opening curly bracket
+    \{   U+007B open curly bracket
+    \}   U+007D close curly bracket
 
 Additionally any byte value may be specified via a hex encoding using `\x` as the prefix.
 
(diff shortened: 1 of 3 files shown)


docs/SPEC.md
```diff
@@ -3780,57 +3631,7 @@ from(bucket: "waterhouse/autogen")
...
- |  0028 |   2  | -100 |
- |  0029 |   1  | -100 |
- 
- Example of script:
- ```
- from(bucket: "telegraf/autogen"):
-     |> range(start: -7d)
-     |> chandeMomentumOscillator(n: 10, columns: ["_value"])
- ```
-
```
nathanielc: Looks like your changes are based off an old version, can you rebase your work off latest master?

stdlib/universe/universe.flux
```diff
@@ -295,20 +287,19 @@ tripleEMA = (n, tables=<-) =>
...
-
 toString   = (tables=<-) => tables |> map(fn:(r) => ({r with _value: string(v:r._value)}))
 toInt      = (tables=<-) => tables |> map(fn:(r) => ({r with _value: int(v:r._value)}))
 toUInt     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: uint(v:r._value)}))
 toFloat    = (tables=<-) => tables |> map(fn:(r) => ({r with _value: float(v:r._value)}))
 toBool     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: bool(v:r._value)}))
 toTime     = (tables=<-) => tables |> map(fn:(r) => ({r with _value: time(v:r._value)}))
+
+
+simpleHeatIndex = (tables=<-) => tables
```
nathanielc: Does something like this belong in a package? Maybe a `physics` package or something? Not sure what a good name is.

davidgs: I've actually added a section to the SPEC.md called **Physical Environment** and placed this under it. I was about to also add `idealGasLaw` to the same section, with the accompanying flux function. At some point I'll also add the *complete* heat index calculation uncluding the Rothfusz regression but that will take more time.

docs/SPEC.md
davidgs: Actually, I was looking at how to add the implementation including the regression, and it's so much more complicated that I think I should pull this, and submit a go pkg with both of them in it.

Issue #1623, opened 2019-07-30
Replace usage of github.com/pkg/errors with the flux error wrapping


jsternberg:
The errors that are wrapped with `github.com/pkg/errors` end up losing the flux error code because the error is not wrapped in our native error wrapper.

This issue is just to go through the code and make changes everywhere that uses `errors.Wrap` to use the `errors.Wrap` from the `github.com/influxdata/flux/internal/errors` library instead. A code can be chosen if it makes sense or `codes.Inherit` can be used in most situations.

PR #1737, opened 2019-08-20
refactor: wrap errors with flux.Error instead of github.com/pkg/errors

jsternberg:
This removes the `github.com/pkg/errors` dependency completely and
changes all usages to either directly call `errors.New` from the errors
package or `errors.Wrap` with the inherit code from our own wrapping
library.

This will allow us to later on start annotating errors so that not
everything shows up as an internal error and using our own wrapping will
ensure that actually gets transmitted to the user.

Previously, the `github.com/pkg/errors` wrapping would cause an error in
the wrapping chain because the code that identifies the error code would
not recognize the error as a flux error because it was wrapped by the
errors package.

#1623

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/ast/edit/option_editor_test.go
+++ b/ast/edit/option_editor_test.go
@@ -7,7 +7,6 @@ import (
 	"github.com/influxdata/flux/ast/edit"
 	"github.com/influxdata/flux/parser"
 	"github.com/influxdata/flux/values"
-	"github.com/pkg/errors"
 )
 
 func TestEditor(t *testing.T) {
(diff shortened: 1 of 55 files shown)


jsternberg:
There is a change in the `go.mod` with the line removed.

The `// indirect` is only used if we have all dependencies that use the package are not modules themselves which means that we have to set the dependency because they aren't modules and aren't setting it themselves.

I think what you might mean is that `go.sum` should have changed. The reason it didn't is because one of our dependencies likely uses the package and `go.sum` is a recursive flat listing of every dependency that is referenced.

PR #1740, opened 2019-08-20
refactor(interpreter): Threaded external dependencies map from the ex…

aanthony1243:
…ecution engine down through the interpreter so that runtime functions can use them.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


compiler/runtime.go
```diff
@@ -1,11 +1,14 @@
 package compiler
 
 import (
+	context2 "context"
```
sauren-khosla: change to "context", same with "dependencies" just below

examples/library/library_example_test.go
```diff
@@ -36,6 +38,10 @@ g.from(start: 1993-02-16T00:00:00Z, stop: 1993-02-16T00:03:00Z, count: 5, fn: (n
 	}
 
 	alloc := &memory.Allocator{}
+
+	if p, ok := program.(lang.DependenciesAwareProgram); ok {
+		p.SetExecutorDependencies(execute.Dependencies{"interpreter": dependencies.NewDefaultDependencies()})
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies() also

execute/executor_test.go
```diff
@@ -742,7 +743,7 @@ func TestExecutor_Execute(t *testing.T) {
 			// Construct physical query plan
 			plan := plantest.CreatePlanSpec(tc.spec)
 
-			exe := execute.NewExecutor(nil, zaptest.NewLogger(t))
+			exe := execute.NewExecutor(execute.Dependencies{"interpreter": dependencies.NewDefaultDependencies()}, zaptest.NewLogger(t))
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies() also

internal/cmd/refactortests/cmd/refactortests.go
```diff
@@ -177,6 +181,9 @@ func executeScript(pkg *ast.Package) (string, string, error) {
 	}
 
 	program, err := c.Compile(context.Background())
+	if p, ok := program.(lang.DependenciesAwareProgram); ok {
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

internal/spec/build.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/influxdata/flux/dependencies"
```
sauren-khosla: fix imports here

internal/spec/build.go
```diff
@@ -27,14 +28,14 @@ func FromScript(ctx context.Context, script string, now time.Time) (*flux.Spec,
 		return nil, err
 	}
 	s.Finish()
-	return FromAST(ctx, astPkg, now)
+	return FromAST(ctx, nil, astPkg, now)
```
sauren-khosla: verify we don't need to pass `deps` through here

interpreter/interpreter.go
```diff
@@ -1,8 +1,10 @@
 package interpreter
 
 import (
+	context2 "context"
```
sauren-khosla: fix `context2` and `dependencies2`

lang/compiler.go
```diff
@@ -2,6 +2,7 @@ package lang
 
 import (
 	"context"
+	"github.com/influxdata/flux/dependencies"
```
sauren-khosla: fix imports

lang/compiler_test.go
```diff
@@ -144,6 +145,10 @@ func TestFluxCompiler(t *testing.T) {
 				t.Errorf(`unexpected value for now, want "%v", got "%v"`, tc.now, astProg.Now)
 			}
 
+			if p, ok := program.(lang.DependenciesAwareProgram); ok {
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies() as well as below

lang/query_test.go
```diff
@@ -16,6 +20,8 @@ func runQuery(script string) (flux.Query, error) {
 	if err != nil {
 		return nil, err
 	}
+	program.SetExecutorDependencies(execute.Dependencies{"interpreter": dependencies.NewDefaultDependencies()})
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

querytest/execute.go
```diff
@@ -5,13 +5,19 @@ import (
...
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/lang"
 	"github.com/influxdata/flux/memory"
 )
 
 type Querier struct{}
 
 func (q *Querier) Query(ctx context.Context, w io.Writer, c flux.Compiler, d flux.Dialect) (int64, error) {
 	program, err := c.Compile(ctx)
+	if p, ok := program.(lang.DependenciesAwareProgram); ok {
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/flux_test.go
```diff
@@ -99,9 +100,13 @@ func testFlux(t testing.TB, pkg *ast.Package) {
 
 func doTestRun(t testing.TB, c flux.Compiler) {
 	program, err := c.Compile(context.Background())
+	if p, ok := program.(lang.DependenciesAwareProgram); ok {
+		p.SetExecutorDependencies(execute.Dependencies{"interpreter": dependencies.NewDefaultDependencies()})
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/universe/contains_test.go
```diff
@@ -1,6 +1,8 @@
 package universe_test
 
 import (
+	"context"
```
sauren-khosla: fix imports

stdlib/universe/filter.go
```diff
@@ -105,7 +107,7 @@ func createFilterTransformation(id execute.DatasetID, mode execute.AccumulationM
 	}
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
-	t, err := NewFilterTransformation(d, cache, s)
+	t, err := NewFilterTransformation(d, cache, s, a.Context(), a.Dependencies()["interpreter"].(dependencies.Interface))
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/universe/map.go
```diff
@@ -114,7 +115,8 @@ func createMapTransformation(id execute.DatasetID, mode execute.AccumulationMode
 	}
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
-	t, err := NewMapTransformation(d, cache, s)
+	t, err := NewMapTransformation(d, cache, s, a.Context(), a.Dependencies()["interpreter"].(dependencies.Interface))
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/universe/reduce.go
```diff
@@ -134,7 +136,7 @@ func createReduceTransformation(id execute.DatasetID, mode execute.AccumulationM
 	}
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
-	t, err := NewReduceTransformation(d, cache, s)
+	t, err := NewReduceTransformation(d, cache, s, a.Context(), a.Dependencies()["interpreter"].(dependencies.Interface))
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/universe/state_tracking.go
```diff
@@ -154,7 +156,7 @@ func createStateTrackingTransformation(id execute.DatasetID, mode execute.Accumu
 	}
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
-	t, err := NewStateTrackingTransformation(d, cache, s)
+	t, err := NewStateTrackingTransformation(d, cache, s, a.Context(), a.Dependencies()["interpreter"].(dependencies.Interface))
```
sauren-khosla: we should be able to change this to executetest.NewTestExecuteDependencies()

stdlib/universe/table_fns_test.go
```diff
@@ -117,11 +119,12 @@ func mustLookup(s values.Scope, valueID string) values.Value {
 func evalOrFail(t *testing.T, script string, mutator flux.ScopeMutator) values.Scope {
 	t.Helper()
 
-	_, s, err := flux.Eval(script, func(s values.Scope) {
-		if mutator != nil {
-			mutator(s)
-		}
-	})
+	_, s, err := flux.Eval(script, context.TODO(), dependencies.NewDefaultDependencies(),
+		func(s values.Scope) {
```
sauren-khosla: must've accidentally took part of Nathaniel's big merge when rebasing here?

stdlib/universe/typeconv_test.go
```diff
@@ -1,7 +1,9 @@
 package universe
 
 import (
+	"context"
```
sauren-khosla: fix imports

cmd/flux/cmd/repl.go
```diff
@@ -29,6 +32,9 @@ type querier struct{}
 
 func (querier) Query(ctx context.Context, c flux.Compiler) (flux.ResultIterator, error) {
 	program, err := c.Compile(ctx)
+	if p, ok := program.(lang.DependenciesAwareProgram); ok {
+		p.SetExecutorDependencies(fexecute.Dependencies{"interpreter": dependencies.NewDefaultDependencies()})
```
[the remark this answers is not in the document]
aanthony1243: this one is not a test.  but there are many such cases where we should make that change.

stdlib/http/post_test.go
```diff
@@ -52,7 +54,7 @@ status = http.post(url:"%s/path/a/b/c", headers: {x:"a",y:"b",z:"c"}, data: byte
 status == 204 or fail()
 `, ts.URL)
 
-	if _, _, err := flux.Eval(script, addFail); err != nil {
+	if _, _, err := flux.Eval(script, context.TODO(), dependencies.NewDefaultDependencies(), addFail); err != nil {
```
aanthony1243: FYI this one cannot be the test dependencies ATM because it expects a different error code.  This one expects a regular client that will talk to a mock test server, rather than our test dependency setup that uses a mock client.

stdlib/universe/map.go
aanthony1243: no, this isn't test code.

stdlib/universe/reduce.go
aanthony1243: same here, not test code.

stdlib/universe/state_tracking.go
aanthony1243: not a test.

stdlib/universe/table_fns_test.go
aanthony1243: that was intentional.  All of this rebased off the latest master which includes the recent scope refactor.  

it's more obvious here because I changed the formatting around a bit.

compile.go
```diff
@@ -299,7 +301,7 @@ func evalBuiltInPackages() error {
 			return errors.Wrapf(err, codes.Inherit, "package has invalid builtins %q", astPkg.Path)
 		}
 
-		itrp := interpreter.NewInterpreter()
+		itrp := interpreter.NewInterpreter(context.TODO(), dependencies.NewDefaultDependencies())
```
nathanielc: This can safely be `context.Background()`.

How will `NewDefaultDependencies` work? Does every dependency have to have a default?

What about a pattern where a `ErrorDependencies` is used for this case? My thinking is that if any of the builtins try and access the deps during startup they will not be available and we should error.

During `evalBuiltInPackages` no deps can/should be used. So concretely `http.post` will have a dep on a `SecretLookup` interface, but it doesn't use that dep unless `http.post` is called during startup which it should not be. 

Then on subsequent calls to `interpreter.Eval` new (fully initialized deps) can be provided and calling `http.post` will use those deps.

compiler/runtime.go
```diff
@@ -541,7 +546,7 @@ func (e *callEvaluator) Eval(scope Scope) (values.Value, error) {
 	if err != nil {
 		return nil, err
 	}
-	return f.Function().Call(args.Object())
+	return f.Function().Call(e.ctx, e.deps, args.Object())
```
nathanielc: Should the context and deps come from `Eval` here instead of being compiled in?

In other words, should we revert `complier/Compile` to not know about context or dep but instead teach the `runtime/Eval` interface about them? This way we do not have to worry about caching bad context or deps as they are passed in with each run of the runtime.

execute/row_fn.go
```diff
@@ -22,15 +24,20 @@ type dynamicFn struct {
...
-func newDynamicFn(fn *semantic.FunctionExpression, scope compiler.Scope) dynamicFn {
+func newDynamicFn(fn *semantic.FunctionExpression, scope compiler.Scope, ctx context.Context, deps dependencies.Interface) dynamicFn {
 	return dynamicFn{
 		compilationCache: compiler.NewCompilationCache(fn, scope),
 		inRecord:         values.NewObject(),
 		recordName:       fn.Block.Parameters.List[0].Key.Name,
 		references:       findColReferences(fn),
 		recordCols:       make(map[string]int),
+		deps:             deps,
+		ctx:              ctx,
```
nathanielc: Similar question here, should we move the context and deps off the struct but rather pass them in via `Eval` each time?

In general its best practice to not store a context on the heap (i.e. a struct). This is because by definition context carries only information about the current call stack so storing it on the heap instead of the stack creates opportunity for bugs. For example  if context lives on the heap beyond the stack lifetime how do you ensure it gets cleaned up? If you always keep it on the stack (i.e. a function parameter) you avoid this whole class of bugs.

execute/row_fn_test.go
```diff
@@ -94,6 +96,8 @@ func TestRowMapFn_Eval(t *testing.T) {
 						},
 					},
 					prelude(),
+					context.TODO(),
```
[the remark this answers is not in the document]
nathanielc: For test cases Background makes sense unless `*testing.T` has a context?

interpreter/interpreter.go
```diff
@@ -17,12 +19,16 @@ type Interpreter struct {
 	polyTypes   map[semantic.Node]semantic.PolyType
 	sideEffects []SideEffect // a list of the side effects occurred during the last call to `Eval`.
 	pkg         string
+	deps        dependencies2.Interface
+	ctx         context2.Context
 }
 
-func NewInterpreter() *Interpreter {
+func NewInterpreter(ctx context2.Context, deps dependencies2.Interface) *Interpreter {
```
nathanielc: Similar suggestion here to not store the context but rather to pass it in for calls to `Eval`.

compile.go
sauren-khosla: Note: made the change here to `context.Background()` but still need to have discussion around the rest of this comment.

compiler/compiler.go
```diff
@@ -416,6 +416,7 @@ func CompileFnParam(fn *semantic.FunctionExpression, scope Scope, paramType, ret
 	}
 	paramName := fn.Block.Parameters.List[0].Key.Name
 
+	// TODO(adam): thread a context and dependency through to here from the transformations that call them.
```
nathanielc: is this still relevant?

compiler/compiler.go
```diff
@@ -70,12 +70,12 @@ func polyType(t semantic.PolyType, err error) semantic.PolyType {
 }
 
 // compile recursively compiles semantic nodes into evaluators.
-func compile(n semantic.Node, typeSol semantic.TypeSolution, scope Scope, funcExprs map[string]*semantic.FunctionExpression) (Evaluator, error) {
+func compile(scope Scope, funcExprs map[string]*semantic.FunctionExpression, n semantic.Node, typeSol semantic.TypeSolution) (Evaluator, error) {
```
nathanielc: Should these changes just be reverted since they just change the order of parameters?

execute/executetest/dependencies.go
```diff
@@ -0,0 +1,51 @@
...
+
+type defaultDependencies struct {
+	httpclient *http.Client
+}
+
+func (d defaultDependencies) HTTPClient() (*http.Client, error) {
+	return d.httpclient, nil
+}
+
+func NewTestDependenciesInterface() dependencies.Interface {
```
nathanielc: Why is this in `executetest` instead of a `dependenciestest` package?

dependencies/dependencies.go
```diff
@@ -0,0 +1,46 @@
...
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			MaxIdleConnsPerHost:   runtime.GOMAXPROCS(0) + 1,
```
nathanielc: Where do these values come from? Should we explain some of them?

internal/spec/build.go
nathanielc: We do need the deps here, simplest way is to require them to be passed in.

lang/compiler.go
```diff
@@ -132,7 +133,7 @@ func CompileTableObject(to *flux.TableObject, now time.Time, opts ...CompileOpti
 
 func WalkIR(astPkg *ast.Package, f func(o *flux.Operation) error) error {
 
-	if spec, err := spec.FromAST(context.Background(), astPkg, time.Now()); err != nil {
+	if spec, err := spec.FromAST(context.Background(), nil, astPkg, time.Now()); err != nil {
```
nathanielc: We we need deps here? How is WalkIR used? I am not familiar with it. Can you add a doc comment to the function once you figure it out?

lang/compiler.go
```diff
@@ -313,7 +314,13 @@ func (p *AstProgram) Start(ctx context.Context, alloc *memory.Allocator) (flux.Q
 		astPkg.Files = append([]*ast.File{p.opts.extern}, astPkg.Files...)
 	}
 
-	s, err := spec.FromAST(ctx, astPkg, p.Now)
+	deps, ok := p.Dependencies[dependencies.InterpreterDepsKey]
+	if !ok {
+		// TODO(Adam): this should be more of a noop dependency package
+		panic("no interpreter dependencies found")
```
nathanielc: Should this just be an explicit error in the meantime instead of a panic?

compiler/compiler.go
sauren-khosla: Nope, deleting.

lang/compiler.go
aanthony1243: it's used in BucketsAccessed in InfluxDB.   This will not require deps, our plan is to have the `Default` deps be the always-error case that we discussed offline.   This will be done as a followup.  i'd rather not have the deps be nil anywhere to prevent panics.

PR #1742, opened 2019-08-21
fix(stdlib/universe): resolved holt-winters panic

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/stdlib/universe/holt_winters/holt_winters.go
+++ b/stdlib/universe/holt_winters/holt_winters.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/apache/arrow/go/arrow/array"
 	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/internal/mutable"
 )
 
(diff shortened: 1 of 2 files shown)


stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -63,7 +64,7 @@ func (r *HoltWinters) Do(vs *array.Float64) *array.Float64 {
 	r.vs = vs
 	l := vs.Len() // l is the length of both times and values
 	if l < 2 || r.seasonal && l < r.s || r.n <= 0 {
-		return mutable.NewFloat64Array(r.alloc).NewFloat64Array()
+		return arrow.NewFloat(nil, nil)
```
nathanielc: I don't follow the fix, as I understand it the error is when only a single row is provided to holt-winters. Why does this change make is safe? Also do we need to pass in `r.alloc` to `arrow.NewFloat`?

sauren-khosla: This return statement is only triggered if there is insufficient data to predict future tendencies using the Holt-Winters algorithm. We need to return a Float array, as that's what the other end of `Holt-Winters` (`universe/holt_winters.go`) is expecting (there are function calls such as `.Len()`, which won't work if we just return `nil`), though if there is nothing in the array we simply won't use it (nothing will be appended to the builder from the array). From what I understand, the issue lies in the `mutable` package and accessing data from the argument passed in.

Issue #1682, opened 2019-08-12
Pre-authorization secret lookup request function  `secrets.Get(key)`


aanthony1243:
calls the internal secrets API.

DOD:
* Add secrets Flux package
* Add `.get` method to pacakge
* Add tests for `.get` method.
* Add tests for authorization of `.get` method.

PR #1744, opened 2019-08-21
feat(stdlib): add secrets package that will construct a secret object

jsternberg:
The secret object will be constructed from a key. In the future, we'll
do a prevalidation of the access to the secret once we start threading
the secret service through standard functions. For now, this just
returns the object and we'll get an error when attempting to retrieve
the secret during the actual usage.

#1682

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/secrets/get.go
```diff
@@ -0,0 +1,38 @@
...
+// GetFunc is a function that calls Get.
+var GetFunc = makeGetFunc()
+
+func makeGetFunc() values.Function {
+	sig := semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"key": semantic.String,
+		},
+		Required: semantic.LabelSet{"key"},
+		Return:   semantic.Tvar(1),
```
nathanielc: This doesn't need to be a tvar, we can fully specify the type as an object with a `secretKey` property.

values/secret.go
```diff
@@ -0,0 +1,20 @@
+package values
```
nathanielc: Why does the `values` package need to know about secrets? I think these helper methods should exist in the `secrets` package itself.

jsternberg: My intention was that the code to access secrets should be in a common location that's abstracted away from influxdb, but I'm not sure that's actually needed so I'll move this code to the `secrets` package.

PR #1745, opened 2019-08-21
fix(options): make packages aware of options

nathanielc:
This change makes option mutations visible to builtin packages.
Previously option mutations were lost to builtin packages since they
maintained their own scope.
Now a few things happen:
* Scopes understand how to set options differently from normal variable
bindings.
* Each package maintains its own set of options.
* The interpreter tracks all mutations to options and applies them
retroactively to any function values that are passed to other functions.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


interpreter/interpreter.go
```diff
@@ -175,7 +177,77 @@ func (itrp *Interpreter) doStatement(stmt semantic.Statement, scope values.Scope
...
+		init, err := itrp.doExpression(a.Init, scope)
+		if err != nil {
+			return nil, err
+		}
+		// Use an empty string as the package name as we don't know its name.
+		// This will have one of two behaviors:
+		//     1. The option key will be found in the prelude and applied there.
+		//     2. The option key will not be found in the prelude and the
+		//        interpreter will handle adding the new option to the current package.
+		return itrp.setOption(scope, "", a.Identifier.Name, init)
```
jlapacik: The more I think about it, I'm not sure I like this. That is, conflating update and extend semantics. If there weren't any options defined in the prelude, then we wouldn't need this behavior as option updates would only be specified via member assignments. This could be achieved by defining `now` in the `system` package. Thoughts? Is having to specify `system.now()` a bad user experience?

nathanielc:
@jlapacik I agree, I explored using a special PackageExpression in contrast to MemberExpression. It wasn't something easily achievable. I think we should consider it for the future. 

As for not allowing any options in the prelude, that might be ok. Let's have an explicit discussion about it. My initial perspective is that we are making a user visible change to simplify logic on the backend, which feels like the wrong direction. There is likely a better way to handle the update/create semantics of options that doesn't require banning them in the prelude and doesn't use a hacky empty string signal. 

My thinking is when we bring packages as first class concepts in the type system instead of pretending they are objects, that will give us the chance to redo option semantics as well.

PR #1754, opened 2019-08-23
refactor(repl): repl now gets its context and dependencies set at startup and passed down.

aanthony1243:
…

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -25,6 +25,8 @@ import (
 type REPL struct {
 	scope   values.Scope
 	querier Querier
+	ctx     context.Context
+	deps    dependencies.Interface
 
 	cancelMu   sync.Mutex
 	cancelFunc context.CancelFunc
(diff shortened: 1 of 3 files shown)


repl/repl.go
```diff
@@ -34,7 +36,7 @@ type Querier interface {
 	Query(ctx context.Context, deps dependencies.Interface, compiler flux.Compiler) (flux.ResultIterator, error)
 }
 
-func New(q Querier) *REPL {
+func New(ctx context.Context, deps dependencies.Interface, q Querier) *REPL {
 	return &REPL{
 		scope:   values.NewScope(),
 		querier: q,
```
sauren-khosla: should you return 
``&REPL { 
scope: values.NewScope(), 
querier: q, 
ctx: ctx,
deps: deps
}`` here?

PR #1758, opened 2019-08-23
fix(stdlib/alerts): add test case for notify function

nathanielc:
This change fixes various bugs in order to get the notify test function
passing.

* Changes alerts.log to an option
* Changes alerts.write to use experimental.to
* Fixes bug with overriding option used in the same package

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -230,11 +230,26 @@ func (itrp *Interpreter) mutateFunctionScope(f function) (function, error) {
 	// copy the scope so we can safely mutate it
 	f.scope = f.scope.Copy()
 	copyPackages(f.scope)
+	mutatedPkg := false
 	for _, mut := range itrp.modifiedOptions {
+		// Check if the function is defined in the scope of package that was mutated
+		if f.pkg.Name() == mut.Package {
+			if !mutatedPkg {
+				f.pkg = f.pkg.Copy()
+			}
+			mutatedPkg = true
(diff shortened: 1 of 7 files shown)


interpreter/interpreter.go
```diff
@@ -230,12 +230,27 @@ func (itrp *Interpreter) mutateFunctionScope(f function) (function, error) {
 	// copy the scope so we can safely mutate it
 	f.scope = f.scope.Copy()
 	copyPackages(f.scope)
+	mutatedPkg := false
 	for _, mut := range itrp.modifiedOptions {
+		// Check if the function is defined in the scope of package that was mutated
```
jlapacik: Can you explain this new addition? I'm not quite sure I follow.

nathanielc: Yeah, given this is a builtin package

```
package foo

option bar = 1

do = () => bar
```

Then I do this in main:

```
package main

import "foo"
option foo.bar = 2

foo.do() // should return 2
```

Before the change it would return `1`. This was because the function `do` would close over its package scope. Calling `SetOption` on the `foo` package later had need effect on the package scope from within the `foo` package because the package scope is implemented by copying the package as an object.

The fix is to check if the scope of the function is of a package that had one of its options mutated. If so we need to do set the option on the package, and we need to recreate the package scope from the modified package.

PR #1766, opened 2019-08-23
feat(stdlib/http): add ability to validate URLs before making http.post requests

nathanielc:
This change also includes fixes to make adding new dependencies easier.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


dependencies/dependencies.go
```diff
@@ -12,40 +13,47 @@ const InterpreterDepsKey = "interpreter"
 type Interface interface {
 	HTTPClient() (*http.Client, error)
 	SecretService() (SecretService, error)
+	URLValidator() (url.Validator, error)
 }
 
 type DefaultDependencies struct {
-	httpclient    *http.Client
-	secretservice SecretService
+	Deps Deps
```
aanthony1243: maybe I'm overthinking the design but if `deps` are shared between components they should be implemented as immutable.  

I don't mind double-encapsulating deps to do this, but making deps a private member of DefaultDependencies.  Then we can do `NewDefaultDependencies(deps)`  to construct it.

dependencies/dependencies.go
```diff
@@ -12,40 +13,47 @@ const InterpreterDepsKey = "interpreter"
...
+	return nil, errors.New(codes.Unimplemented, "url validator uninitialized in dependencies")
 }
 
-func NewDependenciesInterface(hc *http.Client, ss SecretService) Interface {
-	return &DefaultDependencies{
-		httpclient:    hc,
-		secretservice: ss,
+// NewDefaultDependecies produces a set of dependencies.
+// Not all dependencies have valid defaults and will not be set.
+func NewDefaultDependencies() DefaultDependencies {
```
aanthony1243: I think the default dependencies should be all-nil so that we get errors by default when we use them.  or else we need to have a special type of dependencies struct that has this behavior.   

Then we could add a `NewBasicDependencies` or something to represent this case.  

The issue is that there are a handful of places such as package evaluation that need to initialize dependencies to avoid nil references, but we don't want those dependencies to be functional in any way (e.g. loading a package should not send HTTP requests)
Using NewDependenciesInterface

PR #1768, opened 2019-08-26
refactor(dependencies): move secret service to its own package

affo:


affo:
I have checked and this PR doesn't need any change in OSS

PR #1778, opened 2019-08-26
feat(alerts): pager duty endpoint

### LINKED ISSUE #1690: Create pagerduty endpoint
Create a PagerDuty endpoint as described here https://github.com/influxdata/flux/blob/master/docs/Outputs.md

NOTE: The example there uses Slack, but the pattern extends to how PagerDuty should operate.

NOTE: This has been specified to use the v2 PagerDuty API

Here are a list of parameters to the `endpoint` function:

* url - string - Should default to `https://events.pagerduty.com/v2/enqueue` using an option like the Slack example

Here are a list of properties that should exist on the return of the mapFn object:

* routingKey - string - The pagerduty routing key for a given pagerduty service. This will need to be a secret, ignore that for now.
* links - list of objects - Each object must have a `href` and a `text` property.
* client - string - Name of the client calling PagerDuty. Default to `flux`.
* clientURL - string - URL to contact the client. Deafults to empty string
* dedupKey - string - This needs to be a unique value per status, not sure how to get this yet. Leave it empty for now.
* customDetails - object - Arbitrary data the user can add to the pager duty message
* class - string - The class/type of the event, for example `ping failure` or `cpu load`. Maybe we should default this to the check name?
* group - string - Logical grouping of components of a service, for example `app-stack`
* eventAction - string - Must be `trigger` or `resolve`.
* severity - string - One of `info`, `warning`, `critical`. Severity can be omitted if the `eventAction` is `resolve`.
* source - string - Source of the alert, typically the `host` tag value.
* summary - string - Summary text to display
* timestamp - string - Timestamp formatted as `2006-01-02T15:04:05.000000000Z07:00`

See this Kapacitor code for how to call the PagerDuty API. https://github.com/influxdata/kapacitor/blob/master/services/pagerduty2/service.go#L286 Basically you need to construct a Flux object that represents the JSON to post to the PagerDuty API. 

See also https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2

AlirieGray:
Closes #1690 

This PR creates an endpoint to send an event to the Pagerduty V2 API to trigger an alert.

NOTE: not ready to merge until Flux bug fix with type checking is merged (after this, the commented out lines in runtime.go will be uncommented)

- [ ] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
This is looking good, can you rebase the change and fix the commit messages to be semantic? Also there is failure with staticcheck that needs to be fixed. Otherwise this looks good. Thanks!

execute/dispatcher.go
```diff
@@ -65,6 +65,7 @@ func (d *poolDispatcher) Start(n int, ctx context.Context) {
 			// Setup panic handling on the worker goroutines
 			defer func() {
 				if e := recover(); e != nil {
+					debug.PrintStack()
```
jlapacik: This can be removed now.

stdlib/pagerduty/pagerduty.flux
```diff
@@ -0,0 +1,93 @@
...
+    return http.post(headers: headers, url: pagerdutyURL, data: enc)
+}
+
+// `endpoint` creates the endpoint for the Slack external service.
+// `url` - string - URL of the slack endpoint. Defaults to: "https://slack.com/api/chat.postMessage", if one uses the webhook api this must be acquired as part of the slack API setup, and this URL will be secret.
+// `token` - string - token for the slack endpoint.  This can be ignored if one uses the webhook url acquired as part of the slack API setup, but must be supplied if the chat.postMessage API is used.
+// The returned factory function accepts a `mapFn` parameter.
+// The `mapFn` parameter must be a function that returns an object with `pagerdutyURL`, `token`, `client`, `client_url`,`class`,`group`, `severity`, `component`, `source`, `summary`, and `timestamp` as defined in the sendEvent function.
+// Note that while sendEvent accepts a dedup key, endpoint gets the dedupkey from the groupkey of the input table instead of it being handled by the `mapFn`.
+endpoint = (url=defaultURL, token="") =>
```
nathanielc: This comment appears to be a copy paste from Slack, can you update it to be pagerduty specific?

stdlib/pagerduty/pagerduty.flux
```diff
@@ -0,0 +1,93 @@
...
+    (mapFn) =>
+        (tables=<-) => tables
+            |> dedupKey() // TODOO(docmerlin): add a way to specify a column and just default to the group key
```
nathanielc: Is this TODO still relevant? If not can we remove this comment?

stdlib/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,211 @@
...
+			return nil, fmt.Errorf("missing argument: severity")
+		}
+
+		if v.Type().Nature() == semantic.String {
+			return values.NewString(actionFromSeverity(v.Str())), nil
+		}
+		return nil, fmt.Errorf("could not parse severity")
+	},
+	false,
+)
```
nathanielc: Based on this implementation I would expect this function can be defined in pure Flux, something like:

```
actionFromSeverity = (severity) => if strings.toLower(v:severity) == "ok"" then "resolve" else "trigger"
```

Can you change it to be pure Flux?

stdlib/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,211 @@
...
+
+	err := tbl.Do(func(cr flux.ColReader) error {
+		l := cr.Len()
+		if err := execute.AppendCols(cr, builder); err != nil {
+			return err
+		}
+		colIDX, err := builder.AddCol(flux.ColMeta{
+			Label: dedupKeyColName,
+			Type:  flux.TString,
+		})
```
nathanielc: This will add a column to the builder for each col reader in the incoming table which is not what we want. The column should be added once outside the `tbl.Do` loop.

An example of when this would break is when we read data from storage we read it in 1000 row chunks, each chunk becomes its own ColReader. So a table with 3000 rows would have three col readers and would add the column to the builder three times.

testserver/main.go
```diff
@@ -0,0 +1,17 @@
+package main
```
nathanielc: Is this code used in any tests? If not can you remove it?

stdlib/pagerduty/pagerduty.flux
```diff
@@ -0,0 +1,93 @@
+package pagerduty
+
+import "http"
+import "json"
+
+builtin dedupKey
```
nathanielc: Can you add a comment there explaining what this function does?

stdlib/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// because pagerduty restricts the dedupKey size we are hashing to reduce chance of collisions, by distributing the key more evenly
+	dedupKeyHash := sha256.Sum256([]byte(dedupKey))
+
+	err = tbl.Do(func(cr flux.ColReader) error {
+		l := cr.Len()
+		for j := range builder.Cols() {
+			if j == colIDX {
+				for i := 0; i < l; i++ {
+					if err := builder.AppendValue(j, values.NewString(hex.Dump(dedupKeyHash[:]))); err != nil {
```
nathanielc: As a small optimization can you move this part `values.NewString(hex.Dump(dedupKeyHash[:]))`  outside of the `tbl.Do` loop as it doesn't change and is allocation heavy. 

Then you can simply pass the same value in to each row. Values are immutable so you don't have to worry about sharing it.

Issue #1782, opened 2019-08-28
function duration causes a panic


russorat:
```
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> filter(fn: (r) => r._field == "usage_user")
  |> filter(fn: (r) => r.cpu == "cpu-total")
  |> filter(fn: (r) => r.host == "rsavage.prod")
  |> map(fn: (r) => ({ r with _value:  duration(v: "1h")}))
```

```
{
"code":"internal error",
"message":"panic",
"error":"unknown type invalid"
}
```

sauren-khosla:
If you try the same query but with a different function in `map` (something that's not `duration`), do you still encounter a panic?

russorat:
this doesn't cause a panic:

```
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> filter(fn: (r) => r._field == "usage_user")
  |> filter(fn: (r) => r.cpu == "cpu-total")
  |> filter(fn: (r) => r.host == "rsavage.prod")
  |> map(fn: (r) => ({ r with _value: float(v: 1)}))
```

PR #1788, opened 2019-08-29
fix(execute): error on duration columns

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

See https://github.com/influxdata/flux/issues/1782


yzhang1991:
I remembered we concluded yesterday in the Slack channel that we will add the real support for this feature?

PR #1798, opened 2019-09-04
fix(dependencies): changed the default http client to be more robust

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/dependencies/dependencies.go
+++ b/dependencies/dependencies.go
@@ -1,7 +1,9 @@
 package dependencies
 
 import (
+	"net"
 	"net/http"
+	"time"
 
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/dependencies/secret"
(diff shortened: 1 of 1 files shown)


dependencies/dependencies.go
```diff
@@ -50,12 +52,29 @@ func (d Dependencies) URLValidator() (url.Validator, error) {
...
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			// DualStack is deprecated
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
```
nathanielc: I think you lost the max connections per host option.

dependencies/dependencies.go
```diff
@@ -50,12 +53,30 @@ func (d Dependencies) URLValidator() (url.Validator, error) {
...
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			// DualStack is deprecated
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		MaxIdleConnsPerHost:   runtime.GOMAXPROCS(0) + 1,
```
nathanielc: What is the reasoning for this default? You had it set to `100` earlier.

sauren-khosla: I pulled this from a commit on another branch, but I just checked the defaults in a doc elsewhere and they changed to 100. I'll update that now.

dependencies/dependencies.go
```diff
@@ -50,12 +52,30 @@ func (d Dependencies) URLValidator() (url.Validator, error) {
...
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			// DualStack is deprecated
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		MaxIdleConnsPerHost:   100,
```
nathanielc: Looks good, but can you add a comment to this saying is not part of the defaults so we know which ones we have changed and which are from Go's defaults?

PR #1809, opened 2019-09-05
fix(stdlib): sleep function now gets canceled when the context is canceled

jsternberg:
The current version of the sleep function will continue sleeping
indefinitely even if the context is canceled. This prevents the sleep
function from running indefinitely and causing the query queue from
being stalled forever.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -582,9 +582,22 @@ func functionName(call *semantic.CallExpression) string {
 	}
 }
 
+// DoFunctionCall will call DoFunctionCallContext with a background context.
 func DoFunctionCall(f func(args Arguments) (values.Value, error), argsObj values.Object) (values.Value, error) {
+	return DoFunctionCallContext(func(_ context.Context, args Arguments) (values.Value, error) {
+		return f(args)
+	}, context.Background(), argsObj)
+}
+
+// DoFunctionCallContext will treat the argsObj as the arguments to a function.
+// It will then invoke that function with the Arguments and return the
+// value from the function.
+//
+// This function verifies that all of the arguments have been consumed
+// by the function call.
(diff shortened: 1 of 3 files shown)


stdlib/universe/sleep.go
```diff
@@ -32,21 +33,37 @@ var sleep = values.NewFunction(
...
-			time.Sleep(time.Duration(dur))
-		}
-		return v, nil
+		return interpreter.DoFunctionCallContext(Sleep, ctx, args)
 	},
 	// sleeping is a side effect
 	true,
 )
+
+func Sleep(ctx context.Context, args interpreter.Arguments) (values.Value, error) {
```
nathanielc: Does this need to be public only for the test case? If so should we adjust the test case to use the `universe` package so it is private?

interpreter/interpreter.go
```diff
@@ -594,6 +594,18 @@ func DoFunctionCall(f func(args Arguments) (values.Value, error), argsObj values
 	return v, nil
 }
 
+func DoFunctionCallContext(f func(ctx context.Context, args Arguments) (values.Value, error), ctx context.Context, argsObj values.Object) (values.Value, error) {
```
nathanielc: Can you document this function?

stdlib/universe/sleep.go
jsternberg: It's not because of the test case. This is the same pattern I did with `runtime.Version`. I thought it was a good pattern because it made a public interface that could be used to access the native go functions.

If desired, I can hide it. I kind of like this pattern though. I can see how we might want to hide it though in circumstances where we might have a native go version but want to replace it with pure flux later.

PR #1811, opened 2019-09-05
docs(execute): add documentation to Transformation interface

fchikwekwe:
Adding documentation to the Transformation interface since there was none before. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ vendor
 /internal/rust/parser/pkg
 /internal/rust/site/node_modules
 /internal/rust/.cache
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ vendor
 /internal/rust/parser/pkg
 /internal/rust/site/node_modules
 /internal/rust/.cache
+.DS_Store
--- a/execute/transformation.go
+++ b/execute/transformation.go
@@ -9,11 +9,17 @@ import (
 	"github.com/influxdata/flux/plan"
 )
 
+// Transformation represents functions that control of how the execution
+// engine will take in a flux Table, perform the necessary data processing
+// and return an output flux Table
 type Transformation interface {
 	RetractTable(id DatasetID, key flux.GroupKey) error
+	// Process takes in a flux Table, performs data processing on it
+	// and produces an output flux Table
 	Process(id DatasetID, tbl flux.Table) error
 	UpdateWatermark(id DatasetID, t Time) error
 	UpdateProcessingTime(id DatasetID, t Time) error
+	// Finish indicates that the Transformation is complete
 	Finish(id DatasetID, err error)
 }
 
--- a/execute/transformation.go
+++ b/execute/transformation.go
@@ -9,17 +9,17 @@ import (
 	"github.com/influxdata/flux/plan"
 )
 
-// Transformation represents functions that will taking in data from a Dataset
-// and perform the necessary data processing on it
+// Transformation represents functions that stream a set of tables, performs
+// data processing on them and produces an output stream of tables
 type Transformation interface {
 	RetractTable(id DatasetID, key flux.GroupKey) error
-	// Process takes in a DatasetID referring to a list of all transformations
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ vendor
 /internal/rust/parser/pkg
 /internal/rust/site/node_modules
 /internal/rust/.cache
+.DS_Store
```
nathanielc: Can you remove this change and instead add this a global gitignore on your laptop? See https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

execute/transformation.go
```diff
@@ -9,11 +9,17 @@ import (
...
 )
 
+// Transformation represents functions that control of how the execution
+// engine will take in a flux Table, perform the necessary data processing
+// and return an output flux Table
 type Transformation interface {
 	RetractTable(id DatasetID, key flux.GroupKey) error
+	// Process takes in a flux Table, performs data processing on it
+	// and produces an output flux Table
 	Process(id DatasetID, tbl flux.Table) error
```
nathanielc: This is correct in concept but the method signature doesn't really communicate the same thing. In other words, if the function produces an output table where is it? Its not on the function signature. Can you explain where the output table is created?

execute/transformation.go
```diff
@@ -9,11 +9,17 @@ import (
 	"github.com/influxdata/flux/plan"
 )
 
+// Transformation represents functions that control of how the execution
+// engine will take in a flux Table, perform the necessary data processing
+// and return an output flux Table
```
nathanielc: What about transformations that process more than one table?

execute/transformation.go
```diff
@@ -9,11 +9,17 @@ import (
...
 type Transformation interface {
 	RetractTable(id DatasetID, key flux.GroupKey) error
+	// Process takes in a flux Table, performs data processing on it
+	// and produces an output flux Table
 	Process(id DatasetID, tbl flux.Table) error
 	UpdateWatermark(id DatasetID, t Time) error
 	UpdateProcessingTime(id DatasetID, t Time) error
+	// Finish indicates that the Transformation is complete
 	Finish(id DatasetID, err error)
```
nathanielc: What does `complete` mean? I can answer that one, `finished` or  `complete` mean that no other methods on the Transformation will be called. I.e. its always the last method called on the transformation.

.gitignore
fchikwekwe: Yeah,  no problem. I can update that.

execute/transformation.go
fchikwekwe: When I look at the `createToTransformation` function in the `experimental` package in IDPE, it returns a `Dataset` and a `Transformation`.  Based on this, and the function signature that you pointed out, it seems that the functions in the `Transformation` interface only return nil/error. I _think_ that the working information for the `Transformation` is stored in a `Dataset` and is returned by the function that creates the `Transformation` and the `Dataset`. In this case, `createToTransformation`.

execute/transformation.go
fchikwekwe: Good point. I can update the language to consider more than one table.

execute/transformation.go
fchikwekwe: That's fair. I can update this to be more clear.

execute/transformation.go
fchikwekwe: @nathanielc I know that this isn't quite right, but I think I'm understanding the code a little bit better. Let me know what you think and I'll update it.

nathanielc: @fchikwekwe That is correct, the Process method is responsible for creating a new output table and adding it to the dataset.

execute/transformation.go
nathanielc: The vocabulary we use is `stream` is a set of tables.

fchikwekwe: @nathanielc I spoke with Chris a little bit about this and the Process comment so I made another change. Let me know what you think and I might mark it ready for review.

PR #1816, opened 2019-09-06
feat(dependencies): replace EnvironmentSecretService with EmptySecret…

yzhang1991:
…Service

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

---
This change is [Reviewable](https://reviewable.io/reviews/influxdata/flux/1816)

### DIFF
--- a/dependencies/secret/empty_secret_service.go
+++ b/dependencies/secret/empty_secret_service.go
@@ -1,0 +1,14 @@
+package secret
+
+import (
+	"context"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+)
+
+func (ess EmptySecretService) LoadSecret(ctx context.Context, k string) (string, error) {
+	return "", errors.Newf(codes.NotFound, "secret key %q not found", k)
+}
+
+type EmptySecretService struct {
+}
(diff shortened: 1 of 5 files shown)


dependencies/secret/empty_secret_service.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+)
+
+func (ess EmptySecretService) LoadSecret(ctx context.Context, k string) (string, error) {
+	return "", errors.Newf(codes.NotFound, "secret key %q not found", k)
+}
+
+type EmptySecretService struct {
```
nathanielc: Can you add a comment here explaining how the service behaves? i.e. it errors loading any secret.

fchikwekwe: @nathanielc do we need to write testing for updates like this? It didn't look like the previous version `environment_secret_service` had any tests. Just on `secret_service`

nathanielc: We could, but the test doesn't get us much as all it does is validate the method returns an error. Behavior which we can clearly audit in the code. I would not be opposed to adding a simple unit test for the service, but I wouldn't require it either. I'll leave it up to the author in this case.

PR #1817, opened 2019-09-06
fix(pagerduty): add severity level formatting helper function

AlirieGray:
Severity levels on status objects can be one of the following: `ok, info, warn, crit, unknown` according to the spec shown here: https://github.com/influxdata/idpe/blob/master/docs/content/rfcs/alerts_and_notifications.md#bucket-data-model

However, pagerduty only accepts: `critical, error, warning, info`

This PR adds a helper function called `severityFromLevel` turns a level from the status object into a pagerduty severity

For example, changing `warn` -> `warning`

The `endpoint` and `sendEvent` functions will both take data in the form that Pagerduty expects. This means that the user or the generated monitoring code will need to call the helper functions `severityFromLevel` and `actionFromLevel` if passing data directly from a status object

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/pagerduty/pagerduty.flux
+++ b/stdlib/pagerduty/pagerduty.flux
@@ -12,18 +12,19 @@ option defaultURL = "https://events.pagerduty.com/v2/enqueue"
 
 // severity levels on status objects can be one of the following: ok,info,warn,crit,unknown
 // but pagerduty only accepts critical, error, warning or info.
-// formatSeverity turns a level from the status object into a pagerduty severity
-formatSeverity = (severity) => {
-    sev = strings.toLower(v:severity) 
-    if sev == "warn" then "warning"
-    else if sev == "crit" then "critical"
-    else if sev == "info" then "info"
-    else if sev == "ok" then "info"
-    else "unknown"
+// severityFromLevel turns a level from the status object into a pagerduty severity
+severityFromLevel = (level) => {
+    lvl = strings.toLower(v:level)
+    sev = if lvl == "warn" then "warning" 
+        else if lvl == "crit" then "critical" 
+        else if lvl == "info" then "info" 
+        else if lvl == "ok" then "info" 
+        else "error"
+    return sev
 }
 
-// `actionFromSeverity` converts a severity to an action; "ok" becomes "resolve" everything else converts to "trigger".
-actionFromSeverity = (severity)=> if strings.toLower(v:severity) == "ok" then "resolve" else "trigger"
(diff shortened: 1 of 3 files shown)


AlirieGray:
Unsure what to do about the "ok" level and the "error" severity

stdlib/pagerduty/pagerduty.flux
```diff
@@ -10,6 +10,18 @@ builtin dedupKey
 option defaultURL = "https://events.pagerduty.com/v2/enqueue"
 
 
+// severity levels on status objects can be one of the following: ok,info,warn,crit,unknown
+// but pagerduty only accepts critical, error, warning or info.
+// formatSeverity turns a level from the status object into a pagerduty severity
+formatSeverity = (severity) => {
```
nathanielc: Should this be called `severityFromLevel`  and then name the argument `level`? That would make it clear which way this function converts levels to severities. 

```
severityFromLevel(level) => ...
```

stdlib/pagerduty/pagerduty.flux
```diff
@@ -10,6 +10,18 @@ builtin dedupKey
...
+    sev = strings.toLower(v:severity) 
+    if sev == "warn" then "warning"
+    else if sev == "crit" then "critical"
+    else if sev == "info" then "info"
+    else if sev == "ok" then "info"
+    else "unknown"
+}
+
 // `actionFromSeverity` converts a severity to an action; "ok" becomes "resolve" everything else converts to "trigger".
 actionFromSeverity = (severity)=> if strings.toLower(v:severity) == "ok" then "resolve" else "trigger"
```
nathanielc: Should this function change now? It seems like this function should be called `actionFromLevel` to be consistent with the above changes. 

```
actionFromLevel = (level) => ...
```

Also note that its not possible to define `actionFromSeverity` since both `ok` and `info` levels are converted to the `info` severity.

AlirieGray:
From thinking about the level/severity naming issue that @nathanielc raised, it seems like the best place to do this transformation is actually in the monitoring code in influxdb. That way, this endpoint can remain just a Pagerduty endpoint that anyone can call with the data that the Pagerduty API expects, rather than tightly coupling it to the shape of our monitoring data.

nathanielc:
@AlirieGray  I think these should be part of the package. The `endpoint` functions job is to translate between our monitoring data shape and what pager duty expects. In that sense the `sendEvent` function can expect all pagerduty specific properties and the `endpoint` function can use these two helper methods to perform the translation.

PR #1834, opened 2019-09-09
flux(http.post): flush tcp buffers on the http response body to make …

docmerlin:
…sure that we can reuse sockets


docmerlin:
This no longer applies.  Golang now flushes up to 256Megs on close().

PR #1836, opened 2019-09-10
fix(parser): push error for bad string expression

affo:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -917,7 +917,13 @@ func (p *parser) parseStringExpression() *ast.StringExpression {
 			}
 		default:
 			// TODO bad expression
-			return nil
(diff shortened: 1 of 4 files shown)


internal/parser/parser.go
```diff
@@ -917,7 +917,13 @@ func (p *parser) parseStringExpression() *ast.StringExpression {
 			}
 		default:
 			// TODO bad expression
```
aanthony1243: drop the todo?

affo: I left the TODO because we have different points in which we should use a hypothetic  `BadExpression` node, and this could be one of them.

affo:
@jlapacik I don't know if there's an issue about this 🤔

PR #1843, opened 2019-09-11
fix(internal/spec): do a pointer comparison for table objects instead of a deep compare

jsternberg:
The conversion from an AST to a spec would evaluate the AST and then
return a spec from the side effects that were related to table objects.

In order to do this, the code that converted a table object into a spec
would also check to see if that table object had already been seen
before.

This caused a performance problem because the `Equal` method for a
`TableObject` would recursively check the `TableObject` which also
included checking the parents. If you had a reasonably complex query,
this could take a very long time as that code was not very efficient and
involved lots of small memory allocations.

This changes it so that instead of recursively checking, it just checks
the memory location which should be the same when they are duplicates
anyway.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/spec/build.go
+++ b/internal/spec/build.go
@@ -159,7 +159,7 @@ func isDuplicateTableObject(ctx context.Context, op *flux.TableObject, objs []*f
 	defer s.Finish()
 
 	for _, tableObject := range objs {
-		if op.Equal(tableObject) {
+		if op == tableObject {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


jsternberg:
Benchmarks:

```
benchmark              old ns/op      new ns/op     delta
Benchmark_Build-12     7707517459     34494225      -99.55%

benchmark              old allocs     new allocs     delta
Benchmark_Build-12     140373067      391713         -99.72%

benchmark              old bytes      new bytes     delta
Benchmark_Build-12     4959824792     22660985      -99.54%
```

I manually verified with code coverage that the benchmark covered that piece of code and took the true conditional path.

PR #1848, opened 2019-09-11
fix(parser): introduce ParenExpression

affo:
I chose to make the `SourceLocation` returned by `Node.Location()` mutable in order to make it possible to override the source location.

This is needed in order to change the location of a retuned expression once it is inside of parenthesis.

If I try to introduce the change in the concrete expression, then I don't have enough context and I don't know if I am inside of parens.

If I introduce the change at an higher level (`parseParenExpression`), then too many things get the parenthesis in (for example: `f = (a) => ...`, the identifier `a` becomes `(a)`).

That seemed to me a good trade-off.

I also had to make changes to the Rust part and create a method to set the location for an expression.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


affo:
@jsternberg `ParenExpression` introduced.

This is way better

internal/parser/parser.go
```diff
@@ -1036,8 +1036,8 @@ func (p *parser) parseObjectLiteral() ast.Expression {
 }
 
 func (p *parser) parseParenExpression() ast.Expression {
-	pos, _ := p.open(token.LPAREN, token.RPAREN)
```
jsternberg: Why the name change?

semantic/analyze.go
```diff
@@ -281,6 +281,9 @@ func analyzeExpression(expr ast.Expression) (Expression, error) {
 		return analyzeIdentifierExpression(expr)
 	case *ast.StringExpression:
 		return analyzeStringExpression(expr)
+	case *ast.ParenExpression:
+		// Unwrap the parenthesis and analyze underlying Expression.
+		return analyzeExpression(expr.Expression)
```
jsternberg: 👍

internal/parser/parser.go
affo: cause I forgot to revert it back, fixing now

Issue #1861, opened 2019-09-13
Turn the *memory.Allocator into an implementation of arrow's memory allocator


jsternberg:
The `*memory.Allocator` type should implement the arrow memory allocator interface. This will make it easier for the allocator to be used by arrow code and it would allow us to embed the underlying arrow allocator instead of the hack that we have in the `arrow` package.

PR #1862, opened 2019-09-13
feat: convert the flux memory allocator into an arrow allocator

jsternberg:
This converts the flux memory allocator to support the same interface as
the arrow memory allocator so the two can be used interchangeably.

The purpose of this is to allow greater integration of arrow allocators
and to open up the possibility of pooled memory for buffers to reduce
memory load. At the moment, the flux memory allocator is shared among
the entire query and is commonly wrapped by a temporary arrow allocator
that combines the two. This change reverses that dynamic by having the
flux memory allocator invoke the arrow allocator directly since that is
now possible with the changed interface.

This will allow us to set a single arrow allocator for the entire query
instead of creating a new one for each builder like we currently do.

BREAKING CHANGE: The `*memory.Allocator` interface has changed so
`Allocate` and `Free` are now just `Account`.

#1861

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/memory/allocator.go
+++ b/memory/allocator.go
@@ -4,9 +4,17 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 )
+
+// DefaultAllocator is the default memory allocator for Flux.
+//
+// This implements the memory.Allocator interface from arrow.
+var DefaultAllocator = memory.DefaultAllocator
+
+var _ memory.Allocator = (*Allocator)(nil)
 
 // Allocator tracks the amount of memory being consumed by a query.
 type Allocator struct {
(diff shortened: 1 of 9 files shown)


memory/allocator.go
```diff
@@ -4,28 +4,76 @@ import (
...
 
+	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 )
 
+// DefaultAllocator is the default memory allocator for Flux.
+var DefaultAllocator = memory.DefaultAllocator
+
 // Allocator tracks the amount of memory being consumed by a query.
```
wolffcm: It would be good to say that this implements the Arrow allocator interface in this comment.

wolffcm: Or even add a line a `var _ memory.Allocator ...` that enforces it.

jsternberg: I'll add a comment, but I can't really add the above because it's already that type in the type system.

PR #1864, opened 2019-09-13
feat(stdlib/internal/promql): make flux quantile() function account for quantile values above 1 or below 0 in the same way the promQL version does

### LINKED ISSUE #1044: promql: Make percentile() edge-case behavior match PromQL
PromQL's `quantile()` can be *almost* emulated using Flux's `percentile()`, except that PromQL returns `-Inf` / `+Inf` if one chooses quantile values below `0` or above `1`, whereas Flux doesn't.

We could either ignore this (unlikely to come up in real use cases), change the behavior in Flux, or additionally add a PromQL-compatible version into Flux.

fchikwekwe:
Updates the flux quantile function so that it accounts for quantile values that are above 1 by returning `+Inf` and for quantile values that are below 0 by returning `-Inf` in the same way that the promQL quantile aggregate function does. 

Closes #1044
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/internal/promql/flux_gen.go
+++ b/stdlib/internal/promql/flux_gen.go
@@ -21,11 +21,11 @@ var pkgAST = &ast.Package{
 			Errors: nil,
 			Loc: &ast.SourceLocation{
 				End: ast.Position{
-					Column: 19,
-					Line:   98,
+					Column: 96,
+					Line:   117,
 				},
 				File:   "promql.flux",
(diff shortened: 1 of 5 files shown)


stdlib/experimental/promql/promql.flux
```diff
@@ -0,0 +1,20 @@
+package promql
```
jsternberg: Can we place this method in the `internal/promql` package?

stdlib/experimental/promql/promql_quantile_neg_test.flux
```diff
@@ -0,0 +1,44 @@
+package promql_test
```
jsternberg: promql tests can go with the other tests in the `stdlib/testing/testdata` folder.

stdlib/experimental/promql/promql.flux
fchikwekwe: Yeah, let me go ahead and update that.

stdlib/experimental/promql/promql_quantile_neg_test.flux
fchikwekwe: True. I'll update that

stdlib/experimental/promql/promql.flux
fchikwekwe: @jsternberg I moved the files around so that they are in the correct folders now.

fchikwekwe:
alchemy please merge

PR #1878, opened 2019-09-16
fix(stdlib): fix some of the memory leaks within the standard library

jsternberg:
Fix some of the remaining leaks within the standard library and enable
tests to ensure that memory leaks are checked for every end to end test.

There were two functions that I could not figure out how to fix for
memory leaks. The `join` function was too complex for me to feel
comfortable with deciding when to release memory. The `cov` function
uses the `join` function so I similarly could not fix that one. These
tests are specially skipped until we are able to address the memory
leaks in `join` with a possible refactor.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/testing/diff.go
+++ b/stdlib/testing/diff.go
@@ -311,6 +311,7 @@ func (t *DiffTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 	// to prematurely declare the other table as finished so we
 	// don't do more work on something that failed anyway.
 	if t.finished[id] {
+		tbl.Done()
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


stdlib/flux_test.go
```diff
@@ -149,8 +164,14 @@ func doTestInspect(t testing.TB, c flux.Compiler) {
 			t.Error(err)
 		}
 	}
+	r.Done()
```
yzhang1991: There is a `defer r.Done()` above?

jsternberg: This ensures that the query is done before we take the statistics which is needed to ensure that the statistics are computed. The `defer` is done above to ensure that this always happens even on error conditions. It is safe to call `Done()` multiple times.

Issue #1795, opened 2019-09-04
Refactor dependency injection in Flux


affo:
This should solve different issues:
 1. make it so that injecting dependency is clean an clear;
 2. make it so that injecting dependency is extendable once one uses Flux as a library;
 3. enable dynamic queries.

For `3`, dependencies should include the `memory.Allocator` used for the current query.

Dependencies are a mixture of static ones (bucket lookup service, storage reader, points writer, the secret service etc.) and dynamic (the allocator, the logger?, etc.). We need to find the best pattern to incorporate them.

Given that there is at least one dynamic dependency, there should be a new dependency object that gets created per request.

Possible ways of implementing this:
 1. use Go's `context.Context` and use private keys to inject the deps into it (very flexible);
 2. some framework for Dependency Injection (DI) in Go (idea to explore);
 3. Object embedding. Flux define its deps, InfluxDB embeds those and adds its ones and casts the dep object to extract them when it needs to. Other Flux users can do the same.

I have already implemented a refactor for InfluxDB and Flux to remove `execute.Dependencies` and only use `dependencies.Interface` (implementing strategy `3`). That is a good starting point to move to another implementation if needed.

The changes are WIP, this is the link to them:
 - Flux: https://github.com/influxdata/flux/compare/exp/dynamic-queries
 - InfluxDB: https://github.com/influxdata/influxdb/compare/exp/dynamic-queries

This issue also makes the implementation  for the new design of compilation and execution unnecessary (#1499  and its children).

jsternberg:
I had worked on a prototype to how dependency injection might work using context. I like context because it is very flexible and I think it is more user friendly from a programming point of view. Since it is a common Go library and idiom, it would work well. Further, I believe that moving to using the `context.Context` would be a good middle ground as it would be easily possible to inject all of the dependencies in `execute.Dependencies` into the context and have them live in the dependencies so that the code transition is lighter.

I kind of fear that another major refactor of the code without a clear transition plan will be too difficult.

https://github.com/influxdata/flux/compare/refactor/dependency-injection?expand=1

Now the exact method of how the dependencies are structured inside of the context is very flexible. For example, we could just inject the `execute.Dependencies` themselves into the context. I am not a big fan of the string keys and how the dependencies are usually organized on a per procedure basis, but it would be something that was possible. But the above pattern is very flexible on the internal details of how the stuff is stored in the context and I think it would provide a good base to start with.

affo:
[...]

PR #1879, opened 2019-09-17
feat(dependencies): new dependency injection framework

affo:
#1795 

 - No more `dependencies` package (moved to `flux`);
 - No more `execute.Dependencies`;
 - `Inject` pattern;
 - Fix to `Program` to make it inject `tableFind` dependencies;
 - Huge refactor to remove `dependencies.Interface` to function signatures.

I flattened the deps in a single object.

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -343,8 +343,7 @@ func evalBuiltInPackages() error {
 		}
 
 		itrp := interpreter.NewInterpreter(pkg)
-		ctx := NewEmptyDependencies().Inject(context.Background())
-		if _, err := itrp.Eval(ctx, semPkg, preludeScope.Nest(pkg), stdlib); err != nil {
+		if _, err := itrp.Eval(context.Background(), semPkg, preludeScope.Nest(pkg), stdlib); err != nil {
 			return errors.Wrapf(err, codes.Inherit, "failed to evaluate builtin package %q", astPkg.Path)
 		}
 	}
--- a/compile.go
+++ b/compile.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/codes"
-	"github.com/influxdata/flux/dependencies"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/parser"
@@ -40,16 +39,16 @@ func Parse(flux string) (*ast.Package, error) {
 }
 
 // Eval accepts a Flux script and evaluates it to produce a set of side effects (as a slice of values) and a scope.
-func Eval(ctx context.Context, deps dependencies.Interface, flux string, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+func Eval(ctx context.Context, flux string, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
 	astPkg, err := Parse(flux)
 	if err != nil {
 		return nil, nil, err
 	}
-	return EvalAST(ctx, deps, astPkg, opts...)
+	return EvalAST(ctx, astPkg, opts...)
 }
 
 // EvalAST accepts a Flux AST and evaluates it to produce a set of side effects (as a slice of values) and a scope.
-func EvalAST(ctx context.Context, deps dependencies.Interface, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
 	semPkg, err := semantic.New(astPkg)
 	if err != nil {
 		return nil, nil, err
@@ -65,7 +64,7 @@ func EvalAST(ctx context.Context, deps dependencies.Interface, astPkg *ast.Packa
 		opt(scope)
 	}
 
-	sideEffects, err := itrp.Eval(ctx, deps, semPkg, scope, StdLib())
+	sideEffects, err := itrp.Eval(ctx, semPkg, scope, StdLib())
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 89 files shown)


lang/compiler.go
```diff
@@ -449,3 +446,15 @@ func getRules(plannerPkg values.Object, optionName string) ([]string, error) {
 	})
 	return rs, nil
 }
+
+// WalkIR applies the function `f` to each operation in the compiled spec.
+// WARNING: this function evaluates the AST using an unlimited allocator.
+// In case of dynamic queries this could lead to unexpected memory usage.
```
affo: I think we should open up an issue for this. Thoughts?

lang/compiler.go
```diff
@@ -313,40 +306,44 @@ func (p *AstProgram) Start(ctx context.Context, alloc *memory.Allocator) (flux.Q
...
 	}
-	deps, ok := p.Dependencies[dependencies.InterpreterDepsKey]
-	if !ok {
-		// TODO(Adam): this should be more of a noop dependency package
-		return nil, fmt.Errorf("no interpreter dependencies found")
+	// The program must inject execution dependencies to make it available
+	// to function calls during the evaluation phase (see `tableFind`).
+	deps := ExecutionDependencies{
+		Allocator: alloc,
+		Logger:    p.Logger,
```
affo: This is a key point to make `tableFind` work

dependencies.go
```diff
@@ -12,56 +13,74 @@ import (
...
+type Dependency interface {
+	Inject(ctx context.Context) context.Context
+}
+
+type key int
+
+const dependenciesKey key = iota
 
-type Interface interface {
+type Dependencies interface {
```
jsternberg: I'd like to refactor this later, but I think this is the correct way to do this for now.

dependencies/dependenciestest/dependencies.go
```diff
@@ -1,10 +1,10 @@
 package dependenciestest
 
 import (
+	"github.com/influxdata/flux"
```
jsternberg: This import is in the wrong place.

examples/library/library_example_test.go
```diff
@@ -36,11 +36,8 @@ g.from(start: 1993-02-16T00:00:00Z, stop: 1993-02-16T00:03:00Z, count: 5, fn: (n
 		panic(err)
 	}
 
+	ctx = executetest.NewTestExecuteDependencies().Inject(ctx)
```
jsternberg: Should this be changed to `InjectTestExecuteDependencies`?

This is a fairly large change and I'd like to reduce moving parts so please don't change this at the moment, but we can consider this in a follow-up PR.

internal/cmd/refactortests/cmd/refactortests.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+
```
jsternberg: Random extra newline here.

interpreter/interptest/eval.go
```diff
@@ -4,14 +4,14 @@ import (
 	"context"
 
 	"github.com/influxdata/flux/ast"
-	"github.com/influxdata/flux/dependencies"
+
```
jsternberg: Random extra newline.

lang/compiler.go
```diff
@@ -216,25 +215,19 @@ func (*TableObjectCompiler) CompilerType() flux.CompilerType {
 	panic("TableObjectCompiler is not associated with a CompilerType")
 }
 
-type DependenciesAwareProgram interface {
-	SetExecutorDependencies(execute.Dependencies)
+type LoggingProgram interface {
```
jsternberg: Is there a way we can inject the logger into the context too?

stdlib/experimental/bigtable/bigtable_test.go
```diff
@@ -16,8 +18,8 @@ import (
 	"github.com/influxdata/flux/stdlib/universe"
 	"github.com/influxdata/flux/values"
 	"github.com/stretchr/testify/assert"
-	"testing"
-	"time"
+
```
jsternberg: Extra newline.

stdlib/experimental/object_keys.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	flux "github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
-	"github.com/influxdata/flux/dependencies"
+
```
jsternberg: Extra newline.

stdlib/universe/state_tracking.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/compiler"
-	"github.com/influxdata/flux/dependencies"
+
```
jsternberg: Random newline.

examples/library/library_example_test.go
affo: Yep, could be an improvement.

lang/compiler.go
affo: Yes, I think we can. The logger is actually needed to create the Executor and to make it log if something goes bad in the end. I inject the logger below (https://github.com/influxdata/flux/pull/1879/files#r325281074), in order to let a function create an Executor. The easiest way of doing this, should be to pass the logger to a Program upon Start.

Because actually, there is no need to inject it for making it available to the Program.

affo: I think we should defer this to a new PR

compile.go
```diff
@@ -344,7 +343,8 @@ func evalBuiltInPackages() error {
 		}
 
 		itrp := interpreter.NewInterpreter(pkg)
-		if _, err := itrp.Eval(context.Background(), dependencies.NewEmpty(), semPkg, preludeScope.Nest(pkg), stdlib); err != nil {
+		ctx := NewEmptyDependencies().Inject(context.Background())
```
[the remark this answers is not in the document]
affo: Now `flux.GetDependencies` returns empty dependencies if they are not injected.

Issue #1881, opened 2019-09-17
Design promql join function


jlapacik:
This issue is to track the design of `promql.join`. This join function should be based on `groupKeyJoin` as proposed [here](https://github.com/influxdata/flux/issues/1219). As `promql` is an internal package, we can design a specific join to meet the needs of the promql transpiler and later rewrite it in terms of a more expansive `groupKeyJoin`.

PR #1886, opened 2019-09-17
docs(promql): group key join for promql

jlapacik:
https://github.com/influxdata/flux/issues/1881.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/internal/promql/join.md
+++ b/stdlib/internal/promql/join.md
@@ -1,0 +1,92 @@
+### join
+
+`join` merges two input streams into a single output stream.
+Records that have the same group key and `_time` values will be joined in the output.
+The output groups will be the same as the input groups.
+
+`join` has the following properties:
+
+| Name | Type     | Description |
+| ---- | -------- | ----------- |
+| a    | object   | Table stream |
+| b    | object   | Table stream |
+| fn   | function | Defines the function that joins the rows of each table. The return value is an object which defines the output record structure. This function must preserve the table grouping. |
+
+Ex.
+```
+import "internal/promql"
+
+left = from(bucket: "my-bucket")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._field == "a")
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "value_a"})
+
+right = from(bucket: "my-bucket")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._field == "b")
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "value_b"})
+
+promql.join(a: left, b: right, fn: (a, b) =>
+    return {a with value_b: b.value_b}
+)
+```
+
+`join` assumes that each input table is sorted by `_time`.
+The planner will ensure this by inserting a `sort` procedure after each one of join's immediate predecessors.
+`join` will only call `fn` on rows that pass the join condition.
+
+Here are some examples of compiling PromQL queries into their equivalent Flux forms involving `promql.join`.
+
+```
+left_metric + right_metric
+```
+compiles to:
+```
+lhs = left_metric
+    |> rename(columns: {_value: "lv"})
+
+rhs = right_metric
+    |> rename(columns: {_value: "rv"})
+
+promql.join(a: lhs, b: rhs fn: (a, b) => ({a with rv: b.rv}))
+    |> map(fn: (r) => ({r with _value: r.lv + r.rv}))
+    |> drop(columns: ["lv", "rv"])
+```
+
+```
+left_metric + on(tagA,tagB) right_metric
+```
+compiles to:
+```
+lhs = left_metric
+    |> rename(columns: {_value: "lv"})
+    |> group(columns: ["tagA", "tagB"])
+
+rhs = right_metric
+    |> rename(columns: {_value: "rv"})
+    |> group(columns: ["tagA", "tagB"])
+
+promql.join(a: lhs, b: rhs, fn: (a, b) => ({a with rv: b.rv}))
+    |> map(fn: (r) => ({r with _value: r.lv + r.rv}))
+    |> drop(columns: ["lv", "rv"])
+```
+
+```
+left_metric + ignoring(tagA,tagB) right_metric
+```
+compiles to:
+```
+lhs = left_metric
+    |> rename(columns: {_value: "lv"})
+    |> group(columns: ["tagA", "tagB", "_time", "lv"], mode: "except")
+
+rhs = right_metric
+    |> rename(columns: {_value: "rv"})
+    |> group(columns: ["tagA", "tagB", "_time", "rv"], mode: "except")
+
+promql.join(a: lhs, b: rhs, fn: (a, b) => ({a with rv: b.rv}))
+    |> map(fn: (r) => ({r with _value: r.lv + r.rv}))
+    |> drop(columns: ["lv", "rv"])
+```


stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+    |> rename(columns: {_value: "a"})
+
+right = from(bucket: "my-bucket")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._field == "B")
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "b"})
+
+promql.join(tables: {left, right}, fn: (t) => {
+    return {t.left with b: t.right.b}
```
wolffcm: If there are common column labels for both tables, what is the behavior?  Based on how `with` behaves within `map`, I guess that columns on the RHS of `with` will overwrite any columns with the same name on the LHS.

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+
+`join` assumes that each input table is sorted by `_time`.
+The planner will ensure this by inserting a `sort` procedure after each one of join's immediate predecessors.
+`join` will only call `fn` on rows that pass the join condition.
```
wolffcm: What is the "join condition"?  Is it just that the `_time` columns match?

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+
+Here are some examples of compiling PromQL queries into their equivalent Flux forms involving `promql.join`.
+
+```
+left_metric + right_metric
```
wolffcm: Ah, PromQL syntax, so concise, so beautiful.

stdlib/internal/promql/join.md
wolffcm: Reading this it occurs to me again that having data that is "pivoted" out of the gate (i.e., no `_field` column and `_value` renamed to be the name of the field) would simplify things a lot.

stdlib/internal/promql/join.md
wolffcm: For rows that do not have a corresponding row with the same `_time` in the other table, I guess you just discard the row?  Is that what PromQL does for implicit joins?  I could imagine also doing a left/right/full outer join where all or some of the rows are retained, and using null values for missing rows.

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "b"})
+
+promql.join(tables: {left, right}, fn: (t) => {
+    return {t.left with b: t.right.b}
+}
+```
+
+`join` assumes that each input table is sorted by `_time`.
+The planner will ensure this by inserting a `sort` procedure after each one of join's immediate predecessors.
```
wolffcm: I like the idea of explicitly inserting a `sort` operation when we need the input to be sorted.  If we go this route we could also add a transformation that would remove unneeded sorts.

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+Records that have the same group key and `_time` values will be joined in the output.
+The output groups will be the same as the input groups.
+
+`join` has the following properties:
+
+| Name    | Type     | Description |
+| ------- | -------- | ----------- |
+| tables  | object   | Specifies the two table streams to join. |
+| fn      | function | Defines the function that joins the rows of each table. The return value is an object which defines the output record structure. |
+
```
wolffcm: You could also add a `predFn` parameter that would define the join predicate, and make it default to `(left, right) => left._time == right._time`.

stdlib/internal/promql/join.md
jlapacik: Yes in this case they would be overwritten.

stdlib/internal/promql/join.md
jlapacik: A row on the left will only join with a row on the right if both rows have the same group key value as well as the same `_time`  value. Everything else is discarded. And yes this is what PromQL does. PromQL doesn't support the other types of joins. Flux can support these other join types with an extra predicate parameter, but to keep things simple, for the PromQL case, I chose not to do that.

stdlib/internal/promql/join.md
jlapacik: Yes I agree, that's ultimately how I would like the Flux join function defined. However that will also require us to provide some internal join implementations that are optimized for specific scenarios for which the planner will need to be made aware. I didn't want this implementation to have the overhead of calling an anonymous function for each row combination so I left that out.

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+```
+compiles to:
+```
+lhs = left_metric
+    |> rename(columns: {_value: "lv"})
+
+rhs = right_metric
+    |> rename(columns: {_value: "rv"})
+
+promql.join(tables: {lhs, rhs}, fn: (t) => ({t.lhs with rv: t.rv}))
```
aanthony1243: so the `map` is technically not needed here as we could to the `+` in the join fn.    

However, why the join fn?   we could just have `join` output  `t` for each row.  Then a call to `map` to finish it off?  

I guess it's for efficiency, yeah?

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "a"})
+
+right = from(bucket: "my-bucket")
+    |> range(start: -1h)
+    |> filter(fn: (r) => r._field == "B")
+    |> drop(columns: ["_field"])
+    |> rename(columns: {_value: "b"})
+
+promql.join(tables: {left, right}, fn: (t) => {
```
aanthony1243: is `{left, right}`  implicit syntax for object literals?   I do like that syntax but should we be more explicit about what's happening here?  maybe name the streams `t1`  and `t2`?

jlapacik: Yes that's sugar for `{left: left, right: right}`. I can be more explicit.

stdlib/internal/promql/join.md
jlapacik: True the addition operation could be defined within the join fn. The reason for the `fn` parameter is that it allows the user to specify the exact output schema they want which is a much more declarative design. The current join handles the merging of rows internally and any column name conflicts. I have found that to be really awkward. This way the user tells us what data they want out vs `join` giving them data that they must then reshape.

stdlib/internal/promql/join.md
jsternberg: This seems more complex than I would like and would impact the current performance because functions are slow to execute in the current system.

I'm not aware of any system that requires a function to specify how the data is joined and I don't think we should start. It is far too complex and will likely make it too difficult of a function to use in my opinion.

Additionally, I see that there ends up being a bit of a fundamental problem with joining data where you don't know all of the data. If there's any conflicts for column names, you end up with data overwritten.

I think that we should only have an `on` keyword which specifies what fields we join on (such as `_time` for most use cases), we automatically join on group keys, and any other columns get prefixed the same as the current join.

stdlib/internal/promql/join.md
jsternberg: I think this goes a bit beyond our current capabilities since we cannot optimize the sort at the moment. Since this is an internal function, we can say that it assumes the table is sorted, but does not perform a sort since that is too costly of an operation.

I also don't like that this makes any assumptions about time. The columns that join will join on should be specified by the `on` keyword as was stated in the original proposal.

stdlib/internal/promql/join.md
jsternberg: Join shouldn't perform any processing and shouldn't act as yet another map. We cannot have them specify a schema as it won't work for the unknown schemas the transpiler works with.

stdlib/internal/promql/join.md
jsternberg: Although I can be pretty easily convinced to hardcode `on` as `["_time"]` temporarily and the `on` will follow once the proof of concept has happened and we generalize this to more uses.

jlapacik: @jsternberg this join design is specialized specifically for PromQL. To keep the implementation simple and efficient, it's based on groupKeyJoin and automatically joins on _time since that is how PromQL operates in all cases. This can easily be swapped out when we refactor the current flux join, but these are the assumptions that were made to keep the scope small.

stdlib/internal/promql/join.md
jlapacik: This join is much more selective than the current flux join. Therefore in the majority of cases, we will be performing relatively few function calls. 

In my opinion, join is similar to map. You could consider it a map over two streams. I really find the implicit column renaming of the current join to be very awkward. I'd rather have the user declare exactly what they want out.

And I don't want to go too off topic, but type inference can actually catch naming conflicts at compile time in this form.

jsternberg: In order to keep the scope limited, I guess the main concern I have is how would the transpiler generate the function call? If the transpiler would always generate the same function using `with`, then I don't see a purpose in the `fn` parameter.

I also am not aware of any prior work that treats join as similar to map. It will also be harder to make join efficient if it includes an implicit call to `map` and it also means that we will have another place where `map` can cause issues. What happens if I do this?

```
join(tables: {left, right}, fn: (l, r) => {host: "server{l._value}"})
```

And `host` ends up being part of the group key. It is much simpler if we declare that join will just always prefix field key names. I think the existing join behavior does some weird hybrid behavior where it only does the prefix if there's a conflict. I think the prefix should always happen.

jlapacik: I'm not sure I follow your example. We would have the same issue if a user put a similar map call after join.

I've given examples of how to transpile PromQL using this join and yes they would always generate the same type of function call. But putting this join in an internal package gives us the ability to try it without breaking the public API at a later point.

And in terms of efficiency, we can add fast internal implementations which the planner can rewrite to for common use cases, so I don't see that being an issue. Even without specialized implementations, the selective nature of groupKeyJoin makes it so that we won't be evaluating the anonymous function for very many rows in most cases.

stdlib/internal/promql/join.md
juliusv: There is an exception for the `or` (set union) and `unless` (set difference) binary operators.

`or` produces output series (Flux rows) for label sets (Flux group keys) that are present *either* on the LHS or RHS. The sample value (Flux `_value`) is taken from the LHS if the series is present there, from the RHS otherwise.

`unless` propagates anything from the LHS into the output unless there is a matching label set (Flux group key) on the RHS.

I guess this would need some type of outer join that fills up missing rows on either side with `null` values, on which one could then do conditional postprocessing to achieve those operators?

stdlib/internal/promql/join.md
```diff
@@ -0,0 +1,91 @@
...
+    |> rename(columns: {_value: "rv"})
+    |> group(columns: ["tagA", "tagB"])
+
+promql.join(tables: {lhs, rhs}, fn: (t) => ({t.lhs with rv: t.rv}))
+    |> map(fn: (r) => ({r with _value: r.lv + r.rv}))
+    |> drop(columns: ["lv", "rv"])
+```
+
+```
+left_metric + ignore(tagA,tagB) right_metric
```
juliusv: `ignore` -> `ignoring`

juliusv:
@jlapacik I noticed this got merged, but didn't see a response to https://github.com/influxdata/flux/pull/1886/files#r327665695 yet - just wanted to make sure that comment was seen. I'm also going to think more about the rest and double-check that I didn't miss something.

stdlib/internal/promql/join.md
juliusv: Just to clarify, the full transpilation of this query will require additional postprocessing, at least also dropping the metric name (`_field`) column here. In other cases (like the example below that uses `ignoring()`), the ignored tags also need to be dropped from the final result, etc.. So the `join()` will always just be a fragment (though crucial part) of what's needed for the transpilation.

juliusv:
To mirror PromQL behavior there's also the requirement to produce a join error in some cases, though so far I think we've said that we could solve them by adding specialized post-processing functions that check uniqueness constraints on the joined output.

I just want to mention these cases again so they're known:

When there are multiple series on one side of a binary operation that would match an element from the other side (based on the provided grouping / matching labels), then the user needs to explicitly say that they want to group by the extra dimensionality present on the "many" side (via group_right or group_left modifiers). Otherwise this produces an error.

* Error in the code: https://github.com/prometheus/prometheus/blob/3cffd8103509e5e423f4bb58b784ba9f0805d333/promql/engine.go#L1446-L1447
  * Example that triggers the error, since the right side has the extra "cpu" dimension and thus multiple series from the RHS match a single series on the LHS, but we don't explicitly provide a "group_right" modifier: http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=sum%20without(cpu)%20(node_cpu_seconds_total)%20%2F%20ignoring(cpu)%20group_left%20node_cpu_seconds_total&g0.tab=1
* Error in the code: https://github.com/prometheus/prometheus/blob/3cffd8103509e5e423f4bb58b784ba9f0805d333/promql/engine.go#L1493
  * Example that triggers the error, since the left side has more dimensionality, but we don't explicitly provide a "group_left" modifier: http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=node_cpu_seconds_total%20%2F%20ignoring(cpu)%20sum%20without(cpu)%20(node_cpu_seconds_total)&g0.tab=1

These errors are similar, but handled differently in the PromQL code because one of the cases (right-hand side having more dimensionality without explicit grouping instructions) can be caught earlier.

Replicating the exact error message is probably not as important, as long as we can detect and error out on these cases at all somehow.

stdlib/internal/promql/join.md
jlapacik: @juliusv I did not include transpilation examples for `or` and `unless` but they can indeed be transpiled to Flux using a combination of left and right join operations, as well as union. I can add more examples to this doc

stdlib/internal/promql/join.md
jlapacik: I forgot to drop the `_field` and ignored columns in these examples. I will update.

jlapacik:
@juliusv sorry for the late response. I didn't include transpilation examples for every PromQL binary operator, however this join design does support all transpilations. Like you mention above, `join` is just one component in the transpiled query. I will add more examples to this doc.

As for the errors you are correct, that logic was intentionally left out of `join`. Instead it should be handled after the `join` operation by a separate transformation.

stdlib/internal/promql/join.md
juliusv: Ah ok, if this can be done somehow, then great :) Are you generally taking over the transpilation work for vector-to-vector binops? (I'd be happy about that, just good to know.)

PR #1900, opened 2019-09-20
fix(stdlib): sql does not leak connections anymore

jsternberg:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/source_iterator.go
+++ b/execute/source_iterator.go
@@ -27,16 +27,16 @@ type SourceDecoder interface {
 // CreateSourceFromDecoder takes an implementation of a SourceDecoder, as well as a dataset ID and Administration type
 // and creates an execute.Source.
 func CreateSourceFromDecoder(decoder SourceDecoder, dsid DatasetID, a Administration) (Source, error) {
-	return &sourceIterator{decoder: decoder, id: dsid}, nil
+	return &sourceDecoder{decoder: decoder, id: dsid}, nil
 }
 
-type sourceIterator struct {
+type sourceDecoder struct {
 	decoder SourceDecoder
 	id      DatasetID
 	ts      []Transformation
 }
 
-func (c *sourceIterator) Do(ctx context.Context, f func(flux.Table) error) error {
+func (c *sourceDecoder) Do(ctx context.Context, f func(flux.Table) error) error {
 	err := c.decoder.Connect(ctx)
 	if err != nil {
 		return err
@@ -66,11 +66,11 @@ func (c *sourceIterator) Do(ctx context.Context, f func(flux.Table) error) error
 	return nil
 }
 
-func (c *sourceIterator) AddTransformation(t Transformation) {
+func (c *sourceDecoder) AddTransformation(t Transformation) {
 	c.ts = append(c.ts, t)
 }
 
-func (c *sourceIterator) Run(ctx context.Context) {
+func (c *sourceDecoder) Run(ctx context.Context) {
 	err := c.Do(ctx, func(tbl flux.Table) error {
 		for _, t := range c.ts {
 			err := t.Process(c.id, tbl)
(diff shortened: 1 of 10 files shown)


stdlib/sql/from.go
```diff
@@ -106,76 +109,107 @@ func (s *FromSQLProcedureSpec) Copy() plan.ProcedureSpec {
...
+	// This iterator does not maintain any internal connections.
+}
 
-	reader := *c.reader
+func (c *sqlIterator) Err() error {
+	return c.err
+}
+
+// read will use the RowReader to construct a flux.Table.
+func read(ctx context.Context, reader execute.RowReader, alloc *memory.Allocator) (flux.Table, error) {
```
aanthony1243: I think we should promote this to a public, exported function, maybe UnGroupedRead(...)?

it'd be cool to have another helper function  GroupedRead()  that takes a function parameter that can compute a group key given a row from the row reader.  This can be done in a separate PR though.

execute/source_iterator.go
```diff
@@ -85,3 +85,93 @@ func (c *sourceIterator) Run(ctx context.Context) {
 		t.Finish(c.id, err)
 	}
 }
+
+// CreateSourceFromIterator takes an implementation of a SourceIterator as well as a dataset ID
+// and creates an execute.Source.
+func CreateSourceFromIterator(iterator SourceIterator, dsid DatasetID) (Source, error) {
+	return &sourceIterator2{iterator: iterator, id: dsid}, nil
```
aanthony1243: I think it's confusing that neither `sourceIterator` nor `sourceIterator2`  implement the interface `SourceIterator`.   What if we rename `SourceIterator`  to `TableIterator`  ?    

And then we can also rename `sourceIterator`  to `decoderSource`  and `sourceIterator2` to `tableIteratorSource`.  

Long-term, we should eliminate the tech-debt and keep just the TableIterator types.

jsternberg: I was going to rename these but I just kind of forgot.

stdlib/sql/from.go
jsternberg: This was mostly meant as a helper function because the real thing I wanted was for the implementations of `RowReader` to just return tables, but they were already written and I didn't feel like refactoring multiple locations so I did this. It might be helpful to expose it though.

execute/source_iterator.go
```diff
@@ -8,14 +8,14 @@ import (
...
-// Connect implements the logic needed to connect directly to the data source.
+// connect implements the logic needed to connect directly to the data source.
 //
 // Fetch implements a single fetch of data from the source (may be called multiple times).  Should return false when
 // there is no more data to retrieve.
 //
 // Decode implements the process of marshaling the data returned by the source into a flux.Table type.
 //
-// In executing the retrieval process, Connect is called once at the onset, and subsequent calls of Fetch() and Decode()
+// In executing the retrieval process, connect is called once at the onset, and subsequent calls of Fetch() and Decode()
```
fchikwekwe: Where are the calls of `Fetch()` and `Decode()` moved to? Or are they removed in this update?

fchikwekwe: Oh, I see now. Sorry I was confused by `SourceDecoder` and `sourceDecoder` having such similar names and the renaming of those functions later in the file.

execute/source_iterator.go
jsternberg: Ah, I didn't understand this message. I think that's probably a good idea to just have this implement `TableIterator` and make a default source that consumes a `TableIterator`.

jsternberg: I modified `SourceIterator` to use the same signature but added `context.Context`. The `TableIterator` doesn't take a context, but the sources need the context so they know when to cancel.

So the interfaces are functionally the same, but slightly different because of the context parameter. It worked a lot better and was much easier to implement.

stdlib/sql/from.go
aanthony1243: yeah, it's more of a forward-thinking vision that once you implement the row reader interface, you can use this function for free.

PR #1901, opened 2019-09-21
refactor(stdlib): add max and total allocations metrics to benchmarks

jsternberg:
This adds a custom metric for the max and total allocations as recorded
by the memory allocator to the flux end to end benchmarks.

It also adds a `total_allocated` to the statistics struct which records
a counter of all allocations without accounting for released memory.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
It's to aggregate them so they can be divided for the per op division later.

PR #1910, opened 2019-09-23
[WIP] feat(stdlib/experimental/prometheus): add prometheus metrics endpoint as data source in flux

fchikwekwe:
Still a lot of changes to be made to this scraper. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/experimental/prometheus/from.go
```diff
@@ -0,0 +1,339 @@
...
+		p.nextMetric()
+		return true, nil
+	}
+	// p.met = p.metrics[0]
+	return false, nil
+}
+
+func (p *PrometheusIterator) nextMetric() error {
+	fmt.Println(p.i)
+	next := p.metrics[p.i]
```
aanthony1243: this seems like a lot of unnecessary copying.  Not sure if this is accountable for the slowdown, but just advance `p.i` here (or just drop this fn and advance `i` on line 220. )

then, in `Decode` just use `p.metrics[p.i]` in the few places it's needed

stdlib/experimental/prometheus/from.go
```diff
@@ -0,0 +1,339 @@
...
+
+	return nil
+}
+
+func (p *PrometheusIterator) Decode(ctx context.Context) (table flux.Table, err error) {
+
+	// unpacking Fields
+	var fieldVal string
+	var val interface{}
+	for f, v := range p.met.Fields {
```
aanthony1243: the outcome of this loop is  fieldVal, val := p.met.Fields[len(p.met.Fields) - 1] .   Is that what's intended?  Are there multiple Fields to decode here?

stdlib/experimental/prometheus/from.go
```diff
@@ -0,0 +1,339 @@
...
+	p.metrics = make([]Metric, 0)
+
+	// read metrics
+	for name, family := range metricFamilies {
+		for _, m := range family.Metric {
+			// read tags
+			tags := makeLabels(m)
+
+			// read fields
+			fields := make(map[string]interface{})
```
aanthony1243: this assignment is unnecessary.  if you want to create this variable and don't care about its initial value, use `var fields map[string]interface{}`

stdlib/experimental/prometheus/from.go
fchikwekwe: Is the assignment the thing that is slow here? Or is it indexing the slice?

fchikwekwe: I thought both of those should be O(1). Sorry if your point went over my head here.

stdlib/experimental/prometheus/from.go
fchikwekwe: Yeah, I was trying to figure out how to unpack the map values when I don't have the key. This was a hack-y way to unpack those values.

fchikwekwe: Since keys in this map are strings, I don't think we can use an int to help with unpacking it.

stdlib/experimental/prometheus/from.go
```diff
@@ -0,0 +1,314 @@
...
+				p.metrics = append(p.metrics, met)
+			}
+		}
+	}
+	return nil
+}
+
+func (p *PrometheusIterator) Fetch(ctx context.Context) (bool, error) {
+	p.url = "http://demo.robustperception.io:9090/metrics"
+	resp, err := http.Get(p.url)
```
jsternberg: This might be causing an infinite loop. `Fetch` is called repeatedly until it returns `false` and I'm not convinced this will ever return `false`.

It might be best to rewrite this to use the new `SourceIterator` which has an easier interface to implement.

stdlib/experimental/prometheus/prometheus.flux
```diff
@@ -0,0 +1,3 @@
+package prometheus
+
+builtin from
```
jsternberg: Minor nitpick. I think we should call this something else because we are not retrieving data from prometheus which is something we would likely want to support. This is acting as a prometheus metrics scraper so I think it needs a different name to signify that difference.

stdlib/experimental/prometheus/from.go
fchikwekwe: I haven't been experiencing an infinite loop here 🤔 There's a `return false, nil` at the bottom of this method when the conditional is satisfied. But you're right that I should update this to the new `SourceIterator`. I'll pull that in and see if helps with the speed.

stdlib/experimental/prometheus/prometheus.flux
fchikwekwe: Yeah I just changed this to scrape.

stdlib/experimental/prometheus/from.go
jsternberg: Ah! Now I understand what's going on. The reason this is slow is because you're making n http requests instead of making a single request and buffering the result while you read through it.

It would definitely be much easier to implement `SourceIterator` as it would make the code a lot more straightforward and less confusing.

fchikwekwe: Okay, yeah that makes sense. I will try and implement `SourceIterator` here.

fchikwekwe: Thanks Jonathan! I haven't implemented `SourceIterator` (I will try and do that before I open a real PR), but even separating the http.Get from the rest of the fetch logic made it much faster.

promql/aggregate_expr.go
```diff
@@ -17,7 +17,8 @@ const (
 )
 
 type aggregateFn struct {
-	name      string
+	name string
+	// packageName string
```
wolffcm: I guess this is left over from other work, and you can remove it.

stdlib/experimental/prometheus/scrape.go
```diff
@@ -0,0 +1,334 @@
...
+				p.metrics = append(p.metrics, met)
+			}
+		}
+	}
+	return nil
+}
+
+func (p *PrometheusIterator) Fetch(ctx context.Context) (bool, error) {
+	// Iterate over all Metrics in the slice
+	p.i++
```
wolffcm: Is there an off-by-one issue here?  It seems like the metric in `p.metrics[0]` will never be returned.

stdlib/experimental/prometheus/scrape.go
```diff
@@ -0,0 +1,334 @@
...
+	// Unpacking Fields map
+	var val interface{}
+	for _, v := range p.met.Fields {
+		val = v
+	}
+
+	groupKey := execute.NewGroupKeyBuilder(nil)
+	groupKey.AddKeyValue("_measurement", values.New("prometheus"))
+	groupKey.AddKeyValue("_field", values.New(p.met.Name))
+
```
wolffcm: At least in the metrics that we get when telegraf scrapes `/metrics` endpoints in acceptance and production, we usually make the `_measurement` be the prometheus metric name (like `http_query_request_bytes`), and `_field` is the metric type (like `counter` or `gauge`).

Probably we want to do the same thing?

stdlib/experimental/prometheus/scrape.go
```diff
@@ -0,0 +1,334 @@
...
+	builder.AddCol(flux.ColMeta{
+		Label: "_field", // promethus metric name
+		Type:  flux.TString,
+	})
+	builder.AddCol(flux.ColMeta{
+		Label: "url",
+		Type:  flux.TString,
+	})
+
+	builder.AppendValue(0, values.New(p.met.Timestamp.String()))
```
wolffcm: You are appending a string to a timestamp column here.  Unless I'm mistaken I think you want to use `values.ConvertTime` instead?

stdlib/experimental/prometheus/scrape.go
wolffcm: Oh, I like the idea of adding `url` as a column to the output, probably we want to put that in the group key as well?

stdlib/experimental/prometheus/scrape.go
```diff
@@ -0,0 +1,334 @@
...
+			}
+
+		}
+	} else if m.Untyped != nil {
+		if !math.IsNaN(m.GetUntyped().GetValue()) {
+			fields["value"] = float64(m.GetUntyped().GetValue())
+		}
+	}
+	return fields
+}
```
wolffcm: It would be nice to see some unit tests here, I think you could use the `httptest` package to mock a server with a `metrics` endpoint and verify that the source produces the expected `executetest.Table` objects.

stdlib/experimental/prometheus/scrape.go
wolffcm: Also I think Adam alluded to this in one of his earlier comments, but you could just store `p.i` and `p.metrics` as state, and then index into the slice via `p.metrics[p.i]` in the `Decode` method.  Then you would avoid having to copy a `Metrics` object every time you decode.

stdlib/experimental/prometheus/scrape.go
fchikwekwe: So I spoke with Daniel Nelson recently and he told that telegraf has changed their prometheus scraper to omit the `metric type`  put **prometheus** as the measurement column. I originally had it the way you suggested to match the chronograf example that you showed me.

promql/aggregate_expr.go
fchikwekwe: Oh, I thought I deleted that. 🤦🏾‍♀️

stdlib/experimental/prometheus/scrape.go
fchikwekwe: Yes, test writing is my next step. That's a good suggestion. I was thinking about how to test the metrics endpoint consistently.

stdlib/experimental/prometheus/scrape.go
wolffcm: Ok, that's great.  Thanks for clarifying this.

stdlib/experimental/prometheus/from.go
fchikwekwe: Nevermind. Chris explained the issue here. I'll go ahead and fix it 👍🏾

stdlib/experimental/prometheus/scrape.go
fchikwekwe: Thanks for pointing out the off by one! And yeah, thanks for helping me understand the copying issue. Fixing.

Issue #1904, opened 2019-09-23
Investigating accumulating vs discarding mode


jsternberg:
The flux datasets have a concept of accumulating and discarding mode. At the moment, the tables run in accumulating mode and there is no way to customize that.

This may cause problems. I am not aware of any particular reason why we would operate in accumulating mode at the moment, but it means that the memory allocated for the constructed tables from a transformation do not get released until the query is over. While this may not matter much with the current allocator because the current allocations would only be released to the Go garbage collector, it does mean that the memory is unable to be garbage collected during execution and it is impossible for us to optimize memory usage with pooled allocations because the memory is never released to be reused by the same query.

Investigate whether everything works if we switched to operating in discarding mode. Additionally, investigate if this improves anything. I think it is unlikely to help anything without other work being done, but I am curious if there's any initial improvements we get from this.

jsternberg:
I'm not seeing any difference between these, but I realized there might be a reason. I think that the table implementations discard themselves after they are invoked to release the memory. This has never come up before because we don't reuse the tables.

I'm going to do a small experiment with making the aggregating mode work as it was aggregating by not discarding the table and see what I get.

jsternberg:
I think I've confirmed the above. I added this code to `execute/dataset.go`:

```
diff --git a/execute/dataset.go b/execute/dataset.go
index 031c6030..34032647 100644
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -151,7 +151,11 @@ func (d *dataset) triggerTable(key flux.GroupKey) error {
 		}
 		fallthrough
 	case AccumulatingMode:
-		if err := d.processTable(b); err != nil {
+		b, err := CopyTable(b)
+		if err != nil {
+			return err
+		}
+		if err := d.processTable(b.Copy()); err != nil {
 			return err
 		}
 	}
```

The max allocated increased for each benchmark by quite a bit and the total allocated didn't change at all. I had assumed the total allocated wouldn't change, but I was curious in the first attempt why max allocated hadn't changed. Now I know. The reason is because accumulating mode doesn't actually accumulate anything.

PR #1912, opened 2019-09-23
feat(execute): switch to using discarding mode for the transformations

jsternberg:
This switches to using discarding mode with the transformations. The
intention of discarding mode was that it would discard the data after it
had been processed. For the current query engine, we do not have any
other stream primitives that would cause us to need the data after it
had been completed because we have no other triggers that actually
function.

But this hadn't caused a problem mostly because accumulating mode, the
previous method that was used, didn't actually accumulate anything.
Since tables will automatically release any memory they are using after
being processed, accumulating mode was essentially the same as
discarding mode, but didn't bother to remove anything from the group
lookup cache.

This change makes it clear that accumulating mode does not function as
it is stated to function and switches to using discarding mode because
it is the only method we actually support anyway.

#1904

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -2,6 +2,8 @@ package execute
 
 import (
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/plan"
 	uuid "github.com/satori/go.uuid"
 )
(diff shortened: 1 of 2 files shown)


jsternberg:
Because accumulating mode was, as described above, not actually working, there aren't any actual performance gains gotten from this. There may even be some performance losses because now we actually delete stuff from the cache. I didn't see anything significant on the benchmark comparison though so I think this is likely safe to merge.

aanthony1243:
all that being said, I see what's happening here and don't see any problems to fix.

PR #1917, opened 2019-09-24
refactor(internal/mutable): create mutable builders for all array types

jsternberg:
This creates mutable builders for all of the array types that we use.
This also adds a common interface for all of the mutable builders.

It adds a boolean and string array. The boolean array is complete, but
the string array does not implement the contract correctly. It does not
use the allocator for the mutable memory and will instead construct the
array using an untracked slice and only consume the memory once it
constructs an array.

The intention of this is that we would be able to remove our custom
allocator code and use only the allocator interface. This would allow us
to create different allocator implementations instead of being tied to
the single `*memory.Allocator` struct.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/mutable/array.go
+++ b/internal/mutable/array.go
@@ -1,0 +1,78 @@
+package mutable
+
+import (
+	"sync/atomic"
+
+	"github.com/apache/arrow/go/arrow/array"
+	"github.com/apache/arrow/go/arrow/memory"
+)
+
+// Array specifies the common interface for all mutable arrays.
+type Array interface {
+	// AppendNull will append a null value to the array.
+	AppendNull()
+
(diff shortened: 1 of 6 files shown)


jsternberg:
This is something I had in my local working directory that I'm hoping will be helpful. I don't have the exact plan for what we plan to do with the allocator, but I think it will be more flexible if we can remove our reliance on the custom allocator interface that is used by the `ColListTableBuilder`. This moves us in a direction that allows for that.

At the minimum, the boolean builder exposes a bitmap that can be used by transformations and would be memory that is tracked. This would be particularly useful for doing memory optimizations on things like filter.

PR #1919, opened 2019-09-25
feat(promql): group key join

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,561 @@
...
+		// The record obtained from calling fn may be added to output
+		if err := appendRowToBuilder(builder, obj); err != nil {
+			return nil, err
+		}
+		idxB, readerB = recordsB.next()
+		goto NEXT
+	}
+DONE:
+	return builder.Table()
+}
```
wolffcm: Kind of a corner case I'm curious about here: what if you had two tables A and B with the same group key and a bunch of rows in both sides with the same value in the `_time` column?  (I know that storage cannot produce such data, but it can exist in Flux.) . In that case I would expect a cross join for that subset of rows.  

E.g, if you had 10 rows in `A` with `_time=1000` and another 10 such rows in `B`, then there would be 100 corresponding rows in the join output.

It's unlikely to come up in practice very often now, but it would be good to have a test for it.  It will become more frequent if the join predicate were expressed as an arbitrary function passed in by the user.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,561 @@
...
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+const joinKind = "promJoin"
+
+func init() {
+	signature := semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"a": flux.TableObjectType,
```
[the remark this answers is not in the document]
jlapacik: I would like to keep the signature as minimal as possible. Hence the one-letter parameter names. Is that confusing? My thinking is that `join` is the main descriptor of the operation, so let's try to keep the parameter names as short as possible.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,561 @@
...
+			entry.(*cacheEntry).a = tbl
+		case c.b:
+			entry.(*cacheEntry).b = tbl
+		}
+	} else {
+		switch id {
+		case c.a:
+			c.data.Set(tbl.Key(), &cacheEntry{a: tbl})
+		case c.b:
+			c.data.Set(tbl.Key(), &cacheEntry{b: tbl})
```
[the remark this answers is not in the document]
jlapacik: This is for the case where there is no cache entry for this group key. This means that this is the first table associated with the key. If an entry has already been added to the cache, we just update that entry.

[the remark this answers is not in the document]
wolffcm: Yeah, of course.  I was confused.

stdlib/internal/promql/join.go
jlapacik: Thanks for catching this. I totally overlooked it.

stdlib/internal/promql/join.go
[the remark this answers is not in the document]
wolffcm: I am thinking forward a bit if we ever want to expression concepts from SQL like `left outer join`, etc.  Using "left" and "right" makes more sense to me coming from that background.  I don't really feel all the strongly about it though.

stdlib/internal/promql/join.go
```diff
@@ -326,84 +326,74 @@ func (c *mergeJoinCache) joinTables(a, b flux.Table) (flux.Table, error) {
 	ar := make([]flux.ColReader, 0)
 	br := make([]flux.ColReader, 0)
 
+	// Release column readers after joining
+	defer a.Done()
+	defer b.Done()
+
 	// Retain column readers for joining
 	if err := a.Do(func(cr flux.ColReader) error {
 		cr.Retain()
```
wolffcm: This is basically retaining all the column readers until this method returns, right?  Which kind of defeats the purpose of batching.  I think it's fine for now, but at some point we won't want to keep the entirety of both tables in memory.

stdlib/internal/promql/join.go
```diff
@@ -464,7 +464,7 @@ func appendRowToBuilder(builder *execute.ColListTableBuilder, obj values.Object)
 	obj.Range(func(name string, v values.Value) {
 		idx := execute.ColIdx(name, builder.Cols())
 		if idx < 0 {
-			err = fmt.Errorf("column %s not found", name)
+			err = errors.Newf(codes.NotFound, "column %s not found", name)
```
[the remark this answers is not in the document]
jsternberg: I've been using `FailedPrecondition` for it. It implies a user error, but it indicates that the data may be the cause of the problem because the section is dynamic which seemed to fit into `FailedPrecondition`. InfluxDB translates both of them to `Invalid` though so it doesn't matter much, but it would lend consistency with my other PR.

stdlib/internal/promql/join.go
jsternberg: If the purpose of this is just to retain everything, then `execute.CopyTable` would be more idiomatic.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+const joinKind = "promJoin"
```
jsternberg: Can you name this with the full pathname? We're going to run into this issue more often and we should set a standard for this. Something like `internal/promql.join`?

stdlib/internal/promql/join.go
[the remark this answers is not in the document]
jlapacik: I didn't know this would result in a 404. I'll change the codes to internal.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+	return joinKind
+}
+func (s *MergeJoinProcedureSpec) Copy() plan.ProcedureSpec {
+	return &MergeJoinProcedureSpec{Fn: s.Fn.Copy()}
+}
+
+func createMergeJoinTransformation(id execute.DatasetID, mode execute.AccumulationMode, spec plan.ProcedureSpec, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
+	s, ok := spec.(*MergeJoinProcedureSpec)
+	if !ok {
+		return nil, nil, errors.Newf(codes.Invalid, "invalid spec type %T", spec)
```
jsternberg: This should be an internal error. There's nothing the user can do about this and it indicates a logic error.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+	rowA := make(map[string]values.Value, 8)
+	rowB := make(map[string]values.Value, 8)
+
+	_, ra, ta := iterA.next()
+	i, rb, tb := iterB.next()
+
+	var firstRow = true
+
+NEXT:
+	if ra == -1 || rb == -1 {
```
jsternberg: Can you encapsulate this section where it finds the same time into a function? Maybe into an `advance()` function and then document it to say "advance to the next row to be joined".

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+		goto NEXT
+	}
+
+	iterA.row(rowA)
+
+	for r, t := rb, tb; r != -1 && t == ta; _, r, t = iterB.next() {
+		iterB.row(rowB)
+
+		// Evaluate fn over both input rows
+		obj, err := c.fn.Eval(
```
jsternberg: Can you move this to an `evalRow` function?

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+			values.NewObjectWithValues(rowA),
+			values.NewObjectWithValues(rowB),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		// Build schema if this is the first row being joined
+		if firstRow {
+			schema := make([]flux.ColMeta, 0, obj.Len())
```
jsternberg: Make this a `buildSchema` function.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,553 @@
...
+	return -1, -1, -1
+}
+
+func (iter *colReaderIterator) reset(i, r int) {
+	iter.i = i
+	iter.r = r
+}
+
+func (iter *colReaderIterator) row(m map[string]values.Value) {
+	for k := range m {
```
jsternberg: I don't think this should ever be needed. The tables will always have a consistent schema so you are always going to have the same values that you are filling in.

jlapacik: I don't think this is true at the moment. For instance, if `fn` returns two rows of different types (perhaps by an if-then expression), the type system currently cannot catch that. I'll add a comment so that it's clear.

jsternberg: Ah I didn't think of that. Never mind then.

stdlib/internal/promql/join.go
jlapacik: Isn't this equivalent to `obj, err := evalRow(...)`?

stdlib/internal/promql/join.go
jsternberg: Ah, I think I see why `CopyTable` doesn't work. @jlapacik can you add to the `BufferedTable` interface the following?

```go
Buffer(i int) ColReader
NumBuffers() int
```

That way we can reuse that code and it would be more useful.

stdlib/internal/promql/join.go
jsternberg: It wouldn't be very complex, but it would make it more explicit and allow you to pass `rowA` and `rowB` directly rather than the multiple lines needed for this invocation. I just found the function a bit difficult to read and I thought it would help to split it up into the various components.

stdlib/internal/promql/join.go
jlapacik: @wolffcm @jsternberg the only time these column readers are not present in memory is when they are coming directly from storage. Otherwise they are already in memory and retaining them only increases the reference count.

@jsternberg we would need to add these methods to the actual column reader interface would we not? Otherwise I would have to call `CopyTable` first, which again increases all the reference counts simultaneously.

jsternberg: The `CopyTable` function returns a `flux.BufferedTable` interface. It's reasonably safe to update that one. If you want to avoid that for this PR so it's not mixed in with a "breaking change", I'm fine with the current implementation as long as a `TODO: update to use buffered table when buffered table exposes the buffers` or some other random piece of text.

stdlib/internal/promql/join.go
jlapacik: Can you provide a bit more explanation?

jsternberg: It seems to me like the purpose of this is to determine which index they have a matching time. It appears that this is implementing an inner join only, but I wasn't clear about that.

What I mean is that instead of using `NEXT` and a goto, this can be changed into a function called `advance()` that will call `next()` on each of the iterators until it finds a match. The inside implementation of that can use goto. I think goto is kind of awesome for this kind of thing. I just think it should be isolated to a single function so I know where it starts and where it ends.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,561 @@
...
+func NewMergeJoinTransformation(d execute.Dataset, cache *mergeJoinCache) *mergeJoinTransformation {
+	return &mergeJoinTransformation{
+		d:     d,
+		cache: cache,
+	}
+}
+
+func (t *mergeJoinTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	t.mu.Lock()
+	t.cache.insert(id, tbl)
```
jsternberg: This won't work. You must call `Do` on this table if you plan on wanting to get more tables.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	return -1, -1, -1
+}
+
+func (iter *colReaderIterator) reset(i, r int) {
+	iter.i = i
+	iter.r = r
+}
+
+func (iter *colReaderIterator) row(m map[string]values.Value) {
```
jsternberg: I see that the next and reset logic are required mostly because of this method. It might be easier to have this iterator work a bit less like an iterator and more like an abstraction on top of the column readers for finding the correct row.

You can modify this method to:

```go
func (iter *colReaderIterator) row(i int, m map[string]values.Value)
```

You can then have it so it uses the logic to find the correct column reader and offset. This can work reasonably efficiently if you keep track of the current chunk that you are in.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,554 @@
...
+	var readers []flux.ColReader
+	if err := tbl.Do(func(cr flux.ColReader) error {
+		cr.Retain()
+		readers = append(readers, cr)
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	timeCol := execute.ColIdx(execute.DefaultTimeColLabel, tbl.Cols())
```
jsternberg: This can be checked before retaining the readers.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,554 @@
...
+	if timeCol == -1 {
+		return errors.New(codes.Invalid, "no _time column found")
+	}
+
+	t.cache.insert(id, tbl.Key(), NewRowIterator(readers, timeCol))
+	return nil
+}
+
+func (t *mergeJoinTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey) error {
+	panic("not implemented")
```
jsternberg: Can you switch this to `return errors.New(codes.Unimplemented)`?

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,554 @@
...
+
+	return t.d.UpdateProcessingTime(pt)
+}
+
+func (t *mergeJoinTransformation) Finish(id execute.DatasetID, err error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if err != nil {
+		t.d.Finish(err)
```
jsternberg: Can you switch these two if statements to be:

```go
if err != nil || t.done {
    ...
}
```

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,554 @@
...
+		i++
+		goto NEXT
+	}
+
+	if ta > tb {
+		j++
+		goto NEXT
+	}
+
+	for k := j; ta == b.time(k); k++ {
```
jsternberg: What is this for loop doing? Can you add a comment to clarify? Specifically, it seems to me like the area above this has already verified that `ta == tb`. What additional logic does this add? I think what it is doing is allowing for the left side to be joined with multiple rows on the right side, but then I wonder why isn't the reverse possible. Shouldn't we be checking to see that if there are multiple times on the right side that join with times on the left side that we join those? Or do we want that behavior at all? Maybe we want to say that a join will not duplicate any rows.

I'd like a comment on this to clarify the intended behavior.

stdlib/internal/promql/join.go
```diff
@@ -0,0 +1,554 @@
...
+	row map[string]values.Value
+
+	timeCol int
+	offsets []int
+	readers []flux.ColReader
+}
+
+// columns returns the columns of the iterator
+func (iter *RowIterator) columns() []flux.ColMeta {
+	return iter.readers[0].Cols()
```
jsternberg: Can you get these columns from the table instead and passed in explicitly? Indexing into the reader at index 0 makes me nervous. I'm not sure if there's a guarantee that `Do` will always return at least one table. There is certainly no test that mandates it.

Issue #1897, opened 2019-09-20
All errors in flux repo should be `flux` errors


aanthony1243:
Uses of Errors.New  and fmt.Errorf  should be replaced with flux.Error  types.

jsternberg:
The standard way to convert errors is to use the `github.com/influxdata/flux/internal/errors` package which contains the utilities for this.

If you see the `errors.New` function being used from `import "errors"`, then the standard way to convert it is:

```go
- errors.New("my message")
+ errors.New(codes.Internal, "my message")
```

The code will not always be internal. If the message is meant for users, then `codes.Invalid` is usually more appropriate. You can look in the `codes` package at the root of the Flux repository for all of the available codes. An `Internal` code will get filtered the same way that the standard library codes will be filtered.

If you see `fmt.Errorf` then you would do:

```go
- fmt.Errorf("my message: %s", "hello")
+ errors.Newf(codes.Internal, "my message: %s", "hello")
```

Same as the above, `Internal` is only an example. If you are unsure of the code, then `Internal` will prevent the message from showing up to end users and is considered safe. But, be aware that an internal message is intended only for internal use and will not be seen by users. It corresponds to a 500 http error while invalid corresponds to 400.

I'm going to take on doing this for the `stdlib/universe` package.

aanthony1243:
@nathanielc  I'm not sure how this is a `compute` job, as it involves modifying and annotating large portions of the flux runtime code.

nathanielc:
@aanthony1243 Agreed, must have gotten lumped in with some other issues.

lyondhill:
A patter is established for this kind of work here: https://github.com/influxdata/influxdb/blob/master/task_errors.go

PR #1920, opened 2019-09-26
fix(stdlib/universe): update the universe package to use flux errors throughout

jsternberg:
The universe package now only uses annotated flux errors throughtout the
entire package. A utility has also been included in the `internal/cmd`
directory to check a package to ensure it does not use `errors.New` or
`fmt.Errorf`.

#1897

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	go.uber.org/zap v1.9.1
 	golang.org/x/exp v0.0.0-20190510132918-efd6b22b2522
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
+	golang.org/x/tools v0.0.0-20190628153133-6cdbf07be9d0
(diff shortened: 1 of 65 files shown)


jsternberg:
It's not really possible to pass in a different errors package mostly because they don't have the same signatures. The utility was just something I used myself and I'm not adding it to CI at the moment or have any plans for that. I just thought it might be useful.

That errors package is also in an internal package so we can't leak the usage of that errors package to anybody else. We can't infect somebody else's code. I thought this might not work for influxdb because it would have to create flux errors, but this isn't how it works. Any flux code in influxdb should use influxdb errors.

I don't think `codes.Critical` is necessary because I don't think it's different from `codes.Internal`.

We can talk about this more later though.

PR #1921, opened 2019-09-26
fix(csv): improve csv error message for serialized Flux error

wolffcm:


csv/result.go
```diff
@@ -368,7 +384,9 @@ func readMetadata(r *csv.Reader, c ResultDecoderConfig, extraLine []string) (tab
 				}
 				return tableMetadata{}, errors.Wrap(err, codes.Inherit, "failed to read error value")
 			}
-			return tableMetadata{}, stderrors.New(line[1])
+			// TODO: We should determine the correct error code here:
+			//   https://github.com/influxdata/flux/issues/1916
+			return tableMetadata{}, &serializedFluxError{err: errors.New(codes.Internal, line[1])}
```
jsternberg: So that I understand correctly, this is intended as a temporary hack just so we can identify this single line of code so we don't rush a serialization change through the csv encoder? If that's accurate, then this is approved.

wolffcm: Yes, that's right.

PR #1922, opened 2019-09-26
feat(memory): add a memory manager to the memory allocator

jsternberg:
This adds a `memory.Manager` interface that will allow the
`memory.Allocator` to request more memory when more is needed. If this
value is not set, the previous behavior of failing instantaneously will
continue.

This is needed for influxdata/influxdb#15265.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -133,9 +133,9 @@ func CreateAllocatingFromSource(spec plan.ProcedureSpec, id execute.DatasetID, a
 }
 
 func (s *AllocatingFromProcedureSpec) Run(ctx context.Context) {
-	if err := s.alloc.Allocate(s.ByteCount); err != nil {
-		panic(err)
-	}
+	// Allocate the amount of memory as specified in the byte count.
+	// This memory is not used or returned.
+	_ = s.alloc.Allocate(s.ByteCount)
 }
 
 func (s *AllocatingFromProcedureSpec) AddTransformation(t execute.Transformation) {
--- a/memory/allocator.go
+++ b/memory/allocator.go
@@ -2,6 +2,7 @@ package memory
 
 import (
 	"fmt"
+	"sync"
 	"sync/atomic"
 
 	"github.com/apache/arrow/go/arrow/memory"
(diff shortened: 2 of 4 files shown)


memory/allocator.go
```diff
@@ -170,6 +213,28 @@ func (a *Allocator) allocator() memory.Allocator {
...
+	// it will return an error with the details.
+	RequestMemory(want int64) (got int64, err error)
+
+	// FreeMemory will inform the memory manager that this memory
+	// is not being used anymore.
+	// It is not required for this to be called similar to how
+	// it is not necessary for a program to free the memory.
+	// It is the responsibility of the controller to reclaim the
+	// unfreed memory when a resource is dead.
+	FreeMemory(bytes int64)
```
nathanielc: This part seems a bit odd. The Manager is not aware of the controller, so how can it delegate freeing to the controller?

Should we update this interface with a `Release` method to make this more explicit?

jsternberg: I can change the wording. The intention was to say that it is the responsibility of the memory manager itself to reclaim its memory. A release method wouldn't really help as I would be unlikely to implement it.

The original reason I put this here was in case it ever did get used, I wouldn't make a breaking interface change. And it's not a problem if the underlying implementation doesn't implement it because memory is managed by the underlying resource manager anyway.

I don't think release would have a meaning for this interface.

jsternberg: Updated the text to remove mention of the controller and to make clear that this is purely an optional method. It is the responsibility of the manager to identify when the resource is dead and to reclaim the memory itself similar to how an OS would reclaim memory from a dead program.

execute/executetest/source.go
```diff
@@ -133,9 +133,7 @@ func CreateAllocatingFromSource(spec plan.ProcedureSpec, id execute.DatasetID, a
 }
 
 func (s *AllocatingFromProcedureSpec) Run(ctx context.Context) {
-	if err := s.alloc.Allocate(s.ByteCount); err != nil {
-		panic(err)
-	}
+	_ = s.alloc.Allocate(s.ByteCount)
```
nathanielc: Why are we ignoring the error now?

Can you add a comment explaining why its safe to do so?

jsternberg: It's mostly that `Allocate` doesn't return an error anymore. That got changed in another PR so this was just incorrect and I found it. I changed the interface to match arrow's so this returns a `[]byte` and it panics if there was an error.

nathanielc: Ok, why do we not need the returned `[]byte`?

jsternberg: There's nowhere for it to be returned to. This is just a special source for testing that the query panics and recovers when you overallocate.

nathanielc: ok, that makes sense, can you add that as a comment?

jsternberg: It's documented here:

```
// AllocatingFromProcedureSpec is a procedure spec AND an execution node
// that allocates ByteCount bytes during execution.
type AllocatingFromProcedureSpec struct {
    ByteCount int

    alloc *memory.Allocator
    ts    []execute.Transformation
}
```

nathanielc: That comment doesn't really explain this 

>There's nowhere for it to be returned to. This is just a special source for testing that the query panics and recovers when you overallocate.

I am probably missing some context that makes this obvious, but since I am missing it its likely others are too, as such I'd like to see a comment added explaining why it is safe to ignore the returned value.

jsternberg: Updated.

PR #1944, opened 2019-10-07
feat(rust/parser): demonstrate building Rust from Go

nathanielc:
DO NOT MERGE this is just an example of how the pieces fit together.

The Go binary can be built and run using:
```
go generate ./internal/rust/...
go run ./internal/rust/ <path to Flux file>
```

### DIFF
--- a/internal/rust/parser/parser.go
+++ b/internal/rust/parser/parser.go
@@ -1,0 +1,19 @@
+package parser
+
+//go:generate cargo build --release
+
+// #cgo LDFLAGS: -L${SRCDIR}/target/release -ldl -Wl,-Bstatic -lflux_parser -Wl,-Bdynamic
+// #include <stdlib.h>
+// void flux_parse_json(const char*);
+import "C"
+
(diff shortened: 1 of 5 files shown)


internal/rust/parser/parser.go
```diff
@@ -0,0 +1,19 @@
+package parser
+
+//go:generate cargo build --release
```
jsternberg: This won't work because of the permissions on the directory when this is included as a library. But, we may be able to do this. The key would be wrapping this command with either a script or a go binary that ensures the current directory is writable. If the current directory is not writable, it would attempt to make it writable so that we could write out the produced binary.

But I'm not sure if that will influence the hash that gets produced for `go.sum` since I am not sure how that is calculated.

Alternatively, we can choose a place in `/usr/local` or choose a location that is writable in the go cache and use that. I'll give that a try since it might be needed for the ARM builds if we choose to only include `x86_64` precompiled binaries in the module.

PR #1949, opened 2019-10-08
fix(monitor): set defaults for data properly

russorat:
The `monitor.check` function expects a `data` object with specific fields. this sets the default for the function from an empty map to a map with the expected fields.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
@russorat Is the expectation that each property of the object will be defaulted? If so then this is incorrect. Flux will default the entire object as a whole, but individual fields will not be defaulted.

At a higher level what does this solve?

russorat:
My assumption was that since there is a default value for the `data` param in the `monitor.check` function, that it was optional. but leaving it out causes the following:

image

So either the function should be updated to not throw an error or the default should be updated to provide a map with the expected fields.

nathanielc:
@russorat I expect the `data` argument to be required. Why is it safe to call `check` without a `data` parameter?

In other words that error is correct because it is invalid to call `check` without a `data` parameter. The error message could be better, we are actively working on that.

russorat:
ok makes sense. so can we remove the default for the data param completely?

nathanielc:
@russorat yes, that is the correct solution here, and it will provide a more clear error message that the parameter was missing.

PR #1966, opened 2019-10-10
fix(stdlib): median column test input is now compatible with influxdb

jsternberg:
influxdata/influxdb#13909


jsternberg:
Going to include this as part of a larger refactor to move the test cases out of `testdata` and into their appropriate packages. It is also part of a refactor to ensure that the tests are wrapped in subtests by package so there are no name conflicts.

PR #1967, opened 2019-10-10
feat(internal/cmd/gorust): create a tool that measures performance of calling Rust from Go

wolffcm:


nathanielc:
Does it make sense to add this as a Go benchmark test?

wolffcm:
> Does it make sense to add this as a Go benchmark test?

Yes, I'll change this to be a benchmark instead.

This doesn't pass in CI because the step to build the Rust library is missing, I'll see if there's a way to do that.

wolffcm:
@nathanielc I'm getting build failures in CI:
```
cargo:warning=/usr/include/features.h:364:12: fatal error: 'sys/cdefs.h' file not found
```
I guess that we don't build the Rust parser in CI at all yet.  The Docker image used is `nathanielc/flux-build:latest`, it would need to be updated with a C library.  There could also be issues with the build cache involved.  Is it worth adding this to the Flux repo now, or should we save this branch to merge it when our Rust build chain plan is more complete?

nathanielc:
Lets wait to merge once we figure out the build bits.

As for updating the `nathanielc/flux-build` image, we should move that off of my account :) The Dockerfile for it is in the flux repo, but its built under my account and its a bit awkward since it requires that you merge changes to the Dockerfile before you can leverage them in PRs.

libflux/go/libflux/bench_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func ParseAndDeserializeFB(fluxFile string) error {
+	_ = libflux.ParseIntoFbs(fluxFile)
+	return nil
+}
+
+
+func GoParse(fluxFile string) error {
+	_ = parser.ParseSource(fluxFile)
```
jsternberg: This might not test the go parser. I think this uses the rust parser when `libflux` is used as a tag. I think you can fix this by switching the import to `internal/parser` instead.

libflux/go/libflux/bench_test.go
```diff
@@ -0,0 +1,111 @@
...
+func ParseReturnHandle(fluxFile string) error {
+	 p := libflux.Parse(fluxFile)
+	 p.Free()
+	return nil
+}
+
+func ParseReturnJSON(fluxFile string) error {
+	p := libflux.Parse(fluxFile)
+	defer p.Free()
+	_, err := p.MarshalJSON()
```
jsternberg: Can you change this to:

```go
if _, err := p.MarshalJSON(); err != nil {
     return err
}
```

Alternatively, you can do:

```go
_, err := p.MarshalJSON()
return err
```

libflux/go/libflux/bench_test.go
wolffcm: Yes, you were exactly right about this.

PR #1974, opened 2019-10-11
refactor(stdlib): move test files into their appropriate packages and group them together

jsternberg:
This groups the test files for packages into their own appropriate
packages so they are closer to the source of what they are testing and
it also ensures that the package path is included as part of the
package. The test package is also now constructed so all of the tests
are included in the same package and they are, instead, broken up when
preparing the test case to be run. This way, they can be grouped
together in the same subtest and avoid any name conflicts.


internal/cmd/builtin/cmd/generate.go
```diff
@@ -132,10 +138,8 @@ func generate(cmd *cobra.Command, args []string) error {
 			if importPath != pkgName {
 				testPackages = append(testPackages, importPath)
 			}
-			// Isolate tests files into their own package
-			packs := splitTestPackages(testPkg)
 			// Generate test AST file using non *_test package name since this is Go code that needs to be part of the normal build.
-			if err := generateTestASTFile(dir, strings.TrimSuffix(testPkg.Package, "_test"), packs); err != nil {
+			if err := generateTestASTFile(dir, strings.TrimSuffix(testPkg.Package, "_test"), []*ast.Package{testPkg}); err != nil {
```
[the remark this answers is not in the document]
jlapacik: Never mind. I see what's going on.

stdlib/flux_test.go
```diff
@@ -20,28 +20,32 @@ func init() {
...
+	},
+	"http": {
+		"http_endpoint": "need ability to test side effects in e2e tests: https://github.com/influxdata/flux/issues/1723)",
+	},
+	"testing/chronograf": {
+		"measurement_tag_keys":   "unskip chronograf flux tests once filter is refactored (https://github.com/influxdata/flux/issues/1289)",
+		"aggregate_window_mean":  "unskip chronograf flux tests once filter is refactored (https://github.com/influxdata/flux/issues/1289)",
+		"aggregate_window_count": "unskip chronograf flux tests once filter is refactored (https://github.com/influxdata/flux/issues/1289)",
+	},
+	"testing/pandas": {
```
jlapacik: Should these tests be in the `strings` package?

stdlib/strings/trim_test.flux
```diff
@@ -9,32 +9,32 @@ inData = "
...
-,,0,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,2018-05-22T19:53:36Z,k9ngm,used_percent,disk,disk1,apfs,host.local,/
+,,0,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,2018-05-22T19:54:16Z,13F2,used_percent,disk,disk1,apfs,host.local,/
 "
 
 t_string_trim = (table=<-) =>
 	(table
 		|> range(start: 2018-05-22T19:53:26Z)
 		|> map(fn: (r) =>
-        			({newValue: strings.trimSpace(v: r._value)})))
+        			({r with _value: strings.trimSpace(v: r._value)})))
```
jlapacik: Why did this test case change, but the output schema didn't?

stdlib/test_packages.go
```diff
@@ -4,34 +4,38 @@ package stdlib
...
 	secrets "github.com/influxdata/flux/stdlib/influxdata/influxdb/secrets"
 	v1 "github.com/influxdata/flux/stdlib/influxdata/influxdb/v1"
 	promql "github.com/influxdata/flux/stdlib/internal/promql"
+	regexp "github.com/influxdata/flux/stdlib/regexp"
 	strings "github.com/influxdata/flux/stdlib/strings"
 	chronograf "github.com/influxdata/flux/stdlib/testing/chronograf"
 	kapacitor "github.com/influxdata/flux/stdlib/testing/kapacitor"
 	pandas "github.com/influxdata/flux/stdlib/testing/pandas"
-	testdata "github.com/influxdata/flux/stdlib/testing/testdata"
+	promql1 "github.com/influxdata/flux/stdlib/testing/promql"
```
jlapacik: Why are there two promql packages?

stdlib/flux_test.go
jsternberg: They seemed to be directly related to some tests involving pandas. I don't really know enough about them though to know whether they should be moved. I mostly just tried to move `testdata` to `universe` and then break up anything in there. I've mostly left anything else unmoved, but this will open things up to us moving them if it is more appropriate.

stdlib/strings/trim_test.flux
jsternberg: The test was previously skipped. It looks like it was skipped because of referencing an import inside of map which was fixed awhile ago, but the test itself never seems to have been fully tested.

stdlib/test_packages.go
jsternberg: There's `internal/promql` which involves tests of those functions from that package. I decided instead of attempting to merge things, I would instead just put the transpiler tests into testing as "integration tests". Since this was a structure change, I didn't want to also make more decisions than I needed to at this moment.

jlapacik: I see those are transpiler tests. Got it.

stdlib/flux_test.go
jlapacik: Ok to me it seems like they belong with the `strings` package. But I don't really know enough about them either.

jsternberg:
It looks like, predictably, there would be nothing to delete a generated go file from a directory that no longer had any flux files. Oversight on my part. I'll remove.

Issue #1933, opened 2019-10-02
sqlite connector


nicgrobler:
Hi,

I have been playing with Flux and extended your "sql" package to include sqlite3 - so it supports:  
- sql.from()
- sql.to()  

If interested, and you are not already doing it, I am happy to submit a PR.

I added it because SQLite is an extremely useful "medium" which many platforms support without the need for a full-blown relational DB, with server.

nathanielc:
@nicgrobler We would love to see a PR, we are not actively working to add sqlite support but see it as a valuable addition. Thanks. Let us know if you get stuck on anything.

nicgrobler:
@nathanielc thanks - couple of questions have come up, which need some strategic decisions from you guys that go way beyond simply adding SQLite (I have 2 working branches, but they are very different, for the reasons below):
 
1. As most flavours of "sql DB" out there do not support boolean types (essentially stored as flavours of INT)we need to decide whether we support "SQL booleans", or not: Saying "No" means that our own code will be very much less ambiguous, and simple as a result. If we choose "Yes", then we need to add a way for the user (extra function argument) to specify which columns are in fact bools, and therefore override the reader to parse INT -> BOOL.

2. If we don't decide, and I simply add sqlite to the existing code, and we rely on accepting Bools when supported, it is very easy to end up in the following situation:

Postgres (bool) -> Flux (bool) -> MySQL (int), so far "OK" -> flux (int) oh dear -> Postgres (int) if this is the same DB as originally used, we now have an issue.

So I guess what I'm asking is:
1. Do we support SQL booleans?

2. If Yes, is the following API change acceptable: 
```
sql.from(driverName: , dataSourceName: , query: , boolColumns["bool_as_int_column",...])
```
3. I suggest that boolColumns is optional (most DBs will not be using bools) to not break existing API - but then we need to let the users know clearly, that if their SQL source has a boolean column, and they don't explicitly list it, the query will fail with a type error.

Hopefully this makes sense? I only ask because adding SQLite into the existing code took me less than 4 hours, but I did several days worth of testing, which exposed these issues (I tested Flux against live instances of mysql, influxdb, mysql, and sqlite).  
SQLite (and Postgres) is by nature a very happy little DB that simply stores whatever you want, wherever you want, which forces me to ask you which direction you want to go in.

nathanielc:
[...]

PR #1984, opened 2019-10-14
WIP:feat(stdlib/sql): Sqlite3 support #1933 and fixes #1938

### LINKED ISSUE #1938: Sql errors not being returned to user
Hi,

```
	if err != nil {
		return err
	}
	for i := range valStrings {
			return nil
		}
	}

	return err
}
```
```
```
to this:
```
```

nicgrobler:
### This PR adds
 
- Basic use cases tested against running postgres & mysql instances.
 
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


nicgrobler:
not going to pollute this further - will close, squash and resubmit in new PR.

Issue #1952, opened 2019-10-09
Parse type expressions


jlapacik:
This issue is for creating a small parser for polytype and monotype expressions. Note this issue is for testing purposes. We need a way to succinctly express the types that we expect in test cases. Test cases should be easy to write as well as easy to read and understand. They should also be easy to debug if necessary.

Ex.
```
f = (x, y) => x + y
```
The polytype of `f` should be expressed as:
```
forall ['a] where 'a: Addable (x: 'a, y: 'a) -> 'a
```

See [here](https://github.com/jlapacik/influxible/blob/master/infer_test.go) for examples and [here](https://github.com/jlapacik/ftel/blob/master/antlr/flux.g4) for the grammar.

Note this issue does not require adding type expressions to the Flux grammar/parser. That can be done at a later date at which point type inference test cases will just use the Flux parser to parse their expected types.

Note, the above links (and example) are just examples. This issue will define the grammar used for type expressions.

affo:
This solution _could be_ explored: https://github.com/goffrie/plex.

We could follow the ragel lexer + hand-made parser too

jlapacik:
Some clarification: The output of parsing should be an AST that is distinct from the actual types defined as part of #1951. There should however be a 1-1 mapping between these AST nodes and the type structures defined in #1951. By having this parser generate its own AST, we can easily integrate it with the Flux parser when it comes time to add type expressions to the Flux grammar.

After this issue has been implemented, all that will be needed is a simple function to instantiate a type structure from the aforementioned AST.

nathanielc:
As discussed in #1993 we are going simplify the grammar slightly for type variables.

Instead of parsing type variables as  `'a` we will expect the form `t0` and `t1` etc. The production rule will be:

```
tvr
    : 't' [0-9]+
    ;
```

jlapacik:
Update: It turns out we will in fact need to support multiple kind constraints on type variables. Hence the grammar should be updated to support polytype expressions like:
```
forall [t0] where t0: SomeKind + AnotherKind + ... + LastKind (x: t0) -> t0
```
In particular you can constrain free type variables with multiple kinds using `+`.

PR #1986, opened 2019-10-14
feat(rust/semantic): Define initial Flux data types

### LINKED ISSUE #1951: Define the main data types of the Flux type system
This issue is to define the 3 main data types used in the Flux type system. They are `MonoType`, `PolyType`, and `Kind`.

A monotype is a specific named type such as integer, float, or string. It should probably be represented using an enum like so:
```Rust
enum MonoType {
    Int,
    Flt,
    Str,
    Tvar,
    ...
}
```

A polytype is a parametric type. Parametric types are parametrized or quantified by zero or more free variables that can be replaced with arbitrary monotypes. This substitution of a polytype's free variables is what allows for polymorphism in Flux. Note that polytypes are a superset of monotypes. A monotype is a polytype with zero free variables, i.e. there's only one way to instantiate it. A polytype should be defined something like this:
```Rust
use std::collections::HashMap;

struct PolyType {
    freeVars: Vec<Tvar>,
    constraints: Option<HashMap<Tvar, Kind>>,
    expr: MonoType,
}
```

Note, every variable in Flux is assigned a parametric type, while every expression in Flux is given a monomorphic type.

The final concept of the Flux type system is that of kinds. A kind represents a class of types (think type classes in Haskell). Kinds are used to constrain the free variables of a polytype. For example, if we know that one of a polytype's free variables represents a type that can be used with the `+` operator, then we know that type is of kind `Addable`. In other words, it belongs to the class of types that `+` accepts. Note, kinds or type classes are what allow for fine grained operator overloading in Flux. The `Kind` data type should probably be defined as an enum as well.
```Rust
enum Kind {
    Addable,
    Divisible,
    Comparable,
    Nullable,
    ....
}
```

Note all 3 of the above data types should implement (not derive) custom equality and display methods corresponding to the `PartialEq` and `Display` traits respectively.

jlapacik:
Closes #1951.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+            Row::Nil => (HashMap::new(), None),
+            Row::Var(tv) => (HashMap::new(), Some(*tv)),
+            Row::Extension { head: h, tail: t } => {
+                let (mut r, tv) = t.flatten();
+                r.insert(h.k.clone(), &h.v);
+                (r, tv)
+            }
+        }
+    }
+}
```
jlapacik: `flatten` is used only to compare two record types for equality. It should always be a private internal method.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}:{}", self.tv, self.kind)
+    }
+}
+
+// List is a wrapper around a generic vector
+#[derive(Debug, Clone, PartialEq)]
+struct List<'a, T> {
+    values: &'a Vec<T>,
+}
```
jlapacik: This is an internal type whose only purpose is for displaying a comma separated list of values.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            MonoType::Bool => f.write_str("bool"),
+            MonoType::Int => f.write_str("int"),
+            MonoType::Uint => f.write_str("uint"),
+            MonoType::Flt => f.write_str("flt"),
+            MonoType::Str => f.write_str("str"),
+            MonoType::Dur => f.write_str("dur"),
+            MonoType::Time => f.write_str("time"),
+            MonoType::Rgx => f.write_str("rgx"),
```
jlapacik: Is `rgx` too compact?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    Nul,
+}
+
+impl fmt::Display for Kind {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Kind::Add => f.write_str("Addable"),
+            Kind::Div => f.write_str("Divisible"),
+            Kind::Cmp => f.write_str("Comparable"),
+            Kind::Nul => f.write_str("Nullable"),
```
nathanielc: Should we use the full names for the types themselves? I didn't know that `Cmp` was for `Comparable` without reading this display code first.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum MonoType {
+    Bool,
+    Int,
+    Uint,
+    Flt,
+    Str,
+    Dur,
+    Time,
+    Rgx,
```
nathanielc: Similar question here, I would rather spell these all the way out, i.e. `Float`, `String` and `Regexp`.  `Uint` is fine as that is very common.

internal/rust/semantic/src/types.rs
nathanielc: Yes, I would like these printed names to be spelled out.  Check the Flux SPEC all the these type have an official `name` defined in the spec. We should use that name.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+            MonoType::Rec(obj) => obj.fmt(f),
+            MonoType::Fun(fun) => fun.fmt(f),
+        }
+    }
+}
+
+// Tvar stands for type variable.
+// A type variable holds an unknown type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Tvar(i64);
```
nathanielc: Why does this type need `Hash`?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+// A row is either Nil meaning it has no properties,
+// or it is an extension of a row.
+//
+// A row may extend what is referred to as a row
+// variable. A row variable is a type variable that
+// represents an unknown record type.
+//
+#[derive(Debug, Clone)]
+pub enum Row {
+    Nil,
```
nathanielc: Should we call this `Empty` instead of `Nil`? Meaning is there a difference between a `nil` record and an `empty` record?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    Var(Tvar),
+    Extension { head: Property, tail: Box<Row> },
+}
+
+impl fmt::Display for Row {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Row::Nil => f.write_str("{}"),
+            Row::Var(tv) => tv.fmt(f),
```
nathanielc: Since this is a row variable should it be wrapped in `{}` for formatting?

internal/rust/semantic/src/types.rs
nathanielc: Makes sense, seems like later we should be able to implement an allocation free comparison that knows how to do the transitive lookup of properties within a row. But that can come later as an optimization.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+                let (mut r, tv) = t.flatten();
+                r.insert(h.k.clone(), &h.v);
+                (r, tv)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Property {
```
nathanielc: Can you add a comment here?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+}
+
+impl fmt::Display for Property {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}:{}", self.k, self.v)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Function {
```
nathanielc: Can you comment this type as well?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+        args.append(&mut opt);
+        match &self.retn {
+            Some(t) => write!(f, "({}) -> {}", List { values: &args }, t),
+            None => write!(f, "({}) -> !", List { values: &args }),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct BoundTvar {
```
nathanielc: What is a BoundTvar? Is it just used to format a Tvar with its bound Kind? 
I think my confusion was I assumed this was public, I see now that it is private. Maybe a simple comment explaining it purpose local to this file would be helpful.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+//     that instantiates both polytypes with the
+//     same type variables.
+//
+//     Note this requires a substitution, so remove
+//     this derivation once substitutions are defined.
+//
+#[derive(Debug, Clone, PartialEq)]
+pub struct PolyType {
+    pub free: Vec<Tvar>,
+    pub bnds: Option<HashMap<Tvar, Kind>>,
```
nathanielc: Can a Tvar have multiple Kinds? Meaning is this valid? 

```
forall ['a] where 'a is Addable and 'a is Comparable (x:'a) -> 'a
```

If so I think we need a different structure here. Would a list of tvar kind pairs be sufficient?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+
+impl fmt::Display for BoundTvar {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}:{}", self.tv, self.kind)
+    }
+}
+
+// List is a wrapper around a generic vector
+#[derive(Debug, Clone, PartialEq)]
+struct List<'a, T> {
```
nathanielc: Its going to take my a bit to remember that upper case types are not public implicitly anymore :) 

It makes sense that this is private, but I was confused why it was "public" for a bit.

nathanielc: Should we call this `DisplayList` or something to make it more clear that is used only as an aid for displaying a list of things?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+// Display a generic list whose values can be displayed
+impl<'a, T: fmt::Display> fmt::Display for List<'a, T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.values.is_empty() {
+            return Ok(());
+        }
+        let size = self.values.len();
+        let list = &self.values[..size - 1];
+        for v in list {
+            if let Err(e) = write!(f, "{}, ", v) {
```
nathanielc: nit, Should we remove the trailing comma? It should be a simple fix and will likely improve readability.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    fn display_type_array() {
+        assert_eq!(
+            "[int]",
+            MonoType::Arr(Box::new(Array(MonoType::Int))).to_string()
+        );
+    }
+    #[test]
+    fn display_type_record() {
+        assert_eq!(
+            "{a:int | {b:str | t0}}",
```
nathanielc: Do we want to flatten these for display?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+        write!(f, "{}:{}", self.k, self.v)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Function {
+    pub req: HashMap<String, MonoType>,
+    pub opt: HashMap<String, MonoType>,
+    pub pipe: Option<Property>,
+    pub retn: Option<MonoType>,
```
nathanielc: Should the return type be optional? For now I would like to simplify and require a return type for functions. The Flux SPEC says that functions must have a return type so I think we should follow that for now. We can investigate allowing `void` return types later.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+}
+
+// Tvar stands for type variable.
+// A type variable holds an unknown type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Tvar(i64);
+
+impl fmt::Display for Tvar {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "t{}", self.0)
```
nathanielc: Should this include the single quote to be consistent with the type expression grammar? Meaning lie this `'t0`?

Should we map numeric types to the alphabet? Meaning use `'a`, `'z`, `'aa` and `'ab` for `'t0`, `'t25`, `'t26` and `'t27` respectively? Not sure this is a good idea, just thinking out loud.

internal/rust/semantic/src/types.rs
fchikwekwe: I agree with that. It would make the code more readable.

internal/rust/semantic/src/types.rs
nathanielc: Nevermind I missed that the slicing above was an exclusive bound and therefore was dropping the final item.

internal/rust/semantic/src/types.rs
jlapacik: That doesn't seem necessary since this is just for testing. Perhaps this should be part of the test module instead of a struct method.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,666 @@
...
+    Int,
+    Uint,
+    Flt,
+    Str,
+    Dur,
+    Time,
+    Rgx,
+    Var(Tvar),
+    Arr(Box<Array>),
+    Rec(Box<Row>),
```
fchikwekwe: Why `Rec` for Row? Just curious.

internal/rust/semantic/src/types.rs
jlapacik: Substitutions will be implemented as hash maps where the keys are type variables.

internal/rust/semantic/src/types.rs
jlapacik: I was just following the convention which is `Nil`. The comment specifies that it represents the record with no properties. Note there is no ambiguity as records cannot be null.

internal/rust/semantic/src/types.rs
jlapacik: That would be unconventional so I would prefer not to.

internal/rust/semantic/src/types.rs
jlapacik: Yes a comment would be helpful. It's only used for properly displaying a polytype.

internal/rust/semantic/src/types.rs
jlapacik: At the moment as well the foreseeable future type variables will only be bound by a single kind. Having said that, a list might be better.

internal/rust/semantic/src/types.rs
jlapacik: Yes we do.

internal/rust/semantic/src/types.rs
jlapacik: My bad. I was not aware that was specified in the spec. But I like that much much much better.

internal/rust/semantic/src/types.rs
jlapacik: I would like the display code to be as small and concise as possible. If possible it should be immediately obvious to someone who is reading the code, how types are displayed in my opinion. So my vote would be to not include the single quote or any alphabet mapping.

internal/rust/semantic/src/types.rs
jlapacik: No reason really. I just thought it was less stutter than `Row(Box<Row>)`.

internal/rust/semantic/src/types.rs
nathanielc: >Note there is no ambiguity as records cannot be null.

I think we should call it `Empty` then as to not accidentally confuse that records could be nil.

internal/rust/semantic/src/types.rs
nathanielc: Sounds good

internal/rust/semantic/src/types.rs
nathanielc: What kind constraints does this code have?

```
change = (x,y) => (x - y) / (x + y)
```

I would expect a type signature like this:

```
forall ['a] where 'a is Addable and 'a is Subtractable and 'a is Divisible (x:'a, y:'a) -> 'a
```

internal/rust/semantic/src/types.rs
nathanielc: ok, should we change the type expression grammar to not use single quotes? I am operating under the assumption that the displayed types should be parse-able type expressions. That way as developers or users we speak the same language regarding types.

internal/rust/semantic/src/types.rs
nathanielc: I think we should pick either `Row` or `Record` for the name of our record/row type. That way its not confusing if they are different as they are really the same thing.

internal/rust/semantic/src/types.rs
jlapacik: In this case the type signature would be:
```
forall ['a] where 'a is Divisible (x:'a, y:'a) -> 'a
```
as `Divisible` is the smallest class satisfying the above constraints.

nathanielc: Ok, that makes sense. Does the logic about which Kind is smallest get applied during unification? Meaning that when we see Addable unify with Divisible we have logic that knows that they do unify but the resulting kind is the Divisible one? Not sure I am asking that clearly, in other words where does the logic about which Kind is the smallest satisfying kind live?

jlapacik: You're being quite clear :). This logic is not present at the moment but will live with the kind itself. So each kind will know how to "unify" with other kinds.

nathanielc: Sounds good. Thanks!

internal/rust/semantic/src/types.rs
jlapacik: Yes the displayed types should be valid type expressions. The single quote form and `t` prefix form should both be parse-able. This should be specified as part of #1952.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,705 @@
...
+//
+// A row is either Empty meaning it has no properties,
+// or it is an extension of a row.
+//
+// A row may extend what is referred to as a row
+// variable. A row variable is a type variable that
+// represents an unknown record type.
+//
+#[derive(Debug, Clone)]
+pub enum Row {
```
affo: I guess this is what `Object` is now in our Go implementation?

Or maybe not Object in the sense of `values.Object`, because that's a value and not a type.
Do we have an equivalent in Go?

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+impl fmt::Display for PolyType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self.bnds {
+            Some(bnds) => {
+                let mut bounds = Vec::new();
+                for tv in &self.free {
+                    if let Some(kind) = bnds.get(tv) {
+                        bounds.push(BoundTvar {
```
affo: I guess you could avoid building up a new vector to hold those variables and directly push the result of formatting in the final string, that should use less memory and blablabla.

But I guess this is only for debugging and does not need to be very efficient

internal/rust/semantic/src/types.rs
jlapacik: Yes you're right, this is the replacement for what is currently the Object type in Go. It looks a little different than the Go type as it is designed to be extensible. But it serves the same purpose.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+// MonoType represents a specific named type
+#[derive(Debug, Clone, PartialEq)]
+pub enum MonoType {
+    Bool,
+    Int,
+    Uint,
+    Float,
+    Str,
```
nathanielc: nit, to be consistent this should be String.

internal/rust/semantic/src/types.rs
nathanielc: why do we want both forms? I'll create a separate issue to discuss this fully, I don't want to block this PR on the grammar of the type expressions.

internal/rust/semantic/src/types.rs
```diff
@@ -0,0 +1,705 @@
...
+            },
+            self.retn
+        )
+    }
+}
+
+// BoundTvar represents a constrained type variable.
+// Used solely for displaying the generic constraints of a polytype.
+#[derive(Debug)]
+struct BoundTvar {
```
nathanielc: nit, BoundTvar made me think of the type variable being bound in the environment not as being constrained by a kind. Maybe a name like TvarKind or TKPair or something like that that communicates is constrained by a kind would be more clear? 

Its also fine as BoundTvar since the comment makes it clear. Do what ever you think works.

internal/rust/semantic/src/types.rs
nathanielc: Issue https://github.com/influxdata/flux/issues/1993

internal/rust/semantic/src/types.rs
jlapacik: Oops I thought the name `String` was reserved, but it's not.

PR #1987, opened 2019-10-14
feat(stdlib/sql): add sqlite3 support

### LINKED ISSUE #1933: sqlite connector

### LINKED ISSUE #1938: Sql errors not being returned to user

nicgrobler:
Adds support for SQLite3 to the stdlib/sql package. SQLite is a fully SQL compliant relational DB, in a file (or in memory, if preferred). Adding this to flux would allow it to interact with this DB, and present a lot of opportunity to extract information from many 3rd party "infrastructure tools", and combine this with Influxdb, or other data sources.

Closes #1933
Fixes #1938 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/sql/from.go
+++ b/stdlib/sql/from.go
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
 	_ "github.com/lib/pq"
+	_ "github.com/mattn/go-sqlite3"
 )
 
 const FromSQLKind = "fromSQL"
@@ -117,6 +118,8 @@ func createFromSQLSource(prSpec plan.ProcedureSpec, dsid execute.DatasetID, a ex
 	switch spec.DriverName {
 	case "mysql":
 		newRowReader = NewMySQLRowReader
+	case "sqlite3":
+		newRowReader = NewSqliteRowReader
 	case "postgres", "sqlmock":
 		newRowReader = NewPostgresRowReader
 	default:
@@ -189,7 +192,6 @@ func read(ctx context.Context, reader execute.RowReader, alloc *memory.Allocator
 			return nil, err
 		}
 	}
-
 	for reader.Next() {
 		row, err := reader.GetNextRow()
 		if err != nil {
--- a/stdlib/sql/sql_test.go
+++ b/stdlib/sql/sql_test.go
@@ -231,6 +231,96 @@ func TestMySQLParsing(t *testing.T) {
 	}
 }
 
+func TestSQLiteParsing(t *testing.T) {
+	testCases := []struct {
+		name       string
+		columnName string
+		columnType flux.ColType
+		data       [][]uint8
+		want       [][]values.Value
+	}{
+		{
+			name:       "ints",
+			columnName: "_int",
+			columnType: flux.TInt,
+			data:       stringSliceToByteArrays([]string{"6", "1", "643", "42", "1283", "4", "0", "18"}),
+			want:       [][]values.Value{{values.NewInt(6)}, {values.NewInt(1)}, {values.NewInt(643)}, {values.NewInt(42)}, {values.NewInt(1283)}, {values.NewInt(4)}, {values.NewInt(0)}, {values.NewInt(18)}},
+		},
+		{
+			name:       "floats",
+			columnName: "_float",
+			columnType: flux.TFloat,
+			data:       stringSliceToByteArrays([]string{"6", "1", "643", "42", "1283", "4", "0", "18"}),
+			want:       [][]values.Value{{values.NewFloat(6)}, {values.NewFloat(1)}, {values.NewFloat(643)}, {values.NewFloat(42)}, {values.NewFloat(1283)}, {values.NewFloat(4)}, {values.NewFloat(0)}, {values.NewFloat(18)}},
+		},
+		{
+			name:       "strings",
+			columnName: "_string",
+			columnType: flux.TString,
+			data:       stringSliceToByteArrays([]string{"6", "1", "643", "42", "1283", "4", "0", "18"}),
+			want:       [][]values.Value{{values.NewString("6")}, {values.NewString("1")}, {values.NewString("643")}, {values.NewString("42")}, {values.NewString("1283")}, {values.NewString("4")}, {values.NewString("0")}, {values.NewString("18")}},
+		},
+		{
+			name:       "datetime",
+			columnName: "_datetime",
+			columnType: flux.TTime,
+			data: stringSliceToByteArrays([]string{
+				"2019-06-03 13:59:00",
+				"2019-06-03 13:59:01",
+				"2019-06-03 13:59:02",
+				"2019-06-03 13:59:03",
+				"2019-06-03 13:59:04",
+				"2019-06-03 13:59:05",
+				"2019-06-03 13:59:06",
+				"2019-06-03 13:59:07"}),
+			want: [][]values.Value{
+				{values.NewTime(values.ConvertTime(createTestTimes()[0].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[1].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[2].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[3].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[4].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[5].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[6].(time.Time)))},
+				{values.NewTime(values.ConvertTime(createTestTimes()[7].(time.Time)))},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+
+			TestReader := &SqliteRowReader{}
+
+			TestReader.NextFunc = func() func() bool {
+				i := 0
+				vals := make([]interface{}, len(tc.data))
+				for i, v := range tc.data {
+					vals[i] = v
+				}
+				return func() bool {
+					if i < len(tc.data) {
+						TestReader.SetColumns([]interface{}{vals[i]})
+						i++
+						return true
+					}
+					return false
+				}
+			}()
+			TestReader.InitColumnNames([]string{tc.columnName})
+			TestReader.SetColumnTypes([]flux.ColType{tc.columnType})
+
+			i := 0
+			for TestReader.Next() {
+				row, _ := TestReader.GetNextRow()
+				if !cmp.Equal(tc.want[i], row) {
+					t.Fatalf("unexpected result -want/+got\n\n%s\n\n", cmp.Diff(tc.want[i], row))
+				}
+				i++
+			}
+		})
+	}
+}
+
 func createTestTimes() []interface{} {
 	str := []string{"2019-06-03 13:59:00",
 		"2019-06-03 13:59:01",
--- a/stdlib/sql/sqlite.go
+++ b/stdlib/sql/sqlite.go
@@ -3,7 +3,6 @@ package sql
 import (
 	"database/sql"
 	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/influxdata/flux"
@@ -13,7 +12,7 @@ import (
 	"github.com/influxdata/flux/values"
 )
 
-type MySQLRowReader struct {
+type SqliteRowReader struct {
 	Cursor      *sql.Rows
 	columns     []interface{}
 	columnTypes []flux.ColType
@@ -22,8 +21,7 @@ type MySQLRowReader struct {
 	CloseFunc   func() error
 }
 
-// Next prepares MySQLRowReader to return rows
-func (m *MySQLRowReader) Next() bool {
+func (m *SqliteRowReader) Next() bool {
 	if m.NextFunc != nil {
 		return m.NextFunc()
 	}
@@ -45,16 +43,14 @@ func (m *MySQLRowReader) Next() bool {
 	return next
 }
 
-func (m *MySQLRowReader) GetNextRow() ([]values.Value, error) {
+func (m *SqliteRowReader) GetNextRow() ([]values.Value, error) {
 	row := make([]values.Value, len(m.columns))
 	for i, col := range m.columns {
 		switch col := col.(type) {
 		case bool, int64, uint64, float64, string:
 			row[i] = values.New(col)
 		case []uint8:
-			// Hack for MySQL, might need to work with charset?
-			// Can't do boolean with MySQL - stores BOOLEANs as TINYINTs (0 or 1)
-			// No way to distinguish if intended int or bool
+			// this allows easier testing using existing methods
 			switch m.columnTypes[i] {
 			case flux.TInt:
 				newInt, err := UInt8ToInt64(col)
@@ -68,7 +64,6 @@ func (m *MySQLRowReader) GetNextRow() ([]values.Value, error) {
 					return nil, err
 				}
 				row[i] = values.NewFloat(newFloat)
-			// This works, but you can also just just add the DSN parameter parseTime=true (see line 136)
 			case flux.TTime:
 				t, err := time.Parse(layout, string(col))
 				if err != nil {
@@ -89,20 +84,24 @@ func (m *MySQLRowReader) GetNextRow() ([]values.Value, error) {
 	return row, nil
 }
 
-func (m *MySQLRowReader) InitColumnNames(names []string) {
-	m.columnNames = names
+func (m *SqliteRowReader) InitColumnNames(n []string) {
+	m.columnNames = n
 }
 
-func (m *MySQLRowReader) InitColumnTypes(types []*sql.ColumnType) {
+func (m *SqliteRowReader) InitColumnTypes(types []*sql.ColumnType) {
 	stringTypes := make([]flux.ColType, len(types))
 	for i := 0; i < len(types); i++ {
 		switch types[i].DatabaseTypeName() {
-		case "INT", "BIGINT", "SMALLINT", "TINYINT":
+		case "INT", "INTEGER", "BIGINT", "SMALLINT", "TINYINT":
 			stringTypes[i] = flux.TInt
 		case "FLOAT", "DOUBLE":
 			stringTypes[i] = flux.TFloat
-		case "DATETIME":
+		case "DATETIME", "TIMESTAMP", "DATE":
 			stringTypes[i] = flux.TTime
+		case "TEXT":
+			stringTypes[i] = flux.TString
+		case "BOOL", "BOOLEAN":
+			stringTypes[i] = flux.TInt
 		default:
 			stringTypes[i] = flux.TString
 		}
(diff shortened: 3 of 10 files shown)


nathanielc:
@nicgrobler Thanks! I see you have signed the CLA so thanks. We will review this and get back to you.

nicgrobler:
No worries - Trashed first PR as I wasn’t heroic enough to read the contributing guide in detail..which is actually very helpful when you read all of it:-) So cleaned up commit messages and resubmitted, in the required style. Sorry!

stdlib/sql/to.go
```diff
@@ -207,38 +222,95 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+		have a default limit on the number of parameters which can be queued before calling Exec; SQLite, for example, has a default of 999 (can only be changed
+		at compile time).
+
+		So if the row width is 10 columns, the maximum Batchsize would be:
+		(999 - row_width) / row_width = 89 rows.
+
+		Sending more would result in the call to Exec returning a "too many SQL variables" error, and the transaction would be rolled-back / aborted
+	*/
+
+	if batchSize < numberCols {
```
yzhang1991: @nathanielc can `numberCols` be greater than 999 in flux?

stdlib/sql/to.go
```diff
@@ -207,38 +222,95 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+		return numberCols
+	}
+	return (batchSize - numberCols) / numberCols
```
yzhang1991: I could be wrong if I missed something to understand this 999 cap thing right,  but I think this should be
```suggestion
	return batchSize / numberCols
```

stdlib/sql/to.go
```diff
@@ -207,38 +222,95 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+	batchSize := correctBatchSize(t.spec.Spec.BatchSize, len(cols))
+
 	labels := make(map[string]idxType, len(cols))
 	var questionMarks, newSQLTableCols []string
 	for i, col := range cols {
 		labels[col.Label] = idxType{Idx: i, Type: col.Type}
 		questionMarks = append(questionMarks, "?")
 		colNames = append(colNames, col.Label)
-
+		drivername := t.spec.Spec.DriverName
```
yzhang1991:
```suggestion
		driverName := t.spec.Spec.DriverName
```

stdlib/sql/to.go
```diff
@@ -207,38 +222,95 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+		// the following allows driver-specific type errors (of which there can be MANY) to be returned, rather than the default of invalid type
+		translateColumn, err := getTranslationFunc(drivername)
+		if err != nil {
+			return nil, nil, nil, err
+		}
 		switch col.Type {
 		case flux.TFloat:
-			newSQLTableCols = append(newSQLTableCols, fmt.Sprintf("%s FLOAT", col.Label))
+			v, err := translateColumn()(flux.TFloat, col.Label)
```
yzhang1991: Why switch?

`translateColumn()(col.Type, col.Label)`??

stdlib/sql/to_privates_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/influxdata/flux/internal/errors"
+)
+
+// additional and seperate tests that can be run without needing functions to be Exported in sql, just to be testable
+func TestCorrectBatchSize(t *testing.T) {
+	// given the combination of row width and supplied batchSize argument from user, verify that it is modified as required
+	userBatchSize := 1000
+	rowWidth := 10
+	correctedSize := correctBatchSize(userBatchSize, rowWidth)
+	if !cmp.Equal(99, correctedSize) {
```
yzhang1991: I am confused by this result.

stdlib/sql/to_privates_test.go
```diff
@@ -0,0 +1,252 @@
...
+	// verify that this works as it should - used here to avoid exporting this functionality
+	columnLabel := "apples"
+	// verify invalid return error
+	_, err := getTranslationFunc("bananas")
+	if !cmp.Equal(errors.New(codes.Internal, "invalid driverName: bananas").Error(), err.Error()) {
+		t.Log(cmp.Diff(errors.New(codes.Internal, "invalid driverName: bananas").Error(), err.Error()))
+		t.Fail()
+	}
+
+	// verify that valid returns expected hapiness for SQLITE
```
yzhang1991:
```suggestion
	// verify that valid returns expected happiness for SQLITE
```

stdlib/sql/to_privates_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if !cmp.Equal(nil, err) {
+		t.Log(cmp.Diff(nil, err))
+		t.Fail()
+	}
+	if !cmp.Equal(columnLabel+" DATETIME", v) {
+		t.Log(cmp.Diff(columnLabel+" DATETIME", v))
+		t.Fail()
+	}
+	// as SQLITE has NO BOOLEAN column type, we need to return an error rather than doing implicit conversions
+	_, err = sqlT()(flux.TBool, columnLabel)
```
yzhang1991: Is this the behavior that we want? Why can't we just make it work?

yzhang1991:
Also, I wonder if we should somehow use our FilesystemService to audit sqlite3's DSN. @nathanielc So far I don't even know if that's possible.
Some sqlite3 DSN examples:
* test.db
* file:test.db?cache=shared&mode=memory
* :memory:
* file::memory:

stdlib/sql/sqlite.go
```diff
@@ -0,0 +1,169 @@
...
+		case "INT", "INTEGER", "BIGINT", "SMALLINT", "TINYINT":
+			stringTypes[i] = flux.TInt
+		case "FLOAT", "DOUBLE":
+			stringTypes[i] = flux.TFloat
+		case "DATETIME", "TIMESTAMP", "DATE":
+			stringTypes[i] = flux.TTime
+		case "TEXT":
+			stringTypes[i] = flux.TString
+		case "BOOL", "BOOLEAN":
+			stringTypes[i] = flux.TInt
```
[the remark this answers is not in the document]
yzhang1991: I understand that there is no `BOOL` in sqlite3. But isn't it preferable that when we see the data in flux it really becomes bool?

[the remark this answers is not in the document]
nicgrobler: No - there is no book type in SQLite, but people sometimes alias the column. If you store true or false in the db, it quietly converts it into an int. so any column labelled as bool, will in fact be an int. the confusion, compared to something like MySQL is that SQLite will allow you to define the column type to be anything you like - unlike MySQL which changes the column type to be tinyint- so if you try to read this into a flux.TBool, it will always fail.

stdlib/sql/sqlite.go
```diff
@@ -0,0 +1,169 @@
...
+func (m *SqliteRowReader) SetColumnTypes(types []flux.ColType) {
+	m.columnTypes = types
+}
+
+func (m *SqliteRowReader) SetColumns(i []interface{}) {
+	m.columns = i
+}
+
+func (m *SqliteRowReader) Close() error {
+	if err := m.Cursor.Err(); err != nil {
```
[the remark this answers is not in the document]
nicgrobler: Yep - I wondered the same, but thought it likely it was there for a reason that I was too stupid to understand

stdlib/sql/sqlite.go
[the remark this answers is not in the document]
yzhang1991: Yeah, I see that makes sense from a technical perspective. I will pass this question to the team for comments. Because as a user my expectation here might be seeing boolean values in flux, although that requires an implicit conversion.

stdlib/sql/sqlite.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// SqliteTranslateColumn translates flux colTypes into their corresponding SQLite column type
+func SqliteTranslateColumn() translationFunc {
+	c := map[string]string{
+		flux.TFloat.String():  "FLOAT",
+		flux.TInt.String():    "INT",
+		flux.TUInt.String():   "INT",
+		flux.TString.String(): "TEXT",
+		flux.TTime.String():   "DATETIME",
```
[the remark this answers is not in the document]
nicgrobler: Nope - there is no bool support in SQLite. This allows the user to take the choice to explicitly use flux to convert the bool column into an INT type, and then feed to SQLite. If we try to support bools, given sqlite’s dynamic typing, we would be doing all kinds of implicit type conversions under the covers

stdlib/sql/to_privates_test.go
yzhang1991: This question is more from a product perspective and is not necessarily a request for change.

stdlib/sql/to.go
```diff
@@ -207,38 +222,95 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+
+func correctBatchSize(batchSize, numberCols int) int {
+	/*
+		BatchSize for the DB is the number of parameters that can be queued within each call to Exec.
+		As each row you send has a parameter count equal to the number of columns (i.e. the number of "?" used in the insert statement), and some DBs
+		have a default limit on the number of parameters which can be queued before calling Exec; SQLite, for example, has a default of 999 (can only be changed
+		at compile time).
+
+		So if the row width is 10 columns, the maximum Batchsize would be:
+		(999 - row_width) / row_width = 89 rows.
```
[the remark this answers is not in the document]
nicgrobler: Typo In the comment, should be 99. you are correct.

stdlib/sql/to.go
nicgrobler: Good point - if it can, then we would be inserting rows which exceed the number of Parma’s that the DB will accept in an insert. Likely this would also impact many other dbs. Perhaps returning an error so user can ‘do something’ about limiting the width of the data?

stdlib/sql/to.go
nicgrobler: No - you need to take account of the initial column width - the column names used as part of the insert statement - if you do it this way, you exceed the limit. At least, that is what I found during testing.

stdlib/sql/sqlite.go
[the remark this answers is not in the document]
yzhang1991: I think maybe you are right... I have been thinking about as a user what kind of behavior I will expect wrt. boolean columns... It came up several times in the PR, some inputs from the product guys may be helpful lol.

stdlib/sql/to.go
nicgrobler: 👍🏻

stdlib/sql/to.go
yzhang1991: Do you mean the column names are also in this 999 allowed variables?
If I have a SQL says `INSERT INTO T (col) VALUES (1)` it actually takes 2 slots (1 column name + 1 column value)?

nicgrobler: Yes exactly - that initial slot you used for col, would need to be taken into account

stdlib/sql/to_privates_test.go
nicgrobler: The calculation is correct but it is not clear why - perhaps it should link to the explanation (which will need to be updated) in the code about batchsize?

stdlib/sql/to_privates_test.go
nicgrobler: I wrote a version that handled all of the implicit conversions from and to bools for SQLite, and it can work, but it then makes flux behave in a dynamic way, and only solves the issue for SQLite - most other db (oracle, sybase etc) do not support bools either

stdlib/sql/mysql.go
```diff
@@ -169,3 +172,24 @@ func NewMySQLRowReader(r *sql.Rows) (execute.RowReader, error) {
 
 	return reader, nil
 }
+
+// MysqlTranslateColumn translates flux colTypes into their corresponding MySQL column type
+func MysqlTranslateColumn() translationFunc {
```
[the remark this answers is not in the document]
nicgrobler: Yes - agree on this and the other two.

stdlib/sql/mysql.go
```diff
@@ -169,3 +172,24 @@ func NewMySQLRowReader(r *sql.Rows) (execute.RowReader, error) {
...
+		flux.TString.String(): "TEXT(16383)",
+		flux.TTime.String():   "DATETIME",
+		flux.TBool.String():   "BOOL",
+		// BOOL is a synonym supplied by MySQL for "convenience", and MYSQL turns this into a TINYINT type under the hood
+		// which means that looking at the schema afterwards shows the columntype as TINYINT, and not bool!
+	}
+	return func(f flux.ColType, colname string) (string, error) {
+		s, found := c[f.String()]
+		if !found {
+			return "", errors.Newf(codes.Internal, "DB does not support column type %s", f.String())
```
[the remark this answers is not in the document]
nicgrobler: Yes, no argument.

stdlib/sql/sqlite.go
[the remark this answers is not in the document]
nicgrobler: in the original issue (linked), @nathanielc answers this question very well - I also fell into the ‘solving bools’ hole, and I think his answer really helps.

stdlib/sql/sqlite_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+func TestNulTranslation(t *testing.T) {
+	// write values to "DB" - as would be done by sql.to() - using sqlite3 as it allows maximum levels of type abuse, and therefore likely the largest
+	// potential for issues
+	db, err := sql.Open("sqlite3", ":memory:")
+	if !cmp.Equal(err, nil) {
+		t.Fatalf("unexpected result -want/+got\n\n%s\n\n", cmp.Diff(nil, err))
+	}
+
+	// create table
```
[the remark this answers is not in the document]
nicgrobler: 👍🏻

stdlib/sql/sql_test.go
```diff
@@ -231,6 +231,96 @@ func TestMySQLParsing(t *testing.T) {
...
+func TestSQLiteParsing(t *testing.T) {
+	testCases := []struct {
+		name       string
+		columnName string
+		columnType flux.ColType
+		data       [][]uint8
+		want       [][]values.Value
+	}{
+		{
+			name:       "ints",
```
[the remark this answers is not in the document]
nicgrobler: I will add the tests for these to this set of tests as part of the changes requested.

nicgrobler:
Hi @yzhang1991 I have implemented the changes requested, and also tried to fix some of the existing tests (there before I started adding sqlite) to use the existing RowReaders as they would be in "real life". A lot of comments were for naming style (colName instead of colname, for example) and I wanted to remove these issues to clear a lot of the noise from here, so you guys can better see what's going on. Hope this helps.

stdlib/sql/sqlite.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/values"
+)
+
+type SqliteRowReader struct {
+	Cursor      *sql.Rows
+	columns     []interface{}
+	columnTypes []flux.ColType
+	columnNames []string
+	NextFunc    func() bool
```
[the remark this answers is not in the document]
jsternberg: For sqlite3 and that specifically, it would likely be better to just use a memory source and just test using the actual library. The only reason for these in the MySQL one is because we don't want to require a MySQL server be present when running `go test`.

For sqlite, that's not really an issue.

stdlib/sql/from.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/semantic"
 	_ "github.com/lib/pq"
+	_ "github.com/mattn/go-sqlite3"
```
jsternberg: These imports will probably need to be refactored so they aren't here. That applies for all of the database drivers. The reason for this is because we sometimes want a driver to be available and sometimes we don't. When we include this library in our cloud offering, the sqlite3 connector needs to be gone because it's a security vulnerability and there's no use case for it in a cloud environment.

But for flux itself, I can see how sqlite3 would be really useful and so I like this change. But this import needs to only exist within the tests and within the `cmd/flux` package. It cannot be here.

stdlib/sql/to.go
```diff
@@ -353,6 +440,8 @@ func ExecuteQueries(tx *sql.Tx, s *ToSQLOpSpec, colNames []string, valueStrings
 	if s.DriverName != "sqlmock" {
 		_, err := tx.Exec(query, *valueArgs...)
 		if err != nil {
+			// this err which is extremely helpful as it comes from the SQL driver should be
+			// bubbled up further up the stach so user can see the issue
```
jsternberg: `stach` -> `stack`

stdlib/sql/to_test.go
```diff
@@ -1,9 +1,10 @@
 package sql_test
 
 import (
-	"github.com/influxdata/flux/plan"
 	"testing"
 
+	"github.com/influxdata/flux/plan"
+
```
jsternberg: Extra newline here. Thanks for moving it under the `testing` import though. This just needs to be grouped with the other imports below it.

stdlib/sql/from.go
jsternberg: For the moment, leave the other database drivers there and move this import to the packages I mentioned.

nicgrobler: Will get this moved - didn’t occur to me, but makes complete sense

nicgrobler: @jsternberg apologies for the question, but you are essentially talking about having the ability to build different versions of Flux yes (one with sqlite, one without, for example)? If so, then simply moving this import somewhere else won't work - as the main sql code (example shown below) shows:  
```
        // Retrieve the row reader implementation for the driver.
	var newRowReader func(rows *sql.Rows) (execute.RowReader, error)
	switch spec.DriverName {
	case "mysql":
		newRowReader = NewMySQLRowReader
	case "sqlite3":
		newRowReader = NewSqliteRowReader
	case "postgres", "sqlmock":
		newRowReader = NewPostgresRowReader
	default:
		return nil, errors.Newf(codes.Invalid, "sql driver %s not supported", spec.DriverName)
	}
```
So there would need to be a way of uncoupling the driver-specific functionality from here yes? I am sure you guys have already solved this, and have some way of "modularising" functionality so that you can pick-and-choose connectors at compile time... I just can't see where you are already doing this.

I just want to make sure I understand the approach you are taking (or want to take) so I don't waste time submitting useless garbage changes that don't do what you want ;-)

nicgrobler: ok - moved import to cmd/flux/main.go (and tests), and removed from everywhere else. commenting it out results in:
```
> sql.from(driverName:"sqlite3",dataSourceName:"hack",query:"select * from b")
Result: _result
Error: sql: unknown driver "sqlite3" (forgotten import?)
```
no other side-effects. removing comments, and it all works again as expected. Reason I am not 100% sure about this is that there appears to be no other "3rd party" imports being handled like this - so either I'm being super-dense, or so far, no other connector falls under this same treatment?

jsternberg: I created this issue anyway for ourselves since when you said that the mysql driver wasn't working, I did not feel like I could just say "no, it is definitely working" without manually checking.

Thanks for checking on that.

https://github.com/influxdata/flux/issues/2006

nicgrobler:
@yzhang1991 @jsternberg just fyi - the avalanche of failures in my build now is because I now need to add sqlite to the validator / deps (even in a barebones way). I am working on this, and will have it done soon. will then rebase. apologies.

nicgrobler:
> @yzhang1991 @jsternberg just fyi - the avalanche of failures in my build now is because I now need to add sqlite to the validator / deps (net/url can handle the dsn parsing). I am working on this, and will have it done soon. will then rebase. apologies.

All changes made - I integrated the code with the url validation, and added a few extra steps to cover sqlite-specific url checks.

nicgrobler:
Hi @yzhang1991 @jsternberg - I have marked this PR as WIP because the url_validation changes made in various places, including within the sql package **will** conflict with stuff in this code (we have both made changes to the "sql/to_test.go" file), and this way no accidental merge madness takes place.

I have tracked the commits made and the conflicts are easy enough to fix, but pretty sure you don't want to spend time picking through conflicts. When, and if, you come to this code, ping me and I will gladly pull changes and fix conflicts on my side.

yzhang1991:
Hi @nicgrobler,

Thanks for following up. I have one pending PR that I expect will land on master soon - #1989 after which I think the sql test part will be stable. I expect that to be merged by the end of today. I will ping you as soon as this merge takes place.

nicgrobler:
👍🏻No stress. Just trying not to giving you a bad day by breaking your stuff;-)

nicgrobler:
closing this PR as so messed up updating forked repo, and therefore messed up commit history. Will link to new PR so all of this history is still here, if needed.

PR #1989, opened 2019-10-14
refactor(test): url validation in transformations

yzhang1991:
Refactor the sql.to and kafka.to url validation tests so that they now share a test harness.

---
This change is [Reviewable](https://reviewable.io/reviews/influxdata/flux/1989)


execute/source.go
```diff
@@ -26,11 +26,10 @@ type Source interface {
 }
 
 type CreateSource func(spec plan.ProcedureSpec, id DatasetID, ctx Administration) (Source, error)
-type CreateNewPlannerSource func(spec plan.ProcedureSpec, id DatasetID, ctx Administration) (Source, error)
 
-var procedureToSource = make(map[plan.ProcedureKind]CreateNewPlannerSource)
+var procedureToSource = make(map[plan.ProcedureKind]CreateSource)
```
yzhang1991: `CreateSource` and `CreateNewPlannerSource` define the same function signature, and `CreateSource` has no references. This seems to be a duplicated definition for me so I removed the longer one.

Same for `CreateTransformation` below.

stdlib/kafka/to_test.go
```diff
@@ -548,77 +549,51 @@ func TestToKafka_Process(t *testing.T) {
 }
 
 func TestToKafka_NewTransformation(t *testing.T) {
-	testCases := []struct {
```
yzhang1991: There is basically no change to the test cases themselves. I just made the test cases to use the new harness.
The reviewer may check https://github.com/influxdata/flux/blob/f3e476b1ff1941ee636d53ec4506451a63b9c3ec/stdlib/kafka/to_test.go#L552 instead of trying hard to look through the diff.

stdlib/sql/to_test.go
```diff
@@ -309,82 +308,60 @@ func TestToSQL_Process(t *testing.T) {
 }
 
 func TestToSql_NewTransformation(t *testing.T) {
-	testCases := []struct {
```
yzhang1991: Same, test case did not change, just using the new harness.
https://github.com/influxdata/flux/blob/f3e476b1ff1941ee636d53ec4506451a63b9c3ec/stdlib/sql/to_test.go#L310

querytest/tf_validation.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"testing"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/dependencies/dependenciestest"
+	"github.com/influxdata/flux/dependencies/url"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/plan"
+	"github.com/stretchr/testify/assert"
```
nathanielc: So far the Flux code base doesn't use the `github.com/stretchr/testify/assert`  testing framework. It looks like it was introduced when we added the `querytest` package and we missed it. Flux uses the https://github.com/google/go-cmp testing package. Can you refactor the querytest package to use that package instead? Then we can reduce our dependencies back down to a smaller set. Thanks

PR #1991, opened 2019-10-15
fix(stdlib/experimental/prometheus): allow data collected by experime…

fchikwekwe:
This is a wrapper function to allow data collected via `experimental.prometheus.scrape()` to be used with the histogramQuantile function. The experimental wrapper takes a field argument as a string, where the user can specify an `_field` to filter by, and a quantile argument as a float. 

This change was added to experimental since it is a breaking change. The current plan is for it to be promoted in the future. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/prometheus/prometheus.flux
```diff
@@ -1,7 +1,17 @@
...
 
 // scrape enables scraping of a prometheus metrics endpoint and converts 
 // that input into flux tables. Each metric is put into an individual flux 
 // table, including each histogram and summary value.  
 builtin scrape
 
+// histogramQuantile enables the user to calculate quantiles on a set of given values
+// This function assumes that the given histogram data is being scraped or read from a 
+// Prometheus source. 
+histogramQuantile = (tables=<-, field, quantile) =>
```
wolffcm: I wonder if we should remove the `field` parameter to this altogether.  Most of our transformations require the user to do any filtering up front.  This parameter makes it so that the user cannot get a percentile of several fields at once, which they may want to do.  

If the schema of the prometheus data indicated the type of metric in the group key, then it would be useful to filter on it, e.g.: `|> filter(fn: (r) => r.metric_type == "histogram")`, so that we don't try to compute a percentile from non-histogram data.  But we don't have that, unfortunately.

@nathanielc Thoughts on this? ☝️ What are the general principles we should keep in mind with designing an API like this?

stdlib/testing/prometheus/histogramQuantile_test.flux
```diff
@@ -0,0 +1,54 @@
...
+import "testing"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,string,string,string,string,double
+#group,false,false,false,true,true,false,true,false
+#default,_result,,,,,,,
+,result,table,_time,_measurement,_field,url,le,_value
+,,0,2018-05-22T13:00:00Z,prometheus,prometheus_test_metric,http://prometheus.test,100,0.1
```
wolffcm: I would have thought that the `_value` column would be all whole numbers (despite being of type `double` which is the only scalar numeric type supported by Prometheus I guess), since it represents a count... am I missing something?

stdlib/experimental/prometheus/prometheus.flux
```diff
@@ -1,7 +1,17 @@
...
 // table, including each histogram and summary value.  
 builtin scrape
 
+// histogramQuantile enables the user to calculate quantiles on a set of given values
+// This function assumes that the given histogram data is being scraped or read from a 
+// Prometheus source. 
+histogramQuantile = (tables=<-, field, quantile) => 
+    tables
+        |> filter(fn: (r) => r._measurement == "prometheus" and r._field == field)
+        |> group(mode: "except", columns: ["le", "_value"])
```
nathanielc: Do we need to exclude `_time` as well?

stdlib/experimental/prometheus/prometheus.flux
nathanielc: I agree we should not do the filtering as part of this function. Users may want to compute the quantile of various different fields for example which can be done using correct grouping.

stdlib/experimental/prometheus/prometheus.flux
nathanielc: Can you remind me how the data comes grouped from the `scrape` call? Is `le` in the group key and that is what we are trying to fix here? If so maybe we should use the experimental.group function and subtract the `le` column from the group key but otherwise not change the group key. This will enable users to compute historgram quantiles across many different fields based on what ever grouping they setup before call this function.

stdlib/testing/prometheus/histogramQuantile_test.flux
nathanielc: I expected the same :point_up:

stdlib/experimental/prometheus/prometheus.flux
fchikwekwe: Yeah, that makes sense.

fchikwekwe: I'm not sure that `experimental.group` has the capability to subtract a column from the `groupKey` and leave all others as is. I just spoke with @wolffcm and he said there's not a mode for that yet. 

I'll open an issue to add another mode that would make it easy to remove one column from the `groupKey`, but for now I think we might have to stick with regular `group()`.

PR #2000, opened 2019-10-15
refactor(libflux): move the rust code into a single crate

### LINKED ISSUE #1994: Refactor Rust parser into a single crate
Refactor the Rust parser into a single crate named `flux`. This will be the basis behind `libflux`. The crate should allow for building both an rlib and a static library.

At the moment, the Rust parser is split into 3 different crates. This would change it so it was one crate under the single `flux` namespace.

jsternberg:
The rust code is now a single crate named `flux` and it will produce a
`libflux` file for use when we integrate it with Go.

A header file has been added to be used as the C interface that Go will
use. No other modifications, other than a few changes to make it work by
referencing internal crates, have been made to the source code.

Fixes #1994.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
It might be best if we leave some of those for when we are further integrating it with Go. The `Makefile` looks very WASM focused at the moment. I have adapted it over though anyway just in case we want to keep it and make the change later.

This doesn't remove any code in `internal/rust` yet.

nathanielc:
> This doesn't remove any code in internal/rust yet.

I think we need to do a clean break. That way we can use `git mv` to track changes for the inflight code changes.

We also need to preserve the WASM building bits.

jsternberg:
I moved it over. The WASM building bits are preserved. I just renamed stuff. I don't seem to have the necessary tools to run the commands properly so I'll have to figure that out.

PR #2012, opened 2019-10-16
refactor(values): add additional duration functions

jsternberg:
These functions will be updated with month-aware functionality and they
should be used in preference to converting the duration into a
`time.Duration`.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


values/time.go
```diff
@@ -65,6 +72,38 @@ func (t Time) Time() time.Time {
 	return time.Unix(0, int64(t)).UTC()
 }
 
+// Add will add the two Duration values together to
+// create a combined Duration.
+func (d Duration) Add(other Duration) Duration {
```
jsternberg: This method may not be valid. See #2014 for details.

jsternberg: I removed it for now.

values/time.go
```diff
@@ -65,6 +72,45 @@ func (t Time) Time() time.Time {
...
+// to a positive duration that is within the interval duration.
+func (d Duration) Normalize(every Duration) Duration {
+	offset := d
+	if offset < 0 {
+		offset += every * ((offset / -every) + 1)
+	} else if offset > every {
+		offset -= every * (offset / every)
+	}
+	return offset
+}
```
affo: I am quite not following this.

For the case `offset > every` isn't it the same as `offset % every`?

Would it be better to write tests for this?

jsternberg: Let me remove this one. It was a direct copy/paste from window and doesn't make sense without that refactor.

PR #2013, opened 2019-10-16
test(promql): update .flux PromQL tests to pass in influxdb

wolffcm:
With the exception of `histogramQuantile_test.flux`, all of these test cases were automatically updated with a program that modifies the Flux AST for the tests and then rewrites the `.flux` files.  I will make a separate PR for that tool.

1. If there was no call to `range` it was added
2. If `inData` was missing a `_measurement` column, then it was added to both `inData` and `outData` csv.

`histogramQuantile_test.flux` had a separate issue where `_time` was in the group key, and it was confusing storage.


fchikwekwe:
My incoming PR updating `histogramQuantile` and the `histogramQuantile_test.flux` file should help with the groupKey issue. 

Thanks for working on this issue. It will be good to be able to run these PromQL tests.

wolffcm:
No, I updated `histgramQuantile` so that it works now as well.

PR #2017, opened 2019-10-17
fix(stdlib): make keep/drop throw an error if merging tables with different schemas

wolffcm:
This fixes an issue that happens when we are keeping or dropping columns such that we lose a column in the group key.  When this happens we may need to merge separate tables that now have the same group key.

Previously we would not detect if the tables had different schemas and an internal error would be thrown in `AppendMappedRecordWithNulls`.  The fix here is to throw a more user-friendly error.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/schema_functions_test.go
+++ b/stdlib/universe/schema_functions_test.go
@@ -398,6 +398,162 @@ func TestDropRenameKeep_Process(t *testing.T) {
 			}},
 		},
 		{
+			name: "drop key col merge tables",
+			spec: &universe.SchemaMutationProcedureSpec{
+				Mutations: []universe.SchemaMutation{
+					&universe.DropOpSpec{
+						Columns: []string{"b"},
+					},
+				},
+			},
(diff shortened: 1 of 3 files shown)


stdlib/universe/schema_functions_test.go
```diff
@@ -429,6 +585,91 @@ func TestDropRenameKeep_Process(t *testing.T) {
...
+				Data: [][]interface{}{
+					{"one", 3.0},
+					{"one", 13.0},
+					{"one", 5.0},
+					{"one", 15.0},
+				},
+			}},
+		},
+		{
+			name: "keep one key col merge error column count",
```
yzhang1991: Should there be a test case for "keep one key col merge error column type"?

wolffcm: I had omitted it because the `keep` path is pretty much the same as the drop one, but I added it for completeness.

PR #2018, opened 2019-10-17
feat(values): turn duration value into a vector

### LINKED ISSUE #1981: Refactor values.Duration to be a struct
The `values.Duration` type is an `int64`. It needs to be refactored to also contain information for non-regular time such as months and years. An `int64` isn't going to be enough to represent this as we will need to be able to do something like the following.

```go
d := values.Duration("1y") // a duration of 1 year, this is not meant to be the code for this
ts := values.Time(0) // January 1, 1970
newTS := ts.Add(d) // January 1, 1971
```

The interim duration struct will probably be:

```go
type Duration struct {
    Nanoseconds int64
}
```

We will make further modifications to this struct to represent the parts that are needed as part of a duration.

jsternberg:
The duration value is meant to be a vector of the necessary components
for implementing a duration. This changes `values.Duration` into a
struct which will be that vector.

The only attribute implemented is nanoseconds because we have not
implemented any other dimension of the vector.

BREAKING CHANGE: JSON encoding of duration values will now be output as
string instead of an integer.

Fixes #1981.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -730,7 +730,7 @@ func (a Arguments) GetDuration(name string) (Duration, bool, error) {
 	if !ok {
 		return ConvertDuration(0), false, nil
 	}
-	return Duration(v.Duration()), true, nil
+	return v.Duration(), true, nil
 }
 
 func (a Arguments) GetRequiredDuration(name string) (Duration, error) {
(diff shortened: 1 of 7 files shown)


sanderson:
@jsternberg With this change, will we also remove `experimental.addDuration()` and `experimental.subDuration()`?

jsternberg:
@sanderson I don't think so. It appears that function adds a duration to a time value which this wouldn't effect. During our work on refactoring the type system, we should make it possible to do `now() + 1h` without the use of `addDuration` or `subDuration`.

PR #2019, opened 2019-10-17
feat(libflux): add initial c binding for parsing an AST

### LINKED ISSUE #1995: Create initial C binding to integrate the Rust parser with Go
We will need to create the initial C binding to integrate the Rust parser with Go. The initial version of this will include a call in to the parser and the ability to encode the data as JSON so it can be passed natively to Go.

It should be possible to optionally use the Rust parser by compiling with the tag `libflux`. This will allow us to begin integration without breaking any existing workflows.

jsternberg:
Fixes #1995.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
This is ready for review. It is not integrated into the CI and I'll make a separate issue for that. It works locally when I run with `-tags libflux` and it doesn't stop the current system from working. It works on both Linux and Mac OS X. It is very unlikely to work on Windows right now.

libflux/go/libflux/parser_linux.go
```diff
@@ -0,0 +1,4 @@
+package libflux
+
+// #cgo LDFLAGS: -ldl
```
yzhang1991: What is this?

libflux/go/libflux/parser.go
```diff
@@ -0,0 +1,49 @@
...
+
+// free is a utility method for calling Free
+// on a resource.
+func free(f freeable) {
+	f.Free()
+}
+
+// ASTFile is a parsed AST.
+type ASTFile struct {
+	ptr *C.struct_flux_ast_t
```
yzhang1991: How did this `struct_flux_ast_t` map to `ast::File` in Rust?

libflux/go/libflux/parser_linux.go
yzhang1991: Can I ask what library does `-ldl` include?

libflux/go/libflux/parser.go
affo: Should be the flat buffers serialization in the future I guess 🤔

yzhang1991: OK so in the PR it is just a test run and we do not look into any details in the struct.

libflux/go/libflux/parser_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestParse(t *testing.T) {
+	text := `
+package main
+
+from(bucket: "telegraf")
+	|> range(start: -5m)
+	|> mean()
+`
+	ast := libflux.Parse(text)
+	ast.Free()
```
affo: It is not clear to me why you `Free` again. Is that only to show that the operation is idempotent?

Because `Parse` sets the finalizer to free memory. That's my doubt.

libflux/go/libflux/parser.go
affo: Yeah, I think that @jsternberg wants to show us that we can call Rust from Go and properly free memory.

He demonstrates that by only using a pointer to something. In the future that should be a fbs structure. I am only guessing though and I was curious to have a look at this PR :)

libflux/go/libflux/parser_test.go
jsternberg: It's mostly because I didn't have anything to test. I wanted to show that the methods worked properly without panicking or, in the case of C code, it would segfault. Also, the garbage collector is unlikely to run before the test is over so the memory would not be freed. At the same time, the memory not being freed when the program ends is not a big deal. It's mostly the only reason finalizers exist.

libflux/go/libflux/parser.go
jsternberg: It maps to this definition in the C file.

```
struct flux_ast_t;
```

The C binding generated by Go will map that to the `C.struct_flux_ast_t` type. Since the struct is a forward declaration, the only thing it can be used for is a pointer. When you're writing C bindings, this is common to create handles that can be passed back into the library and to avoid leaking the internal details of the implementation since the attributes in the struct become part of the public ABI.

libflux/go/libflux/parser_linux.go
jsternberg: It's `libdl` which manages `dlopen`, `dlsym`, and `dlclose`. For whatever reason, this was required on Linux and wasn't required on Mac OS X. The Rust library appeared to use it for whatever reason.

libflux/go/libflux/parser.go
jsternberg: To clarify, the implementation is very likely to always be an empty struct. It may be possible in certain circumstances to expose a struct's inner details. It is possible to tell Rust how to arrange the memory so the C definition and the Rust definition share the same memory layout, but it is probably not something we should use too frequently.

We would use the handles to access the data. See this [gist](https://gist.github.com/jsternberg/3adac9a7b3386d7eb53846e47955604b#sharing-memory) for a rough outline of what I mean.

PR #2020, opened 2019-10-18
refactor(libflux/ast): rename AST enums to be consistent with type names

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/ast/mod.rs
+++ b/libflux/src/ast/mod.rs
@@ -5,7 +5,7 @@ use std::vec::Vec;
 
 use crate::scanner;
 
-use chrono::prelude::DateTime;
+use chrono;
 use chrono::FixedOffset;
 
 use serde::de::{Deserialize, Deserializer, Error, Visitor};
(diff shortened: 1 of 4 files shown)


libflux/src/ast/mod.rs
```diff
@@ -96,62 +96,62 @@ where
...
+    Unary(Box<UnaryExpr>),
+    PipeExpr(Box<PipeExpr>),
+    Call(Box<CallExpr>),
+    Conditional(Box<ConditionalExpr>),
+    StringExpr(Box<StringExpr>),
+    Paren(Box<ParenExpr>),
+
+    Integer(IntegerLit),
+    Float(FloatLit),
+    StringLit(StringLit),
```
jlapacik: `String(StringLit)`?

nathanielc: There are StringLit and StringExpr types so those would conflict so I used the long names on those constructors. Same goes for PipeLit and PipeExpr.

nathanielc:
@jlapacik I answered above, but it was not a conflict with Rust `String` type rather a conflict between our ast type `StringLit` and `StringExpr`.

PR #2021, opened 2019-10-18
feat(rust/ast): add walk implementation

nathanielc:
Needs #2020 to merge first

Adds visitor/walk pattern to walk the Rust AST structures.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/src/ast/walk/mod.rs
```diff
@@ -0,0 +1,363 @@
+#[cfg(test)]
+mod tests;
+
+use crate::ast::*;
+use std::cell::{RefCell, RefMut};
+use std::rc::Rc;
+
+// Node represents any structure that can appear in the AST
+#[derive(Debug)]
+pub enum Node<'a> {
```
jsternberg: This seems like it should be in the `ast` module.

libflux/src/ast/walk/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+// When used with the walk function, Visit will be called for every node
+// in depth-first order. After all children for a Node have been visted,
+// Done is called on that Node to signal that we are done with that Node.
+//
+// If Visit returns None, walk will not recurse on the children.
+pub trait Visitor: Sized {
+    // Visit is called for a node.
+    // The returned visitor will be used to walk children of the node.
+    // If visit returns None, walk will not recurse on the children.
+    fn visit(&self, node: &Node) -> Option<Self>;
```
jsternberg: To clarify, the reason why this works is because if you attempt to do this with a unique pointer, it won't work correctly. So you instantiate the visitor trait using an `Rc` version of the type instead?

libflux/src/ast/walk/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            Statement::Expr(s) => Node::ExprStmt(s),
+            Statement::Variable(s) => Node::VariableAssgn(s),
+            Statement::Option(s) => Node::OptionStmt(s),
+            Statement::Return(s) => Node::ReturnStmt(s),
+            Statement::Bad(s) => Node::BadStmt(s),
+            Statement::Test(s) => Node::TestStmt(s),
+            Statement::Builtin(s) => Node::BuiltinStmt(s),
+        }
+    }
+    fn from_func_body(fb: &FunctionBody) -> Node {
```
jsternberg: Can you spell these out to match with the type name? `from_function_body`, `from_statement`, `from_property_key`, etc?

libflux/src/ast/walk/mod.rs
nathanielc: Yes, after thinking it through the visitor trait itself doesn't care how you do it but you need to return an owned visitor from the visit function. Therefore if you want to reuse yourself you need multiple ownership  hence using `Rc` reference counting. 

I also ended up using RefCell to hide the mutability of the closure since its not part of the visitor itself.

libflux/src/ast/walk/mod.rs
nathanielc: It is not needed in the `ast` module and is only used here so I thought it best to just leave it here. The `ast` doesn't need a generic concept of a `Node` because the use of enums means that we can always be specific about what ast nodes can go where. There is never a branch of the AST were all nodes are possible.

jsternberg: Sure. We'll refactor later if this isn't the case. I approved so we're good.

libflux/src/ast/walk/tests.rs
```diff
@@ -0,0 +1,306 @@
...
+    let file = parse_string("test_walk", source);
+    let mut nodes = Vec::new();
+    walk(
+        &create_visitor(&mut |n| nodes.push(format!("{}", n))),
+        Node::File(&file),
+    );
+    assert_eq!(want, nodes);
+}
+
+#[test]
```
jlapacik: I believe unit tests usually go in the same file as the code they are testing.

libflux/src/ast/walk/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    // Property
+    Property(&'a Property),
+
+    // StringExprPart
+    TextPart(&'a TextPart),
+    InterpolatedPart(&'a InterpolatedPart),
+
+    // Assignment
+    VariableAssgn(&'a VariableAssgn),
```
goller: Is `Assgn` an intentional spelling?  Should this be Assign?

libflux/src/ast/walk/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            Expression::Paren(e) => Node::ParenExpr(e),
+            Expression::Integer(e) => Node::IntegerLit(e),
+            Expression::Float(e) => Node::FloatLit(e),
+            Expression::StringLit(e) => Node::StringLit(e),
+            Expression::Duration(e) => Node::DurationLit(e),
+            Expression::Uint(e) => Node::UintLit(e),
+            Expression::Boolean(e) => Node::BooleanLit(e),
+            Expression::DateTime(e) => Node::DateTimeLit(e),
+            Expression::Regexp(e) => Node::RegexpLit(e),
+            Expression::PipeLit(e) => Node::PipeLit(e),
```
goller: Seems like the other expression literals do not keep the `Lit` suffix. Should this be `Expression:Pipe`? or perhaps the other literals should be suffixed?

libflux/src/ast/walk/mod.rs
nathanielc: Its an intentional spelling of `Assignment`, open to suggestions on a better abbreviation on that one.

libflux/src/ast/walk/mod.rs
nathanielc: This is an exception, there are two Pipe types one is `PipeLit` the other is `PipeExpr` so I couldn't use the bare `Pipe` constructor name.  So I used the full name in this case. 

Same for `StringLit` and `StringExpr`.

libflux/src/ast/walk/mod.rs
goller: I'm thinking `Assign`

when I role-play talking about it:

"It's a Variable Assign"

libflux/src/ast/walk/tests.rs
nathanielc: They can be in either, for really long tests its nice to put them in their own file so the files are easier to navigate. These tests are not too long so I could move them back. I was following the convention of the parser and ast modules, which do have very large tests.

Issue #1953, opened 2019-10-09
Define auxiliary data structures for type inference


jlapacik:
Define the following data structures for type inference:

* Type Environment
A type environment is a map of in scope variable names to their assigned polytype. Inference will add a new entry to the environment for each variable assignment. It will also perform a lookup for each identifier expression that it encounters during the inference process.

* Substitution
A substitution is map of type variables to equivalent monotypes. A substitution is produced when unifying two monotypes. Substitutions are applied to monotypes, polytypes, and type environments. A valid substitution is one in which no key (or type variable) in the substitution is referenced by any of the values in the substitution. A substitution must always preserve this invariant. And finally substitutions can and will be merged together during inference, eventually producing a substitution for the entire program.

* Constraints
There are two types of constraints, equality and kind constraints. An equality constraint asserts that two monotypes are equal and will be unified at some point, producing a substitution. A kind constraint asserts that a monotype belongs to a certain type class. 

* Constraint Map
A constraint map is a map of type variables to kinds. A type variable that is present in the constraint map during inference must not be present as a key in the global substitution. In the case that a type variable is assigned to a monotype in the global substitution, that type variable must be removed from the constraint map if present.

affo:
@jlapacik 

> A valid substitution is one in which no key (or type variable) in the substitution is referenced by any of the values in the substitution

Is this to avoid cyclic substitutions by construction?

Say I have

```
t0 -> Int
```

And then I find

```
t0 + t1
```

This would add a contraint like 

```
t1 -> t0
```

But it cannot be in the substitution like that, because we would violate the invariant, so, we should directly "expand" the substitution:

```
t0 -> Int
t1 -> Int
```

How far I got from something reasonable? 😅

affo:
@jlapacik 

>  A type variable that is present in the constraint map during inference must not be present as a key in the global substitution. In the case that a type variable is assigned to a monotype in the global substitution, that type variable must be removed from the constraint map if present.

Let me rephrase to understand if I got it. You are saying that once you have a valid substitution for a type variable that matches some kind, then there is no sense in keeping a constraint on that type variable. That would only make the constraint map heavier.

jlapacik:
@affo you are correct in both cases. For substitutions, an equivalent definition is that substitutions must be idempotent. That is, for any substitution `s` and any type `x`, `s(x) = s(s(x))`.

[...]

PR #2025, opened 2019-10-18
Type inference plumbing

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

#1953 .

This PR does several things.

1. It defines empty structs for each of the basic types. Previously each of the basic types (int, float, string, ...) were just variants of the `MonoType` enum.

2. Defines the data structure for type substitutions as well as the `Substitutable` trait.

3. Defines the data structure for mapping type variables to their kind constraints.

4. Defines the `Constraint` enum which represents either a type constraint or a kind constraint. Type inference will produce a set of these constraints for each expression.

### DIFF
--- a/libflux/src/semantic/infer.rs
+++ b/libflux/src/semantic/infer.rs
@@ -1,0 +1,56 @@
+use crate::semantic::types::{Kind, MonoType};
+use std::ops;
+
+// Type constraints are produced during type inference and come
+// in two flavors.
+//
+// A kind constraint asserts that a particular type is of a
+// particular kind or family of types.
+//
+// An equality contraint asserts that two types are equivalent
+// and will be unified at some point.
+//
+#[derive(Debug, PartialEq)]
+enum Constraint {
+    Kind(MonoType, Kind),
+    Equal(MonoType, MonoType),
+}
+
+#[derive(Debug, PartialEq)]
+struct Constraints(Vec<Constraint>);
+
+// Constraints can be added using the '+' operator
+impl ops::Add for Constraints {
+    type Output = Constraints;
+
+    fn add(self, cons: Constraints) -> Self::Output {
+        Constraints(self.0.into_iter().chain(cons.0.into_iter()).collect())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::semantic::types::Tvar;
+
+    #[test]
+    fn add_constraints() {
+        let c0 = Constraints(vec![
+            Constraint::Equal(MonoType::Var(Tvar(0)), MonoType::Var(Tvar(1))),
+            Constraint::Kind(MonoType::Var(Tvar(1)), Kind::Addable),
+        ]);
+        let c1 = Constraints(vec![
+            Constraint::Equal(MonoType::Var(Tvar(2)), MonoType::Var(Tvar(3))),
+            Constraint::Kind(MonoType::Var(Tvar(3)), Kind::Divisible),
+        ]);
+        assert_eq!(
+            c0 + c1,
+            Constraints(vec![
+                Constraint::Equal(MonoType::Var(Tvar(0)), MonoType::Var(Tvar(1))),
+                Constraint::Kind(MonoType::Var(Tvar(1)), Kind::Addable),
+                Constraint::Equal(MonoType::Var(Tvar(2)), MonoType::Var(Tvar(3))),
+                Constraint::Kind(MonoType::Var(Tvar(3)), Kind::Divisible),
+            ])
+        );
+    }
+}
--- a/libflux/src/semantic/sub.rs
+++ b/libflux/src/semantic/sub.rs
@@ -1,0 +1,50 @@
+use crate::semantic::types::{MonoType, Tvar};
+use std::{collections::HashMap, fmt};
+
+// A substitution defines a function that takes a monotype as input
+// and returns a monotype as output. The output type is interpreted
+// as being equivalent to the input type.
+//
+// Substitutions are idempotent. Given a substitution s and an input
+// type x, we have s(s(x)) = s(x).
+//
+#[derive(Debug, PartialEq)]
+pub struct Subst(HashMap<Tvar, MonoType>);
+
+impl fmt::Display for Subst {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("substitution:\n")?;
+        for (k, v) in &self.0 {
+            write!(f, "\t{}: {}\n", k, v)?;
+        }
+        Ok(())
+    }
+}
+
+impl Subst {
+    pub fn empty() -> Subst {
+        Subst(HashMap::new())
+    }
+
+    pub fn init(values: HashMap<Tvar, MonoType>) -> Subst {
+        Subst(values)
+    }
+
+    pub fn lookup(&self, tv: Tvar) -> Option<&MonoType> {
+        self.0.get(&tv)
+    }
+
+    pub fn merge(self, with: Subst) -> Subst {
+        let applied: HashMap<Tvar, MonoType> = self
+            .0
+            .into_iter()
+            .map(|(k, v)| (k, v.apply(&with)))
+            .collect();
+        Subst(applied.into_iter().chain(with.0.into_iter()).collect())
+    }
+}
+
+// A type is substitutable if a substitution can be applied to it.
+pub trait Substitutable {
+    fn apply(self, sub: &Subst) -> Self;
+}
--- a/libflux/src/semantic/types.rs
+++ b/libflux/src/semantic/types.rs
@@ -1,4 +1,9 @@
-use std::{cmp, collections::HashMap, fmt};
+use crate::semantic::sub::{Subst, Substitutable};
+use std::{
+    cmp,
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    fmt, result,
+};
 
 // PolyType represents a generic parametrized type.
 //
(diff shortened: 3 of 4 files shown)


libflux/src/semantic/infer.rs
```diff
@@ -0,0 +1,49 @@
...
+        let c0 = Constraints(vec![
+            Constraint::Equal(MonoType::Var(Tvar(0)), MonoType::Var(Tvar(1))),
+            Constraint::Kind(MonoType::Var(Tvar(1)), Kind::Addable),
+        ]);
+        let c1 = Constraints(vec![
+            Constraint::Equal(MonoType::Var(Tvar(2)), MonoType::Var(Tvar(3))),
+            Constraint::Kind(MonoType::Var(Tvar(3)), Kind::Divisible),
+        ]);
+        assert_eq!(
+            c0 + c1,
```
affo: Oh gosh, so cool. Summing up constraints

libflux/src/semantic/types.rs
```diff
@@ -116,27 +209,490 @@ impl fmt::Display for MonoType {
...
+    }
+}
+
+impl Unifiable for MonoType {
+    fn unify(self, with: Self, cons: Const) -> Result {
+        match self {
+            MonoType::Bool(t) => {
+                if let MonoType::Var(tv) = with {
+                    tv.unify(self, cons)
+                } else if let MonoType::Bool(s) = with {
```
nathanielc: My reading of this code says that the:

```
if let MonoType::Var(tv) = with {
                    tv.unify(self, cons)
}
```

code can be moved out of the `match self` block. Maybe this is because of the `Self` reference in the `unifiable` trait again?

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
     }
 }
 
+// Kind constraints on type variables.
+//
+// Ex. forall [t0] where t0:Comparable + Equatable (x:t0, y:t0) -> bool
+//                       ^_______________________^
+//
+#[derive(Debug, Clone, PartialEq)]
+pub struct Const {
```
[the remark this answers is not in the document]
jlapacik: Maybe? I have 3 different structs with some version of Constraint in their names, so I definitely need a better naming scheme.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+// the types are equal after application of the substitution.
+//
+// More formally, s = unify(x,y) => s(x) = s(y) .
+//
+// Note here that unification also takes a set of constrained type
+// variables and produces a possibly updated set of constraints along
+// with the substitution equating the two types.
+//
+trait Unifiable {
+    fn unify(self, with: Self, cons: Const) -> Result;
```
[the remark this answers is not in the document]
jlapacik: Yes `Self` refers to the implementing type. So you could **not** pass an integer to the unify method of `Bool`.

libflux/src/semantic/types.rs
```diff
@@ -75,22 +155,35 @@ impl fmt::Display for Kind {
...
             Kind::Comparable => f.write_str("Comparable"),
+            Kind::Equatable => f.write_str("Equatable"),
             Kind::Nullable => f.write_str("Nullable"),
         }
     }
 }
 
+impl cmp::Ord for Kind {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
+        self.to_string().cmp(&other.to_string())
```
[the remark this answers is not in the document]
jlapacik: Yes there should be a comment here. This is for display purposes only.

libflux/src/semantic/types.rs
```diff
@@ -75,22 +155,35 @@ impl fmt::Display for Kind {
...
+    }
+}
+
+impl cmp::PartialOrd for Kind {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
```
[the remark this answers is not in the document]
jlapacik: Yes sorry for the confusion. Same reason as above. Display purposes only. Specifically displaying the kind constraints of a polytype's free variables.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+trait Constrainable {
+    fn constrain(self, with: Kind, cons: Const) -> Result;
+}
+
+type Result = result::Result<(Subst, Const), Error>;
+
+#[derive(Debug)]
+enum Error {
+    Constrain { t: MonoType, with: Kind },
+    Unify { t: MonoType, with: MonoType },
```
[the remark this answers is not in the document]
jlapacik: Yes sorry this wasn't clear. I thought constrain was better verbiage for kinds since unify connotes equality more or less. I guess the doc comment for `Constrainable` didn't explain this very well?

I agree I have too many names with Constraint in them. I'll try to come up with more descriptive ones.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+//
+// Unlike unification which asserts an equality constraint on two
+// types, a kind constraint asserts the containment of a type within
+// a type class or family.
+//
+trait Constrainable {
+    fn constrain(self, with: Kind, cons: Const) -> Result;
+}
+
+type Result = result::Result<(Subst, Const), Error>;
```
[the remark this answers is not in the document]
jlapacik: Yes you're right this probably does need to be public.

Unification returns a substitution that is a unifier for two types. But when we unify types we may also discover partial information about type variables in the form of kind constraints. This is why we return both.

libflux/src/semantic/types.rs
```diff
@@ -116,27 +209,490 @@ impl fmt::Display for MonoType {
...
+
+impl Substitutable for Bool {
+    fn apply(self, _: &Subst) -> Self {
+        self
+    }
+}
+
+impl Unifiable for Bool {
+    fn unify(self, _: Self, cons: Const) -> Result {
+        Ok((Subst::empty(), cons))
```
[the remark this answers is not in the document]
jlapacik: You raise a good point. I guess this begs the question, does `Unifiable` make sense as a trait at all? Unlike `Substitutable` it's not implemented on any external types. Perhaps it's just confusing?

I also defined empty structs for the basic types like `Int`, `Float`, etc. I did this to isolate each type's unification logic. But does it make sense to have a `struct Bool;` type, or should there just be a single `MonoType::Bool` variant with the unification logic for `Bool` implemented as part of `MonoType::unify`?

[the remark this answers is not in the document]
nathanielc: > I also defined empty structs for the basic types like Int, Float, etc. I did this to isolate each type's unification logic. But does it make sense to have a struct Bool; type, or should there just be a single MonoType::Bool variant with the unification logic for Bool implemented as part of MonoType::unify?

This is what I expected when I first saw the MonoType definition. I was confused at first why the basic types were being reintroduced as their own struct. 

Maybe try it out? Remove the trait and see if it simplifies the logic?

libflux/src/semantic/types.rs
affo: Yep, I agree.

@jlapacik I also think that two nested `match` expression would make the code more readable.

libflux/src/semantic/types.rs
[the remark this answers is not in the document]
affo: @jlapacik I like your proposal of just using variants and implement the logic directly into `Monotype`.

libflux/src/semantic/types.rs
```diff
@@ -240,82 +821,114 @@ pub struct Function {
...
+                Some(p) => Some(p.apply(sub)),
+            },
+            retn: self.retn.apply(sub),
+        }
+    }
+}
+
+impl Unifiable for Function {
+    fn unify(self, _: Self, _: Const) -> Result {
+        unimplemented!();
```
affo: those `unimplemented!` are left for future implementation in another PR?

libflux/src/semantic/infer.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::ops;
+
+#[derive(Debug, PartialEq)]
+enum Constraint {
+    Kind(MonoType, Kind),
+    Equal(MonoType, MonoType),
+}
+
+#[derive(Debug, PartialEq)]
+struct Constraints(Vec<Constraint>);
```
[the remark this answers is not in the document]
jlapacik: If it's just an alias to an external type, you can't implement any external traits on it. I wouldn't be able to implement `ops::Add` or `fmt::Display`.

libflux/src/semantic/sub.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// A Substitution is implemented as a map from type variables
+// to monotypes.
+//
+// Note this implementation implies that substitutions are
+// idempotent. Given a monotype x, and a substitution s, s(s(x)) = s(x).
+//
+#[derive(Debug)]
+pub struct Subst(HashMap<Tvar, MonoType>);
```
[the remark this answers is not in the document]
jlapacik: Same as above. I implemented the `fmt::Display` trait for substitutions. If it weren't for that implementation you are right, this could be just an alias. The Go code has methods for displaying substitutions and constraint sets. I found them to be quite useful when debugging.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+    pub kinds: HashMap<Tvar, HashSet<Kind>>,
+}
+
+impl fmt::Display for Const {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        DisplayList {
+            values: self
+                .kinds
+                .iter()
+                .collect::<BTreeMap<_, _>>()
```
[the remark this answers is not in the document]
jlapacik: By collecting into a BTree we obtain a sorted iterator. This is for consistent(vs random) displaying of output.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+}
+
+impl<'a> fmt::Display for TvarConst<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{}:{}",
+            self.tv,
+            DisplayList {
+                values: self.kinds.iter().collect::<BTreeSet<_>>().iter().collect(),
```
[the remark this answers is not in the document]
jlapacik: Same as above. A little trick to obtain sorted output.

libflux/src/semantic/types.rs
```diff
@@ -58,13 +49,102 @@ impl fmt::Display for PolyType {
...
+//                       ^_______________________^
+//
+#[derive(Debug, Clone, PartialEq)]
+pub struct Const {
+    pub kinds: HashMap<Tvar, HashSet<Kind>>,
+}
+
+impl fmt::Display for Const {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        DisplayList {
```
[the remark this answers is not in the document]
jlapacik: I completely overlooked `join`. That's exactly what I was looking for. Thank you.

libflux/src/semantic/types.rs
jlapacik: Yes correct.

jlapacik:
@nathanielc @affo thank you for the reviews. I will implement the unification for type variables so that it's more clear how these methods interact with each other.

libflux/src/semantic/types.rs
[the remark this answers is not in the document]
affo: Ooooo, got it!

libflux/src/semantic/types.rs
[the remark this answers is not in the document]
affo: Great! Can you put a comment, just to be explicit?

If you directly use that data structures you loose performance on insert/update right? From O(1) to O(logn).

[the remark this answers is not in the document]
jlapacik: Absolutely. And yes insert is O(logn). I'm not too concerned with the performance characteristics of display though.

libflux/src/semantic/infer.rs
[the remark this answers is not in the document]
affo: That's a good point, I thought you could, but _an alias is just an alias_ 😅

jlapacik:
I have updated the comments and removed unnecessary traits. I also removed the empty structs for basic types. All the unification logic for basic types is now part of `MonoType`. I also implemented unification for type variables in the last commit with some test cases.

libflux/src/semantic/types.rs
```diff
@@ -212,133 +191,71 @@ impl Substitutable for MonoType {
...
-                    t.unify(*fun, cons)
-                } else {
-                    Err(Error::Unify {
-                        t: MonoType::Fun(t),
-                        with,
-                    })
-                }
-            }
-            MonoType::Var(t) => t.unify(with, cons),
+        match (self, with) {
```
affo: Boom combinatory `match`. Didn't know about this

libflux/src/semantic/types.rs
```diff
@@ -116,27 +205,251 @@ impl fmt::Display for MonoType {
...
+        self,
+        kinds: HashSet<Kind>,
+    ) -> impl FnOnce(HashSet<Kind>, Const) -> Result {
+        move |k, c| {
+            if k.is_subset(&kinds) {
+                Ok((Subst::empty(), c))
+            } else {
+                Err(Error::Constrain {
+                    t: self,
+                    with: k.difference(&kinds).map(|&kind| kind).collect(),
```
nathanielc: what does the `map` call do here? I looks like the identity as far as I can tell.

libflux/src/semantic/types.rs
```diff
@@ -116,27 +205,251 @@ impl fmt::Display for MonoType {
...
+        }
+    }
+
+    fn unify_with_tvar(self, tv: Tvar, cons: Const) -> Result {
+        let mut cons = cons;
+        // Gather the kind constraints for both type variables
+        let kinds: HashSet<Kind> = cons
+            .0
+            .get(&self)
+            .unwrap_or(&HashSet::new())
```
nathanielc: nit, in reading up on what this does, I discovered that `unwrap_or` is evaluated eagerly so we are creating a new hashset with each call to `unify_with_tvar`. The docs recommend using `unwrap_or_else` which accepts a closure which is evaluated lazily.

libflux/src/semantic/types.rs
```diff
@@ -116,27 +205,251 @@ impl fmt::Display for MonoType {
...
+            .union(cons.0.get(&tv).unwrap_or(&HashSet::new()))
+            .map(|&kind| kind)
+            .collect();
+        let sub = Subst::init(maplit::hashmap! {self => MonoType::Var(tv)});
+        if kinds.len() > 0 {
+            // Update the kind constraints
+            cons.0.remove(&self);
+            cons.0.remove(&tv);
+            cons.0.insert(tv, kinds);
```
nathanielc: insert will replace the existing value, why do we remove it first?

jlapacik: Typo

libflux/src/semantic/types.rs
jlapacik: I must have missed `unwrap_or_else`. I will update. Note however that while this creates a new hashset value, it won't actually allocate any heap memory until we insert the first element.

libflux/src/semantic/types.rs
jlapacik: Basically. The error enum owns the kinds passed into it, so I have to collect a set of kinds vs a set of references to kinds. The error enum could probably use a little more work.

nathanielc: ok so it converts `[Kind]` to `[&Kind]`? Should we instead use `&[Kind]`?

libflux/src/semantic/types.rs
nathanielc: That is good to know, I see that in the docs now. In that case i say this reads more clearly so we should leave it as `unwrap_or` since we do not need to avoid the allocation.

libflux/src/semantic/types.rs
jlapacik: It converts `[&Kind]` to `[Kind]`. I think it makes sense for errors to consume their kinds. The pattern is that kinds are owned, they're not referenced, so the same should hold for errors.

libflux/src/semantic/types.rs
```diff
@@ -14,57 +19,138 @@ use std::{cmp, collections::HashMap, fmt};
...
+// Const is used to keep track of the generic constraints for each type
+// variable during type inference.
+//
+// Note that during inference we might infer that a type variable is of
+// a particular kind (type class) without inferring an exact monotype
+// for said type variable. This information is stored in the Const data
+// structure.
+//
+#[derive(Debug, Clone, PartialEq)]
+pub struct Const(HashMap<Tvar, HashSet<Kind>>);
```
affo: Why not `Constraints`? `Const` makes me think about "constant".

libflux/src/semantic/types.rs
```diff
@@ -116,9 +217,136 @@ impl fmt::Display for MonoType {
...
+                with, cons
+            ),
+            MonoType::Int => self.satisfies_at_most(maplit::hashset![
+                Kind::Addable,
+                Kind::Subtractable,
+                Kind::Divisible,
+                Kind::Comparable,
+                Kind::Equatable,
+                Kind::Nullable
+            ])(with, cons),
```
affo: It seems like every time you constrain something a new hashset gets created. Could we store those constant hashsets in something generated once for all?

libflux/src/semantic/types.rs
```diff
@@ -116,9 +217,136 @@ impl fmt::Display for MonoType {
...
+            (MonoType::Var(tv), t) => tv.unify(t, cons),
+            (t, MonoType::Var(tv)) => tv.unify(t, cons),
+            (MonoType::Arr(t), MonoType::Arr(s)) => t.unify(*s, cons),
+            (MonoType::Row(t), MonoType::Row(s)) => t.unify(*s, cons),
+            (MonoType::Fun(t), MonoType::Fun(s)) => t.unify(*s, cons),
+            (t, with) => Err(Error::cannot_unify(t, with)),
+        }
+    }
+
+    fn constrain(self, with: HashSet<Kind>, cons: Const) -> Result {
```
affo: Why do you need `cons: Const` here? Just to return a `Result`? 🤔

libflux/src/semantic/types.rs
```diff
@@ -141,16 +369,109 @@ impl Fresher {
...
+            // The monotype must satisfy all the constraints of
+            // the type variable.
+            let (s, cons) = t.constrain(kinds, cons)?;
+            Ok((sub.merge(s), cons))
+        } else {
+            Ok((sub, cons))
+        }
+    }
+
+    fn constrain(self, with: HashSet<Kind>, cons: Const) -> Result {
```
affo: Ok, I see. You need `Const` above because you need it here, right?

But there is no trait to implement, so... 🤔

libflux/src/semantic/types.rs
```diff
@@ -81,19 +82,51 @@ fn display_kinds(tv: Tvar, kinds: &HashSet<Kind>) -> String {
     )
 }
 
+impl Const {
+    fn update(self, tv: Tvar, with: HashSet<Kind>) -> Self {
+        let mut constraints = self;
+        if let Some(kinds) = constraints.0.get(&tv) {
+            let new: HashSet<Kind> = kinds.union(&with).map(|&kind| kind).collect();
```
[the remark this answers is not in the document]
jlapacik: Unfortunately no. I believe the compiler said it's missing the `From` trait.

libflux/src/semantic/types.rs
jlapacik: There's already a `Constraints` struct for the constraints that inference generates. Maybe `TvarKinds`?

libflux/src/semantic/types.rs
jlapacik: It's also needed for Array, Row, and Function.

libflux/src/semantic/types.rs
jlapacik: This could be handled more efficiently using nested match expressions. This ways is nice because all of the constraint and error logic is in a single closure. Let's optimize this later.

PR #2027, opened 2019-10-18
feat(internal/rust): add semantic analysis

### LINKED ISSUE #1955: Define semantic nodes in Rust
This issue is to port the semantic graph (currently defined in Go) over to Rust.

Note, semantic expressions should have their monotype associated with them. An initial type variable should be given to each expression when generating a semantic graph from an AST. Each expression node should be updated with its inferred type using the final substitution obtained after type inference.

### LINKED ISSUE #1972: Convert Rust AST to Rust Semantic Graph
Generate the semantic graph from the AST. Each node should be annotated with a type variable as its mono type.

Test cases will be of the form AST -> Semantic Graph

affo:
Closes #1972 
Closes #1955  

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/semantic/analyze.rs
+++ b/libflux/src/semantic/analyze.rs
@@ -1,0 +1,2576 @@
+use crate::ast;
+use crate::semantic::nodes::*;
+use crate::semantic::types::{Fresher, MonoType};
+use std::result;
+
+type SemanticError = String;
+type Result<T> = result::Result<T, SemanticError>;
+
+// analyze analyzes an AST package node and returns its semantic analysis.
+// The function requires a Fresher to instantiate fresh type veriables for Expressions.
+// The function explicitly moves the ast::Package because it adds information to it.
+// We follow here the principle that every compilation step should be isolated and should add meaning
+// to the previous one. In other terms, once one analyzes an AST he should not use it anymore.
+// If one wants to do so, he should explicitly pkg.clone() and incur consciously in the memory
+// overhead involved.
+pub fn analyze(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
+    analyze_package(pkg, fresher)
+    // TODO(affo): run checks on the semantic graph.
+}
+
+fn analyze_package(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
+    let files = pkg
+        .files
+        .into_iter()
+        .map(|f| analyze_file(f, fresher))
+        .collect::<Result<Vec<File>>>()?;
+    Ok(Package {
+        loc: pkg.base.location,
+        package: pkg.package,
+        files: files,
+    })
+}
+
+fn analyze_file(file: ast::File, fresher: &mut Fresher) -> Result<File> {
+    let package = analyze_package_clause(file.package, fresher)?;
+    let imports = file
+        .imports
+        .into_iter()
+        .map(|i| analyze_import_declaration(i, fresher))
+        .collect::<Result<Vec<ImportDeclaration>>>()?;
+    let body = file
+        .body
+        .into_iter()
+        .map(|s| analyze_statement(s, fresher))
+        .collect::<Result<Vec<Statement>>>()?;
+    Ok(File {
+        loc: file.base.location,
+        package,
+        imports,
+        body,
+    })
+}
+
+fn analyze_package_clause(
+    pkg: Option<ast::PackageClause>,
+    fresher: &mut Fresher,
+) -> Result<Option<PackageClause>> {
+    if pkg.is_none() {
+        return Ok(None);
+    }
+    let pkg = pkg.unwrap();
+    let name = analyze_identifier(pkg.name, fresher)?;
+    Ok(Some(PackageClause {
+        loc: pkg.base.location,
+        name,
+    }))
+}
+
+fn analyze_import_declaration(
+    imp: ast::ImportDeclaration,
+    fresher: &mut Fresher,
+) -> Result<ImportDeclaration> {
+    let alias = match imp.alias {
+        None => None,
+        Some(id) => Some(analyze_identifier(id, fresher)?),
+    };
+    let path = analyze_string_literal(imp.path, fresher)?;
+    Ok(ImportDeclaration {
+        loc: imp.base.location,
+        alias,
+        path,
+    })
+}
+
+fn analyze_statement(stmt: ast::Statement, fresher: &mut Fresher) -> Result<Statement> {
+    match stmt {
+        ast::Statement::Option(s) => Ok(Statement::Option(analyze_option_statement(s, fresher)?)),
+        ast::Statement::Builtin(s) => {
+            Ok(Statement::Builtin(analyze_builtin_statement(s, fresher)?))
+        }
+        ast::Statement::Test(s) => Ok(Statement::Test(analyze_test_statement(s, fresher)?)),
+        ast::Statement::Expr(s) => Ok(Statement::Expr(analyze_expression_statement(s, fresher)?)),
+        ast::Statement::Return(s) => Ok(Statement::Return(analyze_return_statement(s, fresher)?)),
+        // TODO(affo): we should fix this to include MemberAssignement.
+        //  The error lies in AST: the Statement enum does not include that.
+        //  This is not a problem when parsing, because we parse it only in the option assignment case,
+        //  and we return an OptionStmt, which is a Statement.
+        ast::Statement::Variable(s) => Ok(Statement::Variable(analyze_variable_assignment(
+            s, fresher,
+        )?)),
+        ast::Statement::Bad(_) => {
+            Err("BadStatement is not supported in semantic analysis".to_string())
+        }
+    }
+}
+
+fn analyze_assignment(assign: ast::Assignment, fresher: &mut Fresher) -> Result<Assignment> {
+    match assign {
+        ast::Assignment::Variable(a) => Ok(Assignment::Variable(analyze_variable_assignment(
+            a, fresher,
+        )?)),
+        ast::Assignment::Member(a) => {
+            Ok(Assignment::Member(analyze_member_assignment(a, fresher)?))
+        }
+    }
+}
+
+fn analyze_option_statement(stmt: ast::OptionStmt, fresher: &mut Fresher) -> Result<OptionStmt> {
+    Ok(OptionStmt {
+        loc: stmt.base.location,
+        assignment: analyze_assignment(stmt.assignment, fresher)?,
+    })
+}
+
+fn analyze_builtin_statement(stmt: ast::BuiltinStmt, fresher: &mut Fresher) -> Result<BuiltinStmt> {
+    Ok(BuiltinStmt {
+        loc: stmt.base.location,
+        id: analyze_identifier(stmt.id, fresher)?,
+    })
+}
+
+fn analyze_test_statement(stmt: ast::TestStmt, fresher: &mut Fresher) -> Result<TestStmt> {
+    Ok(TestStmt {
+        loc: stmt.base.location,
+        assignment: analyze_variable_assignment(stmt.assignment, fresher)?,
+    })
+}
+
+fn analyze_expression_statement(stmt: ast::ExprStmt, fresher: &mut Fresher) -> Result<ExprStmt> {
+    Ok(ExprStmt {
+        loc: stmt.base.location,
+        expression: analyze_expression(stmt.expression, fresher)?,
+    })
+}
+
+fn analyze_return_statement(stmt: ast::ReturnStmt, fresher: &mut Fresher) -> Result<ReturnStmt> {
+    Ok(ReturnStmt {
+        loc: stmt.base.location,
+        argument: analyze_expression(stmt.argument, fresher)?,
+    })
+}
+
+fn analyze_variable_assignment(
+    stmt: ast::VariableAssgn,
+    fresher: &mut Fresher,
+) -> Result<VariableAssgn> {
+    Ok(VariableAssgn {
+        loc: stmt.base.location,
+        id: analyze_identifier(stmt.id, fresher)?,
+        init: analyze_expression(stmt.init, fresher)?,
+    })
+}
+
+fn analyze_member_assignment(stmt: ast::MemberAssgn, fresher: &mut Fresher) -> Result<MemberAssgn> {
+    Ok(MemberAssgn {
+        loc: stmt.base.location,
+        member: analyze_member_expression(stmt.member, fresher)?,
+        init: analyze_expression(stmt.init, fresher)?,
+    })
+}
+
+fn analyze_expression(expr: ast::Expression, fresher: &mut Fresher) -> Result<Expression> {
+    match expr {
+        ast::Expression::Function(expr) => Ok(Expression::Function(Box::new(analyze_function_expression(*expr, fresher)?))),
+        ast::Expression::Call(expr) => Ok(Expression::Call(Box::new(analyze_call_expression(*expr, fresher)?))),
+        ast::Expression::Member(expr) => Ok(Expression::Member(Box::new(analyze_member_expression(*expr, fresher)?))),
+        ast::Expression::Index(expr) => Ok(Expression::Index(Box::new(analyze_index_expression(*expr, fresher)?))),
+        ast::Expression::PipeExpr(expr) => Ok(Expression::Call(Box::new(analyze_pipe_expression(*expr, fresher)?))),
+        ast::Expression::Binary(expr) => Ok(Expression::Binary(Box::new(analyze_binary_expression(*expr, fresher)?))),
+        ast::Expression::Unary(expr) => Ok(Expression::Unary(Box::new(analyze_unary_expression(*expr, fresher)?))),
+        ast::Expression::Logical(expr) => Ok(Expression::Logical(Box::new(analyze_logical_expression(*expr, fresher)?))),
+        ast::Expression::Conditional(expr) => Ok(Expression::Conditional(Box::new(analyze_conditional_expression(*expr, fresher)?))),
+        ast::Expression::Object(expr) => Ok(Expression::Object(Box::new(analyze_object_expression(*expr, fresher)?))),
+        ast::Expression::Array(expr) => Ok(Expression::Array(Box::new(analyze_array_expression(*expr, fresher)?))),
+        ast::Expression::Identifier(expr) => Ok(Expression::Identifier(analyze_identifier_expression(expr, fresher)?)),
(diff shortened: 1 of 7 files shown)


libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum Statement<'a> {
+    Expr(ExpressionStatement<'a>),
+    Var(VariableAssignment<'a>),
+    Opt(OptionStatement<'a>),
+    Ret(ReturnStatement<'a>),
+    Test(TestStatement<'a>),
+    Built(BuiltinStatement<'a>),
+}
```
jlapacik: These variant names will probably have to be updated to reflect the AST.

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+use chrono::FixedOffset;
+use std::vec::Vec;
+
+pub type Location<'a> = &'a ast::SourceLocation;
+
+// TypeInformation represents the type information for a semantic node.
+#[derive(Debug, PartialEq, Clone)]
+pub struct TypeInformation {
+    pub mono_type: types::MonoType,
+}
```
jlapacik: Is there a reason for wrapping the monotype vs attaching the monotype to the node directly?

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+    pub loc: Location<'a>,
+
+    pub member: MemberExpression<'a>,
+    pub init: Expression<'a>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct StringExpression<'a> {
+    pub loc: Location<'a>,
+    pub type_info: TypeInformation,
```
jlapacik: Could this field just be of type `MonoType`?

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+    pub right: Expression<'a>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CallExpression<'a> {
+    pub loc: Location<'a>,
+    pub type_info: TypeInformation,
+
+    pub callee: Expression<'a>,
+    pub arguments: ObjectExpression<'a>,
```
jlapacik: I don't think this should be an object expression. This says that every function implicitly takes a single object as a parameter. But this isn't consistent with the ability of a function to specify default or pipe arguments. Nor is it consistent with extensible records which is what our objects are now.

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+    pub operator: ast::OperatorKind,
+    pub argument: Expression<'a>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Property<'a> {
+    pub loc: Location<'a>,
+
+    pub key: PropertyKey<'a>,
+    pub value: Option<Expression<'a>>,
```
jlapacik: Why is this optional?

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2688 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::semantic::types::{Fresher, Tvar};
+    use pretty_assertions::assert_eq;
+
+    // TestFresher mocks the fresher so that we don't have to test Tvar ids.
+    // That is not straightforward because the IncrementingFresher is used at different depths in
+    // the semantic graph.
+    struct TestFresher {}
```
jlapacik: I believe the derivative crate will allow you to do something like this:
```Rust
#[derive(Derivative)]
#[derivative(PartialEq)]
pub struct Expression {
    ....
    #[derivative(PartialEq="ignore")]
    pub mono_type: MonoType,
}
```
so that you can ignore the monotype. Do we want to do that? We could ignore the source location as well.

On an "unrelated" note, I now see why you made `Fresher` a trait.

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2688 @@
...
+    let imports = file
+        .imports
+        .iter()
+        .map(|i| analyze_import_declaration(i, fresher))
+        .collect::<Result<Vec<ImportDeclaration>>>()?;
+    let body = file
+        .body
+        .iter()
+        .map(|s| analyze_statement(s, fresher))
+        .collect::<Result<Vec<Statement>>>()?;
```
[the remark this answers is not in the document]
affo: 🙏

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2688 @@
+use crate::ast;
+use crate::semantic::nodes::*;
+use crate::semantic::types::{Fresher, IncrementingFresher, MonoType};
+use std::result;
+
+type SemanticError = String;
+type Result<T> = result::Result<T, SemanticError>;
```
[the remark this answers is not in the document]
affo: 🙏

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2688 @@
...
+        ast::Expression::Bool(lit) => Ok(Expression::Bool(analyze_boolean_literal(&lit, fresher)?)),
+        ast::Expression::Flt(lit) => Ok(Expression::Flt(analyze_float_literal(&lit, fresher)?)),
+        ast::Expression::Int(lit) => Ok(Expression::Int(analyze_integer_literal(&lit, fresher)?)),
+        ast::Expression::Uint(lit) => Ok(Expression::Uint(analyze_unsigned_integer_literal(&lit, fresher)?)),
+        ast::Expression::Regexp(lit) => Ok(Expression::Regexp(analyze_regexp_literal(&lit, fresher)?)),
+        ast::Expression::Dur(lit) => Ok(Expression::Dur(analyze_duration_literal(&lit, fresher)?)),
+        ast::Expression::Time(lit) => Ok(Expression::Time(analyze_time_literal(&lit, fresher)?)),
+        ast::Expression::PipeLit(_) => Err("a pipe literal may only be used as a default value for an argument in a function definition".to_string()),
+        // This should not happen, but it could for BadExpression.
+        _ => Err("unsupported expression in semantic analysis".to_string())
```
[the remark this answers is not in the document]
affo: @jlapacik this shouldn't.

The ast shouldn't contain any error node at this point.

The check should be run before passing the ast to `analyze`.

libflux/src/semantic/analyze.rs
affo: Mmmmm, do we really want to make equality ignore the source location?

Two  identical identifiers, for instance, in two different positions are not the same identifier.

Even though I like this Rust directive, I wouldn't like to change the nodes' behavior for tests.

I am waiting for your feedback on this.
(I would really like to have some like Go's comparison options)

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+extern crate chrono;
+use crate::ast;
+use crate::semantic::types;
+
+use chrono::prelude::DateTime;
+use chrono::Duration;
+use chrono::FixedOffset;
+use std::vec::Vec;
+
+pub type Location<'a> = &'a ast::SourceLocation;
```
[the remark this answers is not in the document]
affo: Yep, the idea is to point to that location and avoid cloning it. It should be useless.

libflux/src/semantic/nodes.rs
affo: It was mostly to implement `update` once and forall.

Do you preder to remove the method and attache only the monotype?

libflux/src/semantic/types.rs
```diff
@@ -119,14 +119,38 @@ impl fmt::Display for MonoType {
...
 impl fmt::Display for Tvar {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "t{}", self.0)
     }
 }
 
+// Fresher is used to get fresh type variables.
+pub trait Fresher {
+    fn get(&mut self) -> Tvar;
+}
```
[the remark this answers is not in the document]
affo: I maintained the trait implementation for mocking the fresher.

libflux/src/semantic/nodes.rs
affo: There can be cases in which the objects do not specify values. For example:

```
{a, b}
```

Is equivalent to

```
{a: a, b: b}
```

Now I see your point. I guess I could remove the  optional expression and assign an IdentifierExpr in that case.

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+    pub operator: ast::Operator,
+    pub argument: Expression<'a>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Property<'a> {
+    pub loc: Location<'a>,
+
+    pub key: PropertyKey<'a>,
+    pub value: Expression<'a>,
```
affo: I have updated this not to be optional (I realized that that was useless as you noted).

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,567 @@
...
+    pub right: Expression<'a>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CallExpr<'a> {
+    pub loc: Location<'a>,
+    pub type_info: TypeInformation,
+
+    pub callee: Expression<'a>,
+    pub arguments: Vec<Property<'a>>,
```
affo: Now Arguments are a list of properties

affo:
@jlapacik I passed to a more OO model in order to simplify method signatures and use a fresher in the `Analyzer`.

That  way it can be mocked easily. I fixed the `ObjectExpr` thing in `CallExpr` and fixed the naming coherently with the changes in `ast`.

libflux/src/semantic/nodes.rs
affo: I have done that in the last commit.

libflux/src/semantic/nodes.rs
affo: Mmmm, I guess the ability of a function to specify default or pipe arguments should be in `FunctionExpr`, and that's where it is.

`CallExpr`, instead, is only to specify how a function is called.

> Nor is it consistent with extensible records which is what our objects are now

For this, how should I change the representation then? In my last commit I opted for `Vec<Property>`.

libflux/src/semantic/nodes.rs
affo: see https://github.com/influxdata/flux/pull/2027#discussion_r336613955

libflux/src/semantic/analyze.rs
affo: Thinking more about it, the type can be excluded from the comparison, but I'd avoid to exclude the location for the reason above.

We could say that, in general, no two nodes in the same place should have a different type. So, we can safely exclude the type.

libflux/src/semantic/analyze.rs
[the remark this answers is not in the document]
jlapacik: I agree the AST shouldn't contain any error nodes. So if we get an AST with a BadExpression node, we should error. The reason I say this should explicitly match on BadExpression, is that if a new node is added to the AST, then this won't compile.

libflux/src/semantic/types.rs
```diff
@@ -119,14 +119,28 @@ impl fmt::Display for MonoType {
...
 
+// Fresher returns a fresh type variable with incrementing id.
+pub struct Fresher(i64);
+
+impl Fresher {
+    pub fn new() -> Fresher {
+        Fresher(0)
+    }
+
+    pub fn get(&mut self) -> Tvar {
```
jlapacik: Could you name this `fresh` instead of `get`? It's just a little more explicit about what is actually happening. I usually don't expect a `get` method to mutate the object.

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2604 @@
...
+            ast::Statement::Return(s) => Ok(Statement::Return(self.analyze_return_statement(&s)?)),
+            // TODO(affo): we should fix this to include MemberAssignement.
+            //  The error lies in AST: the Statement enum does not include that.
+            //  This is not a problem when parsing, because we parse it only in the option assignment case,
+            //  and we return an OptionStmt, which is a Statement.
+            ast::Statement::Variable(s) => {
+                Ok(Statement::Variable(self.analyze_variable_assignment(&s)?))
+            }
+            // This should not happen, but it could for BadStatement.
+            _ => Err("unsupported statement in semantic analysis".to_string()),
```
jlapacik: Similar to `analyze_expression` but the only option here is that we get a `BadStatement` node right? If that's the case, can we explicitly match on `BadStatement`?

To be clear, I think we should still error, I just prefer the stronger compile time guarantees that matching on all cases gives us.

affo: Yeah, wise 👍

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2604 @@
...
+            _ => Err("missing return statement in block".to_string()),
+        }
+    }
+
+    fn analyze_call_expression<'a>(&mut self, expr: &'a ast::CallExpr) -> Result<CallExpr<'a>> {
+        let callee = self.analyze_expression(&expr.callee)?;
+        // TODO(affo): I'd prefer these checks to be in ast.Check().
+        if expr.arguments.len() > 1 {
+            return Err("arguments are more than one object expression".to_string());
+        }
```
jlapacik: This can be removed now.

jlapacik: Actually never mind. I misread this. Although I would definitely prefer that this be changed in the AST. But not necessary for this PR.

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2580 @@
+use crate::ast;
+use crate::semantic::nodes::*;
+use crate::semantic::types::{Fresher, MonoType};
+use std::result;
+
+type SemanticError = String;
+type Result<T> = result::Result<T, SemanticError>;
+
+// analyze analyzes an AST package node and returns its semantic analysis.
+// The function requires a Fresher to instantiate fresh type veraibles for Expressions.
```
jlapacik: typo: variables

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2580 @@
...
+// The function explicitely moves the ast::Package because it adds information to it.
```
jlapacik: typo: explicitly

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2580 @@
...
+// We follow here the principle that every compilation step should be isolated and should add meaning
+// to the previous one. In other terms, once one analyzes an AST he should not use it anymore.
+// If one wants to do so, he should explicitely pkg.clone() and incur consciously in the memory
```
jlapacik: typo: explicitly

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2580 @@
...
+// overhead involved.
+pub fn analyze(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
+    analyze_package(pkg, fresher)
+    // TODO(affo): run checks on the semantic graph.
+}
+
+fn type_info(fresher: &mut Fresher) -> MonoType {
+    MonoType::Var(fresher.fresh())
+}
```
jlapacik: I think we can replace calls to `type_info` now with `MonoType::Var(fresher.fresh())` to be more explicit.

libflux/src/semantic/nodes.rs
```diff
@@ -0,0 +1,576 @@
...
+
+    pub value: String,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct InterpolatedPart {
+    pub loc: ast::SourceLocation,
+
+    pub expression: Expression,
+}
```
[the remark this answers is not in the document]
jlapacik: TextPart and InterpolatedPart aren't considered expressions on their own and so they aren't given types. However InterpolatedPart contains an expression. This expression will be constrained to be of type `string` initially. Eventually though it will just be constrained to be of kind `StringFormattable`(insert better name here).

libflux/src/semantic/analyze.rs
```diff
@@ -0,0 +1,2604 @@
+use crate::ast;
+use crate::semantic::nodes::*;
+use crate::semantic::types::{Fresher, MonoType};
+use std::result;
+
+type SemanticError = String;
+type Result<T> = result::Result<T, SemanticError>;
+
+pub struct Analyzer {
```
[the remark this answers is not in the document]
affo: This is no more there, sorry 😅

libflux/src/semantic/types.rs
```diff
@@ -119,14 +119,28 @@ impl fmt::Display for MonoType {
...
+pub struct Tvar(pub i64);
 
 impl fmt::Display for Tvar {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "t{}", self.0)
     }
 }
 
+// Fresher returns a fresh type variable with incrementing id.
+pub struct Fresher(i64);
```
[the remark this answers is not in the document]
affo: yep, it needs to because it is used in `analyze`.

I could move it to analyze and make it private because I guess it wouldn't be used anywhere else.

affo:
@jlapacik I fixed typos and addressed the `typ` minor change.

@nathanielc 
In the last commit I made the `Fresher` private by moving it to `analyze`. This implies that `fn analyze` cannot take a `Fresher` in. Could this be a problem @jlapacik ? 🤔 

I can revert that last commit if there is an explicit need to take a fresher in.

libflux/src/semantic/nodes.rs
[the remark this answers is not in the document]
affo: Yeah, my assumption was that the expression in the interpolated part will be constrained to work, but in the end it would be turned into a string.

libflux/src/semantic/types.rs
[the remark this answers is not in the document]
jlapacik: Sorry I missed this. `analyze` will need to take a `Fresher` as input and so it should be public so that the same `Fresher` can be used when analyzing multiple packages.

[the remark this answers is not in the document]
affo: Yeah. I guessed that was the reason. I can revert the last commit then.

affo:
@jlapacik @nathanielc rebased to fix conflicts and reverted the last commit. Is this good to go?

PR #2029, opened 2019-10-18
feat(ast): create a flatbuffers schema for AST

wolffcm:
This PR adds a FlatBuffers schema for Flux AST and commits generated code for both Rust and Go.

Generating the FlatBuffers is part of the `generate` target of make, and I made sure that `checkgenerate` works right too.

I put the schema in `flux/ast/ast.fbs` and the generated code lives here:
- `flux/ast/fbast/*.go` for the generated Go code
-  `flux/libflux/src/ast/fbast.rs` for the generated Rust code


ast/gen.go
```diff
@@ -0,0 +1,4 @@
+package ast
+
+//go:generate flatc --go ./ast.fbs
```
jsternberg: Is there any way you could generate these files to an internal package? So that the files go to the package: `ast/internal/fbast`.

libflux/Makefile
```diff
@@ -6,9 +6,12 @@ build:
 	wasm-pack build --dev --scope influxdata
 
 generate:
+	flatc --rust -o src/ast/ ../ast/ast.fbs
```
jsternberg: Should we integrate this into `build.rs`? I think the makefile is probably fine since I don't think it is necessary to require `flatc` be present to build the Rust module though.

Can we get this with proper Makefile targets and sources though? I would also prefer if we kept `ast_generated.rs` and did not publicly expose the module. This would simplify the makefile and I don't think the specific name matters since the module wouldn't be publicly exposed anyway.

ast/Makefile
```diff
@@ -0,0 +1,26 @@
+# List any generated files here
+TARGETS = fbast/*.go
```
nathanielc: @jsternberg Just did some work to make the Makefile nonrecusive, I am not sure if that has been merge yet but a heads up that this will need updating at some point soon.

ast/ast.fbs
```diff
@@ -0,0 +1,359 @@
...
+
+// Expressions
+
+// Unions cannot appear in vectors or other unions,
+// so provide a wrapped version where needed.
+table WrappedExpression {
+  expr:Expression;
+}
+
+enum OperatorKind : byte {
```
nathanielc: In the Rust code I removed the `Kind` suffix from all of these types since they were not needed. Are they needed here or can we shorten this to `Operator`?

libflux/Cargo.toml
```diff
@@ -16,6 +16,7 @@ wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
 chrono = { version = "0.4", features = ["serde"] }
 regex = "1"
 maplit = "1.0.2"
+flatbuffers = "0.6.0"
```
nathanielc: Does this mean we depend on some Rust runtime code for flatbuffers?

wolffcm: Yes, there is some runtime code that flatbuffers uses for both Rust and Go.

ast/Makefile
jsternberg: It should be decently easy to integrate this with the nonrecursive Makefile. If the other PR gets merged first, I'll help Chris to update this. If this gets merged first, I'll rebase my PR.

ast/ast.fbs
```diff
@@ -0,0 +1,359 @@
+namespace fbast;
```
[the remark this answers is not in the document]
wolffcm: Unions are kind of like enums, but the possible values must be table types.  (The flatbuffers table type is kind of like a struct, but the fields are all optional in the serialization)

When a union is nested in a larger data structure like a table, it's really two fields; the value itself, and a `_type` field that lets the client know what the value is, similar to a tagged struct in C.  This makes it a little awkward to work with the API presented by the generated code, eg:
```rust
    let add = BinaryExpression::create(
        &mut b,
        &BinaryExpressionArgs {
            operator: OperatorKind::AdditionOperator,
            left_type: Expression::IntegerLiteral,
            left: Some(lit1.as_union_value()),
            right_type: Expression::IntegerLiteral,
            right: Some(lit2.as_union_value()),
            ..BinaryExpressionArgs::default()
        },
    );
```
However, we can wrap this in a more easy to work with API.

ast/ast.fbs
wolffcm: Yes, it can be shortened.

libflux/Makefile
wolffcm: Hmm, I've some questions about this.  I didn't know about `build.rs`.

So the lexer is generated as part of `cargo build`, then?  Does that mean if `scanner.rl` is out of sync with its generated code, that `make checkgenerate` will not catch it?

(I was also a little surprised to see that there are two identical copies of `scanner.rl` in our repo)

I'm fine with generating the code from the Makefile or from `build.rs`, but it seems like the mechanism we use to do it should be the same as for the lexer.

jsternberg: For the Rust code, the scanner is not checked in and is constructed with `build.rs`. I am wondering if this is the correct way to do it though. It assumes that everyone who uses the code and performs the build will also need these same tools. So if you want to build the Rust parser for Flux inside of InfluxDB, then the InfluxDB build environment will need these tools present to generate the code.

I do agree that we should do the same thing for both of these. Maybe we should convert the scanner to generate the code as part of the makefile and commit the generated code just like we do with Go.

wolffcm: > I would also prefer if we kept ast_generated.rs and did not publicly expose the module.

The generated code is not publicly exposed now, as it's nested within the `ast` module and doesn't have the keyword `pub`.  

My understanding of Rust modules is that if I left the file with its original name of `ast_generated.rs`, then the module would have to be named `ast_generated` which doesn't convey very well what it is.  I think maybe I'm misunderstanding your question though, we should talk tomorrow.

jsternberg: Ok. I think this is fine.

jsternberg: One caveat. Can you move the `mv` to the same line and do `&& mv ...`?

Issue #2006, opened 2019-10-16
Ensure that all sql drivers are tested in integration tests


jsternberg:
All of the sql drivers need integration tests that can be used against the actual database implementations. We are going to need this kind of framework in the future for other implementations like the promql transpiler, influxql, and whatever other sources we interact with.

This will likely involve creating docker sidecars as part of the integration test.

This issue is specifically about testing mysql and postgres.

This will need to be separated from the other unit tests in some way just so we don't need to run them on our local machines. If the tests don't take too long, it is ok if they are run by default in CI. If they do take more time to run (such as maybe >1 minute), then it is ok if they only run when a commit lands on master.

The only requirement is that they _must_ be run before we perform any kind of release.

onelson:
The initial lift towards this was done in https://github.com/influxdata/flux/pull/4317 and there were a handful of follow-up PRs to add additional backend support for the problem cases encountered along the way.

The status quo gives us the ability to run local integration tests against a non-exhaustive list of supported backends. We have internal docs on how to manually test some cloud-based backends (such as Google BigQuery, Snowflake).

The docker setup for certain backends required modification of the container fs in clumsy ways to configure them. The workflow also relies on `docker exec` to seed initial test data in already running instances. These aspects are barriers to getting this going in CI since approaches to smooth this over involve a combination of publishing custom docker images per backend (to help automate the configuration steps) or perhaps a CI environment where we can do docker-in-docker (DinD) to more closely resemble the local workflow.

Futhermore, some backends (notably Vertica and HANA) take minutes (3-5) to start up and become ready for connections, adding to the pain should we want to make this a barrier in CI.

PR #2030, opened 2019-10-21
feat(stdlib/sql): add sqlite3 support

### LINKED ISSUE #1933: sqlite connector

### LINKED ISSUE #1938: Sql errors not being returned to user

nicgrobler:
Adds support for SQLite3 to the stdlib/sql package. SQLite is a fully SQL compliant relational DB, in a file (or in memory, if preferred). Adding this to flux would allow it to interact with this DB, and present a lot of opportunity to extract information from many 3rd party "infrastructure tools", and combine this with Influxdb, or other data sources.

Closes #1933
Fixes #1938
#2006

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/sql/source_validator.go
+++ b/stdlib/sql/source_validator.go
@@ -12,6 +12,15 @@ import (
 
 // helper function to validate the data source url (postgres, sqlmock) / dsn (mysql) using the URLValidator.
 func validateDataSource(validator url.Validator, driverName string, dataSourceName string) error {
+
+	/*
+		NOTE: some parsers don't return an error for an "empty path" (a path consisting of nothing at all, or only whitespace) - not an error as such, but here we rely on the driver implementation "doing the right thing"
+		better not to, and flag this as an error because calling any SQL DB with an empty DSN is likely wrong.
+	*/
+	if strings.TrimSpace(dataSourceName) == "" {
+		return errors.Newf(codes.Invalid, "invalid data source url: %v", "empty path supplied")
+	}
+
 	var u *neturl.URL
 	var err error
 
@@ -50,11 +59,6 @@ func validateDataSource(validator url.Validator, driverName string, dataSourceNa
 		u, err = neturl.Parse(dataSourceName)
 		if err != nil {
 			return errors.Newf(codes.Invalid, "invalid data source url: %v", err)
-		}
-		// NOTE: parser doesn't return an error for an empty path, or a path consisting of only whitespace - not an error as such, but here we rely on the driver implementation "doing the right thing"
-		// better not to, and flag this as an error
-		if strings.TrimSpace(dataSourceName) == "" {
-			return errors.Newf(codes.Invalid, "invalid data source url: %v", "empty path supplied")
 		}
 	default:
 		return errors.Newf(codes.Invalid, "sql driver %s not supported", driverName)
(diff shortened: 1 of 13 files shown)


nicgrobler:
this is the result of all comments and requested changes in closed PR #1987
which was closed as fork ended up with commit history issues.

stdlib/sql/to.go
```diff
@@ -216,38 +230,109 @@ func (t *ToSQLTransformation) Finish(id execute.DatasetID, err error) {
...
+		driverName := t.spec.Spec.DriverName
+		// the following allows driver-specific type errors (of which there can be MANY) to be returned, rather than the default of invalid type
+		translateColumn, err := getTranslationFunc(driverName)
+		if err != nil {
+			return nil, nil, nil, err
+		}
 		switch col.Type {
 		case flux.TFloat:
-			newSQLTableCols = append(newSQLTableCols, fmt.Sprintf("%s FLOAT", col.Label))
+			v, err := translateColumn()(col.Type, col.Label)
```
yzhang1991: Maybe we can use `fallthrough` here since the `translateColumn()` code for all the types is the same?

stdlib/sql/to_test.go
```diff
@@ -3,9 +3,12 @@ package sql_test
 import (
 	"testing"
 
+	_ "github.com/mattn/go-sqlite3"
+
```
yzhang1991: Generally, we do not use newlines between project imports, we only separate them from the stdlib imports

stdlib/sql/source_validator.go
```diff
@@ -33,13 +34,35 @@ func validateDataSource(validator url.Validator, driverName string, dataSourceNa
...
+			return nil
+		}
+		// we have a dsn that MIGHT be valid, so need to parse it - if it fails here, it is likely to be invalid
+		u, err = neturl.Parse(dataSourceName)
+		if err != nil {
+			return errors.Newf(codes.Invalid, "invalid data source url: %v", err)
+		}
+		// NOTE: parser doesn't return an error for an empty path, or a path consisting of only whitespace - not an error as such, but here we rely on the driver implementation "doing the right thing"
+		// better not to, and flag this as an error
+		if strings.TrimSpace(dataSourceName) == "" {
```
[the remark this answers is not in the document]
nicgrobler: although the MySQL parser will return an error (due to lack of '/'), it doesn't actually check for empty string - I definitely think this most basic check should be applied to all SQL driver types until we come across some type that allows empty dsn where it is not an error (sqlite will create an un-named file - but this is clearly a bad idea) so have moved this logic to the top.

stdlib/sql/to_test.go
nicgrobler: yep, changed ;-)

stdlib/sql/to.go
nicgrobler: done - no logic changes at all, just collapsed switch into shorter version.

yzhang1991:
Hi @nicgrobler ,

Thanks so much for this work! I have checked again the PR and it is now merged.

PR #2035, opened 2019-10-21
Feat/rust ast check

yzhang1991:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/ast/check/mod.rs
+++ b/libflux/src/ast/check/mod.rs
@@ -1,9 +1,7 @@
-use crate::ast::{PropertyKey, SourceLocation, walk};
+use crate::ast::{walk, PropertyKey, SourceLocation};
 use std::fmt;
 
-// check() will inspect a single node and annotate it with any AST errors.
-// Unlike the ast.Check() in our Go parser, the check function in Rust
-// returns a list of errors generated during the check process PLUS
+// check() inspects an AST node and returns a list of found AST errors plus
 // any errors existed before ast.check() is performed.
 pub fn check(node: walk::Node) -> Vec<Error> {
     let mut errors = vec![];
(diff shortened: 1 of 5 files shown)


libflux/src/ast/check/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+use crate::ast::SourceLocation;
+use std::fmt;
+
+#[cfg(test)]
+mod tests;
+
+// check will inspect a single node and annotate it with any AST errors.
+// Unlike the ast.Check() in our Go parser, the check function in Rust
+// returns a list of errors generated during the check process PLUS
+// any errors existed before ast.check() is performed.
```
nathanielc: nit, Can you update this comment to describe what it does without comparing it to the Go check function? We want this comment to serve as the documentation for the function and its likely that who ever is reading it isn't familiar with the Go behavior.

libflux/src/ast/check/tests.rs
```diff
@@ -0,0 +1,114 @@
...
+            },
+            end: Position {
+                line: 2,
+                column: 14
+            },
+            source: Some(String::from("{c: 2, a}"))
+        },
+        message: String::from("cannot mix implicit and explicit properties")
+    }];
+    assert!(want.iter().zip(got).all(|(want, got)| want == &got));
```
nathanielc: if you change this to `assert_eq!(want, got)` does it work? If so I would prefer that method as Rust assert_eq method has good diffing logic for what is different between the values.

libflux/src/ast/check/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+                    location: n.base().location.clone(),
+                    message: err.clone()
+                });
+            }
+            match n {
+                walk::Node::BadStmt(n) => {
+                    errors.push(Error {
+                        location: n.base.location.clone(),
+                        message: format!("invalid statement {}: {}",
+                                         n.base.location, n.text)
```
nathanielc: Since we now store the location on the error type itself I don't think we want it in the error message. Thoughts?

yzhang1991: Makes sense.

libflux/src/ast/check/mod.rs
yzhang1991: Done

libflux/src/ast/check/tests.rs
yzhang1991: Done

PR #2041, opened 2019-10-22
ci: integrate libflux into circleci

### LINKED ISSUE #2022: Integrate libflux build into continuous integration
The continuous integration system needs to have libflux integrated into it so `libflux` is built and used as part of the tests.

There is a small complication with this. Because the Makefiles are split up, it is a bit difficult at the moment to ensure that all of the steps run in the proper order. For details about why this happens, you can reference the paper [Recursive Makefiles considered harmful](http://aegis.sourceforge.net/auug97.pdf).

While recursive make is fine for simple workflows, it becomes impractical once a project starts needing to rely on dependencies that are produced by another Makefile. For integrating `libflux`, we are encountering that exact problem.

This issue is done when we are building `libflux` and running tests with `-tags libflux` inside of CircleCI.

jsternberg:
This updates the Makefiles so there is a top level Makefile that drives
everything and all of the logic is contained within there. The project
isn't complicated enough that a more advanced tool is needed for
managing the build.

A single Makefile is used in order to avoid the problem of recursive
makefiles where the dependency graph is split and doesn't necessarily
build properly. Since the top level build needs to know whether or not
`libflux` is built, that logic has to be in the top level makefile
otherwise we incur a call to `make` and the dependency graph isn't
capable of determining whether the result was even changed.

This enables `-tags libflux` to be used.

Fixes #2022.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


fchikwekwe:
Does this make any changes to the way that we build flux? Besides the addition of the `-tags libflux` flag?

jsternberg:
It shouldn't make any changes to building flux yet. Mostly because the tag is optional so you can go about your normal workflow and then use the makefile and `-tags libflux` when running go test.

PR #2043, opened 2019-10-22
fix(memory): reorder variables in the allocator for atomic operations

jsternberg:
The variables used by atomic operations have to be byte aligned for
these operations to succeed. The easiest way to ensure this happens is
to put these variables at the top of the struct.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/memory/allocator.go
+++ b/memory/allocator.go
@@ -19,6 +19,15 @@ var _ memory.Allocator = (*Allocator)(nil)
 
 // Allocator tracks the amount of memory being consumed by a query.
 type Allocator struct {
+	// Variables accessed with atomic operations should be at
+	// the beginning of the struct to ensure byte alignment is correct.
+	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
+	allocationLimit int64
+	bytesAllocated  int64
+	maxAllocated    int64
+	totalAllocated  int64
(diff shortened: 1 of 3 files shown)


jsternberg:
Note: I'm going to include an upgrade to arrow when apache/arrow#5714 is merged as part of this PR. But this PR can be reviewed separately from that PR.

This is needed for 32 bit InfluxDB 1.x to work correctly. For anybody looking at this, Flux does not officially support 32 bit outside of InfluxDB 1.x and we may pull 32 bit support for Flux in future releases of InfluxDB 1.x.

memory/allocator.go
```diff
@@ -19,6 +19,14 @@ var _ memory.Allocator = (*Allocator)(nil)
 
 // Allocator tracks the amount of memory being consumed by a query.
 type Allocator struct {
+	// Variables accessed with atomic operations should be at
+	// the beginning of the struct to ensure byte alignment is correct.
```
jlapacik: I think a link to this https://golang.org/pkg/sync/atomic/#pkg-note-BUG in the comment would be helpful.

jsternberg: Good idea. I didn't even know this note existed but it's perfect for the comment.

PR #2048, opened 2019-10-23
feat(libflux): integrate libflux parser with the go parsing api

### LINKED ISSUE #2009: Integrate conditional compilation using libflux for parsing
Modify the parser package to use `libflux` when the `libflux` tag is present in the build. This will involve creating the initial cgo bootstrap for libflux and then using it. The conditional compilation is so that we can start with testing and using the cgo parser optionally and potentially begin integration of it with InfluxDB.

The cgo wrapper for `libflux` should be in exactly one package and the parser would invoke it from that directory. My thought is that the structure should be:

```
/flux
    /libflux
        /go # implements the cgo interaction
    /parser
        parser.go # pure go implementation
        libflux_parser.go # uses libflux to implement the parser
```

jsternberg:
The libflux parser can be optionally enabled by using `-tags libflux`. If
that tag is used, the `libflux` go wrapper will be used to perform the
parsing which is written in Rust with a C ABI.

This adds a marshal json method to the Rust code as we are still working on
creating a flatbuffers API and using it within both Go and Rust.

This also updates the C API to generate bindings to the C types from the
header file so these types can be formally used instead of types that don't
match. The semantics are still the same. A pointer to a Rust type is cast
in an unsafe way to a mutable raw pointer that gets managed outside of
Rust, but the type names are more consistent which should lend itself to
readability.

Fixes #2009.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/include/influxdata/flux.h
+++ b/libflux/include/influxdata/flux.h
@@ -1,6 +1,8 @@
 #ifndef _INFLUXDATA_FLUX_H
 #define _INFLUXDATA_FLUX_H
 
+#include <stddef.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -8,12 +10,39 @@ extern "C" {
 // flux_ast_t is the AST representation of a flux query.
 struct flux_ast_t;
 
+// flux_buffer_t is a reference to a byte-slice.
+struct flux_buffer_t {
+	// data is a pointer to the data contained within the buffer.
+	void *data;
+
+	// len is the length of the buffer.
+	size_t len;
+};
+
+// flux_error_t represents a flux error.
+struct flux_error_t;
+
 // flux_parse will take in a string and return the AST representation
 // of the query.
 struct flux_ast_t *flux_parse(const char *);
 
-// flux_ast_free will free memory associated with an AST handle.
-void flux_ast_free(struct flux_ast_t *);
+// flux_ast_marshal_json will marshal json and fill in the given buffer
+// with the data. If successful, memory will be allocated for the data
+// within the buffer and it is the caller's responsibility to free this
+// data. If an error happens it will be returned. The error must be freed
+// using flux_free if it is non-null.
+struct flux_error_t *flux_ast_marshal_json(struct flux_ast_t *, struct flux_buffer_t *);
+
+// flux_buffer_free will free the memory that was allocated for a buffer.
+// This should only be called if the caller is the one who owns the data.
+void flux_buffer_free(struct flux_buffer_t *);
+
+// flux_error_str will return a string representation of the error.
+// This will allocate memory for the returned string.
+const char *flux_error_str(struct flux_error_t *);
+
+// flux_free will free a resource.
+void flux_free(void *);
 
 #ifdef __cplusplus
 }
--- a/libflux/src/lib.rs
+++ b/libflux/src/lib.rs
@@ -9,24 +9,58 @@ pub mod scanner;
 pub mod semantic;
 
 use std::ffi::*;
-use std::os::raw::c_char;
+use std::error::Error;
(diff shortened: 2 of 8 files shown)


libflux/src/lib.rs
```diff
@@ -9,24 +9,58 @@ pub mod scanner;
...
+    buffer.len = data.len();
+    buffer.data = Box::into_raw(data.into_boxed_slice()) as *mut c_void;
+    return std::ptr::null_mut();
+}
+
+#[no_mangle]
+pub extern "C" fn flux_error_str(err: *mut flux_error_t) -> *mut c_char {
+    let e = unsafe { &*(err as *mut ErrorHandle) };
+    let s = CString::new(e.err.description()).unwrap();
+    return s.into_raw();
```
fchikwekwe: Why is the unsafe pointer required/preferred and how is it handled? I'd just like to understand more about your thought process here. Still learning how pointers work in Rust.

fchikwekwe: I did see that you mentioned that this leads to better, more-consistent type names. Just wondering if there were additional thoughts that led to this approach.

jsternberg: The pointer type doesn't really matter since those types don't really exist anyway. In C, you are allowed to forward declare a type and use it as a pointer without declaring what's inside of it since what's inside of it ends up becoming part of the ABI (application binary interface). That ends up meaning that an exposed struct requires updating the library major version whenever you modify one that is publicly exposed so you should avoid publicly exposing structs that the caller doesn't need. It also prevents anybody from messing with the internals of your struct if you want to handle it in a certain way.

The reason why an unsafe pointer is required is because of Rust's memory management. If we don't call `into_raw()` then Rust will see that the lifetime of the box is finished and it will free the memory associated with it. When we turn it into a raw pointer, we are telling Rust that we are taking ownership of that memory and that we will handle the lifetime. This is needed because we are returning memory from Rust to Go and then using it with an indeterminate lifetime. It's unsafe because the memory guarantees that Rust gives are gone, but we need to do it because we need to control the lifetime of this pointer manually rather than let the Rust runtime take care of it.

The type names are mostly just to make it easier to follow the code and see that it matches the C code. You can have all of these return `void *` and they would work fine since the ABI's would be compatible. Once a box is turned into a raw pointer, it really doesn't matter what's inside of it. That's also the unsafe part because there's no guarantee that `*mut flux_error_t` is actually an error type other than us giving a pinky-promise that we won't do anything bad.

libflux/include/influxdata/flux.h
```diff
@@ -1,19 +1,48 @@
...
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 // flux_ast_t is the AST representation of a flux query.
 struct flux_ast_t;
 
+// flux_buffer_t is a reference to a byte-slice.
+struct flux_buffer_t {
```
wolffcm: As an FYI, it will be just a bit more complicated to make a flatbuffer available to Go.  Since the flatbuffer builder builds from the end of the buffer backwards, there will be some unused space at the beginning of the buffer, but Go will still need a pointer to the first byte to pass to `Free`.  So the analogous data structure for flatbuffers will need an `offset` field.

jsternberg: That's good to know. We'll figure out if we need that. The pointer is not always supposed to be freed as sometimes it's a reference and there isn't really a way to deal with that. We can also always return the offset as part of a reference variable that gets set by the C function equivalent or we'll find some way to return the offset.

libflux/src/lib.rs
fchikwekwe: Thanks for the detailed explanation. Helped a lot

PR #2052, opened 2019-10-23
fix(execute): add annotated errors to the execute package where it effects normal usage

jsternberg:
The aggregate, selector, and row functions get used within the universe
package. While we annotated the errors in universe, this didn't effect
transformations that had the bulk of their implementation within the
execute package.

This specifically targets very common errors that could occur so that
they don't show up as internal errors in InfluxDB.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -1,11 +1,10 @@
 package execute
 
 import (
-	"errors"
-	"fmt"
-
 	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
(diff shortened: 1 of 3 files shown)


jsternberg:
@nathanielc I am planning to backport this to `v0.50.1`. I encountered it today when accidentally using an auto-generated query within cloud that tried to sum strings. This is specifically targeted to those errors most likely to be hit by a user so this doesn't cover the entire execute package.

Issue #2051, opened 2019-10-23
Allow option evaluation without full AST evaluation


gavincabbage:
Currently, extracting options requires evaluation of the full AST through a call to [`EvalAST`](https://github.com/influxdata/flux/blob/01f7fa654ff215c6cd35c0d47cdd48c41ff63eb8/compile.go#L50). This issue proposes a new evaluation helper, `EvalOptions`, which will rewrite the AST to only include option statements before evaluation. This will allow options to be extracted without evaluating the full AST.

The signature can be the same as that of `EvalAST`, differing only in behavior, i.e.

```
EvalOptions(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error)
```

gavincabbage:
Satisfied by #2054

PR #2054, opened 2019-10-24
feat(compile): add EvalOptions

gavincabbage:
A naive implementation to satisy #2051, this PR adds a new `EvalOptions` function that calls `EvalAST` with a filtered deep copy of the AST that excludes all statement types except options. This will allow options to be extracted and inspected without evaluating the full AST. A simple test is included to validate the option filtering logic.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -72,6 +72,39 @@ func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]
 	return sideEffects, scope, nil
 }
 
+// EvalOptions is like EvalAST, but only evaluates options.
+func EvalOptions(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+	return EvalAST(ctx, options(astPkg), opts...)
+}
+
+// options returns a shallow copy of the AST, trimmed to include only option statements.
+func options(astPkg *ast.Package) *ast.Package {
+	trimmed := &ast.Package{
+		BaseNode: astPkg.BaseNode,
+		Path:     astPkg.Path,
+		Package:  astPkg.Package,
+	}
+	for _, f := range astPkg.Files {
+		var body []ast.Statement
+		for _, s := range f.Body {
+			if opt, ok := s.(*ast.OptionStatement); ok {
+				body = append(body, opt)
+			}
+		}
+		if len(body) > 0 {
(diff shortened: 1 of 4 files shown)


compile.go
```diff
@@ -72,6 +72,38 @@ func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]
 	return sideEffects, scope, nil
 }
 
+// EvalOptions is like EvalAST, but only evaluates options.
+func EvalOptions(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+	return EvalAST(ctx, options(astPkg), opts...)
+}
+
+func options(astPkg *ast.Package) *ast.Package {
```
nathanielc: Can you add a simple comment here explaining what the function does?

compile.go
```diff
@@ -72,6 +72,38 @@ func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]
...
+	trimmed := &ast.Package{
+		BaseNode: astPkg.BaseNode,
+		Path:     astPkg.Path,
+		Package:  astPkg.Package,
+	}
+	for _, f := range astPkg.Files {
+		var body []ast.Statement
+		for _, s := range f.Body {
+			if s.Type() == "OptionStatement" {
```
nathanielc: Can you use a type switch on the statement instead of calling `Type()`? That is more consistent with how we check the type of the nodes and it makes it easier to refactor later as the comparison is against a real type instead of a string literal.

gavincabbage: I'm glad you called this one out, I was going to ask you about this exact thing! Testing the requested changes now.

compile.go
```diff
@@ -72,6 +72,39 @@ func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]
 	return sideEffects, scope, nil
 }
 
+// EvalOptions is like EvalAST, but only evaluates options.
+func EvalOptions(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+	return EvalAST(ctx, options(astPkg), opts...)
+}
+
+// options returns a copy of the AST, trimmed to include only option statements.
```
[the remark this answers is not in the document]
gavincabbage: Absolutely. Good call.

PR #2059, opened 2019-10-25
fix(libflux): correctly lex µs

yzhang1991:
Using `unsigned char` instead of `char` in Ragel C generated function will fix this issue!!

### DIFF
--- a/libflux/src/scanner/mod.rs
+++ b/libflux/src/scanner/mod.rs
@@ -1,6 +1,6 @@
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
-pub type CChar = i8;
+pub type CChar = u8;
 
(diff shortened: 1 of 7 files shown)


libflux/src/scanner/mod.rs
```diff
@@ -1,6 +1,6 @@
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
-pub type CChar = i8;
+pub type CChar = u8;
```
affo: Wow, that's cool!

affo:
@yzhang1991 this is the test I am mentioning https://github.com/influxdata/flux/blob/c1bac5abee883417532f50c35ca4dd215d3caade/libflux/src/scanner/tests.rs#L850

PR #2060, opened 2019-10-25
feat(libflux): semantic.Walk

affo:
This implements the walk pattern for the semantic graph.

The Visitor directly supports nesting and `walk` invokes `nest` once it encounters a `Block`.

`nest` can return a "new" visitor:
 - if it does, that visitor is used for the new scope;
 - if it doesn't, the same visitor is used for the new scope;

The default implementation for `nest` is to return `None`.

__UPDATE__: We decided not to use `nest` and to make walk use only _one_ visitor. Nesting can be implemented with a stack on the internal state of the visitor, as shown in the test cases.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,997 @@
...
+// When used with the walk function, Visit will be called for every node
+// in depth-first order. After all children for a Node have been visted,
+// Done is called on that Node to signal that we are done with that Node.
+//
+// If Visit returns None, walk will not recurse on the children.
+pub trait Visitor: Sized {
+    // Visit is called for a node.
+    // The returned visitor will be used to walk children of the node.
+    // If visit returns None, walk will not recurse on the children.
+    fn visit(&self, node: &mut Node) -> Option<Self>;
```
jlapacik: Why does `visit` need a reference to `self` if it's returning another visitor? Why can't the signature be:
```Rust
fn visit(self, node: &mut Node) -> Option<Self>;
```

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,997 @@
...
+    // Nest gets called when the walk encounters a Block.
+    // By default it returns None.
+    // If some visitor is returned, that will be used to continue walking
+    // down the current Block.
+    fn nest(&self) -> Option<Self> {
+        None
+    }
```
jlapacik: Nesting I believe is only ever used once in the Go code and it's used as a way to get around the limitations imposed by the default traversal algorithm. I would honestly like to see it removed unless it is deemed absolutely necessary.

In the Go code nesting is used by the type checker during constraint generation. It's purpose it to add a new stack frame to the type environment when visiting a function block. However a new stack frame must be added ***before*** we visit the block. As a result, type inference cannot use the visitor pattern to generate constraints.

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,997 @@
...
+    // Done is called for a node once it has been visited along with all of its children.
+    fn done(&self, _: &mut Node) {} // default is to do nothing
+}
+
+// Walk recursively visits children of a node.
+// Nodes are visited in depth-first order.
+pub fn walk<T>(v: &T, mut node: Node)
```
jlapacik: Why pass the visitor as a trait object?

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,997 @@
...
+
+type FuncVisitor<'a> = Rc<RefCell<&'a mut dyn FnMut(&mut Node)>>;
+
+impl<'a> Visitor for FuncVisitor<'a> {
+    fn visit(&self, node: &mut Node) -> Option<Self> {
+        let mut func: RefMut<_> = self.borrow_mut();
+        (&mut *func)(node);
+        Some(Rc::clone(self))
+    }
+}
```
jlapacik: Holy cow!

libflux/src/semantic/walk.rs
affo: I think it's better to let `visit` only borrows `self`.

This enables stuff like

```
let v = SomeVisitor{};
walk(&v, node);
println!("{}", v.something);
```

Otherwise, `visit` would own `v` and that couldn't be used after `walk`.

libflux/src/semantic/walk.rs
affo: Got it

libflux/src/semantic/walk.rs
affo: I think I don't understand the question 🤔 . Can you articulate more?

affo:
@jlapacik 

> I might have a visitor that needs to modify the default traversal. Is there a way to do that?

Yep, I can think about an alternative design that lets you modify the traversal. Do you think we need "standard traversals" (depth/breadth-first, ...) or something more sophisticated?

> Each call to visit returns a new visitor, but where does this visitor go? It does not get returned from walk. How can I retrieve the information I've gathered from my walk?

I see your point. I am actually writing down a test that uses and external variable as "global" state for multiple visitors and I am trying to see where I can go from there.

@jlapacik In general, it seems that the requirements for `Walk` are a bit different from the ones we had in the Go codebase, or, put it in another way, `Walk` can be designed better now that we are rewriting it. I would like to better understand those requirements and give it a shot. We can start by listing the actual use cases of walking the semantic graph and see what we exactly need to explore the graph and evaluate it.

At the moment, I'll try to complete the test I am writing to prove that there is a way to have a global state across visitors and I'll rework `nest`.

affo:
@jlapacik The last commit adds a test with global error handling across visitor, which is what I think you where referring to above.

Considerations: I think with that those pattern we are somehow forcing people to use `Rc` and `RefCell` for visiting and sharing state, which doesn't seem to be a best practice in Rust, in general.

However, it seems that `Rc<RefCell<T>>` is a common pattern in graph traversal and manipulation: https://users.rust-lang.org/t/is-rc-refcell-a-code-smell/27366. Which makes me think that we are doing it right somehow 😅

libflux/src/semantic/walk.rs
jlapacik: I guess then a better question is why doesn't `walk` return a visitor? It seems a little weird that `visit` returns a visitor, but walk returns nothing?

affo: Mmmm ok, but what would you made of the _last_ visitor? wouldn't you need a "trace" of the visitors? Also, how can you access those visitors between calls?

libflux/src/semantic/walk.rs
jlapacik: Sorry not a good question.

jlapacik:
@affo we discussed in a slack call but just to reiterate, I think the `visit` method should consume(own) it's visitor. In particular I think the Visitor trait should maybe look like this:
```Rust
pub trait Visitor: Sized {
    fn visit(self, node: &mut Node) -> WalkOption<Self>;
}

pub enum WalkOption<T: Visitor> {
    Done(T),
    Continue(T),
}
```

Walk would return a visitor as well:
```Rust
pub fn walk<T>(v: T, node: &mut Node) -> WalkOption<T>
where
    T: Visitor,
{
    match v.visit(node) {
        // Done walking graph with visitor - return immediately.
        WalkOption::Done(w) => WalkOption::Done(w),
        // Continue walking graph with visitor.
        WalkOption::Continue(w) => ...,
    }
}
```
The output of walking a semantic graph then is a visitor in its final state.

Note a visitor can override the default graph traversal algorithm by calling `walk` inside of `visit`. Returning `WalkOption::Done` would ensure the graph is not walked further (and that each node is visited at most once).

Also note that implementing `Visitor` on a closure might look something like this:
```Rust
impl Visitor for Box<dyn FnMut(&mut Node)> {
    fn visit(mut self, node: &mut Node) -> WalkOption<Self> {
        self(node);
        Opt::Continue(self)
    }
}
```

Thoughts? Is this what you were thinking too? I removed the `done` method from the Visitor trait too as it seems like it's no longer necessary if we are always returning an updated visitor. Does this seem correct?

affo:
@jlapacik I really like where this is heading.

`WalkOption` is great. Agreed on walk owning the visitor. I am continuing with the new implementation discussed. Now thinking of providing a more "complete" Visitor, roughly:


```rust
trait Visitor {
fn visit_this(self, n: This) -> Self {
   self
}

fn visit_that(self, n: That) -> Self {
  self
}
...
}
```

Now we can provide default implementations for the various branches.

implementing a visitor that way would be cleaner and save the user from pattern match 🤔 

I think I'll start with implementing what we agreed and wait for your thoughts on this.

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1131 @@
...
+    fn from_assignment(a: &'a mut Assignment) -> Node {
+        match *a {
+            Assignment::Variable(ref mut v) => Node::VariableAssgn(v),
+            Assignment::Member(ref mut m) => Node::MemberAssgn(m),
+        }
+    }
+    // Duplicate returns a new Node that points to the same semantic node of `self`.
+    // It is not named "copy" or "clone" to avoid collision with those traits.
+    // This is required to let `visit` own the node.
+    fn duplicate<'b>(&'b mut self) -> Node<'b> {
```
affo: I haven't found a better way of doing this. I tried to implement `Clone` but I had problems with lifetimes. This is the best thing I could do. This gives the great advantage that `visit` calls can own nodes.

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1131 @@
...
+                walk(v, Node::from_expr(&mut n.init));
+            }
+        };
+    }
+    v.done(node);
+}
+
+impl<F> Visitor for F
+where
+    F: for<'a> FnMut(Node<'a>),
```
affo: @jlapacik you'll love this 😄

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1131 @@
...
+"#,
+            );
+            let mut v = NestingCounter { count: 0 };
+            walk(&mut v, Node::Package(&mut pkg));
+            assert_eq!(v.count, 4);
+        }
+
+        // RepeatedPlusChecker checks if there is more than one addition in the same Block.
+        // If so, it errors.
+        struct RepeatedPlusChecker {
```
affo: @jlapacik @nathanielc Here it is, a scoped visitor

affo:
@jlapacik @nathanielc I am still trying to figure out if we can somehow get rid of that `Node`. In the end, that's only a wrapper for a pointer to abstract away multiple types. Therefore, I wanted `walk` and `visit` to own it.

I implemented `duplicate` for doing that, but I don't know if that's satisfactory.

libflux/src/semantic/walk.rs
affo: @nathanielc this is Higher-Ranked Trait Bounds (woah).

According to the book: 

> This lets the Rust compiler find the minimum lifetime to invoke our closure and satisfy the borrow checker's rules.

The problem here happens when you use that closure.

```
walk<'a>(&mut v: FnMut(Node<'a>), node: Node<???IDK???>) {
  ...
}
```

> When a function has an explicit lifetime parameter, that lifetime must be at least as long as the _entire_ call to that function.

It seems that, due to lifetime elision, the compiler cannot determine the lifetime of `node` compared to the parameter of the visitor and it cannot know that it lives more than the function.

```
walk(&mut v: for <'a> FnMut(Node<'a>), node: Node<???IDK???>) {
  ...
}
```

This is ok, because `<'a>` is not associated to the lifetime of the function call, but only to the node itself ---> the compiler can find a good lifetime to go on.


That's rather obscure to me now. I plan to study more lifetimes and try different combinations to see what works. However, this solution seems correct.

nathanielc: Thanks, maybe a quick comment with a link to https://doc.rust-lang.org/nomicon/hrtb.html will help future readers learn why this is the case for closures? When I first read the code I didn't think it was specific to closures and so it was harder to understand its meaning. (I still don't fully understand, but I see why closures need special handling).

libflux/src/semantic/walk.rs
jlapacik: Is this better than using Rc<Node> in your opinion?

libflux/src/semantic/walk.rs
jlapacik: Can you explain this a little more. I don't see why the lifetime of Node is less than that of the visitor?

libflux/src/semantic/walk.rs
affo: Yeah, sure. There is and intermediate step, we could use `&mut Node`.

affo: We don't actually need multiple ownership, a reference to a mutable variable should be enough.

affo:
@nathanielc I am exploring using plain `Rc<Node>` in signatures.

This leads to a runtime exception caused by multiple mutable references to the same node:

```rust
let mut node = node;
    if v.visit(Rc::clone(&node)) {
        match Rc::get_mut(&mut node).unwrap() {
            Node::Package(ref mut n) => {
                for mut file in n.files.iter_mut() {
                    walk(v, Rc::new(Node::File(&mut file)));
                }
            }
        ...
        }
    ...
    }
```

`visit` clones the `Rc` creating a new reference. However, the visitor can invoke `get_mut` to mutate the node. This doesn't work, because the object has more than one reference to it, and you cannot `get_mut` when `ref count > 1`.

I cannot create a new `Rc` for visit, because, in that way, I would own the `Node` in the `Rc`.

If `Rc` is a strong requirement and we need mutability for types, I say that the only way around is to use `RefCell`.

At this point I would switch to `Rc<RefCell<Node>>` for interior mutability.

affo:
@nathanielc @jlapacik the last change introduces `Rc<RefCell<Node>>`.

If that's ok, I'll add some further  documentation to the visitor  trait to explain how to  create a visitor and add the doc with pointers to the nomicon for `for<'a>`.

affo:
@jlapacik @nathanielc Thank you!

I am adding a bit more documentation to the visitor so that external users have documentation and don't need to fall back to see tests and then I'll merge.

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1284 @@
...
+                    walk(v, Rc::new(RefCell::new(Node::Block(&mut *next))));
+                }
+                Block::Expr(ref mut estmt, ref mut next) => {
+                    walk(v, Rc::new(RefCell::new(Node::ExprStmt(estmt))));
+                    walk(v, Rc::new(RefCell::new(Node::Block(&mut *next))))
+                }
+                Block::Return(ref mut expr) => {
+                    walk(v, Rc::new(RefCell::new(Node::from_expr(expr))))
+                }
+            },
```
affo: walk on new block types

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1284 @@
...
+                    "File",
+                    "ExprStmt",
+                    "FunctionExpr",
+                    "Block::Return",
+                    "IntegerLit",
+                ],
+            )
+        }
+        #[test]
+        fn test_function_expr_multiline_block() {
```
affo: new test to check multiline block

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1284 @@
...
+            Node::RegexpLit(_) => write!(f, "RegexpLit"),
+            Node::ExprStmt(_) => write!(f, "ExprStmt"),
+            Node::OptionStmt(_) => write!(f, "OptionStmt"),
+            Node::ReturnStmt(_) => write!(f, "ReturnStmt"),
+            Node::TestStmt(_) => write!(f, "TestStmt"),
+            Node::BuiltinStmt(_) => write!(f, "BuiltinStmt"),
+            Node::Block(n) => match n {
+                Block::Variable(_, _) => write!(f, "Block::Variable"),
+                Block::Expr(_, _) => write!(f, "Block::Expr"),
+                Block::Return(expr) => write!(f, "Block::Return"),
```
affo: `fmt` specifies the block type

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1284 @@
...
+            Node::UintLit(n) => &n.loc,
+            Node::BooleanLit(n) => &n.loc,
+            Node::DateTimeLit(n) => &n.loc,
+            Node::RegexpLit(n) => &n.loc,
+            Node::ExprStmt(n) => &n.loc,
+            Node::OptionStmt(n) => &n.loc,
+            Node::ReturnStmt(n) => &n.loc,
+            Node::TestStmt(n) => &n.loc,
+            Node::BuiltinStmt(n) => &n.loc,
+            Node::Block(n) => n.loc(),
```
affo: `loc` for `Block` returns the location of the statement in the block, not the entire block. Should I comment on this maybe? 🤔

libflux/src/semantic/walk.rs
```diff
@@ -0,0 +1,1284 @@
...
+                    plus: vec![false],
+                    err: "".to_string(),
+                }
+            }
+        }
+
+        impl Visitor for RepeatedPlusChecker {
+            fn visit(&mut self, node: Rc<RefCell<Node>>) -> bool {
+                match *node.borrow() {
+                    Node::FunctionExpr(_) => {
```
affo: now this must match `FunctionExpr` cause we have multiple lines in a function body and every line is a `Block`

libflux/src/semantic/walk.rs
jlapacik: That makes sense. It's clear from the `impl`

Issue #1973, opened 2019-10-11
Instantiate Flux types from AST for type expressions


jlapacik:
#1952 will define an AST for type expressions. There should be a 1-1 mapping from the nodes of this AST to the type structures defined in #1951. We will need a function to perform this translation so that we can generate a type structure from a textual type expression.

Note this function will only be used in test cases. As such it should be defined inside of the main test module.

nathanielc:
Note concretely we mean converting the AST for this expression:

```
forall [t0] where to is Addable (a:t0) -> t0
```

to the Flux types defined in `libflux/src/semantic/type.rs`.

fchikwekwe:
What additional work is this issue asking for beyond outputting an AST representation of PolyTypes as defined in `libflux/src/semantic/type.rs`? I believe that I may have already resolved much of this issue with #1952.

fchikwekwe:
PR #2066 will resolve this issue for now once closed. Instead of creating a separate AST and then parsing type expressions into semantic types, I parsed them directly into semantic types. Due to this, there will have to be more work in the future to create a separate AST and then add type expressions into the Flux grammar.

fchikwekwe:
PR #2066 is now merged.

PR #2066, opened 2019-10-25
feat(libflux/semantic): allow lexing and parsing of string polytypes

fchikwekwe:
This is a parser to test polytype expressions for our implementation of Algorithm W. 

This parser provides the ability to lex a given PolyType string value, create tokens and parse those tokens into semantic types. Future work is still required as outlined in my comments on #1973. 

The output AST represents a 1 to 1 mapping to the types defined in #1951. 

Further notes and details are outlined as comments within `semantic/parser.mod.rs`. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/semantic/parser/grammar.md
+++ b/libflux/src/semantic/parser/grammar.md
@@ -1,0 +1,46 @@
+# Polytype Parser Grammar 
+
+[Modified from Joshua Lapacik's work here](https://github.com/jlapacik/ftel/blob/master/antlr/flux.g4)
+
+```
+? = optional; once or none 
+* = zero or more times
+| = or 
+(values) = a list of possibly repeating values
+single quotes ('') indicate a literal value
+
+
+polytype    = 'forall' '[' vars? ']' ('where' constraints)? monotype
+
+vars        = type_var (',' type_var)* 
+constraints = constraint ( (',' | 'and') constraint)* 
+constraint  = type_var (':') kinds
+kinds       = kind ( '+' kind)*
+kind        = IDENTIFIER 
+monotype    = type_var | primitive | array | row | function
+
+type_var    = 't' ([0-9])*
+primitive   = INT | FLOAT | STRING | BOOL | DURATION | TIME | REGEXP 
+array       = '[' monotype ']'
+row         = '{' properties? '}'
+function    = '(' arguments? ')' '->' monotype
+properties  = property ( '|' property )* ( '|' type_var)?
+property    = IDENTIFIER ':' monotype
+arguments   = argument ( ',' argument )*
+argument    = required | optional | pipe
+required    = IDENTIFIER ':' monotype
+optional    = '?' IDENTIFIER ':' monotype
+pipe        = '<-' IDENTIFIER? ':' monotype
+
+
+INT         = 'int'
+UINT        = 'uint'
+FLOAT       = 'float'
+STRING      = 'string'
+BOOL        = 'bool'
+DURATION    = 'duration'
+TIME        = 'time'
+REGEXP      = 'regexp'
+IDENTIFIER  = [a-zA-Z] ([0-9a-zA-Z])*
+WHITESPACE  = [ \t\r\n]+ -> skip
+```
--- a/libflux/src/semantic/parser/mod.rs
+++ b/libflux/src/semantic/parser/mod.rs
@@ -1,0 +1,1494 @@
+use std::{ str::Chars, iter::Peekable, slice::Iter, collections::{ HashMap, HashSet } };
+
+use crate::semantic::types::{ PolyType, MonoType, Tvar, Kind, TvarKinds, Row, Array, Function, Property };
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+// TokenType holds all possible TokenType values 
+pub enum TokenType {
+    ERROR,
+    EOF,
+    IDENTIFIER,
+    WHITESPACE,
+
+    // Keywords and Primitives
+    FORALL,
+    WHERE,
+    INT,
+    UINT,
+    FLOAT,
+    STRING,
+    BOOL,
+    DURATION, 
+    TIME,
+    REGEXP,
+
+    // Operators
+    LEFTCURLYBRAC,
+    RIGHTCURLYBRAC,
+    LEFTSQUAREBRAC,
+    RIGHTSQUAREBRAC,
+    LEFTPAREN,
+    RIGHTPAREN,
+    QUESTIONMARK,
+    COLON, 
+    COMMA, 
+    PIPE, 
+    ARROW, 
+    PLUS,
+    WITH
+}
+
+// Lexer holds the iterator for the source string, the list of output tokens and keeps track of
+// the current string value of any identifier or keyword to append the the Token.
+struct Lexer<'a> {
+    source: Peekable<Chars<'a>>, 
+    tokens: Vec<Token>, 
+    current_string: String,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+// Token holds the token's type and the text value for any keyword or identifier to be used later 
+// by the parser. 
+pub struct Token {
+    token_type: TokenType,
+    text: Option<String>,
+}
+
+// Lex instantiates the Lexer with default values and initializes lexing. 
+// This function is not meant to be used directly. The user should pass 
+// source into Parse(), which in turn calls this function. 
+pub fn Lex(source: &str) -> Vec<Token> {
+    let mut lexer = Lexer { 
+        source: source.chars().peekable(),
+        tokens: Vec::new(),
+        current_string: String::new(),
+        };
+    lexer.lex_tokens(); 
+    lexer.tokens
+}
+
+impl Lexer<'_> {
+    // lex_tokens calls lex_token while there are still characters to lex
+    fn lex_tokens(&mut self){
+        while let Some(token_type) = lex_token(self) {
+            if token_type == TokenType::WHITESPACE {
+                self.ignore(); 
+            } else {
+                self.emit(token_type); 
+            }
+
+            if token_type == TokenType::EOF {
+                break; 
+            }
+        }
+    }
+
+    // next grabs the next character from the string while there are still characters to lex
+    fn next(&mut self) -> Option<char> {
+        match self.source.next() {
+            None => None, 
+            Some(letter) => {
+                if letter.is_alphanumeric() {
+                    self.current_string.push(letter);
+                }
+                Some(letter)
+            }
+        }
+    }
+
+    // emit instatiates a Token and pushes that token along with its TokenType and text, if applicable,
+    // into the Lexer's tokens vector
+    fn emit(&mut self, token: TokenType) {
+        if !self.current_string.is_empty() {
+            let text = self.current_string.clone();
+            self.tokens.push(Token{ token_type: token, text: Some(text) })
+        } else {
+            self.tokens.push(Token{ token_type: token, text: None })
+        }
+        self.current_string = String::new();
+    }
+
+    // ignore empties self.current_string. This is used for whitespace tokens since they are not currently
+    // taken into consideration when parsing
+    fn ignore(&mut self) {
+        self.current_string = String::new();
+    }
+
+    fn keyword_or_ident(&mut self) -> TokenType {
+        while let Some(letter) = self.source.peek() {
+            if letter.is_alphanumeric() {
+                self.next();
+                continue; 
+            }
+            break;
+        }
+
+        let current: &str = &self.current_string; 
+        match current {
+            "forall" => TokenType::FORALL, 
+            "where" => TokenType::WHERE, 
+            "int" => TokenType::INT, 
+            "float" => TokenType::FLOAT, 
+            "string" => TokenType::STRING, 
+            "bool" => TokenType::BOOL, 
+            "uint" => TokenType::UINT, 
+            "duration" => TokenType::DURATION, 
+            "time" => TokenType::TIME, 
+            "regexp" => TokenType::REGEXP, 
+            _ => TokenType::IDENTIFIER
+        }
+    }
+}
+
+// lex_token lexes and returns a single token
+fn lex_token(lexer: &mut Lexer) -> Option<TokenType> {
+    match lexer.next() {
+        Some(letter) if letter.is_alphanumeric() =>  Some(lexer.keyword_or_ident()), 
+        Some(letter) if letter.is_whitespace() => Some(TokenType::WHITESPACE), 
+        Some(letter) if letter == '{' => Some(TokenType::LEFTCURLYBRAC), 
+        Some(letter) if letter == '}' => Some(TokenType::RIGHTCURLYBRAC), 
+        Some(letter) if letter == '[' => Some(TokenType::LEFTSQUAREBRAC), 
+        Some(letter) if letter == ']' => Some(TokenType::RIGHTSQUAREBRAC), 
+        Some(letter) if letter == '(' => Some(TokenType::LEFTPAREN), 
+        Some(letter) if letter == ')' => Some(TokenType::RIGHTPAREN), 
+        Some(letter) if letter == '?' => Some(TokenType::QUESTIONMARK), 
+        Some(letter) if letter == ':' => Some(TokenType::COLON), 
+        Some(letter) if letter == ',' => Some(TokenType::COMMA), 
+        Some(letter) if letter == '+' => Some(TokenType::PLUS), 
+        Some(letter) if letter == '|' => Some(TokenType::WITH),
+        Some(letter) if letter == '<' => {
+            match lexer.next() {
+                Some('-') => Some(TokenType::PIPE),
+                _ => Some(TokenType::ERROR), 
+            }
+        },
+        Some(letter) if letter == '-' => {
+            match lexer.next() {
+                Some('>') => Some(TokenType::ARROW), 
+                _ => Some(TokenType::ERROR), 
+            }
+        },
+        Some(_) => Some(TokenType::ERROR), 
+        _ => Some(TokenType::EOF), 
+    }
+}
+
+struct Parser<'a> {
+    tokens: Peekable<Iter<'a, Token>>,
+}
+
+// Parse passes the source text through the Lexer, It then initializes parsing 
+// and returns a PolyType representation.
+
+// This is the only function meant to be accessed by the end user. It handles 
+// both lexing and parsing of string polytypes. 
+pub fn Parse(source: &str) -> Result<PolyType, &'static str> {
+    let tokens = Lex(source); 
+    let mut parser = Parser{ tokens: tokens.iter().peekable() }; 
+    parser.parse_polytype()
+}
+
+impl Parser<'_> {
+    // next grabs the next token using the Iter()'s next method and unpacks
+    // the value if there are still tokens to parse
+    fn next(& mut self) -> Token {
+        match self.tokens.next() {
+            Some(token) => (*token).clone(),
+            None => Token {
+                token_type: TokenType::EOF,
+                text: None
+            },
+        }
+    }
+    
+    // peek returns a preview of the next Token using Iter()'s peek method
+    // and unpacks the value if there are still tokens to parse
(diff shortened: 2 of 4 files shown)


libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1729 @@
...
+                }
+                if !keyword_match{
+                    // token is not a keyword, must be an identifier
+                    self.emit(TokenTypes::IDENTIFIER);
+                    break;
+                }
+                break;
+            }
+        }
+    }
```
nathanielc: The logic between these two blocks in the `if {} else {}` are the same, can you simplify it to not repeat  the code? Maybe create a helper function or change the flow of this code so that its the final step of the function?

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1729 @@
...
+                expr: expr,
+                })
+            }
+            Err(e) => Err(e)
+        }
+        
+    }
+
+
+    /**Production rules mentioned below can be found in the accompanying grammar.md file */
```
nathanielc: Do all of the `parse_*` names match the names of the production rules in the grammar? If not can you make them match? I don't have a preference of whether we change the functions or the grammar, choose whatever name you think is best.

fchikwekwe: Yeah, that's a reasonable request. It also eliminates the need for the repetitive comments on  the various parsing methods. I think I will keep the grammar as is and change the functions.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+    }
+
+    fn emit(&mut self, token: TokenTypes) {
+        if !self.current_string.is_empty() {
+            let text = self.current_string.clone();
+            // if the token requires a string, it should always be in self.current_string
+            self.tokens.push(Token{ token_type: token, text: Some(text) })
+        } else {
+            self.tokens.push(Token{ token_type: token, text: None })
+        }
```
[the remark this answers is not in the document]
fchikwekwe: At least one case comes to mind where the parser logic checks to see if the next token has text or not. I think the logic could be changed, but I'd need to look into it a little more. 

As far as simplifying the logic, I'm not sure if this change would have a big impact. I already unpack the text in a helper function, so from the perspective of anyone using the Tokens as is, this seems like it would be minor. But perhaps it might cost less in terms of memory to just use an empty string rather than creating an `Option`.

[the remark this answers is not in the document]
fchikwekwe: For that initial case that I mentioned, I could also check if `String.is_empty()` if I change this per your suggestion.

libflux/src/semantic/parser/mod.rs
fchikwekwe: Thanks for catching that. I meant to go back and refactor this before opening my PR.

libflux/src/semantic/parser/mod.rs
fchikwekwe: Ended up making some small changes to both.

libflux/src/semantic/parser/mod.rs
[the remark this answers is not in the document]
fchikwekwe: As as example, the `parse_type_vars` function uses the text Option.

libflux/src/semantic/parser/mod.rs
fchikwekwe: I think that this update should be a little more DRY.

libflux/src/semantic/parser/grammar.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+INT         = 'int'
+UINT        = 'uint'
+FLOAT       = 'float'
+STRING      = 'string'
+BOOL        = 'bool'
+DURATION    = 'duration'
+TIME        = 'time'
+REGEX       = 'regular expression'
```
jsternberg: Would it be good to have this be `regex` or `regexp` instead? Spaces can mess things up when it comes to scanners and it may make it more clear if this is a single word like the rest of them.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+// Lexer holds the iterator for the source string, the list of output tokens and keeps track of
+// the current string value of any identifier or keyword to append the the Token.
+struct Lexer<'a> {
+    source: Peekable<Chars<'a>>, 
+    tokens: Vec<Token>, 
+    current_string: String,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+// Token hold the token's type and the text value for any keyword or identifier to be used later
```
jsternberg: *holds

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+        let addable = "Addable".to_string();
+        let subtractable = "Subtractable".to_string();
+        let divisible = "Divisible".to_string();
+        let comparable = "Comparable".to_string();
+        let nullable = "Nullable".to_string();
+        let equatable = "Equatable".to_string();
+
+        match token.text {
+            Some(text) => {
+                if text == addable {
```
jsternberg: Is it possible in Rust to create a map of strings and use that for this? This seems like a candidate for a map of strings to kinds.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+    let mut lexer = Lexer { 
+        source: source.chars().peekable(),
+        tokens: Vec::new(),
+        current_string: String::new(),
+        };
+    lexer.lex_tokens(); 
+    lexer.tokens
+}
+
+// StateFn maintains the lexers recursive state
```
jsternberg: What recursive state does the lexer have? I'm not too familiar with the language, but lexers shouldn't have recursive state. A parser may have recursive state, but most lexers don't. I'm looking and I don't see any location where `StateFn` is set to anything other than `lex_token`.

libflux/src/semantic/parser/grammar.md
fchikwekwe: I think that in practice I actually went for 'regex' for this. Sorry for the oversight.

libflux/src/semantic/parser/mod.rs
fchikwekwe: 🤦🏾‍♀️ whoops!

libflux/src/semantic/parser/mod.rs
fchikwekwe: `lex_token()` and `StateFn()` maintain the lexing behavior by calling one another while `Lexer.next()` iterates through the characters of the `source` string. I may be incorrect to call this behavior recursive, but that is how it appeared to me.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+        ("is", TokenTypes::IS),
+        ("and", TokenTypes::AND),
+        ("int", TokenTypes::INT),
+        ("float", TokenTypes::FLOAT),
+        ("string", TokenTypes::STRING),
+        ("bool", TokenTypes::BOOL),
+        ("uint", TokenTypes::UINT),
+        ("duration", TokenTypes::DURATION),
+        ("time", TokenTypes::TIME),
+        ("regex", TokenTypes::REGEX)
```
jlapacik: should be "regexp"

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+        if !keyword_match {
+            // token is not a keyword, must be an identifier
+            self.emit(TokenTypes::IDENTIFIER);
+        }        
+    }
+}
+
+// lex_token recursively lexes characters using StateFn
+fn lex_token(lexer: &mut Lexer) -> Option<StateFn> {
+    let keyword_pairs: Vec<(&str, TokenTypes)> = vec![
```
jlapacik: keyword_pairs seems like it should be a match expression instead of a map. Maybe inside of `keyword_or_ident`?

libflux/src/semantic/parser/grammar.md
```diff
@@ -0,0 +1,47 @@
...
+polytype    = 'forall' '[' vars? ']' ('where' constraints)? monotype
+
+vars        = typeVar (',' typeVar)* 
+constraints = constraint ( (',' | 'and') constraint)* 
+constraint  = typeVar ('is' | ':') kinds
+kinds       = kind ( '+' kind)*
+kind        = IDENTIFIER 
+monotype    = typeVar | typ | array | row | function
+
+type_var    = IDENTIFIER
```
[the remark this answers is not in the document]
fchikwekwe: We made a change to this so that all strings which aren't keywords are identifiers. I forgot to change it in the grammar. I'll go ahead and push that change.

libflux/src/semantic/parser/grammar.md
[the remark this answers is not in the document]
fchikwekwe: I went with 'regex' if that's okay.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1696 @@
...
+#[derive(Debug, PartialEq, Copy, Clone)]
+// TokenTypes holds all possible TokenType values used by the Lexer and Parser. 
+pub enum TokenTypes {
+    ERROR,
+    EOF, 
+
+    // Keywords and Kinds
+    FORALL,
+    WHERE,
+    IS, // not currently in use for Kinds; use ':' instead
```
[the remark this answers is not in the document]
fchikwekwe: Sounds good. 👍🏾

libflux/src/semantic/parser/mod.rs
fchikwekwe: Okay, I'll change this for consistency's sake.

libflux/src/semantic/parser/grammar.md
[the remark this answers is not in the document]
fchikwekwe: Actually, I'll change it to 'regexp' to be consistent with your code.

libflux/src/semantic/parser/mod.rs
fchikwekwe: Gotcha, thats a good suggestion.

libflux/src/semantic/parser/mod.rs
fchikwekwe: What if instead of converting this to a map, I got rid of the variables and simplified the match expression? Do you still think a map would be better?

fchikwekwe: I just don't like the idea of creating a whole map and iterating over it when I can do this with a map statement.

jsternberg: If a match expression works, go with that. I was thinking it would be a global map and usually you can just access the map. I don't think you need to iterate over it.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1729 @@
...
+                    },
+                }
+            Some(StateFn{ f: lex_token })
+            }
+        }
+        _ => {
+            lexer.emit(TokenTypes::EOF);
+            None
+        }
+    }
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, this reduced the visual complexity of this section by a lot. Thanks for the suggestion.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1658 @@
...
+            "duration" => { self.emit(TokenType::DURATION); },
+            "time" => { self.emit(TokenType::TIME); },
+            "regexp" => { self.emit(TokenType::REGEXP); },
+            _ => { keyword_match = false; } 
+        }
+        
+        if !keyword_match {
+            // token is not a keyword, must be an identifier
+            self.emit(TokenType::IDENTIFIER);
+        }
```
nathanielc: Wow, this is great. Much simpler thanks! 

One last simplification now is that the `keyword_match` boolean is not needed as this last if expression can be moved inside the `_ => ` section of the match expression.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1658 @@
...
+    INT,
+    UINT,
+    FLOAT,
+    STRING,
+    BOOL,
+    DURATION, 
+    TIME,
+    REGEXP,
+    IDENTIFIER,
+    WHITESPACE,
```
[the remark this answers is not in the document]
fchikwekwe: I'm still using it in my next update to know when to ignore. Jonathan and I paired for a little bit and he suggested a refactor that has lex_tokens handling more of the lexing behavior. I'll push my latest update son that you can see that change.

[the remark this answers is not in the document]
fchikwekwe: By it, I mean the `WHITESPACE` TokenType

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+        }
+
+        match monotype {
+            // if there's no monotype, there's no pipe argument 
+            Err(_e) => None, 
+            Ok(monotype) => {
+                // string is optional, so still return Property
+                match string {
+                    None => Some(Property {
+                        k: String::new(),
```
jlapacik: I think this should be the actual string `"<-"` since property names aren't optional.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+                retn: MonoType::Var(Tvar(0))
+            }))
+        };
+
+        assert_eq!(Ok(output), parse_tokens(tokens));
+    }
+
+    #[test]
+    fn parse_row_polytypes() {
+        let parse_text = "   forall [t1, t2] where t1: Nullable, t2: Comparable {test: t1, testAgain: bool, testLast: [uint]} ";
```
jlapacik: Just to reiterate this row type should be specified as:

    {test: t1 | testAgain: bool | testLast: [uint]}

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+                        tail: MonoType::Row(Box::new(Row::Empty)),
+                    }))
+                }))
+            }))
+        };
+
+        assert_eq!(Ok(output), parse_tokens(tokens));
+
+
+
```
jlapacik: nit, unnecessary white space

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+        }
+        match inner_prop {
+            None => Err("Unable to parse row MonoType"), 
+            Some(rows) => Ok(rows)
+        }
+    }
+}
+
+
+/** TESTS */
```
jlapacik: nit, can remove this comment.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+            Some(token) => (**token).clone(),
+            None => Token {
+                token_type: TokenType::EOF,
+                text: None
+            },
+        }
+    }
+
+
+    /*Production rules for each of the following methods can be found in the accompanying grammar.md file.
```
jlapacik: nit, can you remove this comment, or make it a doc comment over `parse_polytype` just to keep things consistent?

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+                Some('>') => Some(TokenType::ARROW), 
+                _ => Some(TokenType::ERROR), 
+            }
+        },
+        Some(_) => Some(TokenType::ERROR), 
+        _ => Some(TokenType::EOF), 
+    }
+}
+
+/** PARSER */
```
jlapacik: nit, can remove this

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1755 @@
...
+    RIGHTPAREN,
+    QUESTIONMARK,
+    COLON, 
+    COMMA, 
+    PIPE, 
+    ARROW, 
+    PLUS
+}
+
+/* LEXER */
```
jlapacik: nit, can remove this

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1718 @@
...
+        if self.next().token_type != TokenType::LEFTSQUAREBRAC {
+            return Err("Missing left square bracket")
+        }
+
+        let free_vars = self.parse_vars();
+        
+        let ok_free_vars = match free_vars {
+            Err(e) => { return Err(e) }
+            Ok(free_vars) => free_vars
+        };
```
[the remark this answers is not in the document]
fchikwekwe: Yeah I haven't worked with the `?` operator yet. Thanks for the suggestion

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1718 @@
...
+        match expr {
+            Ok(expr) => {
+                Ok(PolyType {
+                free: ok_free_vars,
+                cons: cons,
+                expr: expr,
+                })
+            }
+            Err(e) => Err(e)
+        }
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, this is a good suggestion.

libflux/src/semantic/parser/mod.rs
fchikwekwe: Do you only want the pipe to be included in the string in cases where the pipe argument doesn't have a variable name?

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1740 @@
...
+            } else {
+                return Err("Invalid arguments for this function.")
+            }
+            
+            token = self.next(); // check if next is right paren or comma
+            if token.token_type != TokenType::COMMA {
+                // if the next token is not comma, must be right paren
+                break;
+            } 
+            token = self.next(); // if its a comma, then there are more args to parse
```
fchikwekwe: @jsternberg you suggested that I try and use a match statement in here. Was that just a suggestion to make this a little more idiomatic or was there another reason for that change? I'm not seeing the added value in this particular case.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1718 @@
...
+    }
+}
+
+/** PARSER */
+
+struct Parser<'a> {
+    tokens: Peekable<Iter<'a, Token>>,
+}
+
+// parse_tokens is a testing function for directly parsing a vector of tokens.
```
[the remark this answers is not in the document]
fchikwekwe: I thought it might have been useful to have a few tests that just focused on the Lexer or the Parser. Testing parse tests them both together. I'll also need to refactor several of my tests.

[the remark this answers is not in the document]
fchikwekwe: I can go ahead and change this. We just lose the ability to test the parser in isolation.

libflux/src/semantic/parser/mod.rs
```diff
@@ -0,0 +1,1506 @@
...
+// }
+
+// Parse passes the source text through the Lexer, receives tokens and removes
+// whitespace. Its then initializes parsing and returns a PolyType AST.
+// This is the only function meant to be accessed by the end user. It handles 
+// both lexing and parsing of string polytypes. 
+pub fn Parse(source: &str) -> Result<PolyType, &'static str> {
+    let tokens = Lex(source); 
+    let mut parser = Parser{ tokens: tokens.iter().peekable() }; 
+    parser.parse_polytype()
```
nathanielc: One way to refactor this is to keep `parse_tokens` but write `Parse` in terms of `parse_tokens. 

```
pub fn Parse(source: &str) -> Result<PolyType, &'static str> {
    let tokens = Lex(source); 
    parse_tokens(tokens)
```

this way the decoupled nature of the code still exists so we can test things in isolation, while still presenting a simple API to end users of the code.

fchikwekwe: I went ahead and just removed `parse_tokens()`. Should be simple to add it back in if the need arises.

PR #2067, opened 2019-10-25
feat(values): add month support when adding durations to a time value

### LINKED ISSUE #2044: Make duration add operation month-aware
The duration add operation in `values/time.go` needs to be made month-aware. At the moment, any duration operation will treat a month as 30 days and the context of the time will not be used. This issue is about making the `(values.Time).Add(d values.Duration)` operation aware of which month it is in and following the spec for the rules regarding edge cases.

The easiest way to do this is likely to use `time.Time` values which are date-aware.

jsternberg:
The `Add` method follows the rules as specified in the spec for how
adding a duration to a time value should happen. It also clarifies that
adding a duration value twice is not the same thing as adding the
equivalent value once. For example, `t + 1mo + 1mo` is not the same as
`t + 2mo`. This is because of the rules regarding the end of the month.
At the same time, doing `t + 1mo * 2` is the same thing as adding `2mo`
because of the rules for how multiplication works with duration values.

The `Add` method is now aware of month boundaries and can be used
properly with years and months now.

This also adds support for using months for a window interval. If you do
`window(every: 1mo)`, it will create windows on month boundaries. The
`every` parameter for `window()` is limited to only using one of the two
base units: months or nanoseconds.

Fixes #2044.

- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -219,10 +219,12 @@ Examples:
     2018-01-31T00:00:00Z + 1mo      // 2018-02-28T00:00:00Z, February 31th is rolled back to the last day of the month, February 28th in 2018.
 
     // Addition and subtraction of durations to date times does not commute
-    2018-02-28T00:00:00Z + 1mo + 1d // 2018-03-29T00:00:00Z
-    2018-02-28T00:00:00Z + 1d + 1mo // 2018-04-01T00:00:00Z
-    2018-01-01T00:00:00Z + 2mo - 1d // 2018-02-28T00:00:00Z
-    2018-01-01T00:00:00Z - 1d + 3mo // 2018-03-31T00:00:00Z
+    2018-02-28T00:00:00Z + 1mo + 1d  // 2018-03-29T00:00:00Z
+    2018-02-28T00:00:00Z + 1d + 1mo  // 2018-04-01T00:00:00Z
+    2018-01-01T00:00:00Z + 2mo - 1d  // 2018-02-28T00:00:00Z
+    2018-01-01T00:00:00Z - 1d + 3mo  // 2018-03-31T00:00:00Z
+    2018-01-31T00:00:00Z + 1mo + 1mo // 2018-03-28T00:00:00Z
+    2018-01-31T00:00:00Z + 2mo       // 2018-03-31T00:00:00Z
 
     // Addition and subtraction of durations to date times applies months, days and seconds in that order.
     2018-01-28T00:00:00Z + 1mo + 2d // 2018-03-02T00:00:00Z
@@ -231,6 +233,14 @@ Examples:
     2018-02-01T00:00:00Z + 2mo2d    // 2018-04-03T00:00:00Z
     2018-01-01T00:00:00Z + 1mo30d   // 2018-03-02T00:00:00Z, Months are applied first to get February 1st, then days are added resulting in March 2 in 2018.
     2018-01-31T00:00:00Z + 1mo1d    // 2018-03-01T00:00:00Z, Months are applied first to get February 28th, then days are added resulting in March 1 in 2018.
+
+    // Multiplication works
+    2018-01-01T00:00:00Z + 1mo * 1  // 2018-02-01T00:00:00Z
+    2018-01-01T00:00:00Z + 1mo * 2  // 2018-03-01T00:00:00Z
+    2018-01-01T00:00:00Z + 1mo * 3  // 2018-04-01T00:00:00Z
+    2018-01-31T00:00:00Z + 1mo * 1  // 2018-02-28T00:00:00Z
+    2018-01-31T00:00:00Z + 1mo * 2  // 2018-03-31T00:00:00Z
+    2018-01-31T00:00:00Z + 1mo * 3  // 2018-04-30T00:00:00Z
 
 [IMPL#657](https://github.com/influxdata/platform/issues/657) Implement Duration vectors
 
--- a/values/time.go
+++ b/values/time.go
@@ -58,16 +58,58 @@ func (t Time) Round(d Duration) Time {
 	return t + Time(d.Duration()-r.Duration())
 }
 
-func (t Time) Truncate(d Duration) Time {
-	if !d.IsPositive() {
-		return t
-	}
-	r := t.Remainder(d)
-	return t - Time(r.Duration())
-}
-
 func (t Time) Add(d Duration) Time {
-	return t + Time(d.Duration())
+	newT := t
+	if d.months > 0 {
+		// Determine if the number of months is positive or negative.
+		months := d.months
+		if d.negative {
+			months = -months
+		}
+
+		// Retrieve the current date and increment the values
+		// based on the number of months.
+		ts := t.Time()
+		year, month, day := ts.Date()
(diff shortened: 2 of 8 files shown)


values/time.go
```diff
@@ -67,7 +67,41 @@ func (t Time) Truncate(d Duration) Time {
...
 
 func (t Time) Add(d Duration) Time {
-	return t + Time(d.Duration())
+	newT := t
+	if d.months > 0 {
+		// Retrieve the current date and increment the values
+		// based on the number of months.
+		ts := t.Time()
+		year, month, day := ts.Date()
+		year += int(d.months / 12)
```
wolffcm: Shouldn't you subtract here instead for negative durations?

values/time_test.go
```diff
@@ -80,6 +81,115 @@ func TestTime_Truncate(t *testing.T) {
...
+			t:    "2000-01-01T00:00:00Z",
+			d:    "1y",
+			want: "2001-01-01T00:00:00Z",
+		},
+		{
+			// Is a leap year. Multiple of 400.
+			t:    "2000-01-31T00:00:00Z",
+			d:    "1mo",
+			want: "2000-02-29T00:00:00Z",
+		},
```
wolffcm: It seems like you need test cases for negative durations here?

values/time.go
jsternberg: Yea I'll add test cases for negative durations. I messed up and forgot to do that.

docs/SPEC.md
```diff
@@ -222,10 +222,11 @@ Examples:
...
     // Addition and subtraction of durations to date times does not commute
-    2018-02-28T00:00:00Z + 1mo + 1d // 2018-03-29T00:00:00Z
-    2018-02-28T00:00:00Z + 1d + 1mo // 2018-04-01T00:00:00Z
-    2018-01-01T00:00:00Z + 2mo - 1d // 2018-02-28T00:00:00Z
-    2018-01-01T00:00:00Z - 1d + 3mo // 2018-03-31T00:00:00Z
+    2018-02-28T00:00:00Z + 1mo + 1d  // 2018-03-29T00:00:00Z
+    2018-02-28T00:00:00Z + 1d + 1mo  // 2018-04-01T00:00:00Z
+    2018-01-01T00:00:00Z + 2mo - 1d  // 2018-02-28T00:00:00Z
+    2018-01-01T00:00:00Z - 1d + 3mo  // 2018-03-31T00:00:00Z
+    2018-01-31T00:00:00Z + 1mo + 1mo // 2018-03-28T00:00:00Z
```
nathanielc: Lets add this one too `2018-01-31T00:00:00Z + 2mo // 2018-03-31T00:00:00Z`

jsternberg: Updated.

values/time_test.go
jsternberg: Added the test cases and made it work with negative values.

jsternberg:
Updated this PR. I also checked and the semantic analyze stuff doesn't produce durations with month values so it is not possible to hit this code yet. We have to change `semantic/analyze.go` before it would hit that.

jsternberg:
I may have run into an issue with some edge cases that are allowed by the spec, but seem to cause a problem in implementation. I did some research on it and put the answers I found here: https://gist.github.com/jsternberg/063283aa9794cfb5b66c951cb47e94c2

nathanielc:
> One of the first intervals was 1970-05-13 to 1970-06-26 when using 1mo2w

Can you clarify what was confusing about this interval? To me it was not unexpected. If I hand wave I see the duration of  `1mo2w` as roughly a month and a half. And this interval goes from just before half way through May to just before the end of June and so it reads really close to a month and a half.

> I do not believe we can use multiplication to compute the interval which means that the only way we can compute the start of an interval is to repeatedly add the duration to the offset.

Does repeated addition not produce these odd intervals? If so why not, I would expect it to produce even more odd intervals because the _error_ accumulates.

For example using a quick script using repeated addition I get a very similar interval of `1970-05-13 to 1970-06-27` which is only one day different. (note I am using AddDate() which doesn't honor our last day of month, but in this case it shouldn't matter as the code did not hit that case in the short example).

https://play.golang.org/p/CbeSegSk2A_j


> The math for this is pretty simple when the window interval duration is only composed of nanoseconds or only composed of months. It can be computed in constant time because it involves using the modulo operator.

Can you share some simple code for this simplified operation so we can speak more concretely about the difference? I think I see how this can be done for months but I want to be sure we are on the same page for the simple case first.

jsternberg:
> Can you clarify what was confusing about this interval? To me it was not unexpected. If I hand wave I see the duration of 1mo2w as roughly a month and a half. And this interval goes from just before half way through May to just before the end of June and so it reads really close to a month and a half.

That's the thing. If I were a user, I wouldn't expect `1mo2w` to be roughly a month and a half. I'd expect it to be a month and then 2 weeks. I see the month, but 2 weeks after 13 days is the 27th not the 26th. But I think that even if we did say this inaccuracy was fine, the multiplication method just means the operation ends up being `O(log n)` instead of `O(n)` because we can attempt a binary search for the start. It still doesn't result in a constant time.

> Does repeated addition not produce these odd intervals? If so why not, I would expect it to produce even more odd intervals because the error accumulates.
>
> For example using a quick script using repeated addition I get a very similar interval of 1970-05-13 to 1970-06-27 which is only one day different. (note I am using AddDate() which doesn't honor our last day of month, but in this case it shouldn't matter as the code did not hit that case in the short example).

It produces no odd intervals because there's no error to accumulate. It always adds a month and then 2 weeks to the current date that it has so if you look at a single interval, it always makes sense. It's still not possible to intuitively find the start interval for a single time, but that's not expected.

> Can you share some simple code for this simplified operation so we can speak more concretely about the difference? I think I see how this can be done for months but I want to be sure we are on the same page for the simple case first.

Here's the code for truncation for only nanoseconds or only months.

```go
// truncateByNsecs will truncate the time to the given number
// of nanoseconds.
func truncateByNsecs(t Time, d Duration) Time {
	remainder := int64(t) % d.Nanoseconds()
	return t - Time(remainder)
}

// truncateByMonths will truncate the time to the given
// number of months.
func truncateByMonths(t Time, d Duration) Time {
	ts := t.Time()
	year, month, _ := ts.Date()

	// Determine the total number of months and truncate
	// the number of months by the duration amount.
	total := int64(year*12) + int64(month-1)
	remainder := total % d.Months()
	total -= remainder

	// Recreate a new time from the year and month combination.
	year, month = int(total/12), time.Month(total%12)+1
	ts = time.Date(year, month, 1, 0, 0, 0, 0, time.UTC)
	return values.ConvertTime(ts)
}
```

The complexity for months is mostly hidden around the internal implementation of `time.Time` so, if written from scratch, it wouldn't be that simple, but this is reasonably simple. But the logic for these is the same. Find the number of the unit, find the remainder, subtract the remainder, convert it back.

jsternberg:
@nathanielc and I talked and we both agreed that durations with mixed base units caused too much trouble, but it only caused trouble when used as the `every` parameter. The PR has been updated to reflect this so the only restriction on mixed duration units comes when creating a window.

PR #2073, opened 2019-10-28
feat(semantic): interpret months as part of the semantic duration

### LINKED ISSUE #2069: Make the interpreter read in duration values with months
The flux interpreter will read in duration values and translate them into a `time.Duration` before converting it into a `values.Duration`. This loses the special units such as months.

This issue is about making it so the duration units are part of the semantic node. This can be complicated because the `semantic` package cannot depend on the `values` package so there might need to be some refactoring involved with this.

jsternberg:
This changes the semantic duration literal to keep the months and other
duration units from the AST rather than converting it into a
`time.Duration`.

This also updates the semantic node's json encoding to emit these
magnitudes and units similar to the AST rather than a string
representation of the duration.

In total, this causes the semantic analyzer to create the
`Duration` value properly.

BREAKING CHANGE: The semantic node's duration literal has a different
json encoding and a different struct signature.

Fixes #2069.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/values/time.go
+++ b/values/time.go
@@ -108,6 +108,12 @@ func (t Time) Time() time.Time {
 // Mul will multiply the Duration by a scalar.
 // This multiplies each component of the vector.
 func (d Duration) Mul(scale int) Duration {
+	// If the duration is zero, do nothing.
+	// This prevents a zero value from becoming negative
+	// which is not possible.
+	if d.IsZero() {
+		return d
+	}
 	if scale < 0 {
 		scale = -scale
 		d.negative = !d.negative
@@ -121,6 +127,12 @@ func (d Duration) Mul(scale int) Duration {
 // It returns false if the number is zero.
 func (d Duration) IsPositive() bool {
 	return !d.negative && !d.IsZero()
+}
+
+// IsNegative returns true if this is a negative number.
+// It returns false if the number is zero.
+func (d Duration) IsNegative() bool {
+	return d.negative
 }
 
 // IsZero returns true if this is a zero duration.
@@ -149,8 +161,8 @@ func (d Duration) Equal(other Duration) bool {
 }
 
 // Duration will return the nanosecond equivalent
-// of this duration. It will assume that months are
-// the equivalent of 30 days.
+// of this duration. It will assume that all months are
+// the same length.
 //
 // It is recommended not to use this method unless
 // it is absolutely needed. This method will lose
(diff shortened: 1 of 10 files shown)


wolffcm:
Why is a change to `semantic` a breaking change?  My understanding was that it could be made into an `internal` package.

jsternberg:
It changes the serialization for the semantic node which I don't believe is used anywhere. But, since it is a publicly exposed package, I was careful and just marked it as breaking.

It also changes the struct of a public package. It may indicate that parts of semantic should become part of an internal package so that changes like this aren't breaking changes.

values/time.go
```diff
@@ -108,6 +108,12 @@ func (t Time) Time() time.Time {
 // Mul will multiply the Duration by a scalar.
 // This multiplies each component of the vector.
 func (d Duration) Mul(scale int) Duration {
+	// If the duration is zero, do nothing.
+	// This prevents a zero value from becoming negative
+	// which is not possible.
+	if d.IsZero() {
```
wolffcm: I'm curious, is there anywhere where if a negative zero managed to creep in, would it have adverse effects at runtime?

values/time.go
```diff
@@ -149,47 +161,51 @@ func (d Duration) Equal(other Duration) bool {
...
-// the equivalent of 30 days.
+// of this duration. It will assume that all months are
+// the same length.
 //
 // It is recommended not to use this method unless
 // it is absolutely needed. This method will lose
 // any precision that is present in the Duration
 // and it should only be used for interfacing with
 // outside code that is not month-aware.
 func (d Duration) Duration() time.Duration {
```
wolffcm: I wonder if this should have a scarier name, given that it could produce misleading results.

values/time.go
```diff
@@ -211,29 +227,81 @@ func (d Duration) String() string {
...
+}
+
+func (d Duration) String() string {
+	values := d.AsValues()
+	if len(values) == 0 {
+		return "0ns"
+	}
+
+	var buf []byte
+	writeInt := func(sb *strings.Builder, v int64, unit string) {
```
wolffcm: I thought I reviewed very similar code somewhere else.  Where was it?  If there is some duplication here, maybe it could be removed somehow?  If not, no big deal.

jsternberg: I think it was taken from the month addition PR since I needed it here.

values/time.go
jsternberg: Probably. I view this as likely a temporary method, but it always seems these things stick around for longer than "temporary". It was previously used though and I would prefer not to change the call site for a bunch of places when I don't want the function to exist anymore.

values/time.go
jsternberg: I don't think so. I think it would just end up with the `String()` function printing `0ns` when the duration was negative zero instead of `-0ns`. It could also result in `IsNegative()` returning true when the value is zero which isn't supposed to be possible.

PR #2074, opened 2019-10-28
feat(flux): Visitor uses Rc for nodes

bthesorceror:
given that the visitor might want to keep a reference to a node
the visit function and done function should be called with a Rc
to that node so the visitor can clone.

### DIFF
--- a/libflux/src/ast/walk/mod.rs
+++ b/libflux/src/ast/walk/mod.rs
@@ -231,23 +231,35 @@ impl<'a> Node<'a> {
 // Done is called on that Node to signal that we are done with that Node.
 //
 // If Visit returns None, walk will not recurse on the children.
-pub trait Visitor: Sized {
+//
+// Note: the Rc in visit and done is to allow for multiple ownership of a node, i.e.
+//       a visitor can own a node as well as the walk funciton. This allows
+//       for nodes to persist outside the scope of the walk function and to
+//       be cleaned up once all owners have let go of the reference.
+pub trait Visitor<'a>: Sized {
     // Visit is called for a node.
     // The returned visitor will be used to walk children of the node.
     // If visit returns None, walk will not recurse on the children.
-    fn visit(&self, node: &Node) -> Option<Self>;
+    fn visit(&self, node: Rc<Node<'a>>) -> Option<Self>;
     // Done is called for a node once it has been visited along with all of its children.
-    fn done(&self, _: &Node) {} // default is to do nothing
+    fn done(&self, _: Rc<Node<'a>>) {} // default is to do nothing
 }
 
 // Walk recursively visits children of a node.
 // Nodes are visited in depth-first order.
-pub fn walk<T>(v: &T, node: Node)
+pub fn walk<'a, T>(v: &T, node: Node<'a>)
 where
-    T: Visitor,
+    T: Visitor<'a>,
 {
-    if let Some(w) = v.visit(&node) {
-        match node {
+    walk_rc(v, Rc::new(node));
+}
+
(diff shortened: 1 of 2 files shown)


libflux/src/ast/walk/mod.rs
```diff
@@ -231,109 +231,114 @@ impl<'a> Node<'a> {
...
+// Note: the Rc in visit is to allow for multiple ownership of a node, i.e.
+//       a visitor can own a node as well as the walk funciton. This allows
+//       for nodes to persist outside the scope of the walk function and to
+//       be cleaned up once all owners have let go of the reference.
+pub trait Visitor<'b>: Sized {
     // Visit is called for a node.
     // The returned visitor will be used to walk children of the node.
     // If visit returns None, walk will not recurse on the children.
-    fn visit(&self, node: &Node) -> Option<Self>;
+    fn visit<'a>(&self, node: Rc<Node<'b>>) -> Option<Self>;
```
nathanielc: nit, can you remove the `<'a>` from the function since its not used and then rename `'b` to `'a` for consistency?

libflux/src/ast/walk/mod.rs
```diff
@@ -231,109 +231,114 @@ impl<'a> Node<'a> {
...
 }
 
 // Walk recursively visits children of a node.
 // Nodes are visited in depth-first order.
-pub fn walk<T>(v: &T, node: Node)
+pub fn walk<'a, T>(v: &T, node: Rc<Node<'a>>)
 where
-    T: Visitor,
+    T: Visitor<'a>,
 {
```
nathanielc: Should we create separate public/private functions to encapsulate the creation of the Rc for the first node?

Something like:

```
pub fn walk<T>(v: &T, node: Node<'a>)
where
    T: Visitor<'a>,
{
    walk_rec(visitor, Rc::new(node))
}


fn walk_rec<'a, T>(v: &T, node: Rc<Node<'a>>) 
where
    T: Visitor<'a>,
{
   // normal recursive walk function definition
}

```

Thoughts?

libflux/src/ast/walk/mod.rs
```diff
@@ -231,109 +231,114 @@ impl<'a> Node<'a> {
 // Done is called on that Node to signal that we are done with that Node.
 //
 // If Visit returns None, walk will not recurse on the children.
-pub trait Visitor: Sized {
+//
+// Note: the Rc in visit is to allow for multiple ownership of a node, i.e.
```
nathanielc: nit, `visit and done`

libflux/src/ast/walk/mod.rs
bthesorceror: Would the private method then recurse to the public on? If that's the case we could remove all the Rc creation in walk_rec.

libflux/src/ast/walk/mod.rs
```diff
@@ -231,23 +231,35 @@ impl<'a> Node<'a> {
...
+//       be cleaned up once all owners have let go of the reference.
+pub trait Visitor<'a>: Sized {
     // Visit is called for a node.
     // The returned visitor will be used to walk children of the node.
     // If visit returns None, walk will not recurse on the children.
-    fn visit(&self, node: &Node) -> Option<Self>;
+    fn visit(&self, node: Rc<Node<'a>>) -> Option<Self>;
     // Done is called for a node once it has been visited along with all of its children.
-    fn done(&self, _: &Node) {} // default is to do nothing
+    fn done(&self, _: Rc<Node>) {} // default is to do nothing
```
nathanielc: missed this the first time, the Node here should have the explicit lifetime `'a` as well to be consistent.

PR #2077, opened 2019-10-29
docs(spec): restrict intervals to only use one type of base unit for a duration

jsternberg:
The spec now specifies that the base units for a duration unit are
months and nanoseconds and each of the other units corresponds to one of
these two rather than leaving this unspecified. It also continues to say
that both of these may be used with each other.

It adds a restriction for intervals, which are used for windows, that
says that a duration used for an interval can only correspond to one of
these base units. This is because it is possible for us to have regular
intervals when only one of these base units is used, but it would be
difficult to reason and verify about what each window should look like
when these are mixed and there isn't a compelling reason to allow the
duration units to be mixed when it concerns intervals.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -167,30 +167,27 @@ When several durations are specified together, larger units must appear before s
     duration_lit  = { int_lit duration_unit } .
     duration_unit = "y" | "mo" | "w" | "d" | "h" | "m" | "s" | "ms" | "us" | "µs" | "ns" .
 
-| Units    | Meaning                                 |
-| -----    | -------                                 |
-| y        | year (12 months)                        |
-| mo       | month                                   |
-| w        | week (7 days)                           |
-| d        | day                                     |
-| h        | hour (60 minutes)                       |
-| m        | minute (60 seconds)                     |
-| s        | second                                  |
-| ms       | milliseconds (1 thousandth of a second) |
-| us or µs | microseconds (1 millionth of a second)  |
-| ns       | nanoseconds (1 billionth of a second)   |
+| Units    | Meaning                                 | Base |
+| -----    | -------                                 | ---- |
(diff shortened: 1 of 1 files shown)


jsternberg:
I kept the mention of base units in the duration literals because I felt the internal representation had now become the spec representation, but I removed any mentions of intervals from it and removed the language saying it wasn't allowed. That language has now been moved to the section about intervals along with an example that doesn't work.

docs/SPEC.md
```diff
@@ -202,7 +199,7 @@ Examples:
     1h15m  // 1 hour and 15 minutes
     5w
     1mo5d  // 1 month and 5 days
-    -1mo5d // negative 1 month and 5 days
+    -1y    // negative 1 year
```
nathanielc: Is there a reason to change this line?

docs/SPEC.md
```diff
@@ -1197,6 +1194,8 @@ Intervals has the following parameters:
 The Nth interval start date is the initial start date plus the offset plus an Nth multiple of the every parameter.
 Each interval stop date is equal to the interval start date plus the period duration.
 When filtering intervals each potential interval is passed to the filter function, when the function returns false, that interval is excluded from the set of intervals.
+The duration for an interval must contain duration units for only one base unit as defined in the duration literals.
+It is not possible to mix months and nanoseconds in an interval.
```
nathanielc: I think we should be more specific about the `every` parameter, otherwise it will be confusing which parameters allow mixed and which do not.

docs/SPEC.md
jsternberg: Ah the reason was because I didn't want to mix the duration units and then the PR changed so I'll change this back.

PR #2078, opened 2019-10-29
feat(ast): convert Rust AST to FlatBuffers format

wolffcm:
This adds a new module under `ast` called `flatbuffers` that encapsulates both the code generated by `flatc`, code to serialize an AST into the flatbuffers representation and related tests.  I did some moving around of files related to serialization to more closely match other modules.


libflux/src/ast/walk/mod.rs
```diff
@@ -224,29 +224,36 @@ impl<'a> Node<'a> {
...
+/// Done is called on that Node to signal that we are done with that Node.
+///
+/// If Visit returns None, walk will not recurse on the children.
+///
+/// Note: the Rc in visit and done is to allow for multiple ownership of a node, i.e.
+///       a visitor can own a node as well as the walk funciton. This allows
+///       for nodes to persist outside the scope of the walk function and to
+///       be cleaned up once all owners have let go of the reference.
+///
+/// Implementors of the Visitor trait will typically wrap themselves in Rc and RefCell
```
wolffcm: I converted this to a doc comment so that it will be rendered by `cargo doc`, and also added this section here.

libflux/src/lib.rs
```diff
@@ -8,9 +8,9 @@ pub mod parser;
 pub mod scanner;
 pub mod semantic;
 
-use std::ffi::*;
```
wolffcm: I didn't change this file, these are all changes from `cargo fmt`.

PR #2081, opened 2019-10-30
feat(stdlib/http): Add http.get flux method

nisley:
Add GET to http.go

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/http/http.go
```diff
@@ -0,0 +1,235 @@
...
+// the rest. This allows sockets to be reused.
+const maxResponseBody = 512 * 1024 // 512 KB
+
+var httpGetStatusCodeOnly = values.NewFunction(
+	"get",
+	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"url":     semantic.String,
+			"headers": semantic.Tvar(1),
+			"data":    semantic.Bytes,
```
nathanielc: As per your comment below do we need the `data` parameter to this function? Does your use case need it? If not then I say lets remove it until its clear we need it.

stdlib/http/http.go
```diff
@@ -0,0 +1,235 @@
...
+		},
+		Required: []string{"url"},
+		Return:   semantic.Int,
```
nathanielc: How do you plan to use the `http.get` function since it only returns the status code of the request but otherwise ignores the response?

stdlib/http/http.go
```diff
@@ -0,0 +1,235 @@
...
+			defer s.Finish()
+
+			req = req.WithContext(cctx)
+			response, err := dc.Do(req)
+			if err != nil {
+				return 0, err
+			}
+
+			// Read the response body but limit how much we will read.
+			// This is to allow a socket to be reused after it is closed.
```
nathanielc: I am not familiar with this idea of a limit reader that enables the socket to be reused. Can you point to some docs somewhere that explain this?

stdlib/http/http.go
nisley: Fair enough. I was torn since the Go client accepted and used data that perhaps I should include it, but that isn't overly compelling.

stdlib/http/http.go
nisley: This get is almost a straight copy of the flux post request code that only returns statusCode.  My next plan is to implement get(url: "", headers: "", responseType: "string|json")

stdlib/http/http.go
nisley: https://github.com/influxdata/flux/blob/7ad7e2ba275c45e21a5109766b36c8848e9a4e0f/stdlib/http/post.go#L104

nathanielc: I see it comes from e35a585d1. Basically if we can read the response body then we can reuse the connection. If we don't read it then it just need to be closed to tell the other side we don't want the rest of the data. But in that case we can't reuse the connection. But instead of just reading all of the response we only read up to 512KB to capture most use cases while still protecting us against very large responses.

nisley:
The use case without the response body is just a simple pinger.  Sometimes, a simple success request is all one is looking for.

nisley:
closing in favor of https://github.com/influxdata/flux/pull/2094

PR #2087, opened 2019-11-01
docs(spec): design how time zones function in the spec

### LINKED ISSUE #2024: Design how time zones will look and function within Flux
This issue covers designing how time zones will look and function within Flux. We know how it functions in InfluxQL and we need to determine how it will look in Flux as we may want to keep some things and do things differently in others.

This involves updating the spec with the decisions that we make upon review. Before updating the spec, it may be helpful to sketch out some ideas and post them to this issue before doing the actual write up.

TODO write up how they work in influxql for context.

jsternberg:
The spec has been updated to detail how time zones work and function in
the context of existant times and an expansion on how they are specified
within a flux script.

It also describes the behavior for what happens around time zone
switches including some strange edge cases. In general, a window can be
described as adding a certain interval value to a clock. In some cases,
this causes the clock to go back in time and it causes the same hour to
happen again. This time interval where the clock goes back just means
that there are two intervals worth of data in a single interval.

This becomes more interesting when you add offsets and intervals that
are smaller than the skip in time. This can cause multiple distinct time
ranges to be created within the same interval.

Fixes #2024.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1292,12 +1292,20 @@ Examples using known start and stop dates:
 #### Location support
 
 Intervals have location support for changing offsets by default.
-When a location has been specified either with the `location` option or the `location` parameter, a change in offset is used to either expand or shrink an interval.
+When a location has been specified either with the `location` option or the `location` parameter, a change in offset is used to either expand or shrink an interval if the interval is greater than the change in offset.
 The most common example is daylight savings time switches.
 When daylight savings time starts, the zone offset shifts west, becoming a more negative number.
 When this happens, the hour that was lost does not exist and the current window is shortened to accomodate this change.
+This happens at the time that the offset occurs.
+If the remaining time in the window is less than the offset change, then the window ends at the time of the offset switch and the rest of the offset change is subtracted from the next interval.
 When daylight savings time ends, the zone offset shifts east, becoming a more positive number.
 When this happens, one of the hours happens twice and so the interval becomes longer.
+If the offset change happens at the end of a window, then the offset will not be added to that window since the offset change did not occur in the window.
+Conversely, if the offset change happens at the start of a window, the offset change will be applied to that window.
+These adjustments only happen when the offset change is less than the interval duration.
+These changes happen independently for the `every` and `period`.
+These changes only occur if the location has been set to a named location.
+There are no offset changes if the location is UTC or set to a fixed offset.
 
 Examples:
 
@@ -1315,9 +1323,23 @@ Examples:
     // [2019-03-10T03:00:00-06:00, 2019-03-10T04:00:00-06:00)
     // [2019-03-10T04:00:00-06:00, 2019-03-10T06:00:00-06:00)
     // ... skipped ...
-    // [2019-11-03T00:00:00-06:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T01:00:00-07:00)
+    // [2019-11-03T01:00:00-07:00, 2019-11-03T04:00:00-07:00)
+    // [2019-11-04T04:00:00-07:00, 2019-11-03T06:00:00-07:00)
+
+    intervals(every:1h, period:2h, location: "America/Denver")
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T01:00:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T04:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T05:00:00-06:00)
+    // [2019-03-10T04:00:00-06:00, 2019-03-10T06:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T01:00:00-07:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T03:00:00-07:00)
+    // [2019-11-03T01:00:00-07:00, 2019-11-03T04:00:00-07:00)
     // [2019-11-03T02:00:00-07:00, 2019-11-03T04:00:00-07:00)
-    // [2019-11-04T03:00:00-07:00, 2019-11-03T06:00:00-07:00)
+    // [2019-11-04T03:00:00-07:00, 2019-11-03T05:00:00-07:00)
+    // [2019-11-04T04:00:00-07:00, 2019-11-03T06:00:00-07:00)
 
     intervals(every:1h, location: "America/Denver")
     // [2019-03-10T00:00:00-07:00, 2019-03-10T01:00:00-07:00)
@@ -1326,7 +1348,8 @@ Examples:
     // [2019-03-10T04:00:00-06:00, 2019-03-10T05:00:00-06:00)
     // ... skipped ...
     // [2019-11-03T00:00:00-06:00, 2019-11-03T01:00:00-06:00)
-    // [2019-11-03T01:00:00-06:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T01:00:00-07:00)
+    // [2019-11-03T01:00:00-07:00, 2019-11-03T02:00:00-07:00)
     // [2019-11-03T02:00:00-07:00, 2019-11-03T03:00:00-07:00)
     // [2019-11-03T03:00:00-07:00, 2019-11-03T04:00:00-07:00)
 
@@ -1340,19 +1363,30 @@ Examples:
     // ... skipped ...
     // [2019-11-03T00:00:00-06:00, 2019-11-03T00:30:00-06:00)
     // [2019-11-03T00:30:00-06:00, 2019-11-03T01:00:00-06:00)
-    // [2019-11-03T01:00:00-06:00, 2019-11-03T01:30:00-06:00), [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
-    // [2019-11-03T01:30:00-06:00, 2019-11-03T02:00:00-06:00), [2019-11-03T01:30:00-07:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T01:30:00-06:00)
+    // [2019-11-03T01:30:00-06:00, 2019-11-03T01:00:00-07:00)
+    // [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
+    // [2019-11-03T01:30:00-07:00, 2019-11-03T02:00:00-07:00)
     // [2019-11-03T02:00:00-07:00, 2019-11-03T02:30:00-07:00)
     // [2019-11-03T02:30:00-07:00, 2019-11-03T03:00:00-07:00)
-    
+
+    intervals(every:2h, offset:30m, location: "America/Denver")
+    // [2019-03-10T00:30:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T04:30:00-06:00)
+    // [2019-03-10T04:30:00-06:00, 2019-03-10T06:30:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:30:00-06:00, 2019-11-03T02:30:00-07:00)
+    // [2019-11-03T02:30:00-07:00, 2019-11-03T04:30:00-07:00)
+    // [2019-11-03T04:30:00-07:00, 2019-11-03T06:30:00-07:00)
+
     intervals(every:1h, offset:30m, location: "America/Denver")
     // [2019-03-10T00:30:00-07:00, 2019-03-10T01:30:00-07:00)
-    // [2019-03-10T01:30:00-07:00, 2019-03-10T03:00:00-06:00)
-    // [2019-03-10T03:00:00-06:00, 2019-03-10T03:30:00-06:00)
+    // [2019-03-10T01:30:00-07:00, 2019-03-10T03:30:00-06:00)
     // [2019-03-10T03:30:00-06:00, 2019-03-10T04:30:00-06:00)
     // ... skipped ...
-    // [2019-11-03T00:30:00-06:00, 2019-11-03T01:30:00-06:00), [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
-    // [2019-11-03T01:30:00-06:00, 2019-11T01:00:00-07:00), [2019-11-03T02:00:00-07:00, 2019-11-03T02:30:00-07:00)
+    // [2019-11-03T00:30:00-06:00, 2019-11-03T01:30:00-06:00)
+    // [2019-11-03T01:30:00-06:00, 2019-11-03T01:30:00-07:00)
+    // [2019-11-03T01:30:00-07:00, 2019-11-03T02:30:00-07:00)
     // [2019-11-03T02:30:00-07:00, 2019-11-03T03:30:00-07:00)
     // [2019-11-03T03:30:00-07:00, 2019-11-03T04:30:00-07:00)
 
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -271,8 +271,6 @@ Examples:
     2009-10-15T09:00:00       // October 15th 2009 at 9 AM in the default location
     2018-01-01                // midnight on January 1st 2018 in the default location
 
-[IMPL#152](https://github.com/influxdata/flux/issues/152) Implement shorthand time literals
-
 #### String literals
 
 A string literal represents a sequence of characters enclosed in double quotes.
@@ -395,10 +393,16 @@ The `task` option is used by a scheduler to schedule the execution of a Flux que
 
 The `location` option is used to set the default time zone of all times in the script.
 The location maps the UTC offset in use at that location for a given time.
-The default value is set using the time zone of the running process.
+The default value is UTC.
+The value is specified either as the name of a time zone in the IANA Time Zone database or as a fixed offset.
+If the string value is in the format `(+|-)HH:MM` then it will be interpreted as a fixed zone offset.
 
-    option location = fixedZone(offset:-5h) // set timezone to be 5 hours west of UTC
-    option location = loadLocation(name:"America/Denver") // set location to be America/Denver
+    option location = "-05:00" // set location to a fixed offset 5 hours west of UTC
+    option location = "America/Denver" // set location to be America/Denver
+
+    2019-01-01T00:00:00       // midnight in the current location
+    2019-01-01T00:00:00Z      // midnight in utc
+    2019-01-01T00:00:00-05:00 // midnight at a fixed offset 5 hours west of UTC
 
 [IMPL#660](https://github.com/influxdata/platform/issues/660) Implement Location option
 
@@ -457,7 +461,7 @@ Note all numeric types are nullable.
 
 ##### Time types
 
-A _time type_ represents a single point in time with nanosecond precision.
+A _time type_ represents a single point in time with nanosecond precision and a fixed offset from UTC.
 The time type name is `time`.
 The time type is nullable.
 
@@ -1131,6 +1135,7 @@ December  = 12
 ### Time and date functions
 
 These are builtin functions that all take a single `time` argument and return an integer.
+The following functions consider the time offset when returning a value.
 
 * `second` int
     Second returns the second of the minute for the provided time in the range `[0-59]`.
@@ -1284,6 +1289,73 @@ Examples using known start and stop dates:
 
 [IMPL#659](https://github.com/influxdata/platform/issues/659) Implement intervals function
 
+#### Location support
+
+Intervals have location support for changing offsets by default.
+When a location has been specified either with the `location` option or the `location` parameter, a change in offset is used to either expand or shrink an interval.
+The most common example is daylight savings time switches.
+When daylight savings time starts, the zone offset shifts west, becoming a more negative number.
+When this happens, the hour that was lost does not exist and the current window is shortened to accomodate this change.
+When daylight savings time ends, the zone offset shifts east, becoming a more positive number.
+When this happens, one of the hours happens twice and so the interval becomes longer.
+
+Examples:
+
+    intervals(every:1d, location: "America/Denver")
+    // [2019-03-09T00:00:00-07:00, 2019-03-10T00:00:00-07:00)
+    // [2019-03-10T00:00:00-07:00, 2019-03-11T00:00:00-06:00)
+    // [2019-03-11T00:00:00-06:00, 2019-03-12T00:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-02T00:00:00-06:00, 2019-11-03T00:00:00-06:00)
+    // [2019-11-03T00:00:00-06:00, 2019-11-04T00:00:00-07:00)
+    // [2019-11-04T00:00:00-07:00, 2019-11-05T00:00:00-07:00)
+
+    intervals(every:2h, location: "America/Denver")
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T04:00:00-06:00)
+    // [2019-03-10T04:00:00-06:00, 2019-03-10T06:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T02:00:00-07:00, 2019-11-03T04:00:00-07:00)
+    // [2019-11-04T03:00:00-07:00, 2019-11-03T06:00:00-07:00)
+
+    intervals(every:1h, location: "America/Denver")
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T01:00:00-07:00)
+    // [2019-03-10T01:00:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T04:00:00-06:00)
+    // [2019-03-10T04:00:00-06:00, 2019-03-10T05:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T01:00:00-06:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T02:00:00-07:00, 2019-11-03T03:00:00-07:00)
+    // [2019-11-03T03:00:00-07:00, 2019-11-03T04:00:00-07:00)
+
+    intervals(every:30m, location: "America/Denver")
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T00:30:00-07:00)
+    // [2019-03-10T00:30:00-07:00, 2019-03-10T01:00:00-07:00)
+    // [2019-03-10T01:00:00-07:00, 2019-03-10T01:30:00-07:00)
+    // [2019-03-10T01:30:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T03:30:00-06:00)
+    // [2019-03-10T03:30:00-06:00, 2019-03-10T04:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T00:30:00-06:00)
+    // [2019-11-03T00:30:00-06:00, 2019-11-03T01:00:00-06:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T01:30:00-06:00), [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
+    // [2019-11-03T01:30:00-06:00, 2019-11-03T02:00:00-06:00), [2019-11-03T01:30:00-07:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T02:00:00-07:00, 2019-11-03T02:30:00-07:00)
+    // [2019-11-03T02:30:00-07:00, 2019-11-03T03:00:00-07:00)
+    
+    intervals(every:1h, offset:30m, location: "America/Denver")
+    // [2019-03-10T00:30:00-07:00, 2019-03-10T01:30:00-07:00)
+    // [2019-03-10T01:30:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T03:30:00-06:00)
+    // [2019-03-10T03:30:00-06:00, 2019-03-10T04:30:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:30:00-06:00, 2019-11-03T01:30:00-06:00), [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
+    // [2019-11-03T01:30:00-06:00, 2019-11T01:00:00-07:00), [2019-11-03T02:00:00-07:00, 2019-11-03T02:30:00-07:00)
+    // [2019-11-03T02:30:00-07:00, 2019-11-03T03:30:00-07:00)
+    // [2019-11-03T03:30:00-07:00, 2019-11-03T04:30:00-07:00)
+
 
 ### Builtin Intervals
 
@@ -1300,7 +1372,7 @@ The following builtin intervals exist:
     // 1 day intervals excluding Sundays and Saturdays
     weekdays = intervals(every:1d, filter: (interval) => weekday(time:interval.start) not in [Sunday, Saturday])
     // 1 day intervals including only Sundays and Saturdays
-    weekdends = intervals(every:1d, filter: (interval) => weekday(time:interval.start) in [Sunday, Saturday])
+    weekends = intervals(every:1d, filter: (interval) => weekday(time:interval.start) in [Sunday, Saturday])
     // 1 week intervals
     weeks = intervals(every:1w)
     // 1 month interval
@@ -1310,43 +1382,6 @@ The following builtin intervals exist:
     // 1 year intervals
     years = intervals(every:1y)
 
-
-### FixedZone
-
-FixedZone creates a location based on a fixed time offset from UTC.
-
-
-FixedZone has the following parameters:
-
-| Name   | Type     | Description                                                                                                                    |
-| ----   | ----     | -----------                                                                                                                    |
-| offset | duration | Offset is the offset from UTC for the time zone. Offset must be less than 24h. Defaults to 0, which produces the UTC location. |
-
-Examples:
-
-    fixedZone(offset:-5h) // time zone 5 hours west of UTC
-    fixedZone(offset:4h30m) // time zone 4 and a half hours east of UTC
-
-
-[IMPL#156](https://github.com/influxdata/flux/issues/156) Implement FixedZone function
-
-#### LoadLocation
-
-LoadLoacation loads a locations from a time zone database.
-
-LoadLocation has the following parameters:
-
-| Name | Type   | Description                                                                                                                  |
-| ---- | ----   | -----------                                                                                                                  |
-| name | string | Name is the name of the location to load. The names correspond to names in the [IANA tzdb](https://www.iana.org/time-zones). |
-
-Examples:
-
-    loadLocation(name:"America/Denver")
-    loadLocation(name:"America/Chicago")
-    loadLocation(name:"Africa/Tunis")
-
-[IMPL#157](https://github.com/influxdata/flux/issues/157) Implement LoadLoacation function
 
 ## Data model
 
@@ -1675,6 +1710,7 @@ AggregateWindow has the following properties:
 | timeSrc     | string                                          | TimeSrc is the name of a column from the group key to use as the source for the aggregated time. Defaults to "_stop".                                           |
 | timeDst     | string                                          | TimeDst is the name of a new column in which the aggregated time is placed. Defaults to "_time".                                                                |
 | createEmpty | bool                                            | CreateEmpty, if true, will create empty windows and fill them with a null aggregate value.  Defaults to true.                                                   |
+| location    | string                                          | An optional location that will be passed to `window`. |
 
 _NOTE_: make sure that `fn`'s parameter names match the ones specified above (see [why](#Transformations)).
 
@@ -2802,6 +2838,7 @@ Window has the following properties:
 | startColumn | string                                     | StartColumn is the name of the column containing the window start time. Defaults to `_start`.                                                                                                                                                 |
 | stopColumn  | string                                     | StopColumn is the name of the column containing the window stop time. Defaults to `_stop`.                                                                                                                                                    |
 | createEmpty | bool                                       | CreateEmpty specifies whether empty tables should be created. Defaults to `false`.
+| location    | string                                     | Location specifies a custom location to be used when creating windows. Defaults to the `location` option. |
 
 Example:
 ```
@@ -2814,6 +2851,7 @@ from(bucket:"telegraf/autogen")
 ```
 window(every:1h) // window the data into 1 hour intervals
 window(intervals: intervals(every:1d, period:8h, offset:9h)) // window the data into 8 hour intervals starting at 9AM every day.
+window(every:1d, location: "America/Los_Angeles") // use a different location than the global option.
 ```
 
 #### Pivot
@@ -4210,7 +4248,7 @@ Convert a value to a time.
 
 Example: `from(bucket: "telegraf") |> filter(fn:(r) => r._measurement == "mem" and r._field == "used") |> toTime()`
 
-The function `toTime` is defined as `toTime = (tables=<-) => tables |> map(fn:(r) => ({r with _value: time(v:r._value)}))`.
+The function `toTime` is defined as `toTime = (tables=<-, location="") => tables |> map(fn:(r) => ({r with _value: time(v:r._value,location:location)}))`.
 If you need to convert other columns use the `map` function directly with the `time` function.
 
 ##### toUInt
@@ -4222,8 +4260,6 @@ Example: `from(bucket: "telegraf") |> filter(fn:(r) => r._measurement == "mem" a
 The function `toUInt` is defined as `toUInt = (tables=<-) => tables |> map(fn:(r) => ({r with _value: uint(v:r._value)}))`.
 If you need to convert other columns use the `map` function directly with the `uint` function.
 
-
-[IMPL#242](https://github.com/influxdata/platform/issues/242) Update specification around type conversion functions.
 
 #### String operations
 


docs/SPEC.md
```diff
@@ -1284,6 +1289,73 @@ Examples using known start and stop dates:
...
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T00:30:00-07:00)
+    // [2019-03-10T00:30:00-07:00, 2019-03-10T01:00:00-07:00)
+    // [2019-03-10T01:00:00-07:00, 2019-03-10T01:30:00-07:00)
+    // [2019-03-10T01:30:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T03:30:00-06:00)
+    // [2019-03-10T03:30:00-06:00, 2019-03-10T04:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T00:30:00-06:00)
+    // [2019-11-03T00:30:00-06:00, 2019-11-03T01:00:00-06:00)
+    // [2019-11-03T01:00:00-06:00, 2019-11-03T01:30:00-06:00), [2019-11-03T01:00:00-07:00, 2019-11-03T01:30:00-07:00)
```
jsternberg: This was the weirdest thing. When I started trying to get some edge cases that tripped up influxql, I came out with this as the most logical conclusion. Since we automatically adjust a window to add or remove an hour for week long intervals, it made sense to do it in partial amounts for shorter intervals. And when the interval was shorter than the time offset, it made sense to duplicate it multiple times.

I'm not sure that this is sustainable though so we may have to discuss how to deal with this. I'm not sure how you would deal with a minute interval when it comes to daylight savings time although I guess it makes sense that each minute happens twice during that hour. I am concerned about what the proper `_start` and `_stop` values would be though for that time period.

docs/SPEC.md
```diff
@@ -395,10 +393,16 @@ The `task` option is used by a scheduler to schedule the execution of a Flux que
...
-The default value is set using the time zone of the running process.
+The default value is UTC.
+The value is specified either as the name of a time zone in the IANA Time Zone database or as a fixed offset.
+If the string value is in the format `(+|-)HH:MM` then it will be interpreted as a fixed zone offset.
+
+    option location = "-05:00" // set location to a fixed offset 5 hours west of UTC
+    option location = "America/Denver" // set location to be America/Denver
 
-    option location = fixedZone(offset:-5h) // set timezone to be 5 hours west of UTC
-    option location = loadLocation(name:"America/Denver") // set location to be America/Denver
```
nathanielc: Why the change to use a string instead of an object?

docs/SPEC.md
```diff
@@ -457,7 +461,7 @@ Note all numeric types are nullable.
 
 ##### Time types
 
-A _time type_ represents a single point in time with nanosecond precision.
+A _time type_ represents a single point in time with nanosecond precision and a fixed offset from UTC.
```
nathanielc: How does time + duration work in the context of a fixed offset? If you add 24h to a time with a fixed offset the day before day light savings you will not correctly honor the time zone since the offset is fixed.

docs/SPEC.md
jsternberg: On a positive note, they do appear to be distinct time ranges so as long as it is ok for window to create overlapping ranges, which to my knowledge it is, we can use the min start range as the start and the max end range as the end. We will just have to update the bounds object to understand that start and stop aren't the only things that are important.

docs/SPEC.md
```diff
@@ -1284,6 +1289,73 @@ Examples using known start and stop dates:
...
+    // [2019-11-04T00:00:00-07:00, 2019-11-05T00:00:00-07:00)
+
+    intervals(every:2h, location: "America/Denver")
+    // [2019-03-10T00:00:00-07:00, 2019-03-10T03:00:00-06:00)
+    // [2019-03-10T03:00:00-06:00, 2019-03-10T04:00:00-06:00)
+    // [2019-03-10T04:00:00-06:00, 2019-03-10T06:00:00-06:00)
+    // ... skipped ...
+    // [2019-11-03T00:00:00-06:00, 2019-11-03T02:00:00-07:00)
+    // [2019-11-03T02:00:00-07:00, 2019-11-03T04:00:00-07:00)
+    // [2019-11-03T03:00:00-07:00, 2019-11-03T06:00:00-07:00)
```
nathanielc: Why does this one start at 3 AM -07:00 only one hour after the 2AM -07:00?

docs/SPEC.md
nathanielc: UTC is monotonic, meaning it always increases. Given a time in nanoseconds since the epoch plus an offset you can always map that back to a UTC time. 

Conversely you cannot always map a UTC time to a timestamp with an offset since there are multiple time/offset combinations that represent the same instant in time. 

Using UTC to handle windowing removes the ambiguity of _duplicate_ windows because each window has distinct start and stop times.

So if we can map a time with its location to a UTC time we can both determine the _start and _stop boundaries of a window as well as determine which times belong in which windows.

Mapping a single time with its location to a UTC time is straightforward.

The only remaining question is how to add a duration to a time with a location when the offset changes over that duration.

I don't follow how that is defined with these examples:

For the `1d` example it seems we change the duration to be longer or shorter such that the time of day remains constant. `2019-03-10T00:00:00-07:00 + 1d =  2019-03-11T00:00:00-06:00` in `America/Denver`

For the `1h` example is seems that we ignore the change in offset. `2019-03-10T01:00:00-07:00 + 1h = 2019-03-10T03:00:00-06:00`  in `America/Denver`.

Can you clarify the behavior of adding a duration to a time with a location where the duration spans a period where the offset changes?

docs/SPEC.md
jsternberg: I don't think we can represent it as an object. We cannot include the `time.Location` object which is really the only important part. So I opted for a string out of the available types.

docs/SPEC.md
jsternberg: I was thinking it wouldn't. I looked at the time encodings and I didn't see anything that encoded time zones and I think the reason is that time zones describe the behavior of how time moves rather than an instant in time. At least that was my thought. We could potentially make it so native addition always uses the option and add a special function for customizing the location. I'll add that.

docs/SPEC.md
jsternberg: It's a mistake. I copied a pasted intervals and it was difficult to sort through them.

docs/SPEC.md
jsternberg: My goal was that if the location is set, adding a duration isn't the same as "adding a group of nanoseconds". Which meant that if I added an hour to a time, my next time should be `hour + 1`. I based this off of the idea that if I group by a week, the week should always start on the same day. In order to make that true, then we need a definition where the extra hour from going off of DST is part of that week rather than ignored and going on DST removes that hour.

> For the 1h example is seems that we ignore the change in offset. `2019-03-10T01:00:00-07:00 + 1h = 2019-03-10T03:00:00-06:00` in `America/Denver`.

I am not sure what you mean here. If you add one hour to 1 AM before DST starts, you get 2 AM. But 2 AM when DST starts is actually 3 AM because, at exactly 2 AM, the hour springs forward. That's why it lists 3 AM as the end time for this interval. Another way to write this would be that the range is `[2019-03-10T01:00:00-07:00, 2019-03-10T01:59:59.999999999-07:00]`.

jsternberg:
An alternative idea, and this is the one influxql uses, is that the offset fixes only apply if the grouping is greater than the time of the offset switch. So if the offset switch is 1 hour, then grouping by an hour would result in 23 or 25 intervals on the day of DST, but if you are grouping by 2 hours, you would always get 12 intervals.

I remember that this idea did have some weird edge cases with offsets though that were difficult, but maybe not as difficult as the currently proposed solution. The other strange thing is that this essentially makes it so there is different behavior depending on how much of an offset there is which may be awkward. I'm going to try writing out what I mean though and adding it as an additional section. Then we can look at the intervals that are created and see what looks more appropriate.

docs/SPEC.md
affo: I don't know if I am oversimplifying things or if I am missing the point. If so, completely ignore this comment 😅 .

@nathanielc 

> The only remaining question is how to add a duration to a time with a location when the offset changes over that duration.

I would say that a time is always related to a location (if not specified is the local one), so, if you add a duration to a time, you get a time in that precise location (again).

Say you have:

`2019-03-10T00:00:00 @ America/Denver`

This resolves to 

`2019-03-10T00:00:00-XXX`

As far as I  got it, `-XXX` depends on both on the time and the location and can be extracted from a global db.  
For example, `-XXX` is `-07:00` because it's the tenth of March and we are in Denver.

Anyways, now you resolved `@America/Denver` to `-07:00`.

And you can get the UTC time out of it:

`2019-03-10T00:00:00-07:00 = 2019-03-09T17:00:00Z`

So, now you can add 1 day:

`2019-03-09T17:00:00Z + 1d = 2019-03-10T17:00:00Z`

And you can come back to Denver that would now resolve to `-06:00` thanks to the database:

`2019-03-10T23:00:00-06:00`

Is the computation correct in your opinion? 🤔 

By the end of the day we have: 

`2019-03-10T00:00:00 @ America/Denver + 1d = 2019-03-10T23:00:00 @ America/Denver`

If you suppose that the timezone changes exactly at midnight, then you'll obtain that:

`2019-03-10T00:00:00 @ America/Denver + 1d == 2019-03-10T00:00:00 @ America/Denver + 1d1h`

Because, in Denver, it's twice `2019-03-10T23:00:00`. But that seems coherent with what happens in reality 🤔 

Another thing I wanted to say, is that I think that here we can employ a new operator and use `@` as suggested by @nathanielc in a previous PR. That seems idiomatic and has the pro that it can be applied to any time without adding parameters to every function that works with time.

jsternberg: > And you can come back to Denver that would now resolve to -06:00 thanks to the database:
>
> 2019-03-10T23:00:00-06:00
>
> Is the computation correct in your opinion? 🤔

I believe this behavior is incorrect. After you resolve back to Denver, you note that the offset changed and so you need to now adjust the unix timestamp as if the start time had been in the -6 offset instead of -7. This is to keep it so that adding a day always results in midnight being the time.

About the operator, I was thinking that we would remove the ability to use an operator to add times. I was talking about this with @jlapacik at some point and it might not be possible for us to allow different types on the left and right side of the parameter. If this is the case and adding a time isn't too common to begin with, I think we can add functions for performing these operations to the `date` package. I think it would probably be easier than adding another operator token.

nathanielc: >This is to keep it so that adding a day always results in midnight being the time.

I think this makes sense for durations of a day, but durations less than a day it starts to become ambiguous.  

For example if we apply this same reasoning to hours you state this:

> Which meant that if I added an hour to a time, my next time should be hour + 1.

Makes sense, if I add an hour to a date time I should get a date time that has its `hour` unit incremented by one. 

But then its not really possible to do that as some hours of the day simply do not exist

> If you add one hour to 1 AM before DST starts, you get 2 AM. But 2 AM when DST starts is actually 3 AM because, at exactly 2 AM, the hour springs forward. That's why it lists 3 AM as the end time for this interval.

So if I add 1 hour to `1 AM` I would expect to get `2 AM` but I can't because `2 AM` doesn't exist in that location. 

As a result is seems like it is impossible to preserve the invariant that `hour + 1` is your hour later according to the adjusted timezone.

I think that preserving midnight for day durations is a reasonable invariant to keep because you can always add a `day` unit to a date and have a next day that makes sense. The duration may not always be exactly 24hours but there is always a date to represent the new value.


In light of both of these I think we want to do two things to make this much easier to reason about:

1. Reintroduce `day` as its own base unit. Define day units as not strictly 24h but always the next day at the same time of day after adjusting for changes in offsets.
2. Strictly define the nanoseconds unit as adding that number of nanoseconds always regardless of offset changes.

This way we have three base units defined as follows:

1. Month, adding months always increases the month number accordingly. If a date would go past the end of the month then the last day of the month is used.
2. Day, adding days always increases the days number accordingly and the time of day will always be the same as the starting date including when the offset changes.
3. Nanosecond, adding nanoseconds always move the date time forward in time by that exact number of nanoseconds regardless of timezone changes.


The point here is we make it easy for users to reason about how time/duration addition works. 

Thoughts?

jsternberg: I guess I am concerned about offset changes that are not related to daylight savings time which is why I had favored keeping the definition away from defining a particular offset shift. And I wonder what would happen if you wanted to create a window interval for every 8 hours to signify morning, afternoon, and night. This may be common if you were trying to organize data around [work shift hours](https://www.ziprecruiter.com/blog/what-are-shift-work-hours/). It might be possible to do this by creating 3 intervals and using day increases to keep it aligned, but I'm not sure that's very user friendly and I'd like us to figure out the performance implications of binding ourselves to that method.

So I don't really think defining day as a part of the vector really helps except for one specific category of problem with one specific interval that may end up being US-centric because not every location deals with time zones in the hour forward/hour back at the specific time of 2 AM.

It might be useful if we created a research issue to look at time zones and see if there's any prior art for this.

> So if I add 1 hour to 1 AM I would expect to get 2 AM but I can't because 2 AM doesn't exist in that location.
>
> As a result is seems like it is impossible to preserve the invariant that hour + 1 is your hour later according to the adjusted timezone.

I mentioned it before but the original idea that InfluxQL was supposed to implement, but I messed up with some edge cases, was that this invariant would only get preserved if the shift was less than the interval. So if you had an hour shift from daylight savings time, then grouping by an hour or less wouldn't cause it to worry about daylight savings time. If you had an interval larger than that, we would automatically add/remove time values from the interval to accommodate. This would solve the problem of adding a day and it would also allow Flux to be used for creating intervals for people's working hours. It may match more with how daylight savings time is practiced in real life where if you are working a night shift when the clock shifts, you work either an hour more or less during that night.

nathanielc: > It might be useful if we created a research issue to look at time zones and see if there's any prior art for this.

I have been looking at existing systems this is what I have found:

* Go - treats durations a constant lengths of time regardless of timezone
* Python - same as Go
* Excel/Lbre Calc - Most operations work with only dates or times but not both. Operations that work for both dates and times are not timezone aware (or at least I didn't find any).
* R - Has lots of libraries that do things differently. The only library I found that has support for timezones is this one https://lubridate.tidyverse.org/ Its worth a more complete discussion.

The Lubridate library in R is inspired by this Java lib https://www.joda.org/. Lubridate has three different types for working with durations of time, from their site:

* durations, which measure the exact amount of time between two points
 * periods, which accurately track clock times despite leap years, leap seconds, and day light savings time
 * intervals, a protean summary of the time information between two points

This cheat sheet (second page) is particularly insightful https://rawgit.com/rstudio/cheatsheets/master/lubridate.pdf

In short I have been arguing for making Flux durations behave like the Lubridate durations. It seems like @jsternberg has been arguing for Flux durations to behave like Lubridate periods.

Perhaps we should just create two different duration types? 

FWIW I still think my proposal above will work for most cases as it is still possible to describe periods of time even in the face of time zone shifts since you can always anchor a start time at midnight for the time zone and use offsets from there.

jsternberg: Yes, I think periods are more like what we want. Although I would say your proposal doesn't behave like durations because I think your proposal is that it will behave like durations if you don't use days and it will behave like a period if you use a day. That's the part that seems to be weird for me.

I think the Lubridate library is a perfect model. Although I guess the reason why I think we should operate like periods in all circumstances is that, if you want durations, you can just use UTC and avoid time zones entirely. If you want a specific offset, use an offset. If you're setting a time zone, I think that the user has indicated they want to track clock times. I think this part of the document says it best:

> R recognizes ~600 time zones. Each encodes the time zone, Daylight
Savings Time, and historical calendar variations for an area. R assigns
one time zone per vector.
>
> Use the UTC time zone to avoid Daylight Savings.

I am curious what this library does if you add 1 minute repeatedly at a DST switch while in a time zone. It seems like that would be the solution to my weird intervals.

nathanielc:
@jsternberg I tried out the lubridate R lib this is what I found:

(Also starting new comment so its easier to find)

First a DST test of Lubridate periods
```R
# Create time right before a DST shift
> time <- force_tz(ymd_hms("2019-11-03 00:30:00"), "America/Denver")
> time
[1] "2019-11-03 00:30:00 MDT"
# Create a 1m period
> m <- minutes(1)
# Add a 1m period to the time repeatedly
> for (val in 1:125) { print(time); time <- time + m; }
[1] "2019-11-03 00:30:00 MDT"
[1] "2019-11-03 00:31:00 MDT"
[1] "2019-11-03 00:32:00 MDT"
[1] "2019-11-03 00:33:00 MDT"
[1] "2019-11-03 00:34:00 MDT"
[1] "2019-11-03 00:35:00 MDT"
[1] "2019-11-03 00:36:00 MDT"
[1] "2019-11-03 00:37:00 MDT"
[1] "2019-11-03 00:38:00 MDT"
[1] "2019-11-03 00:39:00 MDT"
[1] "2019-11-03 00:40:00 MDT"
[1] "2019-11-03 00:41:00 MDT"
[1] "2019-11-03 00:42:00 MDT"
[1] "2019-11-03 00:43:00 MDT"
[1] "2019-11-03 00:44:00 MDT"
[1] "2019-11-03 00:45:00 MDT"
[1] "2019-11-03 00:46:00 MDT"
[1] "2019-11-03 00:47:00 MDT"
[1] "2019-11-03 00:48:00 MDT"
[1] "2019-11-03 00:49:00 MDT"
[1] "2019-11-03 00:50:00 MDT"
[1] "2019-11-03 00:51:00 MDT"
[1] "2019-11-03 00:52:00 MDT"
[1] "2019-11-03 00:53:00 MDT"
[1] "2019-11-03 00:54:00 MDT"
[1] "2019-11-03 00:55:00 MDT"
[1] "2019-11-03 00:56:00 MDT"
[1] "2019-11-03 00:57:00 MDT"
[1] "2019-11-03 00:58:00 MDT"
[1] "2019-11-03 00:59:00 MDT"
[1] "2019-11-03 01:00:00 MST"
[1] "2019-11-03 01:01:00 MST"
[1] "2019-11-03 01:02:00 MST"
[1] "2019-11-03 01:03:00 MST"
[1] "2019-11-03 01:04:00 MST"
[1] "2019-11-03 01:05:00 MST"
[1] "2019-11-03 01:06:00 MST"
[1] "2019-11-03 01:07:00 MST"
[1] "2019-11-03 01:08:00 MST"
[1] "2019-11-03 01:09:00 MST"
[1] "2019-11-03 01:10:00 MST"
[1] "2019-11-03 01:11:00 MST"
[1] "2019-11-03 01:12:00 MST"
[1] "2019-11-03 01:13:00 MST"
[1] "2019-11-03 01:14:00 MST"
[1] "2019-11-03 01:15:00 MST"
[1] "2019-11-03 01:16:00 MST"
[1] "2019-11-03 01:17:00 MST"
[1] "2019-11-03 01:18:00 MST"
[1] "2019-11-03 01:19:00 MST"
[1] "2019-11-03 01:20:00 MST"
[1] "2019-11-03 01:21:00 MST"
[1] "2019-11-03 01:22:00 MST"
[1] "2019-11-03 01:23:00 MST"
[1] "2019-11-03 01:24:00 MST"
[1] "2019-11-03 01:25:00 MST"
[1] "2019-11-03 01:26:00 MST"
[1] "2019-11-03 01:27:00 MST"
[1] "2019-11-03 01:28:00 MST"
[1] "2019-11-03 01:29:00 MST"
[1] "2019-11-03 01:30:00 MST"
[1] "2019-11-03 01:31:00 MST"
[1] "2019-11-03 01:32:00 MST"
[1] "2019-11-03 01:33:00 MST"
[1] "2019-11-03 01:34:00 MST"
[1] "2019-11-03 01:35:00 MST"
[1] "2019-11-03 01:36:00 MST"
[1] "2019-11-03 01:37:00 MST"
[1] "2019-11-03 01:38:00 MST"
[1] "2019-11-03 01:39:00 MST"
[1] "2019-11-03 01:40:00 MST"
[1] "2019-11-03 01:41:00 MST"
[1] "2019-11-03 01:42:00 MST"
[1] "2019-11-03 01:43:00 MST"
[1] "2019-11-03 01:44:00 MST"
[1] "2019-11-03 01:45:00 MST"
[1] "2019-11-03 01:46:00 MST"
[1] "2019-11-03 01:47:00 MST"
[1] "2019-11-03 01:48:00 MST"
[1] "2019-11-03 01:49:00 MST"
[1] "2019-11-03 01:50:00 MST"
[1] "2019-11-03 01:51:00 MST"
[1] "2019-11-03 01:52:00 MST"
[1] "2019-11-03 01:53:00 MST"
[1] "2019-11-03 01:54:00 MST"
[1] "2019-11-03 01:55:00 MST"
[1] "2019-11-03 01:56:00 MST"
[1] "2019-11-03 01:57:00 MST"
[1] "2019-11-03 01:58:00 MST"
[1] "2019-11-03 01:59:00 MST"
[1] "2019-11-03 02:00:00 MST"
[1] "2019-11-03 02:01:00 MST"
[1] "2019-11-03 02:02:00 MST"
[1] "2019-11-03 02:03:00 MST"
[1] "2019-11-03 02:04:00 MST"
[1] "2019-11-03 02:05:00 MST"
[1] "2019-11-03 02:06:00 MST"
[1] "2019-11-03 02:07:00 MST"
[1] "2019-11-03 02:08:00 MST"
[1] "2019-11-03 02:09:00 MST"
[1] "2019-11-03 02:10:00 MST"
[1] "2019-11-03 02:11:00 MST"
[1] "2019-11-03 02:12:00 MST"
[1] "2019-11-03 02:13:00 MST"
[1] "2019-11-03 02:14:00 MST"
[1] "2019-11-03 02:15:00 MST"
[1] "2019-11-03 02:16:00 MST"
[1] "2019-11-03 02:17:00 MST"
[1] "2019-11-03 02:18:00 MST"
[1] "2019-11-03 02:19:00 MST"
[1] "2019-11-03 02:20:00 MST"
[1] "2019-11-03 02:21:00 MST"
[1] "2019-11-03 02:22:00 MST"
[1] "2019-11-03 02:23:00 MST"
[1] "2019-11-03 02:24:00 MST"
[1] "2019-11-03 02:25:00 MST"
[1] "2019-11-03 02:26:00 MST"
[1] "2019-11-03 02:27:00 MST"
[1] "2019-11-03 02:28:00 MST"
[1] "2019-11-03 02:29:00 MST"
[1] "2019-11-03 02:30:00 MST"
[1] "2019-11-03 02:31:00 MST"
[1] "2019-11-03 02:32:00 MST"
[1] "2019-11-03 02:33:00 MST"
[1] "2019-11-03 02:34:00 MST"
> 
```

Compare to a Lubridate duration

```
> time <- force_tz(ymd_hms("2019-11-03 00:30:00"), "America/Denver")
> time
[1] "2019-11-03 00:30:00 MDT"
> dm <- dminutes(1)
> for (val in 1:125) { print(time); time <- time + dm; }
[1] "2019-11-03 00:30:00 MDT"
[1] "2019-11-03 00:31:00 MDT"
[1] "2019-11-03 00:32:00 MDT"
[1] "2019-11-03 00:33:00 MDT"
[1] "2019-11-03 00:34:00 MDT"
[1] "2019-11-03 00:35:00 MDT"
[1] "2019-11-03 00:36:00 MDT"
[1] "2019-11-03 00:37:00 MDT"
[1] "2019-11-03 00:38:00 MDT"
[1] "2019-11-03 00:39:00 MDT"
[1] "2019-11-03 00:40:00 MDT"
[1] "2019-11-03 00:41:00 MDT"
[1] "2019-11-03 00:42:00 MDT"
[1] "2019-11-03 00:43:00 MDT"
[1] "2019-11-03 00:44:00 MDT"
[1] "2019-11-03 00:45:00 MDT"
[1] "2019-11-03 00:46:00 MDT"
[1] "2019-11-03 00:47:00 MDT"
[1] "2019-11-03 00:48:00 MDT"
[1] "2019-11-03 00:49:00 MDT"
[1] "2019-11-03 00:50:00 MDT"
[1] "2019-11-03 00:51:00 MDT"
[1] "2019-11-03 00:52:00 MDT"
[1] "2019-11-03 00:53:00 MDT"
[1] "2019-11-03 00:54:00 MDT"
[1] "2019-11-03 00:55:00 MDT"
[1] "2019-11-03 00:56:00 MDT"
[1] "2019-11-03 00:57:00 MDT"
[1] "2019-11-03 00:58:00 MDT"
[1] "2019-11-03 00:59:00 MDT"
[1] "2019-11-03 01:00:00 MDT"
[1] "2019-11-03 01:01:00 MDT"
[1] "2019-11-03 01:02:00 MDT"
[1] "2019-11-03 01:03:00 MDT"
[1] "2019-11-03 01:04:00 MDT"
[1] "2019-11-03 01:05:00 MDT"
[1] "2019-11-03 01:06:00 MDT"
[1] "2019-11-03 01:07:00 MDT"
[1] "2019-11-03 01:08:00 MDT"
[1] "2019-11-03 01:09:00 MDT"
[1] "2019-11-03 01:10:00 MDT"
[1] "2019-11-03 01:11:00 MDT"
[1] "2019-11-03 01:12:00 MDT"
[1] "2019-11-03 01:13:00 MDT"
[1] "2019-11-03 01:14:00 MDT"
[1] "2019-11-03 01:15:00 MDT"
[1] "2019-11-03 01:16:00 MDT"
[1] "2019-11-03 01:17:00 MDT"
[1] "2019-11-03 01:18:00 MDT"
[1] "2019-11-03 01:19:00 MDT"
[1] "2019-11-03 01:20:00 MDT"
[1] "2019-11-03 01:21:00 MDT"
[1] "2019-11-03 01:22:00 MDT"
[1] "2019-11-03 01:23:00 MDT"
[1] "2019-11-03 01:24:00 MDT"
[1] "2019-11-03 01:25:00 MDT"
[1] "2019-11-03 01:26:00 MDT"
[1] "2019-11-03 01:27:00 MDT"
[1] "2019-11-03 01:28:00 MDT"
[1] "2019-11-03 01:29:00 MDT"
[1] "2019-11-03 01:30:00 MDT"
[1] "2019-11-03 01:31:00 MDT"
[1] "2019-11-03 01:32:00 MDT"
[1] "2019-11-03 01:33:00 MDT"
[1] "2019-11-03 01:34:00 MDT"
[1] "2019-11-03 01:35:00 MDT"
[1] "2019-11-03 01:36:00 MDT"
[1] "2019-11-03 01:37:00 MDT"
[1] "2019-11-03 01:38:00 MDT"
[1] "2019-11-03 01:39:00 MDT"
[1] "2019-11-03 01:40:00 MDT"
[1] "2019-11-03 01:41:00 MDT"
[1] "2019-11-03 01:42:00 MDT"
[1] "2019-11-03 01:43:00 MDT"
[1] "2019-11-03 01:44:00 MDT"
[1] "2019-11-03 01:45:00 MDT"
[1] "2019-11-03 01:46:00 MDT"
[1] "2019-11-03 01:47:00 MDT"
[1] "2019-11-03 01:48:00 MDT"
[1] "2019-11-03 01:49:00 MDT"
[1] "2019-11-03 01:50:00 MDT"
[1] "2019-11-03 01:51:00 MDT"
[1] "2019-11-03 01:52:00 MDT"
[1] "2019-11-03 01:53:00 MDT"
[1] "2019-11-03 01:54:00 MDT"
[1] "2019-11-03 01:55:00 MDT"
[1] "2019-11-03 01:56:00 MDT"
[1] "2019-11-03 01:57:00 MDT"
[1] "2019-11-03 01:58:00 MDT"
[1] "2019-11-03 01:59:00 MDT"
[1] "2019-11-03 01:00:00 MST"
[1] "2019-11-03 01:01:00 MST"
[1] "2019-11-03 01:02:00 MST"
[1] "2019-11-03 01:03:00 MST"
[1] "2019-11-03 01:04:00 MST"
[1] "2019-11-03 01:05:00 MST"
[1] "2019-11-03 01:06:00 MST"
[1] "2019-11-03 01:07:00 MST"
[1] "2019-11-03 01:08:00 MST"
[1] "2019-11-03 01:09:00 MST"
[1] "2019-11-03 01:10:00 MST"
[1] "2019-11-03 01:11:00 MST"
[1] "2019-11-03 01:12:00 MST"
[1] "2019-11-03 01:13:00 MST"
[1] "2019-11-03 01:14:00 MST"
[1] "2019-11-03 01:15:00 MST"
[1] "2019-11-03 01:16:00 MST"
[1] "2019-11-03 01:17:00 MST"
[1] "2019-11-03 01:18:00 MST"
[1] "2019-11-03 01:19:00 MST"
[1] "2019-11-03 01:20:00 MST"
[1] "2019-11-03 01:21:00 MST"
[1] "2019-11-03 01:22:00 MST"
[1] "2019-11-03 01:23:00 MST"
[1] "2019-11-03 01:24:00 MST"
[1] "2019-11-03 01:25:00 MST"
[1] "2019-11-03 01:26:00 MST"
[1] "2019-11-03 01:27:00 MST"
[1] "2019-11-03 01:28:00 MST"
[1] "2019-11-03 01:29:00 MST"
[1] "2019-11-03 01:30:00 MST"
[1] "2019-11-03 01:31:00 MST"
[1] "2019-11-03 01:32:00 MST"
[1] "2019-11-03 01:33:00 MST"
[1] "2019-11-03 01:34:00 MST"
> 
```

Compare to when the DST shift is backwards 

First using Lubridate period

```
> time <- force_tz(ymd_hms("2020-03-08 00:30:00"), "America/Denver")
> time
[1] "2020-03-08 00:30:00 MST"
> m <- minutes(1)
> for (val in 1:125) { print(time); time <- time + m; }
[1] "2020-03-08 00:30:00 MST"
[1] "2020-03-08 00:31:00 MST"
[1] "2020-03-08 00:32:00 MST"
[1] "2020-03-08 00:33:00 MST"
[1] "2020-03-08 00:34:00 MST"
[1] "2020-03-08 00:35:00 MST"
[1] "2020-03-08 00:36:00 MST"
[1] "2020-03-08 00:37:00 MST"
[1] "2020-03-08 00:38:00 MST"
[1] "2020-03-08 00:39:00 MST"
[1] "2020-03-08 00:40:00 MST"
[1] "2020-03-08 00:41:00 MST"
[1] "2020-03-08 00:42:00 MST"
[1] "2020-03-08 00:43:00 MST"
[1] "2020-03-08 00:44:00 MST"
[1] "2020-03-08 00:45:00 MST"
[1] "2020-03-08 00:46:00 MST"
[1] "2020-03-08 00:47:00 MST"
[1] "2020-03-08 00:48:00 MST"
[1] "2020-03-08 00:49:00 MST"
[1] "2020-03-08 00:50:00 MST"
[1] "2020-03-08 00:51:00 MST"
[1] "2020-03-08 00:52:00 MST"
[1] "2020-03-08 00:53:00 MST"
[1] "2020-03-08 00:54:00 MST"
[1] "2020-03-08 00:55:00 MST"
[1] "2020-03-08 00:56:00 MST"
[1] "2020-03-08 00:57:00 MST"
[1] "2020-03-08 00:58:00 MST"
[1] "2020-03-08 00:59:00 MST"
[1] "2020-03-08 01:00:00 MST"
[1] "2020-03-08 01:01:00 MST"
[1] "2020-03-08 01:02:00 MST"
[1] "2020-03-08 01:03:00 MST"
[1] "2020-03-08 01:04:00 MST"
[1] "2020-03-08 01:05:00 MST"
[1] "2020-03-08 01:06:00 MST"
[1] "2020-03-08 01:07:00 MST"
[1] "2020-03-08 01:08:00 MST"
[1] "2020-03-08 01:09:00 MST"
[1] "2020-03-08 01:10:00 MST"
[1] "2020-03-08 01:11:00 MST"
[1] "2020-03-08 01:12:00 MST"
[1] "2020-03-08 01:13:00 MST"
[1] "2020-03-08 01:14:00 MST"
[1] "2020-03-08 01:15:00 MST"
[1] "2020-03-08 01:16:00 MST"
[1] "2020-03-08 01:17:00 MST"
[1] "2020-03-08 01:18:00 MST"
[1] "2020-03-08 01:19:00 MST"
[1] "2020-03-08 01:20:00 MST"
[1] "2020-03-08 01:21:00 MST"
[1] "2020-03-08 01:22:00 MST"
[1] "2020-03-08 01:23:00 MST"
[1] "2020-03-08 01:24:00 MST"
[1] "2020-03-08 01:25:00 MST"
[1] "2020-03-08 01:26:00 MST"
[1] "2020-03-08 01:27:00 MST"
[1] "2020-03-08 01:28:00 MST"
[1] "2020-03-08 01:29:00 MST"
[1] "2020-03-08 01:30:00 MST"
[1] "2020-03-08 01:31:00 MST"
[1] "2020-03-08 01:32:00 MST"
[1] "2020-03-08 01:33:00 MST"
[1] "2020-03-08 01:34:00 MST"
[1] "2020-03-08 01:35:00 MST"
[1] "2020-03-08 01:36:00 MST"
[1] "2020-03-08 01:37:00 MST"
[1] "2020-03-08 01:38:00 MST"
[1] "2020-03-08 01:39:00 MST"
[1] "2020-03-08 01:40:00 MST"
[1] "2020-03-08 01:41:00 MST"
[1] "2020-03-08 01:42:00 MST"
[1] "2020-03-08 01:43:00 MST"
[1] "2020-03-08 01:44:00 MST"
[1] "2020-03-08 01:45:00 MST"
[1] "2020-03-08 01:46:00 MST"
[1] "2020-03-08 01:47:00 MST"
[1] "2020-03-08 01:48:00 MST"
[1] "2020-03-08 01:49:00 MST"
[1] "2020-03-08 01:50:00 MST"
[1] "2020-03-08 01:51:00 MST"
[1] "2020-03-08 01:52:00 MST"
[1] "2020-03-08 01:53:00 MST"
[1] "2020-03-08 01:54:00 MST"
[1] "2020-03-08 01:55:00 MST"
[1] "2020-03-08 01:56:00 MST"
[1] "2020-03-08 01:57:00 MST"
[1] "2020-03-08 01:58:00 MST"
[1] "2020-03-08 01:59:00 MST"
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
[1] NA
> 
```

Now again with a lubridate duration

```
> time <- force_tz(ymd_hms("2020-03-08 00:30:00"), "America/Denver")
> dm <- dminutes(1)
> time
[1] "2020-03-08 00:30:00 MST"
> for (val in 1:125) { print(time); time <- time + dm; }
[1] "2020-03-08 00:30:00 MST"
[1] "2020-03-08 00:31:00 MST"
[1] "2020-03-08 00:32:00 MST"
[1] "2020-03-08 00:33:00 MST"
[1] "2020-03-08 00:34:00 MST"
[1] "2020-03-08 00:35:00 MST"
[1] "2020-03-08 00:36:00 MST"
[1] "2020-03-08 00:37:00 MST"
[1] "2020-03-08 00:38:00 MST"
[1] "2020-03-08 00:39:00 MST"
[1] "2020-03-08 00:40:00 MST"
[1] "2020-03-08 00:41:00 MST"
[1] "2020-03-08 00:42:00 MST"
[1] "2020-03-08 00:43:00 MST"
[1] "2020-03-08 00:44:00 MST"
[1] "2020-03-08 00:45:00 MST"
[1] "2020-03-08 00:46:00 MST"
[1] "2020-03-08 00:47:00 MST"
[1] "2020-03-08 00:48:00 MST"
[1] "2020-03-08 00:49:00 MST"
[1] "2020-03-08 00:50:00 MST"
[1] "2020-03-08 00:51:00 MST"
[1] "2020-03-08 00:52:00 MST"
[1] "2020-03-08 00:53:00 MST"
[1] "2020-03-08 00:54:00 MST"
[1] "2020-03-08 00:55:00 MST"
[1] "2020-03-08 00:56:00 MST"
[1] "2020-03-08 00:57:00 MST"
[1] "2020-03-08 00:58:00 MST"
[1] "2020-03-08 00:59:00 MST"
[1] "2020-03-08 01:00:00 MST"
[1] "2020-03-08 01:01:00 MST"
[1] "2020-03-08 01:02:00 MST"
[1] "2020-03-08 01:03:00 MST"
[1] "2020-03-08 01:04:00 MST"
[1] "2020-03-08 01:05:00 MST"
[1] "2020-03-08 01:06:00 MST"
[1] "2020-03-08 01:07:00 MST"
[1] "2020-03-08 01:08:00 MST"
[1] "2020-03-08 01:09:00 MST"
[1] "2020-03-08 01:10:00 MST"
[1] "2020-03-08 01:11:00 MST"
[1] "2020-03-08 01:12:00 MST"
[1] "2020-03-08 01:13:00 MST"
[1] "2020-03-08 01:14:00 MST"
[1] "2020-03-08 01:15:00 MST"
[1] "2020-03-08 01:16:00 MST"
[1] "2020-03-08 01:17:00 MST"
[1] "2020-03-08 01:18:00 MST"
[1] "2020-03-08 01:19:00 MST"
[1] "2020-03-08 01:20:00 MST"
[1] "2020-03-08 01:21:00 MST"
[1] "2020-03-08 01:22:00 MST"
[1] "2020-03-08 01:23:00 MST"
[1] "2020-03-08 01:24:00 MST"
[1] "2020-03-08 01:25:00 MST"
[1] "2020-03-08 01:26:00 MST"
[1] "2020-03-08 01:27:00 MST"
[1] "2020-03-08 01:28:00 MST"
[1] "2020-03-08 01:29:00 MST"
[1] "2020-03-08 01:30:00 MST"
[1] "2020-03-08 01:31:00 MST"
[1] "2020-03-08 01:32:00 MST"
[1] "2020-03-08 01:33:00 MST"
[1] "2020-03-08 01:34:00 MST"
[1] "2020-03-08 01:35:00 MST"
[1] "2020-03-08 01:36:00 MST"
[1] "2020-03-08 01:37:00 MST"
[1] "2020-03-08 01:38:00 MST"
[1] "2020-03-08 01:39:00 MST"
[1] "2020-03-08 01:40:00 MST"
[1] "2020-03-08 01:41:00 MST"
[1] "2020-03-08 01:42:00 MST"
[1] "2020-03-08 01:43:00 MST"
[1] "2020-03-08 01:44:00 MST"
[1] "2020-03-08 01:45:00 MST"
[1] "2020-03-08 01:46:00 MST"
[1] "2020-03-08 01:47:00 MST"
[1] "2020-03-08 01:48:00 MST"
[1] "2020-03-08 01:49:00 MST"
[1] "2020-03-08 01:50:00 MST"
[1] "2020-03-08 01:51:00 MST"
[1] "2020-03-08 01:52:00 MST"
[1] "2020-03-08 01:53:00 MST"
[1] "2020-03-08 01:54:00 MST"
[1] "2020-03-08 01:55:00 MST"
[1] "2020-03-08 01:56:00 MST"
[1] "2020-03-08 01:57:00 MST"
[1] "2020-03-08 01:58:00 MST"
[1] "2020-03-08 01:59:00 MST"
[1] "2020-03-08 03:00:00 MDT"
[1] "2020-03-08 03:01:00 MDT"
[1] "2020-03-08 03:02:00 MDT"
[1] "2020-03-08 03:03:00 MDT"
[1] "2020-03-08 03:04:00 MDT"
[1] "2020-03-08 03:05:00 MDT"
[1] "2020-03-08 03:06:00 MDT"
[1] "2020-03-08 03:07:00 MDT"
[1] "2020-03-08 03:08:00 MDT"
[1] "2020-03-08 03:09:00 MDT"
[1] "2020-03-08 03:10:00 MDT"
[1] "2020-03-08 03:11:00 MDT"
[1] "2020-03-08 03:12:00 MDT"
[1] "2020-03-08 03:13:00 MDT"
[1] "2020-03-08 03:14:00 MDT"
[1] "2020-03-08 03:15:00 MDT"
[1] "2020-03-08 03:16:00 MDT"
[1] "2020-03-08 03:17:00 MDT"
[1] "2020-03-08 03:18:00 MDT"
[1] "2020-03-08 03:19:00 MDT"
[1] "2020-03-08 03:20:00 MDT"
[1] "2020-03-08 03:21:00 MDT"
[1] "2020-03-08 03:22:00 MDT"
[1] "2020-03-08 03:23:00 MDT"
[1] "2020-03-08 03:24:00 MDT"
[1] "2020-03-08 03:25:00 MDT"
[1] "2020-03-08 03:26:00 MDT"
[1] "2020-03-08 03:27:00 MDT"
[1] "2020-03-08 03:28:00 MDT"
[1] "2020-03-08 03:29:00 MDT"
[1] "2020-03-08 03:30:00 MDT"
[1] "2020-03-08 03:31:00 MDT"
[1] "2020-03-08 03:32:00 MDT"
[1] "2020-03-08 03:33:00 MDT"
[1] "2020-03-08 03:34:00 MDT"
>
```

In summary it appears the `period` just skips time when a DST shifts forward. Conversely the `period` returns what is effectively a `nil` date when a DST shifts backward. To me this behavior was surprising. I think it will be unexpected to a user that if they define a timezone that it will skip chunks of their data and return invalid dates. 

Conversely the `duration` type moves forward in time by 1m physical intervals. 

> Use the UTC time zone to avoid Daylight Savings.

I agree, additionally you could use a fixed offset timezone

jsternberg:
Ah, I think I may understand what the library is doing. Look at this:

```
[1] "2019-11-03 00:57:00 MDT"
[1] "2019-11-03 00:58:00 MDT"
[1] "2019-11-03 00:59:00 MDT"
[1] "2019-11-03 01:00:00 MST"
[1] "2019-11-03 01:01:00 MST"
[1] "2019-11-03 01:02:00 MST"
```

This didn't make sense to me at first, but I think it might. Instead of having `1:00 MDT` and then repeating `1:00 MST` an hour later, it switches the time zone an hour early and makes that one interval last an hour and a minute. This seems a bit weird, but I'm not sure it's inappropriate as a decision. And as a reminder, if they want to do intervals by the minute, they can just use the UTC location which is the default anyway. As long as the rules are consistent for these edge cases, we can likely respond with, "That's what we had to do to make it work for that time period, but using this feature with your interval doesn't make a lot of sense so don't use it."

Then the spring forward:

```
[1] "2020-03-08 01:59:00 MST"
[1] NA
```

I think this makes sense, but I agree that this isn't what we would actually do. I think in this circumstance, we would just discard non-existant intervals. So in effect, it would skip directly to 3 AM. I likely have to think about that a bit more.

jsternberg:
I tried to write up some thoughts. They're not the most coherent and definitely not complete, but it tries to summarize the different methods that we have been batting around.

Underlying goal is that it is possible to do things like aggregating by days or hours in your time zone in a consistent way that continues through a time zone shift. So if you have an interval that goes from 12 am to 1 am on every day, it will continue being there every day of the year even when the time zone changes. This is the same for a full day, 8 hour intervals, and others that aren't necessarily related to `window(every:1d)`.

The rules also need to be consistent, but it is not necessary that all possible intervals make complete sense because not all intervals in general make sense. But, the most important in my mind are that 1d, 8h, 1h, and 1m all work appropriately.

The rules need to be consistent mostly because mixing different every, period, and offset values need to have underlying math that is consistent otherwise we will quickly descend into edge cases on top of edge cases. For that reason, the following intervals should be computed and make sense across a daylight savings time boundary.

    intervals(every:1d)
    intervals(every:8h)
    intervals(every:1h)
    intervals(every:1m)
    intervals(every:1d, period:8h)
    intervals(every:1d, period:1h)
    intervals(every:1d, period:1m)
    intervals(every:8h, period:1d)
    intervals(every:8h, period:1h)
    // intervals skipped for brevity
    
All of these generated intervals should then be repeated with an offset of `30m`.

## Possible Options

### Ignore DST switches when the time change is less than or equal to the offset change.

Advantages
* Simple and would work well for the most common use cases.
* A period of `1m` wouldn't produce a weird hour and minute long interval.

Disadvantages
* Might work very weirdly if `every` and `period` aren't the same. For example, if `every` is `1m` but the period is `2h`. This may not be a disadvantage though.

### Lengthen or shorten any interval that encounters the offset switch

Advantages
* Simple and would work well for the most common use cases.
* No exceptions so the behavior doesn't change depending on what the time interval is.

Disadvantages
* A period of `1m` would produce a weird hour and minute long interval when daylight savings time ends.

### Treat day as a separate vector similar to months

Advantages
* Works for aggregating by days.

Disadvantages
* Doesn't really work with anything else.
* Only applies to daylight savings time.

### Treat each part of the duration as its own component of the vector

This would be if we also added hours as its own vector so advancing the time by 2 hours would always add 2 hours to the actual number.

Advantages
* Works for aggregation even by days.
* Works for more than daylight savings time.

Disadvantages
* Much more complicated and makes adding different vectors difficult.
* Creates some weird results such as what happens when aggregating by minutes during daylight savings time. There are two disparate intervals that both are 1:00 AM to 1:01 AM when daylight savings time ends.

My favorite is "Ignore DST switches when the time change is less than or equal to the offset change". I am going to try seeing if I can write up more detail of what that would look like.

jsternberg:
Edited the above because I forgot an additional reason why "Ignore DST switches when the time change is less than or equal to the offset change" might be better. With that method, you could look at the aggregates by day and see that one day has more requests than the others. You could then zoom in. Once you zoom in enough, the offset change is no longer factored for the window so you can see the aggregation exactly by hour while the other methods mostly always factor in that extra hour no matter what you do.

PR #2092, opened 2019-11-01
feat: type inference test macros

### LINKED ISSUE #2072: Type inference test harness
Writing a type inference test case should involve specifying two things:

1. The flux program to type check/infer.
2. The expected types of the that program.

The flux program itself should be given as a uff-8 encoded string literal and the expected types should be specified using the type expression grammar as defined as part of #1952.

The output of this issue should be a function for comparing the expected output to the actual output of type inference as well as an example for how to write tests.

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR defines macros for writing type inference test cases. Along with a few small examples.

Closes #2072 .

### DIFF
--- a/libflux/src/semantic/tests.rs
+++ b/libflux/src/semantic/tests.rs
@@ -1,0 +1,211 @@
+//! This is the main test module for type inference.
+//!
+//! This module defines two macros:
+//!
+//! * `test_infer!`
+//! * `test_infer_err!`
+//!
+//! `test_infer!` types a flux program and asserts that the inferred types
+//! are equivalent to what was expected.
+//!
+//! `test_infer_err!` asserts that a flux program does not type check.
+//!
+//! Note macros are a great tool for emulating table driven tests. With
+//! macros we can define a single pattern that each test case follows. We
+//! may also define certain inputs as optional. This allows us the ability
+//! to introduce new test inputs while maintaining backwards compatibility
+//! with previously specified tests.
+//!
+//! With macros we can also provide custom test output that is much easier
+//! to inspect in the case of a failed assertion than it is with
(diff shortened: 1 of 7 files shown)


libflux/src/semantic/tests.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+macro_rules! test_infer {
+    (
+        $name:ident,
+        $(env: $env:expr,)?
+        src: $src:expr,
+        exp: $exp:expr,
+    ) => {
+        #[test]
+        fn $name() {
```
jsternberg: So just a quick comment, but can we have it so `test_infer` is called from within a function? I'm not too keen on the macro generating the function itself. The entirety of the body seems fine. It's just I'd like:

```
fn test_blah_blah() {
    test_infer!()
}
```

or whatever the proper syntax would be.

libflux/src/semantic/tests.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+/// # Example
+///
+/// ```
+/// test_infer_err! { undeclared variable,
+///     src: "x = f",
+/// }
+/// ```
+///
+macro_rules! test_infer_err {
```
nathanielc: Do we want to include anything about the expected error?

jlapacik: In the future yes. For right now I think it's a little premature until we formalize the structure of type errors. But we should (and we definitely will) have enough test coverage to ensure we don't have any false positives.

nathanielc: Great, I am thinking something like this would be a good project soon https://github.com/elm/error-message-catalog/issues

Basically proof that error messages make sense in the context of their script.

PR #2094, opened 2019-11-01
feat(stdlib/http): Add an http.get flux method (returns statusCode and body)

nisley:
Open this PR in favor of closing https://github.com/influxdata/flux/pull/2081

I could probably use some advice on the test as I don't know much about writing them - the ones I have are basically copies of http.post but changed for gets.  The getWithBody has a different return type and I don't know if that means there needs to be significant change.

Following on from the last PR - the http.get that mirrors the http.post and only returns statusCode is useful for Ping type situations when one isn't interested in the body.  I considered many things including redoing http.post, but that may have unintended consequences for alerts and notifications so I thought it best to simply mirror http.post with get and add getWithBody. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


nisley:
> @nisley I like were this is going, a couple of thoughts:
> 
> 1. Can you add these functions as part of the `experimental/http` package? That way we can iterate on their API a bit before committing to not breaking them?
> 2. I would prefer to see a single `http.get` function with an optional `body` parameter instead of two different functions.

Can you tell me or show me an example of how to have flux return Int when no body parameter is passed in and return String when something is passed (my return at this point is ["200", "bodystring"] - even better would be [Int, "bodytring"] - but at any rate, I got lost trying to figure out how to return different semantic types based on parameter values.

nisley:
I think I found an example for in lgamma that returns different semantic types in one return, but I don't believe the return signature ends up exactly matching post.
```
"lgamma": values.NewFunction(
			"lgamma",
			semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
				Parameters: map[string]semantic.PolyType{"x": semantic.Float},
				Required:   semantic.LabelSet{"x"},
				Return:     semantic.NewObjectPolyType(map[string]semantic.PolyType{"lgamma": semantic.Float, "sign": semantic.Int}, semantic.LabelSet{"lgamma", "sign"}, nil),
			}),
```

nisley:
As for experimental/http.get() - I'd really like to keep both get and post in the same file, but when I build and try to run the flux repl with the below it complains that it is missing builtin.
Should I be able to do this?:
func init() {
	flux.RegisterPackageValue("experimental/http", "get", get)
	flux.RegisterPackageValue("http", "post", post)
}

nathanielc:
There is probably a way to make that work, but I would rather see the code in separate files for now until we are ready to promote the code to stable status.

nisley:
Ok, I think this is set with the requested changes.

stdlib/experimental/http/http_experimental.go
```diff
@@ -0,0 +1,156 @@
...
+var get = values.NewFunction(
+	"get",
+	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"url":          semantic.String,
+			"headers":      semantic.Tvar(1),
+			"responseType": semantic.String,
+		},
+		Required: []string{"url"},
+		Return:   semantic.NewObjectPolyType(map[string]semantic.PolyType{"statusCode": semantic.Int, "body": semantic.String, "response": semantic.String}, semantic.LabelSet{"status"}, semantic.LabelSet{"status", "body", "res"}),
```
nathanielc: I would like to see the return type here simplified to just the `statusCode` and the `body` properties.

The `statusCode` as an int makes sense the `body` should be a `semantic.Bytes`. 

Additionally these should always be returned, therefore we can remove the `responseType` parameter.

nisley: What about headers? I still see them as a valuable option.

nathanielc: Do you need the headers right now? If not then I want to wait because we don't have a good way to represent the headers.

stdlib/experimental/http/http.flux
```diff
@@ -0,0 +1,5 @@
+package http
+
+// Get submits an HTTP get request to the specified URL with headers and different returns based on responseType
+// At a minimum, HTTP status code is returned. BODY and ALL (which includes the response headers) are also options
+builtin get
```
nathanielc: This comment is out of date after some of the changes we have made. Can you update it?

stdlib/experimental/http/http_experimental.go
```diff
@@ -0,0 +1,120 @@
...
+// http get mirrors the http post originally completed for alerts & notifications
+var get = values.NewFunction(
+	"get",
+	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"url":     semantic.String,
+			"headers": semantic.Tvar(1),
+		},
+		Required: []string{"url"},
+		Return:   semantic.NewObjectPolyType(map[string]semantic.PolyType{"statusCode": semantic.Int, "body": semantic.Bytes, "headers": semantic.String}, semantic.LabelSet{"status", "body"}, nil),
```
nathanielc: Can you remove `headers` from the return type here as we are not setting it below anymore?

stdlib/http/http.flux
```diff
@@ -6,6 +6,7 @@ import "experimental"
 // The HTTP status code is returned.
 builtin post
 
+
```
nathanielc: nit,  can you undo this change?

stdlib/http/http.go
```diff
@@ -0,0 +1,128 @@
...
+		// return status code
+		return values.NewInt(int64(statusCode)), nil
+	},
+	true, // post has side-effects
+)
+
+func init() {
+	flux.RegisterPackageValue("http", "post", post)
+
+}
```
nathanielc: Can you rename the `http.go` function back to `post.go` so its clear nothing has changed with the http.post bits?

stdlib/experimental/http/http_experimental.go
nisley: headers in this case is a parameter into the function -  it can be passed in by the user to be built into the request payload

stdlib/http/http.flux
nisley: Nothing is changed in this file?  As far as I can tell it's the original?

stdlib/experimental/http/http.flux
nisley: done

stdlib/http/http.go
nisley: done

stdlib/http/post.go
```diff
@@ -16,111 +11,118 @@ import (
...
-				return nil, err
-			}
+func init() {
+	flux.RegisterPackageValue("http", "post", post)
 
-			// return status code
-			return values.NewInt(int64(statusCode)), nil
-		},
-		true, // post has side-effects
-	))
```
nathanielc: can you revert the changes to this file as they are not related to this PR anymore?

nisley: I'm sure there's an easier way to revert then copy/paste from master, but I've reverted to the original and tested again.

nisley:
alchemy please merge

stdlib/experimental/http/get_test.go
```diff
@@ -0,0 +1,94 @@
...
+			w.Write([]byte(err.Error()))
+		}
+		w.WriteHeader(204)
+	}))
+	defer ts.Close()
+
+	script := fmt.Sprintf(`
+import "experimental/http"
+
+status = http.get(url:"%s/path/a/b/c", headers: {x:"a",y:"b",z:"c"})
```
pauldix: Would be better if this returned a response object that had `status` and `headers` and `body`. Also, would be nice to be able to pass in `urls` that can be fetched in parallel by Go. The return for that would be an array of responses, where the position in the array is the same as the position in the passed in the urls.

pauldix: Also would be good to have a `timeout` parameter that is a duration. So you could have `5s` or `125ms` or whatever.

pauldix: Oh, would also be nice to have `time` as a duration that can ideally be converted into an int in milliseconds or whatever.

pauldix: Actually, now that I think of it, it would probably be better to keep `Get` consistent for a single url. But add another function named `GetAll` that takes an array of urls and returns an array of responses.

PR #2098, opened 2019-11-05
fix: update make release to make sure local and remote are in sync 

fchikwekwe:
There was an issue with `make release` being able to cut a release without local changes being up to date and in sync with remote. 

These changes check the current commit against the latest commit on the remote version of the branch. They also check that all the tags on both local and remote are the same. If there are inconsistencies, then the script will error. 

**For reviewers:** I was not sure how to test this file without accidentally cutting a bunch of releases. If you have suggestions on how I can test these changes, please let me know. Otherwise, please keep that in mind as you review. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -5,6 +5,27 @@ cd $DIR
 
 set -e
 
+remote=$(git rev-parse "@{u}") # "@{u}" gets the current upstream branch
+local=$(git rev-parse @) # '@' gets the current local branch 
+
+# check if local commit syncs with remote 
+if [ "$remote" != "$local" ]; then
+    echo "Error: local commit does not match remote. Exiting release script."
+    exit 1
+fi 
+
+# remove any excess brackets, space/tab characters and 'origin' branch and sort the tags
+git_remote_tags () { git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]] ; }
+git_local_tags () { git show-ref --tags | grep -v '{}' | grep -v 'origin'| sort | tr -d [[:blank:]] ; }
+
+# check if local tags are different from remote tags
+if ! diff -q <(git_remote_tags) <(git_local_tags) &>/dev/null; then
+    echo "Error: local tags do not match remote. Exiting release script."
+    exit 1
+fi
+
+# cut the next Flux release
 version=$(go run github.com/influxdata/changelog nextver)
-git tag -s -m "Release $version" $version
+git tag -s -m "Release $version" "$version"
 git push origin "$version"
+


release.sh
```diff
@@ -5,6 +5,38 @@ cd $DIR
 
 set -e
 
+git remote update
+
+upstream=${1:-'@{u}'}
+remote=$(git rev-parse "$upstream")
+local=$(git rev-parse @)
```
affo: could you put some comments on the usage of `@` and `@{u}`?

release.sh
```diff
@@ -5,6 +5,38 @@ cd $DIR
...
+fi 
+
+remoteTags=$(mktemp)
+git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]] >> $remoteTags
+
+localTags=$(mktemp)
+git show-ref --tags | grep -v '{}' | grep -v 'origin' | sort | tr -d [[:blank:]] >> $localTags
+
+# check if local tags match remote tags
+match=$(diff "$remoteTags" "$localTags")
```
affo: This is great! This would check that every tag is on the exact same commit both locally and remotely. Thanks

release.sh
```diff
@@ -5,6 +5,38 @@ cd $DIR
...
+    then
+        echo "Error: local commit does not match remote. Exiting release script." 
+        exit 1
+fi 
+
+remoteTags=$(mktemp)
+git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]] >> $remoteTags
+
+localTags=$(mktemp)
+git show-ref --tags | grep -v '{}' | grep -v 'origin' | sort | tr -d [[:blank:]] >> $localTags
```
affo: I think you want `grep -v "$upstream"`, right?

Could you also comment on what are doing here?

release.sh
```diff
@@ -5,6 +5,38 @@ cd $DIR
...
+
+# check if local commit syncs with remote 
+if [ "$remote" != "$local" ]
+    then
+        echo "Error: local commit does not match remote. Exiting release script." 
+        exit 1
+fi 
+
+remoteTags=$(mktemp)
+git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]] >> $remoteTags
```
affo: Can't you just store the result in a local variable? 🤔

affo: This shouldn't be necessary actually. I think `mktemp` is a good solution so that you can exploit `diff`

release.sh
fchikwekwe: Yeah, this definitely needs a comment.

release.sh
fchikwekwe: I made this a little bit more concise, but yes I'll comment whats going on here.

release.sh
fchikwekwe: Thanks @jsternberg. I was looking for a way to avoid creating files, but I wasn't sure how to work around it. This solution works well for my purposes.

jsternberg: Just because I found this very interesting, here's the reason why I used `printf` (although given the second example, you can completely avoid `printf` in this scenario because you can avoid variables altogether).

For demonstrating the above: https://stackoverflow.com/a/49519466
For why you shouldn't use echo: https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819

release.sh
```diff
@@ -3,8 +3,31 @@
...
+if ! diff -s --left-column <(printf "%s" "$remoteTags") <(printf "%s" "$localTags") >/dev/null 2>&1
+    then 
+        echo "Error: local tags do not match remote. Exiting release script."
+        exit 1
+fi
+
+# cut the next Flux release
 version=$(go run github.com/influxdata/changelog nextver)
-git tag -s -m "Release $version" $version
+git tag -s -m "Release "$version"" "$version"
```
nathanielc: What is this change for? It looks like it might pass three parameters instead of two to the call which would break it.

fchikwekwe: Oh, that's my bad. I was trying to make sure that all my variables were wrapped in double quotes. I can change that back.

fchikwekwe: Also, do we need the `set -e` line? It looks like it is supposed to quit out of the script if any command has a non-zero exit status. I added some error messages that might be helpful to the developer when they are working on debugging release issues. If it is necessary, I can add it back, but then I may as well remove the error messages and exit statuses as well.

fchikwekwe: @nathanielc ^^

nathanielc: I think we still want the `set -e` line as its provides some needed safety that commands are working as expected. The error messages will still be triggered when reached.

release.sh
```diff
@@ -3,8 +3,33 @@
...
+    then 
+        echo "Error: local tags do not match remote. Exiting release script."
+        exit 1
+fi
+
+# cut the next Flux release
 version=$(go run github.com/influxdata/changelog nextver)
-git tag -s -m "Release $version" $version
+git tag -s -m "Release $version" "$version"
 git push origin "$version"
```
nathanielc: One way to test this is to change these lines to `echo ....` and then run the script knowing that it will not do anything except echo these last commands. 

Also you can clone the repo into a temp directory, then clone from that directory into yet another directory. Then the first becomes the upstream of the second. 

Something like:

```
/tmp/flux-test/
      origin/ git clone from github
      working/ git clone from ./origin
```

Then you can add commits and tags to the repo in `origin` and then run the script in `working` to see if it works. All of this can remain local without having to touch github.

release.sh
fchikwekwe: Ahh, yeah I was wondering what the difference between `echo` and `printf` was in this case, besides `printf` being more C-like and having more functionality. Thanks for sharing.

release.sh
fchikwekwe: Yeah I've been commenting out the final lines and printing the comment on like 31 to test the rest of script for now. That just feels like a potentially brittle way of ensuring the code works. Jonathan offered to help me test it on a cloned repo sometime this afternoon.

release.sh
```diff
@@ -5,6 +5,31 @@ cd $DIR
...
+        printf "%s" "Error: local commit does not match remote. Exiting release script." $'\n'
+        exit 1
+fi 
+
+# remove any excess brackets, space/tab characters and 'origin' branch, then sort the tags
+remoteTags=$(git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]])
+localTags=$(git show-ref --tags | grep -v '{}' | grep -v 'origin'| sort | tr -d [[:blank:]])
+
+# check if sorted local tags match remote tags
+if ! diff -q <(printf "%s" "$remoteTags") <(printf "%s" "$localTags") >/dev/null 2>&1
```
jsternberg: Can you change the above to:

```
git_remote_tags() {
    git ls-remote ...
}
git_local_tags() {
    git show-ref ...
}
if !diff -q <(git_remote_tags) <(git_local_tags) &> /dev/null; then
    ...
fi
```

I think it makes it easier to read and then there doesn't have to be an explanation for why `printf`. I also prefer `&> /dev/null` because it means the same thing as what you have, but it's more explicit in my opinion.
```

File: release.sh [hunk 8]
```diff
@@ -5,6 +5,31 @@ cd $DIR
...
+    then 
+        printf "%s" "Error: local tags do not match remote. Exiting release script." $'\n'
```
jsternberg: You can just use `echo` for this. The `printf` is mostly just for untrusted input.

release.sh
wolffcm: Would it be better to just grep for tags of the form `v[0-9]+\.[0-9]+\.[0-9]+` here, and exclude all the others?

release.sh
fchikwekwe: Succesfully tested! Thanks for the suggestion @nathanielc!

release.sh
fchikwekwe: @wolffcm, perhaps. Let me try it and see what happens

release.sh
```diff
@@ -5,6 +5,38 @@ cd $DIR
...
+remoteTags=$(mktemp)
+git ls-remote --tags origin | grep -v '{}' | sort | tr -d [[:blank:]] >> $remoteTags
+
+localTags=$(mktemp)
+git show-ref --tags | grep -v '{}' | grep -v 'origin' | sort | tr -d [[:blank:]] >> $localTags
+
+# check if local tags match remote tags
+match=$(diff "$remoteTags" "$localTags")
+
+if [ ! -z $match ]
```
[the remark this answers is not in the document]
affo: @jsternberg this is the main reason why I look into Google for everything when I write bash. Thanks for the explanation :)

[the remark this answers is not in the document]
affo: I guess you can also use double brackets right ? `if [[ ... ]]`

PR #2102, opened 2019-11-06
feat(libflux): added semantic expression constraints to libflux

### LINKED ISSUE #1956: Generate type constraints for semantic expressions and blocks
~~Each semantic expression node as well as block nodes should implement the trait defined in #1954.~~

#2061 defines unimplemented `infer` methods on expression nodes that are used for generating type constraints. This issue is for implementing those `infer` methods. Note constraints will also need to be generated for function blocks, however #2061 provides an implementation. That implementation will be used when generating the constraints for function expressions.

Note #1958 is required for generating constraints for semantic function blocks.

jsternberg:
Semantic expression constraints have been added for most expression
types in libflux. The only ones missing are member expressions, object
expressions, and function/call expressions.

This also slightly modifies the ragel scanner so it does not emit
warnings anymore for unused variables.

Fixes #1956.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/semantic/nodes.rs
+++ b/libflux/src/semantic/nodes.rs
@@ -17,9 +17,10 @@ use crate::semantic::{
     fresh::Fresher,
     infer::{Constraint, Constraints},
     sub::{Substitutable, Substitution},
-    types::{Kind, MonoType, PolyType, Tvar},
+    types::{Array, Kind, MonoType, PolyType, Tvar},
 };
 
+use crate::semantic::types::Kind::*;
 use chrono::prelude::DateTime;
 use chrono::Duration;
 use chrono::FixedOffset;
@@ -60,6 +61,16 @@ impl Error {
     }
     fn invalid_statement(msg: String) -> Error {
         Error { msg }
+    }
+    fn unsupported_binary_operator(op: &ast::Operator) -> Error {
+        Error {
+            msg: format!("unsupported binary operator {}", op.to_string()),
+        }
+    }
+    fn unsupported_unary_operator(op: &ast::Operator) -> Error {
+        Error {
+            msg: format!("unsupported unary operator {}", op.to_string()),
+        }
     }
 }
 
@@ -218,7 +229,7 @@ pub struct File {
 }
 
 impl File {
-    fn infer(&mut self, mut env: Environment, f: &mut Fresher) -> Result {
+    fn infer(&mut self, env: Environment, f: &mut Fresher) -> Result {
         // TODO: add imported types to the type environment
         self.body.iter_mut().try_fold(
             (env, Constraints::empty()),
@@ -431,7 +442,21 @@ pub struct StringExpr {
 
 impl StringExpr {
     fn infer(&mut self, env: Environment, f: &mut Fresher) -> Result {
-        unimplemented!();
+        let mut env = env;
+        let mut constraints = Vec::new();
+        for p in &mut self.parts {
+            if let StringExprPart::Interpolated(ref mut ip) = p {
+                let (e, cons) = ip.expression.infer(env, f)?;
+                constraints.append(&mut Vec::from(cons));
+                constraints.push(Constraint::Equal(
+                    ip.expression.type_of().clone(),
+                    MonoType::String,
+                ));
+                env = e
+            }
+        }
+        constraints.push(Constraint::Equal(self.typ.clone(), MonoType::String));
+        return Ok((env, Constraints::from(constraints)));
     }
 }
 
@@ -466,8 +491,18 @@ pub struct ArrayExpr {
 }
 
 impl ArrayExpr {
-    fn infer(&self, env: Environment, f: &mut Fresher) -> Result {
-        unimplemented!();
+    fn infer(&mut self, mut env: Environment, f: &mut Fresher) -> Result {
+        let mut cons = Vec::new();
+        let elt = MonoType::Var(f.fresh());
+        for el in &mut self.elements {
+            let (e, c) = el.infer(env, f)?;
+            cons.append(&mut c.into());
+            cons.push(Constraint::Equal(el.type_of().clone(), elt.clone()));
+            env = e;
+        }
+        let at = MonoType::Arr(Box::new(Array(elt)));
+        cons.push(Constraint::Equal(at, self.typ.clone()));
+        return Ok((env, cons.into()));
     }
 }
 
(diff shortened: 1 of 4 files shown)


libflux/src/semantic/tests.rs
```diff
@@ -204,6 +204,636 @@ fn instantiation_1() {
...
+            ("f", "forall [] regexp"),
+        ],
+    }
+}
+#[test]
+fn instantiate_string_expr() {
+    let src = r#"message = "Hello, ${name}!""#;
+    test_infer! {
+        env: &[
+            ("name", "forall [t0] t0"),
```
jlapacik: We don't have types like this, at least not yet. Specifically we don't have value polymorphism, until we introduce polymorphic literals. So I would prefer we didn't include this specific test case as it's not really testing what we want to be testing.

I think what we want is something like this:
```Rust
test_infer! {
    env: &[
        ("name", "forall [] t0"),
    ],
    src: src,
    exp: &[
        ("name", "forall [] string"),
        ("message", "forall [] string"),
    ],
}
```
Note this test macro only prints out new additions to the environment. We'll need to update this macro to specify a "before" type environment, and an "after" type environment in order to support  this test case.

libflux/src/semantic/tests.rs
```diff
@@ -204,6 +204,636 @@ fn instantiation_1() {
...
+        exp: &[
+            ("message", "forall [] string"),
+        ],
+    }
+    test_infer_err! {
+        env: &[
+            ("name", "forall [] int"),
+        ],
+        src: src,
+    }
```
jlapacik: I like this test case. Can you add more for the other basic non-string datatypes?

libflux/src/semantic/tests.rs
```diff
@@ -204,6 +204,636 @@ fn instantiation_1() {
...
+            a = ["a", "b", "c"]
+        "#,
+        exp: &[
+            ("a", "forall [] [string]"),
+        ],
+    }
+    test_infer! {
+        env: &[
+            ("a", "forall [t0] t0"),
+            ("b", "forall [t1] t1"),
```
jlapacik: As mentioned above, these aren't valid Flux types so I would prefer not to use them in test cases. For this particular test case perhaps you should test that the empty array type is `forall [] [t0]`?

Again, I think the way you want to write this test is like:
```Rust
test_infer! {
    before: &[
        ("a", "forall [] t0"),
        ("b", "forall [] t1"),
    ],
    src: r#"
        c = [a, b]
    "#,
    after: &[
        ("a", "forall [] t2"),
        ("b", "forall [] t2"),
        ("c", "forall [] [t2]"),
    ],
}
```
This is a deficiency of the testing framework and I will address it.

libflux/src/semantic/nodes.rs
```diff
@@ -542,8 +572,61 @@ pub struct BinaryExpr {
...
+        } {
+            cons.push(Constraint::Kind(self.typ.clone(), kind));
+        }
+
+        cons.append(&mut Vec::from(match self.operator {
+            // The following operators require both sides to be equal.
+            ast::Operator::AdditionOperator
+            | ast::Operator::SubtractionOperator
+            | ast::Operator::MultiplicationOperator
+            | ast::Operator::DivisionOperator
```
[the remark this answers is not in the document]
jsternberg: Including multiplication?

libflux/src/semantic/tests.rs
jsternberg: Yes, this was something I was asking about. One thing that would be nice is if I could put anything in the expected array to see how it changed.

libflux/src/semantic/nodes.rs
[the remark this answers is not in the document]
jlapacik: Yes, I could be assuming wrong, but any type that can be an operand of `*` can also be an operand of `/`

jlapacik:
@jsternberg I have thought it over and I don't think there should be tests that express polytypes of the form

    forall [t0] t0

or

    forall [] t0

These polytypes cannot be expressed in pure flux. The latter polytype is assigned to a function's parameters and added to the type environment before the function body is inferred. However that is an implementation detail of the algorithm itself. You cannot write a flux expression that has that type.

In order to test that type variables are instantiated correctly, I think we should use function expressions as that is the only context in which the type `forall [] t0` will ever exist.

jsternberg:
Ok. I'll work on inferring function expressions then.

libflux/src/semantic/tests.rs
jlapacik: As mentioned [here](https://github.com/influxdata/flux/pull/2102#issuecomment-551848574), after giving it some consideration, I think we should test cases like these indirectly via function expressions.

libflux/src/semantic/tests.rs
jlapacik: This comment is not relevant anymore because of [this comment](https://github.com/influxdata/flux/pull/2102#issuecomment-551848574)

libflux/src/semantic/nodes.rs
```diff
@@ -542,8 +572,61 @@ pub struct BinaryExpr {
...
+            | ast::Operator::PowerOperator
+            | ast::Operator::ModuloOperator => Constraints::from(vec![
```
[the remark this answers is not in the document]
jlapacik: What types are valid operands for the `PowerOperator` and `ModuloOperator`?

libflux/src/semantic/nodes.rs
```diff
@@ -573,8 +603,72 @@ pub struct BinaryExpr {
...
+            | ast::Operator::EqualOperator => {
+                let kind = match self.operator {
+                    ast::Operator::EqualOperator
+                    | ast::Operator::NotEqualOperator => Equatable,
+                    _ => Comparable,
+                };
+                Constraints::from(vec![
+                    Constraint::Kind(self.left.type_of().clone(), kind),
+                    Constraint::Kind(self.right.type_of().clone(), kind),
+                    Constraint::Equal(self.typ.clone(), MonoType::Bool),
```
jlapacik: You're missing a constraint here. Both sides should be constrained to be equal.

libflux/src/semantic/nodes.rs
```diff
@@ -573,8 +603,72 @@ pub struct BinaryExpr {
...
+                ])
+            }
+            _ => Constraints::empty(),
+        }));
+
+        // If no constraints were added, then this is an
+        // unsupported binary operator.
+        if cons.is_empty() {
+            return Err(Error::unsupported_binary_operator(&self.operator));
+        }
```
jlapacik: I would like to see this error thrown on invalid match arms in the expression above.

libflux/src/semantic/nodes.rs
```diff
@@ -664,8 +781,19 @@ pub struct UnaryExpr {
...
+        let (env, acons) = self.argument.infer(env, f)?;
+        let cons = match self.operator {
+            ast::Operator::NotOperator => Constraints::from(vec![
+                Constraint::Equal(self.argument.type_of().clone(), MonoType::Bool),
+                Constraint::Equal(self.typ.clone(), MonoType::Bool),
+            ]),
+            ast::Operator::ExistsOperator => {
+                Constraints::from(Constraint::Equal(self.typ.clone(), MonoType::Bool))
+            }
+            _ => Constraints::empty(),
```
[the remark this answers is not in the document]
jlapacik: Unary operators will need the same kind constraints that binary operators do. Assuming an operator functions as both unary and binary.

libflux/src/semantic/tests.rs
```diff
@@ -230,6 +230,748 @@ fn instantiation_1() {
...
+                "a" => "forall [] float",
+                "b" => "forall [] int",
+            ],
+            src: &src,
+            exp: map![
+                // TODO(jsternberg): This should be an error
+                // when polymorphic literals are implemented.
+                "c" => "forall [] bool",
+            ],
+        }
```
jlapacik: This would pass if we had polymorphic literals. It should fail now though.

libflux/src/semantic/nodes.rs
jsternberg: This is not one of the constraints we have in the Go code. It requires polymorphic literals for this to be the case.

libflux/src/semantic/tests.rs
jsternberg: This needs to be tested. It's in the Go code as something that is allowed.

jsternberg: Also, this should not pass with polymorphic literals. These are explicit float and integer types. This isn't an example of two literals being compared.

jlapacik: This is correct, this specific test case would not pass with polymorphic literals. What I meant to say was that you may compare an integer literal and a float literal with polymorphic literals. However without them, the type system should not allow it.

I realize this is currently in the Go code, but keep in mind there are many inconsistencies in the current type system code. This is one of the areas where the Go code is not correct. This type system is considered a breaking change for this very reason and is why we need to define how flux versioning works before promoting it. So it is entirely safe and reasonable to implement these constraints correctly. We do not have to maintain parity with the Go code.

libflux/src/semantic/nodes.rs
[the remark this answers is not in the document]
jsternberg: Just noting for the future this might not be the case. Or at the minimum, we will likely want to revisit this because this same principle would probably apply to addition and subtraction too.

libflux/src/semantic/nodes.rs
jsternberg: I'll refactor the match a bit to bring this into the match instead. This is meant to handle the case when `Constraints::empty()` gets returned above, but it can be made more explicit.

libflux/src/semantic/tests.rs
jsternberg: Ok. I'll change it. But I think that means we're going to need to implement polymorphic literals before this gets promoted to be used as the primary type system. I don't think we can promote a type system that doesn't allow the behavior that's meant to be allowed here. It would be safer to allow this and then constrain it later once polymorphic literals exist because then there are fewer prerequisites to integrating this into the Go code.

PR #2104, opened 2019-11-07
fix(ast): flux AST FlatBuffers definition, generated code, and test

yzhang1991:
This is the pre-requisite for landing the AST FlatBuffers Go deserializer.

### DIFF
--- a/ast/flatbuffers_test.go
+++ b/ast/flatbuffers_test.go
@@ -33,9 +33,9 @@ func TestFlatBuffers(t *testing.T) {
 	stmt := fbast.ExpressionStatementEnd(b)
 
 	fbast.WrappedStatementStart(b)
-	fbast.WrappedExpressionAddExprType(b, fbast.StatementExpressionStatement)
-	fbast.WrappedExpressionAddExpr(b, stmt)
-	wrappedStmt := fbast.WrappedExpressionEnd(b)
+	fbast.WrappedStatementAddStatementType(b, fbast.StatementExpressionStatement)
+	fbast.WrappedStatementAddStatement(b, stmt)
+	wrappedStmt := fbast.WrappedStatementEnd(b)
 
 	fbast.FileStartBodyVector(b, 1)
 	b.PrependUOffsetT(wrappedStmt)


yzhang1991:
Hi @wolffcm ,

I think several changes are needed before my Go deserializer work lands, so I made them a separate PR and want to get some inputs from you first. I recommend that you go through this PR commit by commit so the changes may be clearer to you.

For the first commit, I just ran `go generate ./ast` in the flux repo, then I observed code changes in a lot of files. The `flatc` version I ran is:
```
flatc version 1.11.0
```

The second commit is self-explanatory.

And for the last commit, this is your FlatBuffers definition for `CallExpression`:
```
table CallExpression {
  base_node:BaseNode;
  callee:Expression;
  arguments:ObjectExpression;
}
```
The `arguments` is an `ObjectExpressionType`, while in **ast/mod.rs, L481**:
```rust
// CallExpr represents a function call
#[serde(rename = "CallExpression", tag = "type")]
pub struct CallExpr {
    #[serde(skip_serializing_if = "BaseNode::is_empty")]
    #[serde(default)]
    #[serde(flatten)]
    pub base: BaseNode,
    pub callee: Expression,
    pub arguments: Vec<Expression>,
}
```
`arguments` is `Vec<Expression>` type.
Moreover, in **ast/ast.go, L637**:
```go
// CallExpression represents a function call
type CallExpression struct {
	BaseNode
	Callee    Expression   `json:"callee"`
	Arguments []Expression `json:"arguments,omitempty"`
}
```
`Arguments` is `[]Expression` type.
I am wondering if this was a mistake, that we need to make the FlatBuffers definition for `CallExpression.Arguments` a `[WrappedExpression]` type. However, in https://github.com/influxdata/flux/blob/master/libflux/src/ast/flatbuffers/tests.rs#L614, I see you got `&ast_args[0]` for comparison in the test. That made me think you may have some special thinking in making it `ObjectExpression` rather than `[WrappedExpression]` in FlatBuffers, so I want to ask you about that.

yzhang1991:
The CI failure resulted from the FlatBuffers definition change for the `CallExpression`. Once my question in the comment above is addressed then I can easily fix that.

PR #2105, opened 2019-11-07
refactor(libflux): remove warnings from libflux and set ci to use strict mode

jsternberg:
This adds a strict mode to compiling libflux that will turn any warning
into an error. This means that the developer will either have to
explicitly handle the warning in code by using `#[allow(warning_name)]`
or they will have to follow the text of the warning to fix it.

This adds certain `#[allow]` statements into the code because we are
still filling in implementation.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/ast/mod.rs
+++ b/libflux/src/ast/mod.rs
@@ -1,4 +1,7 @@
 pub mod check;
+
+// Disable warnings because the flatbuffers module is generated code.
+#[allow(warnings)]
 pub mod flatbuffers;
 pub mod walk;
 
--- a/libflux/src/lib.rs
+++ b/libflux/src/lib.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "strict", deny(warnings))]
+
 extern crate chrono;
 #[macro_use]
 extern crate serde_derive;
@@ -14,6 +16,7 @@ use std::os::raw::{c_char, c_void};
 
 use parser::Parser;
 
+#[allow(non_camel_case_types)]
 pub mod ctypes {
     include!(concat!(env!("OUT_DIR"), "/ctypes.rs"));
 }
--- a/libflux/src/semantic/mod.rs
+++ b/libflux/src/semantic/mod.rs
@@ -1,10 +1,21 @@
 mod analyze;
+pub use analyze::analyze;
+
 mod env;
 mod fresh;
 mod infer;
+
+// TODO(jsternberg): Once more work is done on the infer methods,
+// this should be removed and the warnings fixed.
+#[allow(warnings)]
 pub mod nodes;
-mod parser;
+
 mod sub;
-mod tests;
 mod types;
 pub mod walk;
+
+#[cfg(test)]
+mod parser;
+
+#[cfg(test)]
+mod tests;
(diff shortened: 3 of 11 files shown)


libflux/src/ast/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod check;
+#[allow(warnings)]
```
nathanielc: can you add a comment here why we are allowing warnings? I assume its because it is generated code?

libflux/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "strict", deny(warnings))]
```
nathanielc: what does this do?

libflux/src/semantic/mod.rs
```diff
@@ -1,10 +1,21 @@
 mod analyze;
+pub use analyze::analyze;
+
 mod env;
 mod fresh;
 mod infer;
+
+// TODO(jsternberg): Once more work is done on the infer methods,
+// this should be removed and the warnings fixed.
+#[allow(warnings)]
```
nathanielc: Can you create an issue to clean up these warnings?

jlapacik:
Can the changes to semantic/tests.rs be done in a separate PR?

libflux/src/lib.rs
jsternberg: This converts warnings into errors if the strict feature is enabled. This prevents us from enabling warnings for all builds and allows us to only have this enabled for CI. The Makefile also uses this mode by default.

libflux/src/ast/mod.rs
jsternberg: Sure.

libflux/src/semantic/mod.rs
jsternberg: Sure.

jsternberg: https://github.com/influxdata/flux/issues/2106

jsternberg:
@jlapacik the changes for `semantic/tests.rs` are only for dealing with warnings right now. It's still the same code, but if an environment wasn't specified, it got upset that the mutable hashmap was never used. This just fixes that. The refactor I talked to you about is already separated.

libflux/src/semantic/tests.rs
```diff
@@ -74,19 +73,18 @@ fn parse_program(src: &str) -> ast::Package {
 ///
 macro_rules! test_infer {
     (
-        $(env: $env:expr,)?
+        env: $env:expr,
```
jlapacik: This should be optional. Can you just move `m` inside the optional block?

jsternberg: It is optional. You can specify multiple patterns. This is the pattern with the implementation. I had to do it this way because there didn't seem to be a way to say "if this pattern isn't included, do this instead." So instead, there's one pattern which includes it and one pattern that doesn't. The pattern that doesn't constructs an empty hashmap and passes that in.

libflux/src/ast/mod.rs
wolffcm: Not all of the flatbuffers module is generated, only its submodule `ast_generated`, so this could be tightened up.

PR #2108, opened 2019-11-07
refactor(libflux): refactor the test_infer macros for more complete type checking

jsternberg:
The `test_infer` macros have been changed to call a common `infer_types`
function that is in the test module. This should make it easier for
editors to analyze the function and help with readability.

The syntax for specifying type variables has changed to use a map macro.
You can now specify things like this:

    map!{
        "var1" => "forall [] t0",
        "var2" => "forall [] t1",
    }

This should be clearer to read.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/semantic/tests.rs
+++ b/libflux/src/semantic/tests.rs
@@ -63,6 +63,60 @@ fn validate(t: PolyType) -> Result<PolyType, String> {
     return Ok(t);
 }
 
+fn parse_env_map(env: HashMap<&str, &str>) -> HashMap<String, PolyType> {
+    env.into_iter()
+        .map(|(name, expr)| {
+            let init = parse(expr).unwrap();
+            let poly = validate(init).unwrap();
+            return (name.to_string(), poly);
+        })
+        .collect()
+}
+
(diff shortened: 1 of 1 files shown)


jsternberg:
Updated this to use the new test code. Now the primary thing this does is just move the main implementation into a function and the macros just wrap around it.

libflux/src/semantic/tests.rs
```diff
@@ -208,39 +185,47 @@ macro_rules! test_infer_err {
                     .fold(String::new(), |acc, (name, poly)| acc
                         + &format!("\t{}: {}\n", name, poly))
             )
-        };
+        }
     }};
     ( src: $src:expr $(,)? ) => {{
-        let env: Vec<(&str, &str)> = Vec::new();
+        let env = HashMap::new();
```
jlapacik: Can just inline this.

jsternberg: I could, but I'm already having enough difficulty with reading the long lines. Rust seems to be a more horizontal language than I am used to.

PR #2113, opened 2019-11-08
feat(rust/algw): unify function types

affo:
Ok, this is a first PR with the port of the algorithm.

There is one test that passes, but I'd like the algorithm to be reviewed and to add many more tests.

### DIFF
--- a/libflux/src/semantic/nodes.rs
+++ b/libflux/src/semantic/nodes.rs
@@ -17,7 +17,7 @@ use crate::semantic::{
     fresh::Fresher,
     infer::{Constraint, Constraints},
     sub::{Substitutable, Substitution},
-    types::{Array, Kind, MonoType, PolyType, Tvar},
+    types::{Array, Function, Kind, MonoType, PolyType, Tvar},
 };
 
 use crate::semantic::types::Kind::*;
@@ -519,8 +519,74 @@ pub struct FunctionExpr {
 }
 
 impl FunctionExpr {
-    fn infer(&mut self, env: Environment, f: &mut Fresher) -> Result {
-        unimplemented!();
+    fn infer(&mut self, mut env: Environment, f: &mut Fresher) -> Result {
+        let mut cons = Constraints::empty();
+        let mut pipe = None;
+        let mut req = HashMap::new();
+        let mut opt = HashMap::new();
+        // This params will build the nested env when inferring the function body.
+        let mut params = HashMap::new();
+        for param in &mut self.params {
+            match param.default {
+                Some(ref mut e) => {
+                    let (nenv, ncons) = e.infer(env, f)?;
+                    cons = cons + ncons;
+                    let id = param.key.name.clone();
+                    // We are here: `f = (a=1) => {...}`.
+                    // So, this PolyType is actually a MonoType, whose type
+                    // is the one of the default value ("1" in "a=1").
+                    let typ = PolyType {
+                        vars: Vec::new(),
+                        cons: HashMap::new(),
+                        expr: e.type_of().clone(),
+                    };
+                    params.insert(id.clone(), typ);
+                    opt.insert(id, e.type_of().clone());
+                    env = nenv;
+                }
+                None => {
+                    // We are here: `f = (a) => {...}`.
+                    // So, we do not know the type of "a". Let's use a fresh TVar.
+                    let id = param.key.name.clone();
+                    let ftvar = f.fresh();
+                    let typ = PolyType {
+                        vars: Vec::new(),
+                        cons: HashMap::new(),
+                        expr: MonoType::Var(ftvar),
+                    };
+                    params.insert(id.clone(), typ.clone());
+                    // Piped arguments cannot have a default value.
+                    // So check if this is a piped argument.
+                    if param.is_pipe {
+                        pipe = Some(types::Property {
+                            k: id,
+                            v: MonoType::Var(ftvar),
+                        });
+                    } else {
+                        req.insert(id, MonoType::Var(ftvar));
+                    }
+                }
+            };
+        }
+        // Add the parameters to some nested environment.
+        let mut nenv = Environment::new(env);
+        for (id, param) in params.into_iter() {
+            nenv.add(id, param);
+        }
+        // And use it to infer the body.
+        let (nenv, bcons) = self.body.infer(nenv, f)?;
+        // Now pop the nested environment, we don't need it anymore.
+        let env = nenv.pop();
+        let retn = self.body.type_of().clone();
+        let func = MonoType::Fun(Box::new(Function {
+            req,
+            opt,
+            pipe,
+            retn,
+        }));
+        cons = cons + bcons;
+        cons.add(Constraint::Equal(func, self.typ.clone()));
+        return Ok((env, cons));
     }
 
     pub fn pipe(&self) -> Option<&FunctionParameter> {
@@ -581,6 +647,16 @@ impl Block {
             Block::Variable(ass, _) => &ass.loc,
             Block::Expr(es, _) => es.expression.loc(),
             Block::Return(expr) => expr.loc(),
+        }
+    }
+    pub fn type_of(&self) -> &MonoType {
+        let mut n = self;
+        loop {
+            n = match n {
+                Block::Variable(_, b) => b.as_ref(),
+                Block::Expr(_, b) => b.as_ref(),
+                Block::Return(e) => return e.type_of(),
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


libflux/src/semantic/types.rs
```diff
@@ -842,8 +842,71 @@ impl Substitutable for Function {
...
+                } else if let Some(g_arg_type) = g.opt.remove(arg_name.as_str()) {
+                    apply_then_unify(f_arg_type, g_arg_type, sub, cons, fresh)
+                } else {
+                    Ok(sub)
+                }
+            })?;
+        // 3. Lastly unify pipe args. Note that pipe arguments are optional.
+        //    However if a pipe arg was used in a calling context, i.e it's an un-named pipe arg,
+        //    then the other type must specify a pipe arg too, otherwise unification fails (checked above).
+        if let (Some(Property { k: _, v: pf }), Some(Property { k: _, v: pg })) = (f.pipe, g.pipe) {
```
affo: should I also check the name here?

libflux/src/semantic/types.rs
```diff
@@ -842,8 +842,71 @@ impl Substitutable for Function {
...
+            return Err(Error::cannot_unify(&f, &g));
+        }
+        // Given two function types f and g, the process for unifying their arguments is as follows:
+        // 1. If a required arg of f is not present in the arguments of g, fail unification (checked above),
+        //    otherwise unify both argument types.
+        let sub: Substitution =
+            f.req
+                .into_iter()
+                .try_fold(Substitution::empty(), |sub, (arg_name, f_arg_type)| {
+                    if let Some(g_arg_type) = g.req.remove(arg_name.as_str()) {
```
affo: I am using `remove` because that's the only way to own something from an HashMap

libflux/src/semantic/types.rs
```diff
@@ -1281,4 +1345,35 @@ mod tests {
...
+            expr: MonoType::Fun(f),
+        } = fn_type
+        {
+            let sub = f.unify(call_type, &mut cons, &mut Fresher::new()).unwrap();
+            assert_eq!(
+                sub,
+                Substitution::from(maplit::hashmap! {Tvar(0) => MonoType::Int})
+            );
+            // the constraint on t0 gets removed.
+            assert_eq!(cons, maplit::hashmap! {},);
```
affo: Is this right? 🤔

libflux/src/semantic/types.rs
```diff
@@ -1281,4 +1345,163 @@ mod tests {
...
+        ) = (f, g)
+        {
+            // this extends the first map with the second by generating a new one.
+            let mut cons = f_cons.into_iter().chain(g_cons).collect();
+            let sub = f.unify(*g, &mut cons, &mut Fresher::new()).unwrap();
+            assert_eq!(
+                sub,
+                Substitution::from(maplit::hashmap! {
+                    Tvar(0) => MonoType::Int,
+                    Tvar(1) => MonoType::Float,
```
affo: Ok, this fails, `t1` doesn't get constrained to `float`

libflux/src/semantic/tests.rs
```diff
@@ -34,6 +34,7 @@ use crate::semantic::types::{MaxTvar, MonoType, PolyType};
 use crate::ast;
 use crate::parser::parse_string;
 
+#[cfg(test)]
```
affo: make this dependency a dev one

jsternberg: This isn't needed anymore.

libflux/src/semantic/types.rs
jlapacik: Note you will need to keep track of if you unified pipe arguments above as part of the required and optional arguments. If you did, then no need to unify pipe arguments here as they've already been unified above. If you didn't, then the pipe argument names must match and their types must unify.

libflux/src/semantic/types.rs
```diff
@@ -959,8 +959,71 @@ impl MaxTvar for Function {
...
+                    Ok(sub)
+                }
+            })?;
+        // 3. Lastly unify pipe args. Note that pipe arguments are optional.
+        //    However if a pipe arg was used in a calling context, i.e it's an un-named pipe arg,
+        //    then the other type must specify a pipe arg too, otherwise unification fails (checked above).
+        if let (Some(Property { k: _, v: pf }), Some(Property { k: _, v: pg })) = (f.pipe, g.pipe) {
+            sub = apply_then_unify(pf, pg, sub, cons, fresh)?;
+        }
+        Ok(sub)
```
jlapacik: Don't forget to unify the return types. I would probably just unify those right off the bat since they're the simplest.

libflux/src/semantic/types.rs
jlapacik: I think this is just because you forgot to unify the return types.

libflux/src/semantic/types.rs
```diff
@@ -959,8 +959,71 @@ impl MaxTvar for Function {
...
+        fresh: &mut Fresher,
+    ) -> Result<Substitution, Error> {
+        let f = self;
+        let mut g = other;
+        // We start with a validation phase from steps 1 and 3:
+        //  1. Verify that a required argument of f is in the required/optional arguments of g (and viceversa).
+        //  3. If a pipe arg was used then the other type must specify a pipe arg too.
+        // This is done at the beginning both to avoid starting unifying when there is no need to,
+        // and in order to error using f and g, otherwise, the borrow checker would not make us
+        // error while we have borrowed some of their internal pieces (while iterating for example).
```
jlapacik: If this is not possible and we have to clone, I'm ok with that. We can tackle performance once we have correctness.

libflux/src/semantic/nodes.rs
```diff
@@ -520,7 +520,40 @@ pub struct FunctionExpr {
...
+            if fp.is_pipe {
+                pipe = Some(types::Property {
+                    k: k.clone(),
+                    v: v.clone(),
+                });
+            }
+            m.insert(k, v);
+        }
+
+        let (env, bcons) = self.body.infer(env, f)?;
```
jlapacik: The environment you pass to the `infer` function here must contain the function parameters.

libflux/src/semantic/mod.rs
```diff
@@ -14,7 +14,8 @@ mod sub;
 mod types;
 pub mod walk;
 
-#[allow(warnings)]
+#[cfg(test)]
 mod parser;
-#[allow(warnings)]
+
+#[cfg(test)]
```
affo: I saw what you have done here, but I still don't like the `#[cfg(test)]` above those modules 🤔

libflux/src/semantic/tests.rs
```diff
@@ -2743,7 +2734,7 @@ fn multiple_constraints() {
             e = f(a: 2019-10-30T00:00:00Z, b: 2019-10-31T00:00:00Z)
         "#,
         exp: map![
-            "x" => "forall [t0] where t0: Comparable + Equatable (a: t0, b: t0) -> bool",
```
affo: this one fails, don't know why 🤔

libflux/src/semantic/types.rs
```diff
@@ -139,7 +139,22 @@ impl PolyType {
...
+                    pkinds.sort();
+                    if kinds != pkinds {
+                        return false;
+                    }
+                } else {
+                    return false;
+                }
+            }
+            true
+        }
```
affo: bug fix for equality

libflux/src/semantic/nodes.rs
```diff
@@ -519,8 +519,74 @@ pub struct FunctionExpr {
...
+                }
+                None => {
+                    // We are here: `f = (a) => {...}`.
+                    // So, we do not know the type of "a". Let's use a fresh TVar.
+                    let id = param.key.name.clone();
+                    let ftvar = f.fresh();
+                    let typ = PolyType {
+                        vars: Vec::new(),
+                        cons: HashMap::new(),
+                        expr: MonoType::Var(ftvar.clone()),
```
jlapacik: Just a note, type variables are `Copy` so no need to explicitly clone. However cloning is in no way expensive for type variables.

libflux/src/semantic/nodes.rs
```diff
@@ -519,8 +519,74 @@ pub struct FunctionExpr {
...
+                }
+            };
+        }
+        // Add the parameters to some nested environment.
+        let mut nenv = Environment::new(env);
+        for (id, param) in params.into_iter() {
+            nenv.add(id, param);
+        }
+        // And use it to infer the body.
+        let (nenv, bcons) = self.body.infer(nenv, f)?;
```
jlapacik: I love how functions must always have a return value and so we can treat the body as one big let expressions. So nice.

libflux/src/semantic/nodes.rs
```diff
@@ -710,8 +786,54 @@ pub struct CallExpr {
...
+        // update the environment and the constraints, and use the inferred types to
+        // build the fields of the type for this call expression.
+        let (mut env, mut cons) = self.callee.infer(env, f)?;
+        let mut req = HashMap::new();
+        let mut pipe = None;
+        for Property {
+            key: ref mut id,
+            value: ref mut expr,
+            ..
+        } in &mut self.arguments
```
jlapacik: Whoa I've never seen this kind of destructuring before in a for loop. But I think I love it.

libflux/src/semantic/nodes.rs
```diff
@@ -710,8 +786,54 @@ pub struct CallExpr {
...
+                pipe,
+                // The return type of a function call is the type of the call itself.
+                // Remind that, when two functions are unified, their return types are unified too.
+                // As an example take:
+                //   f = (a) => a + 1
+                //   f(a: 0)
+                // The return type of `f` is `int`.
+                // The return type of `f(a: 0)` is `t0` (a fresh type variable).
+                // Upon unification a substitution "t0 => int" is created, so that the compiler
+                // can infer that, for instance, `f(a: 0) + 1` is legal.
```
jlapacik: Great comment

libflux/src/semantic/types.rs
jlapacik: Thanks for this. I'm wondering if we should remove the display logic that sorts kinds since that made this bug kind of confusing?

libflux/src/semantic/types.rs
```diff
@@ -959,8 +974,123 @@ impl MaxTvar for Function {
...
+    /// Unify 1 and 3: no problem, required arguments are satisfied. Take care that, if you unify
+    /// 3 with 1, you will find `a` in 1's pipe argument.
+    ///
+    /// Unify 1 and 4: should fail because `d` != `a`.
+    ///
+    /// Unify 2 and 3: should fail because `a` is not in the arguments of 2.
+    ///
+    /// Unify 2 and 4: should succeed, the same as 1 and 2.
+    ///
+    /// Unify 3 and 4: should fail because `a` is not in the arguments of 4.
```
jlapacik: This is a great comment by the way.

libflux/src/semantic/nodes.rs
affo: yeah! `..` is like `and all the rest I don't care about`.

Issue #1608, opened 2019-07-25
Improve how the rust scanner manages token positions


affo:
It seems like instead of storing the file offset inside the token, we could just keep the start `Position` (including the line and column) and the end `Position` directly in the token returned from `scan()`.

Then we wouldn't need to do the mapping from file offset to `Position` which does a binary search for each call to `pos()` and has (log n) performance.

Moreover, this would simplify the logic into `scanner.rl` that wouldn't need to create a dynamic list for offsets, but should simply store the offset of the current line in order to attach a proper line/column to the matched token.

We discussed how this might increase the memory footprint since tokens may be cloned frequently. It seems like if tokens are immutable, this could be avoided.

PR #2124, opened 2019-11-09
refactor(libflux): remove usage of malloc from Rust scanner

wolffcm:
We currently use a linked list (which needs `alloc` and `free`) to keep track of where newlines are when scanning tokens.  This was problematic because the wasm build has poor support for these C library functions.

This PR does the following:
- Avoids using `malloc` inside of the C code that hosts the Ragel scanner, by tracking newlines in- and outside of tokens.  The returned token now contains both offset and line/column for its beginning and end.
- Makes some minor tweaks to the docker file that does the wasm build on Mac.
- Removes the `Scanner::pos()` method (which translates from buffer offset to line/col) from the public interface of the scanner, since this info is now available directly from the token.
- The parser still needs to go from line/column back to an offset to get the raw source for each AST node.  To achieve this, I re-implemented the `Scanner::offset()` method using a hash map.
- We previously were not able to correctly compute line/column when doing an `unread` (needed to correctly handle regular expressions), this fixes that issue and adds a test.
- We also had a similar issue when creating a `BadStatement` or `BadExpression` due to an invalid token.

I took care to ensure that we get the same answer as before by keeping both mechanisms for determining line/column around and using `assert_eq!` that they are the same (and then deleting the linked list code once I was convinced)

### DIFF
--- a/libflux/src/scanner/scanner.h
+++ b/libflux/src/scanner/scanner.h
@@ -75,18 +75,13 @@ WASM_EXPORT int scan(
     const unsigned char *data,
     const unsigned char *pe,
     const unsigned char *eof,
-
     const unsigned char **last_newline,
     unsigned int *cur_line,
-
     unsigned int *token,
     unsigned int *token_start,
     unsigned int *token_start_line,
     unsigned int *token_start_col,
     unsigned int *token_end,
     unsigned int *token_end_line,
-    unsigned int *token_end_col,
-
-    const unsigned int **newlines,
-    unsigned int *newlines_len
+    unsigned int *token_end_col
 );
--- a/libflux/src/scanner/scanner.h
+++ b/libflux/src/scanner/scanner.h
@@ -69,6 +69,24 @@ enum TOK {
 #define WASM_EXPORT __attribute__ ((visibility("default")))
 
 // Scan reads the input and reports the next lexical token. Returns the execution state.
-WASM_EXPORT int scan(int mode, const unsigned char **p, const unsigned char *data, const unsigned char *pe,
-        const unsigned char *eof, unsigned int *token, unsigned int *token_start, unsigned int *token_end,
-        const unsigned int **newlines, unsigned int *newlines_len);
+WASM_EXPORT int scan(
+    int mode,
+    const unsigned char **p,
+    const unsigned char *data,
+    const unsigned char *pe,
+    const unsigned char *eof,
+
+    const unsigned char **last_newline,
+    unsigned int *cur_line,
+
+    unsigned int *token,
+    unsigned int *token_start,
+    unsigned int *token_start_line,
+    unsigned int *token_start_col,
+    unsigned int *token_end,
+    unsigned int *token_end_line,
+    unsigned int *token_end_col,
+
+    const unsigned int **newlines,
+    unsigned int *newlines_len
+);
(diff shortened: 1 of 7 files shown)


affo:
Fixes #1608

libflux/src/scanner/scanner.rl
```diff
@@ -52,7 +66,9 @@
         regex_lit => { tok = REGEX; fbreak; };
 
         # We have to specify whitespace here so that leading whitespace doesn't cause a state transition.
-        whitespace+;
+        whitespace;
```
affo: Why have you removed the `+`?

libflux/src/scanner/scanner.rl
```diff
@@ -8,8 +7,22 @@
...
     include WChar "unicode.rl";
 
-    # For comments on newlines tracking, see below.
-    newline = '\n' @{ push(&nls, fpc - data + 1); no_nls++; };
+    action advance_line {
+        // We do this for every newline we find.
+        // This allows us to return correct line/column for each token
+        // back to the caller.
+        (*cur_line)++;
+        *last_newline = fpc + 1;
```
affo: I don't like this name 😅 

Could it be something like `last_newline_col`? Just to make it explicit that it refers to the column position

libflux/src/scanner/scanner.rl
```diff
@@ -8,8 +7,22 @@
...
+    }
+
+    action advance_line_between_tokens {
+        // We do this for each newline we find in the whitespace between tokens,
+        // so we can record the location of the first byte of a token.
+        last_newline_before_token = *last_newline;
+        cur_line_token_start = *cur_line;
+    }
```
affo: OK, now I think I got it, this seems like pretty smart and advanced Ragel use.

So, `advance_line` gets called while you are lexing tokens, so even inside of a single token (e.g. a multiline string), while `advance_line_between_tokens` gets called only outside of token lexing (as if tokens are atomic units).

So `advance_line_between_tokens` saves the last position of the last newline before lexing the next token.  
`advance_line` will update the lines while lexing the next token. When you get the next token you can use `(cur_line_token_start, last_newline_before_token)` for start pos and `(cur_line, last_newline)` for end pos.

I got confused by names, but I guess they are ok 👍 😅

libflux/src/scanner/scanner.rl
wolffcm: The whitespace token defined above already has a trailing `+`, so it's redundant here.

libflux/src/scanner/scanner.rl
wolffcm: Yes, that's exactly right.

libflux/src/scanner/scanner.h
```diff
@@ -69,6 +69,19 @@ enum TOK {
...
+    const unsigned char **last_newline,
+    unsigned int *cur_line,
+    unsigned int *token,
+    unsigned int *token_start,
+    unsigned int *token_start_line,
+    unsigned int *token_start_col,
+    unsigned int *token_end,
+    unsigned int *token_end_line,
+    unsigned int *token_end_col
+);
```
fchikwekwe: Much more readable.

libflux/src/scanner/scanner.rl
fchikwekwe: This is a great way to get around using a linked list. Pretty cool use of Ragel.

PR #2125, opened 2019-11-11
test: type inference coverage

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

First commit constrains unary expressions to be the same type as their operands. The second commit adds greater type inference test converage.

### DIFF
--- a/libflux/src/semantic/tests.rs
+++ b/libflux/src/semantic/tests.rs
@@ -255,96 +255,294 @@ fn literals() {
 // variables that are type variables.
 
 #[test]
-fn string_expr() {
-    let src = r#"message = "Hello, ${name}!""#;
-    //    test_infer! {
-    //        env: map![
-    //            "name" => "forall [] t0",
-    //        ],
-    //        src: src,
-    //        exp: map![
-    //            "name" => "forall [] string",
-    //            "message" => "forall [] string",
-    //        ],
-    //    }
-    test_infer_err! {
-        env: map![
-            "name" => "forall [] int",
-        ],
-        src: src,
-    }
+fn string_interpolation() {
+    // test_infer! {
+    //     env: map![
+    //         "name" => "forall [] t0",
+    //     ],
+    //     src: src,
+    //     exp: map![
+    //         "name" => "forall [] string",
+    //         "message" => "forall [] string",
+    //     ],
+    // }
     test_infer! {
         env: map![
             "name" => "forall [] string",
         ],
-        src: src,
+        src: r#"
+            message = "Hello, ${name}!"
+        "#,
         exp: map![
             "message" => "forall [] string",
         ],
     }
+    test_infer_err! {
+        env: map![
+            "name" => "forall [] bool",
+        ],
+        src: r#"
+            "Hello, ${name}!"
+        "#,
+    }
+    test_infer_err! {
+        env: map![
+            "name" => "forall [] int",
+        ],
+        src: r#"
+            "Hello, ${name}!"
+        "#,
+    }
+    test_infer_err! {
+        env: map![
+            "name" => "forall [] uint",
+        ],
+        src: r#"
+            "Hello, ${name}!"
+        "#,
+    }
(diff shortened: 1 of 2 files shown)


libflux/src/semantic/tests.rs
```diff
@@ -255,715 +255,2377 @@ fn literals() {
...
+            "Hello, ${name}!"
+        "#,
+    }
+    test_infer_err! {
+        env: map![
+            "name" => "forall [t0] (x: t0) -> t0",
         ],
+        src: r#"
+            "Hello, ${name}!"
+        "#,
```
affo: So, no type except from `string` is `Stringable`?

Do we require the user to manually `string(v: ...)`?

libflux/src/semantic/tests.rs
```diff
@@ -255,715 +255,2377 @@ fn literals() {
...
+        exp: map![
+            "c" => "forall [] uint",
+        ],
+    }
+    test_infer! {
         env: map![
             "a" => "forall [] float",
-            "b" => "forall [] int",
+            "b" => "forall [] float",
```
affo: what if you remove `b` from the prelude? Would it error or we expect infer to infer `b` too?

libflux/src/semantic/tests.rs
```diff
@@ -255,715 +255,2377 @@ fn literals() {
...
+        "#,
+    }
+    test_infer_err! {
+        env: map![
+            "a" => "forall [] float",
+            "b" => "forall [] int",
+        ],
+        src: r#"
+            c = a + b
+        "#,
```
affo: this will drive users crazy ahahahah. Cannot sum int and float. Think we need a `Number` kind for which addition is defined.

libflux/src/semantic/tests.rs
```diff
@@ -255,715 +255,2377 @@ fn literals() {
...
+        }
+        test_infer_err! {
+            env: map![
+                "a" => "forall [] float",
+                "b" => "forall [] int",
+            ],
+            src: &src,
+        }
+    }
+    for op in vec![">=", "<=", ">", "<"] {
```
affo: smart 😎

libflux/src/semantic/tests.rs
```diff
@@ -255,715 +255,2377 @@ fn literals() {
...
+    //         "b" => "forall [] t1",
+    //         "c" => "forall [] float",
+    //     ],
+    //     src: r#"
+    //            d = if a then b else c
+    //        "#,
+    //     exp: map![
+    //         "a" => "forall [] bool",
+    //         "b" => "forall [] float",
+    //         "d" => "forall [] float",
```
affo: why are tests with `TVar` skipped?

libflux/src/semantic/tests.rs
jlapacik: Right now yes, but we will add a `Stringable` kind to the type system eventually so that one can interpolate other expression types into string expressions.

libflux/src/semantic/tests.rs
jlapacik: It would error since `b` was not declared.

libflux/src/semantic/tests.rs
jlapacik: You are absolutely right. The plan is to add such a kind so that something like this would be supported.

libflux/src/semantic/tests.rs
jlapacik: We will test tvar instantiation indirectly using function expressions. The plan is to refactor these commented out tests once we can type function expressions.

affo: I see, we need unification first.

affo:
@jlapacik Your replies to my comments confirm your intent in producing this fix. The framework you designed and brought up is able to add ton of value to Flux. Thank you so much! 🍾

jlapacik:
Thank you @affo

PR #2126, opened 2019-11-11
feat(tableutil): add internal table utility for streaming tables

jsternberg:
This creates a new utility for constructing a table stream. Its primary
focus is to allow sources that create tables to easily create buffered
table streams. This is demonstrated by updating the internal point
generator so that it will now natively construct arrow arrays using the
arrow table buffer and stream the chunks downstream using the stream
method.

These utilities are being kept as internal packages while we experiment
with possible methods of constructing tables quickly and efficiently.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/execute/tableutil/stream.go
+++ b/internal/execute/tableutil/stream.go
@@ -1,0 +1,187 @@
+package tableutil
+
+import (
+	"context"
+	"sync/atomic"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+)
+
+// SendFunc is used to send a flux.ColReader to a table stream so
+// it can be read by the table consumer.
+type SendFunc func(flux.ColReader)
+
+// Stream will call StreamWithContext with a background context.
+func Stream(f func(ctx context.Context, fn SendFunc) error) (flux.Table, error) {
+	return StreamWithContext(context.Background(), f)
+}
+
+// StreamWithContext will create a table that streams column readers
+// through the flux.Table. This method will return only after
+// the function buffers the first column reader.
+// This first column reader is used to identify the group key
+// and columns for the entire table stream.
+//
+// Implementors using this *must* return at least one table.
+// If the function returns without returning at least one table,
+// then an error will be returned. If the first table that is returned
+// is empty, then this will return an empty table and further buffers
+// will not be used.
+func StreamWithContext(ctx context.Context, f func(ctx context.Context, fn SendFunc) error) (flux.Table, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	ch := make(chan streamBuffer)
+
+	// Create the send method.
+	send := func(cr flux.ColReader) {
+		select {
+		case ch <- streamBuffer{cr: cr}:
+		case <-ctx.Done():
+			// We could not send the column reader because this was cancelled.
+			cr.Release()
+		}
+	}
+
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		defer close(ch)
+		if err := f(ctx, send); err != nil {
+			ch <- streamBuffer{err: err}
+		}
+	}()
+
+	select {
+	case sp := <-ch:
+		cr, err := sp.cr, sp.err
+		if cr == nil {
+			if err == nil {
+				err = errors.New(codes.Internal, "empty table stream")
+			}
+			cancel()
+			return nil, err
+		}
+
+		// Retrieve the group key and columns from the column reader.
+		key, cols := cr.Key(), cr.Cols()
+
+		// If the table is empty, signal to the context
+		// that we are not expecting more tables just
+		// in case the implementor does something wrong.
+		// We also release the column reader since we don't need
+		// it anymore and set it to nil.
+		empty := cr.Len() == 0
+		if empty {
+			cancel()
+			cr.Release()
+			cr = nil
+		}
+		return &streamTable{
+			first:  cr,
+			key:    key,
+			cols:   cols,
+			cancel: cancel,
+			ch:     ch,
+			done:   done,
+			empty:  empty,
+		}, nil
+	case <-ctx.Done():
+		cancel()
+		return nil, ctx.Err()
+	}
+}
+
+// streamBuffer is a column reader or error sent
+// from the streaming function.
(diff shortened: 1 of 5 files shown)


jsternberg:
I forgot to include the benchmark change for this. I wrote a benchmark for `gen.Input` to show the change.

```
benchmark                     old ns/op       new ns/op     delta
BenchmarkInput/1000-12        57067289        12255450      -78.52%
BenchmarkInput/100000-12      6977867133      19307882      -99.72%
BenchmarkInput/1000000-12     64813674783     21678660      -99.97%

benchmark                     old allocs     new allocs     delta
BenchmarkInput/1000-12        420137         8436           -97.99%
BenchmarkInput/100000-12      40033514       10735          -99.97%
BenchmarkInput/1000000-12     400040503      10735          -100.00%

benchmark                     old bytes       new bytes     delta
BenchmarkInput/1000-12        41218925        7172487       -82.60%
BenchmarkInput/100000-12      7173513832      16520755      -99.77%
BenchmarkInput/1000000-12     68339805312     16558400      -99.98%
```

internal/execute/tableutil/stream.go
```diff
@@ -0,0 +1,184 @@
...
+// streamTable is an implementation of flux.Table
+// that will stream buffers from a column reader.
+type streamTable struct {
+	used   int32
+	first  flux.ColReader
+	key    flux.GroupKey
+	cols   []flux.ColMeta
+	cancel func()
+	ch     <-chan streamBuffer
+	done   <-chan struct{}
```
yzhang1991: Who will check this channel? I just did not find it and it is private.

internal/execute/tableutil/stream.go
```diff
@@ -0,0 +1,184 @@
...
+	"context"
+	"sync/atomic"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+)
+
+// SendFunc is a function that will be used to construct a stream.
+type SendFunc func(flux.ColReader)
```
yzhang1991: Can you make the comment more informative? This function is to send a `ColReader`/ TableBuffer to a table stream so it could be read by the table consumer.

I asked because I was confused and did not get what does this function do until I read all the code below in this file.

internal/execute/tableutil/stream.go
```diff
@@ -0,0 +1,184 @@
...
+// the function buffers the first column reader.
+// This first column reader is used to identify the group key
+// and columns for the entire table stream.
+//
+// Implementors using this *must* return at least one table.
+// If the function returns without returning at least one table,
+// then an error will be returned. If the first table that is returned
+// is empty, then this will return an empty table and further buffers
+// will not be used.
+func StreamContext(ctx context.Context, f func(ctx context.Context, fn SendFunc) error) (flux.Table, error) {
```
yzhang1991: It sounds more like `StreamWithContext`. At first glance, I thought this is to create some context called `StreamContext`

internal/execute/tableutil/stream.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			return err
+		}
+		if err := f(cr); err != nil {
+			cr.Release()
+			return err
+		}
+		cr.Release()
+	}
+	<-s.done
```
yzhang1991: What about the other `return err` cases in the channel, no writes to the `done` channel? We cannot re-enter this method because `used = 1` right? Again, who will check the done channel?

internal/gen/input.go
```diff
@@ -368,30 +371,64 @@ func (dg *dataGenerator) Do(f func(tbl flux.Table) error) error {
...
+				// and then reuse them for each one with a slice.
+				// The first table should always be the biggest because of
+				// the size constraint.
+				// These are constructed lazily below.
+				keyValues := make([]array.Interface, len(s.Cols()))
+				defer func() {
+					for _, vs := range keyValues {
+						if vs != nil {
+							vs.Release()
+						}
```
yzhang1991: Question: why there is no Release like this in `TestStream`?

yzhang1991:
Hi @jsternberg , can I see the code for your benchmark? I think that will be a good learning material for me.

internal/execute/tableutil/stream.go
jsternberg: I'll change the name. I thought I was mimicking parts of the standard library, but I see that it uses `WithContext` for what I thought.

jsternberg:
The code for the benchmark is part of the PR. It is in `internal/gen/input_test.go`.

yzhang1991:
Yeah, I just realized that. I thought it was just a test.

internal/execute/tableutil/stream.go
jsternberg: It is used to prevent the `Do` method from returning until the goroutine associated with streaming ends.

yzhang1991: Ah, I confused myself with the syntax. Sorry

internal/execute/tableutil/stream.go
```diff
@@ -0,0 +1,184 @@
...
+		s.first = nil
+		return nil
+	}
+	s.first.Release()
+	s.first = nil
+
+	for sp := range s.ch {
+		cr, err := sp.cr, sp.err
+		if err != nil {
+			return err
```
yzhang1991: Is this early return going to be a problem anyway for the done channel?

jsternberg: I do not think so. The done channel is just used to prevent this from leaving until the streaming function ends. When an error happens, the streaming function is supposed to be done. If the streaming function continues sending things, then the context gets canceled by this return and we do not want to wait for the done channel. We just need to signal to the sender that we're ignoring it now.

internal/gen/input.go
jsternberg: It's because this piece of code is being clever. Because of the uniform nature of how this constructs tables, it knows that the first buffer will always be the maximum size. So when it constructs group key columns, it attempts to construct them as optimally as possible. It then keeps a reference here and retains a new copy to use for the table buffer below. But, since it is maintaining its own reference, it has to release it when it is done.

So if you generate `1024 * 10` column readers, it only generates one buffer's worth of key values and just shares it between the 10 column readers.

PR #2129, opened 2019-11-11
chore(libflux): expose semantic::fresh

bthesorceror:


nathanielc:
@bthesorceror Why do you need the fresher to be public?

bthesorceror:
@nathanielc currently we don't have a function to turn source into a semantic::nodes::File, to get around that I need to call analyze which requires an argument of type Fresher.

PR #2130, opened 2019-11-11
feat(libflux): expose function to analyze from string

affo:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/semantic/analyze.rs
+++ b/libflux/src/semantic/analyze.rs
@@ -4,17 +4,21 @@ use crate::semantic::nodes::*;
 use crate::semantic::types::MonoType;
 use std::result;
 
-type SemanticError = String;
-type Result<T> = result::Result<T, SemanticError>;
+pub type SemanticError = String;
+pub type Result<T> = result::Result<T, SemanticError>;
 
 // analyze analyzes an AST package node and returns its semantic analysis.
-// The function requires a Fresher to instantiate fresh type veriables for Expressions.
 // The function explicitly moves the ast::Package because it adds information to it.
 // We follow here the principle that every compilation step should be isolated and should add meaning
 // to the previous one. In other terms, once one analyzes an AST he should not use it anymore.
 // If one wants to do so, he should explicitly pkg.clone() and incur consciously in the memory
 // overhead involved.
-pub fn analyze(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
+pub fn analyze(pkg: ast::Package) -> Result<Package> {
+    analyze_with(pkg, &mut Fresher::new())
+}
+
+// analyze_with runs analyze using the provided Fresher.
+pub fn analyze_with(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
     analyze_package(pkg, fresher)
     // TODO(affo): run checks on the semantic graph.
 }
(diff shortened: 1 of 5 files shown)


libflux/src/semantic/mod.rs
```diff
@@ -19,3 +19,23 @@ mod parser;
 
 #[cfg(test)]
 mod tests;
+
+use crate::ast;
+use crate::parser;
+use crate::semantic::analyze::Result;
+use crate::semantic::fresh::Fresher;
+
+pub fn analyze_string(source: &str) -> Result<nodes::Package> {
```
jsternberg: Should this be `analyze_source`?

affo: I wanted to keep consistency with `parse_string` 🤔

libflux/src/semantic/mod.rs
```diff
@@ -19,3 +19,23 @@ mod parser;
...
+    if errs.len() > 0 {
+        return Err(format!("got errors on parsing: {:?}", errs));
+    }
+    let ast_pkg = ast::Package {
+        base: file.base.clone(),
+        path: "".to_string(),
+        package: "main".to_string(),
+        files: vec![file],
+    };
+    analyze(ast_pkg, &mut Fresher::new())
```
jsternberg: It might be better if we expose a version of the `analyze` function that doesn't take a fresher instance. Like we can change `analyze` to take in only an ast package and then have an `analyze_with` method that takes a fresher instance. That way we make things more composable.

affo: Yeah, I see. Anyways, exposing a function that does all the error checking and so on is important. I think that can be done in another PR.

What is weird, is that you could `pub use analyze` without exposing the `Fresher` 🤔

affo: (in the meantime) I changed this to not depend on the internal `Result` type but on a generic stringy one

jsternberg: That part was ok. I think returning a string is probably going to result in a mistake so please change it back to the result type. I think I am changing my review to say that it is better to expose a version of analyze that doesn't take a fresher instance and one that does.

libflux/src/semantic/analyze.rs
```diff
@@ -4,17 +4,21 @@ use crate::semantic::nodes::*;
...
 // to the previous one. In other terms, once one analyzes an AST he should not use it anymore.
 // If one wants to do so, he should explicitly pkg.clone() and incur consciously in the memory
 // overhead involved.
-pub fn analyze(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
+pub fn analyze(pkg: ast::Package) -> Result<Package> {
+    analyze_with(pkg, &mut Fresher::new())
+}
+
+// analyze_with runs analyze using the provided Fresher.
+pub fn analyze_with(pkg: ast::Package, fresher: &mut Fresher) -> Result<Package> {
```
jsternberg: This should not be public since fresher is an implementation detail.

affo: That needs to be public for other modules to use it, that's not public outside of `semantic`.

jsternberg: Looks like `analyze` is already exposed. Never mind.

```rust
mod analyze;
pub use analyze::analyze;
```

Issue #2133, opened 2019-11-11
Optimize pivot for memory usage


jsternberg:
The pivot transformation, like most of the transformations, has a bad pattern of memory access. The reason for this is because it uses the default table builder which doesn't have a convenient way to reserve memory. It has a `GrowXXX` method for growing a column to a length, but there's no way to reserve a certain amount of memory when you know the potential size of the input.

This issue is to optimize pivot. This will likely involve a few things.

1. The table that is constructed should directly construct arrow array buffers. This means it should probably not use the default table builder.
2. It should perform the pivot when it has the available data. This likely means it buffers all of the data in memory and then performs a pivot.

jlapacik:
To clarify, you are proposing that pivot consumes the entire stream, and then allocates memory for the table builder at once to avoid the constant reallocations that occur right now as pivot dynamically grows its columns?

jsternberg:
Correct. Except that right now the pivot algorithm also uses random access so work will have to be done to allow pivot to construct the columns in a way where it doesn't need random access. I think this is possible in a pretty simple way if we do the following process.

1. Identify the columns that have to be merged associated with their row keys.
2. Ensure the row key for each of these is sorted.
3. Perform a join on these into a single table.

There's some additional complications to this. It becomes harder when the row key is multiple rows and it also becomes harder when multiple columns are specified. The algorithm will also likely require that the column key be part of the group key so we might have to ensure that the input is already grouped or have the planner verify that the input will be grouped through a schema aware procedure.
1. Identity the row key and ensure it is sorted

jsternberg:
I think we can possibly constrain this issue to optimize pivot in cases where a single row key and a single column key are given. It will likely require rewriting the core algorithm.

jsternberg:
Related: https://github.com/influxdata/influxdb/issues/14139

jsternberg:
I have a beginning implementation of this done. It merges the row key columns to produce a single row key that can then be used for merging the values. I've implemented it for time values at the moment and the value columns are just completely null values, but I wanted to do a benchmark and test it out to see the difference.

```
jsternberg:[~/g/s/g/i/flux] $ (perf/optimize-pivot *$) benchcmp old.txt new.txt
benchmark                  old ns/op       new ns/op      delta
BenchmarkPivot/1000-12     33557264224     6604501459     -80.32%

benchmark                  old allocs     new allocs     delta
BenchmarkPivot/1000-12     136673348      3992899        -97.08%

[...]

PR #2135, opened 2019-11-12
refactor: expand the interface for BufferedTable

jsternberg:
The `BufferedTable` now exposes the buffers within the table through two
new methods that give the number of buffers and return the buffers
themselves.

BREAKING CHANGE: The interface for `BufferedTable` changed so
implementors of the interface will need to implement the two new
methods.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -79,6 +79,14 @@ func (tb *tableBuffer) Empty() bool {
 	return len(tb.buffers) == 0
 }
 
+func (tb *tableBuffer) Buffer(i int) flux.ColReader {
+	return tb.buffers[i]
+}
+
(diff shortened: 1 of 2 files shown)


jsternberg:
There's only one implementation at the moment. I have been mulling using this for improving some transformations. Notably, I think it will be useful for improving [pivot](https://github.com/influxdata/flux/issues/2133). I also think that this would have been helpful for implementing `promql.join`. We ended up using a separate implementation to read and retain the column readers because there was no way to get the buffers out of the buffered table so there was no way to use `CopyTable` to read the table and retain a reference.

jlapacik:
Yes I remember now. Thank you.

Issue #2090, opened 2019-11-01
Translate Rust semantic graph into FlatBuffers representation


wolffcm:
This is the analogous issue for #2004.  This can largely follow the example there.

DOD:
- There is a new module under `semantic` called `flatbuffers` that has a `serialize` method.
- Tests exist showing that each kind of semantic graph node is correctly translated into its FlatBuffers representation.

nathanielc:
Note this does not include the monotype information on the semantic structs. See #2091 for the sequencing plan, this issue is part of Phase 1.

PR #2139, opened 2019-11-12
feat(semantic): define the flatbuffers schema for semantic graph

### LINKED ISSUE #2089: Write FlatBuffers schema for semantic graph
This is the analogous issue for the AST serialization, #2003.  Mostly this issue can follow the example there.

Definition of done:
- a `semantic.fbs` schema file lives somewhere in `flux/semantic/internal/`
- Generated code for Rust and Go is committed, with basic tests to ensure it compiles.
- Invocation of `flatc` should be driven from Makefile.

fchikwekwe:
Similar to PR #2078, this PR will:

- define a FlatBuffers schema for the semantic graph
- generate the necessary Go and Rust files
- test the FB schema to ensure proper functionality
- drive Rust file generation from the Makefile 
- drive Go file generation whenever `make generate` is run 

This PR closes #2089. 
Closing this PR will unblock issue #2090 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/src/semantic/flatbuffers/tests.rs
```diff
@@ -0,0 +1,67 @@
...
+        &FloatLiteralArgs {
+            value: 3.5,
+            ..FloatLiteralArgs::default()
+        },
+    );
+
+    let increment = UnaryExpression::create(
+        &mut builder,
+        &UnaryExpressionArgs {
+            operator: Operator::AdditionOperator,
```
wolffcm: You know, it turns out that Flux does not have unary plus.  I just tried it out, I got an error.  Unary minus is a thing in Flux though.

(Unary plus definitely exists in C)

This test is still valid for your purposes though.

semantic/semantic.fbs
```diff
@@ -0,0 +1,354 @@
...
+    loc:SourceLocation;
+    operator:Operator; 
+    left: Expression;
+    right: Expression;
+}
+
+table CallExpression {
+    loc:SourceLocation;
+    callee:Expression;
+    arguments:[ObjectExpression];
```
wolffcm: I made the AST FlatBuffers equivalent of this be a single `ObjectExpression`; it might be good to do the same here.

In the Go semantic graph, it's just a single `ObjectExpression` as well:
https://github.com/influxdata/flux/blob/master/semantic/graph.go

However in Rust, it's this:
https://github.com/influxdata/flux/blob/master/libflux/src/semantic/nodes.rs#L684

semantic/semantic.fbs
```diff
@@ -0,0 +1,354 @@
...
+table ObjectExpression {
+    loc:SourceLocation;
+    with:IdentifierExpression;
+    properties:[Property];
+}
+
+table UnaryExpression {
+    loc:SourceLocation;
+    operator:Operator;
+    argument:WrappedExpression;
```
wolffcm: This can be an unwrapped `Expression`, can't it?

semantic/semantic.fbs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+table Property {
+    loc:SourceLocation;
+    key:PropertyKey; 
+    value:WrappedExpression;
```
wolffcm: Same here, I don't think this needs to be wrapped.  I think unions only need to be wrapped if they are in an array.

libflux/src/semantic/flatbuffers/tests.rs
fchikwekwe: Oh, that's good to know.

fchikwekwe: Do you think I should change it to something that might actually show up in Flux?

semantic/semantic.fbs
fchikwekwe: yeah, it can. I'll make that change.

libflux/src/semantic/flatbuffers/tests.rs
wolffcm: It's probably a good idea to change it, since it might confuse someone into thinking we have unary plus in Flux.

fchikwekwe: 💯 will do

semantic/semantic.fbs
```diff
@@ -0,0 +1,354 @@
...
+
+table ExternalVariableAssignment {
+    loc:SourceLocation; 
+    identifier:Identifier;
+}
+
+table ExternBlock {
+    loc:SourceLocation;
+    node:Block;
+}
```
jlapacik: These are no longer concepts in the Rust semantic graph.

fchikwekwe: Ahh, okay. I'll go ahead and remove these.

PR #2152, opened 2019-11-13
feat(semantic): infer imported package types

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/semantic/nodes.rs
+++ b/libflux/src/semantic/nodes.rs
@@ -72,6 +72,11 @@ impl Error {
             msg: format!("unsupported unary operator {}", op.to_string()),
         }
     }
+    fn unknown_import_path(path: &str) -> Error {
+        Error {
+            msg: format!("\"{}\" is not a known import path", path),
+        }
+    }
 }
 
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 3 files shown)


libflux/src/semantic/tests.rs
```diff
@@ -230,6 +334,79 @@ fn instantiation_1() {
...
+                "a" => "forall [] bool",
+                "b" => "forall [] time",
+            ],
+        ],
+        src: r#"
+            import foo "path/to/foo"
+
+            foo.a + foo.b
+        "#,
+    }
```
nathanielc: Can you add a simple test case where you do not specify the `as` name of the import? Like `import "path/to/foo"` and then use it as `foo`?

libflux/src/semantic/nodes.rs
```diff
@@ -229,37 +251,57 @@ pub struct File {
...
+                            Ok((env, cons + rest))
+                        }
+                        Statement::Return(_) => Err(Error::invalid_statement(String::from(
+                            "cannot have return statement in file block",
+                        ))),
+                    },
+                )?;
+
+        for name in imports {
+            env.remove(name);
```
nathanielc: Is my understanding correct that it is safe to remove the names here because we have already recursed into all children of the file and so the env names are no longer needed?

jlapacik: Yes correct. We do not want these names to stay in the environment as they are not visible to other files in the same package.

PR #2153, opened 2019-11-13
ci: build libflux wasm target in CI

wolffcm:
This adds a `libflux-wasm` target to the Makefile and builds adds that target to the CircleCI configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ build: libflux
 
 clean:
 	rm -rf bin
-	cd libflux; $(CARGO) clean
+	cd libflux && $(CARGO) clean && rm -rf pkg
 
 cleangenerate:
 	rm -rf $(GENERATED_TARGETS)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -151,14 +151,18 @@ bench:
 release:
 	./release.sh
 
+libflux-wasm: libflux/pkg/flux_bg.wasm
 
+libflux/pkg/flux_bg.wasm:
```
jsternberg: It seems to me like this will cause it not to build the dependency if it has ever been built before. If the wasm build doesn't produce the dependencies, we may need to remove this target and decide that we will always build the wasm library whenever it is requested.

wolffcm: Hmm, `wasm-pack` just calls `cargo` under the hood in addition to some other stuff.  So I would think that it can use a similar way to detect when dependencies have changed.  It doesn't seem to generate the `libflux.d` file for the `wasm` target though, I'm not sure why.

I've updated the PR to just always build the wasm library as you've suggested.

jsternberg: Cool. If we can, we'll fix this later. If it becomes a big problem that it always attempts a rebuild than we'll deal with that then.

PR #2156, opened 2019-11-14
test: type inference with function types

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/semantic/nodes.rs
+++ b/libflux/src/semantic/nodes.rs
@@ -616,55 +616,79 @@ impl BinaryExpr {
 
         let cons = match self.operator {
             // The following operators require both sides to be equal.
-            ast::Operator::AdditionOperator
-            | ast::Operator::SubtractionOperator
-            | ast::Operator::MultiplicationOperator
-            | ast::Operator::DivisionOperator
-            | ast::Operator::PowerOperator
-            | ast::Operator::ModuloOperator => {
-                let mut constraints = vec![
-                    Constraint::Equal(self.left.type_of().clone(), self.typ.clone()),
-                    Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
-                ];
-                if let Some(kind) = match self.operator {
-                    ast::Operator::AdditionOperator => Some(Addable),
-                    ast::Operator::SubtractionOperator => Some(Subtractable),
-                    ast::Operator::MultiplicationOperator
-                    | ast::Operator::DivisionOperator
-                    | ast::Operator::PowerOperator
-                    | ast::Operator::ModuloOperator => Some(Divisible),
-                    _ => None,
-                } {
-                    constraints.push(Constraint::Kind(self.typ.clone(), kind));
-                }
-                Constraints::from(constraints)
-            }
-            // The following require the type to be a boolean.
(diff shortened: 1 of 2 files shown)


libflux/src/semantic/nodes.rs
```diff
@@ -616,55 +616,79 @@ impl BinaryExpr {
...
+            ]),
+            ast::Operator::SubtractionOperator => Constraints::from(vec![
+                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                Constraint::Equal(self.left.type_of().clone(), self.typ.clone()),
+                Constraint::Kind(self.typ.clone(), Kind::Subtractable),
+            ]),
+            ast::Operator::MultiplicationOperator => Constraints::from(vec![
+                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                Constraint::Equal(self.left.type_of().clone(), self.typ.clone()),
+                Constraint::Kind(self.typ.clone(), Kind::Divisible),
```
affo: Could you explain this?

libflux/src/semantic/tests.rs
```diff
@@ -2723,3 +2392,391 @@ fn record_with_scoped_labels() {
...
+    test_infer! {
+        src: "f = (x) => x",
+        exp: map![
+            "f" => "forall [t0] (x: t0) -> t0",
+        ],
+    }
+}
+#[test]
+#[ignore]
+fn polymorphic_instantiation() {
```
affo: booooom. I really like your macros, tests are very expressive in their intent 👍

libflux/src/semantic/tests.rs
```diff
@@ -2723,3 +2392,391 @@ fn record_with_scoped_labels() {
...
+            "e" => "forall [] time",
+            "g" => "forall [] regexp",
+            "h" => "forall [] [int]",
+            "i" => "forall [t0] [t0]",
+            "j" => "forall [] {a: int | b: float}",
+        ],
+    }
+}
+#[test]
+#[ignore]
```
affo: Yeah, this ones will work once I merge "my" contribution 😅

libflux/src/semantic/nodes.rs
affo: I mean, why iuf you find a multiplication you require the Kind to be "Divisible"? Isn't it to much of a constraint? 🤔 

Ooook, but I see that `Int` for example is `Divisible`.

```
MonoType::Int => match with {
                Kind::Addable
                | Kind::Subtractable
                | Kind::Divisible
                | Kind::Comparable
                | Kind::Equatable
                | Kind::Nullable => Ok(Substitution::empty()),
            },
```

jlapacik: Right now `Divisible` is the constraint for multiplication. Not sure if that will need to change in the future, but at least for right now every type that is divisible can also be multiplied. This is not true for `Addable` and `Subtractable` though, since strings are addable but not subtractable.

libflux/src/semantic/tests.rs
jlapacik: Thank you.

Issue #2131, opened 2019-11-11
Allow procedure specs to expose table schemas in the planner


jsternberg:
There are some transformations that could be more efficiently implemented if we knew the incoming table schema. An example is the sort function where, if we knew that the table was already sorted, we wouldn't need to perform a sort at all. Another example is the pivot function which will attempt to join rows from different tables based on a set of columns. If we know those columns are already sorted, we don't have to do any sort to join these.

This shows up in a lot of optimization areas. We should find a way to expose table schemas to the planner so that the planner can inform the execution.

jsternberg:
I thought of a good potential implementation of this. The interfaces are below with godoc comments.

```go
// SchemaAwareProcedureSpec is a ProcedureSpec that
// is able to answer questions about the schema for
// the tables it produces.
type SchemaAwareProcedureSpec interface {
	ProcedureSpec

	// GetTableSchema will return the known TableSchema of the procedure.
	// It will be passed in the plan.Node of the procedure
	// spec which can be used to determine the parents.
	GetTableSchema(n Node) TableSchema
}
```

This is an optional interface that a procedure spec could implement. This allows a planner rule to retrieve the table schema for a particular node in the graph. The reason why the `plan.Node` needs to be passed in is because a procedure spec may need to refer to its predecessors when determining its own table schema. This is seen in the circumstances like narrow transformations which do not modify the schema, but just pass on the schema to the next transformation.

```go
// TableSchema is the known information about the table schemas
// that are produced by a procedure.
//
// The schema can be queried to get information about these
// tables. But, the schema may have incomplete knowledge.
//
// When filling out a schema for a procedure, include as many
// methods as you can for the information that can be known
// before the procedure is executed. This information may be
// known statically or dynamically, but it must be capable
// of being known before the procedure is executed and be
// consistent with any changes that may happen between retrieving
// the schema and executing the procedure.
//
// When both filling out the schema and reading it, keep in mind
// that a source may produce tables with different schemas.
// These methods should only return Yes or No if it applies to
// all tables that the source would produce.
//
// If an attribute is not filled out, then it is assumed that
// the procedure is not aware of that part of the schema.
// This causes the function to always return maybe.
type TableSchema struct {
	// IsKeyColumnFunc defines how this schema responds to an IsKeyColumn query.
	IsKeyColumnFunc func(label string) Answer

	// IsValueColumnFunc defines how this schema responds to an IsValueColumn query.
	IsValueColumnFunc func(label string) Answer

[...]

PR #2161, opened 2019-11-14
perf(stdlib): optimize pivot for common scenarios

jsternberg:
This optimizes pivot for the common scenario of a single row and single
column key where the column key is part of the group key.

It will perform a pivot by reading all of the relevant tables and it
will then merge all of the keys together in sorted order. After it has
done that, it will build each column by matching the value with the
appropriate key. This causes the algorithm to run in `O(n)` time and
avoid unnecessary reallocations because it determines the size before it
attempts to build any arrays.

#2133

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
Benchmark for this:

```
jsternberg:[~/g/s/g/i/flux] $ (perf/optimize-pivot *$) benchcmp old.txt new.txt
benchmark                  old ns/op       new ns/op      delta
BenchmarkPivot/1000-12     33557264224     5613306182     -83.27%

benchmark                  old allocs     new allocs     delta
BenchmarkPivot/1000-12     136673348      3813245        -97.21%

benchmark                  old bytes      new bytes      delta
BenchmarkPivot/1000-12     9367265416     3319302120     -64.56%
```

It likely still needs some tests, but this PR doesn't enable the code. I think we're likely going to need #2131 for this to work effectively so we can confidently enable this code in specific scenarios.

jsternberg:
I updated this so it works with the new way that tools are versioned. This is ready for review again.

jsternberg:
Test cases from the v1 version of pivot were adapted for v2. The v2 one only works if the column key is part of the group key and that the row key has no nulls so those tests were removed.

stdlib/universe/pivot.go
```diff
@@ -382,3 +390,285 @@ func (t *pivotTransformation) Finish(id execute.DatasetID, err error) {
...
+	}
+
+	if err := tb.Validate(); err != nil {
+		return nil, err
+	}
+	return tableutil.FromBuffer(tb), nil
+}
+
+//go:generate -command tmpl ../../gotool.sh github.com/benbjohnson/tmpl
+//go:generate tmpl -data=@../../internal/types.tmpldata -o pivot.gen.go pivot.gen.go.tmpl
```
nathanielc: nit, can you move these to the top of the file?

jsternberg: Sure. I put it at the bottom because that's where the code originally was, but it makes sense at the top too.

jsternberg:
I think I probably do need to create tests for the new functionality that isn't being tested. I'll try to write those today. If this begins to get too delayed because I'm having difficulty writing the tests, we can split the work.

jsternberg:
Wrote tests for arrowutil and the other stuff that's in here. I also cleaned up the iterators a bit. Previously, I had next be called after reading the value, but that's not really expected in typical Go APIs. I updated it so next is always called before reading the value. That made the merge code a bit different (and the previous api made it easier), but I don't think there's much of a meaningful difference and having the iterators with a weird API specifically for one algorithm seemed a bit strange to me.

jsternberg:
I tried to add more patch coverage. I think part of the issue is a lot of the generated file is not actually tested because it generates the code paths for the row key to be a non-time column. I think I am just going to have to go with what I currently have for now. I believe the test coverage to be sufficient enough.

PR #2162, opened 2019-11-14
feat: flatbuffers ast to go ast

yzhang1991:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


ast/ast.go
```diff
@@ -64,6 +72,14 @@ func (l *SourceLocation) Copy() *SourceLocation {
...
+	l.File = string(buf.File())
+	l.Start.FromBuf(buf.Start(nil))
+	l.End.FromBuf(buf.End(nil))
+	l.Source = string(buf.Source())
+	return &l
+}
+
 // Node represents a node in the InfluxDB abstract syntax tree.
 type Node interface {
 	node()
```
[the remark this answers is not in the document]
yzhang1991: The argument types and the return types are not the same.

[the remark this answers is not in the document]
wolffcm: Yeah, I came to the same conclusion for the semantic graph as well.

ast/flatbuffers_test.go
```diff
@@ -1,12 +1,120 @@
...
+			Package: packageName,
+			Files:   []*ast.File{file},
+		}
+		//astRust := rparser.ParseSource(src)
+
+		if !cmp.Equal(astFbs, astGo, CompareOptions...) {
+			t.Errorf("AST roundtrip vs. Go unexpected packages -fbs/+go:\n%s",
+				cmp.Diff(astFbs, astGo, CompareOptions...))
+		}
+		//if !cmp.Equal(astFbs, astRust, CompareOptions...) {
```
wolffcm: Can you link to the related issues in GitHub here?

wolffcm: Actually, if you can create a more liberal `CompareOptions` that would allow Rust-JSON AST comparison to succeed, it might be good to do that here, instead of commenting out code.  I'm okay with either way.

libflux/go/libflux/parser.go
```diff
@@ -63,3 +76,14 @@ func Parse(s string) *ASTFile {
...
 	return f
 }
+
+func ParseIntoFbs(s string) *ast.Package {
+	cstr := C.CString(s)
+	defer C.free(unsafe.Pointer(cstr))
+
+	ptr := C.flux_parse_fb(cstr)
+	f := &AstBuf{ptr: ptr}
+	runtime.SetFinalizer(f, free)
```
wolffcm: Do you need `SetFinalizer` here?  You can just use `defer` instead, right?  In the `Parse` method above, we're returning an `*ASTFile` that contains a pointer that needs to be freed.  But here you're returning an AST that lives entirely in memory managed by Go.

yzhang1991: Yes, you are right.

PR #2174, opened 2019-11-19
feat(libflux): port immutable walk and fix mutable walk

affo:
This PR ports immutable walk with multiple ownership:

 - port immutable walk as in https://github.com/influxdata/flux-lsp/blob/master/src/visitors/semantic/walk.rs
 - fix back mutable walk to be compile-time safe using `&mut`.
 - create a module that contains both and tests both


libflux/src/semantic/mod.rs
```diff
@@ -11,7 +11,7 @@ mod infer;
 pub mod nodes;
 
 mod sub;
-mod types;
+pub mod types;
```
affo: This is required. How can we let somebody visit the semantic graph that contains types if types are not public?

jlapacik: You're right this should be public

PR #2180, opened 2019-11-20
chore(libflux): remove ragel as a build dependency

### LINKED ISSUE #2178: Remove ragel dependency for libflux
Remove ragel dependency for libflux by committing generated code.

bthesorceror:
Closes #2178

- Check in generated scanner code


Makefile
```diff
@@ -128,7 +128,7 @@ tidy:
 checktidy:
 	./etc/checktidy.sh
```
wolffcm: Can you add `libflux/scanner.c` to `GENERATED_TARGETS`?  This way it will get deleted when `make cleangenerate` is run.

Makefile
```diff
@@ -128,7 +128,7 @@ tidy:
...
 
-checkgenerate:
+checkgenerate: libflux/scanner.c
```
wolffcm: I'm curious why you needed to add this?  Or, maybe adding `scanner.c` to `GENERATED_TARGETS` means you won't need to add it here?

Makefile
bthesorceror: Making this change caused CI to fail, reverted back to original solution purposed by @nathanielc

PR #2183, opened 2019-11-21
feat(libflux): inject types in the semantic graph

affo:
This PR adds type injection in the semantic graph.

 - `inject.rs` injects types
 - `semantic::analyze_source` now inject types in the analyzed source;
 - added e2e test for `analyze_source`
 - `Substitutable` change to take `&self` -> this allows to avoid unnecessary type cloning.
 - util methods in `walk::{Node, NodeMut}`.

### DIFF
--- a/libflux/src/semantic/mod.rs
+++ b/libflux/src/semantic/mod.rs
@@ -19,9 +19,21 @@ mod tests;
 
 use crate::ast;
 use crate::parser::parse_string;
-use crate::semantic::analyze::Result;
+use crate::semantic::analyze::analyze_with;
+use crate::semantic::analyze::Result as AnalysisResult;
+use crate::semantic::analyze::SemanticError;
+use crate::semantic::env::Environment;
+use crate::semantic::fresh::Fresher;
+use crate::semantic::nodes::Error as InferError;
+use crate::semantic::nodes::{infer_pkg_types, inject_pkg_types, Importer};
 
-pub fn analyze_source(source: &str) -> Result<nodes::Package> {
+impl From<InferError> for SemanticError {
+    fn from(err: InferError) -> SemanticError {
+        err.msg.clone()
+    }
+}
+
+pub fn analyze_source(source: &str) -> AnalysisResult<nodes::Package> {
     let file = parse_string("", source);
     let errs = ast::check::check(ast::walk::Node::File(&file));
     if errs.len() > 0 {
@@ -33,5 +45,9 @@ pub fn analyze_source(source: &str) -> Result<nodes::Package> {
         package: "main".to_string(),
         files: vec![file],
     };
-    analyze(ast_pkg)
+    let mut f = Fresher::new();
+    let mut sem_pkg = analyze_with(ast_pkg, &mut f)?;
+    // TODO(affo): add a stdlib Importer.
+    let (_, sub) = infer_pkg_types(&mut sem_pkg, Environment::empty(), &mut f, &Importer::new())?;
+    Ok(inject_pkg_types(sem_pkg, &sub))
 }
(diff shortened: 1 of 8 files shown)


libflux/src/semantic/mod.rs
```diff
@@ -37,6 +49,14 @@ pub fn analyze_source(source: &str) -> Result<nodes::Package> {
...
-    analyze(ast_pkg)
-    // TODO infer
+    let mut sem_pkg = analyze(ast_pkg)?;
+    // TODO(affo): add a stdlib Importer.
+    let (_, sub) = infer_pkg_types(
+        &mut sem_pkg,
+        Environment::empty(),
+        &mut Fresher::new(),
+        &Importer::new(),
+    )?;
```
jlapacik: I thought `analyze` took a `Fresher` argument? I think this is the reason the below test case isn't passing. `analyze` and `infer_pkg_types` must take the same mutable `Fresher` reference.

libflux/src/semantic/walk/walk_mut.rs
```diff
@@ -142,6 +144,125 @@ impl<'a> NodeMut<'a> {
...
+            NodeMut::FloatLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::StringLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::DurationLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::UintLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::BooleanLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::DateTimeLit(ref mut n) => n.typ = n.typ.apply(sub),
+            NodeMut::RegexpLit(ref mut n) => n.typ = n.typ.apply(sub),
+            _ => (),
+        }
+    }
```
jlapacik: Ok I think I understand. Is this why you changed `Substitutable`s to take `&self` instead `self`? Because otherwise you'd have to clone here?

But `Node` and `NodeMut` are public types right? Visible from external crates? By implementing `replace_type` on `Node` won't we then have to make the `Substitution` type a public item of this crate? I don't think we want that. Substitutions should only be used internally during type inference.

Perhaps a `visitor` isn't the right job for this. Perhaps `Expression` types should just be `Substitutable`? Something like this?

```Rust
impl Substitutable for Expression {
    fn apply(mut self, sub: &Substitution) -> Self {
        match self {
            Expression::Integer(mut lit) => {
                let typ = lit.typ.apply(&sub);
                lit.typ = typ;
                Expression::Integer(lit)
            }
            ...
            Expression::Binary(mut exp) => {
                let typ = exp.typ.apply(&sub);
                exp.typ = typ;
                Expression::Binary(exp)
            }
            ...
        }
    }
    fn free_vars(&self) -> Vec<Tvar> {
        self.type_of().free_vars()
    }
}
```

libflux/src/semantic/types.rs
```diff
@@ -296,8 +294,8 @@ impl Substitutable for MonoType {
             | MonoType::String
             | MonoType::Duration
             | MonoType::Time
-            | MonoType::Regexp => self,
-            MonoType::Var(tvr) => sub.apply(tvr),
+            | MonoType::Regexp => self.clone(),
```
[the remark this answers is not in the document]
affo: Yeah, but this clone seems like trivial because  the enum has no parameter. I was worried of cloning complex monotypes as  `Function` that can recurse.

libflux/src/semantic/walk/walk_mut.rs
affo: About visibility, I have doubts about this, because `cargo` doesn't warn you or even error if you expose as public something that's "partially" public. So, I don't exactly know what the outcome of invoking `replace_type` without possibly passing a `Substitution` would be. I'll check.

Anyways, your strategy seems pretty elegant. I could `apply` to the expression and replace the reference to the Node in `NodeMut`. This seems to be a great desing, because the Expression itself should be able to move its pieces out to build the return value, thus avoiding cloning! I will try it out today. I am still doubtful about partial moves and lifetimes for pointers in `NodeMut`, but I'll try.

Thanks for highlighting this alternative approach Joshua 😎

PR #2190, opened 2019-11-22
feat(arrow): add support for lazy arrow arrays

jsternberg:
This adds support for lazy arrow arrays. The lazy array implements the
`array.Interface` interface, but exposes an interface that allows it to
be converted to the appropriate underlying arrow type if it needs to be
materialized.

This follows a similar pattern to the new error API where there is an
`As` function in the arrow package and it will first check if the array
interface is assignable to the array type directly or if it can be
coerced with `As`. So to type cast an arrow array, you would do this:

    var arr *array.Int64
    if arrow.As(myArr, &arr) {
        return arr
    }

Functions for each of the arrow array types have been added to do this
more easily and to panic when it doesn't work so these can be integrated
easily with the `flux.ColReader`.

This allows `Repeat` to create an array implementation that can
materialize the arrow array at a later time rather than immediately.

Related to influxdata/influxdb#15985.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/arrow/repeat.go
+++ b/arrow/repeat.go
@@ -2,6 +2,8 @@ package arrow
 
 import (
 	"fmt"
+	"sync"
+	"sync/atomic"
 
 	"github.com/apache/arrow/go/arrow"
 	"github.com/apache/arrow/go/arrow/array"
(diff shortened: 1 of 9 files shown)


jsternberg:
A potential caveat with this is I'm not sure if this is the best design or not. It was the least intrusive way I could think to implement it, but I'm not sure the implementation works for anything other than lazy arrays. I have a potential other solution for the influxdb problem that uses shared buffers, but this one hands down performs better in all situations mostly because I can do the shared buffers _and_ this one and lose absolutely nothing. Since in most cases we will never materialize the group key and the most common area where we create string arrays is for group keys, this helps to drastically reduce the amount of memory that is allocated for any query.

arrow/repeat.go
```diff
@@ -10,60 +12,202 @@ import (
 	"github.com/influxdata/flux/values"
 )
 
-// Repeat will construct an arrow array that repeats
-// the value n times.
+// Repeat will construct an arrow array that repeats the value n times.
 func Repeat(v values.Value, n int, mem memory.Allocator) array.Interface {
```
wolffcm: I wonder if this could be taken a step further and *never* materialize an array for columns that have all the same value?  Since you could just store a scalar value that would be the value returned for any call to `Value`?

jsternberg:
This is being worked on as a different solution.

PR #2192, opened 2019-11-22
feat(libflux/semantic): serialize semantic graph flatbuffers

fchikwekwe:
Semantic graph Flatbuffers need to be serialized in Rust. This commit serializes and performs testing. 

Additional testing may be added to support round trip with deserialization on the Go side. 

The Rust semantic graph `DurationLit` node only contains a single `Duration` value, while the AST and the Go Semantic Graph allow for a list of type `Duration` to accommodate the possible multiple duration values. To temporarily work around this difference, I'm serializing the Flatbuffers `DurationLiteral` with a single duration value stored in a vector. This needs to be fixed before Algorithm W can be deployed in order to not break calendar durations. I'll open another PR to handle that fix to the Rust semantic graph. 

I also worked with @wolffcm to make changes to the code that generates deserializing code on the Go side to accommodate for the difference between CallExpression arguments in the Rust (vector of Properties) and in the Go (pointer to an ObjectExpression) semantic graphs. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/flux/ast/flatbuffers/mod.rs
+++ b/libflux/src/flux/ast/flatbuffers/mod.rs
@@ -692,7 +692,6 @@ impl<'a> ast::walk::Visitor<'a> for SerializingVisitor<'a> {
                         package,
                         imports,
                         body,
-                        ..fbast::FileArgs::default()
                     },
                 );
                 v.files.push(f);
--- a/libflux/src/flux/semantic/flatbuffers/mod.rs
+++ b/libflux/src/flux/semantic/flatbuffers/mod.rs
@@ -1,3 +1,1151 @@
 #![allow(clippy::all)]
 pub mod semantic_generated;
 pub mod types;
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use crate::ast;
+
+use crate::semantic;
+use crate::semantic::walk;
+use flatbuffers::{UnionWIPOffset, WIPOffset};
+use semantic_generated::fbsemantic;
+
+extern crate chrono;
+use chrono::Offset;
+
+pub fn serialize(semantic_pkg: &mut semantic::nodes::Package) -> Result<(Vec<u8>, usize), String> {
+    let mut v = new_serializing_visitor_with_capacity(1024);
+    walk::walk(&mut v, Rc::new(walk::Node::Package(semantic_pkg)));
+    v.finish()
+}
+
+fn new_serializing_visitor_with_capacity<'a>(_capacity: usize) -> SerializingVisitor<'a> {
+    SerializingVisitor {
+        inner: Rc::new(RefCell::new(SerializingVisitorState::new_with_capacity(
+            _capacity,
+        ))),
+    }
+}
+
+struct SerializingVisitor<'a> {
+    inner: Rc<RefCell<SerializingVisitorState<'a>>>,
+}
+
+impl<'a> semantic::walk::Visitor<'_> for SerializingVisitor<'a> {
+    fn visit(&mut self, _node: Rc<walk::Node<'_>>) -> bool {
+        let v = self.inner.borrow();
+        if let Some(_) = &v.err {
+            return false;
+        }
+        Some(SerializingVisitor {
+            inner: Rc::clone(&self.inner),
+        });
+        true
+    }
+
+    fn done(&mut self, node: Rc<walk::Node<'_>>) {
+        let mut v = &mut *self.inner.borrow_mut();
+        if let Some(_) = &v.err {
+            return;
+        }
+        let node = &*node;
+        let loc = v.create_loc(node.loc());
+        match node {
+            walk::Node::IntegerLit(int) => {
+                let int_typ = int.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, int_typ);
+
+                let int = fbsemantic::IntegerLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::IntegerLiteralArgs {
+                        loc,
+                        value: int.value,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack
+                    .push((int.as_union_value(), fbsemantic::Expression::IntegerLiteral))
+            }
+            walk::Node::UintLit(uint) => {
+                let uint_typ = uint.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, uint_typ);
+
+                let uint = fbsemantic::UnsignedIntegerLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::UnsignedIntegerLiteralArgs {
+                        loc,
+                        value: uint.value,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack.push((
+                    uint.as_union_value(),
+                    fbsemantic::Expression::UnsignedIntegerLiteral,
+                ))
+            }
+            walk::Node::FloatLit(float) => {
+                let float_typ = float.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, float_typ);
+
+                let float = fbsemantic::FloatLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::FloatLiteralArgs {
+                        loc,
+                        value: float.value,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack
+                    .push((float.as_union_value(), fbsemantic::Expression::FloatLiteral))
+            }
+            walk::Node::RegexpLit(regex) => {
+                let regex_val = v.create_string(&regex.value);
+                let regex_typ = regex.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, regex_typ);
+
+                let regex = fbsemantic::RegexpLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::RegexpLiteralArgs {
+                        loc,
+                        value: regex_val,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack.push((
+                    regex.as_union_value(),
+                    fbsemantic::Expression::RegexpLiteral,
+                ))
+            }
+            walk::Node::StringLit(string) => {
+                let string_val = v.create_string(&string.value);
+                let string_typ = string.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, string_typ);
+
+                let string = fbsemantic::StringLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::StringLiteralArgs {
+                        loc,
+                        value: string_val,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack.push((
+                    string.as_union_value(),
+                    fbsemantic::Expression::StringLiteral,
+                ))
+            }
+
+            walk::Node::DurationLit(dur_lit) => {
+                let mut dur_vec: Vec<WIPOffset<fbsemantic::Duration>> = Vec::new();
+                let magnitude = match dur_lit.value.num_nanoseconds() {
+                    Some(mag) => mag,
+                    None => {
+                        v.err = Some(String::from("Empty duration value"));
+                        return;
+                    }
+                };
+                let dur = fbsemantic::Duration::create(
+                    &mut v.builder,
+                    &fbsemantic::DurationArgs {
+                        magnitude,
+                        unit: fbsemantic::TimeUnit::ns,
+                    },
+                );
+                dur_vec.push(dur);
+                let value = Some(v.builder.create_vector(dur_vec.as_slice()));
+
+                let dur_typ = dur_lit.typ.clone();
+                let (typ, typ_type) = types::build_type(&mut v.builder, dur_typ);
+
+                let dur_lit = fbsemantic::DurationLiteral::create(
+                    &mut v.builder,
+                    &fbsemantic::DurationLiteralArgs {
+                        loc,
+                        value,
+                        typ: Some(typ),
+                        typ_type,
+                    },
+                );
+                v.expr_stack.push((
+                    dur_lit.as_union_value(),
+                    fbsemantic::Expression::DurationLiteral,
+                ))
+            }
+
+            walk::Node::DateTimeLit(datetime) => {
+                let val = datetime.value.to_rfc3339();
+                let val = v.create_string(&val);
+
+                let secs = datetime.value.timestamp();
+                let nano_secs = datetime.value.timestamp_subsec_nanos();
+                let offset = datetime.value.offset().fix().local_minus_utc();
+
+                let time = fbsemantic::Time::create(
+                    &mut v.builder,
+                    &fbsemantic::TimeArgs {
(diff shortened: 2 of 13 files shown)


fchikwekwe:
There are a few build issues here due to some changes I made to the flatbuffers schema. I'm gonna try and fix those and then update this PR. Feel free to still review if you'd like, but there will another update soon-ish.

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1147 @@
...
+
+impl<'a> semantic::walk::Visitor<'_> for SerializingVisitor<'a> {
+    fn visit(&mut self, _node: Rc<walk::Node<'_>>) -> bool {
+        let v = self.inner.borrow();
+        if let Some(_) = &v.err {
+            return false;
+        }
+        Some(SerializingVisitor {
+            inner: Rc::clone(&self.inner),
+        });
```
rockstar: I wonder if you could explain this block to me, as I haven't done much with `Rc` itself just yet. It looks like we're creating a `SerializingVisitor` with an `inner` that is a strong reference counte of `self.inner`, right, but we're not holding a handle to the visitor, and adding it to a `Option` of `Some` but also not retaining a handle to it. It looks like this block would be a no-op short of incrementing the reference counter, so I think I'm missing something.

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1147 @@
...
+                );
+                v.stmts
+                    .push((opt.as_union_value(), fbsemantic::Statement::OptionStatement));
+            }
+
+            walk::Node::Block(block) => {
+                // Block statements must be convered in this enum in order to also walk all child nodes.
+                // However, block statements are consumed by the function expression node so that
+                // blocks are only constructed once all child nodes have been traversed.
+            }
```
rockstar: Should we add an `unreachable!()` here? To ensure we never actually find ourselves in that block?

libflux/src/flux/semantic/nodes.rs
```diff
@@ -128,8 +128,8 @@ pub enum Assignment {
 impl Assignment {
     fn apply(self, sub: &Substitution) -> Self {
         match self {
-            Assignment::Variable(ass) => Assignment::Variable(ass.apply(&sub)),
-            Assignment::Member(ass) => Assignment::Member(ass.apply(&sub)),
+            Assignment::Variable(assign) => Assignment::Variable(assign.apply(&sub)),
+            Assignment::Member(assign) => Assignment::Member(assign.apply(&sub)),
```
rockstar: This change made me smirk. It's appreciated.

libflux/src/flux/semantic/flatbuffers/semantic_generated.rs
```diff
@@ -2998,6 +3004,30 @@ pub mod fbsemantic {
                 None
             }
         }
+
+        #[inline]
+        #[allow(non_snake_case)]
```
rockstar: Is this lint rule part of the generated code, or did it get added manually?

libflux/src/flux/ast/flatbuffers/mod.rs
```diff
@@ -553,6 +553,7 @@ impl<'a> ast::walk::Visitor<'a> for SerializingVisitor<'a> {
                         init_,
                     },
                 );
+                print!("done with return \n");
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, it is. I'll remove it.

libflux/src/flux/semantic/flatbuffers/semantic_generated.rs
fchikwekwe: Yeah, everything in this file is generated.

libflux/src/flux/semantic/nodes.rs
fchikwekwe: Yeah, I had to read that file too many times to leave it as is.

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: Hmm, I haven't used that warning before. Let me look into it.

fchikwekwe:
@yzhang1991 @wolffcm @rockstar I made some changes that should alleviate conflicts and build issues. Feel free to take a look now.

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: Yeah, so the process followed by the SerializingVisitor is something that I borrowed from @wolffcm when he implemented this on the AST side. By `holding a handle` to the visitor, are you referring to the place where its dereferenced?

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1150 @@
...
+    fn done(&mut self, node: Rc<walk::Node<'_>>) {
+        let mut v = &mut *self.inner.borrow_mut();
+        if let Some(_) = &v.err {
+            return;
+        }
+        let node = &*node;
+        let loc = v.create_loc(node.loc());
+        match node {
+            walk::Node::IntegerLit(int) => {
+                let int_typ = int.typ.clone();
```
wolffcm: I don't think we should need to always clone the type to serialize it.  I think `build_type` should be made to accept a reference.  If this isn't an easy fix, can you file a bug?

libflux/src/flux/semantic/flatbuffers/mod.rs
wolffcm: This arm will be reached (and should be) it's just that it doesn't need to do anything.  Blocks are kind of confusing since each one is essentially a node in a linked list.  What we're doing here is waiting to do anything until we've handled every statement in the linked list of Blocks.

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1150 @@
...
+
+    package: Option<WIPOffset<fbsemantic::Package<'a>>>,
+    package_clause: Option<WIPOffset<fbsemantic::PackageClause<'a>>>,
+
+    import_decls: Vec<WIPOffset<fbsemantic::ImportDeclaration<'a>>>,
+    files: Vec<WIPOffset<fbsemantic::File<'a>>>,
+    blocks: Vec<WIPOffset<fbsemantic::Block<'a>>>,
+    stmts: Vec<(WIPOffset<UnionWIPOffset>, fbsemantic::Statement)>,
+    vars: Vec<(WIPOffset<UnionWIPOffset>, fbsemantic::Var<'a>)>,
+    cons: Vec<(WIPOffset<UnionWIPOffset>, fbsemantic::Constraint<'a>)>,
```
wolffcm: I think this is unused, right?

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1150 @@
...
+                let identifier = fbsemantic::Identifier::create(
+                    &mut v.builder,
+                    &fbsemantic::IdentifierArgs { loc, name },
+                );
+                v.identifiers.push(identifier)
+            }
+
+            walk::Node::Property(prop) => {
+                // the value for a property is always an expression
+                let key = v.identifiers.pop();
```
wolffcm: Would be good to use your handy helper `pop_ident` here.

libflux/src/flux/semantic/flatbuffers/tests.rs
```diff
@@ -0,0 +1,952 @@
...
+    ];
+    let pkg = ast::Package {
+        base: ast::BaseNode {
+            ..ast::BaseNode::default()
+        },
+        path: String::from("./"),
+        package: String::from("test"),
+        files: f,
+    };
+    let mut pkg = match analyze(pkg) {
```
wolffcm: Does this run type inference, so that the semantic graph you're serializing has types?  I guess that it won't, since there are some undefined symbols there.  Can you file a follow up issue to modify this test to run type inference?

libflux/scanner.c
```diff
@@ -92,7 +92,7 @@ static const unsigned char _flux_trans_keys[] = {
 	119u, 121u, 194u, 100u, 104u, 109u, 110u, 115u, 
 	117u, 119u, 121u, 194u, 48u, 57u, 115u, 181u, 
 	170u, 181u, 186u, 128u, 150u, 152u, 182u, 184u, 
-	255u, 192u, 255u, 128u, 255u, 173u, 130u, 133u,
```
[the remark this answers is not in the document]
fchikwekwe: Good catch. I'll change this.

libflux/src/flux/semantic/analyze.rs
```diff
@@ -262,7 +262,13 @@ fn analyze_function_params(
 
 fn analyze_function_body(body: ast::FunctionBody, fresher: &mut Fresher) -> Result<Block> {
     match body {
-        ast::FunctionBody::Expr(e) => Ok(Block::Return(analyze_expression(e, fresher)?)),
+        ast::FunctionBody::Expr(expr) => {
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, I did pull and rebase after merging those changes. I can do it again for safekeeping.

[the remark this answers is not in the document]
fchikwekwe: I also rebased an earlier version of these changes onto this feature branch a few commits ago so that I could start using those changes on this branch. Maybe that's why its showing up like this.

semantic/semantic.fbs
```diff
@@ -110,26 +112,26 @@ union Assignment {
...
-  UnaryExpression,
-  BooleanLiteral,
-  DateTimeLiteral,
-  DurationLiteral,
-  FloatLiteral,
-  IntegerLiteral,
-  StringLiteral,
-  RegexpLiteral,
-  UnsignedIntegerLiteral,
+    StringExpression,
```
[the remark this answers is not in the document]
fchikwekwe: I think that there might have been some automatic formatting here? That or I copy and pasted this from somewhere. I'll fix that. I think Flatbuffers convention is two space indentation.

libflux/src/flux/semantic/flatbuffers/tests.rs
wolffcm: Thinking this over, I don't think you need to run type inference in this test.  There are already unit tests for serializing types, so what you have here is fine.  I think we will want to run type inference as part of the roundtrip testing though.

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1150 @@
...
+        });
+        true
+    }
+
+    fn done(&mut self, node: Rc<walk::Node<'_>>) {
+        let mut v = &mut *self.inner.borrow_mut();
+        if let Some(_) = &v.err {
+            return;
+        }
+        let node = &*node;
```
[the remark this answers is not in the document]
fchikwekwe: Hmm yeah, I think this was the result of small changes while debugging. I'll try removing it.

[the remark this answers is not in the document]
fchikwekwe: Okay, so it looks like removing this also removes the reference counter around the node and breaks things a little bit. I'm not super sure why that is, but I'll leave it in for now and try to play around with in in my next PR.

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: Yeah I'll file a bug for this. I played around with fixing it. Doesn't seem like a ton of work, but it does cascade into a few other bugs a bit.

fchikwekwe: https://github.com/influxdata/flux/issues/2292

libflux/src/flux/semantic/flatbuffers/mod.rs
```diff
@@ -1,3 +1,1150 @@
...
+                    string.as_union_value(),
+                    fbsemantic::Expression::StringLiteral,
+                ))
+            }
+
+            walk::Node::DurationLit(dur_lit) => {
+                let mut dur_vec: Vec<WIPOffset<fbsemantic::Duration>> =
+                    Vec::new(); 
+                let magnitude = match dur_lit.value.num_nanoseconds() {
+                    Some(mag) => mag,
```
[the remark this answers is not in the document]
fchikwekwe: Here it is: https://github.com/influxdata/flux/issues/2290

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: Yeah, that is true. This node does get visited so that the linked list it represents can also be traversed.

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: yeah, let me remove it.

semantic/semantic.fbs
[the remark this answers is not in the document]
fchikwekwe: Hmm...locally I don't have the weird whitespace stuff that's seen here. I'll try and figure out why its appearing this way.

libflux/src/flux/semantic/flatbuffers/mod.rs
rockstar: _Should_ this return an `Option<Self>`? Is there a reason why the interface should/shouldn't be homogeneous?

libflux/src/flux/semantic/flatbuffers/mod.rs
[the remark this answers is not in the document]
rockstar: I'd be curious to know how it messes with the reference count, mostly out of curiosity with the `Rc` data structure (I've already mentioned my lack of familiarity with it). Does the `&*` signal some sort of op overload that calls `Deref::deref`? What's the error that we get? Maybe there's a more explicit call in `Rc` that might make this operation more explicit? For example, instead of calling `clone` on an `Rc` instance, `Rc::clone` seems preferred, because it makes the reference counting requirement more explicit; maybe there's a better way to express what is required here.

libflux/src/flux/semantic/flatbuffers/mod.rs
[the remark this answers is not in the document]
rockstar: Can we put that bug in the code itself?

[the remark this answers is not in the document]
fchikwekwe: Like add it as a comment? Sure.

libflux/src/flux/semantic/flatbuffers/mod.rs
fchikwekwe: Hmm...I'm curious about that as well. I think I'm going to go ahead and merge this as is, but I can easily update it in my next PR. Is the idea here that we just need to remove the option?

libflux/src/flux/semantic/flatbuffers/mod.rs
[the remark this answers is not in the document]
fchikwekwe: @rockstar I think fixing this would be a good opportunity for us to pair together. I'll hit you up and see if we can set up a time.

PR #2193, opened 2019-11-22
feat(libflux/semantic): update semantic Flatbuffers schema to match rust semantic graph more closely

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


semantic/semantic.fbs
```diff
@@ -155,24 +155,14 @@ table ArrayExpression {
 table FunctionExpression {
     loc:SourceLocation;
     defaults:ObjectExpression;
```
wolffcm: I think in Rust the function expression does not have a `defaults` field, they are attached to the parameter instead.

wolffcm: Also, don't you want to put `params:[FunctionParameter]` in here?

semantic/semantic.fbs
```diff
@@ -155,24 +155,14 @@ table ArrayExpression {
...
-    loc:SourceLocation;
-    parameters:FunctionParameters;
-    body:Block;
-}
-
-table FunctionParameters {
-    loc:SourceLocation;
-    list:[FunctionParameter];
-    pipe:Identifier;
+    Block:Block;
```
wolffcm: [nit] the field names in the rest of this file are lower case.

semantic/semantic.fbs
fchikwekwe: Yeah, I rebased and some of my changes were dropped. I can fix that.

semantic/semantic.fbs
```diff
@@ -267,11 +252,6 @@ table IdentifierExpression   {
     name:string;
 }
 
-table Identifier {
-    loc:SourceLocation;
-    name:string;
-}
-
```
jlapacik: Why was `Identifier` removed? `Identifier` and `IdentifierExpression` are distinct concepts. For instance, `IdentifierExpression` is an actual expression and therefore has a type associated with it. The `Identifier` associated with an object property or a function parameter however does not.

wolffcm: @jlapacik Yes, you are right.  I didn't catch this when I reviewed.  I'll make another PR that fixes this.

PR #2200, opened 2019-11-26
feat(libflux): create utility program for building libflux

jsternberg:
The utility program will build libflux in the go build cache and then
print out command line options to be used with a compiler. This matches
the same interface as `pkg-config` and can be used in a similar way.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,4 @@
-# Top level Makefile for the entire project
-#
-# This Makefile encodes the "go generate" prerequeisites ensuring that the proper tooling is installed and
+# This Makefile encodes the "go generate" prerequisites ensuring that the proper tooling is installed and
 # that the generate steps are executed when their prerequeisites files change.
 #
 # This Makefile follows a few conventions:
@@ -43,7 +41,8 @@ GENERATED_TARGETS = \
 	semantic/internal/fbsemantic \
 	libflux/src/ast/flatbuffers/ast_generated.rs \
 	libflux/src/semantic/flatbuffers/semantic_generated.rs \
-	libflux/scanner.c
+	libflux/scanner.c \
+	libflux/go/libflux/flux.h
 
 generate: $(GENERATED_TARGETS)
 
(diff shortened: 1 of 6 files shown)


jsternberg:
Some caveats I wanted to add to this discussion. I think this PR is probably good enough for getting us started, but it's not my favorite and there are other methods that I hadn't considered at the time I wrote this that I would like to experiment with in the future.

One is that it would be possible to have a utility that generates `.pc` files for `pkg-config` and then we would be able to use go's built in functionality. But, that would require a separate build step to put the `.pc` file and build it somewhere else. An advantage of this method is that you can directly feed this as part of a makefile. That may actually be a downside though. I don't know yet at the moment.

Another alternative is that we could write a `pkg-config` pass through that supports the same options, catches any references to `flux`, does the same things as this, and passes through any extra libraries to the real `pkg-config`. Then you would set the environment variable to customize the `pkg-config` binary path during the build process and you would use go's builtin support for `pkg-config`. I really like this idea, but I also don't want to delay actually integrating libflux integration.

I put this in internal because I do not consider it a user-facing script. There will likely be a better place than internal for this at some point since it is a dev-facing script, but I also don't want to signal that this is the final version because I personally don't like the solution and I am not sure how well it is going to work in practice. I would like us to have more experience integrating this before we say this is the official way to embed libflux into your go library.

jsternberg:
Can you explain to me what the placeholder is? I didn't understand that.

Makefile
```diff
@@ -79,6 +78,9 @@ libflux: libflux/target/debug/libflux.a
 libflux/target/debug/libflux.a:
 	cd libflux && $(CARGO) build $(CARGO_ARGS)
 
+libflux/go/libflux/flux.h: libflux/include/influxdata/flux.h
+	$(GO_GENERATE) ./libflux/go/libflux
```
yzhang1991: I think I missed some context here, why did you add this extra generate step?

affo: The reason is far from  being trivial.

`flux.h` was in an `include` dir by itself, we needed it to be alongside `.go` files to get it exported if vendored when building (because that's how `go mod vendor` works like that). So, we added this other generation step that copies over `flux.h` in a dir where there are `.go` files. We decided to copy, and not move it, because having an `include` dir states out clearly which are the headers of the lib.

Hope this clears things out

PR #2204, opened 2019-11-28
fix(e2e-tests): fix e2e usage write test so that its query is properly pushed down

affo:
The original version of the queries used a function invocation that broke being pushed down to storage.

This is a workaround to make everything work.

FYI @ischolten heads up if the queries look good to you.


ischolten:
@affo should we wait to merge this until the billing queries are updated?

jsternberg:
I'm going to merge this now as it's blocking a release.

PR #2209, opened 2019-12-02
refactor(libflux): into workspace

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR makes the libflux crate into a workspace with a single member.


jlapacik:
That is right @wolffcm

PR #2212, opened 2019-12-02
feat(stdlib/experimental) Add timeout to http.get and return response…

nisley:
Add an optional timeout parameter to http.get and return response headers as an object.

### DIFF
--- a/stdlib/experimental/http/get_test.go
+++ b/stdlib/experimental/http/get_test.go
@@ -110,7 +110,7 @@ func TestGet_Timeout(t *testing.T) {
 	script := fmt.Sprintf(`
 import "experimental/http"
 
-resp = http.get(url:"%s/path/a/b/c", timeout: 4)
+resp = http.get(url:"%s/path/a/b/c", timeout: 4s)
 `, ts.URL)
 
 	ctx := flux.NewDefaultDependencies().Inject(context.Background())
--- a/stdlib/experimental/http/get_test.go
+++ b/stdlib/experimental/http/get_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux"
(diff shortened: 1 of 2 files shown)


stdlib/experimental/http/http_experimental.go
```diff
@@ -13,22 +14,27 @@ import (
...
+const defaultTimeout = int64(30)
+
 // http get mirrors the http post originally completed for alerts & notifications
 var get = values.NewFunction(
 	"get",
 	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
 		Parameters: map[string]semantic.PolyType{
 			"url":     semantic.String,
 			"headers": semantic.Tvar(1),
+			"timeout": semantic.Int,
```
nathanielc: Can you use a duration type instead of an integer? That way user can specify `4s` or `500ms` in which ever units works best for their use case.

stdlib/experimental/http/http_experimental.go
```diff
@@ -94,25 +114,38 @@ var get = values.NewFunction(
...
+		return values.NewObjectWithValues(map[string]values.Value{
+			"statusCode": values.NewInt(int64(statusCode)),
+			"headers":    headers,
+			"body":       values.NewBytes(body)}), nil
 
 	},
 	true, // get has side-effects
 )
 
+func HeaderToArray(header http.Header) (headerObj values.Object) {
```
nathanielc: nit, can you rename this function to `headerToObject`? Using lower case means its not a public function.

stdlib/experimental/http/http_experimental.go
nisley: Done!

stdlib/experimental/http/http_experimental.go
nisley: Done!

stdlib/experimental/http/get_test.go
```diff
@@ -92,3 +93,32 @@ http.get(url:"http://127.1.1.1/path/a/b/c", headers: {x:"a",y:"b",z:"c"})
...
+		}
+		time.Sleep(time.Second * 8)
+		w.WriteHeader(204)
+	}))
+	defer ts.Close()
+
+	script := fmt.Sprintf(`
+import "experimental/http"
+
+resp = http.get(url:"%s/path/a/b/c", timeout: 4s)
```
nathanielc: Can you make this timeout something really small? That way the test still runs fast. Something like `10ms` or maybe `50ms`. It will timeout really fast but that is ok for the test.

nisley: Great idea. Done.

PR #2219, opened 2019-12-03
refactor(libflux): rename the directory for the flux crate to flux

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jlapacik:
You'll have to update the Makefile

jsternberg:
Updated the Makefile. I also noticed that cargo was being executed from the subdirectory rather than the top level. I figured out how to run cargo from the `libflux` directory and changed the invocation to do that.

Issue #45, opened 2018-09-19
Filter does not drop empty tables when it is not pushed down to the storage engine


jsternberg:
The storage engine will not include empty tables when they are filtered out because of a tag, but the filter function inside of flux just leaves the table empty so you end up with different results depending on whether that optimization happened or not.

nathanielc:
We need to specify how to treat empty tables:

Should empty tables be filtered?
Should they be filtered only if its filtering on the group key?

How could we return empty tables from storage if no data exists for a range or because of a _value filter?

The answer seems to depend on how filter interacts with the group key.

aanthony1243:
this came up in discussion today.  One need for empty tables is so that `count()`  gets a correct number where a user will want to know if a group key has 0 results.  This is probably the most prominent case, but any aggregator will generate a similar discussion.  Empty tables mean something when you aggregate.

juliusv:
This is hitting me when filtering, then aggregating, and then processing the result further. I end up getting a lot of tables with `null` values in the second step, which breaks further processing (e.g. I want to apply another `filter()` on the aggregated result to only keep aggregations with a certain `_value`, but having `null` values in filter conditions results in an error).

juliusv:
Counter to my comment above, keeping empty windows can also be useful in some situations of my PromQL->Flux transpilation work. When implementing Flux versions of PromQL functions such as [`time()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#time) or [`absent()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#absent), there may be no underlying data to work with, but I still require one output row per original window, even if it's empty. Especially for `time()` I would want to create windows that aren't even based on any selection from a datastore at all. If that's not possible, I could still generate windows manually in the corresponding custom Flux functions, but it would be nice to reuse `window()` for windowing in those cases.

jsternberg:
Fixed by #2278.

PR #2220, opened 2019-12-03
docs(spec): update the spec to clarify filter's behavior with empty tables

jsternberg:
The `filter` transformation has been changed so it will drop empty
tables by default. This matches with the influxdb behavior for pushing
down a filter to the storage engine. If the user wants to keep the empty
tables, they can switch the empty behavior to keep the table.

Related to #45.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -2106,12 +2106,16 @@ Filter applies a predicate function to each input record.
 Only the records for which the predicate evaluates to _true_ will be included in the output tables.
 One output table is produced for each input table.
 The output tables will have the same schema as their corresponding input tables.
+If `onEmpty` is set to `drop`, then empty tables will be dropped.
+If `onEmpty` is set to `keep`, then empty tables will be output to the next transformation.
+The default is to drop empty tables.
 
 Filter has the following properties:
 
-| Name | Type                | Description                                                                                        |
-| ---- | ----                | -----------                                                                                        |
-| fn   | (r: record) -> bool | Fn is a predicate function. Records which evaluate to true, will be included in the output tables. |
+| Name    | Type                | Description                                                                                        |
+| ----    | ----                | -----------                                                                                        |
+| fn      | (r: record) -> bool | Fn is a predicate function. Records which evaluate to true, will be included in the output tables. |
+| onEmpty | string              | The behavior for empty tables. This can be `keep` or `drop`. The default is `drop`.
 
 _NOTE_: make sure that `fn`'s parameter names match the ones specified above (see [why](#Transformations)).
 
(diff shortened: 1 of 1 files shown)


wolffcm:
Does this in fact match with storage's behavior?  My understanding is that storage will prune empty tables if the predicate is on a series key, but that empty tables _will_ be produced if the predicate is on `_value`... do I have this wrong?

jsternberg:
@stuartcarnie can you comment on the `_value` predicate? I didn't think about that.

If that's the case, it seems we're either going to have to do that pruning ourselves or get storage to do it. Or it can be part of the reader code that influxdb has for reading from the storage engine.

The reason why I think that drop should be the default is mostly a question of "what is going to be the most common?" and "what should be the most common for efficiency?" and it seems to me like dropping fits that description better than keeping the empty tables. But, if we want to switch `keep` to be the default, that's fine too but it will have performance implications.

docs/SPEC.md
```diff
@@ -2125,6 +2128,14 @@ from(bucket:"telegraf/autogen")
                 r.service == "app-server")
 ```
 
+Example:
+
+```
+from(bucket:"telegraf/autogen")
+    |> range(start:-12h)
+    |> filter(fn: (r) => r._measurement == "cpu", empty: "keep")
```
nathanielc: What does `keep` mean? For example given the above query, should data be returned with `_measurement` equal to `mem`? It would be an empty table so should it be kept or dropped? 

Does keep mean: only keep tables with a group key that matches the predicate but could be empty? What if a predicate function applies to columns outside the group key, does that mean no empty tables would ever be kept?

jsternberg: It was my opinion that the group key didn't matter. It had to do with the length of the table. If the table was empty, it was dropped. If it wasn't, then it was kept.

I'm not 100% sure I understand the question, but the above would return the full table if `_measurement` was equal to `cpu` and it would return an empty table for any series where `_measurement` was equal to anything else. So you would still get tables for all of the other series, but they would be empty.

nathanielc: > So you would still get tables for all of the other series, but they would be empty.

This is not what storage does and seems like the wrong behavior. Does the user really want all series in their bucket when they set `empty: 'keep'`? That seems unlikely and not useful.

jsternberg: Yes that is not what storage does. The default is proposed as `drop` which is the behavior that storage has if I understand correctly. @wolffcm did make a good point that this may not be true for value comparisons though which I hadn't thought of.

At the moment, the push down filter that influxdb uses has `empty: "drop"` as its behavior (potentially sans-value comparisons) and the flux filter has `empty: "keep"` as its behavior. This proposal is to make `drop` the default and to have the option to specify the behavior to allow for counting empty windows.

PR #2227, opened 2019-12-04
refactor(tableutil): modify the stream implementation

jsternberg:
The stream implementation now takes the key and columns when creating
the stream instead of using the first column reader. This makes it
easier to write the functionality and it makes it easier to construct an
empty table by just not sending anything. It is also possible to now
repeatedly send nothing and still get an empty table.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/execute/tableutil/stream.go
+++ b/internal/execute/tableutil/stream.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"sync/atomic"
 
+	"github.com/apache/arrow/go/arrow/array"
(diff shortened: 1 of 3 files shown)


jsternberg:
For a bit more context, I have been experimenting with the API for this while trying to work on optimizing functions. When I attempted to use this with filter, I found it difficult to use because I didn't know how to say "well it might be empty?"

I found this interface to be easier to use and a bit more direct.

PR #2228, opened 2019-12-04
feat(semantic/builtins): add types for some builtins

wolffcm:
This adds Flux types for some of the stdlib builtins (Flux functions and constants that are defined in Go code).  The unit test ensures that the type expressions all parse to polytypes correctly.

Not all of the builtins are here, I will file followup issues for what remains.

The `modf` function returns an object with a field named `int`, so I had to extend the type expression parser to allow quoted identifiers to resolve conflicts with keywords.

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -73,13 +73,13 @@ pub fn builtins() -> Builtins<'static> {
                  }),
                  "addDuration" => Node::Builtin("forall [] (d: duration, to: time) -> time"),
                  "subDuration" => Node::Builtin("forall [] (d: duration, from: time) -> time"),
-                 "group" => Node::Builtin("forall [t0, t1] (<-tables: [t0], mode: string, columns: [string]) -> t1"),
+                 "group" => Node::Builtin("forall [t0] (<-tables: [t0], mode: string, columns: [string]) -> [t0]"),
                  "objectKeys" => Node::Builtin("forall [t0] (o: t0) -> [string]"),
                  "set" => Node::Builtin("forall [t0, t1, t2] (<-tables: t0, o: t1) -> t2"),
                  // must specify exactly one of bucket, bucketID
                  // must specify exactly one of org, orgID
                  // if host is specified, token must be too.
-                 "to" => Node::Builtin("forall [t0] (?bucket: string, ?bucketID: string, ?org: string, ?orgID: string, ?host: string, ?token: string) -> t0"),
+                 "to" => Node::Builtin("forall [t0] (<-tables: [t0], ?bucket: string, ?bucketID: string, ?org: string, ?orgID: string, ?host: string, ?token: string) -> [t0]"),
             }),
             "generate" => Node::Package(maplit::hashmap! {
                 "from" => Node::Builtin("forall [] (start: time, stop: time, count: int, fn: (n: int) -> int) -> [{ _start: time | _stop: time | _time: time | _value:int }]"),
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -1,0 +1,338 @@
+use maplit::hashmap;
+use std::collections::hash_map;
+use std::collections::HashMap;
+use std::iter::Iterator;
+use std::vec::Vec;
+
+pub struct Builtins<'a> {
+    pkgs: HashMap<&'a str, Node<'a>>,
+}
+
+impl<'a> Builtins<'a> {
+    fn iter(&'a self) -> NodeIterator {
+        return NodeIterator::new(self);
+    }
+}
+
+enum Node<'a> {
+    Package(HashMap<&'a str, Node<'a>>),
+    Builtin(&'a str),
+}
+
+pub fn builtins() -> Builtins<'static> {
+    Builtins {
+        pkgs: hashmap! {
+            "csv" => Node::Package(maplit::hashmap! {
+                // This is a "provide exactly one argument" function
+                "from" => Node::Builtin("forall [t0] (?csv: string, ?file: string) -> [t0]"),
+            }),
+            "date" => Node::Package(maplit::hashmap! {
+                 "second" => Node::Builtin("forall [] (t: time) -> int"),
+                 "minute" => Node::Builtin("forall [] (t: time) -> int"),
+                 "hour" => Node::Builtin("forall [] (t: time) -> int"),
+                 "weekDay" => Node::Builtin("forall [] (t: time) -> int"),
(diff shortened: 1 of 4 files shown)


libflux/src/flux/semantic/builtins.rs
```diff
@@ -0,0 +1,338 @@
...
+                "trunc" => Node::Builtin("forall [] (x: float) -> float"),
+                "y0" => Node::Builtin("forall [] (x: float) -> float"),
+                "y1" => Node::Builtin("forall [] (x: float) -> float"),
+
+                "float64bits" => Node::Builtin("forall [] (f: float) -> uint"),
+                "float64frombits" => Node::Builtin("forall [] (b: uint) -> float"),
+                "ilogb" => Node::Builtin("forall [] (x: float) -> int"),
+                "frexp" => Node::Builtin("forall [] (f: float) -> {frac: float | exp: int}"),
+                "lgamma" => Node::Builtin("forall [] (x: float) -> {lgamma: float | sign: int}"),
+                "modf" => Node::Builtin(r#"forall [] (f: float) -> {"int": float | frac: float}"#),
```
nathanielc: Would it be possible from context to parse this without the quotes? Meaning if we changed the lexer to not treat type names as keywords but as identifiers and then the parser can understand when an identifier is a type name or not? 

This works for now, but when we go to incorporate this type language into Flux itself it would be good if type names were not parsed as keywords.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -0,0 +1,338 @@
...
+                    "promqlMonth" => Node::Builtin("forall [] (timestamp: float) -> float"),
+                    "promHistogramQuantile" => Node::Builtin("forall [t0, t1] (<-tables: [t0], ?quantile: float, ?countColumn: string, ?upperBoundColumn: string, ?valueColumn: string) -> [t1]"),
+                    "resets" => Node::Builtin("forall [t0, t1] (<-tables: [{_value: float | t0}]) -> [{_value: float | t1}]"),
+                    "timestamp" => Node::Builtin("forall [t0] (<-tables: [{_value: float | t0}]) -> [{_value: float | t0}]"),
+                    "promqlYear" => Node::Builtin("forall [] (timestamp: float) -> float"),
+                    "join" => Node::Builtin("forall [t0, t1, t2] (left: [t0], right: [t1], fn: (left: t0, right: t1) -> t2) -> [t2]"),
+                }),
+            }),
+            "json" => Node::Package(maplit::hashmap! {
+                "encode" => Node::Builtin("forall [t0] (v: t0) -> bytes"),
```
jlapacik: Does json.encode work on any input type or just records?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -0,0 +1,338 @@
...
+                 // must specify exactly one of bucket, bucketID
+                 // must specify exactly one of org, orgID
+                 // if host is specified, token must be too.
+                 "to" => Node::Builtin("forall [t0] (?bucket: string, ?bucketID: string, ?org: string, ?orgID: string, ?host: string, ?token: string) -> t0"),
+            }),
+            "generate" => Node::Package(maplit::hashmap! {
+                "from" => Node::Builtin("forall [] (start: time, stop: time, count: int, fn: (n: int) -> int) -> [{ _start: time | _stop: time | _time: time | _value:int }]"),
+            }),
+            "http" => Node::Package(maplit::hashmap! {
+                "post" => Node::Builtin("forall [t0] (url: string, ?headers: t0, ?data: bytes) -> int"),
```
jlapacik: It seems like the `headers` field should be an object. If so, we'll need to create an issue for expressing that constraint.

libflux/src/flux/semantic/builtins.rs
wolffcm: It might mean removing the distinction between keywords and identifiers in the type expression grammar.  If we ever allow users to name their own types in Flux, then we would probably need to remove this distinction anyway.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -0,0 +1,338 @@
...
+                 }),
+                 "addDuration" => Node::Builtin("forall [] (d: duration, to: time) -> time"),
+                 "subDuration" => Node::Builtin("forall [] (d: duration, from: time) -> time"),
+                 "group" => Node::Builtin("forall [t0, t1] (<-tables: [t0], mode: string, columns: [string]) -> t1"),
+                 "objectKeys" => Node::Builtin("forall [t0] (o: t0) -> [string]"),
+                 "set" => Node::Builtin("forall [t0, t1, t2] (<-tables: t0, o: t1) -> t2"),
+                 // must specify exactly one of bucket, bucketID
+                 // must specify exactly one of org, orgID
+                 // if host is specified, token must be too.
+                 "to" => Node::Builtin("forall [t0] (?bucket: string, ?bucketID: string, ?org: string, ?orgID: string, ?host: string, ?token: string) -> t0"),
```
jlapacik: Is this missing a `tables` parameter?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -0,0 +1,338 @@
...
+                             ?valueColumns: [string]
+                         ) -> [t1]
+                     "#),
+                 }),
+                 "prometheus" => Node::Package(maplit::hashmap! {
+                     "scrape" => Node::Builtin("forall [t0] (url: string) -> [t0]"),
+                 }),
+                 "addDuration" => Node::Builtin("forall [] (d: duration, to: time) -> time"),
+                 "subDuration" => Node::Builtin("forall [] (d: duration, from: time) -> time"),
+                 "group" => Node::Builtin("forall [t0, t1] (<-tables: [t0], mode: string, columns: [string]) -> t1"),
```
jlapacik: Group doesn't modify its input records right? Should this be `forall [t0] (<-tables: [t0], ... ) -> t0`?

libflux/src/flux/semantic/builtins.rs
wolffcm: It appears to work for any input type.

libflux/src/flux/semantic/builtins.rs
wolffcm: Yes, thanks for catching that.

libflux/src/flux/semantic/builtins.rs
wolffcm: Yes, you are right.

libflux/src/flux/semantic/builtins.rs
wolffcm: I filed this issue: https://github.com/influxdata/flux/issues/2231

Issue #2211, opened 2019-12-02
Optimize filter for non-group key filters


jsternberg:
When a filter is used with a table and only references parts of the group key, it is optimized to run the filter once and pass through the table without any copying. Filter was not optimized for when the filter was referencing things that weren't part of the group key so it uses an inefficient table builder.

Optimize filter for memory usage in the case of a function that references something not part of the group key.

fchikwekwe:
@jsternberg What does trying to optimize filter entail? I have not worked on any of these memory usage issues, so I'm a little in the dark here.

jsternberg:
Filter is currently optimized in the scenario where all of the keys referenced in the function are part of the group key. That part will be as optimized as possible because there's no way to get more performant than not allocating anything.

If that situation isn't true, such as `filter(fn: (r) => r._value > 0)` where `_value` is not part of the group key, it will use the `ColListTableBuilder` to produce a table using the old method. This is the scenario I am thinking about optimizing.

I think the optimization should be pretty simple. If I were to rewrite it, I would do the following:

1. Instead of building the table on the fly, produce a bitmap the size of the column reader. Evaluate each row and set the bit on the row if it passes.
2. Use the number of set bits to resize the array so that memory allocation only happens once.
3. Copy over the values to the new array for the value columns.
4. Use a slice for the group key columns.
5. Send to the passthrough dataset since filter cannot regroup.

Additionally, I think it would be good if we decided what to do with filter and whether to drop empty tables or not. It is something we have to resolve sooner rather than later, but this probably isn't the issue for it.

jsternberg:
I gave this an initial go and got some improvement, but I ended up finding that the allocations was just the low hanging fruit. Filter is _very_ slow and it might be worth evaluating why as I believe this covers code from map too. I am also not convinced that the libflux work will resolve the issue as, based on my testing, I did not find that type inference was the blocker. It seems that the compilation cache does its job and while type inference could potentially be the blocker for smaller datasets or ones with high cardinality, a workload that has a small number of series and a lot of points causes some substantial problems.

I have uploaded the torch graph for cpu usage inside of a zip file.

[torch.svg.zip](https://github.com/influxdata/flux/files/3922369/torch.svg.zip)

This flamegraph indicates to me that a lot of time is spent in methods that would seemingly be trivial. The `Set` method takes a substantial amount of time for setting the value on a record and validating the compiled function seems to be pretty bad.

PR #2232, opened 2019-12-04
feat(stdlib): improve filter performance when filtering on values

jsternberg:
This improves filter's performance when the function will evaluate
values that are not part of the group key.

Related to #2211.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
Here's a performance difference:

```
jsternberg:[~/g/s/g/i/flux] $ (feat/optimize-filter-values $=) benchcmp old.txt new.txt
benchmark                          old ns/op       new ns/op       delta
BenchmarkFilter_Values/1000-12     73667911676     59678115452     -18.99%

benchmark                          old allocs     new allocs     delta
BenchmarkFilter_Values/1000-12     1089074700     905505262      -16.86%

benchmark                          old bytes       new bytes       delta
BenchmarkFilter_Values/1000-12     29404196856     21521477224     -26.81%
```

The main bulk of the slowness appears to be when calling the function so this change primarily gets rid of the problems with filter itself.

stdlib/universe/filter_test.go
```diff
@@ -1015,3 +1017,59 @@ func TestFilter_Process(t *testing.T) {
...
+			func(id execute.DatasetID, alloc *memory.Allocator) (execute.Transformation, execute.Dataset) {
+				t, d, err := universe.NewFilterTransformation(context.Background(), spec, id, alloc)
+				if err != nil {
+					b.Fatal(err)
+				}
+				return t, d
+			},
+		)
+	}
+}
```
fchikwekwe: Helpful. I havent done much benchmarking in Go, so its good to have an example.

PR #2234, opened 2019-12-04
feat: libstd crate

jlapacik:
This PR defines a flatbuffer schema for a type environment and creates a new `libstd` crate for typing the flux standard library.

Note the typing of the standard lib is not yet implemented. Also note that the `Importer` struct was turned into a trait with the type `Environment` implementing said trait. `libstd` serializes both the prelude and the rest of the stdlib as type environments.


libflux/src/flux/semantic/import.rs
```diff
@@ -0,0 +1,5 @@
+pub trait Importer<T> {
```
nathanielc: Why is Importer parametric on the type it imports? It looks like all uses are for the `PolyType` type. Is that correct? In that case does it make sense to change this to be more specific?

libflux/src/flux/semantic/mod.rs
```diff
@@ -29,15 +29,21 @@ use crate::semantic::analyze::Result as AnalysisResult;
...
 
 impl From<InferError> for SemanticError {
     fn from(err: InferError) -> SemanticError {
         err.msg.clone()
     }
 }
 
+struct TypeImporter {}
+
+impl Importer<PolyType> for TypeImporter {}
```
wolffcm: Can you unpack this for me?  I don't follow how you can say `TypeImporter` implements `Importer` if it doesn't define an `import` method.

wolffcm: Oh, never mind, I just figured out it has a default implementation that always returns `None`.

semantic/semantic.fbs
```diff
@@ -1,5 +1,14 @@
 namespace fbsemantic;
 
+table TypeEnvironment {
```
wolffcm: The schema has a `root_type` declaration at the bottom:
```
root_type Package;
```
I'm not sure if it will be possible to create a flatbuffer that has something else at the root.  I'm curious if you've tried this?

libflux/src/flux/semantic/flatbuffers/types.rs
```diff
@@ -204,6 +225,39 @@ where
...
+    let ty = build_polytype(builder, assignment.1);
+    fb::TypeAssignment::create(
+        builder,
+        &fb::TypeAssignmentArgs {
+            id: Some(id),
+            ty: Some(ty),
+        },
+    )
+}
+
```
wolffcm: It would be good to have a round-trip test for the type env.

libflux/src/flux/semantic/mod.rs
wolffcm: So it's this type that will eventually be replaced with the importer for the stdlib, right?

libflux/src/libstd/lib.rs
```diff
@@ -0,0 +1,14 @@
+use flux::semantic::env::Environment;
+use flux::semantic::flatbuffers::semantic_generated::fbsemantic as fb;
+
+use flatbuffers;
+
+pub fn prelude() -> Option<Environment> {
+    let buf = include_bytes!(concat!(env!("OUT_DIR"), "/prelude.data"));
```
wolffcm: So `include_bytes!` will put the contents of `${OUT_DIR}/prelude.data` directly into this rust source file?

libflux/src/libstd/build.rs
```diff
@@ -0,0 +1,24 @@
...
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+fn prelude() -> Vec<u8> {
+    Vec::new()
+}
+
+fn stdlib() -> Vec<u8> {
+    Vec::new()
```
wolffcm: So I guess it's in these two functions that type inference of the stdlib is going to occur?

libflux/src/flux/semantic/mod.rs
jlapacik: Yes that's correct.

semantic/semantic.fbs
jlapacik: Yes it's possible via the `get_root` function:

    let buf = include_bytes!(concat!(env!("OUT_DIR"), "/prelude.data"));
    flatbuffers::get_root::<fb::TypeEnvironment>(buf).into()

That's why we can serialize and deserialize `PolyType`s independent of the semantic node types.

libflux/src/libstd/lib.rs
jlapacik: Yes correct.

libflux/src/libstd/build.rs
jlapacik: Right.

libflux/src/flux/semantic/flatbuffers/types.rs
jlapacik: 👍 I'll include one.

jlapacik:
@wolffcm everything you stated is correct. As for actually performing type inference over the stdlib, that is the next piece and I'm still working out the details.

PR #2241, opened 2019-12-05
feat(semantic): add Numeric and Row kind constraints

wolffcm:
This adds two new kind constraints to type inference:
- `Numeric` for int, uint and float.  There are a few transformations where we would error if a certain column like `_value` is not numeric.
- `Row` to indicate a TVar must be a Row type.

I also updated the builtin types that were already merged to reflect the changes.

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -9,7 +9,7 @@ pub struct Builtins<'a> {
 }
 
 impl<'a> Builtins<'a> {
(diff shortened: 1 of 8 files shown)


jlapacik:
You just have to update a couple of the type inference test cases now that `uints` are Subtractable.

wolffcm:
Yeah, I was just looking into that.  uints should be binary subtractable, but I don't think they should be unary subtractable.  I'm thinking I'll rollback that part of the change and file an issue.

Issue #1660, opened 2019-08-08
Remove bucketID and orgID parameters from to()


jpacik:
Currently `to()` takes parameters `bucket` and `bucketID` that are specified as mutually exclusive. That is, you cannot call `to()` with both parameters. This cannot be supported by the type system. Instead `to()` should take a single bucket parameter representing a bucket name, and there should be a separate function for converting a bucket ID to its associated name. Similarly for the parameters `org` and `orgID`.

See influxdata/influxdb#14594.

nathanielc:
I like the idea that @jsternberg had in slack, use string type for bucket names and integer type for bucketIDs using the hex syntax of defining the bucket ID.

Examples:

```
from(bucket:"foo") // find bucket by name
from(bucket: 0x03cbbf8eb3a9e000) // find bucket by id
```

This means we need to allow for specialized functions which I don't we can support yet. Or we need to add a special kind to the type system which seems to specific a solution. But the user experience is good and so worth exploring how we could implement it.

wolffcm:
There is a similar problem with the `host` and `token` parameters: if `host` is specified, then `token` must be also.  Maybe this is okay to catch at runtime though?

wolffcm:
Also, a similar issue exists with `experimental.to()`.

jlapacik:
`host` and `token` could also be specified together as an object `{host: string | token: string}`?

PR #2244, opened 2019-12-05
feat(semantic/builtins): add types for some universe builtins

wolffcm:
This is another part of work that was started here: https://github.com/influxdata/flux/pull/2228

This includes the builtins in the `universe` package, from letters I through Z.

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -24,6 +24,7 @@ pub fn builtins() -> Builtins<'static> {
         pkgs: hashmap! {
             "csv" => Node::Package(maplit::hashmap! {
                 // This is a "provide exactly one argument" function
+                // https://github.com/influxdata/flux/issues/2249
                 "from" => Node::Builtin("forall [t0] where t0: Row (?csv: string, ?file: string) -> [t0]"),
             }),
             "date" => Node::Package(maplit::hashmap! {
@@ -85,6 +86,7 @@ pub fn builtins() -> Builtins<'static> {
                  // must specify exactly one of bucket, bucketID
                  // must specify exactly one of org, orgID
                  // if host is specified, token must be too.
+                 // https://github.com/influxdata/flux/issues/1660
                  "to" => Node::Builtin("forall [t0] where t0: Row (<-tables: [t0], ?bucket: string, ?bucketID: string, ?org: string, ?orgID: string, ?host: string, ?token: string) -> [t0]"),
             }),
             "generate" => Node::Package(maplit::hashmap! {
@@ -101,6 +103,7 @@ pub fn builtins() -> Builtins<'static> {
                     }),
                     "v1" => Node::Package(maplit::hashmap! {
                         // exactly one of json and file must be specified
+                        // https://github.com/influxdata/flux/issues/2250
                         "json" => Node::Builtin("forall [t0] where t0: Row (?json: string, ?file: string) -> [t0]"),
                         "databases" => Node::Builtin(r#"
                             forall [] () -> {
@@ -114,9 +117,11 @@ pub fn builtins() -> Builtins<'static> {
                         "#),
                     }),
                     // This is a one-or-the-other parameters function
+                    // https://github.com/influxdata/flux/issues/1659
                     "from" => Node::Builtin("forall [t0, t1] (?bucket: string, ?bucketID: string) -> [{_measurement: string | _field: string | _time: time | _value: t0 | t1}]"),
                     // exactly one of (bucket, bucketID) must be specified
                     // exactly one of (org, orgID) must be specified
+                    // https://github.com/influxdata/flux/issues/1660
                     "to" => Node::Builtin(r#"
                         forall [t0, t1] where t0: Row, t1: Row (
                             <-tables: [t0],
@@ -177,7 +182,6 @@ pub fn builtins() -> Builtins<'static> {
                         ?valueColumns: [string]
                     ) -> [t0]"#),
             }),
-
             "math" => Node::Package(maplit::hashmap! {
                 "pi" => Node::Builtin("forall [] float"),
                 "e" => Node::Builtin("forall [] float"),
@@ -295,6 +299,7 @@ pub fn builtins() -> Builtins<'static> {
                 "assertEquals" => Node::Builtin("forall [t0] (name: string, got: [t0], want: [t0]) -> [t0]"),
                 "assertEmpty" => Node::Builtin("forall [t0] () -> [t0]"),
                 "diff" => Node::Builtin("forall [t0] (got: [t0], want: [t0], ?verbose: bool) -> [{_diff: string | t0}]"),
+            }),
             "universe" => Node::Package(maplit::hashmap! {
                 "inf" => Node::Builtin("forall [] duration"),
                 "int" => Node::Builtin("forall [t0] (v: t0) -> int"),
@@ -315,6 +320,7 @@ pub fn builtins() -> Builtins<'static> {
                 "#),
                 // This function would almost have input/output types that match, but:
                 // input column may start as int, uint or float, and always ends up as float.
+                // https://github.com/influxdata/flux/issues/2252
                 "kaufmansAMA" => Node::Builtin(r#"
                     forall [t0, t1] where t0: Row, t1: Row (
                         <-tables: [t0],
@@ -323,6 +329,7 @@ pub fn builtins() -> Builtins<'static> {
                     ) -> [t1]
                 "#),
                 // either column list or predicate must be provided
+                // https://github.com/influxdata/flux/issues/2248
                 "keep" => Node::Builtin(r#"
                     forall [t0, t1] where t0: Row, t1: Row (
                         <-tables: [t0],
@@ -330,13 +337,11 @@ pub fn builtins() -> Builtins<'static> {
                         ?fn: (column: string) -> bool
                     ) -> [t1]
                 "#),
-                // the "fn" parameter to keyValues is suspicous: https://github.com/influxdata/flux/issues/2235
                 "keyValues" => Node::Builtin(r#"
-                    forall [t0, t1, t2, t3] where t0: Row, t3: Row (
+                    forall [t0, t1, t2] where t0: Row, t2: Row (
                         <-tables: [t0],
-                        ?keyColumns: [string],
-                        ?fn: t1
-                    ) -> [{_key: string | _value: t2 | t3}]
+                        ?keyColumns: [string]
+                    ) -> [{_key: string | _value: t1 | t2}]
                 "#),
                 "keys" => Node::Builtin(r#"
                     forall [t0, t1] where t0: Row, t1: Row (
@@ -363,6 +368,8 @@ pub fn builtins() -> Builtins<'static> {
                         ?infinity: bool
                     ) -> [float]
                 "#),
+                // Note: mergeKey parameter could be removed from map once the transpiler is updated:
+                // https://github.com/influxdata/flux/issues/816
                 "map" => Node::Builtin("forall [t0, t1] (<-tables: [t0], fn: (r: t0) -> t1, ?mergeKey: bool) -> [t1]"),
                 "max" => Node::Builtin("forall [t0] where t0: Row (<-tables: [t0], ?column: string) -> [t0]"),
                 "mean" => Node::Builtin(r#"
@@ -397,9 +404,11 @@ pub fn builtins() -> Builtins<'static> {
                     ) -> [t0]
                 "#),
                 // start and stop should be able to constrained to time or duration with a kind constraint:
-                // https://github.com/influxdata/flux/issues/2243
+                //   https://github.com/influxdata/flux/issues/2243
+                // Also, we should remove the column arguments so we can reuse t0 in the return type:
+                //   https://github.com/influxdata/flux/issues/2253
                 "range" => Node::Builtin(r#"
-                    forall [t0, t1] where t0: Row, t3: Row (
+                    forall [t0, t1, t2, t3] where t0: Row, t3: Row (
                         <-tables: [t0],
                         start: t1,
                         ?stop: t2,
@@ -408,6 +417,8 @@ pub fn builtins() -> Builtins<'static> {
                         ?stopColumn: string
                     ) -> [t3]
                 "#),
+                // This function could be updated to get better type inference:
+                //   https://github.com/influxdata/flux/issues/2254
                 "reduce" => Node::Builtin(r#"
                     forall [t0, t1, t2] where t0: Row, t1: Row, t2: Row (
                         <-tables: [t0],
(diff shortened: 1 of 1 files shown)


libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
                 "splitRegexp" => Node::Builtin("forall [] (r: regexp, v: string, i: int) -> [string]"),
                 "getString" => Node::Builtin("forall [] (r: regexp) -> string"),
             }),
+            "universe" => Node::Package(maplit::hashmap! {
+                "inf" => Node::Builtin("forall [] duration"),
+                "int" => Node::Builtin("forall [t0] (v: t0) -> int"),
```
jlapacik: We expect to get a runtime error if we can't parse the type that's passed in as an integer, correct?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "#),
+                "join" => Node::Builtin(r#"
+                    forall [t0, t1] where t0: Row, t1: Row (
+                        <-tables: t0,
+                        ?method: string,
+                        ?on: [string]
+                    ) -> [t1]
+                "#),
+                // This function would almost have input/output types that match, but:
+                // input column may start as int, uint or float, and always ends up as float.
```
rockstar: I appreciate this comment.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "kaufmansAMA" => Node::Builtin(r#"
```
jlapacik: Could you create an issue for this? It seems like low hanging fruit to update this function signature so that it can be correctly typed.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                    ) -> [t1]
+                "#),
+                // the "fn" parameter to keyValues is suspicous: https://github.com/influxdata/flux/issues/2235
+                "keyValues" => Node::Builtin(r#"
+                    forall [t0, t1, t2, t3] where t0: Row, t3: Row (
+                        <-tables: [t0],
+                        ?keyColumns: [string],
+                        ?fn: t1
+                    ) -> [{_key: string | _value: t2 | t3}]
+                "#),
```
jlapacik: Per the issue you linked, can we just remove `fn` from the type signature?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                // either column list or predicate must be provided
+                "keep" => Node::Builtin(r#"
+                    forall [t0, t1] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        ?columns: [string],
+                        ?fn: (column: string) -> bool
+                    ) -> [t1]
+                "#),
+                // the "fn" parameter to keyValues is suspicous: https://github.com/influxdata/flux/issues/2235
```
rockstar: Is this a `TODO` or an `XXX`? Do we have a convention for documenting things in code that are either things that need to be done or technical debt accounting?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                    ) -> [t2]
+                "#),
+                "relativeStrengthIndex" => Node::Builtin(r#"
+                    forall [t0, t1] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        n: int,
+                        ?columns: [string]
+                    ) -> [t1]
+                "#),
+                // Either fn or columns should be specified
```
rockstar: Is it worth trying to find a way to annotate this?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "#),
+                "logarithmicBins" => Node::Builtin(r#"
+                    forall [] (
+                        start: float,
+                        factor: float,
+                        count: int,
+                        ?infinity: bool
+                    ) -> [float]
+                "#),
+                "map" => Node::Builtin("forall [t0, t1] (<-tables: [t0], fn: (r: t0) -> t1, ?mergeKey: bool) -> [t1]"),
```
jlapacik: I noticed you're not constraining the type variables to be `Row`s. Is that intentional?

libflux/src/flux/semantic/builtins.rs
wolffcm: I don't know that we have a formal convention.  Usually it's either a `TODO(<username>)` comment, or an issue.  In this case I made an issue for it.

I should just address the issue, because it's a really simple fix.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "range" => Node::Builtin(r#"
+                    forall [t0, t1] where t0: Row, t3: Row (
+                        <-tables: [t0],
+                        start: t1,
+                        ?stop: t2,
+                        ?timeColumn: string,
+                        ?startColumn: string,
+                        ?stopColumn: string
+                    ) -> [t3]
+                "#),
```
jlapacik: `range` should add `_start` and `_stop` labels to the output rows. We should create an issue to remove the parameters `timeColumn`, `startColumn`, and `stopColumn`.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "reduce" => Node::Builtin(r#"
+                    forall [t0, t1, t2] where t0: Row, t1: Row, t2: Row (
+                        <-tables: [t0],
+                        fn: (r: t0, accumulator: t1) -> t1,
+                        identity: t1
+                    ) -> [t2]
+                "#),
```
jlapacik: Shouldn't this return `[t1]`?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                    forall [t0] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        ?every: duration,
+                        ?period: duration,
+                        ?offset: duration,
+                        ?timeColumn: string,
+                        ?startColumn: string,
+                        ?stopColumn: string,
+                        ?createEmpty: bool
+                    ) -> [t1]
```
jlapacik: Why does window produce different output rows?

libflux/src/flux/semantic/builtins.rs
```diff
@@ -275,6 +275,267 @@ pub fn builtins() -> Builtins<'static> {
...
+                "#),
+                "true" => Node::Builtin("forall [] bool"),
+                "uint" => Node::Builtin("forall [t0] (v: t0) -> uint"),
+                // This is suspicous: each input table stream is not restricted to the same schema.
+                // What's the right signature?
+                "union" => Node::Builtin(r#"
+                    forall [t0] where t0: Row, t1: Row (
+                        tables: [[t0]]
+                    ) -> [t1]
+                "#),
```
jlapacik: I think they should be the same schema. From the type checker's point of view, each record type looks like this:

    { ... known labels ... | t0 } // where t0 is a row variable representing unknown labels

So if there is a contradiction on the known label types, type inference will error. But as long as there aren't any inconsistencies for the known label types, type inference won't care.

libflux/src/flux/semantic/builtins.rs
wolffcm: There are several functions that have this problem, and I think probably we'll need to change the functions in some way to make them friendlier to our type system, rather than increasing its complexity.

Josh filed a similar issue here: https://github.com/influxdata/flux/issues/1660

I think I'll file an umbrella issue for the "mutually exclusive parameters" problem, and document all the examples I can find.

libflux/src/flux/semantic/builtins.rs
wolffcm: Yes, that's right.  In that case the current signature is ok, yes?

jlapacik: Yeah just making sure.

libflux/src/flux/semantic/builtins.rs
jlapacik: Thank you.

libflux/src/flux/semantic/builtins.rs
wolffcm: Sure, I'll just remove it here, and save removing it from the Go code to a separate PR linked to that issue.

libflux/src/flux/semantic/builtins.rs
wolffcm: Yes.  In cases where we have a `column` or `columns` parameter, I left in the `Row` constraint.  But in this case we just have a `mergeKey` parameter, so I was on the fence about it.  It seems like moving forward we'll want to use this one on more than just table streams.

We can remove the `mergeKey` parameter, right?  Since we can now use `with` syntax, there's no reason to use it anymore I think.

libflux/src/flux/semantic/builtins.rs
wolffcm: Ok, I'll file an issue and mention it here in a comment.

libflux/src/flux/semantic/builtins.rs
wolffcm: Sadly, no.  This function will clobber the value columns unless they are explicitly added to the accumulator by the user, so it's tough to constrain the output type.

There's lots of examples where, if we could model the group key with a type variable, that we could get a better result from inference.

libflux/src/flux/semantic/builtins.rs
wolffcm: In this case it's only because we add the `startColumn` and the `stopColumn` to the output if they don't already exist.

So do you think we should be removing most of the `column` arguments from our functions and just expect the defaults?  Stuff like this really hurts type inference.

libflux/src/flux/semantic/builtins.rs
wolffcm: Ok, so the type is fine as-is then, that's great.  I'll update the comment.

libflux/src/flux/semantic/builtins.rs
jlapacik: I see. I just read the spec. Now that we have the `with` syntax for extending records, I think this function signature should be updated. Specifically I don't see any need to implicitly drop columns. This could be a very powerful function that many transformations could be written in terms of.

libflux/src/flux/semantic/builtins.rs
jlapacik: I think we can even return `[t0]` as each table will be constrained to have the same schema.

libflux/src/flux/semantic/builtins.rs
jlapacik: Ah that's right. In that case we should either define window to expect a `_start` and `_stop` column, or just add them as `range` does. And yes, also like range, we should remove the `column` arguments and specify the expected labels via the type signature. We should probably create an umbrella issue for all such identifiers.

libflux/src/flux/semantic/builtins.rs
jlapacik: Good catch. Yes we can remove the mergeKey parameter.

libflux/src/flux/semantic/builtins.rs
wolffcm: There weren't too many so I just filed separate issues for each instance.  Probably they should each be handled individually.

libflux/src/flux/semantic/builtins.rs
wolffcm: Oddly, it looks like it was already removed back in June but has reappeared somehow
https://github.com/influxdata/flux/pull/1416/files#diff-7abc02a1cd9fb8d4c37207d081429b3e

libflux/src/flux/semantic/builtins.rs
wolffcm: Well, we'll need to first change the current behavior of `reduce` so that it behaves more like `map`.  As it is now, the accumulator fields are appended to the group key to produce the output table.  This makes it hard to infer the output type, but also ensures that it's a narrow transformation.

I'll file an issue that describes the tradeoff.

libflux/src/flux/semantic/builtins.rs
jlapacik: I see. It looks like the transpiler uses it. Perhaps it's safe to remove right now, but I'm not sure. I'm fine waiting to remove it when we fix the transpiler.

libflux/src/flux/semantic/builtins.rs
jlapacik: Thanks, that would be great. We can discuss it more on the issue, but I'm not concerned about it possibly losing its "narrowness". Whether a transformation is narrow or wide is a physical property and the planner should choose the best implementation.

libflux/src/flux/semantic/builtins.rs
wolffcm: I'm a little confused by this, how about this example:
```
// type is [{host: string | t0}]:
ts0 = from(bucket: "foo") |> range(start: -5m) |> filter(fn: (r) => r.host == "bar")
// type is [{region: string | t0}]:
ts1 = from(bucket: "foo") |> range(start: -5m) |> filter(fn: (r) => r.region == "europe")

union(tables: [ts0, ts1])
```

Does the element type of the `tables` argument to `union` then become `{host: string | region: string | t0}`?  And does that mean that we assume that both `ts0` and `ts1` have a `region` and `host` field?  Because it's not necessarily the case.

wolffcm: (all of these `t0`s are logically different in this example)

libflux/src/flux/semantic/builtins.rs
wolffcm: Ok, I reopened the issue, and will link it here.

libflux/src/flux/semantic/builtins.rs
jlapacik: That's right. So we will infer that the records from each table passed into `union` have the type:

    {host: string | region: string | t0}

You are also correct that the tables from `ts0` might not have a `region` column. This is fine because strings, like all column types, are nullable and so if the `region` column is referenced after the union, any null values will be passed through without error.

libflux/src/flux/semantic/builtins.rs
wolffcm: I created this issue: https://github.com/influxdata/flux/issues/2255

There are so many functions like this, I'm not goint to link to it for every example of this.

jlapacik: That works for me. As long as we have an issue for it. Thank you.

libflux/src/flux/semantic/builtins.rs
wolffcm: Ok, that's great, I forgot about the `Nullable` kind.

wolffcm:
I've pushed another commit that annotates some of the more problematic types.

I also filed several issues for the resulting work, there's a lot of it.  I'm going to create a `Improve Flux Builtin Types` epic and put all the issues into it.

PR #2258, opened 2019-12-09
feat(values): reduce memory allocations for operations in values

### LINKED ISSUE #2211: Optimize filter for non-group key filters

jsternberg:
An object will now cache the type for an object and not recompute it as
long as the type doesn't change. This includes a change to `Set` which
will check the old value's type and the new value's type and it will not
force the type to be recomputed as long as the type doesn't change. This
helps to reduce memory allocations when using `Set` to change the value
inside of functions like `filter()` where the type doesn't change.
Previously, it was using a lot of memory to recompute the type over and
over again.

This also changes `values.NewBool` to return one of two memoized values
for true and false. This avoids an interface allocation each time that a
boolean is created.

Fixes #2211.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/values/bool.go
+++ b/values/bool.go
@@ -1,0 +1,22 @@
+package values
+
+import "github.com/influxdata/flux/semantic"
+
+var (
+	trueValue Value = value{
+		t: semantic.Bool,
+		v: true,
+	}
+	falseValue Value = value{
+		t: semantic.Bool,
+		v: false,
+	}
+)
+
+func NewBool(v bool) Value {
+	if v {
+		return trueValue
+	} else {
+		return falseValue
+	}
+}
--- a/values/object.go
+++ b/values/object.go
@@ -20,7 +20,9 @@ type object struct {
 	labels semantic.LabelSet
 	values []Value
 	ptyp   map[string]semantic.PolyType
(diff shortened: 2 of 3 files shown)


jsternberg:
Benchmark comparison:

```
jsternberg:[~/g/s/g/i/flux] $ (feat/optimize-object-set $) benchcmp old.txt new.txt
benchmark                          old ns/op       new ns/op       delta
BenchmarkFilter_Values/1000-12     70260034867     21914358279     -68.81%

benchmark                          old allocs     new allocs     delta
BenchmarkFilter_Values/1000-12     905505262      155743661      -82.80%

benchmark                          old bytes       new bytes      delta
BenchmarkFilter_Values/1000-12     21482642584     6985710944     -67.48%
```

There's still work to do here, but this will have to be good enough for now until I've got some more cycles for dealing with performance related stuff.

PR #2265, opened 2019-12-10
refactor(ast): add environment variable to use rust parser

wolffcm:
Also added metadata field to ast.File so we can see which parser was used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,12 @@ GO_ARGS=-tags '$(GO_TAGS)'
 
 # Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
+export GO_BUILD=env FLUX_PARSER_TYPE=rust GO111MODULE=on go build $(GO_ARGS)
+export GO_TEST=env FLUX_PARSER_TYPE=rust GO111MODULE=on go test $(GO_ARGS)
 export GO_TEST_FLAGS=
 # Do not add GO111MODULE=on to the call to go generate so it doesn't pollute the environment.
 export GO_GENERATE=go generate $(GO_ARGS)
-export GO_VET=env GO111MODULE=on go vet $(GO_ARGS)
+export GO_VET=env FLUX_PARSER_TYPE=rust GO111MODULE=on go vet $(GO_ARGS)
 export CARGO=cargo
 export CARGO_ARGS=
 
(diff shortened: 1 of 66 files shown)


wolffcm:
@rockstar Most of these changes are actually generated code.  There are a couple of files that `flatc` generates in Rust and Go, but most of them are the `_gen.go` files that contain the Flux standard library as Go struct literals, so we can export the standard library to `influxdb`.  There is a tool called `builtin` that generates these files during the `generated` step of the build.

Makefile
```diff
@@ -16,12 +16,12 @@ GO_ARGS=-tags '$(GO_TAGS)'
 
 # Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
+export GO_BUILD=env FLUX_PARSER_TYPE=rust GO111MODULE=on go build $(GO_ARGS)
```
yzhang1991: Question:

Why is this set at compile-time?

parser/libflux_parser.go
```diff
@@ -6,11 +6,16 @@ import (
 	"encoding/json"
 
 	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/internal/parser"
 	"github.com/influxdata/flux/internal/token"
 	"github.com/influxdata/flux/libflux/go/libflux"
 )
 
 func parseFile(f *token.File, src []byte) (*ast.File, error) {
+	if !useRustParser() {
```
yzhang1991: Another question is, do we want to call `os. Getenv()` every time we parse a file?

Makefile
wolffcm: For `go build` I think you're right, it won't have any affect.  Some of the `go` commands run the built executable (like `go test`) so I wanted to make sure that I preserve the existing behavior of using the Rust parser whenever we have a libflux build.

One thing I didn't do here was set the environment variable for the `generate` step, since it does not use the libflux build either.  So currently we still use the Go parser to generate the `_gen.go` files we use to export the standard library (just like before this PR merged).  But we could probably start using the Rust parser there.

parser/libflux_parser.go
wolffcm: Are you concerned about the performance?  It didn't occur to me that it would be more than reading a global variable, but I guess it is, but a factor of 1000:
https://dinolai.com/notes/golang/golang-benchmark-reading-config-from-os-env-and-file.html

I'l open another PR to cache the value of the environment variable after the first time we read it.

yzhang1991: Yes, my concern is around the performance. I think this is on a critical path.

PR #2266, opened 2019-12-10
feat(bootstrap): methods for type checking package dependencies

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This commit adds a new module `bootstrap` whose main function is `infer_pkg`. This function infers the types in a package/file. If said package/file imports other packages, it first recursively infers the types those imported packages, before type checking the original package.

Note, once all builtin identifiers have had their types defined, this function will be used by the `libstd` crate to infer the types of the entire standard library.

### DIFF
--- a/libflux/src/flux/semantic/bootstrap.rs
+++ b/libflux/src/flux/semantic/bootstrap.rs
@@ -1,0 +1,288 @@
+use std::collections::HashMap;
+use std::fs;
+use std::io;
+use std::path::PathBuf;
+
+use crate::ast;
+use crate::parser;
+use crate::semantic::analyze::analyze_file;
+use crate::semantic::env::Environment;
+use crate::semantic::fresh::Fresher;
+use crate::semantic::import::Importer;
+use crate::semantic::infer;
+use crate::semantic::infer::Constraints;
+use crate::semantic::nodes;
+use crate::semantic::nodes::infer_file;
+use crate::semantic::sub::Substitutable;
+use crate::semantic::types;
+use crate::semantic::types::{MonoType, PolyType, Property, Row};
+
+use walkdir::WalkDir;
+
+const PRELUDE: [&str; 2] = ["universe", "influxdata/influxdb"];
+
+type Error = String;
+
+impl From<nodes::Error> for Error {
+    fn from(err: nodes::Error) -> Error {
+        err.to_string()
+    }
+}
+
+impl From<types::Error> for Error {
+    fn from(err: types::Error) -> Error {
+        err.to_string()
+    }
+}
+
+#[allow(dead_code)]
+// Recursively parse all flux files within a directory.
+fn parse_flux_files(path: &str) -> io::Result<Vec<ast::File>> {
+    let mut files = Vec::new();
+    let entries = WalkDir::new(PathBuf::from(path))
+        .into_iter()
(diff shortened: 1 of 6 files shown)


libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -0,0 +1,288 @@
...
+
+impl From<types::Error> for Error {
+    fn from(err: types::Error) -> Error {
+        err.to_string()
+    }
+}
+
+#[allow(dead_code)]
+// Recursively parse all flux files within a directory.
+fn parse_flux_files(path: &str) -> io::Result<Vec<ast::File>> {
```
nathanielc: Should these methods live in the parser module?

libflux/src/flux/semantic/nodes.rs
```diff
@@ -415,11 +457,20 @@ pub struct BuiltinStmt {
...
+    fn infer<I: Importer>(
+        &self,
+        mut env: Environment,
+        importer: &I,
+    ) -> std::result::Result<Environment, Error> {
+        if let Some(ty) = importer.import(&self.id.name) {
+            env.add(self.id.name.clone(), ty.clone());
+            Ok(env)
+        } else {
+            Err(Error::undefined_builtin(&self.id.name))
```
wolffcm: Ok, cool, so the build will fail if we forget to add types for builtins.

libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -0,0 +1,290 @@
...
+
+type Error = String;
+
+impl From<nodes::Error> for Error {
+    fn from(err: nodes::Error) -> Error {
+        err.to_string()
+    }
+}
+
+impl From<types::Error> for Error {
```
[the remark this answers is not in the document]
jlapacik: I haven't found a document that lays this out, but the pattern I've adopted at least for the semantic module, is to define an error type for each submodule and then convert between errors using the `From` trait. This way I can propagate errors using the `?` operator. This seems very rust-like, but I have yet to find a definitive guide on the subject.

libflux/src/flux/semantic/nodes.rs
jlapacik: Yes.

libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -0,0 +1,290 @@
...
+    for import in &file.imports {
+        dependencies.push(&import.path.value[..]);
+    }
+    dependencies
+}
+
+// dependencies adds all of the packages that must be evaluated(type checked)
+// before a package can be evaluated(type checked) to the deps vector.
+fn dependencies<'a>(pkg: &'a str, deps: &mut Vec<&'a str>, files: &'a HashMap<String, ast::File>) {
+    if let Some(file) = files.get(pkg) {
```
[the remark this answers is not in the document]
jlapacik: Yes to be safe, I should be checking that. I'll update.

libflux/src/flux/semantic/bootstrap.rs
jlapacik: For right now, let's keep it as a utility method for typing the standard lib. And we can update it for user defined packages.

PR #2272, opened 2019-12-11
fix(semantic/analyze): update function block return statements to pro…

fchikwekwe:
Previously, the Rust semantic graph represented `ReturnStmt` nodes within `Block` nodes as Expressions. This is because in the AST a `Block` node can be represented as a list of statements or as an expression with the return value being the value of the expression. This PR makes all function blocks in the semantic graph a list of statements. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/flux/semantic/nodes.rs
+++ b/libflux/src/flux/semantic/nodes.rs
@@ -123,8 +123,8 @@ pub enum Assignment {
 impl Assignment {
     fn apply(self, sub: &Substitution) -> Self {
         match self {
-            Assignment::Variable(ass) => Assignment::Variable(ass.apply(&sub)),
-            Assignment::Member(ass) => Assignment::Member(ass.apply(&sub)),
+            Assignment::Variable(assign) => Assignment::Variable(assign.apply(&sub)),
+            Assignment::Member(assign) => Assignment::Member(assign.apply(&sub)),
         }
     }
 }
@@ -844,11 +844,13 @@ impl FunctionExpr {
 // must always have a return value. This means a function block is by
 // definition an expression.
 //
+// A function block is an expression that evaluates to the argument of
+// its terminating ReturnStmt.
 #[derive(Debug, PartialEq, Clone)]
 pub enum Block {
     Variable(Box<VariableAssgn>, Box<Block>),
     Expr(ExprStmt, Box<Block>),
-    Return(Expression),
+    Return(ReturnStmt),
 }
 
 impl Block {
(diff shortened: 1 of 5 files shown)


libflux/src/flux/semantic/flatbuffers/tests.rs
```diff
@@ -0,0 +1,807 @@
+extern crate flatbuffers;
```
wolffcm: I don't think you want to merge this yet, do you?

libflux/src/flux/semantic/nodes.rs
```diff
@@ -837,18 +837,11 @@ impl FunctionExpr {
     }
 }
 
-// Block represents a function block and is equivalent to a let-expression
```
wolffcm: I think this comment is still accurate.  It might be good to add "A function block is an expression that evaluates to the argument of its terminating `ReturnStmt`."

libflux/src/flux/semantic/walk/walk_mut.rs
```diff
@@ -229,7 +229,7 @@ impl<'a> NodeMut<'a> {
             NodeMut::Block(Block::Expr(ref mut estmt, _)) => {
                 mut_expr_loc(&mut estmt.expression, loc)
             }
-            NodeMut::Block(Block::Return(ref mut expr)) => mut_expr_loc(expr, loc),
+            NodeMut::Block(Block::Return(ref mut expr)) => mut_expr_loc(&mut expr.argument, loc),
```
wolffcm: Unless I'm missing something, I don't think we need to do anything here for any of the block variants.  The `Block` enum has a `loc()` method that just produces the location of it's statement, and the statement will be visited elsewhere in this enum.

So I think for Block this should be:
```rust
NodeMut::Block(_) => (),
```

libflux/src/flux/semantic/flatbuffers/tests.rs
```diff
@@ -0,0 +1,807 @@
...
+        files: f,
+    };
+    let mut pkg = match analyze(pkg) {
+        Ok(pkg) => pkg, 
+        Err(e) => {
+            assert!(false, e);
+            return;
+        }
+    }; 
+    let (vec, offset) = match super::serialize(&mut pkg) {
```
jlapacik: Where is this function defined? It doesn't look like it's been included yet it compiles.

fchikwekwe: whoops. This file shouldn't be in this PR.

libflux/src/flux/semantic/flatbuffers/tests.rs
fchikwekwe: Oops. Nope. Let me remove that.

libflux/src/flux/semantic/flatbuffers/tests.rs
wolffcm: I think it didn't compile because it wasn't actually connected to the module tree with a `mod tests` statement anywhere, so `cargo` just ignores it.

libflux/src/flux/semantic/walk/walk_mut.rs
fchikwekwe: Yeah, that does actually make sense. I'll update that.

fchikwekwe:
Made some updates to hopefully address the issues brought up.

PR #2273, opened 2019-12-12
feat: serialize flux standard library types as part of build process

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR serializes the flux prelude and standard library types as part of the build process. In order for this PR to build successfully, several fixes were required. They are:

1. Corrections to a few of the types given to the builtins https://github.com/influxdata/flux/pull/2273/commits/7078e24b39dfac335837317cf4d94dc9ed02aea6
2. Import aliases were not being added to the type environment https://github.com/influxdata/flux/pull/2273/commits/685c621705816e0091c19cec370b037ef9b52c25
3. Durations are not addable or subtractable, but `range` routinely takes inputs of the form `range(start: -1h)`. I added a new `Signed` kind for the unary `-` operator. Durations are signed but not addable. https://github.com/influxdata/flux/pull/2273/commits/f6c528eec8f34b29bfb0e2fc845e1479838e92de
4. The quantile function in the promql package mistakenly took an integer instead of a float https://github.com/influxdata/flux/pull/2273/commits/489ea1ff52667c9ad87b75f0e6b91c00341c9cc1
5. And finally a subtle bug where sometimes equivalent polytypes would not compare as equal https://github.com/influxdata/flux/pull/2273/commits/48c1391f060758d7ebc3b23646aed79790fbb16e


semantic/semantic.fbs
```diff
@@ -1,5 +1,9 @@
 namespace fbsemantic;
 
+table Fresher {
```
wolffcm: I guess you need this so when we do inference on Flux programs you know where to start creating new type variables, so you don't collide with the ones already in the stdlib?

libflux/src/flux/semantic/mod.rs
```diff
@@ -12,8 +12,8 @@ pub mod nodes;
 pub mod types;
 pub mod walk;
 
-#[cfg(test)]
-mod parser;
+#[allow(dead_code)]
```
wolffcm: I'm curious why you need to allow dead code here.  Is there something that will eventually be made public?  If that's the case, I think it's better to just add `pub` to make dead code reachable, rather than disable this warning for a whole module.

libflux/src/flux/semantic/nodes.rs
```diff
@@ -1603,137 +1606,11 @@ pub fn convert_duration(duration: &Vec<ast::Duration>) -> std::result::Result<Du
...
-    use crate::semantic::analyze::analyze_with;
-    use crate::semantic::import::Importer;
-    use crate::semantic::types::{MonoType, PolyType, Tvar};
+    use crate::semantic::types::{MonoType, Tvar};
     use crate::semantic::walk::{walk, Node};
     use maplit::hashmap;
-    use std::collections::HashMap;
     use std::rc::Rc;
 
-    struct TestImporter {
```
wolffcm: I guess all this testing code became redundant with the new code in bootstrap.rs?

libflux/src/flux/semantic/types.rs
```diff
@@ -47,9 +46,20 @@ impl PartialEq for PolyType {
         let mut f = Fresher::from(max + 1);
         let mut g = Fresher::from(max + 1);
 
-        self.clone()
-            .fresh(&mut f)
-            .equal(&poly.clone().fresh(&mut g))
+        let mut a = self.clone().fresh(&mut f, &mut HashMap::new());
+        let mut b = poly.clone().fresh(&mut g, &mut HashMap::new());
+
+        a.vars.sort();
```
wolffcm: I wonder if there's a way to ensure that we have a deterministic, canonical ordering of type variables within a polytype as soon as its created, so we don't need to do any sorting later on when we need to compare.  Curious about your thoughts on this.

libflux/src/libstd/build.rs
```diff
@@ -1,24 +1,62 @@
...
+    file.write_all(&buf)?;
+    Ok(())
+}
+
+fn main() -> Result<(), Error> {
+    let dir = path::PathBuf::from(env::var("OUT_DIR")?);
+
+    let (pre, lib, fresher) = bootstrap::infer_stdlib()?;
+
+    let path = dir.join("prelude.data");
```
wolffcm: Why does the prelude need to be serialized separately?  Isn't everything in there redundant with what's already in the standard lib?

libflux/src/libstd/build.rs
```diff
@@ -1,24 +1,62 @@
...
+    serialize(Environment::from(pre), fb::build_env, &path)?;
+
+    let path = dir.join("stdlib.data");
+    serialize(Environment::from(lib), fb::build_env, &path)?;
```
wolffcm: `stdlib.data`, `prelude.data` and `fresher.data` are all generated files.  We typically check generated files into source control.  I guess we can't generate them as part of the `generate` step?

Do you need to add a `.gitignore` file for these files, if they won't be checked in?

wolffcm: I guess if you had a binary crate in the workspace for creating the generated files, then you could build only that crate during the generate step, and generate the files.  Then they would be available to be consumed during the Rust build proper.  Would that be possible?  It would be nice to somehow make this conform with our existing build practices.

Also, I'm curious, how long does doing type inference on the standard library take?

libflux/src/libstd/lib.rs
```diff
@@ -8,7 +9,54 @@ pub fn prelude() -> Option<Environment> {
...
+    use flux::semantic::nodes::infer_file;
+
+    #[test]
+    fn deserialize_and_infer() {
+        let prelude = Environment::new(super::prelude().unwrap());
+        let imports = super::imports().unwrap();
+
+        let src = r#"
+            x = from(bucket: "b")
+                |> filter(fn: (r) => r.region == "west")
```
wolffcm: Ok, so I guess we need to fix `range` 😄

libflux/src/libstd/lib.rs
```diff
@@ -8,7 +9,54 @@ pub fn prelude() -> Option<Environment> {
...
+            r#"forall [t0, t1] where t0: Addable [{
+                _value: t0
+                    | _value: t0
+                    | _time: time
+                    | _measurement: string
+                    | _field: string
+                    | region: string
+                    | t1
+                    }]
+            "#,
```
wolffcm: Nice!

libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -6,36 +6,144 @@ use std::path::PathBuf;
...
+
+impl From<&str> for Error {
+    fn from(msg: &str) -> Error {
+        Error {
+            msg: msg.to_string(),
+        }
+    }
+}
+
+#[allow(clippy::type_complexity)]
```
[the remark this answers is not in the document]
jlapacik: I don't think it's the fact that I'm returning a 3-tuple. I don't think clippy would have a problem with `(int, int, int)`. I think it has something to do with the complexity of the types that make up the 3-tuple.

libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -322,9 +433,14 @@ mod tests {
             String::from("a") => parse_string("a.flux", a),
             String::from("b") => parse_string("b.flux", b),
         };
+        let want_err = dependencies("b", &files, Vec::new(), HashSet::new(), HashSet::new())
```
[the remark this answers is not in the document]
jlapacik: Yes you're right.

libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -248,7 +344,7 @@ mod tests {
     use crate::semantic::parser::parse;
 
     #[test]
-    fn infer_program() {
+    fn infer_program() -> Result<(), Error> {
```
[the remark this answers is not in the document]
jlapacik: Yes `main` functions as well as test functions can return results. And yes it will fail if it does not return `Ok(())`. Note the error type must implement `Debug` as that is what will be used to display output in the case of test failure. This allows one to use the `?` operator in `main` and test functions.

semantic/semantic.fbs
jlapacik: Yes.

libflux/src/flux/semantic/mod.rs
jlapacik: By labeling it as `pub` does that get rid of the dead code warning. I thought everything was being used, just across modules. I'll check again.

libflux/src/flux/semantic/nodes.rs
jlapacik: Yes.

libflux/src/flux/semantic/types.rs
jlapacik: We could use a btree to ensure that. This code will only ever be invoked during tests though. Is your concern about performance or regressions?

libflux/src/libstd/lib.rs
jlapacik: Ha, yes that would be nice.

libflux/src/libstd/build.rs
jlapacik: The prelude is composed of all identifiers that are defined in the `universe` and `influxdb` packages. These identifiers can be of any generic type. So the importer for the prelude associates identifier names with polytypes.

The importer for the standard library however associates import paths with polymorphic record (or row) types. A package can be thought of as an object. You use a member expression to access the identifiers defined in that package. That's why they're serialized separately.

libflux/src/libstd/build.rs
jlapacik: Those are binary files. The source files that they're derived from, i.e `builtins.rs` and the flux source files, are checked into source control though. Rust places generated files in the `target` directory when using the `OUT_DIR` environment variable, and my understanding is that the `target` directory is usually git-ignored, which it is in our case.

I am open to a better way though. This was just the first way that came to mind. Although I think we would still ignore those generated binary data files.

wolffcm: Ok, none of this should stop you from merging, but the way the generated scanner works is: it lives in `libflux/scanner.c` and then `flux/build.rs` copies it to the target directory.  The makefile will build `libflux/scanner.c` during the generate step only when it gets out of sync.  

Builds in Rust are already pretty slow, so it might save some time to not infer the stdlib+prelude if we don't have to.  But like I said before, I think this work can be deferred.

libflux/src/flux/semantic/types.rs
wolffcm: My concern was mostly performance.  Seeing an `eq` method that calls sort multiple times gives me pause.  If it's only used in tests, then I'm not that concerned.

libflux/src/libstd/build.rs
wolffcm: Ok, that makes sense to me.  Thanks.

libflux/src/libstd/build.rs
jlapacik: @wolffcm that makes a lot of sense. I can create an issue for this. So basically we only have to re-infer the stdlib, if one of the flux files changes or if `builtins.rs` changes. Is that what you're thinking?

wolffcm: Yes, exactly.  The rule in the makefile would look something like
```makefile
stdlib.data prelude.data fresher.data: $FLUX_FILES builtins.rs
        <command to infer stdlib and prelude>
```

nathanielc: @wolffcm What about if the implementation of algo w itself changes, i.e. we fix a bug? Do we need the makefile to understand all of the code that is part of the `flux` crate?

wolffcm: The `checkgenerate` target should catch any issues there, same as if we fixed a parser bug and the resulting `_gen.go` changes.

libflux/src/flux/semantic/nodes.rs
```diff
@@ -1603,137 +1606,11 @@ pub fn convert_duration(duration: &Vec<ast::Duration>) -> std::result::Result<Du
 mod tests {
     use super::*;
     use crate::ast;
-    use crate::parser::parse_string;
-    use crate::semantic::analyze::analyze_with;
-    use crate::semantic::import::Importer;
-    use crate::semantic::types::{MonoType, PolyType, Tvar};
+    use crate::semantic::types::{MonoType, Tvar};
```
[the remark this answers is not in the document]
jlapacik: Yes, in the `stdlib` crate.

libflux/src/flux/semantic/types.rs
```diff
@@ -80,44 +90,6 @@ impl MaxTvar for PolyType {
...
-    // Fresh takes a polytype and generates an equivalent polytype but with
-    // completely fresh type variables.
-    //
-    // Note in order to be sure that the retured polytype is equivalent to
-    // the one that was passed in, the fresher 'f' must generate type
-    // variables that do not exist in the given polytype. In order to ensure
-    // that this is the case, one should use a fresher that is instantiated
-    // with a type variable that is strictly greater than all other type
-    // variables in the given polytype.
-    //
```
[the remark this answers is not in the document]
jlapacik: Yes should have moved this comment as well.

PR #2274, opened 2019-12-12
fix(scanner): fix token location for scan_with_regex

wolffcm:
If a flux program has a stray `/` in a place where a regexp literal might appear, the scanner will scan to the end of the input looking for the closing `/`.  During scanning we keep track of where newlines appear so we can determine the line/column info for the end of the token returned by the scanner.

In this situation we were getting the location wrong because the token was just TOKEN_DIV, not a regexp literal.  I updated the logic in the Rust scanner to reset the pointer to the last newline for this case.

I also made two changes to how Go parses Flux so that the roundtrip test can now pass:
- if we cannot parse the RHS of a binary expr, we use the location info of the buffered token, so we can propagate location info up the AST
- I added logic to preserve the sortedness of the lines slice maintained by the scanner.  We need to do this since the lines slice can become unordered when unreading multi-line tokens.  This is addressing the same problem that the Rust changes address, but on the Go side.

### DIFF
--- a/libflux/src/flux/scanner/mod.rs
+++ b/libflux/src/flux/scanner/mod.rs
@@ -76,6 +76,16 @@ impl Scanner {
         self._scan(2)
     }
 
+    // unread will reset the Scanner to go back to the Scanner's location
+    // before the last scan_with_regex or scan call. If either of the scan_with_regex methods
+    // returned an EOF token, a call to unread will not unread the discarded whitespace.
+    // This method is a no-op if called multiple times.
+    pub fn unread(&mut self) {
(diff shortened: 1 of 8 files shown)


libflux/src/flux/scanner/mod.rs
```diff
@@ -76,6 +76,16 @@ impl Scanner {
         self._scan(2)
     }
 
+    // unread will reset the Scanner to go back to the Scanner's location
```
wolffcm: I didn't make any changes to this method, I just moved it to be closer to the other public methods.

PR #2277, opened 2019-12-13
refactor(semantic): add type field to semantic graph expressions

wolffcm:
This PR adds a `TypeOf` method to each of the `Expression` interface, and a `typ` field to each implementor of `Expression`.  When we deserialize from flatbuffers, the `typ` fields will be populated with the type produced by Rust type inference.

FlatBuffers does not provide a common interface so monotypes are encapsulated in a struct in the semantic package called `MonoType` that wraps the flatbuffers classes.

I also added partial support for a `String` method for MonoTypes, to get a feel for working with flatbuffers in Go.  My takeaway is that it's painful, and we'll probably want to implement wrapper structs for most of the FlatBuffers structs.


semantic/fb_types.go
```diff
@@ -0,0 +1,16 @@
...
+
+type fbTabler interface {
+	Init(buf []byte, i flatbuffers.UOffsetT)
+	Table() flatbuffers.Table
+}
+
+type FBMonoType struct {
+	mt  fbsemantic.MonoType
+	tbl fbTabler
+}
```
jlapacik: Would you be opposed to calling this `MonoType`? Also why do we need both an `fbTabler` and an `fbsemantic.MonoType`? Isn't `fbsemantic.MonoType` an `fbTabler`?

semantic/flatbuffers.go
```diff
@@ -556,3 +556,41 @@ func (p *FunctionParameter) FromBuf(fb *fbsemantic.FunctionParameter) error {
...
+	TypType() byte
+	Typ(obj *flatbuffers.Table) bool
+}
+
+// getMonoType produces an FBMonoType from the given FlatBuffers expression that has
+// a union "typ" field (which is all the different kinds of expressions).
+// To start with not all expressions may have types, so this will not return an error
+// if the type is missing---this shold change in the future.
+// TODO(cwolff): return an error if the type is missing
+func getMonoType(fbExpr fbTyper) (*FBMonoType, error) {
```
jlapacik: So `getMonoType` takes a flatbuffer as input, checks to see if the current index is the start of a MonoType, and if so, returns it? Do I have this right?

Also the `Typ` method on the interface ensures we only call this on nodes that have a `Typ`?

semantic/fb_types.go
wolffcm: In the Go generated code, `MonoType` is just a kind integer that describes which specific monotype something is:
https://github.com/influxdata/flux/blob/master/semantic/internal/fbsemantic/MonoType.go#L5-L14

In fact there is no interface generated by `flatc` that unites `Arr`, `Fun`, etc as the same kind of thing.  `fbTabler` defines any FlatBuffers type that is defined as a `table`.

Yes, I see that `MonoType` is not already taken in the semantic package, so I can rename it to `MonoType`.

wolffcm: I could have just omitted the `fbsemantic.MonoType` here, since we can use Go's runtime type casting to get at `Arr`, `Fun`, `Var`, etc.  But I thought it might be useful, maybe.

semantic/flatbuffers.go
wolffcm: The code in `fbgen/cmd/semantic.go` looks at each struct type in the Go `semantic` package that implements `Node` and if it has a field of type `*FBMonoType` then it will generate code in `flatbuffers_gen.go` that calls this function. All such fields are accessed as `TypType()` and `Typ()` in the FlatBuffer schema. If this were not the case, then the Go code would not compile.

Essentially, instances of `fbTyper` are all the structs generated by `flatc` that are expressions.  This method just extracts the type from the flatbuffers expression (sadly the only common interface for it is `fbTabler`), and returns it bundled with its `fbsemantic.MonoType` tag.

Also I still need to copy the byte slice here to ensure we don't rely on the original underlying flatbuffer hanging around.

wolffcm:
@jlapacik I wasn't very happy with how awkward it is to use flatbuffers generated code directly in Go, so I made a new package `semantic/types` and created a couple thin wrappers around the FlatBuffers structs to make them easier to use.

There is an example of using this in the `MonoType` and `PolyType` `Strings` methods.

I think this will make it much easier to replace the old types with the new ones, let me know what you think.

PR #2281, opened 2019-12-16
fix: expose tracing flag

yzhang1991:
There is a span created in influxdb too so we need to expose this flag at the flux package level.
I wonder if we could eliminate the `traceconfig` just put a private flag in `tracing.go` @jsternberg


jsternberg:
It's fine to remove the internal package if you want. I approved this anyway.

PR #2285, opened 2019-12-17
fix: eliminate fb copy

yzhang1991:


libflux/go/libflux/parser.go
```diff
@@ -76,5 +77,5 @@ func ParseIntoFbs(s string) *ast.Package {
 	defer C.free(unsafe.Pointer(ptr))
 
 	data := C.GoBytes(ptr.data, C.int(ptr.len))
-	return ast.Package{}.FromBuf(data)
+	return ast.Package{}.FromBuf(data, flatbuffers.UOffsetT(C.int(ptr.offset)))
```
fchikwekwe: What is this line of code doing? I see a couple of offsets here, but its hard to tell.

libflux/src/flux/lib.rs
```diff
@@ -89,13 +90,11 @@ pub unsafe extern "C" fn flux_parse_fb(src_ptr: *const c_char) -> *mut flux_buff
...
-                data: data.as_ptr(),
-                len: data.len(),
-            }))
-        }
+        Ok((vec, offset)) => Box::into_raw(Box::new(flux_buffer_t {
+            data: vec.as_ptr(),
+            offset,
+            len: vec.len(),
+        })),
         Err(_) => 1 as *mut flux_buffer_t,
```
wolffcm: I'm having trouble understanding why the original code worked okay.  Here's what I think the original code was doing:
- taking a slice of the `Vec` that trims the unused first portion
- Then we create a Box that allocates a `flux_buffer_t` on the heap, which contains a pointer to the interior of the `vec`
-  Then the function returns the `Box`.  I guess `Box` will silently convert to a raw pointer, and the caller takes ownership?
- It seems like memory for `vec` will be freed at the end of this function though?

libflux/go/libflux/parser.go
wolffcm: I think it's taking the offset field of `ptr`, typecasting it to a C integer (`C.int(...)`) and then typecasting that to a `flatbuffers.UOffsetT`.  I'm not sure why we need the intermediate cast to a C integer though.

libflux/src/flux/lib.rs
wolffcm: The new code also seems to have the problem that you are returning a pointer to a `flux_buffer_t` whose `data` field references memory that is owned by the Vec, that I think will be freed when it goes out of scope.

Does the `unsafe` qualifier at the beginning of the function mean you don't need `unsafe {...}` within the function body?

I think you are right that there is no copy here though, I was wrong about that.

Maybe the right way to do this is:
```rust
let data = Box::into_raw(vec.into_boxed_slice())
Box::into_raw(Box::new(flux_buffer_t {
            data,
            offset,
            len: vec.len(),
        }))
```
I'm not really sure about it though.  There is also `mem::forget` that could be helpful too.

libflux/go/libflux/analyze.go
```diff
@@ -20,20 +20,19 @@ type SemanticPkg struct {
 }
 
 // MarshalFB serializes the given semantic package into a flatbuffer.
-func (p *SemanticPkg) MarshalFB() ([]byte, error) {
+func (p *SemanticPkg) MarshalFB() ([]byte, uint, func(), error) {
```
wolffcm: I wonder if it would be better to return a pointer to a struct called `Buffer` or something that implements the `Freeable` interface, and use `SetFinalizer` similar to how `ASTPkg` and `SemanticPkg` do to ensure that it's always cleaned up, even if the caller forgets to do it.

`Buffer` could encapsulate the `offset` logic too.

Thoughts?

This would also make the return type of this function a bit simpler

libflux/go/libflux/bench_test.go
```diff
@@ -106,24 +109,25 @@ func ParseAndDeserializeJSON(fluxFile string) error {
 func ParseAndReturnFB(fluxFile string) error {
 	p := libflux.Parse(fluxFile)
```
wolffcm: I found that parsing in Rust takes a looong time, it may be because we do a lot of copying where we don't need to.  So I'm not that surprised that getting rid of copies does not improve things visibly.  Hopefully at some point that will change.

semantic/analyze_libflux.go
```diff
@@ -16,9 +16,10 @@ func AnalyzeSource(fluxSrc string) (*Package, error) {
 		return nil, err
 	}
 	defer sem.Free()
-	fb, err := sem.MarshalFB()
+	fb, offset, freeFn, err := sem.MarshalFB()
 	if err != nil {
 		return nil, err
 	}
-	return DeserializeFromFlatBuffer(fb)
+	defer freeFn()
```
wolffcm: On the current master, this will work fine.  On the `feat/use-algo-w` branch, when deserialize the semantic graph, it will contain a pointer into the flatbuffer for each expression and assignment statement.  We can't free the flatbuffer here since the graph that we return to the caller has pointers into it.  I think we will be stuck with leaving it to the caller to free the memory when it's done.  Curious to know your thoughts on this.

libflux/go/libflux/buffer.go
```diff
@@ -0,0 +1,26 @@
+// +build libflux
+
+package libflux
+
+// #cgo CFLAGS: -I.
+// #cgo LDFLAGS: -L. -llibstd
```
wolffcm: I think you want to use `-lflux` here instead of `-llibstd`, since the function you're using (`flux_free`) is implemented in the `flux` crate.  I assume that this compiled, I wonder why it did..

libflux/go/libflux/analyze.go
yzhang1991: Yeah, that will be a good idea.

semantic/analyze_libflux.go
yzhang1991: Can you explain why do we need to keep so many references to the flatbuffer at different offsets? Why do those expressions still need to refer to the original flatbuffer?

wolffcm: Yes, that's the idea.  The semantic graph will be deserialized but it will contain `MonoType` and `PolyType` structs that are thin wrappers around bytes in the original flatbuffer.  There is no Go-native deserialized format for types, i.e., Flux runtime execution will talk directly to the flatbuffers representation of types (via the thin wrapper).

Only `NativeVariableAssignments` have polytypes, and expressions have monotypes, e.g.:
https://github.com/influxdata/flux/blob/feat/use-algo-w/semantic/graph.go#L636

The PolyType and MonoType structs are defined here, they are the wrappers around the Go code generated by `flatc`:
https://github.com/influxdata/flux/blob/feat/use-algo-w/semantic/monotype.go#L14-L24
https://github.com/influxdata/flux/blob/feat/use-algo-w/semantic/polytype.go#L13-L17

I'm seeing now how this generates issues since the lifetime of the byte array needs to be explicitly managed in Go.  I'd be happy to pair on this if that would be helpful.

libflux/go/libflux/bench_test.go
```diff
@@ -88,11 +88,11 @@ func ParseReturnJSON(fluxFile string) error {
 func ParseAndDeserializeJSON(fluxFile string) error {
 	p := libflux.Parse(fluxFile)
 	defer p.Free()
-	bs, err := p.MarshalJSON()
```
yzhang1991: I left this test case not calling `buf.Free()` so we can test the finalizer.

wolffcm: Ok, I'm not sure if there's a way to test the finalizer in Go though, since my understanding is that the finalizer is not guaranteed to be called if the process exits.  Not sure there is a reliable way to do it.  I wonder if the `-race` flag does anything in this regard?

libflux/go/libflux/buffer.go
```diff
@@ -0,0 +1,45 @@
...
+func unwrapBuffer(buf C.struct_flux_buffer_t) *ManagedBuffer {
+	sh := new(reflect.SliceHeader)
+	sh.Data = uintptr(unsafe.Pointer(buf.data))
+	sh.Len = int(C.uint(buf.len))
+	sh.Cap = sh.Len
+	bs := *(*[]byte)(unsafe.Pointer(sh))
+	ret := &ManagedBuffer{
+		Buffer: bs,
+		Offset: flatbuffers.UOffsetT(uintptr(C.uint(buf.offset))),
+		freeFn: func() {
```
wolffcm: I wonder if this can just be a regular method on `ManagedBuffer`, that will call `C.flux_free` on `buf.data` and then set `buf.data` to zero?  That seems simpler and easier to follow.

libflux/go/libflux/buffer.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (b *ManagedBuffer) Free() {
+	if b.freeFn != nil {
+		b.freeFn()
+		b.freeFn = nil
+	}
+}
+
+func unwrapBuffer(buf C.struct_flux_buffer_t) *ManagedBuffer {
```
wolffcm: `newManagedBuffer` would be a nice name for this that would be a little more Go-idiomatic.

runtime.go
```diff
@@ -125,6 +125,7 @@ func (r *runtime) evalHandle(ctx context.Context, h *libflux.ASTPkg, opts ...Sco
 	// Execute the interpreter over the package.
 	itrp := interpreter.NewInterpreter(nil)
 	sideEffects, err := itrp.Eval(ctx, semPkg, scope, importer)
+	semPkg.Free()
```
wolffcm: I think `defer semPkg.Free()` would be appropriate here: if there is a panic somewhere during evaluation, we still want the free to happen.  Deferred function calls will still happen if a panic occurs.

semantic/flatbuffers.go
```diff
@@ -2,18 +2,28 @@ package semantic
...
-	p := &Package{}
+func DeserializeFromFlatBuffer(mbuf *libflux.ManagedBuffer) (*Package, error) {
+	// Cancel the finalizer on the managed buffer because it will be referenced by the
+	// semantic graph
+	runtime.SetFinalizer(mbuf, nil)
+	fbPkg := fbsemantic.GetRootAsPackage(mbuf.Buffer, mbuf.Offset)
+	p := &Package{
+		freeFn: mbuf.Free,
+	}
+	runtime.SetFinalizer(p, func(p *Package) {
```
wolffcm: I wonder, can you reuse the `libflux.Freeable` interface here?

semantic/flatbuffers_gen.go
```diff
@@ -525,26 +525,26 @@ func (rcv *OptionStatement) FromBuf(fb *fbsemantic.OptionStatement) error {
 	return nil
 }
 
-func (rcv *Package) FromBuf(fb *fbsemantic.Package) error {
+func (p *Package) FromBuf(fb *fbsemantic.Package) error {
```
wolffcm: This file is generated code, if there are no real changes here, it's probably better to roll them back so as not to fail `make checkgenerate`.

wolffcm:
I pushed a commit so that this will build against the latest algo-w.

There are still some issues here, though, the `promql` test `TestJoin` fails, and it seems to be because
- we use the finalizer on the semantic `Package` node to free the buffer returned by Rust
- the test helper `FunctionExpression` (defined [here](https://github.com/influxdata/flux/blob/feat/use-algo-w/execute/executetest/compile.go#L22)) returns a function expression and discards the `Package` node
- this means that the returned function expression contains types that reference a Rust buffer that has been deallocated

We could update `FunctionExpression()` to copy the buffer returned by Rust, but this is a more general problem.  It would be trivially easy to hit this by accident.  We should come up with another solution that let's us remain confident in the memory guarantees that Go provides for us.

PR #2287, opened 2019-12-18
feat(stdlib): implement onEmpty parameter for filter

### LINKED ISSUE #2278: Update filter to add onEmpty parameter
#2220

jsternberg:
The `filter()` function now accepts an `onEmpty` parameter that will
determine the behavior of filter when the filter causes all rows to be
filtered.

Fixes #2278.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/filter_keep_empty_test.flux
+++ b/stdlib/universe/filter_keep_empty_test.flux
@@ -1,0 +1,43 @@
+package universe_test
+
+import "testing"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,dateTime:RFC3339,long,string,string,string,string
+#group,false,false,false,false,true,true,true,true
+#default,_result,,,,,,,
+,result,table,_time,_value,_field,_measurement,host,name
+,,0,2018-05-22T19:53:26Z,15204688,io_time,diskio,host.local,disk0
+,,0,2018-05-22T19:53:36Z,15204894,io_time,diskio,host.local,disk0
+,,0,2018-05-22T19:53:46Z,15205102,io_time,diskio,host.local,disk0
+,,0,2018-05-22T19:53:56Z,15205226,io_time,diskio,host.local,disk0
+,,0,2018-05-22T19:54:06Z,15205499,io_time,diskio,host.local,disk0
+,,0,2018-05-22T19:54:16Z,15205755,io_time,diskio,host.local,disk0
+,,1,2018-05-22T19:53:26Z,648,io_time,diskio,host.local,disk2
+,,1,2018-05-22T19:53:36Z,648,io_time,diskio,host.local,disk2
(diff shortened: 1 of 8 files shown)


stdlib/universe/filter_keep_empty_test.flux
```diff
@@ -0,0 +1,43 @@
...
+#default,_result,,,,,,,,
+,result,table,_start,_stop,_measurement,_field,host,name,_value
+,,0,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,diskio,io_time,host.local,disk0,6
+,,1,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,diskio,io_time,host.local,disk2,0
+"
+
+t_filter_keep_empty = (table=<-) =>
+  table
+  |> range(start: 2018-05-22T19:53:26Z)
+  |> filter(fn: (r) => r["_value"] > 1000, onEmpty: "keep")
```
nathanielc: Can you add a test case that shows how you would use  a combination of filters to get the empty tables from a subset of the data? Meaning say I have data for both `mem` and `cpu` measurements. I want empty table but on the from the `mem` measurement.

jsternberg: Yes I think I can do that.

jsternberg: Done.

stdlib/universe/filter_mixed_empty_test.flux
```diff
@@ -0,0 +1,67 @@
...
+,result,table,_start,_stop,_measurement,_field,host,name,_value
+,,0,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,diskio,io_time,host.local,disk0,6
+,,1,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,diskio,io_time,host.local,disk2,0
+"
+
+t_filter_mixed_empty = (table=<-) =>
+  table
+  |> range(start: 2018-05-22T19:53:26Z)
+  |> filter(fn: (r) => r._measurement == "diskio")
+  |> filter(fn: (r) => r["_value"] > 1000, onEmpty: "keep")
```
nathanielc: Can you update this test case such that the output has an empty table? That way its a positive test that `keep` does what we expect.

jsternberg: I actually can't. Diff doesn't detect the difference between empty tables and non-existant tables correctly. It considers them the same. This is likely a bug in diff, but I didn't want to start touching that. That's the reason why I use `count()` because it is a way to get around diff being unable to do this.

nathanielc: OK, good to know, Can you create an issue fo the `diff` bug?

jsternberg: https://github.com/influxdata/flux/issues/2291

Issue #1985, opened 2019-10-14
Add Months to values.Duration struct


nathanielc:
The values.Duration struct only understands Nanoseconds, we need to update it to understand  months as an independent unit. 

This means updating all values.Duration and values.Time methods that consume a values.Duration type.

This change can ignore time zones.

Issue #2290, opened 2019-12-19
Fix DurationLiterals in Rust Semantic Graph


fchikwekwe:
Right now the Rust semantic graph's value is a single Duration. 

```pub struct DurationLit {
    pub loc: ast::SourceLocation,
    pub typ: MonoType, 
    pub value: Duration,
}
```

While the Rust AST and the Go Semantic graph and AST contain a vector of durations to accommodate the possibility of multiple Duration values in Flux. 

```pub struct DurationLit {
    pub base: BaseNode,
    pub values: Vec<Duration>,
}
```

The Duration and DurationLiteral representations do not have to reflect the ones on the Go side, but they do need to include months in addition to the nanoseconds currently stored. They also need to indicate whether the value is positive or negative. 

@jsternberg suggested following the example from `values/time.go`: 
```// Duration is a vector representing the duration unit components.
type Duration struct {
	// months is the number of months for the duration.
	// This must be a positive value.
	months int64
	// nsecs is the number of nanoseconds for the duration.
	// This must be a positive value.
	nsecs int64
	// negative indicates this duration is a negative value.
	negative bool
}
```

This issue is to: 
- Add a Duration struct to the Rust semantic graph
- Update the semantic graph Flatbuffers to this change (schema, serialization and deserialization) 
- Fix any bugs that this change causes
- Update tests accordingly

fchikwekwe:
Basically, we wanna bring the Rust semantic graph `DurationLiteral` nodes in line with the Rust AST, Go semantic and Go AST representations of DurationLiterals. Flatbuffers handles the differences right now, but it does that by storing a Duration value in nanoseconds only. We want to be able to store the multi-value durations that exist in Flux. `1mo2d3h4s` etc.

jlapacik:
What's the value in storing `1mo2d3h4s` in the semantic graph? Or I guess a better question would be, is there a duration value that can be represented in the AST but not the semantic graph? I'm just trying to get a better understanding on what is being fixed.

jsternberg:
To clarify, the only reason the Go semantic graph stores the individual units is because it was the easiest way to migrate the duration value. Once we get to the semantic graph, we don't really care about any units except for months and nanoseconds.

[...]

PR #2297, opened 2019-12-23
fix(semantic/libflux): fix durations in rust semantic graph

fchikwekwe:
Previously durations in the Rust semantic graph were only storing nanoseconds. This was insufficient because we cannot calculate months based on the number of nanoseconds in Flux as there is no consistent number of nanoseconds that equal a month (calendar durations are discussed here: https://github.com/influxdata/flux/issues/1985,  https://github.com/influxdata/flux/issues/2044). 

It was also problematic because it did not clearly store whether a value is positive or negative. 

This change create a Duration struct that stores months, nanoseconds and whether the value is negative. It also updates Flatbuffers and the analyze package to account for this change. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/src/flux/ast/mod.rs
```diff
@@ -1013,6 +1013,7 @@ pub struct Duration {
 // DurationLit represents the elapsed time between two instants as an
 // int64 nanosecond count with syntax of golang's time.Duration
 // TODO: this may be better as a class initialization
+// All magnitudes in Duration vector should have the same sign
```
rockstar: _should_ or _must_?

libflux/src/flux/semantic/analyze.rs
```diff
@@ -1267,7 +1271,11 @@ mod tests {
                                     value: Expression::Duration(DurationLit {
                                         loc: b.location.clone(),
                                         typ: type_info(),
-                                        value: chrono::Duration::minutes(10),
+                                        value: Duration {
+                                            months: -1,
+                                            nanoseconds: -50,
```
nathanielc: I expected these values to be positive as per the issue description https://github.com/influxdata/flux/issues/2290 Meaning that the `months` and `nanoseconds` values are always positive values and the `negative` field determines if they should be treated as negative.

libflux/src/flux/semantic/flatbuffers/semantic_generated.rs
```diff
@@ -12478,6 +12405,111 @@ pub mod fbsemantic {
...
+        #[inline]
+        fn follow(buf: &'a [u8], loc: usize) -> Self::Inner {
+            Self {
+                _tab: flatbuffers::Table { buf: buf, loc: loc },
+            }
+        }
+    }
+
+    impl<'a> Duration<'a> {
+        #[inline]
```
rockstar: _(I see this is mostly just moved code, but would like to ask this, just in case you know)_

Do we have a rule for when we expect inline attributes to be worth using? What heuristic do we use to balance compile time, binary size, and execution speed?

libflux/src/flux/semantic/nodes.rs
```diff
@@ -1545,13 +1544,23 @@ impl DateTimeLit {
     }
 }
 
+// DurationLit is a pair consisting of length of time and the unit of time measured.
+// It is the atomic unit from which all duration literals are composed.
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[serde(rename = "Duration")]
+pub struct Duration {
+    pub months: i64,
```
nathanielc: we should comment that these fields should always contain positive values

libflux/src/flux/semantic/nodes.rs
```diff
@@ -1578,29 +1588,39 @@ const MINUTES: i64 = SECONDS * 60;
...
-            _ => Err(format!("unrecognized magnitude for duration: {}", d.unit)),
-        })?;
-    Ok(Duration::nanoseconds(d))
+
+// The following durations have month base units
+const MONTHS: i64 = 1;
+const YEARS: i64 = MONTHS * 12;
+
+pub fn convert_duration(ast_dur: &[ast::Duration]) -> std::result::Result<Duration, String> {
+    let negative = ast_dur[0].magnitude.is_negative();
```
nathanielc: what if the list is empty or does it always have to be non-empty?

libflux/src/flux/ast/mod.rs
fchikwekwe: must! I'll change that :)

libflux/src/flux/semantic/nodes.rs
fchikwekwe: I think that the list does have to be non-empty since a duration value is required. Unless it is  possible to have an AST DurationLiteral without a Duration. I can add a check for an empty list here so that we can have a better error message in that case, as well as a test case that handles this.

libflux/src/flux/semantic/nodes.rs
fchikwekwe: Yes, I'll add a comment here. I also need to be taking the absolute value of the final nanoseconds and months values in `convert_duration()`. I'll update that.

libflux/src/flux/semantic/flatbuffers/semantic_generated.rs
fchikwekwe: Hmm...I'm not super familiar with this, so I'm not quite sure. I'd need to read up a bit on what inline does.

libflux/src/flux/semantic/analyze.rs
fchikwekwe: Yes, that makes sense. I'll make that change.

libflux/src/flux/semantic/flatbuffers/semantic_generated.rs
rockstar: I _completely_ missed that this was generated code. Please ignore me. I had thought that we were putting those annotations in.

fchikwekwe:
I made all the requested updates and this PR and checks in CI are passing. Gonna go ahead and merge.

PR #2302, opened 2020-01-02
fix(libflux): remove redundant clones found by a new version of clippy

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/src/flux/semantic/bootstrap.rs
```diff
@@ -376,12 +376,13 @@ mod tests {
                 String::from("x") => parse("forall [] int")?,
             }),
         };
+        let prelude: HashMap<String, PolyType> = HashMap::new();
```
wolffcm: You didn't change the signature of `infer_pkg` so you don't need to create a temporary here, right?

jsternberg: Yes. My mistake. I forgot to remove this.

PR #2307, opened 2020-01-02
feat(libflux/semantic): add source location to type inference errors

nathanielc:
Quick pass at adding source location to type inference errors. This will not catch all errors as the location information is lost when converting semantic nodes to constraints. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/flux/semantic/nodes.rs
+++ b/libflux/src/flux/semantic/nodes.rs
@@ -36,27 +36,8 @@ pub type Result = std::result::Result<(Environment, Constraints), Error>;
 
 #[derive(Debug)]
 pub struct Error {
+    pub loc: ast::SourceLocation,
     pub msg: String,
-}
-
-impl fmt::Display for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(&self.msg)
-    }
-}
-
-impl From<types::Error> for Error {
-    fn from(err: types::Error) -> Error {
-        Error {
-            msg: err.to_string(),
-        }
-    }
-}
-
-impl From<String> for Error {
-    fn from(msg: String) -> Error {
-        Error { msg }
-    }
 }
 
 impl From<Error> for String {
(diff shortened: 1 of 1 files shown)


jsternberg:
I'm taking a look at this now to see if I can find a good solution for this. I printed the output of the errors so I could see what this PR would look like. The following script created the following error:

```
r = ((o) => o)(o: 0)
            x = r
            s = ((o) => o)(o: {list: [0,1]})
            y = s.list
            t = ((o) => ({list: o}))(o: [0,1])
            z = t.list
            f = (x) => {
                y = x
                g = (a) => {
                    z = y
                    return z(b: a)
                }
                return g(a: [0])
            }
            a = f(x: (b) => b)
            b = f(x: (b) => b[0])
            c = f(x: (b) => ({b: b}))
            d = f(x: (b) => 1 + b) // int != [int]
// cannot unify [int] with int at location program@19:13-19:35
```

It is not clear to me that this is the best location for this error. It seems the error comes from `1 + b` where `b` is an `[int]`. The location refers to the variable declaration. But, what should the location information for this error be?

nathanielc:
@jsternberg There is a lot of indirection in that code but if I followed it correctly it comes down to a few things:

* the `f` function is a function that expects a parameter `x` which is a function that operates on list of ints. i.e `f : forall [t0] (x: (b: [int]) -> t0) -> t0`
* the last line where we call `f` with a function `(b) => 1 + b` as the value for `x` we understand that `b` is a list of `[ints]` since that is the type of the parameter.
* You can't add an integer to a list of ints so its a type error.

So the location information should be the binary expression `1 + b` where it communicates that `b` is a list of ints and therefore you `cannot unify [int] with int`. 

Seems like there is room to improve the error message to talk about specific identifiers etc but for now the location information is more helpful. 

Alternatively I would also accept the answer that the whole anon func `(b) => 1 + b ` is wrong since in isolation the type of that function doesn't match the type of the `x` parameter to the `f` function.

jsternberg:
I agree that the type information should probably be the `1 + b`. It seems to me like it would be nice if the message was:

```
cannot unify [int]@program:19:29 with int@program:19:33
```

It seems to me like it might make more sense to include the location information with the type itself that way you can reference a specific variable and the location where it is used. To me, it seems like the error handles inside of a section of code that doesn't really know where the error is happening so I don't think this PR will work. I am not very well aware of the code for this, but it doesn't seem like an easy fix to me.

nathanielc:
Yeah it sounds like you would prefer the constraints to contain their source location, then when two constraints don't unify you can report on both of them. That is similar to how the Go code works.

 This PR could be modified to do that. There are two error types: one in the nodes module and one in the infer module. This PR modifies the one in the nodes module to expect location information. If we also update the error type in the infer module to expect a location then instead of just using a generic location when converting from an infer::Error to an node:Error we can use the location information already present on the infer:Error.

jlapacik:
I agree we should add source location info to the constraints themselves like in the go code. I would be fine with merging this version in the meantime.

PR #2312, opened 2020-01-02
perf(stdlib): optimize window transformation

jsternberg:
This optimizes the window transformation to be more efficient than the 
current window transformation. It changes the access pattern so the time 
values are iterated over in sorted order and the windows are created in 
sorted order. This reduces the amount of random access used when iterating
over values so memory is more efficient.

It also adds a new `RandomAccessGroupLookup` that uses another method for
storing values based on their group key. The window method had a bad access
pattern for the standard `GroupLookup` because the `_start` and
`_stop` values changing would commonly result in an access pattern like 
this:

    1 2 3 4 5 6 -> 1 4 2 5 3 6

This hit one of the group lookups worst access patterns and caused a lot of
time to be spent copying memory to maintain the sorted order of the data
structure.

The window transformation still likely has more room to improve, but this
will at least cause some small amount of improvement over the old 
implementation.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/window_test.go
+++ b/stdlib/universe/window_test.go
@@ -1,7 +1,7 @@
 package universe_test
 
 import (
-	"sort"
+	"math"
 	"strconv"
 	"testing"
 	"time"
@@ -10,6 +10,8 @@ import (
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/internal/gen"
+	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/plan/plantest"
 	"github.com/influxdata/flux/querytest"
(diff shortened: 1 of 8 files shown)


jsternberg:
Benchmarks:

```
benchmark                            old ns/op        new ns/op      delta
BenchmarkWindow_Infinity_1000-12     105918759196     796584582      -99.25%
BenchmarkWindow_ByMinute_100-12      60009869814      4825699308     -91.96%

benchmark                            old allocs     new allocs     delta
BenchmarkWindow_Infinity_1000-12     696462018      1302642        -99.81%
BenchmarkWindow_ByMinute_100-12      152100590      43378993       -71.48%

benchmark                            old bytes       new bytes      delta
BenchmarkWindow_Infinity_1000-12     33036214744     576351637      -98.26%
BenchmarkWindow_ByMinute_100-12      8735916472      3785190209     -56.67%
```

The infinity implementation is a lot better and should help `aggregateWindow`. The normal implementation isn't as good and will likely be the target of future improvements, but it was good enough that I didn't want to delay good just because it wasn't perfect.

stdlib/universe/window_test.go
```diff
@@ -72,24 +74,24 @@ func TestWindowOperation_Marshaling(t *testing.T) {
...
-			},
-			execute.DefaultTimeColLabel,
-			execute.DefaultStartColLabel,
-			execute.DefaultStopColLabel,
-			false,
-		)
-		return fw
-	})
-}
+// func TestFixedWindow_PassThrough(t *testing.T) {
```
nathanielc: Why was this test removed?

jsternberg: I think it was commented out for a compilation error and I wasn't sure what it did. I'll see if I can readd it.

PR #2313, opened 2020-01-02
feat(libflux): add Go/Rust API for getting semantic graph

wolffcm:
I also added a simple (to be expanded) test case for calling type inference from Go.

I wanted to be able to have access to types from the builtins, standard library and prelude, so I had to implement `analyze` in libflux and make that linkable from CGO.  This API lets Go pass in Flux source and get a type-inferred semantic graph (that uses types from the stdlib and prelude).

I think that we'll want to refine the Go/Rust API a bit more than what's on this branch:
- It seems strange that Go needs to call into both `liblibstd` and `libflux`.   Seems like we should encapsulate everything we need from Go into a single crate.
- We'll somehow want to avoid having to copy the prelude and stdlib types very time we type-infer a program.

### DIFF
--- a/libflux/go/libflux/analyze.go
+++ b/libflux/go/libflux/analyze.go
@@ -10,18 +10,20 @@ import "C"
 
 import (
 	"errors"
+	"runtime"
 	"unsafe"
 )
 
-// Analyze parses the given Flux source, performs type inference
-// (taking into account types from prelude and stldlib) and returns
-// a byte slice containing the FlatBuffer serialization of the semantic
-// graph.
-func Analyze(src string) ([]byte, error) {
-	var buf C.struct_flux_buffer_t
-	cstr := C.CString(src)
+// SemanticPkg is a Rust pointer to a semantic package.
+type SemanticPkg struct {
+	ptr *C.struct_flux_semantic_pkg_t
+}
 
-	if err := C.flux_semantic_analyze(cstr, &buf); err != nil {
+// MarshalFB serializes the given semantic package into a flatbuffer.
+func (p *SemanticPkg) MarshalFB() ([]byte, error) {
+	var buf C.struct_flux_buffer_t
+	if err := C.flux_semantic_marshal_fb(p.ptr, &buf); err != nil {
+		defer C.flux_free(unsafe.Pointer(err))
 		cstr := C.flux_error_str(err)
 		defer C.flux_free(unsafe.Pointer(cstr))
 
@@ -32,4 +34,37 @@ func Analyze(src string) ([]byte, error) {
 
 	data := C.GoBytes(buf.data, C.int(buf.len))
 	return data, nil
+}
+
+// Free frees the memory allocated by Rust for the semantic graph.
+func (p *SemanticPkg) Free() {
+	if p.ptr != nil {
+		C.flux_free(unsafe.Pointer(p.ptr))
+	}
+	p.ptr = nil
+}
+
+// Analyze parses the given Flux source, performs type inference
+// (taking into account types from prelude and stldlib) and returns
+// an a SemanticPkg containing an opaque pointer to the semantic graph.
+// The graph can be deserialized by calling MarshalFB.
+//
+// Note that Analyze will consume the AST, so astPkg.ptr will be set to nil,
+// even if there's an error in analysis.
+func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
+	var semPkg *C.struct_flux_semantic_pkg_t
+	defer func() {
+		astPkg.ptr = nil
+	}()
+	if err := C.flux_analyze(astPkg.ptr, &semPkg); err != nil {
+		defer C.flux_free(unsafe.Pointer(err))
+		cstr := C.flux_error_str(err)
+		defer C.flux_free(unsafe.Pointer(cstr))
+
+		str := C.GoString(cstr)
+		return nil, errors.New(str)
+	}
+	p := &SemanticPkg{ptr: semPkg}
+	runtime.SetFinalizer(p, free)
+	return p, nil
 }
--- a/libflux/go/libflux/analyze.go
+++ b/libflux/go/libflux/analyze.go
@@ -1,0 +1,35 @@
+// +build libflux
+
+package libflux
+
+// #cgo CFLAGS: -I.
+// #cgo LDFLAGS: -L. -llibstd
+// #include "flux.h"
+// #include <stdlib.h>
+import "C"
+
+import (
+	"errors"
+	"unsafe"
+)
+
+// Analyze parses the given Flux source, performs type inference
+// (taking into account types from prelude and stldlib) and returns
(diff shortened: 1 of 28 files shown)


libflux/go/libflux/analyze.go
```diff
@@ -0,0 +1,35 @@
+// +build libflux
+
+package libflux
+
+// #cgo CFLAGS: -I.
+// #cgo LDFLAGS: -L. -llibstd
```
jsternberg: My understanding was that the `libstd` crate would be internal and it would only be used for generating a file in the `flux` crate and we would use conditional compilation to bootstrap the `flux` crate.

jsternberg: In the below code added to reference the prelude, I am not seeing a reference to `libstd`. Are you sure that this is needed?

libflux/src/libstd/lib.rs
```diff
@@ -19,6 +24,50 @@ pub fn fresher() -> Fresher {
...
+            return Box::into_raw(Box::new(errh)) as *mut flux_error_t;
+        }
+    }
+}
+
+fn analyze(src: &str) -> Result<Vec<u8>, flux::Error> {
+    let mut f = fresher();
+
+    let ast_file = flux::parser::parse_string("", src);
+    let sem_file = analyze_file(ast_file, &mut f).unwrap();
```
nathanielc: Should this `unwrap` be a  `?` operator instead? It seems like this error could be triggered by bad user input in which case we don't want to panic, rather report the error to the user.

libflux/src/libstd/lib.rs
```diff
@@ -19,6 +24,50 @@ pub fn fresher() -> Fresher {
...
+            std::ptr::null_mut()
+        }
+        Err(err) => {
+            let errh = flux::ErrorHandle { err: Box::new(err) };
+            return Box::into_raw(Box::new(errh)) as *mut flux_error_t;
+        }
+    }
+}
+
+fn analyze(src: &str) -> Result<Vec<u8>, flux::Error> {
```
nathanielc: Should we make this a public method? So that it can be considered the entry point into performing type inference from the Rust API? 

Also I think our terminology is a bit inconsistent at this point. Should this method be called analyze or something else? Previously analyze referred only to the step of converting ast to semantic graph but without the type inference bits.  This method includes parsing, converting ast -> semantic and running type inference. I actually like the name analyze but maybe we should rename the `semantic/analyze` package to `convert` or something different so to not confuse the two.

libflux/src/libstd/lib.rs
```diff
@@ -19,6 +24,50 @@ pub fn fresher() -> Fresher {
...
+    let mut f = fresher();
+
+    let ast_file = flux::parser::parse_string("", src);
+    let sem_file = analyze_file(ast_file, &mut f).unwrap();
+    let mut sem_pkg = flux::semantic::nodes::Package {
```
nathanielc: What if the source code has a `package` statement inside of it? How would that get populated on the package struct?

libflux/src/libstd/lib.rs
```diff
@@ -19,6 +24,50 @@ pub fn fresher() -> Fresher {
...
+            let errh = flux::ErrorHandle { err: Box::new(err) };
+            return Box::into_raw(Box::new(errh)) as *mut flux_error_t;
+        }
+    }
+}
+
+fn analyze(src: &str) -> Result<Vec<u8>, flux::Error> {
+    let mut f = fresher();
+
+    let ast_file = flux::parser::parse_string("", src);
```
nathanielc: Do we want to let the caller determine the file name of the returned ast.File? Then if the caller wants an empty string they just pass one in?

libflux/go/libflux/analyze.go
wolffcm: Hmm, that doesn't seem to be how it works though.  We generate `.data` files that contain serialized types for the standard library and prelude, but they are consumed by `libstd` which exposes public `prelude` and `stdlib` functions.  The `libstd` crate depends on  the `flux` crate, so I don't see how to type infer Flux and include the standard library unless Go talks to the `libstd` crate.

Maybe @jlapacik can shed more light on how this is intended to work.

I'm not sure what you mean by "the below code added to reference the prelude"... maybe I'm misunderstanding your question.

jsternberg: The original idea @jlapacik and I worked out is that the `flux` crate would conditionally compile. The `libstd` crate would use a `nostdlib` or `core-only` feature flag when referencing flux and would generate the data files which would only be referenced when the standard library feature flag was activated. The default would end up being the standard library as enabled. This way, we could have `libstd` reference the `flux` crate while avoiding multiple crates. This is why I figured the linker flag for this should still be to `libflux` because `libflux` is supposed to contain everything. I am not sure if this got implemented in practice so I apologize if I have this incorrect.

wolffcm:
I've updated this PR with the design we discussed during our meeting yesterday.

libflux/go/libflux/flux.h
```diff
@@ -22,29 +19,53 @@ struct flux_buffer_t {
...
-// This will allocate memory for the returned string.
-const char *flux_error_str(struct flux_error_t *);
+// flux_analyze analyzes the given AST and will populate the second pointer argument with
+// a pointer to the resulting semantic graph. It is the caller's responsibility to free the
+// resulting semantic graph with a call to flux_free().
+// If analysis fails, the second pointer argument wil be pointed at 0, and an error will be returned.
+// Any non-null error must be freed by calling flux_free.
+// Regardless of whether an error is returned, this function will consume and free its
+// flux_ast_pkg_t* argument, so it should not be reused after calling this function.
+struct flux_error_t *flux_analyze(struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
```
nathanielc: Why does this method use a double pointer? When using `flux_free` on the `flux_semantic_pkg_t`  do I free the first or second pointer? Would it make the API simpler if this were a single pointer?

wolffcm: I don't think this would work if it was a single pointer, unless we were to return it as a return value.  But we are returning a `flux_error_t*` and it's nice to be consistent with the other functions.  C doesn't have multiple return values, so we are left with using a function parameter.  (Or creating a new struct type to return both an error and the address of the graph)

Here's why a single pointer wouldn't work.  We want Rust to allocate some memory for a semantic graph, and return a pointer to it.  If we were to pass a single pointer to this function, and have Rust update it to point at the semantic graph, Go would never receive this address, because pointers are passed by value.  To make Go aware of the new value, we need to pass the address of a pointer to a semantic graph, and have Rust update *that*.

The caller needs to free the pointer-to-the-graph, not the pointer-to-the-pointer to the graph.  The latter item is managed by Go.

I agree this is really confusing, and I'm open to other ideas.   This was pretty common practice on C codebases I've worked on.

nathanielc: OK thanks for the explanation, maybe we can add the last bit about what to free as a comment and that will be good for me.

PR #2314, opened 2020-01-03
perf(stdlib): optimize limit transformation

jsternberg:
The limit transformation has been optimized to use a passthrough dataset
and to use arrow slices rather than copying the input to a new table.
This reduces the number of memory allocations which makes the
transformation use a more consistent amount of memory and speeds up the
transformation because it doesn't have to wait for memory allocations to
finish.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/limit.go
+++ b/stdlib/universe/limit.go
@@ -1,11 +1,15 @@
 package universe
 
 import (
+	"context"
+
+	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/execute"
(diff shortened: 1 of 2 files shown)


jsternberg:
Benchmarks:

```
benchmark                       old ns/op      new ns/op     delta
BenchmarkLimit_1N_1000-12       608936123      240179582     -60.56%
BenchmarkLimit_100N_1000-12     1527333381     237654834     -84.44%

benchmark                       old allocs     new allocs     delta
BenchmarkLimit_1N_1000-12       3671443        810008         -77.94%
BenchmarkLimit_100N_1000-12     4691770        810008         -82.74%

benchmark                       old bytes      new bytes     delta
BenchmarkLimit_1N_1000-12       282523411      58560502      -79.27%
BenchmarkLimit_100N_1000-12     1027038739     58560394      -94.30%
```

PR #2324, opened 2020-01-06
chore: update PolyType and MonoType String() method to be deterministic

wolffcm:
This PR concerns the new `PolyType` and `MonoType` structs that are backed by flat buffers.

This change updates the `String()` method for `PolyType` and `MonoType` to make them more useful, so that output is always consistent. This will make them easier to use in tests.

### DIFF
--- a/semantic/types/monotype.go
+++ b/semantic/types/monotype.go
@@ -245,6 +245,30 @@ func (mt *MonoType) Property(i int) (*Property, error) {
 	return &Property{fb: p}, nil
 }
 
+// SortedProperties returns the properties for a Row monotype, sorted by
+// key.  It's possible that there are duplicate keys with different types,
+// in this case, this function preserves their order.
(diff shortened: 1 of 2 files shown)


wolffcm:
Yes, you're right, I'll sort those too.  

I opted not to sort row fields, since they are implemented as a linked list in Rust for `PolyType`, do you think there's any chance their order could vary as well?

jlapacik:
@wolffcm that's a good point. Property order can vary except for the relative order of properties with the same key. I.e if `k: "a", v: int` comes before `k: "a", v: float` in one ordering, it must be like that for every ordering.

jsternberg:
It doesn't look like the MonoType `String()` has been updated to be sorted in this PR. I didn't see any changes to that `String()` function.

wolffcm:
@jsternberg Function arguments are now sorted in the `MonoType` `String()` method, maybe you missed it.  I still need to do constraints on polytypes and fields in Rows.

PR #2333, opened 2020-01-07
feat: add lookuptype func for stdlib builtins

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/go/libflux/analyze.go
+++ b/libflux/go/libflux/analyze.go
@@ -45,7 +45,7 @@ func (p *SemanticPkg) Free() {
 }
 
 // Analyze parses the given Flux source, performs type inference
-// (taking into account types from prelude and stldlib) and returns
+// (taking into account types from prelude and stdlib) and returns
 // an a SemanticPkg containing an opaque pointer to the semantic graph.
 // The graph can be deserialized by calling MarshalFB.
 //
@@ -67,4 +67,14 @@ func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
 	p := &SemanticPkg{ptr: semPkg}
 	runtime.SetFinalizer(p, free)
 	return p, nil
+}
+// EnvStdlib takes care of creating a flux_buffer_t, passes the buffer to
+// the Flatbuffers TypeEnvironment and then takes care of freeing the data
+func EnvStdlib() []byte {
+	var buf C.struct_flux_buffer_t
+	C.flux_get_env_stdlib(&buf)
+
+	defer C.flux_free(buf.data)
+
+	return C.GoBytes(buf.data, C.int(buf.len))
 }
--- a/libflux/go/libflux/flux.h
+++ b/libflux/go/libflux/flux.h
@@ -51,6 +51,10 @@ struct flux_error_t *flux_ast_marshal_fb(struct flux_ast_pkg_t *, struct flux_bu
 // and their contents.
 struct flux_semantic_pkg_t;
 
+// flux_get_env_stdlib instantiates a flatbuffers TypeEnvironment and creates a pointer
+// to it to use when performing lookups on the stdlib
+void flux_get_env_stdlib(struct flux_buffer_t *);
+
 // flux_analyze analyzes the given AST and will populate the second pointer argument with
 // a pointer to the resulting semantic graph.
 // It is the caller's responsibility to free the resulting semantic graph with a call to flux_free().
--- a/semantic/lookup.go
+++ b/semantic/lookup.go
@@ -1,0 +1,99 @@
+package semantic
+
+import (
+	flatbuffers "github.com/google/flatbuffers/go"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/libflux/go/libflux"
+	"github.com/influxdata/flux/semantic/internal/fbsemantic"
+)
+
+// LookupBuiltInType returns the type of the builtin value for a given
+// Flux stdlib package. Returns a empty MonoType if lookup fails.
+func LookupBuiltInType(pkg, name string) (MonoType, error) {
+	byteArr := libflux.EnvStdlib()
+	env := fbsemantic.GetRootAsTypeEnvironment(byteArr, 0)
+	var table flatbuffers.Table
+	// perform a lookup of package identifier in flatbuffers TypeEnvironment
+	prop, err := lookup(env, pkg, name)
+	if err != nil {
+		return MonoType{}, err
+	}
+	if !prop.V(&table) {
+		return MonoType{}, errors.Newf(codes.Internal, "Prop value is not valid: %v", err)
+	}
+	monotype, err := NewMonoType(table, prop.VType())
+	if err != nil {
+		return MonoType{}, err
+	}
+	// return fb polytype within semantic wrapper
+	return monotype, nil
+}
+
+// lookup is a helper function that performs a lookup of a package identifier in
+// a flatbuffers TypeEnvironment. It first checks for the package using path string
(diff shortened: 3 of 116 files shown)


libflux/go/libflux/analyze.go
```diff
@@ -68,3 +68,10 @@ func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
 	runtime.SetFinalizer(p, free)
 	return p, nil
 }
+// TODO(faith): fix comment; create the buffer, pass it in and goes takes care of freeing data
+func EnvStdlib() []byte {
+	var buf C.struct_flux_buffer_t
+	C.flux_get_env_stdlib(&buf)
```
wolffcm: You'll want to call `flux_free` to release the buffer allocated by Rust, similar to how `MarshalFB` does above.

semantic/lookup_test.go
```diff
@@ -0,0 +1,163 @@
...
+			id:   "assertEmpty",
+			name: "lookup testing.assertEmpty",
+			want: "(<-tables: [t0]) -> [t0]",
+		},
+		{
+			path: "universe",
+			id:   "chandeMomentumOscillator",
+			name: "lookup testing.assertEmpty",
+			want: "(?columns: [string], n: int, <-tables: [t0]) -> [t1]",
+		},
```
wolffcm: This testing looks good.  I don't think you need to lock down every single builtin here.  Personally I think it's enough to just assert that a few of the builtins are here so you know that lookup is working.

libflux/go/libflux/analyze.go
fchikwekwe: Yeah thanks! I was waiting on this update so I'll go ahead and conform to how we're doing this now.

semantic/lookup.go
```diff
@@ -0,0 +1,99 @@
...
+	if propErr := foundProp(row, prop, name); propErr != nil {
+		return nil, propErr
+	}
+
+	return prop, nil
+}
+
+// foundPackage is a helper function that iterates over type assignments and checks
+// for a given package returning an error if the package is not found
+func foundPackage(env *fbsemantic.TypeEnvironment, obj *fbsemantic.TypeAssignment, pkg string) error {
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, that's a very good point. I wasn't really thinking in terms of optimization, but I agree that the linear search here could become an issue. I'll consider using maps here to make things faster.

semantic/lookup_test.go
fchikwekwe: Are there too many test cases here? Should I remove some? My intention was to provide coverage of every stdlib package and various PolyType representations.

semantic/lookup.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	// return fb polytype within semantic wrapper
+	return monotype, nil
+}
+
+// lookup is a helper function that performs a lookup of a package identifier in
+// a flatbuffers TypeEnvironment. It first checks for the package using path string
+// and then checks for the package identifier, the "name" string in this case, returning
+// the corresponding MonoType if found
+func Lookup(env *fbsemantic.TypeEnvironment, pkg, name string) (*fbsemantic.Prop, error) {
```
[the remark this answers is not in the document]
fchikwekwe: I needed to make this public since testing was in a different package. I guess I could write a wrapper that makes this accessible from the testing package?

semantic/monotype.go
```diff
@@ -19,8 +19,8 @@ type fbTabler interface {
 // MonoType represents a monotype.  This struct is a thin wrapper around
 // Go code generated by the FlatBuffers compiler.
 type MonoType struct {
-	mt  fbsemantic.MonoType
-	tbl fbTabler
+	Mt  fbsemantic.MonoType
```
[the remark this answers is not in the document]
fchikwekwe: Okay, I'll look into changing that.

libflux/go/libflux/flux.h
```diff
@@ -51,6 +51,16 @@ struct flux_error_t *flux_ast_marshal_fb(struct flux_ast_pkg_t *, struct flux_bu
...
+// Take a Flux source program, analyze it, and return its flatbuffer.
+struct flux_error_t *flux_semantic_analyze(const char* src, struct flux_buffer_t *);
+
+// TODO(faith): add a comment 
+void flux_get_env_stdlib(struct flux_buffer_t *);
+
+// flux_buffer_free will free the memory that was allocated for a buffer.
+// This should only be called if the caller is the one who owns the data.
+void flux_buffer_free(struct flux_buffer_t *);
+
```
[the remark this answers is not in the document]
fchikwekwe: Yeah I think this was something that carried over from one of my merges. I'll remove these.

libflux/go/libflux/analyze.go
wolffcm: I think it should be safe... If you are calling `GoBytes` here, then it will do a copy, and it will be safe to free the memory that you got from Rust.

fchikwekwe: Yeah, changed this. Doesn't seem to be affecting the way the lookup performs.

semantic/lookup_test.go
wolffcm: I think what you have here is fine.  

I don't think someone should need to update this test whenever we add something to the stdlib.  It would just be more boilerplate, and we already have roundtrip testing for types.  There will also be tests added to test the functionality of what is added to the stdlib.  If this tests looks like it might be an exhaustive list of all the builtins, it would be confusing if things are missing.

fchikwekwe: I messaged you about this: I'm gonna keep most of these test cases, but clarify that this list is not exhaustive in a comment.

semantic/lookup.go
```diff
@@ -0,0 +1,99 @@
...
+	if propErr := foundProp(row, prop, name); propErr != nil {
+		return nil, propErr
+	}
+
+	return prop, nil
+}
+
+// foundPackage is a helper function that iterates over type assignments and checks
+// for a given package returning an error if the package is not found
+func foundPackage(env *fbsemantic.TypeEnvironment, obj *fbsemantic.TypeAssignment, pkg string) error {
```
wolffcm: Just a style comment here: it would be more idiomatic and easier to read if you created a `*fbsemantic.TypeAssignment` in this method, and returned it as a value, rather than have the caller create it and pass it as a pointer.  So the signature to this method would be something like 
```Go
func foundPackage(env *fbsemantic.TypeEnvironment, pkg string) (*fbsemantic.TypeAssignment, error)
```

Another, nit-like comment, it's good when method names contain verbs, but they're typically in the present tense, so this method might be better named `findPackage`.

semantic/lookup.go
```diff
@@ -0,0 +1,99 @@
...
+				return nil
+			}
+		}
+	}
+	return errors.Newf(codes.Internal, "package not found %v", pkg)
+}
+
+// foundProp is a helper function that iterates over row properties and checks
+// for a given package identifier, returning an error if the identifier is not found
+func foundProp(row *fbsemantic.Row, obj *fbsemantic.Prop, name string) error {
```
wolffcm: (Similar comments as above)

semantic/lookup.go
```diff
@@ -0,0 +1,100 @@
+package semantic
+
+import (
+	"fmt"
```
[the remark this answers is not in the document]
fchikwekwe: I didn't actually need the `fmt` import so I removed it.

semantic/lookup.go
fchikwekwe: Yeah, these functions might go away after I merge #2377 anyway, but those are good notes. If I keep either func, I'll add these changes into that PR.

PR #2339, opened 2020-01-07
refactor(semantic): flatbuffers for basic types

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

Initialize flatbuffers for basic types

### DIFF
--- a/semantic/monotype.go
+++ b/semantic/monotype.go
@@ -115,20 +115,32 @@ var (
 )
 
 func init() {
-	builder := flatbuffers.NewBuilder(1024)
+	BasicBool = buildBasic(fbsemantic.TypeBool)
+	BasicInt = buildBasic(fbsemantic.TypeInt)
+	BasicUint = buildBasic(fbsemantic.TypeUint)
+	BasicFloat = buildBasic(fbsemantic.TypeFloat)
+	BasicString = buildBasic(fbsemantic.TypeString)
+	BasicDuration = buildBasic(fbsemantic.TypeDuration)
+	BasicTime = buildBasic(fbsemantic.TypeTime)
+	BasicRegexp = buildBasic(fbsemantic.TypeRegexp)
+	BasicBytes = buildBasic(fbsemantic.TypeBytes)
+}
+
+func buildBasic(ty fbsemantic.Type) MonoType {
+	builder := flatbuffers.NewBuilder(0)
+
 	fbsemantic.BasicStart(builder)
-	fbsemantic.BasicAddT(builder, fbsemantic.TypeBool)
-	basicBool := fbsemantic.BasicEnd(builder)
+	fbsemantic.BasicAddT(builder, ty)
+	offset := fbsemantic.BasicEnd(builder)
 
-	// TODO (algow): initial all basic values
-
-	// TODO (algow): this probably doesn't work...
-	builder.Finish(basicBool)
-	basicTable := flatbuffers.Table{
+	builder.Finish(offset)
+	table := flatbuffers.Table{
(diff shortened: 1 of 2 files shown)


semantic/monotype.go
```diff
@@ -103,7 +103,15 @@ func (mt MonoType) Kind() Kind {
 }
 
 var (
-	basicTable *flatbuffers.Table
+	tableBool     *flatbuffers.Table
```
jsternberg: Why are these necessary?

semantic/monotype.go
```diff
@@ -117,20 +125,133 @@ var (
...
 
 func init() {
-	builder := flatbuffers.NewBuilder(1024)
-	fbsemantic.BasicStart(builder)
-	fbsemantic.BasicAddT(builder, fbsemantic.TypeBool)
-	basicBool := fbsemantic.BasicEnd(builder)
 
-	// TODO (algow): initial all basic values
+	{
+		builder := flatbuffers.NewBuilder(0)
```
jsternberg: It seems like the below code can be refactored into a single function that returns the monotype. There's only one parameter that changes and that's the type.

semantic/monotype.go
jlapacik: Not entirely sure.

semantic/monotype.go
```diff
@@ -117,20 +115,32 @@ var (
...
-	basicTable = &flatbuffers.Table{
+	fbsemantic.BasicStart(builder)
+	fbsemantic.BasicAddT(builder, ty)
+	offset := fbsemantic.BasicEnd(builder)
+
+	builder.Finish(offset)
+	table := flatbuffers.Table{
 		Bytes: builder.FinishedBytes(),
-		Pos:   basicBool,
+		Pos:   offset,
```
jsternberg: I don't think this is correct. I think the proper code for this is:

```
table := fbsemantic.GetRootAsBasic(builder.FinishedBytes(), 0).Table()
basic, _ := NewMonoType(&table, fbsemantic.MonoTypeBasic)
```

semantic/monotype.go
```diff
@@ -117,20 +115,32 @@ var (
...
+	BasicFloat = buildBasic(fbsemantic.TypeFloat)
+	BasicString = buildBasic(fbsemantic.TypeString)
+	BasicDuration = buildBasic(fbsemantic.TypeDuration)
+	BasicTime = buildBasic(fbsemantic.TypeTime)
+	BasicRegexp = buildBasic(fbsemantic.TypeRegexp)
+	BasicBytes = buildBasic(fbsemantic.TypeBytes)
+}
 
-	// TODO (algow): initial all basic values
+func buildBasic(ty fbsemantic.Type) MonoType {
```
jsternberg: Nitpick: `buildBasicType`

semantic/monotype.go
```diff
@@ -117,20 +115,32 @@ var (
...
+
+	builder.Finish(offset)
+	table := flatbuffers.Table{
 		Bytes: builder.FinishedBytes(),
-		Pos:   basicBool,
+		Pos:   offset,
 	}
-	BasicBool, _ = NewMonoType(basicTable, fbsemantic.MonoTypeBasic)
+
+	basic, _ := NewMonoType(&table, fbsemantic.MonoTypeBasic)
```
jsternberg: Can you catch the error here and panic if it does?

jsternberg:
Can you also add test cases for each of these? I think you should be able to use the `String()` method as @wolffcm advised.

jsternberg:
Not needed anymore.

Issue #2317, opened 2020-01-03
Determine method for creating compound types 


nathanielc:
On the Go side we occasionally need the ability to construct a Flux type as a flatbuffer. This is a tedious process and its not straightforward how it should be done. 

We need to create helper methods for creating new compound Flux types. See https://github.com/influxdata/flux/blob/feat/use-algo-w/semantic/monotype.go#L447 for some incomplete methods along these lines.

nathanielc:
NOTE:

* Can use String method for easy tests
* Does not need to be memory efficient. These methods are called infrequently so if they need to perform copies of flatbuffers that is fine for now. 
* Flatbuffers cannot reference data inside other buffers, as such these methods will always be creating a new flatbuffer

PR #2349, opened 2020-01-08
refactor(semantic): flatbuffers for basic and complex types

### LINKED ISSUE #2316: Create static instance of basic types
We need a static value for each of the basic Flux types represented as flatbuffers on the Go side. See 
https://github.com/influxdata/flux/blob/feat/use-algo-w/semantic/monotype.go#L125

We need to build and initialize all of the basic types defined there during init.

jsternberg:
Co-authored-by: Joshua Lapacik @jlapacik

Fixes #2316 and #2317.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


semantic/monotype.go
```diff
@@ -442,20 +425,226 @@ func (l MonoType) Equal(r MonoType) bool {
...
+	buf := builder.FinishedBytes()
+	basic := fbsemantic.GetRootAsBasic(buf, 0)
+	mt, err := NewMonoType(basic.Table(), fbsemantic.MonoTypeBasic)
+	if err != nil {
+		panic(err)
+	}
+	return mt
+}
+
 func NewArrayType(elemType MonoType) MonoType {
```
nathanielc: Can you comment this method, and also indicate that it will panic and under what conditions?

semantic/monotype.go
```diff
@@ -442,20 +425,226 @@ func (l MonoType) Equal(r MonoType) bool {
...
+	arr := fbsemantic.GetRootAsArr(buf, 0)
+	mt, err := NewMonoType(arr.Table(), fbsemantic.MonoTypeArr)
+	if err != nil {
+		panic(err)
+	}
+	return mt
+}
+
+type ArgumentType struct {
+	Name     []byte
```
nathanielc: Should this be a string to make it easier for callers of these methods? I don't have a strong preference

semantic/types_test.go
```diff
@@ -1,224 +1,219 @@
...
-				}
-			} else {
-				if tc.a == tc.b {
-					t.Errorf("expected types to not be equal: pointers %p %p a: %#v b: %#v", tc.a, tc.b, tc.a, tc.b)
-				}
-			}
-		})
-	}
-}
+// TODO(jsternberg): Readd these tests if they are still applicable.
```
nathanielc: These are no longer applicable and can be deleted.  They were used to validate that the pointer magic we used to compare types was correct. That code has already been removed in the branch.

semantic/monotype.go
```diff
@@ -442,20 +425,226 @@ func (l MonoType) Equal(r MonoType) bool {
...
+		tv.Init(table.Bytes, table.Pos)
+		return buildVarType(builder, tv.I())
+	case fbsemantic.MonoTypeArr:
+		var arr fbsemantic.Arr
+		arr.Init(table.Bytes, table.Pos)
 
-	return MonoType{}
+		elem := monoTypeFromFunc(arr.T, arr.TType())
+		return buildArrayType(builder, elem)
+	case fbsemantic.MonoTypeRow:
```
jlapacik: What about `row.Extends()`? Objects can extend a tvar.

semantic/monotype.go
```diff
@@ -442,20 +425,226 @@ func (l MonoType) Equal(r MonoType) bool {
...
+	fbsemantic.FunStart(builder)
+	fbsemantic.FunAddRetnType(builder, retn.mt)
+	fbsemantic.FunAddRetn(builder, retnOffset)
+	fbsemantic.FunAddArgs(builder, argsOffset)
+	return fbsemantic.FunEnd(builder)
+}
+
+// buildObjectType will construct a row type in the builder
+// and return the offset for the type.
+func buildObjectType(builder *flatbuffers.Builder, properties []PropertyType) flatbuffers.UOffsetT {
```
jlapacik: Shouldn't this take an optional type variable for records that are extended from one?

semantic/monotype.go
jsternberg: I chose a byte since that's the API for what gets returned by flatbuffers and using a string would incur a memory allocation. I figured that it was more performant to go from `string -> []byte` than `[]byte -> string`.

semantic/monotype.go
jsternberg: There's a panic in there, but it's from `NewMonoType` which is getting invoked in a way where it can never panic. So the panic is more to catch that something is really wrong. I'll add a godoc though.

semantic/monotype.go
nathanielc: Works for me

jsternberg:
@jlapacik I added support for copying the extends variable if it is present. I didn't add an option to set it within Go because I didn't think it was necessary, but it is necessary to copy the value if it is present from Rust.

PR #2354, opened 2020-01-09
refactor(values): mutable object values with the new semantic object

### LINKED ISSUE #2318: Reimplement mutable object value with flatbuffers
We use a mutable object value on the Go side in a few places. We need to reimplement that behavior using flatbuffers. See https://github.com/influxdata/flux/blob/feat/use-algo-w/values/object.go#L22

Since using mutable flatbuffers is complicated maybe we can implement only a subset of the previous behavior where the type is known and not mutable but the values are mutable? Not sure if this is possible in all current uses.

jsternberg:
This modifies the API for creating object values since the object's type
needs to be known before we create the object now. It also adds a
utility method for building an object that is easier for creating an
object.

Fixes #2318.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -487,27 +487,51 @@ func (itrp *Interpreter) doArray(ctx context.Context, a *semantic.ArrayExpressio
 }
 
 func (itrp *Interpreter) doObject(ctx context.Context, m *semantic.ObjectExpression, scope values.Scope) (values.Value, error) {
-	obj := values.NewObject()
-	if m.With != nil {
-		with, err := itrp.doExpression(ctx, m.With, scope)
-		if err != nil {
-			return nil, err
-		}
-		with.Object().Range(func(k string, v values.Value) {
-			obj.Set(k, v)
-		})
+	if label, nok := itrp.checkForDuplicates(m.Properties); nok {
+		return nil, errors.Newf(codes.Invalid, "duplicate key in object: %q", label)
 	}
-	for _, p := range m.Properties {
-		v, err := itrp.doExpression(ctx, p.Value, scope)
-		if err != nil {
-			return nil, err
+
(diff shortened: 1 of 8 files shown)


values/object.go
```diff
@@ -87,43 +134,17 @@ func (o *object) String() string {
 }
 
 func (o *object) Type() semantic.MonoType {
-	// TODO (algow): finish implementation of object
-	return semantic.MonoType{}
+	return o.typ
 }
 
 func (o *object) Set(k string, v Value) {
```
nathanielc: Should we rename this to `Replace` or something similar? Since it will not create a new property but instead will only replace the existing value. Can you add a comment describing its behavior?

interpreter/interpreter.go
```diff
@@ -487,27 +487,51 @@ func (itrp *Interpreter) doArray(ctx context.Context, a *semantic.ArrayExpressio
...
-	obj := values.NewObject()
-	if m.With != nil {
-		with, err := itrp.doExpression(ctx, m.With, scope)
-		if err != nil {
-			return nil, err
-		}
-		with.Object().Range(func(k string, v values.Value) {
-			obj.Set(k, v)
-		})
+	if label, nok := itrp.checkForDuplicates(m.Properties); nok {
```
jlapacik: What is this check for?

jlapacik: I guess right now, we won't allow duplicate keys?

jsternberg: That seemed to be the old behavior.

values/object.go
jsternberg: I think `Set` makes sense, but I'll add a godoc comment to the interface making it clear that it is only setting values and doesn't add anything.

PR #2359, opened 2020-01-09
test(semantic): add roundtrip testing for Flatbuffers serialization

wolffcm:
This PR adds roundtrip testing for FlatBuffers serialization of the semantic graph, and also fixes some issues that I found while writing these tests.

I also added some utilities that help with reindexing type variables starting from zero so that changes in the standard lib or the order that that expressions are analyzed do not make the test fail.


wolffcm:
Yes, I knew there would be bugs, and it was bugging *me* that there was no testing for it.

PR #2361, opened 2020-01-10
refactor: update cmp options for semantic package and remove JSON serialization

wolffcm:
This removes JSON serialization support for the semantic graph:
- We don't actually use it anywhere
- We can use flatbuffers serialization instead

This also updates the cmp options so we can compare monotypes and polytypes using their `String()` methods.


semantic/monotype.go
```diff
@@ -366,6 +366,9 @@ func (p *RowProperty) TypeOf() (MonoType, error) {
 
 // String returns a string representation of this monotype.
 func (mt MonoType) String() string {
+	if mt.tbl == nil {
+		return "<monotype: nil>"
+	}
```
jlapacik: Is there a reason not to just return the empty string?

wolffcm: This was from the changes that @jsternberg suggested.  I like that it makes it explicit that there's no type there, and not some other problem.

PR #2376, opened 2020-01-14
feat(libflux/flux): add formatter lib

kelwang:
add the rust formatter

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/flux/formatter/mod.rs
+++ b/libflux/src/flux/formatter/mod.rs
@@ -1,0 +1,747 @@
+use crate::ast::{self, walk::Node};
+use crate::Error;
+
+use chrono::SecondsFormat;
+
+pub struct Formatter {
+    builder: String,
+    indentation: u32,
+    err: Option<Error>,
+}
+
+impl Formatter {
+    // giving too much capacity can significantly boost the performance
+    // see https://github.com/hoodie/concatenation_benchmarks-rs
+    pub fn new(cap: usize) -> Formatter {
+        Formatter {
+            builder: String::with_capacity(cap),
+            indentation: 0,
+            err: None,
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


libflux/src/flux/formatter/tests.rs
```diff
@@ -0,0 +1,225 @@
...
+    format_helper(r#""\xe6\x97\xa5\xe6\x9c\xac\xe8\xaa\x9e""#); // with byte
+}
+
+#[test]
+fn package_import() {
+    format_helper("package foo\n"); // package
+    format_helper("import \"path/foo\"\nimport bar \"path/bar\""); // imports
+    format_helper(
+        "import foo \"path/foo\"\n\nfoo.from(bucket: \"testdb\")\n\
+         \t|> range(start: 2018-05-20T19:53:26Z)",
```
nathanielc: Is there a way to write these tests without the `\t` and `\n` escape chars? That would make them easier to read.

kelwang: my editor has some issue with r#""#, let me try with other editors

libflux/src/flux/formatter/mod.rs
```diff
@@ -0,0 +1,738 @@
...
+impl Formatter {
+    // returns the final string and error msg
+    pub fn output(&self) -> (&str, &str) {
+        if let Some(err) = &self.err {
+            return ("", &err.msg);
+        }
+        (&self.builder, "")
+    }
+
+    fn write_string(&mut self, s: &str) {
```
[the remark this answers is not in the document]
kelwang: thanks, looks like if we alloc the capacity, we can gain a lot of performance

PR #2379, opened 2020-01-14
chore: remove conditional compilation and FLUX_PARSER_TYPE env var

wolffcm:
This commit removes conditional compilation that uses the `libflux` tag, and also the `FLUX_PARSER_TYPE` environment variable.

The `generate` step of the build was also broken, and this fixes that as well.

The Go code that is generated by `flatc` is now all placed into one file.  This helps the Makefile be easier to maintain.

Note that the Rust parser does not retain the flux source file for internal source location nodes, and this now shows up in the `_gen.go` files for the standard library.  I'll file a follow-up issue to fix this.


ast/internal/fbast/doc.go
```diff
@@ -0,0 +1,3 @@
+package fbast
+
+// Package fbast contains code generated by the FlatBuffers compiler for serializing AST.
```
jsternberg: I think the package comment goes above the package directive.

Issue #2084, opened 2019-10-31
Update `make release` to check for local/remote consistency


affo:
I was able to do this:

 * Run `git pull origin master` not `git pull` nor `git fetch`
 * Run `make release`
 * `changelog` doesn't know the last tag because it wasn't fetched at all. Now, suppose that the last remote tag is vX, then `changelog` will create vX locally.
 * Remembered to fetch. Git now sees that I have vX locally and does not care about the remote/local inconsistency.
 * Run `make release` again. Now `changelog` create vY in the same place as vX producing an empty changelog (vY should be vX.1).

To solve this we should `make release` `git pull` before tagging to ensure that local is in sync with remote.

affo:
@nathanielc erroring seems a better approach 👍

fchikwekwe:
Trying to get more context on this: is there a way to know which commit we need to checkout and compare against remote?

fchikwekwe:
I'm also not sure what checking if local and remote are in sync would look like. Right now I'm thinking that this would be a change to `release.sh` to compare the two commits before making the release and commiting it.

PR #2383, opened 2020-01-14
fix: remove 'tags' line from local tags

fchikwekwe:
The `release.sh` file was recently updated (https://github.com/influxdata/flux/issues/2084) to properly compare local and remote tags. This PR updates the file to remove the `refs/tags/list` line from the local tags since this line does not appear in the remote tags list. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


fchikwekwe:
Yeah, I agree that this will have to be updated again if there are any additional lines like this. Perhaps I should open an issue to revisit this file?

wolffcm:
Sure, you can file an issue for it.

PR #2389, opened 2020-01-15
refactor(execute): update execute package to work with algorithm w

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/row_fn.go
+++ b/execute/row_fn.go
@@ -2,7 +2,6 @@ package execute
 
 import (
 	"context"
-	"regexp"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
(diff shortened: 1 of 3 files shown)


execute/row_fn.go
```diff
@@ -13,52 +12,85 @@ import (
...
+		}
+		properties[i] = semantic.PropertyType{
+			Key:   []byte(c.Label),
+			Value: vtype,
+		}
 	}
+	return semantic.NewObjectType(properties), nil
 }
 
 func (f *dynamicFn) prepare(cols []flux.ColMeta, extraTypes map[string]semantic.MonoType) error {
```
jlapacik: Is extraTypes still necessary?

execute/row_fn.go
```diff
@@ -13,52 +12,85 @@ import (
...
+	fn         *semantic.FunctionExpression
 	recordName string
-	record     *Record
 
-	recordCols map[string]int
-	references []string
+	// These are initialized after the function is prepared.
+	preparedFn compiler.Func
+	arg0       values.Object
+	args       values.Object
```
jlapacik: What are `args` and `args0`? Why do we need both?

execute/row_fn.go
jsternberg: I think so. I think it's needed for reduce but I haven't gotten there yet honestly.

execute/row_fn.go
jsternberg: `args` is the object for calling the function and `arg0` is the first argument specifically in the constructed object.

Arguments are sent to functions by using an object expression so `foo({a: 1, b: 2})` is two arguments, but the `Eval` function only takes the argument for the first one. So `arg0` is the object for `r` itself while `args` is for the all parameters.

PR #2390, opened 2020-01-15
fix(semantic): optimize lookup function using hashmap

fchikwekwe:
`LookUpBuiltinType` was creating the entire TypeEnvironment and using linear search for every single lookup. This attempts to create the TypeEnvironment once and store it in a map. Then `LookUpBuiltinType` can be called for every stdlib builtin that needs to be looked up. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/semantic/lookup.go
+++ b/semantic/lookup.go
@@ -1,6 +1,7 @@
 package semantic
 
 import (
+	"fmt"
 	flatbuffers "github.com/google/flatbuffers/go"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
@@ -8,19 +9,16 @@ import (
 	"github.com/influxdata/flux/semantic/internal/fbsemantic"
 )
 
+var stdlibTypeEnvironment = TypeEnvMap(fbsemantic.GetRootAsTypeEnvironment(libflux.EnvStdlib(), 0))
+
 // LookupBuiltInType returns the type of the builtin value for a given
-// Flux stdlib package. Returns a empty MonoType if lookup fails.
+// Flux stdlib package. Returns an error if lookup fails.
 func LookupBuiltInType(pkg, name string) (MonoType, error) {
-	byteArr := libflux.EnvStdlib()
-	env := fbsemantic.GetRootAsTypeEnvironment(byteArr, 0)
 	var table flatbuffers.Table
-	// perform a lookup of package identifier in flatbuffers TypeEnvironment
-	prop, err := lookup(env, pkg, name)
-	if err != nil {
-		return MonoType{}, err
-	}
+	prop := stdlibTypeEnvironment[pkg][name] // query environment map for prop
+
 	if !prop.V(&table) {
-		return MonoType{}, errors.Newf(codes.Internal, "Prop value is not valid: %v", err)
+		return MonoType{}, errors.Newf(codes.Internal, "Prop value is not valid: pkg %v name %v", pkg, name)
 	}
 	monotype, err := NewMonoType(table, prop.VType())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


semantic/lookup_test.go
```diff
@@ -49,55 +49,55 @@ func TestLookupComplexTypes(t *testing.T) {
 			path: "date",
 			id:   "nanosecond",
 			name: "lookup date.nanosecond",
-			want: "(t: time) -> int", 
-		}, 
+			want: "(t: time) -> int",
+		},
 		{
 			path: "date",
```
fchikwekwe: These types of changes were made by `go fmt`

semantic/lookup.go
```diff
@@ -1,26 +1,26 @@
...
 	flatbuffers "github.com/google/flatbuffers/go"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/libflux/go/libflux"
 	"github.com/influxdata/flux/semantic/internal/fbsemantic"
 )
 
+var byteArr = libflux.EnvStdlib()
+var typeEnv = fbsemantic.GetRootAsTypeEnvironment(byteArr, 0)
+var environment = TypeEnvMap(typeEnv)
```
fchikwekwe: Wanted to create the `TypeEnvironment` and then lookup `Prop`s subsequently.

semantic/lookup.go
```diff
@@ -1,26 +1,26 @@
...
-	// perform a lookup of package identifier in flatbuffers TypeEnvironment
-	prop, err := lookup(env, pkg, name)
-	if err != nil {
-		return MonoType{}, err
-	}
+	prop := environment[pkg][name] // query environment map for prop
+
 	if !prop.V(&table) {
-		return MonoType{}, errors.Newf(codes.Internal, "Prop value is not valid: %v", err)
+		return MonoType{}, errors.New(codes.Internal, "Prop value is not valid")
```
fchikwekwe: There isn't really a pertinent error to add here, unless anyone has a suggestion.

semantic/lookup.go
```diff
@@ -30,70 +30,33 @@ func LookupBuiltInType(pkg, name string) (MonoType, error) {
...
-	polytype := typeAssign.Ty(nil)
-	if polytype.ExprType() != fbsemantic.MonoTypeRow {
-		return nil, errors.Newf(
-			codes.Internal,
-			"Expected PolyType expr to be fbsemantic.MonoTypeRow; found %v",
-			polytype.ExprType(),
-		)
-	}
+func TypeEnvMap(env *fbsemantic.TypeEnvironment) map[string]map[string]*fbsemantic.Prop {
+	envMap := make(map[string]map[string]*fbsemantic.Prop)
```
fchikwekwe: It feels so good to reduce the amount of code for this lookup.

semantic/lookup.go
```diff
@@ -30,70 +30,33 @@ func LookupBuiltInType(pkg, name string) (MonoType, error) {
...
-			if string(obj.Id()) == pkg {
-				return nil
-			}
+		newAssign := new(fbsemantic.TypeAssignment)
+		_ = env.Assignments(newAssign, i) // this call assigns a value to newAssign
+		assignId := string(newAssign.Id())
+		polytype := newAssign.Ty(nil)
+		if !polytype.Expr(&table) {
+			panic(fmt.Errorf("PolyType does not have a MonoType; something went wrong %v", string(polytype.ExprType())))
 		}
```
fchikwekwe: I wanted to keep this panic as descriptive/helpful as possible. I don't really expect that this is a case that we would hit.

semantic/lookup.go
wolffcm: You have a package path and the name, it would be good to put that in the error that you return.

semantic/lookup.go
wolffcm: Makes sense, I agree this is very unlikely to happen.  I think that `string(polytype.ExprType())` will just produce a stringified small integer, like `"1"` or something though.  (FYI, there is a map called `EnumNamesMonoType` that maps it to a readable name). You might just put `newAssign.Id()` in there instead, that would indicate which symbol had the problem.

semantic/lookup.go
```diff
@@ -30,70 +30,33 @@ func LookupBuiltInType(pkg, name string) (MonoType, error) {
...
-func foundProp(row *fbsemantic.Row, obj *fbsemantic.Prop, name string) error {
-	l := row.PropsLength()
-	for i := 0; i < l; i++ {
-		if !row.Props(obj, i) {
-			return errors.Newf(codes.Internal, "package identifier %v not found; last position %v", name, i)
-		} else {
-			if string(obj.K()) == name {
-				return nil
-			}
+		// initialize table before use in row
```
wolffcm: It would be good to assert that `ExprType()` here is actually `Row`, and panic if not.

semantic/lookup.go
```diff
@@ -1,26 +1,26 @@
...
 import (
+	"fmt"
 	flatbuffers "github.com/google/flatbuffers/go"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/libflux/go/libflux"
 	"github.com/influxdata/flux/semantic/internal/fbsemantic"
 )
 
+var byteArr = libflux.EnvStdlib()
```
wolffcm: Global variables should be used very sparingly, so in this situation it would be good to only create one of them.  I would nest the calls here, like this
```
var environment = TypeEnvMap(fbsemantic.GetRootAsTypeEnvironment(libflux.EnvStdlib(), 0))
```
It's harder to read, but I think that fewer globals is still better.

The name `environment` will be visible everywhere in the semantic package, so it would be good to rename it to something that's unlikely to clash, like `stdlibTypeEnvironment` or something.

semantic/lookup.go
fchikwekwe: Yeah, the map is a good idea. I'll add the `assignId` and the `ExprType` to the panic statement.

wolffcm:
Looks great, thanks for making those changes.

PR #2391, opened 2020-01-15
fix: uuid security

lyondhill:
Update to remove dependency on insecure version of uuid library
https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMSATORIGOUUID-72488

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -51,6 +51,8 @@ func (src *FromProcedureSpec) AddTransformation(t execute.Transformation) {
 }
 
 func (src *FromProcedureSpec) Run(ctx context.Context) {
+	// uuid.NewV4 can return an error because of enthropy. We will stick with the previous
+	// behavior of panicing on errors when creating new uuid's
 	id := execute.DatasetID(uuid.Must(uuid.NewV4()))
 
 	if len(src.ts) == 0 {
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -51,7 +51,7 @@ func (src *FromProcedureSpec) AddTransformation(t execute.Transformation) {
 }
 
 func (src *FromProcedureSpec) Run(ctx context.Context) {
-	id := execute.DatasetID(uuid.NewV4())
+	id := execute.DatasetID(uuid.Must(uuid.NewV4()))
 
 	if len(src.ts) == 0 {
 		return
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/pkg/term v0.0.0-20180730021639-bffc007b7fd5 // indirect
 	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
 	github.com/prometheus/common v0.6.0
-	github.com/satori/go.uuid v1.2.0
+	github.com/satori/go.uuid v1.2.1-0.20181028125025-b2ce2384e17b
```
jsternberg: Can we get them to do a release for this? It also looks like a breaking change so I am not sure how they want to handle this.

execute/executetest/source.go
```diff
@@ -51,7 +51,7 @@ func (src *FromProcedureSpec) AddTransformation(t execute.Transformation) {
 }
 
 func (src *FromProcedureSpec) Run(ctx context.Context) {
-	id := execute.DatasetID(uuid.NewV4())
+	id := execute.DatasetID(uuid.Must(uuid.NewV4()))
```
fchikwekwe: I don't know much about how UUID's are generated, but if the docs panic when there's an error, that seems like a good way to do this.

go.mod
fchikwekwe: @jsternberg "them" as in the compute team? Or who?

jsternberg: Whoever manages this dependency. The reason why I'm concerned about this is this could break downstream projects, such as influxdb, that use the older dependency and break their API because it forces them to update. This isn't that big of a deal right now, but I would prefer a numbered version rather than a random hash.

lyondhill: I created a new issue for tracking this work #2392

lyondhill:
Additional work will be tracked here #2392

execute/executetest/source.go
jsternberg: @lyondhill can you add a todo here about answering the question of when this panics? I think @fchikwekwe brings up a good point here and, while I don't want to delay the PR, it is a question I would like an answer to.

lyondhill: Will do.

PR #2394, opened 2020-01-15
fix(stdlib/http): add 169.254/16 range to URL validator

pierwill:
Security fix to prevent connections to private address space.

Test written, but not sure it works as it should. Would love feedback!

### Done checklist
- [x] Test cases written

### DIFF
--- a/dependencies/url/validator.go
+++ b/dependencies/url/validator.go
@@ -46,6 +46,7 @@ func init() {
 		"10.0.0.0/8",     // RFC1918
 		"172.16.0.0/12",  // RFC1918
 		"192.168.0.0/16", // RFC1918
+		"169.254.0.0/16", // RFC3927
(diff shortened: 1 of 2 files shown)


dependencies/url/validator.go
```diff
@@ -46,6 +46,7 @@ func init() {
 		"10.0.0.0/8",     // RFC1918
 		"172.16.0.0/12",  // RFC1918
 		"192.168.0.0/16", // RFC1918
+		"169.254.0.0/16", // RFC3927
 		"::1/128",        // IPv6 loopback
 		"fe80::/10",      // IPv6 link-local
```
fchikwekwe: One more thing that you might want to note here, if there are other address spaces that need to be added here you might want to add them together next time. Appreciate this!

PR #2395, opened 2020-01-15
fix(semantic): require successful lookup of stdlib builtins or panic

fchikwekwe:
`LookupBuiltinType` should panic if lookup fails. This issue adds a validation wrapper that panics if there is a failure. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/semantic/lookup.go
+++ b/semantic/lookup.go
@@ -11,12 +11,23 @@ import (
 
 var stdlibTypeEnvironment = TypeEnvMap(fbsemantic.GetRootAsTypeEnvironment(libflux.EnvStdlib(), 0))
 
-// LookupBuiltInType returns the type of the builtin value for a given
-// Flux stdlib package. Returns an error if lookup fails.
-func LookupBuiltInType(pkg, name string) (MonoType, error) {
-	var table flatbuffers.Table
-	prop := stdlibTypeEnvironment[pkg][name] // query environment map for prop
+type envKey struct {
+	Package string
+	Prop    string
+}
 
+// LookupBuiltinType returns the type of the builtin value for a given
+// Flux stdlib package. Returns an error if lookup fails.
+func LookupBuiltinType(pkg, name string) (MonoType, error) {
+	key := envKey{
+		Package: pkg,
+		Prop:    name,
+	}
+	prop, ok := stdlibTypeEnvironment[key]
+	if !ok {
+		return MonoType{}, errors.Newf(codes.Internal, "Expected to find Prop for %v %v, but Prop was missing.", pkg, name)
+	}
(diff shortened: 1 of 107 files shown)


semantic/lookup.go
```diff
@@ -30,70 +30,44 @@ func LookupBuiltInType(pkg, name string) (MonoType, error) {
...
-	return prop, nil
+	return mt
 }
 
-// foundPackage is a helper function that iterates over type assignments and checks
-// for a given package returning an error if the package is not found
-func foundPackage(env *fbsemantic.TypeEnvironment, obj *fbsemantic.TypeAssignment, pkg string) error {
+// TypeEnvMap creates a global map of the TypeEnvironment
+func TypeEnvMap(env *fbsemantic.TypeEnvironment) map[string]map[string]*fbsemantic.Prop {
+	envMap := make(map[string]map[string]*fbsemantic.Prop)
```
nathanielc: Go lets you use a struct as the key to a map, in most cases using a struct is more efficient than nested maps.

So you could create a struct like this :

```
type envKey struct {
    Package string
    Prop string
}
```

And a map like this:

```
envMap := make(map[envKey]*fbsemantic.Prop)
```


Then you can look it up like this:

```
key := envKey {
    Package: pkg,
    Prop: name,
}

prop := envMap[key]
```

This change isn't necessary but it will be more idiomatic Go.

fchikwekwe: Yeah, that looks much more Go-like. I'll go ahead and make it look more like that

fchikwekwe: ^^ Great suggestion. I feel like the code looks a lot better. I was writing it like it was Python.

semantic/lookup.go
```diff
@@ -11,12 +11,21 @@ import (
...
 // Flux stdlib package. Returns an error if lookup fails.
-func LookupBuiltInType(pkg, name string) (MonoType, error) {
-	var table flatbuffers.Table
-	prop := stdlibTypeEnvironment[pkg][name] // query environment map for prop
+func LookupBuiltinType(pkg, name string) (MonoType, error) {
+	key := envKey{
+		Package: pkg,
+		Prop:    name,
+	}
+	prop := stdlibTypeEnvironment[key] // query environment map for prop
```
nathanielc: You should check if the prop was missing so an error can be returned.

```
prop, ok := stdlibTypeEnvironment[key] 
if !ok {
    // return error
}
```

fchikwekwe: Yeah that's a good idea. I figured it didn't matter since we are panicking in most places where we use this function, but you're right its more consistent to return an error here.

PR #2412, opened 2020-01-17
feat(flux): make Eval() and EvalAST() use libflux for parsing and analysis

wolffcm:
The test `flux/compiler_test.go` tests these new methods.

I also made sure that the interpreter tests now pass:
- Some of the tests were verifying that type inference produces errors.
    Since the interpreter no longer runs inference itself, I moved these tests
    into the Rust code.
- Some of the test cases did not pass due to other issues.  I skipped them with the issue number.

### DIFF
--- a/libflux/src/flux/semantic/tests.rs
+++ b/libflux/src/flux/semantic/tests.rs
@@ -2594,6 +2594,77 @@ fn identity_function() {
         ],
     }
 }
+
+#[test]
+fn call_expr() {
+    // missing parameter
+    test_infer_err! {
+        src: r#"
+            plusOne = (x) => x + 1.0
+            plusOne()
+        "#,
+    }
+    // missing pipe
+    test_infer_err! {
+        src: r#"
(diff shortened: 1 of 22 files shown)


libflux/src/flux/semantic/tests.rs
```diff
@@ -2594,6 +2594,77 @@ fn identity_function() {
...
+    // but it's okay.
+    test_infer! {
+        src: r#"
+				f = (x, y) => x(arg: y)
+				f(x: (arg=<-) => arg, y: 0)
+        "#,
+        exp: map![
+            "f" => "forall [t2, t4] (x:(arg:t2) -> t4, y:t2) -> t4",
+        ]
+    }
```
jlapacik: nit, formatting is a little off.

wolffcm: Yeah, in Go code is formatted with tabs, and in Rust it's spaces, I fixed it.

PR #2418, opened 2020-01-23
generate rust dependencies as a by-product of building

adriandt:
1. Create the .deps file in the same rule as the cargo invocation. This creates
the deps as a by-product of building.

2. Include the depended-on files as targets in the dependencies files.

This method has the advantage that when files disappear or are renamed the
make does not fail due to a missing file. Instead, the target that creates the
deps file is re-run, updating the depencies in the process.

This is ths same method used by autotools for dependency management.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,45 +72,39 @@ internal/scanner/scanner.gen.go: internal/scanner/gen.go internal/scanner/scanne
 
 libflux: libflux/target/debug/libflux.a libflux/target/debug/liblibstd.a
 
-# Build the rust static library. Afterwards, fix the .d file that
-# rust generates so it references the correct targets.
-# The unix sed, which is on darwin machines, has a different
-# command line interface than the gnu equivalent.
+# Build the rust static libraries. Afterwards, fix the .d file that rust
+# generates so it references the correct targets. The unix sed, which is on
+# darwin machines, has a different command line interface than the gnu
+# equivalent.
+
+# The dependency file produced by Rust appears to be wrong and uses absolute
+# paths while we use relative paths everywhere. So we need to do some post
+# processing of the file to ensure that the dependencies we load are correct.
+FIX_DEPS = -e "s@${CURDIR}/@@g" -e "s@debug/debug@debug@g" -e "s@\\.dylib@.a@g" -e "s@\\.so@.a@g"
+
+# This second pass over the dependency file addes each file that is depended on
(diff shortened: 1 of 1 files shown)


adriandt:
Testing on ubuntu 16.04. Needs to be tested on Darwin.

Hmmm, what is a semantic commit?

nathanielc:
>  Hmmm, what is a semantic commit?

Have a look here https://www.conventionalcommits.org/en/v1.0.0/ it has a few examples part way down the page. We auto generate the change log based on the information in the commit message. Both the PR title and commit message must follow the format as github will use either in different situations. 

The change looks good. I am going to ping @jsternberg to review this as he has more context than me. 

As for adding the `libflux/scanner.c` as dependency I think that makes sense to do. Yes it will require that users have ragel installed, but it should only run if a change was made to the scanner.rl file, in which case ragel is needed.

Makefile
```diff
@@ -72,46 +72,40 @@ internal/scanner/scanner.gen.go: internal/scanner/gen.go internal/scanner/scanne
...
-# If the .d file does not exist, then the .a file above also
-# does not exist so the dependencies don't matter. If the .d file
-# exists, this will never get called or, at a minimum, it won't modify
-# the files at all. This allows the target below to depend on this
-# file without the file necessarily existing and it will force
-# post-processing of the file if the .d file is newer than our
-# post-processed .deps file.
-libflux/target/debug/libflux.d:
-
-libflux/target/debug/libflux.deps: libflux/target/debug/libflux.d
```
jsternberg: This part needs to be added back in and the change you made should be made to this target instead. The reason this target is important is because it allows cargo to be run outside of make. If you remove this, then running cargo manually and then running make causes the deps file not to be updated properly.

adriandt: Ah yes that's a good feature to have. I wasn't sure why it was that way, but now it makes sense.

adriandt:
This PR lives on in #2422. I just learned that if you `git push -f` a branch that is part of a PR, github will update the PR. Next time I'll do that.

adriandt:
> As for adding the `libflux/scanner.c` as dependency I think that makes sense to do. Yes it will require that users have ragel installed, but it should only run if a change was made to the scanner.rl file, in which case ragel is needed.

@nathanielc The only problem is that when checking out, the source can get a newer timestamp than the destination. So just the fact that target is there is not enough to prevent the call to ragel.

PR #2419, opened 2020-01-23
refactor(runtime): refactor the builtin registration

jsternberg:
The builtin registration was previously happening incrementally so the
order of how things got registered was uncertain. The registration has
now been changed so it encapsulates the work in a runtime builder struct
and it constructs the final runtime when finalize is called instead of
gradually as the packages are registered.

This also modifies the scope so there is a special `values.Option` which
wraps another value. This signals that a variable was defined in a
package as an option. This allows us to evaluate a flux file and
determine from the scope which values were defined as options and which
as normal variables.

The refactor removes most of the need for the global variables by
encapsulating those globals into a single struct which can be used as if
it were not a global variable.

Unfortunately, this is not completely done due to a few other pieces
that need to be fixed before evaluating standard library packages works
correctly.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/package.go
+++ b/interpreter/package.go
@@ -5,25 +5,36 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
 )
 
+// Package is an importable package that can be used from another
+// section of code. The package itself cannot have its attributes
+// modified after creation, but the options may be changed.
 type Package struct {
-	name        string
-	object      values.Object
-	options     values.Object
+	// name is the name of the package.
+	name string
+
+	// object contains the object properties of this package.
+	object values.Object
+
+	// options contains the option overrides for this package.
+	// Values cannot exist in here unless they also exist in
+	// the underlying object.
+	options map[string]values.Value
+
+	// sideEffects contains the side effects caused by this package.
+	// This is currently unused.
 	sideEffects []SideEffect
 }
 
 func NewPackageWithValues(name string, obj values.Object) *Package {
-	if obj == nil {
-		obj = values.NewObjectWithValues(nil)
-	}
 	return &Package{
-		name:    name,
-		options: values.NewObjectWithValues(nil),
-		object:  obj,
+		name:   name,
+		object: obj,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


interpreter/package.go
```diff
@@ -60,23 +80,29 @@ func (p *Package) Type() semantic.MonoType {
...
+		p.options = make(map[string]values.Value)
+	}
+	p.options[name] = v
 }
 func (p *Package) Len() int {
 	return p.object.Len()
 }
 func (p *Package) Range(f func(name string, v values.Value)) {
 	p.object.Range(f)
-	p.options.Range(f)
```
nathanielc: This needs to iterator over the options as well.

values/option.go
```diff
@@ -0,0 +1,10 @@
+package values
+
+type Option struct {
```
nathanielc: Can you add a comment explaining how this works? Why does a value need to know if its an option?

interpreter/package.go
jsternberg: I changed the package so the options are part of the object. Maybe that was incorrect. It seemed to me like when an option is defined, it is defined as a variable that can also be used in the same context. Instead of being in separate objects, the option is wrapped in a `values.Option` to denote it as something special.

So when you do:

```
option prefix = "influxdata_"
my_ident = (name) => prefix + name
```

The prefix becomes part of the object and can be accessed as if it were a normal variable, but it gets wrapped with `values.Option` so `SetOption` will work on it.

values/option.go
jsternberg: Yes. I'll give the explanation above.

interpreter/package.go
```diff
@@ -60,23 +73,32 @@ func (p *Package) Type() semantic.MonoType {
...
 }
 func (p *Package) Set(name string, v values.Value) {
-	p.object.Set(name, v)
+	panic(errors.New(codes.Internal, "package members cannot be modified"))
 }
 func (p *Package) SetOption(name string, v values.Value) {
-	p.options.Set(name, v)
+	// TODO(jsternberg): Setting an invalid option on a package wasn't previously
+	// an error so it continues to not be an error. We should probably find a way
+	// to make it so setting an invalid option is an error.
```
nathanielc: what is an invalid option?

interpreter/package.go
nathanielc: I see, does that mean we need to check the `options` map for overrides during this Range function? Basically copy the `Get` logic into the method as well?

interpreter/package.go
jsternberg: For whatever reason, the old code seemed to allow this:

```
package foo

option prefix = "default_prefix"
```

```
package main

import "foo"

option foo.sprefix = "my_custom_prefix"
```

The option is spelled wrong in the second one. This seemed like a bug to me, but the old interpreter allowed it so I had no plans to change it in this PR.

interpreter/package.go
jsternberg: We do. I seem to have made a mistake here.

interpreter/package.go
nathanielc: Agreed. Creating new options is allowed, but I agree creating one from outside the package seems odd. We can disallow that in a separate change.

PR #2420, opened 2020-01-23
fix(libflux): expose the builtin polytypes when analyzing a stdlib package

jsternberg:
This exposes the builtin polytypes when analyzing a package in the
standard library. It also flattens the builtin map to make it easier to
index. The old code that iterated through flattened the map anyway so it
is easier to read and index if we just refer to each package by their
full package name instead of by each component.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -1,183 +1,185 @@
+use crate::semantic::fresh::{Fresh, Fresher};
+use crate::semantic::import::Importer;
+use crate::semantic::parser::parse;
 use maplit::hashmap;
-use std::collections::hash_map;
+use std::collections::hash_map::Iter;
 use std::collections::HashMap;
-use std::iter::Iterator;
-use std::vec::Vec;
 
 pub struct Builtins<'a> {
-    pkgs: HashMap<&'a str, Node<'a>>,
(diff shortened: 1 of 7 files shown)


jsternberg:
I don't think there is, but the change is also only relevant to the work in this branch so there's not a lot of reason to merge it to master. We would then have to merge master to this branch and this branch is our de facto master for the next couple of weeks while we finish the algorithm w stuff. At least that's what my thought process is.

jlapacik:
I think I would prefer it land in master if possible. That way we ensure that we don't inadvertently regress or break anything.

jsternberg:
This is now targeting master.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -1,183 +1,189 @@
...
+                    h.insert(
+                        name.to_string(),
+                        parse(expr).unwrap(),
+                    );
+                }
+            }
+            _ => {},
+        }
+        return h
+    }
```
jlapacik: This looks mostly correct, although I think we'll want to update any type variables with fresh ones before we insert into the importer. This can be accomplished using a combination of `max_tvar()` and methods defined in `fresh.rs`

PR #2423, opened 2020-01-23
fix: create unique c files to force cgo to consider code changed

adriandt:
Please do not merge, looking for some feedback on this idea. I've been having a problem where cmd/flux/flux does not rebuild after the rust code changes. It seems cgo needs some source code changes to consider the package as changed. So I'm making a unique file using md5 sum. Wonder what others think of this idea ... or if there is a better solution. Seems to be the only thing that works for me, aside from a full blown clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,9 +78,11 @@ libflux: libflux/target/debug/libflux.a libflux/target/debug/liblibstd.a
 # command line interface than the gnu equivalent.
 libflux/target/debug/libflux.a:
 	cd libflux && $(CARGO) build -p flux $(CARGO_ARGS)
+	echo "const char uniq_libflux[] = \"`cat $@ | md5sum | awk '{print $$1;}'`\";" > libflux/go/libflux/libflux.c
 
 libflux/target/debug/liblibstd.a:
 	cd libflux && $(CARGO) build -p libstd $(CARGO_ARGS)
+	echo "const char uniq_liblibstd[] = \"`cat $@ | md5sum | awk '{print $$1;}'`\";" > libflux/go/libflux/liblibstd.c
 
 libflux/go/libflux/flux.h: libflux/include/influxdata/flux.h
 	$(GO_GENERATE) ./libflux/go/libflux
--- a/libflux/go/libflux/.gitignore
+++ b/libflux/go/libflux/.gitignore
@@ -1,0 +1,2 @@
+/libflux.c
+/liblibstd.c


nathanielc:
@adriandt Thanks! I had a similar idea, I'll try this out to see if it works for me.

jsternberg:
I like the idea, but I have a bit of a different way to implement it that I think will be more reliable.

I think it would help if we embedded the hash of the sources into the binary itself and then attempted to reference that within the Go code. The reason is because then it is possible to use the tools separate from the Makefile and still catch things that go wrong. So my idea is that we would have the Rust `build.rs` file hash all of the input files and produce a C file inside of the binary that would amount to something like this.

```
const int INFLUXDATA_FLUX_REVISION_sha1sum = 1;
```

We would then add a similar script to `go generate` that would generate this same sha1 sum and we would add a reference to this constant in the Go code. So something like:

```
const int INFLUXDATA_FLUX_REVISION_sha1sum;
```

We would then reference this symbol inside of the Go code somewhere. If the symbol isn't defined in the Rust code, we get a linker error.

The reason why I like this is mostly because it gives an ABI guarantee and it would integrate with the common tooling for each language. For Go, `go generate` would be used and this would get caught in CI so we could commit the result.

The method in this works for getting Go to rebuild, but it doesn't guarantee that the thing it's linking to is actually correct especially when it comes to downstream users such as InfluxDB.

adriandt:
@jsternberg excellent! Cargo and go build are both new to me, so I'm glad to hear it can be done from inside those systems. I'm surprised that cgo lets you link objects, but doesn't allow you to depend on them in order to get your package rebuilt when they change. It seems to function as though anything from outside the system is immutable.

Makefile
```diff
@@ -78,9 +78,11 @@ libflux: libflux/target/debug/libflux.a libflux/target/debug/liblibstd.a
 # command line interface than the gnu equivalent.
 libflux/target/debug/libflux.a:
 	cd libflux && $(CARGO) build -p flux $(CARGO_ARGS)
+	echo "const char uniq_libflux[] = \"`cat $@ | md5sum | awk '{print $$1;}'`\";" > libflux/go/libflux/libflux.c
 
 libflux/target/debug/liblibstd.a:
 	cd libflux && $(CARGO) build -p libstd $(CARGO_ARGS)
+	echo "const char uniq_liblibstd[] = \"`cat $@ | md5sum | awk '{print $$1;}'`\";" > libflux/go/libflux/liblibstd.c
```
fchikwekwe: I had some trouble downloading `md5sum` so I used `shasum` here when I was borrowing your change. Otherwise, thanks for this. Improves workflow for sure!

adriandt: Great! Yeah `shasum` is probably the better choice for this hack. I'm noticing now that out of the two, only `shasum` comes out of the box on both linux and mac.

adriandt:
Hey @paulh, tagging you here after hearing your standup today. Before the all-hands I was poking around with flux and I had issues getting go binaries to rebuild after changing the rust library. The CGO system won't rebuild unless source changes. It's not enough for the .a to change. My (hackish) solution is to make make a unique C file by hashing the .a.

paulh:
wrong paul...

adriandt:
Closing this PR, as the desire is to do this kind of thing from within CGO and cargo build. Not much interest in depending on the Makefile.

PR #2429, opened 2020-01-24
refactor(execute): execute tests are now passing

jsternberg:
The unit tests for the execute package are now passing. This required
some fixes to filter which weren't previously capable of being tested
because the filter test cases couldn't be run because of the builtins
not being registered yet.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/executetest/compile.go
+++ b/execute/executetest/compile.go
@@ -1,0 +1,36 @@
+package executetest
+
+import (
+	"testing"
+
+	"github.com/influxdata/flux/semantic"
+)
+
+// FunctionExpression will take a function expression as a string
+// and return the *semantic.FunctionExpression.
+//
+// This will cause a fatal error in the test on failure.
+func FunctionExpression(t testing.TB, source string) *semantic.FunctionExpression {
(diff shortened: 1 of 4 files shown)


semantic/monotype.go
```diff
@@ -160,11 +160,12 @@ func monoTypeFromVar(v *fbsemantic.Var) MonoType {
...
-	f, ok := tbl.(*fbsemantic.Fun)
-	if !ok {
-		return nil, errors.New(codes.Internal, "MonoType is not a function")
-	}
+	f := tbl.(*fbsemantic.Fun)
 	return f, nil
+	// if !ok {
+	// 	return nil, errors.New(codes.Internal, "MonoType is not a function")
+	// }
+	// return f, nil
```
nathanielc: If this is no longer needed can you remove it?

jsternberg: I removed this so I could force a panic and find the issue. In the future, I think it would be good to force that panic when using this incorrectly, but I readded it because that isn't the present.

PR #2430, opened 2020-01-24
feat(libflux): handle multi-file packages

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/go/libflux/analyze.go
+++ b/libflux/go/libflux/analyze.go
@@ -81,8 +81,6 @@ func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
 func EnvStdlib() []byte {
 	var buf C.struct_flux_buffer_t
 	C.flux_get_env_stdlib(&buf)
-
 	defer C.flux_free(buf.data)
-
 	return C.GoBytes(buf.data, C.int(buf.len))
 }
--- a/libflux/go/libflux/analyze_test.go
+++ b/libflux/go/libflux/analyze_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+
 	"github.com/influxdata/flux/libflux/go/libflux"
 )
 
@@ -56,3 +57,4 @@ func TestAnalyze(t *testing.T) {
 		})
 	}
 }
+
--- a/libflux/go/libflux/flux.h
+++ b/libflux/go/libflux/flux.h
@@ -33,6 +33,11 @@ struct flux_ast_pkg_t;
 // of the query.
 struct flux_ast_pkg_t *flux_parse(const char *);
 
+// flux_merge_ast_pkgs merges the files of a given input AST package into the file vector of a
+// given output AST package. This function borrows the packages, but it does not own them. The
+// caller of this function still needs to free the package memory on the Go side.
+struct flux_error_t *flux_merge_ast_pkgs(struct flux_ast_pkg_t *, struct flux_ast_pkg_t *);
+
 // flux_parse_json will take in a JSON string for an AST package
 // and populate its second pointer argument with a pointer to an
 // AST package.
@@ -87,3 +92,4 @@ struct flux_error_t *flux_semantic_marshal_fb(struct flux_semantic_pkg_t *, stru
 #endif
 
 #endif
+
--- a/libflux/src/flux/lib.rs
+++ b/libflux/src/flux/lib.rs
@@ -231,3 +231,153 @@ pub unsafe extern "C" fn flux_error_str(err: *mut flux_error_t) -> *mut c_char {
 pub unsafe extern "C" fn flux_free(err: *mut c_void) {
     Box::from_raw(err);
 }
+
+/// flux_merge_ast_pkg_files merges the files of a given input ast::Package into the file
+/// vector of an output ast::Package.
+#[no_mangle]
+pub unsafe extern "C" fn flux_merge_ast_pkgs(
+    out_pkg: *mut flux_ast_pkg_t,
+    in_pkg: *mut flux_ast_pkg_t,
+) -> *mut flux_error_t {
+    // Do not change ownership here so that Go maintains ownership of packages
+    let out_pkg = &mut *(out_pkg as *mut ast::Package);
+    let in_pkg = &mut *(in_pkg as *mut ast::Package);
+
+    match merge_packages(out_pkg, in_pkg) {
+        None => std::ptr::null_mut(),
+        Some(err) => {
+            let err_handle = ErrorHandle { err: Box::new(err) };
+            Box::into_raw(Box::new(err_handle)) as *mut flux_error_t
+        }
+    }
+}
+
+/// merge_packages takes an input package and an output package, checks that the package
+/// clauses match and merges the files from the input package into the output package. If
+/// package clauses fail validation then an option with an Error is returned.
+pub fn merge_packages(out_pkg: &mut ast::Package, in_pkg: &mut ast::Package) -> Option<Error> {
+    let pkg_clause = match &out_pkg.files[0].package {
+        Some(clause) => {
+            &clause.name.name
+        }
+        None => return Some(Error::from("output package does not have a package clause")),
+    };
+
+    // Check that all input files have a package clause that matches the output package.
+    for file in &in_pkg.files {
+        let file_clause = match &file.package {
+            Some(clause) => &clause.name.name,
+            None => return Some(Error::from("current file does not have a package clause")),
+        };
+
+        if pkg_clause != file_clause {
+            return Some(Error::from(format!(
+                "file's package clause: {} does not match package output package clause: {}",
+                file_clause, pkg_clause
+            )));
+        }
+    }
+    out_pkg.files.append(&mut in_pkg.files);
+    None
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{ast, merge_packages};
+
+    #[test]
+    fn ok_merge_multi_file() {
+        let in_script = "package foo\na = 1\n";
+        let out_script = "package foo\nb = 2\n";
+
+        let in_file = crate::parser::parse_string("test", in_script);
+        let out_file = crate::parser::parse_string("test", out_script);
+        let mut in_pkg = ast::Package {
+            base: Default::default(),
+            path: "./test".to_string(),
+            package: "foo".to_string(),
+            files: vec![in_file.clone()],
+        };
+        let mut out_pkg = ast::Package {
+            base: Default::default(),
+            path: "./test".to_string(),
+            package: "foo".to_string(),
+            files: vec![out_file.clone()],
+        };
+        merge_packages(&mut out_pkg, &mut in_pkg);
+        let got = out_pkg.files;
+        let want = vec![out_file, in_file];
+        assert_eq!(want, got);
+    }
+
+    #[test]
+    fn ok_no_in_pkg() {
+        let out_script = "package foo\nb = 2\n";
+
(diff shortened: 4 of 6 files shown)


libflux/go/libflux/create_type_env.go
```diff
@@ -0,0 +1,17 @@
...
+
+
+// EnvStdlib takes care of creating a flux_buffer_t, passes the buffer to
+// the Flatbuffers TypeEnvironment and then takes care of freeing the data
+func EnvStdlib() []byte {
+	var buf C.struct_flux_buffer_t
+	C.flux_get_env_stdlib(&buf)
+	defer C.flux_free(buf.data)
+	return C.GoBytes(buf.data, C.int(buf.len))
+}
```
nathanielc: Why was this moved to its own file?

libflux/go/libflux/flux.h
```diff
@@ -75,6 +75,10 @@ void flux_get_env_stdlib(struct flux_buffer_t *);
 // flux_ast_pkg_t* argument, so it should not be reused after calling this function.
 struct flux_error_t *flux_analyze(struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
 
+// flux_ast_pkg_t merges the files of a given input AST package into the file vector of a
+// given output AST package. If there are errors, then a panic will occur.
```
nathanielc: We can't safely panic through the CGO boundary, it looks like the function signature is to return the error. Is this comment just out of date?

libflux/go/libflux/flux.h
```diff
@@ -75,6 +75,10 @@ void flux_get_env_stdlib(struct flux_buffer_t *);
...
+struct flux_error_t *flux_merge_ast_pkg_files(struct flux_ast_pkg_t *, struct flux_ast_pkg_t *);
```
nathanielc: We should name the args to this function since its not clear which is the input and which is the output without names.

libflux/go/libflux/merge.go
```diff
@@ -0,0 +1,18 @@
...
+
+// #cgo CFLAGS: -I.
+// #cgo LDFLAGS: -L. -llibstd
+// #include "flux.h"
+// #include <stdlib.h>
+import "C"
+
+// Merge packages merges the files of a given input package into a given
+// output package. If there are unrecoverable errors, a panic may occur.
+func MergePackages(inPkg *ASTPkg, outPkg *ASTPkg) *ASTPkg {
```
nathanielc: Should this return an error? Where does the panic come from? I don't see a call to `panic`.

libflux/src/flux/lib.rs
```diff
@@ -223,3 +223,109 @@ pub unsafe extern "C" fn flux_error_str(err: *mut flux_error_t) -> *mut c_char {
 pub unsafe extern "C" fn flux_free(err: *mut c_void) {
     Box::from_raw(err);
 }
+
+/// flux_merge_ast_pkg_files merges the files of a given in_pkg into the file
+/// vector of a given out_pkg. If there are problems with the package identifiers,
+/// then a panic may occur.
+#[no_mangle]
+pub unsafe extern "C" fn flux_merge_ast_pkg_files(
```
nathanielc: Ah, I think I understand my confusion. A panic from Rust will not result in a panic in Go. A panic in Rust will just corrupt the memory of the Go process(which may or may not panic the Go side of things). I am not sure what will happen then. We cannot panic from this Rust function. We must ensure that all errors are returned explicitly. Perhaps this is related to the bug you are looking for.

libflux/src/flux/lib.rs
```diff
@@ -223,3 +223,109 @@ pub unsafe extern "C" fn flux_error_str(err: *mut flux_error_t) -> *mut c_char {
...
+    match merge_files(in_pkg, out_pkg) {
+        Ok(_) => std::ptr::null_mut(),
+        Err(err) => {
+            let err_handle = ErrorHandle { err: Box::new(err) };
+            Box::into_raw(Box::new(err_handle)) as *mut flux_error_t
+        }
+    }
+}
+
+pub fn merge_files(in_pkg: ast::Package, out_pkg: ast::Package) -> Result<Vec<ast::File>, Error> {
```
nathanielc: nit, can you rename this to `merge_packages`?

libflux/go/libflux/flux.h
fchikwekwe: Yeah this comment is out of date.

libflux/go/libflux/merge.go
fchikwekwe: Yeah, another comment out of date.

libflux/src/flux/lib.rs
fchikwekwe: Yeah, there aren't any panics anymore. I'm just propagating errors now. I was panicking originally until I started trying to glue the Go and Rust sides together.

libflux/go/libflux/analyze_test.go
```diff
@@ -56,3 +58,170 @@ func TestAnalyze(t *testing.T) {
...
+
+	err := libflux.MergePackages(pkg1, pkg2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	gotFB, err := pkg2.MarshalFB()
+	got := ast.DeserializeFromFlatBuffer(gotFB)
+
+	if err != nil {
```
nathanielc: nit, this error check should come before the call to DesierlizFromFlatBuffer.

libflux/go/libflux/analyze.go
```diff
@@ -76,13 +76,31 @@ func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
 	return p, nil
 }
 
+// Merge packages merges the files of a given input package into a given output package.
+// This function borrows the input and output packages, but does not own them. Memory
+// must still be freed by the caller of this function.
+func MergePackages(inPkg *ASTPkg, outPkg *ASTPkg) error {
```
wolffcm: I think this function would better located in `parser.go`, since it's dealing with ASTs rather than semantic graphs.

libflux/go/libflux/analyze_test.go
```diff
@@ -2,6 +2,8 @@ package libflux_test
 
 import (
 	"errors"
+	"github.com/influxdata/flux/ast"
```
wolffcm: *nit* these imports are out of order.

libflux/go/libflux/analyze_test.go
fchikwekwe: true, good catch.

libflux/go/libflux/analyze_test.go
```diff
@@ -56,3 +58,170 @@ func TestAnalyze(t *testing.T) {
...
+		Files: []*ast.File{
+			{
+				Name:     "",
+				Metadata: "parser-type=rust",
+				Package: &ast.PackageClause{
+					Name: &ast.Identifier{Name: "foo"},
+				},
+				Body: []ast.Statement{
+					&ast.VariableAssignment{
+						ID:   &ast.Identifier{Name: "b"},
```
wolffcm: Hmm, given the way the inputs are arranged, I would have expected the file with `a = 1` to appear first, and `b = 1` to be second.  I wonder if `MergePackages` should just append all the files in the second argument to the files in the first argument?  This way the order of the files would be preserved.

libflux/go/libflux/analyze_test.go
```diff
@@ -56,3 +58,170 @@ func TestAnalyze(t *testing.T) {
...
+	want := &ast.Package{
+		Package: "foo",
+		Files: []*ast.File{
+			{
+				Name:     "",
+				Metadata: "parser-type=rust",
+				Package: &ast.PackageClause{
+					Name: &ast.Identifier{Name: "foo"},
+				},
+				Body: []ast.Statement{
```
wolffcm: I'm confused by how this test function is named, since there does seem to be some input files here.  I wonder if you parsed the empty string `""`, is that what you're trying to test?

libflux/go/libflux/flux.h
```diff
@@ -75,6 +75,11 @@ void flux_get_env_stdlib(struct flux_buffer_t *);
 // flux_ast_pkg_t* argument, so it should not be reused after calling this function.
 struct flux_error_t *flux_analyze(struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
 
+// flux_merge_ast_pkgs merges the files of a given input AST package into the file vector of a
+// given output AST package. This function borrows the packages, but it does not own them. The
+// caller of this function still needs to free the package memory on the Go side.
+struct flux_error_t *flux_merge_ast_pkgs(struct flux_ast_pkg_t *, struct flux_ast_pkg_t *);
```
wolffcm: *nit* I would move this prototype closer to the other functions that deal with `flux_ast_pkg_t*` (like parsing) just to keep related functions near to each other.

libflux/go/libflux/analyze_test.go
```diff
@@ -56,3 +58,170 @@ func TestAnalyze(t *testing.T) {
...
+	if !cmp.Equal(got, want, asttest.IgnoreBaseNodeOptions...) {
+		t.Errorf(
+			"MergePackages unexpected packages -want/+got:\n %v",
+			cmp.Diff(want, got, asttest.IgnoreBaseNodeOptions...),
+		)
+	}
+}
+
+func TestNoOutFiles(t *testing.T) {
+	pkg1 := libflux.Parse(`
```
wolffcm: There's some boilerplate here that's repeated for each test method, you could use an array of test cases and loop over them to be more concise.

libflux/go/libflux/analyze_test.go
```diff
@@ -56,3 +58,170 @@ func TestAnalyze(t *testing.T) {
...
+		},
+	}
+
+	if !cmp.Equal(got, want, asttest.IgnoreBaseNodeOptions...) {
+		t.Errorf(
+			"MergePackages unexpected packages -want/+got:\n %v",
+			cmp.Diff(want, got, asttest.IgnoreBaseNodeOptions...),
+		)
+	}
+}
```
wolffcm: It would be good to have some tests for the errors that can happen, like when package names do not match, etc.

libflux/go/libflux/analyze_test.go
fchikwekwe: Well parsing the empty string actually throws an error on the rust side because that package would have no package clause. 

Perhaps these "empty packages" tests are obsolete now that I've eliminated the logic that I wrote for the `ParseDirToHandle` function in my last iteration.

fchikwekwe: Yeah, I think I'm just gonna replace those two tests. They don't really make sense anymore.

libflux/go/libflux/analyze_test.go
fchikwekwe: Yes,  I did try that but it was causing an issue in the test information. I wasn't sure if conciseness was worth spending time on another bug.

libflux/go/libflux/analyze_test.go
fchikwekwe: I have some test like that on the Rust side. I can add some similar tests here if that would be helpful.

wolffcm: I think it would be good to test the error path.  There is some code on the Go side that will not get exercised otherwise.

libflux/go/libflux/analyze_test.go
fchikwekwe: Nevermind. Fixed this.

libflux/go/libflux/analyze_test.go
fchikwekwe: I made some changes to the test file. Let me know if you know of any other cases that I should also cover.

Issue #2425, opened 2020-01-23
aggregateWindow not compatible with derivative


russorat:
I noticed a few graphs on the default system dashboard were empty after updating to flux v0.59.4 in influxdb https://github.com/influxdata/influxdb/commit/4fe3aff9a6b92cacc1624720383f59203eeb334b: 
![image](https://user-images.githubusercontent.com/4805997/73031291-8f122d80-3df0-11ea-9f67-403abce84735.png)

The queries are similar and look like this:
```
from(bucket: v.bucket)
  |> range(start: v.timeRangeStart)
  |> filter(fn: (r) => r._measurement == "net")
  |> filter(fn: (r) => r._field == "bytes_recv" or r._field == "bytes_sent")
  |> aggregateWindow(every: v.windowPeriod, fn: last, createEmpty: false)
  |> derivative(unit: 1s, nonNegative: false)
  |> yield(name: "derivative")
```

Removing either `aggregateWindow` or `derivative` allows data to be returned.

It definitely used to work, but something changed and now it doesn't. There are no errors or messages to the user other than no results. Changing the `unit` or `every` seems to have no effect.

jsternberg:
This is probably related to a problem in the window optimizations. We'll revert that change and reevaluate it so we can hopefully reintroduce the change.

It seems like we are missing a test case to ensure that `aggregateWindow` will output the points in the same order.

jsternberg:
This is related to the window optimizations, but not window directly. Derivative itself is broken. When derivative reads a table with multiple buffers, it will always consume the first row and then compute the derivative from the second row onwards. The functionality is supposed to be that derivative consumes the first row from the stream of buffers, but it consumes the first row per buffer.

This wasn't previously an issue because the old implementations would aggregate everything into a single buffer. The optimization doesn't do this because the copying is expensive.

We'll need to add a test case for this workflow and we'll need to fix derivative. Once derivative and the other related transformations are fixed, we'll be able to reintroduce the window optimization.

jsternberg:
#2298

xuoguoto:
Is [this](https://community.influxdata.com/t/duplicate-timestamp-after-aggregatewindow/20777/2) issue I am facing related to this bug?

PR #2431, opened 2020-01-24
revert: "perf(stdlib): optimize window transformation (#2312)"

jsternberg:
This reverts commit 922f5311ca59f8a228077f57574ffc4d65bbd716.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
This was reverted because window output order changes and it breaks uses of `aggregateWindow` paired with derivative. We will need to evaluate how to more forward with this optimization.

jsternberg:
The window change seems to have caused some kind of problem with the time ordering when it is used with `aggregateWindow`. This reverts the window change to go back to the old behavior. In the future, we'll figure out why this didn't merge those together correctly and reintroduce this change.

#2425

PR #2436, opened 2020-01-24
fix: libflux json deserialization uses type properly

jsternberg:
The libflux json deserialization now uses type properly instead of the
untagged property. That causes the enums to consult the type of the enum
instead of trying to do pattern matching.

A consequence of this though is that parts of the AST that hard-code a
specific type, not the enum, won't have their type serialized. But, this
doesn't matter because since the type is hard-coded in both the Go and
Rust AST's, the correct type gets deserialized anyway. The type was
only needed for interfaces anyway.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/flux/ast/mod.rs
+++ b/libflux/src/flux/ast/mod.rs
@@ -101,34 +101,56 @@ where
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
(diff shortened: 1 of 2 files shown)


jsternberg:
@rockstar I agree with your concern. I thought the same thing. I've merged this, but we will have to return to add additional tests for this.

PR #2439, opened 2020-01-24
fix: use LocalRange in compile tests

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compile_test.go
+++ b/compile_test.go
@@ -25,7 +25,7 @@ func TestEval(t *testing.T) {
 			"f": "(x: int) -> int",
 			"y": "42",
 		}
-		scope.Range(func(k string, v values.Value) {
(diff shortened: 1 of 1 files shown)


compile_test.go
```diff
@@ -25,7 +25,7 @@ func TestEval(t *testing.T) {
 			"f": "(x: int) -> int",
 			"y": "42",
 		}
-		scope.Range(func(k string, v values.Value) {
+		scope.LocalRange(func(k string, v values.Value) {
```
nathanielc: what does this fix?

jlapacik: It gets `compile_test.go` to pass. Previously the test would range over the entire prelude, erroring when it found something other than `f` or `y`.

PR #2440, opened 2020-01-24
fix(interpreter): fix package initialization so that stdlib tests can run

nathanielc:
With this change at least one stdlib end to end test can run, it does not pass but it runs without a panic

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -58,7 +58,7 @@ func evalHandle(ctx context.Context, h *libflux.ASTPkg, opts ...ScopeMutator) ([
 		return nil, nil, err
 	}
 
-	itrp := interpreter.NewInterpreter(nil)
+	itrp := interpreter.NewInterpreter(interpreter.NewPackage(""))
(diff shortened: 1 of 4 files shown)


jsternberg:
I don't think this is going to run properly. The problem is that function keeps a reference to package which doesn't exist and can't be used. I'm trying a refactor that will simplify this process so we can remove that code.

nathanielc:
@jsternberg Sounds good, the reason I needed this was the `function` type was panicking because it didn't have a package and I traced it all the way out. So if there is different more appropriate behavior for an `interpreter.function` that doesn't have a package than that makes sense.

jsternberg:
Yea, the reason why it was panicking there is the current implementation has the function keep a reference to the package. The refactor made it so the package was no longer tied to the interpreter. I just didn't want to change the function signature. That means the package associated with the function is an incorrect value and I think that breaks options because it means that functions don't bind to the correct option value.

nathanielc:
Closing in favor of #2441

PR #2441, opened 2020-01-24
refactor(interpreter): modify the interpreter to eval imported packages each time

jsternberg:
The previous method for the interpreter was to eval each package and
then keep the package for future use. But each usage involved creating a
clone of the package and it would involve fixing the objects to refer to
the new package objects in order to keep things like options using the
original reference value.

This modifies the workflow so imports are evaluated each time instead of
attempting to share the values. This greatly simplifies the interpreter
code because we no longer have to clone values before execution and
worry about whether we are referring to the correct object.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compile.go
+++ b/compile.go
@@ -11,7 +11,6 @@ import (
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
-	"github.com/influxdata/flux/libflux/go/libflux"
 	"github.com/influxdata/flux/parser"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
@@ -40,39 +39,12 @@ func Parse(flux string) (*ast.Package, error) {
 // Eval accepts a Flux script and evaluates it to produce a set of side effects (as a slice of values) and a scope.
 func Eval(ctx context.Context, flux string, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
 	h := parser.ParseToHandle([]byte(flux))
-	return evalHandle(ctx, h, opts...)
+	return defaultRuntime.evalHandle(ctx, h, opts...)
 }
 
 // EvalAST accepts a Flux AST and evaluates it to produce a set of side effects (as a slice of values) and a scope.
 func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
-	h, err := parser.ToHandle(astPkg)
-	if err != nil {
-		return nil, nil, err
-	}
-	return evalHandle(ctx, h, opts...)
-}
-
-func evalHandle(ctx context.Context, h *libflux.ASTPkg, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
-	semPkg, err := semantic.AnalyzePackage(h)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	itrp := interpreter.NewInterpreter(nil)
-	// Create a scope for execution whose parent is a copy of the prelude and whose current scope is the package.
-	// A copy of the prelude must be used since options can be mutated.
-	scope := defaultRuntime.prelude.Copy().Nest(nil)
-
-	for _, opt := range opts {
-		opt(scope)
-	}
-
-	sideEffects, err := itrp.Eval(ctx, semPkg, scope, StdLib())
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return sideEffects, scope, nil
+	return defaultRuntime.Eval(ctx, astPkg, opts...)
 }
 
 // EvalOptions is like EvalAST, but only evaluates options.
@@ -114,7 +86,12 @@ type ScopeMutator = func(values.Scope)
 // SetOption returns a func that adds a var binding to a scope.
 func SetOption(pkg, name string, v values.Value) ScopeMutator {
 	return func(scope values.Scope) {
-		scope.SetOption(pkg, name, v)
+		p, ok := scope.Lookup(pkg)
+		if ok {
+			if p, ok := p.(values.Package); ok {
+				values.SetOption(p, name, v)
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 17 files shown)


compile.go
```diff
@@ -40,16 +40,12 @@ func Parse(flux string) (*ast.Package, error) {
 // Eval accepts a Flux script and evaluates it to produce a set of side effects (as a slice of values) and a scope.
 func Eval(ctx context.Context, flux string, opts ...ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
 	h := parser.ParseToHandle([]byte(flux))
-	return evalHandle(ctx, h, opts...)
+	return defaultRuntime.evalHandle(ctx, h)
```
jlapacik: `opts` appears to be unused?

compile.go
```diff
@@ -704,30 +700,74 @@ func ToQueryTime(value values.Value) (Time, error) {
...
+		return nil, errors.Newf(codes.Invalid, "invalid import path %s", path)
+	}
+
+	// Construct the prelude scope from the prelude paths.
+	// If we are importing part of the prelude, we do not
+	// include it as part of the prelude and will stop
+	// including values as soon as we hit the prelude.
+	// This allows us to import all previous paths when loading
+	// the prelude, but avoid a circular import.
+	scope, err := imp.r.newScopeFor(path, imp)
```
jlapacik: Does this mean we are re-evaluating the prelude with every import statement?

jsternberg: The prelude only gets evaluated once and then cached for that execution. So it would be evaluated once per script, but not once per import.

compile.go
jsternberg: I need to figure out what opts is used for and use it. I am unclear on exactly how it is used so I"ll circle back on this.

interpreter/interpreter.go
```diff
@@ -624,14 +624,14 @@ func (itrp *Interpreter) doCall(ctx context.Context, call *semantic.CallExpressi
...
-		af.itrp = itrp
-		f = af
-	}
+	// if af, ok := f.(function); ok {
+	// 	af, err = itrp.mutateFunctionScope(af)
+	// 	if err != nil {
+	// 		return nil, err
+	// 	}
+	// 	af.itrp = itrp
+	// 	f = af
```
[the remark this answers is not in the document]
jsternberg: There are no objects that exist that need to be rebound. The objects that were created are newly created for every invocation. So instead of attempting to copy the package to avoid changing the cached interpreter packages, the interpreter package is unique to that run.

jsternberg:
I am not sure if it is possible to postpone this. I did this mostly because the old code was very complex and I couldn't figure out if my changes were correct or not. I couldn't validate that the old behavior was correct either when I was reading the code.

compile.go
jsternberg: Fixed. It turns out it was easier than I thought it would be.

jlapacik:
This change still seems rather complex to me. Are there any ways to simplify it further? For instance, could we construct the dependency graph of `semantic.Package` nodes during `init()`? That way we don't need to perform any cyclic dependency checks when evaluating the imports.

Also given the dependency graph, we could evaluate imports(i.e dependencies) in a bottom-up fashion for each execution. Thoughts?

jsternberg:
We could run through individual parts and see if we can simplify it further. But I will note that the old code didn't do cyclical import detection at all so this code adds that capability.

We could potentially use the dependency graph to produce imports. I thought it was easier since import returns a package object for me to just do the evaluation when the import happened.

jlapacik:
I don't want to hold this PR up, but I do think the import logic could be simplified by using a dependency graph and a topological sort of a program's dependencies.

jsternberg:
I just pushed an update that fixes setting options. I think it should be more simple now and we can probably remove `SetOption` from things.

It might be hard to get the import logic working through a topological sort because the current import sorting has a cyclical dependency. Every package is assumed to have `import "universe"` but the universe package depends on the system package. So when you evaluate system, you need to know that it shouldn't import universe.

It would be possible to do a topological sort, but I think it would be more complicated personally and the logic would be pretty similar to the import logic in this PR.

jlapacik:
It looks like your change to the `Importer` interface isn't compiling. Once it is, I will approve.

interpreter/interpreter.go
```diff
@@ -169,88 +168,42 @@ func (itrp *Interpreter) doOptionStatement(ctx context.Context, s *semantic.Opti
...
-		if p, ok := v.(*Package); ok {
-			scope.Set(k, p.Copy())
+		pkg, ok := obj.(values.Package)
+		if !ok {
+			return nil, errors.Newf(codes.Invalid, "%s: cannot set option %q on non-package value", a.Location(), a.Member.Property)
 		}
-	})
-	copyPackages(scope.Pop())
+		pkg.SetOption(a.Member.Property, init)
+		v, _ := pkg.Get(a.Member.Property)
```
nathanielc: Since its not immediatley clear can you add a comment that the reason we do a Get right after a Set is so that we return the `values.Option` instead of just the `init` value?

compile.go
```diff
@@ -704,90 +641,86 @@ func ToQueryTime(value values.Value) (Time, error) {
...
 
 func (imp *importer) Copy() *importer {
-	packages := make(map[string]*interpreter.Package, len(imp.pkgs))
-	for k, v := range imp.pkgs {
-		packages[k] = v.Copy()
-	}
-	return &importer{
-		pkgs: packages,
-	}
+	return &importer{r: imp.r}
```
nathanielc: Why do we no longer copy packages for an importer? Should we just remove the `Copy` method entirely from the `importer`?

interpreter/interpreter.go
jsternberg: I changed this code so now there's a `SetOption` utility that returns the option. It's more obvious what it is doing so this awkward code doesn't have to exist.

compile.go
jsternberg: We don't. It seems the function is now unused so I removed it. The original rationale is that we want to copy the runtime, but we want to reevaluate the packages when a copy happens. But the method is unused so there's no need to explain why it doesn't copy packages.

jsternberg:
`go test ./interpreter` now passes. I also fixed a problem with array expressions where the type wasn't set when it was part of a resolved function.

PR #2444, opened 2020-01-31
refactor(libflux): use runtime.KeepAlive to prevent finalizer race conditions

jsternberg:
The go docs for `SetFinalizer` indicate that an object can be garbage
collected even if it is currently within scope as long as the object
itself isn't used anymore. If the garbage collector runs in this
circumstance, it is possible for it to run the finalizer while a cgo
call is happening.

The `runtime.KeepAlive` function forces the compiler to have a reference
to the object until the method is called which forces the object to
remain alive. If this is placed after the call into the C library, it
prevents the handle from being freed while it is in use.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/go/libflux/analyze.go
+++ b/libflux/go/libflux/analyze.go
@@ -30,6 +30,8 @@ func (p *SemanticPkg) MarshalFB() ([]byte, error) {
 		str := C.GoString(cstr)
 		return nil, errors.Newf(codes.Internal, "could not marshal semantic graph to FlatBuffer: %v", str)
 	}
+	// See MarshalFB in ASTPkg for why this is needed.
+	runtime.KeepAlive(p)
 	defer C.flux_free(buf.data)
 
 	data := C.GoBytes(buf.data, C.int(buf.len))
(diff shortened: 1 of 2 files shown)


adriandt:
I'm understanding this better now ... the `SetFinalizer` docs have a better description of what KeepAlive is and why it is needed.

If you're looking at the `KeepAlive` docs you need to know that the containing `File` structure (referenced by `p`) is liable to be garbage collected before the `Read` call is entered because `p` goes out of scope as early as possible, which is after the first argument is evaluated and before the `Read` call is made.

I would expect the compiler to hold onto the `p` reference at least until the statement where it is last used completes (until after the `Read` call), but this is not the case.

adriandt:
In an effort to better understand go and these issues I coded up a small example that demonstrates the problem described in the KeepAlive documentation

https://golang.org/pkg/runtime/#KeepAlive

https://github.com/adriandt/sandbox/blob/master/keepalive/main.go

Issue #2393, opened 2020-01-15
Remove equality type constraint on comparison operators


jlapacik:
Until polymorphic literals are added to the type system, comparison operators such as `<`, `>`, etc must operate on values of the same type. This will most likely cause many queries to break as comparing float columns with integer literals is common.

We should remove the equality constraint for comparison operators, adding it back in when we implement polymorphic literals.

See #2374

jsternberg:
Fixed by #2449.

PR #2449, opened 2020-02-03
refactor: various stdlib fixes for algorithm w

jsternberg:
This contains various fixes to the stdlib to make the tests pass with
algorithm w. It updates some of the builtins with the correct type
signature. Some of them did not match the Go code.

It updates `map` to work with compiled functions.

It also updates an end to end test which relied on polymorphic literals.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -389,7 +389,7 @@ func (e *unaryEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, e
 	if err != nil {
 		return nil, err
 	}
-	values.CheckKind(ret.Type().Nature(), e.t.Nature())
+	// values.CheckKind(ret.Type().Nature(), e.t.Nature())
 	return ret, nil
 }
 
--- a/libflux/src/flux/semantic/nodes.rs
+++ b/libflux/src/flux/semantic/nodes.rs
@@ -976,36 +976,50 @@ impl BinaryExpr {
                 Constraint::Kind(self.typ.clone(), Kind::Divisible),
             ]),
             ast::Operator::GreaterThanOperator => Constraints::from(vec![
-                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                // https://github.com/influxdata/flux/issues/2393
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
                 Constraint::Equal(self.typ.clone(), MonoType::Bool),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Comparable),
+                Constraint::Kind(self.right.type_of().clone(), Kind::Comparable),
             ]),
             ast::Operator::LessThanOperator => Constraints::from(vec![
-                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                // https://github.com/influxdata/flux/issues/2393
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
                 Constraint::Equal(self.typ.clone(), MonoType::Bool),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Comparable),
+                Constraint::Kind(self.right.type_of().clone(), Kind::Comparable),
             ]),
             ast::Operator::EqualOperator => Constraints::from(vec![
-                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                // https://github.com/influxdata/flux/issues/2393
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
                 Constraint::Equal(self.typ.clone(), MonoType::Bool),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Equatable),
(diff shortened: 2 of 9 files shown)


semantic/graph.go
```diff
@@ -499,7 +499,7 @@ type ArrayExpression struct {
 
 	Elements []Expression
 
-	typ MonoType
+	Type MonoType
```
nathanielc: Why the change to make it public? The expression interface has the `TypeOf` method?

stdlib/universe/filter_mixed_empty_test.flux
```diff
@@ -59,7 +59,7 @@ t_filter_mixed_empty = (table=<-) =>
   table
   |> range(start: 2018-05-22T19:53:26Z)
   |> filter(fn: (r) => r._measurement == "diskio")
-  |> filter(fn: (r) => r["_value"] > 1000, onEmpty: "keep")
+  |> filter(fn: (r) => r["_value"] > 1000.0, onEmpty: "keep")
```
nathanielc: This should have worked correct? I though we removed the constraints about equal types for now?

semantic/graph.go
jsternberg: I think `resolveValue` needs to be refactored, but it's in a different package and so I don't have access to modify it. I am not sure `resolveValue` is needed anymore and, in my opinion, it is a bit of a code smell that it converts a value into a semantic node.

stdlib/universe/filter_mixed_empty_test.flux
jsternberg: Oh if we did then I need to rebase. I'll look at this. If it's not done, I can remove that constraint.

compiler/runtime.go
```diff
@@ -389,7 +389,7 @@ func (e *unaryEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, e
 	if err != nil {
 		return nil, err
 	}
-	values.CheckKind(ret.Type().Nature(), e.t.Nature())
+	// values.CheckKind(ret.Type().Nature(), e.t.Nature())
```
ayang64: i don't think you meant to commit commented out code.

jsternberg: We're mostly removing these checks because they are more often wrong than correct. We may re-add them so I've been commenting them out whenever they cause a problem.

stdlib/universe/filter_mixed_empty_test.flux
jsternberg: Updated.

libflux/src/flux/semantic/nodes.rs
```diff
@@ -976,36 +976,44 @@ impl BinaryExpr {
...
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
                 Constraint::Equal(self.typ.clone(), MonoType::Bool),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Equatable),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Comparable),
+                Constraint::Kind(self.right.type_of().clone(), Kind::Equatable),
+                Constraint::Kind(self.right.type_of().clone(), Kind::Comparable),
             ]),
             ast::Operator::LessThanEqualOperator => Constraints::from(vec![
-                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
```
nathanielc: Can you add a link to this issue as part of these comments? https://github.com/influxdata/flux/issues/2393 that way we have the context for why they are commented out for now.

semantic/graph.go
```diff
@@ -934,7 +934,7 @@ func (l *BooleanLiteral) Copy() Node {
 	return nl
 }
 func (e *BooleanLiteral) TypeOf() MonoType {
-	return e.typ
+	return BasicBool
```
jlapacik: I would prefer if `TypeOf` returned the inferred type. I'd like to know if we ever infer anything other than `bool`.

jsternberg: In the future we can remove this. The problem is resolving functions currently replaces values with semantic nodes and so there's no type inference done. We can hopefully remove this in the future because I think you're right.

libflux/src/flux/semantic/nodes.rs
```diff
@@ -976,36 +976,50 @@ impl BinaryExpr {
                 Constraint::Kind(self.typ.clone(), Kind::Divisible),
             ]),
             ast::Operator::GreaterThanOperator => Constraints::from(vec![
-                Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
+                // https://github.com/influxdata/flux/issues/2393
+                // Constraint::Equal(self.left.type_of().clone(), self.right.type_of().clone()),
                 Constraint::Equal(self.typ.clone(), MonoType::Bool),
                 Constraint::Kind(self.left.type_of().clone(), Kind::Comparable),
+                Constraint::Kind(self.right.type_of().clone(), Kind::Comparable),
```
jlapacik: This change probably should have been pointed directly at master since it's not `use-algow` specific

PR #2459, opened 2020-02-05
refactor(stdlib): fix reduce implementation for algorithm w

jsternberg:
This fixes reduce so it works with the refactored row functions and
algorithm w.

This also changes union to expect an array of arrays instead of an array
of objects. This is because table streams are now represented in the
type system with arrays instead of table objects.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/row_fn.go
+++ b/execute/row_fn.go
@@ -197,7 +197,7 @@ func (f *rowFn) eval(ctx context.Context, row int, cr flux.ColReader, extraParam
 		f.arg0.Set(col.Label, ValueForRow(cr, row, j))
 	}
 	for k, v := range extraParams {
-		f.arg0.Set(k, v)
+		f.args.Set(k, v)
 	}
 	return f.preparedFn.Eval(ctx, f.args)
 }
@@ -296,8 +296,6 @@ func (f *RowMapFn) Eval(ctx context.Context, row int, cr flux.ColReader) (values
 
 type RowReduceFn struct {
 	rowFn
-	isWrap bool
-	// wrapObj *Record
 }
 
 func NewRowReduceFn(fn *semantic.FunctionExpression, scope compiler.Scope) (*RowReduceFn, error) {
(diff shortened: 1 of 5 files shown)


values/json.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/semantic"
+)
+
+// JsonValue is a Flux value that can be encoded and decoded to/from JSON.
+type JsonValue struct {
+	Value
+}
```
jlapacik: Why has this been added?

execute/row_fn.go
```diff
@@ -197,7 +197,7 @@ func (f *rowFn) eval(ctx context.Context, row int, cr flux.ColReader, extraParam
 		f.arg0.Set(col.Label, ValueForRow(cr, row, j))
 	}
 	for k, v := range extraParams {
-		f.arg0.Set(k, v)
+		f.args.Set(k, v)
```
jlapacik: Can you remind me what this does again? Why the switch from arg to arg0?

values/json.go
jsternberg: This was for serializing the spec. I think it can be removed. I went into this with the intention to preserve the repl.

execute/row_fn.go
jsternberg: The previous code was wrong. It's too set the values on the arguments. I misunderstood it's purpose previously when fixing it and it's only used by reduce.

stdlib/universe/reduce.go
```diff
@@ -105,10 +92,6 @@ func (s *ReduceProcedureSpec) Copy() plan.ProcedureSpec {
 	ns := new(ReduceProcedureSpec)
 	*ns = *s
 	ns.Fn = s.Fn.Copy()
-	ns.ReducerType = s.ReducerType
-	for k, v := range s.Identity {
-		ns.Identity[k] = v
-	}
```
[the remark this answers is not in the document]
jsternberg: I'm not sure the original copy was needed. The map doesn't appear to change so it should be safe to share them.

stdlib/universe/reduce.go
```diff
@@ -131,130 +114,120 @@ type reduceTransformation struct {
 	cache          execute.TableBuilderCache
 	ctx            context.Context
 	fn             *execute.RowReduceFn
-	neutralElement map[string]values.Value
+	neutralElement values.Object
```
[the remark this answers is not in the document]
jsternberg: Done.

semantic/types.go
```diff
@@ -39,9 +43,44 @@ var natureNames = []string{
...
 	if int(n) < len(natureNames) {
 		return natureNames[n]
 	}
 	return "nature" + strconv.Itoa(int(n))
 }
+
+func (n Nature) MarshalJSON() ([]byte, error) {
+	s := n.String()
+	return json.Marshal(s)
+}
```
jlapacik: These changes are no longer necessary, correct?

jsternberg: Yes that's correct.

PR #2472, opened 2020-02-06
fix(libstd): update type inference test case - add one for union

jlapacik:
The `==` operator no longer requires both operands to be of the same type.
The commit updates the test in libstd accordingly and also adds one for
union.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/libstd/lib.rs
+++ b/libflux/src/libstd/lib.rs
@@ -159,7 +159,31 @@ mod tests {
 
         // TODO(algow): re-introduce equality constraints for binary comparison operators
         // https://github.com/influxdata/flux/issues/2466
-        let want = semantic::parser::parse(
+        let want_a = semantic::parser::parse(
+            r#"forall [t0, t1, t3] where t1: Equatable [{
+                _value: t0
+                    | A: t1
+                    | _time: time
+                    | _measurement: string
+                    | _field: string
+                    | t3
+                    }]
+            "#,
+        )
+        .unwrap();
+        let want_b = semantic::parser::parse(
+            r#"forall [t0, t1, t3] where t1: Equatable [{
+                _value: t0
+                    | B: t1
+                    | _time: time
+                    | _measurement: string
+                    | _field: string
+                    | t3
+                    }]
+            "#,
+        )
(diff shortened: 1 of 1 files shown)


libflux/src/libstd/lib.rs
```diff
@@ -136,6 +138,44 @@ mod tests {
...
+        let mut file = convert_file(ast, &mut f).unwrap();
+        let (got, _) = infer_file(&mut file, prelude, &mut f, &imports, &None).unwrap();
+
+        // TODO(algow): re-introduce equality constraints for binary comparison operators
+        // https://github.com/influxdata/flux/issues/2466
+        let want = semantic::parser::parse(
+            r#"forall [t0, t1, t2, t3] where t1: Equatable, t2: Equatable [{
+                _value: t0
+                    | A: t1
+                    | B: t2
```
nathanielc: This part doesn't look right. The data coming out of union is not guaranteed to the both the `A` and `B` fields. It could have them both but it could also be missing for some of the data.

 What I think has happened here is that the sources `a` and `b`  have been inferred to have both `A` and `B` fields. Can you assert the type of `a` and `b` to validate my assumption? 

It seems like we want `c` to be of type  `{_value:t0, _time: time, _measurement: string, _field: string, | t3}` and not say anything about `A` or `B` since its not guaranteed they exist.

jlapacik: You are correct in that the sources `a` and `b` have been inferred to have both `A` and `B` fields (I will assert just to confirm). The output looks right to me though. If one table has an `A` column and the other has a `B` column, the output will have both `A` and `B` columns with some entries possibly being null.

jlapacik: I asserted the types of `a` and `b`. Note that the identifiers `a` and `b` have not been inferred to have both `A` and `B` fields. Only `c` has been inferred to have both fields.

nathanielc: Great, yeah I was worried it would trickle back up to the type of `a` and `b`. But looks like we are good. Thanks!

PR #2478, opened 2020-02-06
fix(semantic): update type of window function

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

The window function contains two type variables but only one of them was universally quantified. Now both are.

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -773,7 +773,7 @@ pub fn builtins() -> Builtins<'static> {
                 // already exist.
                 // https://github.com/influxdata/flux/issues/2255
                 "window" => r#"
-                    forall [t0] where t0: Row, t1: Row (
+                    forall [t0, t1] where t0: Row, t1: Row (
                         <-tables: [t0],
                         ?every: duration,
                         ?period: duration,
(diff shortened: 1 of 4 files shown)


libflux/src/flux/semantic/builtins.rs
```diff
@@ -773,7 +773,7 @@ pub fn builtins() -> Builtins<'static> {
                 // already exist.
                 // https://github.com/influxdata/flux/issues/2255
                 "window" => r#"
-                    forall [t0] where t0: Row, t1: Row (
+                    forall [t0, t1] where t0: Row, t1: Row (
```
nathanielc: Interesting, so because t1 wasn't listed as a free var it likely conflicted with the tvar chosen for the range start param adding an extra Row constraint where it didn't belong.

Should we add a validate step after parsing these type expressions so that we can catch these systematically?

jlapacik: Yes that's a good idea. We already check for this as part of the rust type inference tests (that all type variables are universally quantified). It should be simple enough to just port that over to the builtins.

jlapacik: The build process has been updated to check for free type variables in the prelude/stdlib. This caught the same typo in `tripleExponentialDerivative`.

Issue #2319, opened 2020-01-03
Remove compilation cache


nathanielc:
We have a compilation cache were we only recompile functions for new input types. This previously used some Go pointer trickery, that is no longer possible using flatbuffers. We can implement the compilation cache by using a hash of the record property keys and basic type. In other words implement a cache using a hash function over a subset of type instead of all possible Flux types.

See https://github.com/influxdata/flux/blob/feat/use-algo-w/compiler/compiler.go#L483 for a starting point.

nathanielc:
After discussion we have concluded that we no longer need the compilation cache. The changes to use the generic values.Value type and the change to no longer call type inference directly from the compile step has removed the need to compile row functions for each input type. A single compilation step is all that is needed. 

This issue is now about removing the CompilationCache entirely and changing call sites to use Compile directly. 

The validation of input types is still required.

jsternberg:
This is going to involve fixing the compiler too. It doesn't really make sense to remove the compilation cache without fixing the compiler at the same time.

PR #2480, opened 2020-02-06
fix: make staticcheck pass

wolffcm:
All of the code that is deleted is dead.


compile.go
```diff
@@ -139,85 +139,6 @@ func isPreludePackage(pkg string) bool {
 	return false
 }
 
-type scopeSet struct {
```
wolffcm: I think this was intended to implement `runtime.Prelude()`, but that is now done in a different way.

compiler/compiler.go
```diff
@@ -565,10 +565,10 @@ func (c *CompilationCache) Compile(in semantic.MonoType) (Func, error) {
 	panic("unimpleneted")
 }
 
-type funcErr struct {
-	F   Func
-	Err error
-}
+//type funcErr struct {
```
wolffcm: A lot of the compilation cache code around this is commented out, so I just followed suit here.

nathanielc: I recommend we just delete the code instead of commenting it out.

semantic/label_set.go
```diff
@@ -35,92 +35,3 @@ func (s LabelSet) String() string {
 func (s LabelSet) isAllLabels() bool {
```
nathanielc: Is label_set used anywhere now? It's purpose was soley used for implementing type inference in Go. I expect it is unused at this point.

wolffcm: It's still used in `values/object.go`, but only there.  I'll move it and make it private.

wolffcm:
Fixes #2319

libflux/include/influxdata/flux.h
```diff
@@ -56,6 +61,10 @@ struct flux_error_t *flux_ast_marshal_json(struct flux_ast_pkg_t *, struct flux_
 // using flux_free if it is non-null.
 struct flux_error_t *flux_ast_marshal_fb(struct flux_ast_pkg_t *, struct flux_buffer_t *);
 
+// flux_get_env_stdlib instantiates a flatbuffers TypeEnvironment and creates a pointer
+// to it to use when performing lookups on the stdlib
+void flux_get_env_stdlib(struct flux_buffer_t *);
```
nathanielc: Why did these `flux.h` files need to change? I am also confused as to why they seem to have changed in different ways. 

Was this because the `generate` step has been getting ignored?

wolffcm: Yes, they got out of sync, because we were ignoring `checkgenerate`. The file `libflux/go/libflux/flux.h` had moved ahead of `libflux/include/...`.  

I also moved `flux_get_env_stdlib` to be earlier in the file so that the declaration of `flux_semantic_pkg_t` was just before `flux_analyze()`, since they are closely related.

nathanielc: Great makes sense. Thanks!

PR #2491, opened 2020-02-07
feat(stdlib): remove generate package from stdlib

### LINKED ISSUE #2476: `CompileFnParam` is not implemented and causes `generate` package to panic
`CompileFnParam` is currently unimplemented. This is causing uses of the "generate" package to panic since the package registration (specifically the call to `createFromGeneratorSource`) refers to this function. 

We either need to get rid of this call to `CompileFnParam` or implement the function. 

DOD: 
-  Tests are unskipped and now pass

jsternberg:
The generate package is removed from stdlib and will, at some point, be
replaced by `internal/gen`.

BREAKING CHANGE: The `generate` package has been removed from the
standard library.

Fixes #2476.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/go/libflux/parser.go
```diff
@@ -4,6 +4,7 @@ package libflux
 // #cgo LDFLAGS: -L. -lflux
 // #include "flux.h"
 // #include <stdlib.h>
+// const int VERSION = 1;
```
wolffcm: What is this for?

jsternberg: It was for forcing a recompile and I need to remove it.

jsternberg:
Closing in favor of #2495. I want to remove it, but @russorat says he uses it some time so I want to create an issue to design out the generate package and make sure we've got different generate functions that fit different use cases before we just wholesale rip this one out.

Issue #2410, opened 2020-01-17
Interpreter seems to ignore option assignments


wolffcm:
This bug is specific to the `feat/use-algo-w` branch.

I'm seeing the interpreter test `TestInterpreter_SetNewOption` fail.  It's a very simple test case:
```js
package alert
option state = "Warning"
state
```
The error produced is `failed to evaluate package: undefined identifier "state"`.

Seems like there is an issue with evaluating `IdentifierExpression` in the interpreter.

jsternberg:
This might be related to the package stuff. At the moment, all of the packages for the interpreter are constructed incorrectly.

PR #2513, opened 2020-02-11
fix(interpreter): test option assignment

jlapacik:
This PR updates a test showing that one can import and mutate an option from an external package.

Fixes https://github.com/influxdata/flux/issues/2410.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/interpreter/package_test.go
+++ b/interpreter/package_test.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/influxdata/flux"
+	_ "github.com/influxdata/flux/builtin"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/dependencies/dependenciestest"
 	"github.com/influxdata/flux/internal/errors"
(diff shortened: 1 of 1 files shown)


jsternberg:
@jlapacik package options are mutable so setting an option should set the value in the original package object instead of adding it to the current scope.

interpreter/package_test.go
```diff
@@ -365,7 +365,6 @@ func TestInterpreter_EvalPackage(t *testing.T) {
 */
 
 func TestInterpreter_SetNewOption(t *testing.T) {
```
jsternberg: Is this test testing anything anymore? It seems to be testing whether we can add an option to an existing package. This isn't possible to do so I'm not sure what this is testing.

jlapacik: What's the correct behavior? I thought that the interpreter did not mutate packages anymore. Is this the case? If so, then either this test should be checking to see if the option has been set in the scope, or we should just remove the test as we already have tests in place for option assignments.

jsternberg: The correct behavior is that setting an option will modify the value in the original package. So that way this would work as this is my understanding of options.

```
package influxdb

option token = ""

builtin _from
from = (bucket, token=token) => _from(bucket: bucket, token: token)

package main

import "influxdata/influxdb"
option influxdb.token = "abc"

influxdb.from(bucket: "telegraf") |> ...
```

The option would modify the value in the `influxdata/influxdb` scope as if it were a pointer to that value. What isn't possible is doing `option influxdb.token2 = "blahblah"` because the option wasn't defined in the original scope. This is allowed even if it is a no-op just because the previous behavior allowed it.

jlapacik: @jsternberg I have updated the test accordingly.

Issue #2511, opened 2020-02-11
Go build should rebuild when the libflux archives change


nathanielc:
Currently the Go build system only rebuilds the `libflux/go/libflux` package if a Go source or C related source file changes. This causes confusion as changes to the Rust code are not immediately available when called from Go as the Go code is still linked against the older Rust code. Its also not clear how to force the Go build to perform a complete rebuild when this happens.

We need a way to automatically inform the Go build process when the Rust code has changed. This PR has a good first idea https://github.com/influxdata/flux/pull/2423 It uses `make` to generate a unique C source file from the output of the Rust build. This has the effect (even though C files are never used) to trigger the Go build to rebuild that package and link against the new libflux.a and liblibstd.a archives.

As proposed in https://github.com/influxdata/flux/pull/2423#issuecomment-578288822 it would be nice if the build tooling worked without requiring a `Makefile` wrapper. A quick search suggests that Rust does not have any post-build hooks https://github.com/rust-lang/cargo/issues/545 and as such it would be difficult to generate the unique C files as part of the `cargo build` process.

In light of the above my recommendation is to use the simple `make` based solution as a starting point.

DOD

* There exist a command that can build the project where changes to the Rust code result in the Go build correctly linking against the new archives

adriandt:
I quickly experimented with using `pkg-config` to generate a unique `-DVAR="value"` define on each run, but:

1. That would be an awful hack.
2. `cgo` does not call `pkg-config` on every invocation of `go build` so it's a no-go. (pun intended).

jsternberg:
@adriandt it should be possible to do this using a generated variable inside of the `libflux/go/libflux` package and it might be possible to leverage the work that @rockstar and I have been working on.

https://github.com/influxdata/pkg-config

We've been working on a utility that you can install and is a drop-in replacement for `pkg-config` that Go can execute. If you put it on your path, Go will use it as if it were the system `pkg-config` or you can use it by setting `PKG_CONFIG=/path/to/the/above/binary` and then execute Go.

I was thinking that `pkg-config` could be extended to support the `--variable` option. We can then generate the `flux.pc` file to include either the source or binary hashes. The `go generate` command can then call `pkg-config --variable sha1sum` or something and insert that into the Go code somewhere so it is unique.

The above `pkg-config` will rebuild the Rust library so `go generate ./libflux` would be the equivalent of calling `cargo build` as if you had called the makefile.

```
const char *flux_sources_hash = "hash value";
```

[...]

PR #2516, opened 2020-02-12
fix: create unique C files to force CGO to consider code changed

adrian-thurston:
Updating .a files in a CGO package does not trigger rebuilding of packages that
depend on it. Unique C files do work though. Generate them using shasum, which
comes from the perl package. refs #2511


adrian-thurston:
Obsolted by `go generate` command.

Issue #2492, opened 2020-02-07
Error when attempting to unmarshal semantic.Node in Filter test


fchikwekwe:
When running `TestFilterOperation_Marshaling`, trying to unmarshal json representation of a flux.Operation causes the following error message: 

`failed to unmarshal operation "filter": json: cannot unmarshal object into Go struct field FunctionBlock.fn.fn.Block.Body of type semantic.Node`

fchikwekwe:
DOD: 
-  Tests are unskipped and now pass

fchikwekwe:
UPDATE: This also affects `TestMapOperation_Marshaling`

wolffcm:
Ah, hmm, we decided to discontinue support for being able to marshal the semantic graph as JSON, since we have flatbuffers now.  But we still serialize the Spec for the REPL I believe, and the Spec for map and filter contain semantic graphs for their lambdas.

This is something else that would need to change to get the  REPL working again I think.  Probably okay to treat this as lower priority.

PR #2518, opened 2020-02-12
fix(universe): remove tests for marshalling semantic graph

jlapacik:
Fixes https://github.com/influxdata/flux/issues/2492.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jlapacik:
I thought so too, but it looks like the rest of the skipped tests are all related to the type environment.

PR #2523, opened 2020-02-13
fix(stdlib): remove unused notification rule fields from slack api

fchikwekwe:
The Slack notification rule apis have some fields that no longer used by InfluxDB. There are unification errors caused by the presence of these fields in the Flux api. We are removing these fields to see if that fixes type checking. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/slack/slack.flux
+++ b/stdlib/slack/slack.flux
@@ -10,13 +10,10 @@ option defaultURL = "https://slack.com/api/chat.postMessage"
 // `message` sends a single message to a Slack channel. It will work either with the chat.postMessage API or with a slack webhook.
 // `url` - string - URL of the slack endpoint. Defaults to: "https://slack.com/api/chat.postMessage", if one uses the webhook api this must be acquired as part of the slack API setup. This URL will be secret. Don't worry about secrets for the initial implementation.
 // `token` - string - the api token string.  Defaults to: "", and can be ignored if one uses the webhook api URL.
-// `username` - string - Username posting the message.
 // `channel` - string - Name of channel in which to post the message. No default.
-// `workspace` - string - Name of the slack workspace to use if there are multiple. Defaults to empty string.
 // `text` - string - The text to display.
-// `iconEmoji` - string - Name of : enclose emoji to use as image of user when posting message, will not show as the avatar icon with a slack webhook.
 // `color` - string - Color to give message: one of good, warning, and danger, or any hex rgb color value ex. #439FE0.
-message = (url=defaultURL, token="", username, channel, workspace, text, iconEmoji, color) => {
+message = (url=defaultURL, token="", channel, text, color) => {
     attachments = [{
         color: validateColorString(color),
         text: string(v: text),
@@ -24,12 +21,9 @@ message = (url=defaultURL, token="", username, channel, workspace, text, iconEmo
     }]
 
     data = {
-        username: username,
         channel: channel,
-        workspace: workspace,
         attachments: attachments,
         as_user: false,
-        icon_emoji: iconEmoji,
     }
 
     headers = {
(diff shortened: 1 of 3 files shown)


stdlib/slack/slack.flux
```diff
@@ -10,26 +10,20 @@ option defaultURL = "https://slack.com/api/chat.postMessage"
...
 // `url` - string - URL of the slack endpoint. Defaults to: "https://slack.com/api/chat.postMessage", if one uses the webhook api this must be acquired as part of the slack API setup. This URL will be secret. Don't worry about secrets for the initial implementation.
 // `token` - string - the api token string.  Defaults to: "", and can be ignored if one uses the webhook api URL.
-// `username` - string - Username posting the message.
 // `channel` - string - Name of channel in which to post the message. No default.
-// `workspace` - string - Name of the slack workspace to use if there are multiple. Defaults to empty string.
 // `text` - string - The text to display.
-// `iconEmoji` - string - Name of : enclose emoji to use as image of user when posting message, will not show as the avatar icon with a slack webhook.
 // `color` - string - Color to give message: one of good, warning, and danger, or any hex rgb color value ex. #439FE0.
-message = (url=defaultURL, token="", username, channel, workspace, text, iconEmoji, color) => {
+message = (url=defaultURL, token="", channel, text, color) => {
```
fchikwekwe: I decided to just change Slack because the pagerduty attribute that is causing errors seems to be nested within another attribute that is used in InfluxDB. We can check if these changes to slack's api work and then I'll spend time trying to figure out how to fix pagerduty.

fchikwekwe:
> Can you explain why this needs to change? How do we know these are not used? Or in other words this is a breaking change, how do we know its safe to make this change?

So, we checked with the Monitoring team before I made this change. These properties are not being used because they are all optional from Slack's point of view. 

Since the username property is no longer present in InfluxDB, we could either add it over there or remove it from the flux APIs. We figured this change would be easier to make.

jlapacik:
InfluxDB generates flux code from the `slack` package in order to create notification rules, but currently the generated code does not type check. https://github.com/influxdata/flux/issues/2341#issuecomment-585274167 is an example of that generated code. There are two options:

1. Leave the slack package as it is, and update the code generation in InfluxDB.
2. Leave the code generation as it is, and update the slack package in Flux.

I think I prefer 2. which is what @fchikwekwe has done. Most likely InfluxDB is the only consumer of this package, and so I think it's best to simplify the flux API instead of complicating the code that generates notification rules.

nathanielc:
@fchikwekwe @jlapacik Thanks for the explanation that helps a lot. I don't want to break the `slack` API. It was designed to be consumed by more than just the front end UI and I don't want to restrict it to just that use case.

That said we can't rely on the generated code being updated because it has already been generated and I don't think its possible to force it to regenerate. That leaves us with change the Slack API to allow those fields to be truly optional. Would using the `exists` operator enable that behavior?

Something like this as an update to the `endpoint` implementation?

```
endpoint = (url=defaultURL, token="") =>
    (mapFn) =>
        (tables=<-) => tables
            |> map(fn: (r) => {
                obj = mapFn(r: r)
                // repeat this pattern for all optional properties
                username = if exists obj.username then obj.username else "" // should this be `nil` instead of empty string?
                return {r with _sent: string(v: 2 == message(
                    url: url,
                    token: token,
                    username: username,
                    channel: obj.channel,
                    workspace: obj.workspace,
                    text: obj.text,
                    iconEmoji: obj.iconEmoji,
                    color: obj.color,
                ) / 100)}
            })
```

To make sure I understand, this used to work because `obj.username` would return `nil` and then when we go to encode the data as json the `nil` field was just skipped. Is that correct?

jlapacik:
@nathanielc your reason for why this worked before algo-w is correct. Unfortunately `exists` will not help us in this situation. `exists obj.username` will still assert that `obj` must have a label called `username`. The issue is that `mapFn` returns a static object which according to the slack package, needs labels `username`, `workspace`, and `iconEmoji`.

jlapacik:
After some discussion with @nathanielc I believe Faith's fix is the correct one. However we will need to create an issue to revert the API once algo-w has been merged. Also this is contingent on InfluxDB being the only consumers of the API. I will confirm that this is indeed the case.

jlapacik:
I have been able to confirm that all requests in the query log that use the flux slack api are generated from notification rules. Therefore it looks like it is safe to temporarily modify the api to be consistent with how those queries are generated.

PR #2529, opened 2020-02-18
refactor: repl can maintain state from multiple invocations

### LINKED ISSUE #2402: Add type environment parameter to Rust type inference
We should add the ability for clients to pass in an optional type environment to Rust type inference.  We already have the ability to serialize type environments to a flatbuffer, so we should make use of that.

I believe that we'll need this working so we can tell Rust inference about  any "extern" declarations (like `v = {bucket: "telegraf"}`, etc) coming from the web UI.

Also, this is needed so that the REPL will work right, so that each new statement that the user submits can include the type environment of previous statements.  This also allows more thorough testing of the interpreter.

This will require updating the Rust API for type inference to have an additional type environment parameter:
https://github.com/influxdata/flux/blob/master/libflux/src/libstd/lib.rs#L28-L64

jsternberg:
The repl can now maintain the analyzer state for a file so it can
analyze parts of a file rather than only a whole file. This allows the
repl to recognize variables that have been defined in one line and used
in another.

Fixes #2402.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
An interesting thing that I found while working on this is that I thought this technique would be helpful for evaluating imports, but I did not find that to be the case. Imports and evaluating packages have different semantics than what the repl does. I found this code was only helpful for the repl.

In the future, I'd like to refactor this. I think this gets us to a minimum viable product though and I don't want to spend too much time refactoring a lot of code that I plan to refactor again.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -786,10 +788,10 @@ pub fn builtins() -> Builtins<'static> {
                     ) -> [t1]
                 "#,
                 "yield" => r#"
-                    forall [t0] where t0: Row (
-                        <-tables: [t0],
+                    forall [t0] (
+                        <-tables: t0,
                         ?name: string
-                    ) -> [t0]
+                    ) -> t0
```
jlapacik: Why did this signature change?

stdlib/testing/testing.flux
```diff
@@ -5,6 +5,7 @@ import c "csv"
 builtin assertEquals
 builtin assertEmpty
 builtin diff
+builtin fail
```
jlapacik: Is this addition absolutely necessary? What's the difference between this and:
```
fail = () => [][0]
```

libflux/src/flux/semantic/builtins.rs
```diff
@@ -347,6 +347,8 @@ pub fn builtins() -> Builtins<'static> {
                 "assertEquals" => "forall [t0] (name: string, <-got: [t0], want: [t0]) -> [t0]",
                 "assertEmpty" => "forall [t0] (<-tables: [t0]) -> [t0]",
                 "diff" => "forall [t0] (<-got: [t0], want: [t0], ?verbose: bool) -> [{_diff: string | t0}]",
+                "fail" => "forall [] () -> bool",
+                "yieldValue" => "forall [t0] (<-v: t0) -> t0",
```
jlapacik: I don't see where this function is defined?

stdlib/testing/fail.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+func init() {
+	flux.RegisterPackageValue("testing", "fail", values.NewFunction(
```
jlapacik: I would prefer not adding this function to the stdlib.

stdlib/universe/table_fns_test.go
```diff
@@ -143,23 +73,87 @@ func TestTableFind_Call(t *testing.T) {
...
+					},
+					Data: [][]interface{}{
+						{mustParseTime(`2018-05-22T19:53:26.000000000Z`), 0.0, "CPU", "user1"},
+						{mustParseTime(`2018-05-22T19:53:36.000000000Z`), 1.0, "CPU", "user1"},
+					},
+				}
+				want.Normalize()
+				return want
+			}(),
+			fn: `f = (key) => key.user == "user1"`,
```
jlapacik: It seems like you're just "rearranging" the code in this file. Is that all that's going on?

stdlib/universe/yield.go
```diff
@@ -17,7 +20,21 @@ type YieldOpSpec struct {
...
+
+		if _, ok := in.(*flux.TableObject); ok {
+			fn, err := flux.FunctionValueWithSideEffect("yield", createYieldOpSpec, yieldSignature)
+			if err != nil {
+				return nil, err
+			}
+			return fn.Function().Call(ctx, args)
+		}
+		return in, nil
+	}, true))
```
jlapacik: So this says that instead of working only on tables, `yield` is essentially the identity function? So now you can yield a value of any type?

jsternberg: This is correct. I can try to get a different way for this to work if we don't want this, but I figured we wanted to change `yield` at some point anyway to do this and I needed a side effect and it isn't possible to inject side effect functions anymore.

stdlib/universe/table_fns_test.go
jsternberg: The previous code executed the query once and stored the table results to be compared with an init function and this seemed like too much. I had to change the functions to execute the csv as a prelude rather than inject a table object into the script to avoid two calls to the interpreter so I figured I might as well make these tests more similar to other ones that do the same thing.

stdlib/testing/fail.go
jsternberg: Is there a reason? It seems like a reasonably useful function since we use it in a lot of packages. Would you feel more comfortable with it if I threw it into an `internal/testing` package instead?

libflux/src/flux/semantic/builtins.rs
jsternberg: My original idea was to create a special yield function for testing but this didn't work so this is a mistake.

stdlib/testing/testing.flux
jsternberg: The error message is more clear. I guess this could be used everywhere instead if you're just looking for any error.

libflux/src/flux/semantic/builtins.rs
jsternberg: See below. I think `yield()` was always supposed to be able to yield anything as a passthrough and so I started using it that way. I can try a different method, but only builtin functions can have side effects so `yield()` was perfect for doing that and being unobtrusive.

jlapacik: Sounds good to me.

jsternberg: I'm working on moving this version of yield to a separate package along with `fail`. We can decide whether to modify `yield` to yield anything later. I figure it was probably a bad idea for me to switch the signature of the function just because I needed it for a unit test.

stdlib/testing/fail.go
jlapacik: From my knowledge `fail` is used only for tests where we want the flux script to produce an error if something unexpected happens. It doesn't seem like a useful function to expose publicly. Especially since you can achieve the equivalent functionality in a single line of pure flux `fail = () => [][0]`

jsternberg: I threw the function into `internal/testutil`. I like it because you could eventually customize the error message. If you prefer the implementation to be pure flux, I can change it to that. An internal package means that we don't have to decide if it is stdlib worthy or not and I still can access it from the tests without booting up the repl.

repl/repl.go
```diff
@@ -17,30 +17,56 @@ import (
...
+	analyzer *libflux.Analyzer
+	importer interpreter.Importer
 
 	cancelMu   sync.Mutex
 	cancelFunc context.CancelFunc
 }
 
+var prelude = []string{
+	"universe",
+}
```
jlapacik: I thought there was another package that was part of the prelude?

libflux/src/libstd/lib.rs
```diff
@@ -47,6 +51,115 @@ pub unsafe extern "C" fn flux_analyze(
     }
 }
 
+struct SemanticAnalyzer {
+    f: Fresher,
+    env: Environment,
+    imports: Environment,
+    importer: Box<dyn Importer>,
```
jlapacik: What is `importer` for? The prelude?

repl/repl.go
jsternberg: I was a bit confused about that. I thought that `influxdata/influxdb` was only imported in influxdb. I'll re-add it since I'm not sure and I think it was included by default.

libflux/src/libstd/lib.rs
jsternberg: This is for the builtins if a package path that matches them exists. Usually this is invoked with main so the builtins aren't present.

jlapacik: In what context is this not invoked with main?

jsternberg: None currently.

jlapacik: Since that's the case for right now, can it be removed from the API?

jsternberg: I don't see the reason why to remove it. If we end up using it outside of the context of main, I'd rather it be correct and I don't believe it currently causes any harm for it to be present and return nothing.

jlapacik: I would just prefer not to have unused code.

jsternberg: The code is used and is needed to make the functionality work as expected. At the moment, it is not used in that context. I would prefer not to debug this in the future unless some specific reason why adding it makes it wrong.

It seems to me like you are asking for the API that specifies the package name to the analyzer to be removed. If I do that, then fixing this, if it is used that way in the future, would require making a breaking change to the API.

PR #2533, opened 2020-02-18
feat: add experimental query pkg

jsteenb2:
the intent of this pkg is to enable usability testing and get feedback on ways to make querying easier for the 80%ers, who are mostly non dev/data analyst folks

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/experimental/query/from.flux
```diff
@@ -0,0 +1,46 @@
...
+
+_sumWindowDet = (table=<-, window=0s, fn=mean) =>
+    if string(v: window) == "0s" then
+        table
+    else
+        table |> aggregateWindow(every: window, fn: fn)
+
+aggregateCounter = (table=<-, pollingInterval=0s, window, groupCols=[]) =>
+    table
+        |> _rateDet(unit: pollingInterval, nonNegative: true)
```
nathanielc: What is the `pollingInterval`? The polling interval should not be needed to correctly understand counters.

stdlib/experimental/query/from.flux
```diff
@@ -0,0 +1,46 @@
...
+        |> group(columns: groupCols)
+        |> _sumWindowDet(window: window)
```
nathanielc: I don't think this does what we want yet. Can you add a test case that shows how the `aggregateCounter` function is used? That will help me understand the intent of the function.

To add a test case see one of the other files that end int `_test.flux`. The necessary parts are a `test` key word statement which needs input data and expected output data for a given function. 

The part that doesn't look correct yet is the `group |> aggregeateWindow` part. I would have expected `aggergateWindow` to not work for this case because there are two separate aggregation steps. I could just be reading the intent wrong, which is where a test case comes in.

PR #2541, opened 2020-02-20
feat: add experimental/aggregate pkg with rate func

jsteenb2:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/aggregate/aggregate_test.flux
```diff
@@ -0,0 +1,116 @@
...
+,,0,2020-02-20T23:00:00Z,2020-02-20T23:01:00Z,host.local,978.9000000000001,2020-02-20T23:00:40Z
+,,0,2020-02-20T23:00:00Z,2020-02-20T23:01:00Z,host.local,22495,2020-02-20T23:00:50Z
+,,0,2020-02-20T23:00:00Z,2020-02-20T23:01:00Z,host.local,7066.2,2020-02-20T23:01:00Z
+"
+
+t_rate = (table=<-) =>
+    table
+        |> range(start: 2020-02-20T23:00:00Z, stop: 2020-02-20T23:01:00Z)
+        |> filter(fn: (r) => r._measurement == "net" and r._field == "bytes_recv")
+        |> aggregate.rate(every: 10s, groupColumns: ["host"], unit: 1s)
```
nathanielc: I'd like to see the test case use an `every` interval that is larger than the raw data interval so we can validate that the `mean` aggregation step is working as expected.

nathanielc: Could you also include data for at least two hosts? So we can see that the computation preserves each host?

nathanielc:
@jsteenb2 

> TODO: need to figure out how to run these tests...

Should just be 

```
$ go generate ./stdlib/... && go test ./stdlib/...
```

PR #2548, opened 2020-02-21
refactor(runtime): move runtime components of the flux package to a runtime package

### LINKED ISSUE #2521: Remove "flux" dependency on "libflux/go/libflux"
The root package `flux` should not depend on Rust code, since it makes almost any client that imports the root package require Rust to build.

Here's a picture of how the repo-root `flux` package depends on `libflux/go/libflux`:

graph

We should make the top-level `flux` package follow the "standard package layout":
https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1

In particular, only the Flux domain types should be in the root package, and nothing in the root package should depend on subpackages, perhaps with the exception of cross-cutting packages like `memory` and `codes`.

The packages `values` and `semantic` currently depend on Rust, but that will be addressed in #2520 and #2519.  The root package can continue to depend on them once those issues are fixed.

The `Eval*` methods that need to execute Flux, as well as `Register*` methods that register stdlib Flux functions should be moved into subpackages.

jsternberg:
Fixes #2521.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
Fixed it. It seems the output from the refactor missed a package and it created the output differently than the generator.

PR #2550, opened 2020-02-21
chore: use non-symlink stdlib path in Rust build

wolffcm:
These are just some minor tweaks that let `pkg-config` builds succeed.


wolffcm:
This was obsoleted by this PR: https://github.com/influxdata/flux/pull/2555

PR #2551, opened 2020-02-21
fix(pagerduty): remove component field from api

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jlapacik:
Note, this is a breaking change, however I have verified that the pagerduty API is only used internally.

jlapacik:
The notification rules in influxdb must be updated to generate the correct code. That will require flux versioning as users will have to regenerate their notifications.

PR #2554, opened 2020-02-24
fix(semantic): remove semantic dependencies on rust/cgo code

fchikwekwe:
This PR removes all code from `package semantic` that depends on rust/cgo and moves it into the `runtime` package. Semantic locations have also been made public in order to allow `flatbuffers_test.go` to be moved the the `semantic_test` package. Moving `flatbuffers_test.go` was necessary to prevent circular dependencies. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


fchikwekwe:
@wolffcm I added an accessor method to keep `PolyType.fb` private. Let me know what you think.

semantic/polytype.go
```diff
@@ -16,6 +16,8 @@ type PolyType struct {
 	fb *fbsemantic.PolyType
 }
 
+func (p PolyType) Fb() *fbsemantic.PolyType { return p.fb }
```
wolffcm: This a small thing, but the caller of this method does not actually need to use the `*fbsemantic.PolyType` for anything, it just needs to know if the pointer is `nil`.  Can you change the name of this method to `IsNil()` and make it return a `bool`?This way we can continue completely hide the underlying representation.  

Also, since this is an exported method, it would be good to have a comment for it.

fchikwekwe: Ok, I'll change that.

semantic/polytype.go
```diff
@@ -16,6 +16,14 @@ type PolyType struct {
...
 }
 
+// For testing we need an exported method that checks whether PolyType.fb is nil.
+func (pt PolyType) IsNil() bool {
+	if pt.fb == nil {
+		return true
+	}
+	return false
+}
+
```
fchikwekwe: @wolffcm is this the change that you were asking for?

semantic/polytype.go
```diff
@@ -16,6 +16,11 @@ type PolyType struct {
 	fb *fbsemantic.PolyType
 }
 
+// For testing we need an exported method that checks whether PolyType.fb is nil.
+func (pt PolyType) IsNil() bool {
+	return pt.fb == nil
+}
+
```
fchikwekwe: @wolffcm is this the change that you were asking for?

wolffcm: Yes, exactly.

PR #2555, opened 2020-02-24
feat: add support for pkg-config

wolffcm:
This patch adds support for `pkg-config`, so that when `go build` is
invoked, `libflux` is built as part of finding the path to the library.


Makefile
```diff
@@ -14,17 +14,20 @@ SHELL := /bin/bash
 GO_TAGS=
 GO_ARGS=-tags '$(GO_TAGS)'
 
-# Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=env FLUX_PARSER_TYPE=rust GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env FLUX_PARSER_TYPE=rust GO111MODULE=on go test $(GO_ARGS)
+
+PKG_CONFIG := bin/$(GOOS)/pkg-config
```
jsternberg: An easier way to do this is to create a script like:

```
#!/bin/bash
exec go run github.com/influxdata/pkg-config "$@"
```

This will cause the script to immediately execute the go command with `go run` from the version specified in `tools.go`. You can then use that script as the `PKG_CONFIG` environment variable. If [this answer](https://stackoverflow.com/questions/23843106/how-to-set-child-process-environment-variable-in-makefile) is to be believed, you should be able to do:

```
export PKG_CONFIG=./scripts/pkg-config`
```

Then you don't need to modify anything else.

tools/tools.go
```diff
@@ -0,0 +1,10 @@
+//+build tools
```
jsternberg: Nitpick: Can you add a space between `//` and `+build`? GoLand complains to me if I don't add a space there.

PR #2556, opened 2020-02-24
feat(semantic): transform semantic nodes back to ast nodes

### LINKED ISSUE #2530: Translate Semantic graph to AST for Filter functions
The `influxdb.source` in vanilla Flux needs to ability to translate semantic graph back to an equivalent AST representation.

NOTE:

This process involves converting semantic to AST, so the round trip of semantic to AST back to semantic should produce the same semantic graph.
Conversely the round trip of AST to semantic to AST does NOT need to produce the same AST.

In other words we are not trying to reproduce the AST the user created, just something semantically equivalent.

jsternberg:
This allows converting semantic nodes back to ast nodes. The original
structure of the ast nodes cannot be restored because information is
lost when converting from ast to semantic, but the structure can be
recovered so that it can be transformed back into semantic.

This method will succeed when comparing:

    semantic -> ast -> semantic

It will not succeed when doing:

    ast -> semantic -> ast

Fixes #2530.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/semantic/ast.go
+++ b/semantic/ast.go
@@ -1,0 +1,236 @@
+package semantic
+
+import (
+	"fmt"
+
+	"github.com/influxdata/flux/ast"
+)
+
+// ToAST will construct an AST from the semantic graph.
+// The AST will not be preserved when going from
+// AST -> semantic -> AST, but going from
+// semantic -> AST -> semantic should result in the same graph.
+func ToAST(n Node) ast.Node {
+	switch n := n.(type) {
+	case *Package:
+		p := &ast.Package{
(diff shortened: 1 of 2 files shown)


jsternberg:
I don't think any of them should. I was a bit concerned over some of the ones at the file level and it does appear like there are semantic nodes that can be generated by the Go parser that cannot be generated by the Rust parser. For example, the Rust parser will always generate a block for functions while the Go parser can do both a block or an expression.

I'm not expecting this to panic though.

Issue #2442, opened 2020-01-30
basic getRecord/tableFind fails


yossigartner:
I'm trying to use the 'getRecord' function as described in the documentation [here](https://docs.influxdata.com/flux/v0.50/stdlib/built-in/transformations/stream-table/getrecord/)

i'm using influxdb cloud Version 2.0.0 (8ff62ab), with flux 0.5v.

the code:
![image](https://user-images.githubusercontent.com/7220867/73436281-e5ccb980-4352-11ea-8c33-e210b328a87e.png)

when i try to 'save' the task, i get the following error:
![image](https://user-images.githubusercontent.com/7220867/73436346-05fc7880-4353-11ea-9f3e-118e6f0f08e1.png)

sanderson:
Is this the full task script or is there more to it?

Yossig:
the full script doesn't event use the r0 variable yet, that's what really bothers me.
(the full script runs OK without this code)

sanderson:
I don't think this is an issue with Flux. I think this is an issue with the task validator that runs when you try to save the task. Can you run this query (line 3 on) successfully in the Data Explorer?

Yossig:
These function are not available in the Data Explorer
![image](https://user-images.githubusercontent.com/7220867/73741373-ca96eb00-4752-11ea-9bcc-a14ea6793eb0.png)

Yossig:
I get this error: "error in query specification while starting program: this Flux script returns no streaming data. Consider adding a "yield" or invoking streaming functions directly, without performing an assignment"

the query:
![image](https://user-images.githubusercontent.com/7220867/73742231-a0462d00-4754-11ea-8be0-a0e62132d941.png)

sanderson:
The error is correct. A task requires a stream of tables as an input and a stream of tables as an output. By using `tableFind()`, you're extracting a single table out of the stream of tables and you no longer have "streaming" data in your task. `getRecord` returns an object literal that represents the first row in the extracted table. It does not return not a stream of tables. 

To get essentially the same result, but structured as a stream of tables, you could:

```js
from(bucket: "test")
  |> range(start: -5m)
  |> filter(fn: (r) => 
      r._measurement == "sample" and
      r.device_id == "999" and
      r._field == "value"
  )
  |> group()
  |> first()
```

mcassaniti:
@sanderson What is the best way to get that into the documentation? I've been using the data explorer to do almost the same query and wondering why I got nothing in return.

[...]

Issue #2387, opened 2020-01-14
Use of one of my custom functions in a map fails


nisley:
I can run and return the value from my custom function getValueInsideTimerange()

When I use it in a map I get
>panic: interface conversion: interface {} is nil, not lang.ExecutionDependencies
![image](https://user-images.githubusercontent.com/1389799/72388218-33ea8780-36da-11ea-8497-9d0b0bdbc6fe.png)

```
getValueInsideTimerange = (earlierTime, laterTime) => {
	abc = from(bucket: "NOAA")
          |> range(start: time(v: 1578833173620), stop: time(v: 1578833173631) )
          |> filter(fn: (r) => r._measurement == "data")
          |> tableFind(fn: (key) => key._field == "speed")
          |> getRecord(idx: 0)
          return abc._value
}

from(bucket: "NOAA")
  |> range(start: -5h)
  |> filter(fn: (r) => r._measurement == "intervals")
  |> map(fn: (r) => 
          ({
            time: r._time,
            datad: float(v: getValueInsideTimerange(earlierTime: 1578833173620, laterTime: 1578833173631) )
          })
        )
```

rockstar:
Here's a stack trace, from slack.

```2020-01-15T16:29:41.320891Z	info	Dispatcher panic	{"log_id": "0KLbyqLW000", "service": "storage-reads", "component": "dispatcher", "error": "panic: interface conversion: interface {} is nil, not lang.ExecutionDependencies"}
goroutine 102350 [running]:
runtime/debug.Stack(0xc013a484e0, 0x69ac800, 0x5ac6155)
	/usr/local/Cellar/go/1.13.4/libexec/src/runtime/debug/stack.go:24 +0x9d
github.com/influxdata/flux/execute.(*poolDispatcher).Start.func1.1(0xc013a48540)
	/Users/nisely/go/pkg/mod/github.com/influxdata/flux@v0.58.0/execute/dispatcher.go:81 +0x192
panic(0x57d9dc0, 0xc013d17230)
	/usr/local/Cellar/go/1.13.4/libexec/src/runtime/panic.go:679 +0x1b2
github.com/influxdata/flux/lang.GetExecutionDependencies(...)
	/Users/nisely/go/pkg/mod/github.com/influxdata/flux@v0.58.0/lang/dependencies.go:27
[...]

PR #2558, opened 2020-02-25
fix: fail gracefully when tableFind does not have an execution context

adriandt:
In the repl, a tablefind result that is captured to a variable, then used in a
map or a filter, causes the repl to attempt to execute a query. It fails with a
panic when it attempts to get the execution context.

A proper solution is part of a bigger design discussion. This patch simply
fails the query gracefully so we can identify the problem more easily and cause
less confusion. The user can be directed to execute on the server process when
encountering this issue.

Design discussion: https://github.com/influxdata/influxdb/issues/13976
Users experiencing this problem: #2442 #2387
Docs: https://github.com/influxdata/docs-v2/issues/735

### DIFF
--- a/stdlib/universe/table_fns.go
+++ b/stdlib/universe/table_fns.go
@@ -88,6 +88,11 @@ func tableFindCall(ctx context.Context, args values.Object) (values.Value, error
 	if err != nil {
 		return nil, errors.Wrap(err, codes.Inherit, "error in table object compilation")
 	}
+
+	if !lang.HaveExecutionDependencies(ctx) {
+		return nil, errors.New(codes.Invalid, "do not have an execution context for tableFind, if using the repl, try executing this code on the server using the InfluxDB API")
+	}
+
 	deps := lang.GetExecutionDependencies(ctx)
 	if p, ok := p.(lang.LoggingProgram); ok {
 		p.SetLogger(deps.Logger)
--- a/stdlib/universe/table_fns_test.go
+++ b/stdlib/universe/table_fns_test.go
@@ -140,6 +140,7 @@ func TestTableFind_Call(t *testing.T) {
 		fn   string
 		// fn      func(key values.Object) (values.Value, error)
 		wantErr error
+		omitExecDeps bool
 	}{
 		{
 			name: "exactly one match 1", // first table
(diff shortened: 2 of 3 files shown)


stdlib/universe/table_fns.go
```diff
@@ -88,6 +88,11 @@ func tableFindCall(ctx context.Context, args values.Object) (values.Value, error
 	if err != nil {
 		return nil, errors.Wrap(err, codes.Inherit, "error in table object compilation")
 	}
+
+	if !lang.HaveExecutionDependencies(ctx) {
```
wolffcm: Unless I'm missing something, `err` is `nil` here, right?  So there is no error to be wrapped or code to inherit?

wolffcm: Another note on this, the error code should be `codes.Invalid` to reflect that the user did something unsupported, since if it is `codes.Internal`, we will hide the message from the user as a security measure.

adriandt: Thank you for the excellent suggestions! Pushed a new version with a more detailed error message, no error wrapping, using Invalid error type and with a test case.

stdlib/universe/table_fns_test.go
```diff
@@ -140,38 +140,52 @@ func TestTableFind_Call(t *testing.T) {
 		fn   string
 		// fn      func(key values.Object) (values.Value, error)
 		wantErr error
+		execDeps bool
 	}{
 		{
 			name: "exactly one match 1", // first table
 			want: tables[0],
 			fn:   `f = (key) => key.user == "user1" and key._measurement == "CPU"`,
+			execDeps: true,
```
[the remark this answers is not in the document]
adriandt: Was wondering if I should do that :) I'll update the PR. It's much nicer with the field left out of the cases that don't concern it.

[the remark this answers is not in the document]
adriandt: I guess I got in the habit of specifying everything, as rust requires you to do (unless you make a default function and explicitly use it). I'd much rather rely on defaults.

adriandt:
@wolffcm when the newly pushed branch passes do I just click "squash and merge" in the "changes approved" box at the bottom of the pull request?

wolffcm:
@adriandt Yep, "squash and merge" will do it right.

PR #2561, opened 2020-02-26
feat: add select function to experimental/query package

sanderson:
This adds a new `select` function to the `experimental/query` package. This function does a lot of string manipulation, which adds an observable cost to using it, but I'm curious to see if we can learn anything about usability.

I'm experimenting with different patterns for selecting data. The `src` parameter expects a string using the following pattern:

```
bucketName:measurement

OR

bucketName:measurement1,measurement2
```

The `fields` parameter expects a comma delimited list of of field keys

It is designed to support one or more measurements, one or more fields, and even measurement and field wildcards (`*`). The following calls all work.

```js
import "experimental/query"

query.select(src: "bucket:measurement", fields: "field1", start: -1h)
query.select(src: "bucket:m1,m2", fields: "field1,field2", start: -1h)
query.select(src: "bucket:*", fields: "*", start: -1h, where: (r) => r.host != "host1")
```

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/experimental/query/from.flux
+++ b/stdlib/experimental/query/from.flux
@@ -1,5 +1,7 @@
 package query
 
+import "strings"
+
 fromRange = (bucket, start, stop=now()) =>
     from(bucket: bucket)
         |> range(start: start, stop: stop)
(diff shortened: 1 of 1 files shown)


stdlib/experimental/query/from.flux
```diff
@@ -24,3 +26,30 @@ inBucket = (
         |> filterMeasurement(measurement)
         |> filter(fn: predicate)
         |> filterFields(fields)
+
+splitData = (v, t=",") => strings.split(v:v, t:t)
+sanitizeList = (v) => strings.replaceAll(v:v, t: ", ", u: ",")
+
+select = (
+    src,
```
jsternberg: Do you think this might be easier if the `src` and `fields` were simplified? I am curious if it would be better to do something like this and split it up:

```
import "influxdata/influxdb"

from = (bucket, start, stop=now()) => influxdb.from(bucket) |> range(start, stop)
select = (tables=<-, m, f) => tables |> filter(fn: (r) => r._measurement == m and r._field == f)
```

This way the query would be simplified to:

```
query.from(bucket: "telegraf", start: -5m)
    |> query.select(m: "cpu", f: "usage_user")
```

If successful, we can make the `from()` a permanent change to `from()`. It is something I have been thinking about.

Another thing that you can do is:

```
select = (tables=<-, m, f="") => {
    s = tables |> filter(fn: (r) => r._measurement == m)
    return if f != "" then s |> filter(fn: (r) => r._field == f) else s
}
```

stdlib/experimental/query/from.flux
```diff
@@ -24,3 +26,30 @@ inBucket = (
...
+    fields,
+    start,
+    stop=now(),
+    where=(r) => true
```
jsternberg: I don't think this adds anything to the query since it is easily replicated with adding `|> filter()` after the query which is more composable.

PR #2564, opened 2020-02-26
feat: influxql-decode-and-series-agg-test

yzhang1991:
This PR added a `-v` flag to `influxql-decode`, defaults to 2 to preserve the old `influxql-decode` behavior.

`-v 1` will ask the tool to generate the line protocol format output, it will be used to load data to v1 and v2 systems for integration testing.

This PR also added a transpiler end-to-end test for series aggregate `difference()`
The flux script matches the output of the transpiler. Changes to the transpiler will be added in a separate PR in the influxdb repo


stdlib/testing/influxql/series_agg_test.flux
```diff
@@ -0,0 +1,704 @@
...
+
+// SELECT difference(f) FROM m GROUP BY *
+t_difference = (tables=<-) => tables
+	|> range(start: influxql.minTime, stop: influxql.maxTime)
+	|> filter(fn: (r) => r._measurement == "m")
+	|> filter(fn: (r) => r._field == "f")
+	|> difference()
+	|> drop(columns: ["_start", "_stop", "_field"])
+	|> map(fn: (r) =>
+		({_time: r._time, difference: r._value}), mergeKey: true)
```
jsternberg: `mergeKey` is deprecated and shouldn't be used. The transpiler should be updated to use `with` instead.

execute/table.go
```diff
@@ -22,6 +22,7 @@ const (
 	DefaultStopColLabel  = "_stop"
 	DefaultTimeColLabel  = "_time"
 	DefaultValueColLabel = "_value"
+	DefaultFieldColLabel = "_field"
```
jsternberg: This is an influxdb concept so it doesn't end up belonging in the flux repository. There's no such thing as a default field column because fields are exclusive to the influxdb source.

If you want a place for this as a constant, the `stdlib/influxdata/influxdb` package is a great place for it. I think it's ok to add this constant to that package.

stdlib/testing/influxql/series_agg_test.flux
yzhang1991: In this case, I will change the transpiler to use the rename() function.
For GROUP BY *, I think I don't have a way to know the name of the tags to add to `with`?

jsternberg: I think rename will end up being better anyway. Right now `rename` performance is the same as everything else, but it is also probably one of the easiest functions to make fast.

internal/cmd/influxql-decode/v1.go
```diff
@@ -0,0 +1,72 @@
...
+
+		var resp influxql.Response
+		if err := json.NewDecoder(f).Decode(&resp); err != nil {
+			return err
+		}
+
+		b := new(bytes.Buffer)
+		for _, res := range resp.Results {
+			for _, s := range res.Series {
+				seriesHeader := fmt.Sprintf("%s,", s.Name)
```
jsternberg: In the future we should update this to use `influxdata/line-protocol`.

yzhang1991: Created an issue https://github.com/influxdata/flux/issues/2566 for this

Issue #2510, opened 2020-02-11
Using Array Style Access to Object with Variable Fails


rawkode:
Apologies for the terrible title, I'm not sure how to describe this except by pasting the code:

Accessing an object through square parens works fine with strings, such as `r["value"]`.

Example:

```
from(bucket: "metrics")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "docker_hub")
  |> filter(fn: (r) => r._field == "pull_count")
  |> map(fn: (r) => ({ r with new_value: r["_value"] }))
```

However, if we pass in a variable for dynamic access - this fails with `type error 7:44-7:57: string != int`

```
key_name = "_value"

from(bucket: "metrics")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "docker_hub")
  |> filter(fn: (r) => r._field == "pull_count")
  |> map(fn: (r) => ({ r with new_value: r["${key_name}"] }))
```

jlapacik:
@rawkode records are not dynamic. Although you can access a record's properties using square parens, `r["value"]` is translated as `r.value`. Using anything other than a string literal inside square parens when accessing a record will result in a type error.

rickspencer3:
1. Is this by design?
2. The error message seems to be totally unrelated to the problem, at least from a user's point of view.

nat45928:
I hit this today as well. This would be a very welcome improvement.

jlapacik:
@rickspencer3 to answer your questions:

1. No this is not by design. Square brackets should be used for array access only, not for objects. It was an early addition to the language that is no longer useful/relevant. It should be removed.
2. In the second example where a dynamic value is used to access the object, the type checker thinks that `r` must be an array, and so it determines that the type within the square brackets should be an integer not a string. The error message can be improved by providing this additional context to the user.

sanderson:
@jlapacik I wouldn't say that it's no longer useful or relevant. I agree that it shouldn't be the primary method for reading object properties, but just recently, we've seen a handful of community use cases where object property keys (in these specific cases, tag keys) contain whitespace and/or special characters. These don't work with dot notation (currently), so bracket notation is the only way to reference those object properties.

```
r = {tag1: "value1", tag2: "value2", "tag with spaces": "value3", "tag.with.period": "value4"}

// won't work
r.tag with spaces

// won't work
r."tag with spaces"

[...]

PR #2565, opened 2020-02-26
feat: experimental geo package

alespour:
This PR contains `experimental/geo` package that provides functions for geographic location filtering and grouping. Internally it uses golang implementation of S2 Geometry Library (https://s2geometry.io/). It is designed to work on a schema with a tag `_ci` which contains S2 cell ID token of level decided by the user and fields `lat` and `lon`.

Latitude and longitude fields are commonly present in geotemporal input data. The tag in the form of S2 Cell ID token can be easily calculated using Google S2 library which is available for many languages. A simple Telegraf processor plugin that adds the tag to line protocol data might also be provided as possible solution (https://github.com/influxdata/telegraf/pull/7087).

Drawbacks:
- The tag and fields keys are hard coded. When/If array access using variable issue (#2510) is fixed, the keys can become functions parameters with default values that are now hard coded.

### Done checklist
- [x] Test cases written


stdlib/experimental/geo/geo.flux
```diff
@@ -0,0 +1,134 @@
...
+
+// Returns level of specified cell ID token.
+builtin getLevel
+
+//
+// Flux
+//
+
+// Gets level of cell ID tag `_ci` from the first record from the first table from the stream.
+_detectCiLevel = (tables=<-) => {
```
jlapacik: Why do we only care about the first record of the first table?

stdlib/experimental/geo/geo.flux
```diff
@@ -0,0 +1,134 @@
...
+      )
+}
+
+// Filters records by a box, a circle or a polygon region.
+// It is an exact filter and must be used after `toRows()` because it requires `lat` and `lon` columns in input row sets.
+strictFilter = (tables=<-, box={}, circle={}, polygon={}) =>
+  tables
+    |> filter(fn: (r) =>
+      containsLatLon(box: box, circle: circle, polygon: polygon, lat: r.lat, lon: r.lon)
+    )
```
jlapacik: Because this function requires `lat` and `lon` to be present on the input rows, the closure passed to `filter` should assert that. For example:
```
strictFilter = (tables=<-, containsLatLon) =>
    tables
        |> filter(fn: (r) => containsLatLon(lat: r.lat, lon: r.lon))
```

stdlib/experimental/geo/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+This filter function is intended to be fast as it uses `_ci` tag to filter records.
+If precise filtering is needed, `strictFilter()` may be used later (after `toRows()`).
+
+Example:
+```
+from(bucket: "rides")
+  |> range(start: 2019-11-01T00:00:00Z)
+  |> filter(fn: (r) => r._measurement == "taxi")
+  |> geo.gridFilter(box: {minLat: 40.51757813, maxLat: 40.86914063, minLon: -73.65234375, maxLon: -72.94921875})
```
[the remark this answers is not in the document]
wolffcm: @jlapacik The standard library `filter` maintains no state itself, it just calls its lambda for each row it receives.  By implementing a `gridFilter` stream function, it's possible to computing the tiling (a set of S2Cells that cover the region) just one time, and then use it for every row. I like the composability of your proposal, but it seems like it would have some performance impact.

stdlib/experimental/geo/README.md
```diff
@@ -0,0 +1,201 @@
+## Package `geo`
+
+The package provides functions for geographic location filtering and grouping.
+It uses golang implementation of S2 Geometry Library [https://s2geometry.io/].
+It is designed to work on a schema with a tags `_ci` which contains S2 cell ID token
+of level decided by the user and fields `lat`, `lon`.
+
+The `_cid` tags hold value of cell ID expressed as token (`s2.CellID.ToToken()`) of corresponding level.
```
wolffcm: I think it would be good to use `_s2cell` instead of `_ci` just so the meaning is really clear.  Likewise, maybe `_tid` should be `_track`?

stdlib/experimental/geo/containslatlon_test.go
```diff
@@ -0,0 +1,153 @@
+package geo_test
+
+import (
+	"context"
+	_ "github.com/influxdata/flux/builtin"
```
wolffcm: nit, we have a convention of putting all the Go stdlib imports first, followed by a blank line, followed by all the other imports in alphabetical order.

stdlib/experimental/geo/README.md
timhallinflux: Telegraf team is recommending that we do NOT have "_"-based field names.  I believe we need to specify the tag name which represents the geohash in the function.  We could default to a tag value -- that we should make the default in the corresponding Telegraf plugin.  Similiarly, we should be able to default the lat/lon field names, but if the use wants to override this...we should allow it.  I believe there are other function signatures that do something similar?  Please suggest one to follow as an example.

stdlib/experimental/geo/geo.flux
alespour: The assumption is that all records in all tables have cell id tag value of the same level.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -53,6 +53,12 @@ pub fn builtins() -> Builtins<'static> {
             "experimental/bigtable" => maplit::hashmap! {
                      "from" => "forall [t0] where t0: Row (token: string, project: string, instance: string, table: string) -> [t0]",
             },
+            "experimental/geo" => maplit::hashmap! {
+                     "containsLatLon" => "forall [] (?box: {}, ?circle: {}, ?polygon: {}, lat: float, lon: float) -> bool",
+                     "getGrid" => "forall [] (?box: {}, ?circle: {}, ?polygon: {}, ?minSize: int, ?maxSize: int, ?level: int, ?maxLevel: int) -> {level: int | set: [string]}",
```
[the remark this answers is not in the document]
alespour: It is just that could not make it work when I tried to specify the records further, when compiling libstd I was getting errors like "Invalid format for optional arguments" or "cannot unify {{}} with {lat:float | lon:float | radius:float | {}}"

stdlib/experimental/geo/README.md
```diff
@@ -0,0 +1,201 @@
...
+#### Function definition
+
+```
+toRows = (tables=<-, correlationKey=["_time"]) =>
+  tables
+    |> pivot(
+      rowKey: correlationKey,
+      columnKey: ["_field"],
+      valueColumn: "_value"
+    )
```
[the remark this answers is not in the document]
alespour: The purpose is to return all fields, `lat` and `lon` just point to a location (and are kind of metadata which cannot be stored to tags), but on the input there are often more fields such as speed etc in live tracking use cases, or tip amount and ride distance for taxi drop off points etc.

stdlib/experimental/geo/README.md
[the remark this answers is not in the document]
alespour: @wolffcm Your comment was correct I thought, now I am confused, as of now, there is a call to builtin fn `getGrid` in the `gridFilter` to have precomputed tiling for given region.

@jlapacik I think `gridFilter` cannot be implemented as suggested because at this time of query execution, `lat` and `lon` are in separate streams/tables, this function filters records by tag value.

stdlib/experimental/geo/geo.flux
timhallinflux: level here refers to the precision of the geohash.  We are going to make an assumption about the geohash...and try to infer it's precision based on the first record.

stdlib/experimental/geo/README.md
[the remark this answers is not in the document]
wolffcm: Yes, thinking it over now, you're right, my original comment was right, and I confused myself.  Since the expensive calls to `getFilter` is done outside of the call to `filter(fn: (r) => ...)`, what you have will perform well.

stdlib/experimental/geo/README.md
wolffcm: I'll double check, but I don't think that it's possible to handle dynamic field names for functions that are written in Flux.  We *can* do it for builtin functions that are written in Go though, but that doesn't help with the way this code is structured. There is an existing function to address this but it's a ways off: https://github.com/influxdata/flux/issues/1122

It might be possible to workaround this limitation with creative use of the `rename` function, but we don't have any examples of that.  How important is it to be able to configure the field names for `lat`, `lon`, `cellid` etc fields?

alespour: `_ci` renamed to `s2cellID` and box/circle/polygon unified into single region parameter.

wolffcm:
@alespour The type changes look great, thanks for the prompt update.  Everything looks good, except per this comment you'll be renaming the tag key to `s2_cell_id`, so once that change is in, I'll go ahead and merge this.

wolffcm:
Just to be clear, only the column name should be assumed to be `s2_cell_id`, the function parameter names should continue to be camelCase.  This seems to be the precedent we have elsewhere in Flux.

PR #2568, opened 2020-02-27
feat(stdlib): add influxdb source

### LINKED ISSUE #2470: Create influxdb.from() source in Flux repository
The only way to communicate with InfluxDB is to use the built-in influxdb from that exists within the server. There's no current way to use the Flux binary to query InfluxDB through the public API.

This can make it harder to interact with InfluxDB from Flux itself.

This issue is about making a Flux-native influxdb source that uses the public HTTP API to retrieve results. We would implement a version that takes the `from() |> range() |> filter()` chain and serializes the query to an AST and uses the public API to execute the query and read the data into a local Flux table.

jsternberg:
The influxdb source will recognize the `from |> range` pattern and it
will convert it to an AST and serialize that to a remote influxdb v2
server.

If a `filter` is included afterwards, it will be converted back to AST
and be included in the influxdb from call.

Fixes #2470.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/interpreter/package.go
+++ b/interpreter/package.go
@@ -18,6 +18,9 @@ type Package struct {
 	// name is the name of the package.
 	name string
 
+	// path is the canonical import path that is used to import this package.
+	path string
+
 	// object contains the object properties of this package.
 	object values.Object
 
@@ -26,16 +29,17 @@ type Package struct {
 	sideEffects []SideEffect
 }
 
-func NewPackageWithValues(name string, obj values.Object) *Package {
+func NewPackageWithValues(name, path string, obj values.Object) *Package {
 	return &Package{
 		name:   name,
+		path:   path,
 		object: obj,
 	}
 }
 
 func NewPackage(name string) *Package {
 	obj := values.NewObject(semantic.NewObjectType(nil))
-	return NewPackageWithValues(name, obj)
+	return NewPackageWithValues(name, "", obj)
 }
 
 func (p *Package) Copy() *Package {
@@ -50,9 +54,17 @@ func (p *Package) Copy() *Package {
 		sideEffects: sideEffects,
 	}
 }
+
+// Name returns the package name.
 func (p *Package) Name() string {
 	return p.name
 }
+
+// Path returns the canonical import path for this package.
+func (p *Package) Path() string {
+	return p.path
+}
+
 func (p *Package) SideEffects() []SideEffect {
 	return p.sideEffects
 }
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -124,7 +124,14 @@ pub fn builtins() -> Builtins<'static> {
             "influxdata/influxdb" => maplit::hashmap! {
                 // This is a one-or-the-other parameters function
                 // https://github.com/influxdata/flux/issues/1659
-                "from" => "forall [t0, t1] (?bucket: string, ?bucketID: string) -> [{_measurement: string | _field: string | _time: time | _value: t0 | t1}]",
+                "from" => r#"forall [t0, t1] (
+                    ?bucket: string,
+                    ?bucketID: string,
+                    ?org: string,
+                    ?orgID: string,
+                    ?host: string,
+                    ?token: string
+                ) -> [{_measurement: string | _field: string | _time: time | _value: t0 | t1}]"#,
                 // exactly one of (bucket, bucketID) must be specified
                 // exactly one of (org, orgID) must be specified
                 // https://github.com/influxdata/flux/issues/1660
--- a/stdlib/influxdata/influxdb/from.go
+++ b/stdlib/influxdata/influxdb/from.go
@@ -4,18 +4,46 @@
 package influxdb
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/csv"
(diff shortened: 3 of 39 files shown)


stdlib/influxdata/influxdb/rules.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return n, true, nil
+}
+
+// DefaultFromAttributes is used to inject default attributes
+// for the various from attributes.
+//
+// This rule is not added by default. Each process must fill
+// out the suitable defaults and add the rule on startup.
+type DefaultFromAttributes struct {
```
affo: I really like this, the influxDB cli can inject user params right?

Anyways, the flux cli could inject params easily. I like it 👍

stdlib/influxdata/influxdb/from.go
```diff
@@ -69,3 +147,300 @@ func (s *FromProcedureSpec) Copy() plan.ProcedureSpec {
...
+		return nil, err
+	}
+	if s.spec.Token != nil {
+		req.Header.Set("Authorization", "Token "+*s.spec.Token)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	return req.WithContext(ctx), nil
+}
+
+func (s *source) newRequestBody() ([]byte, error) {
```
affo: Wondering if all  this logic should be sometime in the future embedded in our 2.x go client.

For now, it only allows to query with a `string` (no `ast`), and returns raw results or deserialized into a structure whose field names matches the column names.

If we add the `ast` kind of  request to that, `flux` could depend on that and decode the raw response.

jsternberg: It should, but to my knowledge, the client is in the influxdb repository and we can't depend on it. That's why there's some duplication that I am hoping eventually can be removed. But the http code to make the queries isn't too difficult on its own and most of it is in the flux repository so it was easier to redo the small section that was in influxdb.

jsternberg:
The `FromKind` procedure spec is the default one that exists so it can be replaced with a physical procedure spec. In this case, the flux project only knows how to do remote calls. There's no embedded storage engine inside of it.

So the `FromRemoteKind` is meant to be the physical procedure that implements remotely reading from an influxdb database while the storage specific one will be in influxdb and injected as part of planner rules.

I've worked on adding some tests. The only thing that should be untested now is the planner rules.

libflux/src/flux/semantic/builtins.rs
```diff
@@ -124,7 +124,14 @@ pub fn builtins() -> Builtins<'static> {
...
                 // This is a one-or-the-other parameters function
                 // https://github.com/influxdata/flux/issues/1659
-                "from" => "forall [t0, t1] (?bucket: string, ?bucketID: string) -> [{_measurement: string | _field: string | _time: time | _value: t0 | t1}]",
+                "from" => r#"forall [t0, t1] (
+                    ?bucket: string,
+                    ?bucketID: string,
+                    ?org: string,
+                    ?orgID: string,
+                    ?host: string,
+                    ?token: string
```
fchikwekwe: I don't really understand this change. Why are we now able to accept these additional optional parameters? Seems like a big change unless I'm missing something. Are org and ordID mutually exclusive here like they are in `to()`.

interpreter/package.go
```diff
@@ -26,16 +29,17 @@ type Package struct {
...
 		name:   name,
+		path:   path,
 		object: obj,
 	}
 }
 
 func NewPackage(name string) *Package {
 	obj := values.NewObject(semantic.NewObjectType(nil))
-	return NewPackageWithValues(name, obj)
+	return NewPackageWithValues(name, "", obj)
```
fchikwekwe: Is an empty path like this the current directory?

stdlib/influxdata/influxdb/from.go
```diff
@@ -4,18 +4,46 @@
...
+	FromRemoteKind = "influxdata/influxdb.fromRemote"
 )
 
-const FromKind = "from"
+// NameOrID signifies the name of an organization/bucket
+// or an ID for an organization/bucket.
+type NameOrID struct {
+	ID   string
+	Name string
+}
```
fchikwekwe: I just wanna understand, so in this iteration of from, we're able to accept an organization name, a bucket name, an organization ID or a bucket ID? Confirming because that's different from what I previously understood.

libflux/src/flux/semantic/builtins.rs
jsternberg: Yes, both org and orgID are mutually exclusive. The `org` is needed so that we can specify the org in the flux binary to retrieve from because flux isn't org-aware the way influxdb is, but influxdb requires the organization. Inside of influxdb, the `org` will get filled in automatically.

The reason for `orgID` is because the influx cli can accept an orgID instead and so we need to send that instead for the influxdb repl.

interpreter/package.go
jsternberg: It means we don't know the path, but this function is mostly obsolete and doesn't work anymore. It's used in some tests where we don't run into this issue with the path.

stdlib/influxdata/influxdb/from.go
jsternberg: We are able to accept a bucket name or a bucket id and we are also able to accept an organization name or organization id. So you can do `from(org, bucket)`, `from(orgID, bucket)`, `from(org, bucketID)`, or `from(orgID, bucketID)`. But you can't do `from(org, orgID)` or `from(bucket, bucketID)` and for flux itself, you are going to need to specify both the bucket and org. For the influx cli, you won't need to specify the org.

stdlib/influxdata/influxdb/from.go
affo: I think the client is in a separate repo: https://github.com/influxdata/influxdb-client-go

Anyways, it was just a suggestion for the future, nothing that should block this PR

PR #2571, opened 2020-03-02
refactor(stdlib): improve the post physical validation error messages for the influxdb source

jsternberg:
The error messages for the influxdb source have been improved to give
better hints about what to do when rules cannot be rewritten to the
proper physical forms for whatever reason.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
These validations caused some of the planner tests to fail since now we have a real influxdb source inside of flux so I just pushed some changes to the tests. Can you re-review that specific commit?

Issue #2570, opened 2020-03-01
Algorithm W: Overwriting an existing field in a map will result in the property being present twice


jsternberg:
When used as a return value for a function, something like:

```
map(fn: (r) => ({r with _value: int(v: r._value)}))
```

This causes the type for the return value to be `{_value:int, _value:float}`. It should not have the same property twice. The first property should overwrite the second one.

jsternberg:
This only seems to happen when overwriting a record with a different type. No issue gets caused when the types are the same. Also, this affects the `testing/usage/storage` test for the end to end tests.

I tried to have the code only take the first value instead of iterating through them all, but this causes some of the pandas tests to fail because the object starts to have something like `{_value:object, _value:object}` and the first object has an invalid value in it for some reason. I didn't investigate that more.

jsternberg:
There seem to be tests for this in the end to end tests but no tests for this in the map tests itself. As part of this issue, we should make sure that some tests for overwriting values with a different type are specifically covered in the map tests.

PR #2574, opened 2020-03-02
test(stdlib): unskip usage tests

jlapacik:
These tests used to hang but are no longer. One test (storage) was
kept skipped as it is failing now with another error (map type error).

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/flux_test.go
+++ b/stdlib/flux_test.go
@@ -50,9 +50,7 @@ var skip = map[string]map[string]string{
 		"partition_strings_splitN":       "pandas. map does not correctly handled returned arrays (https://github.com/influxdata/flux/issues/1387)",
 	},
 	"testing/usage": {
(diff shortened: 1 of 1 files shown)


stdlib/flux_test.go
```diff
@@ -50,9 +50,7 @@ var skip = map[string]map[string]string{
 		"partition_strings_splitN":       "pandas. map does not correctly handled returned arrays (https://github.com/influxdata/flux/issues/1387)",
 	},
 	"testing/usage": {
-		"api":     "todo(algow): stalls",
 		"storage": "todo(algow): stalls",
```
jsternberg: Can you change this message to reference the issue for this? https://github.com/influxdata/flux/issues/2570

jlapacik: Good catch. Yes.

PR #2579, opened 2020-03-03
 feat(stdlib/testing): hand transpile cumulative_sum

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


fchikwekwe:
@adriandt team assigner is weird when you ask for multiple reviews. Feel free to review as well!

fchikwekwe:
@adriandt I think that's probably right. Looking at the way the AST is used when end to end tests are run, it does look like its compiled and checked for correctness. @wolffcm might know a little bit more about why this is as well.

wolffcm:
We generate a `_gen.go` source file for _every_ `.flux` source file in the Flux repo.  This is an artifact of how Go works.  When we import `flux` into `influxdb` (for example) we only get the Go source files, I believe. If we want the Flux sources available to other modules, we need to capture them as Go code.

This lets us run the Flux test scripts `*_test.flux` in the influxdb repo and test them against real InfluxDB sources.  I believe we also need to export the sources for the Flux standard library so that Flux can understand the standard library when it runs there... I'm a little fuzzy on that though, and some of the details there are changing a bit due to the algo-w work.

adriandt:
Thank you @fchikwekwe and @wolffcm !

PR #2580, opened 2020-03-04
fix: ignore properties output by map that are not visible

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/flux/semantic/builtins.rs
+++ b/libflux/src/flux/semantic/builtins.rs
@@ -350,7 +350,7 @@ pub fn builtins() -> Builtins<'static> {
                 "splitAfter" => "forall [] (v: string, t: string) -> [string]",
                 "splitN" => "forall [] (v: string, t: string, n: int) -> [string]",
                 "splitAfterN" => "forall [] (v: string, t: string, i: int) -> [string]",
-                "joinStr" => "forall [] (arr: [string], v: string) -> {}",
+                "joinStr" => "forall [] (arr: [string], v: string) -> string",
                 "strlen" => "forall [] (v: string) -> int",
                 "substring" => "forall [] (v: string, start: int, end: int) -> string",
             },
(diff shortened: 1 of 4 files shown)


libflux/src/flux/semantic/builtins.rs
```diff
@@ -350,7 +350,7 @@ pub fn builtins() -> Builtins<'static> {
                 "splitAfter" => "forall [] (v: string, t: string) -> [string]",
                 "splitN" => "forall [] (v: string, t: string, n: int) -> [string]",
                 "splitAfterN" => "forall [] (v: string, t: string, i: int) -> [string]",
-                "joinStr" => "forall [] (arr: [string], v: string) -> {}",
```
jsternberg: Oh I didn't realize that the reason for the pandas failure was just a type failure.

stdlib/universe/map.go
```diff
@@ -248,7 +248,10 @@ func (t *mapTransformation) createSchema(b execute.TableBuilder, m values.Object
 	}
 
 	props := make(map[string]semantic.Nature, numProps)
-	for i := 0; i < numProps; i++ {
+	// Deduplicate the properties in the return type.
```
jsternberg: Ok so this does work and the error I saw with pandas was a type error unrelated to this. I think this looks good and we will have to add a helper method for this at some point later. This is perfect.

libflux/src/flux/semantic/builtins.rs
jlapacik: Yeah this wasn't caught before because we were looking at the wrong property which had a valid type (integer). We never actually added the invalid property to the map.

Note if we added a new kind constraint asserting that rows returned from any of the transformation functions could only contain basic types, this kind of error could be caught at compile time. Future work.

PR #2583, opened 2020-03-05
fix: don't construct a compiler.compilerScope with a nil value.Scope

adriandt:
In the case of the influx query command, it is possible for a nil scope to be
present in the Spec. In this case the system attempts to build a
compiler.compilerScope using a nil value.Scope as the embedded anonymous field,
resulting in an eventual panic when passed to interface methods that use it as
the receiver object.

Instead return nil when attempting to use nil as the base scope,
indicating there is no scope. The function evaluation copes with with this and
creates an empty scope if none is present.

refs influxdata/influxdb#17062

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -1161,3 +1161,9 @@ func TestCompileAndEval(t *testing.T) {
 		})
 	}
 }
+
+func TestToScopeNil(t *testing.T) {
+	if compiler.ToScope(nil) != nil {
+		t.Fatal("ToScope made non-nil scope from a nil base")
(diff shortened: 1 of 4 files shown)


adriandt:
Not sure what to do about test cases as this is only a problem in the `influx query` command, and we don't test that.

yzhang1991:
Can you run `make checkfmt` to check the result?

adriandt:
> Can you run `make checkfmt` to check the result?

Ah, I'll get this right eventually :) Actually did run it, but neglected to include the changes in the PR.

PR #2591, opened 2020-03-05
test(libflux): C program for exercising C API

### LINKED ISSUE #2289: Build memory checker tool for CI
Since we are using `unsafe` in Rust and a C API to pass data structures between Rust and Go, we should be running some kind of memory checking tool (like `valgrind`) in CI.

One way to do this would be:
- Write a simple C app that exercises all of the C functions exposed by Rust
- Run that C app in `valgrind` and have it look for leaks, double frees, etc
- Make CI fail if `valgrind` finds any issues.

wolffcm:
This creates a simple program that exercises our C API, and adds some targets to our Makefile to build it and run it in valgrind.

Fixes #2289.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ target
 .tern-project
 .idea
 /coverage.txt
+/libflux/c/libflux_memory_tester
 /libflux/pkg
 /libflux/site/node_modules
 /libflux/.cache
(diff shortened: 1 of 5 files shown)


wolffcm:
Note, it's not possible to run valgrind on MacOS, but you can run it on Linux or in a our build container.

PR #2593, opened 2020-03-06
refactor(testing): diff should compare floats with tolerance

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/testing/diff.go
+++ b/stdlib/testing/diff.go
@@ -14,6 +14,7 @@ import (
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/flux/runtime"
+	"gonum.org/v1/gonum/floats"
 )
 
 const DiffKind = "diff"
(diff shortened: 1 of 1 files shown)


jlapacik:
This fixes `TestFluxEndToEnd/testing/influxql/aggregate_group_by_time` in influxdb which is currently skipped. See https://github.com/influxdata/influxdb/issues/17064.

stdlib/testing/diff.go
```diff
@@ -460,6 +461,27 @@ func (t *DiffTransformation) diff(key flux.GroupKey, want, got *tableBuffer) err
 	return nil
 }
 
+// Arbitrary floating point tolerance
```
rockstar: The arbitrary nature of this number rubs me the wrong way, but I realize float precision is a hard thing to handle. Does this have impact on our customers?

jlapacik: It's definitely arbitrary. I tried to pick some sane conservative defaults. Our end-to-end tests I think are the biggest consumers of `diff`. This will affect tables with very minor floating point differences, but it seems like that's what customers would expect in most circumstances.

PR #2598, opened 2020-03-09
fix(http): add response reader as dependency to tune response size

affo:
This PR set the limit for reading a response from Http get and post to 100MB.

~It does it by adding a new injectable dependency, so that it is possible to tune limits by overriding the dependency.~

It wraps the HTTP client with a `RoundTripper` that limits reading from `response.Body`.


affo:
FYI @rickspencer3 

Also added e new experimental function, so that you can:

```
import "experimental/csv"

url="https://whatever/NOAA_db.csv"
csv.from(url: url)
```

A shortcut for what you where doing 👍

stdlib/experimental/http/http_experimental.go
```diff
@@ -31,7 +27,7 @@ var get = values.NewFunction(
 			"timeout": semantic.Duration,
 		},
 		Required: []string{"url"},
-		Return:   semantic.NewObjectPolyType(map[string]semantic.PolyType{"statusCode": semantic.Int, "headers": semantic.Object, "body": semantic.Bytes}, semantic.LabelSet{"status", "headers", "body"}, nil),
+		Return:   semantic.NewObjectPolyType(map[string]semantic.PolyType{"statusCode": semantic.Int, "headers": semantic.Object, "body": semantic.Bytes}, semantic.LabelSet{"statusCode", "headers", "body"}, nil),
```
nathanielc: Why does the `status` need to change to `statusCode`?

jsternberg: It looks like the previous code was an error. The poly type has `statusCode` but the label set had `status`.

dependencies.go
```diff
@@ -24,7 +27,17 @@ const dependenciesKey key = iota
...
+	// that must be used when reading the response body. This may add some security.
+	// Example usage:
+	// ```
+	// deps := flux.GetDependencies(ctx)
+	// client, reader, _ := deps.HTTPClient()
+	// response, _ := client.Do(req)
+	// r := reader(response.Body)
+	// ...
+	// ```
+	HTTPClient() (http.Client, func(body io.Reader) io.Reader, error)
```
jsternberg: It seems to me this can be done by wrapping the `http.Client` itself or modifying the `RoundTripper`.

```
func LimitHTTPBody(client http.Client, size int) http.Client {
    // The client is already a struct so it was already copied
    // which makes this safe.
    if client.Transport == nil {
        client.Transport = http.DefaultTransport
    }
    client.Transport = roundTripLimiter{RoundTripper: client.Transport, max: 1024}
    return client
}

type roundTripLimiter struct {
    http.RoundTripper
    max int
}

func (l roundTripLimiter) RoundTrip(r *http.Request) (*http.Response, error) {
    response, err := l.RoundTripper.RoundTrip(r)
    if err != nil {
        return nil, err
    }
    response.Body = newLimitReader(response.Body, l.max)
    return response, nil
}
```

This method also means that no other code needs to change and I think this does the same thing.

affo: Updating, thank you @jsternberg

dependencies/http/http.go
```diff
@@ -6,11 +6,39 @@ import (
...
+	http.RoundTripper
+	limit int64
+}
+
+func (l roundTripLimiter) RoundTrip(r *http.Request) (*http.Response, error) {
+	response, err := l.RoundTripper.RoundTrip(r)
+	if err != nil {
+		return nil, err
+	}
+	response.Body = http.MaxBytesReader(nil, response.Body, l.limit)
```
jsternberg: Is it ok if null is passed in here?

affo: Yay, it seems so.

The only reference I found is: https://stackoverflow.com/questions/52698194/limit-bytes-to-read-from-http-response

I am updating to discard

affo:
@jsternberg ok, now I think we are ok.

Previous behaviour with `MaxRespBytes` was erroring on a response too large.
Now I wrapped the limited reader into a limited read closer.

Look at the test for current behavior.

PR #2602, opened 2020-03-09
feat: storing comments in the AST and preserving on format

### LINKED ISSUE #1959: Preserve Comments when parsing Flux source code
The Flux parser should preserve comments in the AST and `ast.Format` should be able to reproduce the comments in the AST.

adrian-thurston:
Collecting comments during scanning and storing them in the AST. Including the
comments in AST formatting. In general, comments are placed on their own line
and are therefore assumed to apply to the code that follows. Updates to the
formatting rules can be added as desired.

Comments are collected in the scan function, where they were previously
discarded. The comments are arranged into a linked list as they are encountered
and then attached to the next real token that is generated by the parser.
Comments at the end of the file are attached to the EOF token.

During parsing, these comments are extracted from the tokens and stored in the
AST, generally at locations where the tokens they were attached to would be
stored in the AST, if they were kept. This requires adding comment list
optionals to the AST in various locaitons. To minimize AST changes, they are
also stored in the baseNode, if there is just one comment list to store. For
example, comments ahead of identifiers are stored in the base node. But
comments ahead of ( and ) are added to the AST.

Regexes need some special care. There is an unread() call once in the parser
when we've got a DIV and really want a REGEX open. At this time we need to put
the comments back as well, otherwise they are lost.

Preserving comments ahead of , in array expressions requires adding a new AST
struct to wrap the expressions in the ArrayExpr. We cannot use the base node in
the comment, since the expression may be only a nonterminal and the base node's
comment field may be in use for leading comments.

Preserving comments ahead of WITH requires adding a new AST struct to wrap the
identifier and keep any comments that appear ahead of the WITH keyword.

This commit also includes a boxing of all the variants in enum Statement. Since
we are changing the size of some of the AST data structures, clippy warns about
enum variant sizes.

Fixes #1959

### DIFF
--- a/libflux/src/flux/ast/check/tests.rs
+++ b/libflux/src/flux/ast/check/tests.rs
@@ -57,6 +57,7 @@ fn test_check_collect_existing_error() {
                 source: Some(String::from("a = 1\nb=2\nc=a+b")),
             },
             errors: vec![String::from("error 1")],
+            ..BaseNode::default()
         },
         name: String::from("test_check_collect_existing_error"),
         metadata: String::new(),
@@ -70,7 +71,7 @@ fn test_check_collect_existing_error() {
                     end: Position { line: 1, column: 6 },
                     source: Some(String::from("a = 1")),
                 },
-                errors: vec![],
+                ..BaseNode::default()
             },
             id: Identifier {
                 base: BaseNode {
@@ -80,7 +81,7 @@ fn test_check_collect_existing_error() {
                         end: Position { line: 1, column: 2 },
                         source: Some(String::from("a")),
                     },
-                    errors: vec![],
+                    ..BaseNode::default()
                 },
                 name: String::from("a"),
             },
@@ -93,10 +94,12 @@ fn test_check_collect_existing_error() {
                         source: Some(String::from("1")),
                     },
                     errors: vec![String::from("error 2"), String::from("error 3")],
+                    ..BaseNode::default()
                 },
                 value: 1,
             }),
         }))],
+        eof: None,
     };
     let got = check(walk::Node::File(&file));
     assert_eq!(3, got.len());
--- a/libflux/src/flux/ast/mod.rs
+++ b/libflux/src/flux/ast/mod.rs
@@ -190,19 +190,19 @@ impl Expression {
 #[serde(tag = "type")]
 pub enum Statement {
     #[serde(rename = "ExpressionStatement")]
-    Expr(ExprStmt),
+    Expr(Box<ExprStmt>),
     #[serde(rename = "VariableAssignment")]
     Variable(Box<VariableAssgn>),
     #[serde(rename = "OptionStatement")]
     Option(Box<OptionStmt>),
     #[serde(rename = "ReturnStatement")]
-    Return(ReturnStmt),
+    Return(Box<ReturnStmt>),
     #[serde(rename = "BadStatement")]
-    Bad(BadStmt),
+    Bad(Box<BadStmt>),
     #[serde(rename = "TestStatement")]
     Test(Box<TestStmt>),
     #[serde(rename = "BuiltinStatement")]
-    Builtin(BuiltinStmt),
+    Builtin(Box<BuiltinStmt>),
 }
 
 impl Statement {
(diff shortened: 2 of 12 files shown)


libflux/src/flux/ast/check/tests.rs
```diff
@@ -57,6 +57,7 @@ fn test_check_collect_existing_error() {
                 source: Some(String::from("a = 1\nb=2\nc=a+b")),
             },
             errors: vec![String::from("error 1")],
+            ..BaseNode::default()
```
nathanielc: what is the `..` syntax in Rust? Its new to me.

libflux/src/flux/ast/mod.rs
```diff
@@ -319,6 +333,10 @@ impl BaseNode {
     pub fn is_empty(&self) -> bool {
         self.errors.is_empty() && !self.location.is_valid()
     }
+
+    pub fn add_comments(&mut self, comments: CommentList) {
+        self.comments = comments;
```
nathanielc: I noticed comment list is a singly linked list. Does this method need to check if `self.comments` already has comments? The name `add_comments` implies it might already have comments and we are adding new ones. If this is really a replace/set operation should we rename the function to `set_comments`? Also does the function need to be public?

libflux/src/flux/ast/mod.rs
```diff
@@ -367,6 +385,8 @@ pub struct File {
     #[serde(deserialize_with = "deserialize_default_from_null")]
     pub imports: Vec<ImportDeclaration>,
     pub body: Vec<Statement>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub eof: CommentList,
```
nathanielc: So if I understand correctly, comments are associated with the trailing token/node, but in the case of comments at the end of the file there is no token so we simple capture them here as part of the File node?

So if I am walking the AST looking for comments I can expect to find them in the BaseNode as part of each node  or I will find comments inside various other nodes like the File and Block nodes to capture comments that do not have a clear association with a token/node. Does that sound like I have it right?

libflux/src/flux/ast/mod.rs
adrian-thurston: The name `set_comments` is definitely more appropriate. I updated that. About public ... good question ... just verified that it needs to be public because BaseNode is defined in the `ast` module, but comments are populated in `parser` module.

libflux/src/flux/ast/check/tests.rs
adrian-thurston: It says populate the remaining, unspecified fields with values from some other node. I would have liked to make more use of this, especially in the testing code, however it only works well if there are simple defaults available. When parse tree nodes are made up of directly contained, or boxed-contained values (non-optional) you have to make a default that essentially cascades into a full parse tree for a valid expression, because the default itself must be complete. I decided to avoid that.

It would be great if rust let you made incomplete defaults that you could then draw on using `..` to fill the remaining fields in from reasonable defaults. As far as I could tell, you can't.

This represents my initial learnings in the language, so maybe there are better ways?

libflux/src/flux/ast/mod.rs
adrian-thurston: That's right, the comments at the end of the file are a special case because there is no token in front to associate them with. You can think of it as preserving the EOF token.

In Colm, comments and other ignored things can be attached to the front and back of tokens. It requires altering the scanner quite a bit. After scanning a token you have to scan ahead only comments and attach them to the token, before you send the token off. I actually tried this approach, but kept running into some strange behavior that I couldn't explain. But we decided to go ahead with just comments in front so I dropped that investigation.

For our case, comments after tokens is mostly overkill, with one exception:

let a = 1; // set a to one

It's nice to have comments at the end of a line attached to the end of the last token. For this we will need special case added to the formatter. It's currently not smart enough to preserve this. I don't do much commenting that way, but I believe lots of people do.

adrian-thurston:
@jlapacik Yeah I was hoping for that too. I tried for a bit with a vector of comment lists in the base node so we can accommodate nodes with multiple comment lists like `( <Expr> )`.

The issue I ran into and decided to avoid was verification that the vector had the right number of entries in it. It seemed like it would be too risky for the formatting code, which would have to read the vector at specific hardcoded locations, to get disconnected from the parsing code, which needs to populate it. It seemed safer to use new names in the AST. I feel the pain though. Our abstract syntax tree is a little less abstract.

The other issue was that using a vector made the testing code super hard to read and work with. BasedNodes were populated with a vectors with a seemingly random number of nulls and this had to line up with formatting and parsing.

PR #2603, opened 2020-03-09
fix(libflux): simplify libflux C API and resolve memory leaks

wolffcm:
This PR updates the Go/Rust API to address memory leaks found by valgrind.

It turns out that the Rust source code generated from `flux.h` by bindgen was mostly unused, so I removed it.

We were not actually freeing memory allocated from Rust in calls to `flux_free` from the Go side because freeing `void*` doesn't let the Rust compiler know that it should also free pointer fields embedded within structs.

To address this I created separate `flux_free*` functions for each unique type that must be freed.  I tried to use `Box<dyn Any>` to work around this, but `Any` is not sized, so it didn't work right. See the documentation for `std::boxed`, linked just below, for details.

I based these changes on material here:
https://doc.rust-lang.org/std/boxed/index.html#memory-layout

And here: https://doc.rust-lang.org/nomicon/ffi.html#foreign-function-interface

### DIFF
--- a/libflux/src/flux/lib.rs
+++ b/libflux/src/flux/lib.rs
@@ -16,17 +16,11 @@ pub mod semantic;
 use std::error;
 use std::ffi::*;
 use std::fmt;
-use std::os::raw::{c_char, c_void};
+use std::os::raw::c_char;
 
 use parser::Parser;
 
 pub use ast::DEFAULT_PACKAGE_NAME;
-
-#[allow(non_camel_case_types, missing_docs)]
-pub mod ctypes {
-    include!(concat!(env!("OUT_DIR"), "/ctypes.rs"));
-}
-use ctypes::*;
 
 /// An error handle designed to allow passing `Error` instances to library
 /// consumers across language boundaries.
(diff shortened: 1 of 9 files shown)


libflux/src/flux/lib.rs
```diff
@@ -91,118 +105,127 @@ pub struct flux_buffer_t {
...
+pub unsafe extern "C" fn flux_parse(cstr: *mut c_char) -> Box<ast::Package> {
+    let buf = CStr::from_ptr(cstr).to_bytes();
     let s = String::from_utf8(buf.to_vec()).unwrap();
     let mut p = Parser::new(&s);
     let pkg: ast::Package = p.parse_file(String::from("")).into();
-    Box::into_raw(Box::new(pkg)) as *mut flux_ast_pkg_t
+    Box::new(pkg)
 }
 
+/// Frees an AST pacakge.
```
rockstar: "package"

libflux/src/flux/lib.rs
```diff
@@ -91,118 +105,127 @@ pub struct flux_buffer_t {
...
-    let buffer = &mut *buf; // Unsafe
-    buffer.len = data.len();
-    buffer.data = Box::into_raw(data.into_boxed_slice()) as *mut u8;
-    std::ptr::null_mut()
+    (*buf).len = data.len();
+    (*buf).data = Box::into_raw(data.into_boxed_slice()) as *mut u8;
+    None
 }
 
+/// flux_ast_marshal_fb serializes the given AST package to a flatbuffer.
```
rockstar: Thanks for adding these.

PR #2610, opened 2020-03-10
test: add test-valgrind to CI

wolffcm:
This just adds a new job called `test-valgrind` to the CircleCI config.


wolffcm:
Note, I used dynamic linking for the binary run in valgrind, hence I needed to set `LD_LIBRARY_PATH`.  I only did this out of expediency.

PR #2611, opened 2020-03-10
[DO NOT MERGE] chore: checking to see if I introduce a memory leak that it fails CI

wolffcm:


nathanielc:
Is it possible to have a permanent failing test case on CI and just invert the result? I'd expect we would have to create a special test binary and API. Does that seem worth it? Or would it be too different from the actual test to provide value?

wolffcm:
I think there would be value in it.  It'd be pretty easy to do.  We could add a command line flag to the test binary that skips the `flux_free*` calls.  If valgrind doesn't report a leak, something is wrong.

I'm more worried that we add something to the C API and don't test it in valgrind, and as a result don't find issues.  I'm not sure how to enforce that, save for setting up a code coverage tool that works with C.

PR #2617, opened 2020-03-11
docs: create new guide and ideas for Flux changes

pauldix:
Ideas for changes to Flux to aid in usability and make common queries easier with less ceremony.

### DIFF
--- a/docs/new_flux_ideas_and_guide.md
+++ b/docs/new_flux_ideas_and_guide.md
@@ -360,8 +360,8 @@ However, The difference is that select returns specific rows, rather than a summ
 examples:
 
 ```
-@my_data.h2o_temperature{location, state, bottom_degrees, surface_degrees, _time > -3h}
-    |> select({min(bottom_degrees)}, by: [])
+@my_data.h2o_temperature{location, state, bottom_degrees, surface_degrees, time > -3h}
+    |> select(fn: min("bottom_degrees"), by: [])
 ```
 
 | location | state | bottom_degrees | surface_degrees | time
@@ -374,31 +374,12 @@ we've pulled in two records. We pull in two records because both of those have t
 value of `40`, which is the min for all records in that time. Notice also that `time` is the same value as those 
 individual records, which you don't see in calls to `aggregate`.
 
-If you pass in multiple columns, they are used as secondary selectors (kind of like sorting by multiple columns) 
-for example: 
-
-```
-@my_data.h2o_temperature{location, state, bottom_degrees, surface_degrees, _time > -3h}
-    |> select({min(bottom_degrees), max(surface_degrees)})
-```
-
-produces:
-| location | state | bottom_degrees | surface_degrees | time
-| --- | --- | --- | --- | --- |
-| puget_sound | WA | 40 | 55 | 2020-02-22T17:10 |
-
-Or calling `select` with `{min(bottom_degrees), min(surface_degrees)}` produces: 
-
-| location | state | bottom_degrees | surface_degrees | time
-| --- | --- | --- | --- | --- |
-| puget_sound | WA | 40 | 54 | 2020-02-22T17:40 |
-
 You can also select in groups based on time or by other columns, just like you can with `aggregate`. Here's 
 an example using time grouping and inserting the time window for associated row selections:
 
 ```
 @my_data.h2o_temperature{location, state, bottom_degrees, _time > -3h}
-    |> selector({max(bottom_degrees), by: ["state"], window: 1h, windowColumn: "window")
+    |> selector(fn: max(bottom_degrees), by: ["state"], window: 1h, windowColumn: "window")
 
 // if you include window_column, the result will put the window timestamp for each selected row in that column
 ```
@@ -425,8 +406,8 @@ which block of time the selector was working across.
 
 The following functions can be used in the **Selector Predicate Block**: `first`, `last`, `min`, `max`, `top`, `bottom`
 
-#### Calling Aggregates and Selectors without Aggregate or Select
-The `aggregate` and `select` functions provide a shorthand for computing based on multiple columns. However, the 
+#### Calling Aggregates without the Aggregate function
+The `aggregate` function provides a shorthand for computing multiple aggregates at a time. However, the 
 functions passed in can also be called on tables. For example:
 
 ```
@@ -440,13 +421,15 @@ functions passed in can also be called on tables. For example:
 @my_data.h2o_temperature{time > -3h} |> mean(by: ["location"], window: 1h)
 ``` 
 
+Selectors can only be called with the `select` function.
+
 #### Count, Distinct, and CountDistinct
 The `count` function is an aggregate while the `distinct` function is a selector. The `countDistinct` function is 
 a helper function that will combine operations for you.
  
 ```
 @my_data.h2o_temperature{state, time > -3h}
-    |> select({distinct(state)}, time: 1h, windowColumn: "time")
+    |> select(fn: distinct(state), window: 1h, windowColumn: "time")
 ```
 
 Would return the following result
@@ -465,9 +448,9 @@ per hour:
 
 ```
 @my_data.h2o_temperature{state, time > -3h}
-    |> select({distinct(state)}, time: 1h)
+    |> select(fn: distinct(state), window: 1h)
     |> timeShift(-1m)
-    |> aggregate({count(state)}, time: 1h)
+    |> count("state", window: 1h)
 ```
 
 Would produce:
@@ -611,7 +594,7 @@ the max for the given grouping and time window. If used in a selector, it will r
 max value for the grouping and window.
 
 ```
-max = (table, column) => {}
+max = (table, column, by = [], window = 0) => {}
 ``` 
 
 ### Min
@@ -620,21 +603,21 @@ the min for the given grouping and time window. If used in a selector, it will r
 min value for the grouping and window.
 
 ```
-min = (table, column) => {}
+min = (table, column, by = [], window = 0) => {}
 ``` 
 
 ### Mean
 Mean is an aggregate function (so it can only be used as an argument to `aggregate`).
 
 ```
-mean = (table, column) => {}
+mean = (table, column, by = [], window = 0) => {}
 ```
 
 ### Count
 Count is an aggregate function (so it can only be used as an argument to `aggregate`).
 
 ```
-count = (table, column) => {}
+count = (table, column, by = [], window = 0) => {}
 ```
 
 ### Top
@@ -643,7 +626,7 @@ The top function is passed into `select` to produce the top n rows per grouping 
 ```
 // notice that top takes a view, which is what the selector function sends to it. Top can only be 
 // used as an argument to selector. N is the number of results you'd like
-top = (table, column_name, n) => {}
+top = (table, column, n) => {}
 ```
 
 ### Sort
@@ -711,13 +694,9 @@ Filter will filter a result set to the requested columns and records matching th
 filter = (table, fn) => {}
 
 // calling it
-data |> filter({foo == "bar"}) // will return all columns in data, but only rows matching criteria
+data |> filter(fn: (r) => r.foo == "bar") // will return all columns in data, but only rows where the value of foo is bar
 
-data |> filter({foo == "bar" and time > -3h and time < -1h}) // will return only rows in specified time range
-
-data |> filter({foo, colA, foo == "bar"}) // will return only the foo, colA and system columns
-
-data |> filter(fn: {foo == "bar"}) // example calling with explicit parameter named
+data |> filter({fn: (r) => r.foo == "bar" and r.time > -3h and r.time < -1h) // will return only rows in specified time range
 ```
 
 ### Drop
--- a/docs/new_flux_ideas_and_guide.md
+++ b/docs/new_flux_ideas_and_guide.md
@@ -313,7 +313,7 @@ enough to unique identify a series.
 Say we wanted to get a summary per state: 
 
 ```
-@my_data.h2o_temperature{location, state, bottom_degrees, _time > -3h}
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
     |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["state"]) 
 ```
 
@@ -378,7 +378,7 @@ You can also select in groups based on time or by other columns, just like you c
 an example using time grouping and inserting the time window for associated row selections:
 
 ```
-@my_data.h2o_temperature{location, state, bottom_degrees, _time > -3h}
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
     |> selector(fn: max(bottom_degrees), by: ["state"], window: 1h, windowColumn: "window")
 
 // if you include window_column, the result will put the window timestamp for each selected row in that column
@@ -717,7 +717,7 @@ Keep will keep only the specified columns in the table, while keeping any system
 keep = (table, columns) => {}
 
 // calling it
-data |> keep(["foo", "_time"])
+data |> keep(["foo", "time"])
 ```
 
 ### Rename
@@ -948,7 +948,7 @@ Finally, let's look at how we can get the last value for each series in a measur
 `state == "CA"`,  but only if that value appears in the last five minutes:
 
 ```
-@my_data.h2o_temperature{location, state, surface_degrees, state == "CA" and _time > -5m}
+@my_data.h2o_temperature{location, state, surface_degrees, state == "CA" and time > -5m}
   |> select(fn: last, by: ["_key"]) 
 ```
 
@@ -966,7 +966,7 @@ this: `select(last)`.
 
 We expect the result set to look like this:
 
-| location | state| surface_degrees | _key | _time |
+| location | state| surface_degrees | _key | time |
 | --- | --- | --- | --- | --- |
 | coyote_creek | CA | 50.2 | location=coyote_creek,state=CA | 1600756160 |
 | santa_monica | CA | 63.6 | location=santa_monica,state=CA | 1600756160 |
@@ -980,6 +980,6 @@ from(bucket: "my_data", measurement: "h2o_temperature", columns: ["location", "s
     |> select(fn: last, by: ["_key"])
 ```
 
-The `range` function operates explicitly on the `_time` column to limit results.
+The `range` function operates explicitly on the `time` column to limit results.
 Generally it's preferred to use the shortcut syntax when possible, but showing what the full translation to Flux 
 functions has hopefully served as a gentle introduction to Flux functional concepts.
--- a/docs/new_flux_ideas_and_guide.md
+++ b/docs/new_flux_ideas_and_guide.md
@@ -1,0 +1,1006 @@
+# Getting Started Querying with Flux and InfluxDB
+**NOTE BEFORE YOU GET STARTED:** Forget much of what you know about Flux. This is a new version with some very 
+important differences. As you work through this guide, please keep notes on your thoughts for each section 
+as you encounter it. This ends in a test and you are expected to create a secret Gist with your thoughts and 
+the answers to the questions. You can close with a summary of your feelings after going through the whole thing 
+but please note your thoughts as you go along. It's very likely that you'll think something is crazy or off 
+initially, but maybe think it's ok when you look back later to write your summary. We're interested in 
+seeing this progression of your learning and thought process.
+
+You're allowed to ask questions in the private channel set up for this usability test in Slack. #flux-usability-yourname
+
+Welcome to the getting started guide for Flux. Flux is a functional scripting and query language that looks 
+very similar to Javascript in some parts, but with shortcut syntax for common queries. Its goal is to make 
+querying and working with data, and time series data in particular, easy and fast. While InfluxDB is the first 
+class data source for Flux, it also provides adaptors to other databases, third party APIs, files, and formats that 
+data can be represented in.
+
+This guide doesn't present the entirety of the API or its functionality, but a subset to convey the essential 
+concepts of the language. Go here for a [complete Flux API reference]().
+
+## Data Model
+The data model of InfluxDB consists of Buckets, which hold data. Time series data is written in 
+via the HTTP API and is represented as InfluxDB Line Protocol. It separates time series into measurements, tags, 
+fields, and a time. Here are some examples of the InfluxDB line protocol:
+
+```
+cpu,host=serverA,region=west usage_system=23.2,usage_user=55.1 1566000000
+events,app=foo description="something happened",state=1i 1566000000
+```
+
+In the first line you have a measurement named `cpu`, two tag key/value pairs of `host=serverA` and `region=west`, 
+two fields of `usage_system=23.2` and `usage_user=55.1` and a timestamp at the end of `1566000000`. Measurement 
+names and tag values are always strings. Fields can be either float64, int64, bool, or string. The second line in
+the snippet has data for the `events` measurement and shows a string and integer field value.
+
+The combination of a measurement, tags, and a specific field represent an individual time series, which are an
+ordered list of time, value pairs. For example, `cpu,host=serverA,region=west usage_system` would be one time 
+series from line 1 in the above example. You can generally think of tags as the dimensions on which you can
+query and slice and dice your time series data. However, you can also do that with fields, provided that you
+pay attention to performance and schema design considerations.
+
+With the data in InfluxDB it's important to keep this context of a time series in mind. There are different types 
+of numbers that can be recorded in time series and many operations only make sense when performed on an individual 
+series and then later combined in some other way. Here are some examples of data that is fed in a time series:
+
+### Metrics
+Metrics are regular time series that are values captured at specific intervals of time (like every 10 seconds or 1 
+minute or 1 hour or 1 day). There are different kinds of metrics that are commonly captured
+
+* counter - A value that counts the occurrence of an event. Counters alway increase unless they have 
+            been wrapped or reset. A reset occurs when what ever is doing the counting is restarted in 
+            which case it starts counting again from 0. A wrapped counter is one that has reached the 
+            largest number allowed by the byte representation of the value, in which case it starts over 
+            again at 0. These wraps and resets need to be accounted for when performing operations on counters. 
+            An example of a counter is the number of bytes a network interface has received. 
+* gauge - A value that measures a specific value, which can increase or decrease. An example of a gauge is 
+          taking a temperature reading.
+
+### Events
+Events are single data points in time. This could be an individual request to an API and how long that took, a user 
+logging into a service, a car parking, a light turning on, a container starting or stopping, or anything you can think 
+of. Events are what are called irregular time series given that they can occur at any time. It is possible to create 
+regular time series from an underlying event stream (like counting how many occurred every five minutes for the last hour).
+
+### Working with data in Flux
+When working with data in Flux, it translates the line protocol/time series structure into to a table. You 
+can think of a measurement name as a table with the tag and field keys representing columns. Flux also 
+includes the reserved column, `time`, in every table pulled from InfluxDB. The `time` column is the time at the end 
+of the line protocol when data is written in.
+
+Here's an example of line protocol and its translation to a Flux table:
+
+```
+h2o_temperature,location=santa_monica,state=CA surface_degrees=65.2,bottom_degrees=50.4 1568756160
+h2o_temperature,location=santa_monica,state=CA surface_degrees=63.6,bottom_degrees=49.2 1600756160
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=55.1,bottom_degrees=51.3 1568756160
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=50.2,bottom_degrees=50.9 1600756160
+h2o_temperature,location=puget_sound,state=WA surface_degrees=55.8,bottom_degrees=40.2 1568756160
+h2o_temperature,location=puget_sound,state=WA surface_degrees=54.7,bottom_degrees=40.1 1600756160
+```
+
+| location | state | surface_degrees | bottom_degrees | time |
+| --- | --- | --- | --- | --- |
+| coyote_creek | CA | 55.1 | 51.3 | 1568756160 |
+| coyote_creek | CA | 50.2 | 50.9 | 1600756160 |
+| puget_sound | WA | 55.8 | 40.2  | 1568756160 |
+| puget_sound | WA | 54.7 | 40.1  | 1600756160 |
+| santa_monica | CA | 65.2 | 50.4 | 1568756160 |
+| santa_monica | CA | 63.6 | 49.2 | 1600756160 |
+
+Note that the measurement name doesn't appear anywhere. That's because the table itself represents data from the 
+`h2o_temperature` measurement. Later we'll see how to work with multiple measurements (i.e. tables). Note that the 
+underlying time series would be represented by the combination of `location`, `state`, and the specific field. 
+Results are ordered by default by the columns of the tags and their values and then by time. So we can see that 
+`coyote_creek`, `CA` rows are together in time ascending order.
+
+We'll show later how to specify to the graphing engine what represents an individual time series that you want 
+visualized.
+
+## Basic Queries
+This section will explore some basic queries in Flux to highlight the syntax and a few of the basic concepts of
+the language. For these examples, we'll assume a dataset that has a schema like the one seen above.
+
+### Introduction to basics and getting the last value
+In this example, we'll get the last value of `surface_degrees` for every series in the `h2o_temperature` 
+measurement in the `my_data` bucket. Flux has a shorthand syntax for the common operation of selecting data from 
+a bucket with given criteria for a given time range. Let's look at that first for this example:
+
+```
+@my_data.h2o_temperature{surface_degrees, time}
+```
+
+The `@` symbol at the beginning indicates that we're using the shorthand syntax and that the next set of characters
+will identify the bucket name. After that we have a `.` separator, followed by the measurement name of 
+`h2o_temperature`. Immediately following that we have the **Column Filter Predicate** syntax (the curly braces
+`{` and `}` and what's inside of them) to indicate which columns we want in the result. The identifiers can
+also be wrapped in double quotes, which you'll need to use if they contain spaces:
+
+```
+@"my_data"."h2o_temperature"{"surface_degrees", "time"}
+```
+
+Given the previous data set, this query would produce a result like this: 
+
+| surface_degrees | time |
+| --- | --- |
+| 50.2 | 1600756160 |
+| 54.7 | 1600756160 |
+| 63.6 | 1600756160 |
+
+Since we didn't specify a time range in the query, we get the most recent data point for each series. How far 
+back in time the database looks for a recent value is dependent on the setup of the backend InfluxDB system you 
+are working with (this will typically be less than a few hours). We'll show how to specify explicitly how far 
+back to look for the most recent value in a moment.
+
+In this result set we have three rows, but we can't see their series information (i.e. the tags) because we 
+didn't select them. Selecting them is fairly easy:
+
+```
+@my_data.h2o_temperature{location, state, surface_degrees, time}
+```
+
+Would return the last value of surface_degrees in each series:
+
+| location | state | surface_degrees | time |
+| --- | --- | --- | --- |
+| coyote_creek | CA | 50.2 | 1600756160 |
+| puget_sound | WA | 54.7 | 1600756160 |
+| santa_monica | CA | 63.6 | 1600756160 |
+
+
+In Flux, a tag key and a field key are both called columns. This means that tag keys and field keys must
+be unique within a measurement. Here's an example that explicitly selects all the columns we know about:
+
+```
+@my_data.h2o_temperature{location, state, surface_degrees, bottom_degrees, time}
+``` 
+
+We can also select all columns by leaving the column list empty:
+
+```
+@my_data.h2o_temperature{}
+``` 
+
+Or you can leave out the curly braces:
+
+```
+@my_data.h2o_temperature
+```
+
+If your **Column Filter Predicate** selects only a tag, you'll get the same number of records as you have series for 
+that tag since the last value for each series is what gets defined if you're not filtering by time range or other 
+criteria.
+
+```
+@my_data.h2o_temperature{state}
+```
+
+Returns these results:
+
+| state |
+| --- |
+| CA |
+| CA |
+| WA |
+
+#### Filtering on Criteria
+You can limit the results returned in the query by specifying filtering criteria in the *Column Filter Predicate* block. 
+Here's an example that filters based on the location:
+
+```
+@my_data.h2o_temperature{location, surface_degrees, time, state == "CA"}
+```
+
+| location | state | surface_degrees | time |
+| --- | --- | --- | --- |
+| coyote_creek | CA | 50.2 | 1600756160 |
+| santa_monica | CA | 63.6 | 1600756160 |
+
+Note that `state` is returned in the result because it was listed in the filter criteria. You can have more complex 
+logic to match criteria like:
+
+```
+@my_data.h2o_temperature{location, surface_degrees, time, location == "coyote_creek" OR location == "puget_sound"}
+```
+
+See that while the column names are separated by commas, the filter criteria is separated by the different matching 
+clauses.
+
+
+#### Matching Critera
+There are many operators you can use to filter down data. Here is the list of operators:
+* `==` (equal to)
+* `!=` (not equal)
+* `=~` (regex match)
+* `!~` (not a regex match)
+* `>` (greater than)
+* `>=` (greater than or equal to)
+* `<` (less than)
+* `<=` (less than or equal to)
+* `in`
+
+The regex and in operators are probably the trickiest ones to work with, so here are a few examples:
+
+```
+// this is a comment because it has // before it
+
+// status is a tag and this will match any stats between 200 and 500
+@my_data.http_requests{status =~ /[2-5][0-9][0-9]/}
+
+// will match states in the array
+@my_data.h2o_temperature{state in ["CA", "WA"]}
+```
+
+### Aggregate, Select, Transform and Group Data
+Flux has many built in functions to work with time series data. **Aggregate** functions refer to functions that summarize 
+some period of time. They take N rows and output a single result, examples include count, mean, and percentile, 
+among many others. **Select** functions select a set of rows from an input set based on some criteria. Min, max, top, and 
+bottom are examples of selectors. Some selectors also double as aggregates like min and max.
+
+This is an important distinction. An aggregate produces new data from some source data. That new data is frequently a 
+summary statistic that describes the source data in some meaningful way. A selector specifies criteria for selecting 
+rows from some source data. So you're not transforming the data when using a selector. For example, `count` is not a 
+function that could be used as a selector.
+
+#### Sample Dataset
+Here's an example dataset that we'll use for the rest of the examples in this section. The raw number timestamps 
+have been replaced with human readable times and the values have been set to simple whole numbers to make things 
+easier. The example times only go down to minute precision to make things easier, but you can summarize down to 
+the nanosecond.
+
+```
+h2o_temperature,location=santa_monica,state=CA surface_degrees=65,bottom_degrees=50 2020-02-22T15:01
+h2o_temperature,location=santa_monica,state=CA surface_degrees=64,bottom_degrees=49 2020-02-22T15:31
+h2o_temperature,location=santa_monica,state=CA surface_degrees=63,bottom_degrees=49 2020-02-22T16:01
+h2o_temperature,location=santa_monica,state=CA surface_degrees=62,bottom_degrees=49 2020-02-22T16:31
+h2o_temperature,location=santa_monica,state=CA surface_degrees=61,bottom_degrees=48 2020-02-22T17:01
+h2o_temperature,location=santa_monica,state=CA surface_degrees=60,bottom_degrees=48 2020-02-22T17:31
+h2o_temperature,location=santa_monica,state=CA surface_degrees=60,bottom_degrees=48 2020-02-22T17:46
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=55,bottom_degrees=51 2020-02-22T16:05
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=53,bottom_degrees=50 2020-02-22T17:05
+h2o_temperature,location=puget_sound,state=WA surface_degrees=55,bottom_degrees=40 2020-02-22T17:10
+h2o_temperature,location=puget_sound,state=WA surface_degrees=54,bottom_degrees=40 2020-02-22T17:40
+```
+
+In the data set we can see there are three distinct keys and that the time range of data is from 
+`2020-02-22T15:01` to `2020-02-22T17:46` (about a two and a half hour span). There are more data 
+points for the `santa_monica`, `CA` location than for the others. The differing number of values 
+will be instructive later when we shape and summarize this data.
+
+#### Summarizing and Grouping with Aggregates
+Let's get into an example for summarizing some of our water data. For the sake of these queries, assume that 
+we're executing them at `2020-02-22T18:00`, that way we can use relative times in the query. Say we want to get the 
+last 3 hours of data, with the min, max, and mean of `bottom_degrees` for that time period. With that 
+definition, we'd expect to get 1 row per series in the dataset (assuming that we have min, max, and mean as 
+columns in the resulting table. Here's the query:
+
+```
+@my_data.h2o_temperature{bottom_degrees, time > -3h}
+    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}) 
+```
+
+Will return the following dataset:
+
+| min_bottom_degrees | max_bottom_degrees | mean_bottom_degrees |
+| --- | --- | --- |
+| 48 | 51 | 47.45 |
+
+Even though the sample data set has three series, we get back only a single row that gives the summary statistics we 
+requested. The call to the aggregate function passed in an **Aggregate Summary Block** which looks similar in 
+construction to the **Predicate Filter Block** that we used on the first line of the query. The returned column names 
+are given automatically based on the construction `<function name>_<field name>`. As with the other blocks, field 
+names can be wrapped in double quotes to handle cases like spaces or other special characters in a field name.
+
+Our results don't have the descriptive tag metadata because we didn't ask for it. Selecting `location` and `state` 
+which are the two tags that we have for data in the `h20_measurements` measurement will ensure it comes through in 
+the aggregate and we can then group by those columns:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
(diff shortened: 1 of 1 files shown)


docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
+    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["location", "state"]) 
+```
+
+| location | state | min_bottom_degrees | max_bottom_degrees | mean_bottom_degrees |
+| --- | --- | --- | --- | --- |
+| coyote_creek | CA | 48 | 50 | 48.714 |
+| puget_sound | CA | 40 | 40 | 40 |
```
jsteenb2: locations are flip flopped here for `puget_sound`  and `santa_monica`

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+fill in default values or values based on a summary statistic of the rows like `first`, `min`, `max`, `mean`, etc. You 
+can find the [details for interpolate in the function reference]().
+
+#### Select
+The `select` function works similar to `aggregate` in that it operates on windows of time and groupings of rows.
+However, The difference is that select returns specific rows, rather than a summary data point. Here are a few 
+examples:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, surface_degrees, time > -3h}
```
jsteenb2: I find this select misleading. This immediately reminds me of SQL but in reverse, and starts to play games with my limited brain capacity… Any way it could be renamed to something else that isn’t reminiscent of SQL?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+Would produce:
+
+| state | time
+| --- | --- |
+| 1 | 2020-02-22T16:00
+| 1 | 2020-02-22T17:00
+| 2 | 2020-02-22T18:00
+
+See that we had to use the `timeShift` function to move the timestamps of the windows back so that the resulting
```
jsteenb2: Also, not clear why I’d need `timeShift`? Feels magical to have to know you need to move times...

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+we've pulled in two records. We pull in two records because both of those have the same `min` `bottom_degrees` 
+value of `40`, which is the min for all records in that time. Notice also that `time` is the same value as those 
+individual records, which you don't see in calls to `aggregate`.
+
+You can also select in groups based on time or by other columns, just like you can with `aggregate`. Here's 
+an example using time grouping and inserting the time window for associated row selections:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
+    |> selector(fn: max(bottom_degrees), by: ["state"], window: 1h, windowColumn: "window")
```
jsteenb2: is this suppose to be `select`? or is `selector` something new?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+```
+
+The above example shows assigning a result set to a variable so that it can be used multiple times later. You 
+can also use this construction to do implicit joins across tables. Here's a made up example:
+
+```
+foo_data = @my_data.foo{value as foo_val, and time > -1h}
+bar_data = @my_data.bar{value as bar_val, and time > -1h}
+
+foo_data.foo_val - bar_data.bar_val as difference_to_bar
```
jsteenb2: this is one issue I have had with flux, its implicit return/yield syntax. This seems to make this even more complicated to me 🤔. 

Feels very magical/unobvious that this adds to the returned group imo 🤷‍♂

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+to make this work. However, there is one trick you'll likely need...
+
+### Using Data From a Previous Result in a new Query
+You can query the database for results and use either a record, an individual value, or an entire column in 
+subsequent queries. This is useful for pulling back data from top results, etc. Here's an example:
+
+```
+// get the top 10 locations in terms of surface degrees
+top_temps = @my_data.h2o_temperature{surface_degrees, location}
+    |> selector({top(surface_degrees, 10)})
```
jsteenb2: `s/selector/select`?

docs/new_flux_ideas_and_guide.md
nathanielc: Do you think this would the kind of thing you get over quickly?  Or do you have a suggestion for a better name? 

I share the concern but not sure what to do about it.

jsteenb2: hmm, I'm not worried about me per say, I'm worried about the person who is coming into flux as a newb.  I wonder if @sanderson could help us out here?

select feels so ingrained and has so much baggage from its use in SQL, my biggest concern here.

docs/new_flux_ideas_and_guide.md
nathanielc: The use case where `timeShift` comes up is wanting to compare data from different times. For example say I wanted to plot the average temperature for each 1h for the entire day. I would have a plot with 24 points. Now say I wanted to put yesterdays data on the same plot?

How would you like to describe that query in Flux?


This is how I would expect it to be done:

```
@data.temperature{surface_degrees, location, time > -24h}
    |> mean(by:["location"], window: 1h)

@data.temperature{surface_degrees, location, time > -48h and time < -24h}
    |> mean(by:["location"], window: 1h)
    |> timeShift(1h)
```

Now my plot has both yesterday's and today's data overlaid. 

Open to other paradigms for approaching this use case.

docs/new_flux_ideas_and_guide.md
nathanielc: typo, its supposed to be `select`.

docs/new_flux_ideas_and_guide.md
nathanielc: How would you like to see it done? Are you wanting an explicit `return` statement or something? 

I hadn't heard this feedback before and I quite like the implicit nature as it makes writing simple queries easier, but I want to see it from your perspective if I can. Any suggestions on how this could be more explicit?

docs/new_flux_ideas_and_guide.md
nathanielc: yep

docs/new_flux_ideas_and_guide.md
jsteenb2: hmm, is that `timeShift` suppose to be `timeShift(24h)`... this is interesting, although I don't know that I'd want to use it for this use case... UI concerns are a UI responsibility in my view...  looking at the transformed data, I think you'd have to use a yield to use this programmatically, does that sound about right?

docs/new_flux_ideas_and_guide.md
jsteenb2: I'm not sure off hand, have to noodle on it a bit. The current state of things I can wrap my head around nowadays.

However, this feels very strange to me:

```
foo_data.foo_val - bar_data.bar_val as difference_to_bar
```

in my head I'd expect to only get `difference_to_bar` here and not the result shared in the doc.

```
| foo_val | bar_val | difference_to_bar | time
| --- | --- | --- | --- | 
```

does that make sense why I find it confusing? 

This is one thing I liked about the map func, it has the `with` syntax, which makes things very obvious when reading it that you'll get everything and whatever is extending it.  Perhaps something like that could be used here? I don't know specifically what that may look like though 🤔

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+must be called if you will be visualizing with any kind of graph with an x and y axis. Nothing need be done for 
+tables and single stat panes. In the future, Flux will have functions instructing the front end to visualize and 
+structure the data in any of the ways it supports, like single stat panes, tables, bar charts, line graphs, heatmaps, 
+histograms and many more.
+
+Here's an example of the basic graph call:
+
+```
+@my_data.cpu{host, usage_system, usage_user, time, cpu == "cpu-total" AND time > -1h}
+    |> graph(yCols: ["usage_system", "usage_user"], xCol: "time", by: ["host"])
```
nathanielc: What if we make the UI collect this information from the user? Meaning Flux returns a table and the UI has to either figure out based on context or ask the user what the `x` and `y` columns are and if there is any grouping? This way the decision is still explicit but doesn't require that we design an graphing API inside of Flux.

docs/new_flux_ideas_and_guide.md
nathanielc: @jsteenb2 edited the post, yeah it was supposed to be `timeShift(24h)`.

The use case is not unique to wanting to visualize data. For example I want to create an alert where today's average is significantly different than yesterday's average at the for the same  time of day.

```
today = @data.temperature{surface_degrees, location, time > -24h}
    |> mean(by:["location"], window: 1h)

yesterday = @data.temperature{surface_degrees, location, time > -48h and time < -24h}
    |> mean(by:["location"], window: 1h)
    |> timeShift(24h)

today.surface_degrees - yesterday.surface_degrees
   |> monitor.check(...)
```

Yeah I am ignoring how data is yielded in these examples as its not really related to the use case of `timeShift`

jsteenb2: right on, this makes a lot more sense. I'm still not a fan of the syntax of measurement minus a measurement. Feels awkward, but not a hill for me to die on, just a friendly concern 😀

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+In the first line you have a measurement named `cpu`, two tag key/value pairs of `host=serverA` and `region=west`, 
+two fields of `usage_system=23.2` and `usage_user=55.1` and a timestamp at the end of `1566000000`. Measurement 
+names and tag values are always strings. Fields can be either float64, int64, bool, or string. The second line in
+the snippet has data for the `events` measurement and shows a string and integer field value.
+
+The combination of a measurement, tags, and a specific field represent an individual time series, which are an
+ordered list of time, value pairs. For example, `cpu,host=serverA,region=west usage_system` would be one time 
+series from line 1 in the above example. You can generally think of tags as the dimensions on which you can
+query and slice and dice your time series data. However, you can also do that with fields, provided that you
```
russorat: IMO, this "However..." sentence is not needed here.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+pay attention to performance and schema design considerations.
+
+With the data in InfluxDB it's important to keep this context of a time series in mind. There are different types 
+of numbers that can be recorded in time series and many operations only make sense when performed on an individual 
+series and then later combined in some other way. Here are some examples of data that is fed in a time series:
+
+### Metrics
```
russorat: It's not clear to me why we need to describe types of metrics and events here before we start with Flux.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+          taking a temperature reading.
+
+### Events
+Events are single data points in time. This could be an individual request to an API and how long that took, a user 
+logging into a service, a car parking, a light turning on, a container starting or stopping, or anything you can think 
+of. Events are what are called irregular time series given that they can occur at any time. It is possible to create 
+regular time series from an underlying event stream (like counting how many occurred every five minutes for the last hour).
+
+### Working with data in Flux
+When working with data in Flux, it translates the line protocol/time series structure into to a table. You
```
russorat: "each individual time series is translated to a table". I don't think adding "line protocol" here is helpful. Once the data is in InfluxDB, it's no longer line protocol.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+of. Events are what are called irregular time series given that they can occur at any time. It is possible to create 
+regular time series from an underlying event stream (like counting how many occurred every five minutes for the last hour).
+
+### Working with data in Flux
+When working with data in Flux, it translates the line protocol/time series structure into to a table. You 
+can think of a measurement name as a table with the tag and field keys representing columns. Flux also 
+includes the reserved column, `time`, in every table pulled from InfluxDB. The `time` column is the time at the end 
+of the line protocol when data is written in.
+
+Here's an example of line protocol and its translation to a Flux table:
```
russorat: In general, is it important to understand how line protocol translates to flux queries? If I'm a user, I already understand the data format for my software/sensor, and I am trying to learn how it is returned by Flux. The conversion to LP seems less important in my opinion (it's usually handled behind the scenes by the client lib or telegraf anyway). "Here's an example of six individual time series' returned by a flux query:"

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| location | state | surface_degrees | bottom_degrees | time |
+| --- | --- | --- | --- | --- |
+| coyote_creek | CA | 55.1 | 51.3 | 1568756160 |
+| coyote_creek | CA | 50.2 | 50.9 | 1600756160 |
+| puget_sound | WA | 55.8 | 40.2  | 1568756160 |
+| puget_sound | WA | 54.7 | 40.1  | 1600756160 |
+| santa_monica | CA | 65.2 | 50.4 | 1568756160 |
+| santa_monica | CA | 63.6 | 49.2 | 1600756160 |
+
+Note that the measurement name doesn't appear anywhere. That's because the table itself represents data from the
```
russorat: In my opinion, the first two sentences of this paragraph are the only ones needed here. At this point, i'm not thinking about multiple measurements or sorting, bringing that up here just takes away from the focus.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| santa_monica | CA | 65.2 | 50.4 | 1568756160 |
+| santa_monica | CA | 63.6 | 49.2 | 1600756160 |
+
+Note that the measurement name doesn't appear anywhere. That's because the table itself represents data from the 
+`h2o_temperature` measurement. Later we'll see how to work with multiple measurements (i.e. tables). Note that the 
+underlying time series would be represented by the combination of `location`, `state`, and the specific field. 
+Results are ordered by default by the columns of the tags and their values and then by time. So we can see that 
+`coyote_creek`, `CA` rows are together in time ascending order.
+
+We'll show later how to specify to the graphing engine what represents an individual time series that you want
```
russorat: what is the graphing engine?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+Note that the measurement name doesn't appear anywhere. That's because the table itself represents data from the 
+`h2o_temperature` measurement. Later we'll see how to work with multiple measurements (i.e. tables). Note that the 
+underlying time series would be represented by the combination of `location`, `state`, and the specific field. 
+Results are ordered by default by the columns of the tags and their values and then by time. So we can see that 
+`coyote_creek`, `CA` rows are together in time ascending order.
+
+We'll show later how to specify to the graphing engine what represents an individual time series that you want 
+visualized.
+
+## Basic Queries
```
russorat: I feel like we should get here sooner in the document. Maybe the "series to table" section from above can come later?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+This section will explore some basic queries in Flux to highlight the syntax and a few of the basic concepts of
+the language. For these examples, we'll assume a dataset that has a schema like the one seen above.
+
+### Introduction to basics and getting the last value
+In this example, we'll get the last value of `surface_degrees` for every series in the `h2o_temperature` 
+measurement in the `my_data` bucket. Flux has a shorthand syntax for the common operation of selecting data from
```
russorat: Is it important to call this "shorthand" syntax?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+```
+@my_data.h2o_temperature{surface_degrees, time}
+```
+
+The `@` symbol at the beginning indicates that we're using the shorthand syntax and that the next set of characters
+will identify the bucket name. After that we have a `.` separator, followed by the measurement name of 
+`h2o_temperature`. Immediately following that we have the **Column Filter Predicate** syntax (the curly braces
+`{` and `}` and what's inside of them) to indicate which columns we want in the result. The identifiers can
+also be wrapped in double quotes, which you'll need to use if they contain spaces:
```
russorat: I don't think explaining the double quotes is needed here.

docs/new_flux_ideas_and_guide.md
jsteenb2: I was skeptical of this graphing stuff getting mixed in here. It would be nice to have this direct and focused on making flux easy to use. Would be nice if this focused on  I think adding graphing concerns here takes focus away from that. Just my $0.02 🤷‍♂

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+```
+@"my_data"."h2o_temperature"{"surface_degrees", "time"}
+```
+
+Given the previous data set, this query would produce a result like this: 
+
+| surface_degrees | time |
```
russorat: This result confuses me. I think it would be useful to explain before showing this that if you just add the column names to the filter predicate, it just returns that last value. It would make this less confusing IMO.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| --- | --- |
+| 50.2 | 1600756160 |
+| 54.7 | 1600756160 |
+| 63.6 | 1600756160 |
+
+Since we didn't specify a time range in the query, we get the most recent data point for each series. How far 
+back in time the database looks for a recent value is dependent on the setup of the backend InfluxDB system you
```
russorat: Do we need to think about this yet? IMO, this distracts from the goal of learning Flux.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| 50.2 | 1600756160 |
+| 54.7 | 1600756160 |
+| 63.6 | 1600756160 |
+
+Since we didn't specify a time range in the query, we get the most recent data point for each series. How far 
+back in time the database looks for a recent value is dependent on the setup of the backend InfluxDB system you 
+are working with (this will typically be less than a few hours). We'll show how to specify explicitly how far 
+back to look for the most recent value in a moment.
+
+In this result set we have three rows, but we can't see their series information (i.e. the tags) because we
```
russorat: IMO, we should swap "series information" and "the tags". "but we can't see their tags (i.e. series information) because we..."

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+Would return the last value of surface_degrees in each series:
+
+| location | state | surface_degrees | time |
+| --- | --- | --- | --- |
+| coyote_creek | CA | 50.2 | 1600756160 |
+| puget_sound | WA | 54.7 | 1600756160 |
+| santa_monica | CA | 63.6 | 1600756160 |
+
+
+In Flux, a tag key and a field key are both called columns. This means that tag keys and field keys must
```
russorat: I'm not sure how helpful this detail is for users getting started IMO. I wasn't thinking about unique names here.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+be unique within a measurement. Here's an example that explicitly selects all the columns we know about:
+
+```
+@my_data.h2o_temperature{location, state, surface_degrees, bottom_degrees, time}
```
russorat: I'm not sure this example is needed here. I think the important thing is the "leave the column list empty" bit.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+``` 
+
+We can also select all columns by leaving the column list empty:
+
+```
+@my_data.h2o_temperature{}
+``` 
+
+Or you can leave out the curly braces:
```
russorat: Is this available anywhere there are empty {} in the language? or just here?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+```
+@my_data.h2o_temperature
+```
+
+If your **Column Filter Predicate** selects only a tag, you'll get the same number of records as you have series for
```
russorat: "series for that tag" doesn't make sense to me here for some reason. "the same number of records as you have series since the last value" makes more sense to me.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| santa_monica | CA | 63.6 | 1600756160 |
+
+Note that `state` is returned in the result because it was listed in the filter criteria. You can have more complex 
+logic to match criteria like:
+
+```
+@my_data.h2o_temperature{location, surface_degrees, time, location == "coyote_creek" OR location == "puget_sound"}
+```
+
+See that while the column names are separated by commas, the filter criteria is separated by the different matching
```
russorat: I don't understand this sentence. Based on the example above, i would try `@my_data.h2o_temperature{location == "coyote_creek" OR location == "puget_sound", surface_degrees, time}` here.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+There are many operators you can use to filter down data. Here is the list of operators:
+* `==` (equal to)
+* `!=` (not equal)
+* `=~` (regex match)
+* `!~` (not a regex match)
+* `>` (greater than)
+* `>=` (greater than or equal to)
+* `<` (less than)
+* `<=` (less than or equal to)
+* `in`
```
russorat: what about "not in"?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+// status is a tag and this will match any stats between 200 and 500
+@my_data.http_requests{status =~ /[2-5][0-9][0-9]/}
+
+// will match states in the array
+@my_data.h2o_temperature{state in ["CA", "WA"]}
+```
+
+### Aggregate, Select, Transform and Group Data
+Flux has many built in functions to work with time series data. **Aggregate** functions refer to functions that summarize
```
russorat: I wonder if we can take a page from Telegraf here: https://github.com/influxdata/telegraf/blob/master/docs/AGGREGATORS_AND_PROCESSORS.md#aggregator

Aggregate function "emit new metrics for some period of time"

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+@my_data.http_requests{status =~ /[2-5][0-9][0-9]/}
+
+// will match states in the array
+@my_data.h2o_temperature{state in ["CA", "WA"]}
+```
+
+### Aggregate, Select, Transform and Group Data
+Flux has many built in functions to work with time series data. **Aggregate** functions refer to functions that summarize 
+some period of time. They take N rows and output a single result, examples include count, mean, and percentile, 
+among many others. **Select** functions select a set of rows from an input set based on some criteria. Min, max, top, and
```
russorat: Is Select just another name for Filter?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+bottom are examples of selectors. Some selectors also double as aggregates like min and max.
+
+This is an important distinction. An aggregate produces new data from some source data. That new data is frequently a 
+summary statistic that describes the source data in some meaningful way. A selector specifies criteria for selecting 
+rows from some source data. So you're not transforming the data when using a selector. For example, `count` is not a 
+function that could be used as a selector.
+
+#### Sample Dataset
+Here's an example dataset that we'll use for the rest of the examples in this section. The raw number timestamps 
+have been replaced with human readable times and the values have been set to simple whole numbers to make things
```
russorat: IMO, Just use single digit timestamps like 1, 2, 3, 4 here. They would be easier to understand than the "human readable times".

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+This is an important distinction. An aggregate produces new data from some source data. That new data is frequently a 
+summary statistic that describes the source data in some meaningful way. A selector specifies criteria for selecting 
+rows from some source data. So you're not transforming the data when using a selector. For example, `count` is not a 
+function that could be used as a selector.
+
+#### Sample Dataset
+Here's an example dataset that we'll use for the rest of the examples in this section. The raw number timestamps 
+have been replaced with human readable times and the values have been set to simple whole numbers to make things 
+easier. The example times only go down to minute precision to make things easier, but you can summarize down to 
+the nanosecond.
```
russorat: I don't think this needs to be explained here IMO.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+#### Summarizing and Grouping with Aggregates
+Let's get into an example for summarizing some of our water data. For the sake of these queries, assume that 
+we're executing them at `2020-02-22T18:00`, that way we can use relative times in the query. Say we want to get the 
+last 3 hours of data, with the min, max, and mean of `bottom_degrees` for that time period. With that 
+definition, we'd expect to get 1 row per series in the dataset (assuming that we have min, max, and mean as 
+columns in the resulting table. Here's the query:
+
+```
+@my_data.h2o_temperature{bottom_degrees, time > -3h}
```
russorat: what happens if i don't add `bottom_degrees` here? My guess is that it would return all columns and bottom_degress, min/max/mean_bottom_degrees would all the same value?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+| min_bottom_degrees | max_bottom_degrees | mean_bottom_degrees |
+| --- | --- | --- |
+| 48 | 51 | 47.45 |
+
+Even though the sample data set has three series, we get back only a single row that gives the summary statistics we 
+requested. The call to the aggregate function passed in an **Aggregate Summary Block** which looks similar in 
+construction to the **Predicate Filter Block** that we used on the first line of the query. The returned column names 
+are given automatically based on the construction `<function name>_<field name>`. As with the other blocks, field 
+names can be wrapped in double quotes to handle cases like spaces or other special characters in a field name.
```
russorat: So if i had "Some Field", the resulting column would be "min_Some Field"? Could we be smart enough to match the delimiters found in the column name with _ being default? "Some Field" -> "min Some Field" Some_Field -> "min_Some_Field" and "Some.Field" -> "Min.Some.Field".

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+Our results don't have the descriptive tag metadata because we didn't ask for it. Selecting `location` and `state` 
+which are the two tags that we have for data in the `h20_measurements` measurement will ensure it comes through in 
+the aggregate and we can then group by those columns:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
+    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["location", "state"])
```
russorat: This query has a ton of duplication in it IMO. This is the shorthand syntax, but this is very long an error prone IMO. What happens if I have state on the first line, but forget it in the second? It feels like we should be able to guess what the "by:" fields are based on the first line of the query. Also, I wish I could add something like "bottom_degrees as bd" in the first line and use that in the second. If there is only one field in the first line, we could also assume what is sent to min/max/mean. This could be 
`@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h} |> aggregate({min, max, mean})`

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
+    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["state"]) 
+```
+
+| state | min_bottom_degrees | max_bottom_degrees | mean_bottom_degrees |
+| --- | --- | --- | --- |
+| CA | 48 | 51 | 49.11 |
+| WA | 40 | 40 | 40 |
+
+See that the results don't have the `location` column even though we selected it on the first line. That's because
```
russorat: This is really confusing IMO. To select a column and not have it returned. I think I'd rather the query error tbh.

docs/new_flux_ideas_and_guide.md
russorat: Also, since this is shorthand, `agg` should work for `aggregate` as well.

docs/new_flux_ideas_and_guide.md
russorat: Would this query give the same result?
```
@my_data.h2o_temperature{time > -3h}
    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["state"]) 
```

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+See that the results don't have the `location` column even though we selected it on the first line. That's because 
+it wasn't included in the `by` set of columns in the call to `aggregate`.
+
+The aggregate function can window the data by time intervals. For example, if we wanted to summarize each hour for 
+the last three hours:
+
+```
+@my_data.h2o_temperature{location, bottom_degrees, time > -3h}
+    |> aggregate({min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)}, by: ["location"], window: 1h)
```
russorat: I think it would feel more natural to me to have the window in the first line here. Since i'm thinking about time in the first line with `time > -3h`, it seems like I would be thinking about the window here too. I'm chunking the data into 1h chunks, then applying an aggregate to each of them.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+for nine total rows. But looking at the results, we only see the full three summaries for the `santa_monica` location. 
+This is because the other two series don't have data in every window of time.
+
+You can create rows for the missing summaries by calling `interpolate` after the call to aggregate. It can be used to 
+fill in default values or values based on a summary statistic of the rows like `first`, `min`, `max`, `mean`, etc. You 
+can find the [details for interpolate in the function reference]().
+
+#### Select
+The `select` function works similar to `aggregate` in that it operates on windows of time and groupings of rows.
+However, The difference is that select returns specific rows, rather than a summary data point. Here are a few
```
russorat: I also think the idea of "aggregate generates new summary metrics, and select samples from the underlying data" would help here.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+can find the [details for interpolate in the function reference]().
+
+#### Select
+The `select` function works similar to `aggregate` in that it operates on windows of time and groupings of rows.
+However, The difference is that select returns specific rows, rather than a summary data point. Here are a few 
+examples:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, surface_degrees, time > -3h}
+    |> select(fn: min("bottom_degrees"), by: [])
```
russorat: Why do i need to use `fn:` here but not in the `aggregate` function? Can't I select min/max in the same query?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+individual records, which you don't see in calls to `aggregate`.
+
+You can also select in groups based on time or by other columns, just like you can with `aggregate`. Here's 
+an example using time grouping and inserting the time window for associated row selections:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees, time > -3h}
+    |> selector(fn: max(bottom_degrees), by: ["state"], window: 1h, windowColumn: "window")
+
+// if you include window_column, the result will put the window timestamp for each selected row in that column
```
russorat: windowColumn

docs/new_flux_ideas_and_guide.md
russorat: should this be `select`?

russorat: If location isn't in my `by:` here, why is it returned? I thought fields that weren't in the `by:` got dropped?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| santa_monica | CA | 50 | 2020-02-22T15:01 | 2020-02-22T16:00 |
+| coyote_creek | CA | 51 | 2020-02-22T16:05 | 2020-02-22T17:00 |
+| santa_monica| CA | 48 | 2020-02-22T17:01 | 2020-02-22T18:00 |
+| santa_monica| CA | 48 | 2020-02-22T17:31 | 2020-02-22T18:00 |
+| santa_monica| CA | 48 | 2020-02-22T17:46 | 2020-02-22T18:00 |
+| puget_sound | WA | 40 | 2020-02-22T17:10 | 2020-02-22T18:00 |
+| puget_sound | WA | 40 | 2020-02-22T17:40 | 2020-02-22T18:00 |
+
+What this does is for each 1 hour period, it groups the rows by state and then selects the rows with the max value. 
+Because we included the `windowColumn` argument, that is included in the result with the ending timestamp of the
```
russorat: Is "ending timestamp" the right value to put in the window column? I'm not sure yet tbh.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+What this does is for each 1 hour period, it groups the rows by state and then selects the rows with the max value. 
+Because we included the `windowColumn` argument, that is included in the result with the ending timestamp of the 
+window that each record fell into.
+
+Since so many of the records have tied `max(bottom_degrees)` values, we see that many of the original 
+rows were selected in the result. We can also see the `window` that is inserted to specify 
+which block of time the selector was working across.
+
+The following functions can be used in the **Selector Predicate Block**: `first`, `last`, `min`, `max`, `top`, `bottom`
```
russorat: Does this need to be explicitly called out here? It isn't called out for aggregate

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+#### Calling Aggregates without the Aggregate function
+The `aggregate` function provides a shorthand for computing multiple aggregates at a time. However, the 
+functions passed in can also be called on tables. For example:
+
+```
+// get the most recently written record
+@my_data.h2o_temperature |> last()
+
+// get the most recently written record for each location
+@my_data.h2o_temperature |> last(by: ["location"])
```
russorat: Shouldn't you be able to do this without the last function since thats the default behavior? `@my_data.h2o_temperature{location, <some way to indicate all the fields>}

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+// get the average for each hour for each location
+@my_data.h2o_temperature{time > -3h} |> mean(by: ["location"], window: 1h)
```
russorat: i think about grouping my data into 1h windows, then applying a mean to it.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+``` 
+
+Selectors can only be called with the `select` function.
+
+#### Count, Distinct, and CountDistinct
+The `count` function is an aggregate while the `distinct` function is a selector. The `countDistinct` function is 
+a helper function that will combine operations for you.
+ 
+```
+@my_data.h2o_temperature{state, time > -3h}
```
russorat: I'm lost about how this query relates to the sentence above it. It doesn't seem to count at all.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| state | time
+| --- | --- |
+| CA | 2020-02-22T16:00
+| CA | 2020-02-22T17:00
+| CA | 2020-02-22T18:00
+| WA | 2020-02-22T18:00
+
+We can see that CA reported in all three hours while both CA and WA reported in only the last hour. Also see that 
+distinct discards columns that aren't in the `by` argument or the `distinct` column itself. So we use the `time` 
+name for where we put the window markers for each distinct state. You could then count the number of states reporting
```
russorat: It's not clear to me how the first two sentences explain why I would use `time` for the `windowColumn`.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+
+We can see that CA reported in all three hours while both CA and WA reported in only the last hour. Also see that 
+distinct discards columns that aren't in the `by` argument or the `distinct` column itself. So we use the `time` 
+name for where we put the window markers for each distinct state. You could then count the number of states reporting 
+per hour:
+
+```
+@my_data.h2o_temperature{state, time > -3h}
+    |> select(fn: distinct(state), window: 1h)
+    |> timeShift(-1m)
```
russorat: what is timeShift doing here?

docs/new_flux_ideas_and_guide.md
russorat: I agree. this doesn't make sense to me.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| state | time
+| --- | --- |
+| 1 | 2020-02-22T16:00
+| 1 | 2020-02-22T17:00
+| 2 | 2020-02-22T18:00
+
+See that we had to use the `timeShift` function to move the timestamps of the windows back so that the resulting 
+times would be accurate for `aggregate`.
+
+All of this is encapsulated by the `countDistinct` function. For example, if we wanted to count the number of
```
russorat: Just use this function IMO. Don't make me learn the hard way first.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+Flux makes it easy to do basic math and transformations across tables of data. First, let's look at 
+renaming columns. All of the different predicate blocks recognize the `as` keyword to rename. For 
+example:
+
+```
+@my_data.h2o_temperature{location, state, bottom_degrees as degrees, top_degrees, time}
+
+// or renaming columns in an aggregate:
+@my_data.h2o_temperature{bottom_degrees, time > -1h}
+    |> aggregate({min(bottom_degrees) as min, max(bottom_degrees) as max})
```
russorat: ah so i can do `@my_data.h2o_temperature{bottom_degrees as bd, time > -1h} |> aggregate({min(bd), max(bd)})` yay!

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+```
+temps = @my_data.h2o_temperature{bottom_degrees, surface_degrees, time, location == "coyote_creek"}
+
+// do some math and return a new table with an addition column. Note that as is required on the end. Resulting
+// table will have all the columns of the previous on and this new one
+new_table = temps.surface_degrees - temps.bottom_degrees as bottom_surface_difference
+
+// or you can pipe forward it to another function
+temps.surface_degrees - temps.bottom_degrees as bottom_surface_difference
+    |> sort(by: ["bottom_surface_differrence"])
```
russorat: difference

russorat: You haven't explained what a pipe forward is yet.

docs/new_flux_ideas_and_guide.md
russorat: I feel like I want to put parentheses around this like: `(foo_data.foo_val - bar_data.bar_val) as difference_to_bar`

docs/new_flux_ideas_and_guide.md
barbaranelson: LP is only behind the scenes if it all works perfectly. As soon as you have to start debugging your pipeline, you end up looking at the data in its various transformations (from the sensor format, to LP, to how it is stored in the table, to how you can query it). By hiding the transformation, or expecting the user to just know that it happens, makes it harder to understand what's going on.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+| WA | 40 | 40 | 40 |
+
+See that the results don't have the `location` column even though we selected it on the first line. That's because 
+it wasn't included in the `by` set of columns in the call to `aggregate`.
+
+The aggregate function can window the data by time intervals. For example, if we wanted to summarize each hour for 
+the last three hours:
+
+```
+@my_data.h2o_temperature{location, bottom_degrees, time > -3h}
```
rbetts: Why are some expressions allowed in these predicates but not others. For example -
Why isn’t `b.m{min(temp)}` valid?

Having a hard time understand what’s an expression and what isn’t. Compounded by `b.m` returning `last`. What’s different between min and last?

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+Remember there were three series, so for 1h summaries for the last 3 hours we'd expect to have three rows per group 
+for nine total rows. But looking at the results, we only see the full three summaries for the `santa_monica` location. 
+This is because the other two series don't have data in every window of time.
+
+You can create rows for the missing summaries by calling `interpolate` after the call to aggregate. It can be used to 
+fill in default values or values based on a summary statistic of the rows like `first`, `min`, `max`, `mean`, etc. You 
+can find the [details for interpolate in the function reference]().
+
+#### Select
+The `select` function works similar to `aggregate` in that it operates on windows of time and groupings of rows.
```
rbetts: Oh dear god why overload select?

docs/new_flux_ideas_and_guide.md
rbetts: There’s no windowing here? So again I wonder why some expressions are blessed for a predicate filter and not others.

docs/new_flux_ideas_and_guide.md
rbetts: I really dislike that without a pipe forward I get an implicit last. But with a pipe I get all.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+```
+
+Here are some examples of doing math and introducing the concept of variables:
+
+```
+temps = @my_data.h2o_temperature{bottom_degrees, surface_degrees, time, location == "coyote_creek"}
+
+// do some math and return a new table with an addition column. Note that as is required on the end. Resulting
+// table will have all the columns of the previous on and this new one
+new_table = temps.surface_degrees - temps.bottom_degrees as bottom_surface_difference
```
rbetts: Another place where the blessed
filter expressions are confusing. Why isn’t `b.m{a, b, a - b}` sufficient?

docs/new_flux_ideas_and_guide.md
rbetts: Can I write this without the intermediate variables?

rbetts: I’m super confused about what the join predicate is here.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+```
+h2o_temperature,location=santa_monica,state=CA surface_degrees=65.2,bottom_degrees=50.4 1568756160
+h2o_temperature,location=santa_monica,state=CA surface_degrees=63.6,bottom_degrees=49.2 1600756160
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=55.1,bottom_degrees=51.3 1568756160
+h2o_temperature,location=coyote_creek,state=CA surface_degrees=50.2,bottom_degrees=50.9 1600756160
+h2o_temperature,location=puget_sound,state=WA surface_degrees=55.8,bottom_degrees=40.2 1568756160
+h2o_temperature,location=puget_sound,state=WA surface_degrees=54.7,bottom_degrees=40.1 1600756160
+```
+
+| location | state | surface_degrees | bottom_degrees | time |
```
rbetts: I'd prefer an additional example that showed a table for a measurement that includes series with tag-sets that don't share all the same tag-keys. 

```
m,tag1=key1 field1=1.0
m,tag2=key2 field2=2.0
```

docs/new_flux_ideas_and_guide.md
rbetts: Talking about line protocol specializes flux to influxdb whereas applying it to tabular data makes it more general.

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+```
+@my_data.h2o_temperature
+```
+
+If your **Column Filter Predicate** selects only a tag, you'll get the same number of records as you have series for 
+that tag since the last value for each series is what gets defined if you're not filtering by time range or other 
+criteria.
+
+```
+@my_data.h2o_temperature{state}
```
rbetts: What happens if selecting a field this way: `@b.m{field1}`. My guess is it would return the last value for all unique series with a `field1` field name? 

/aside:  Since the user can't tell what's a tag and what's a field at query time - and we want flux to behave against third party tabular data - I think it would be confusing to make behavior differ by tag/field designations in queries like this.

docs/new_flux_ideas_and_guide.md
rbetts: Me too.

docs/new_flux_ideas_and_guide.md
pauldix: `min` requires a grouping, which isn't included in this part of the predicate. To do that you need to either call the `min` function after, or `selector` or `aggregate` depending on what you're going for.

docs/new_flux_ideas_and_guide.md
pauldix: Hah, I had it as `selector` before because I didn't want people to have the SQL association to `select`, but people tripped up on that. I could call it `selector`, but then `aggregate` should be named to `aggregator`

docs/new_flux_ideas_and_guide.md
rbetts: Is `time > -3h` equivalent to `time <= 3h`?

docs/new_flux_ideas_and_guide.md
pauldix: There is a windowing, by default it's everything in the table unless you specify a value for the `window` parameter. I guess we could update this so you can call `min` inside of the `{}`, but then that would only be for all the data in the range, without any kind of windowing.

docs/new_flux_ideas_and_guide.md
pauldix: No, that's not the case at all. Without the `time > -1h` part in the filter predicate you get `last`. Otherwise you get values in that range.

docs/new_flux_ideas_and_guide.md
pauldix: Sure, maybe the filter expressions should be able to do more

docs/new_flux_ideas_and_guide.md
rbetts: I think `as` should just be supported here anyway and the user can be explicit in what they want?

docs/new_flux_ideas_and_guide.md
pauldix: Yes, I just put them in there to make it a bit more readable.

docs/new_flux_ideas_and_guide.md
rbetts: I'd try to write this first as:

`@my_data.h2o_temperature{location, state, min(bottom_degrees), max(bottom_degrees), mean(bottom_degrees)} |> group(by: ["location", "state"]`

docs/new_flux_ideas_and_guide.md
pauldix: The goal was to give a shortcut syntax to implicit joins where you're joining on all tags and time. Since Flux doesn't really know about tags, it would actually be all String columns and time. I'm not sure the usefulness of this, we could just force people to use the `join` function, but I thought this might be a common operation.

docs/new_flux_ideas_and_guide.md
rbetts: This would be an error in SQL .. something like "not in aggregate function or group by clause"

docs/new_flux_ideas_and_guide.md
pauldix: No, it would return the value for the series with the min key and the greatest timestamp. So you'd get a single value.

docs/new_flux_ideas_and_guide.md
pauldix: No, `-3h` evaluates to `now() - 3h` where `3h` evaluates to `now() + 3h`. Places that expect a time like `time > ...` can take either a time or duration, which will be evaluated like that. Kind of like a macro I guess.

docs/new_flux_ideas_and_guide.md
pauldix: Hah, you could try but it wouldn't be right. Ideally the parser would return a helpful error that leads you to calling `min` with `select` or `aggregate`. You should view each function as taking an input and doing something with it.

docs/new_flux_ideas_and_guide.md
fchikwekwe: If this is a special type of syntax, then yeah maybe its helpful to name it for discussion.

docs/new_flux_ideas_and_guide.md
fchikwekwe: Are there a lot of use cases where the user would want `last` by default? I'm thinking maybe in a task, but its hard for me to think of other times where this would be the preferred default behavior.

docs/new_flux_ideas_and_guide.md
fchikwekwe: I agree with @jsteenb2 here that I would expect to only get `difference_to_bar` back here.

docs/new_flux_ideas_and_guide.md
Anaisdg: Maybe add something like: Columns are created for tag keys, field keys, and time. Note that the measurement name...

docs/new_flux_ideas_and_guide.md
```diff
@@ -0,0 +1,985 @@
...
+### Working with data in Flux
+When working with data in Flux, it translates the line protocol/time series structure into to a table. You 
+can think of a measurement name as a table with the tag and field keys representing columns. Flux also 
+includes the reserved column, `time`, in every table pulled from InfluxDB. The `time` column is the time at the end 
+of the line protocol when data is written in.
+
+Here's an example of line protocol and its translation to a Flux table:
+
+```
+h2o_temperature,location=santa_monica,state=CA surface_degrees=65.2,bottom_degrees=50.4 1568756160
```
Anaisdg, line 74: Maybe it would be easier to use a dummy dataset, where you don't have to memorize the tags and fields?  myMeas, myTag_1=A,myTag_2=a myField_1=int,myField_2=float 1

docs/new_flux_ideas_and_guide.md
adrian-thurston: Curious about this too, what is the rationale behind using last? Is it simply something that isn't *everything* when they neglect to give a time constraint. In other words, a *sane* default?

docs/new_flux_ideas_and_guide.md
adrian-thurston: I think I missed something here ... I understood the implementation to be like Ryan suggests. 
https://github.com/influxdata/flux/blame/pd-new-flux-ideas/docs/new_flux_ideas_and_guide.md#L131

PR #2620, opened 2020-03-12
fix: adjust test-bench config for circle ci

yzhang1991:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
Can we document somewhere, here in the PR at least or as a comment in the config, the reasoning for this change? Was their an issue with concurrently running tests? If so why could we not just remove that conflict? Or was it for another reason that we reduced the parallelism of the tests?

Issue #2592, opened 2020-03-05
Union returns two tables when they share a common group key


desa:
While investigating an issue with `monitor.StateChanges`, we noticed that queries with union were returning two tables that shared a group key. Here is an example of the table returned

```
#group,false,false,true,true,true,false,true,false,true,true,false,true,true,true,false,false,false
#datatype,string,long,string,string,string,string,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,string,long,double,string
#default,thing,,,,,,,,,,,,,,,,
,result,table,_check_id,_check_name,_measurement,_message,_source_measurement,_source_timestamp,_start,_stop,_time,_type,cpu,host,level_value,usage_user,_level
,,0,0542771d5bff9000,cpu-checker,statuses,Check: cpu-checker is: ok,cpu,1583275140000000000,2020-03-03T21:51:09.814263712Z,2020-03-03T22:51:10.062802684Z,2020-03-03T22:39:08.861380357Z,threshold,cpu-total,ip-192-168-3-153.ec2.internal,0,13.700385617056767,ok
,,1,0542771d5bff9000,cpu-checker,statuses,Check: cpu-checker is: info,cpu,1583275320000000000,2020-03-03T21:51:09.814263712Z,2020-03-03T22:51:10.062802684Z,2020-03-03T22:42:06.243786118Z,threshold,cpu-total,ip-192-168-3-153.ec2.internal,1,30.43438761804641,info
```

To work around this while we were debugging, we added a `group()` after the `union` in `allStatuses`, however if this group is removed, what should be a single table is returned as two.

```
package main
// Not OK
import "influxdata/influxdb/monitor"
import "slack"
import "influxdata/influxdb/secrets"
import "experimental"
stateChanges = (fromLevel="any", toLevel, tables=<-) => {    
    otherLevelFilter = if toLevel == "any" or fromLevel == "any" then (r) => false
                   else (r) => r._level != toLevel and r._level != fromLevel and r._measurement == "statuses"
    
    otherStatuses = tables
        |> filter(fn: otherLevelFilter)
        |> map(fn: (r) => ({r with level_value: -10}))
        |> duplicate(column: "_level", as: "l2")
        |> drop(columns: ["_level"])
        |> rename(columns: {"l2": "_level"})
        
    toLevelFilter = if toLevel == "any" then (r) => r._level != fromLevel and exists r._level
                   else (r) => r._level == toLevel
    
[...]

Issue #2623, opened 2020-03-12
Add error detection to query engine message queue


jpacik:
During query processing, tables are pushed to downstream operators via a message queue.
https://github.com/influxdata/flux/blob/a504947c20d0492fb0615f0c9a4e08ab7048dd56/execute/queue.go#L10

This message queue should validate that it receives `process` message types before other message types for each unique dataset ID. Upon receiving an invalid message it should fail here:
https://github.com/influxdata/flux/blob/a504947c20d0492fb0615f0c9a4e08ab7048dd56/execute/queue.go#L29

This ensures that scenarios https://github.com/influxdata/flux/issues/2592 are caught and fail loudly.

PR #2624, opened 2020-03-12
fix(stdlib): pivot message passing

jlapacik:
Fixes https://github.com/influxdata/flux/issues/2592.

The query engine's internal message queue requires that for every
dataset ID, `process` messages are pushed first and `finished`
messages pushed last. The optimized version of pivot invalidated
this requirement.

This commit fixes pivot so that `watermark` and `processing time`
messages are sent downstream after `process` messages.

Note the message queue itself should error if it recognizes that
the above invariant has been violated.

See https://github.com/influxdata/flux/issues/2623.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/universe/pivot.go
+++ b/stdlib/universe/pivot.go
@@ -435,6 +435,9 @@ type pivotTransformation2 struct {
 	alloc  *memory.Allocator
 	spec   PivotProcedureSpec
 	groups *execute.GroupLookup
+
+	watermark  execute.Time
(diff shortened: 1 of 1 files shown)


jlapacik:
Will fix https://github.com/influxdata/influxdb/issues/17064 when propagated to the `feat/use-algo-w` branch on influxdb

PR #2625, opened 2020-03-12
fix(libflux): force the go libflux wrapper to rebuild using go generate

### LINKED ISSUE #2511: Go build should rebuild when the libflux archives change

jsternberg:
A generate script has been added that will inject the current time into
the libflux go wrapper to force Go to recompile the wrapper. This will
force `pkg-config` to rerun and ensure that Go is using the version of
the Rust code from the current directory.

The output of the script is never meant to be checked into version
control and isn't listed on the list of build outputs. It will never be
up to date because the output isn't consistent.

This is a hopefully temporary hack until we get a better method of
generating this output in a way that is consistent. But, this is being
done because it is the easiest method of forcing a build and this
problem only occurs when building locally from source.

Fixes #2511.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ target
 /libflux/pkg
 /libflux/site/node_modules
 /libflux/.cache
+/libflux/go/libflux/hack.gen.go
 /release-notes.txt
--- a/README.md
+++ b/README.md
@@ -59,9 +59,15 @@ $ go build ./cmd/flux
 $ ./flux repl
 ```
 
-If you create or change any flux functions, you will need to rebuild the stdlib:
+If you modify any Rust code, you will need to force Go to rebuild the library.
+
 ```
-$ go generate ./stdlib
+$ go generate ./libflux/go/libflux
+```
+
+If you create or change any flux functions, you will need to rebuild the stdlib and inform Go that it must rebuild libflux:
+```
+$ go generate ./stdlib ./libflux/go/libflux
 ```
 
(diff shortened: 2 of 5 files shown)


jsternberg:
@rockstar asked me if this would have consequences for reproducible builds. It shouldn't for builds using flux as a dependency since they won't generate this file, but it would for the flux binary itself.

PR #2628, opened 2020-03-13
feat: expose format to wasm users

bthesorceror:


libflux/src/flux/formatter/mod.rs
```diff
@@ -1,32 +1,57 @@
...
 
-impl Formatter {
-    // giving too much capacity can significantly boost the performance
-    // see https://github.com/hoodie/concatenation_benchmarks-rs
-    pub fn new(cap: usize) -> Formatter {
+impl Default for Formatter {
+    fn default() -> Self {
         Formatter {
-            builder: String::with_capacity(cap),
+            builder: String::new(),
```
kelwang: the capacity is very helpful

bthesorceror: seems like a bad interface to require the user to have the length of the formatted text before they even run the formatter

libflux/src/flux/formatter/mod.rs
```diff
@@ -1,32 +1,57 @@
...
     // returns the final string and error msg
-    pub fn output(&self) -> (&str, &str) {
+    pub fn output(&self) -> Result<String, String> {
         if let Some(err) = &self.err {
-            return ("", &err.msg);
+            return Err(err.msg.clone());
         }
-        (&self.builder, "")
+
+        Ok(self.builder.clone())
```
kelwang: why clone here?

bthesorceror: Its better to return a none borrowed value for the wasm consumers.

libflux/src/flux/formatter/mod.rs
bthesorceror: doubly so since we don't even require the original script but instead a File node.

kelwang: it's looks exactly like what you changed before. But with this capacity, we can saved thousand times of heap allocations.

bthesorceror: but you have to have that value when you create a formatter

libflux/src/flux/formatter/mod.rs
```diff
@@ -1,32 +1,57 @@
...
-use crate::ast::{self, walk::Node};
+use crate::ast::{self, walk::Node, File};
+use crate::parser::parse_string;
 use crate::Error;
 
 use chrono::SecondsFormat;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub fn format_from_js_file(js_file: JsValue) -> String {
```
kelwang: does &str work?

bthesorceror: for the return value or the function argument?

bthesorceror: for the return value see the above comment

kelwang: Looks like this should be a JsString, but no need to clone it
https://github.com/rustwasm/wasm-bindgen/issues/1705

libflux/src/flux/formatter/mod.rs
```diff
@@ -1,32 +1,57 @@
...
+    "".to_string()
+}
+
+pub fn convert_to_string(file: &File) -> Result<String, String> {
+    let mut formatter = Formatter::default();
+    formatter.format_file(file, true);
+    formatter.output()
+}
+
+pub fn format(contents: String) -> Result<String, String> {
```
nathanielc: This method appears to only be used in the test cases. Should we move it to the test module?

bthesorceror: its meant to be the public interface when we start using the formatter in other places, just to encapsulate the creation of a formatter and executing said formatter

PR #2629, opened 2020-03-13
fix(compiler): should not call CheckKind when evaluating logical expr…

jlapacik:
…essions

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
@jlapacik Can you make a quick audit of all CheckKinds and just remove all of them once and for if they are no longer needed?

jlapacik:
@nathanielc they've all been removed. This was the only one left in place.

nathanielc:
Ok a quick search showed a few commented out ones, should we remove those as well?

jlapacik:
Sure I can uncomment them.

PR #2631, opened 2020-03-16
fix(libflux): make merge_packages allow no package clauses

### LINKED ISSUE #2622: Merging ASTs should allow for empty package clauses
When merging AST packages currently we validate that all files have a package clause and that they are the same. The package clause should be optional, if the package clause is missing then its implied that it is a `main` package. The merge logic should allow for the case that package clauses are missing so long either they are all missing or that if at least one is missing all othess are the `main` package.

wolffcm:
This just allows us to merge packages that have no package clauses, which is commonly the case for the web UI.

Fixes #2622.

### DIFF
--- a/libflux/go/libflux/parser_test.go
+++ b/libflux/go/libflux/parser_test.go
@@ -118,29 +118,32 @@ c = 3`)
 			name:    "packages clauses don't match",
 			outPkg:  barPkg,
 			inPkg:   fooPkg,
-			wantErr: errors.New("failed to merge packages: file's package clause: foo does not match package output package clause: bar"),
+			wantErr: errors.New(`failed to merge packages: file's package clause: "foo" does not match package output package clause: "bar"`),
 		},
 		{
 			name:    "input package has no package clause",
 			outPkg:  barPkg,
 			inPkg:   noClausePkg,
-			wantErr: errors.New("failed to merge packages: current file does not have a package clause"),
+			wantErr: errors.New(`failed to merge packages: current file does not have a package clause, but output package has package clause "bar"`),
 		},
 		{
 			name:    "output package has no package clause",
 			outPkg:  noClausePkg,
 			inPkg:   fooPkg,
-			wantErr: errors.New("failed to merge packages: output package does not have a package clause"),
+			wantErr: errors.New(`failed to merge packages: output package does not have a package clause, but current file has package clause "foo"`),
 		},
 	}
 
 	for _, testCase := range testCases {
-		gotErr := libflux.MergePackages(testCase.outPkg, testCase.inPkg)
-		if gotErr == nil {
-			t.Fatal("\nGot no error, expected:", testCase.wantErr, "\ngot: \n", gotErr)
-		}
-		if diff := cmp.Diff(testCase.wantErr.Error(), gotErr.Error()); diff != "" {
-			t.Fatalf("unexpected error: -want/+got: %v", diff)
-		}
+		testCase := testCase
+		t.Run(testCase.name, func(t *testing.T) {
+			gotErr := libflux.MergePackages(testCase.outPkg, testCase.inPkg)
+			if gotErr == nil {
+				t.Fatal("\nGot no error, expected:", testCase.wantErr, "\ngot: \n", gotErr)
+			}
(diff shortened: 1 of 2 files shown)


libflux/go/libflux/parser_test.go
```diff
@@ -118,29 +118,32 @@ c = 3`)
...
 
 	for _, testCase := range testCases {
-		gotErr := libflux.MergePackages(testCase.outPkg, testCase.inPkg)
-		if gotErr == nil {
-			t.Fatal("\nGot no error, expected:", testCase.wantErr, "\ngot: \n", gotErr)
-		}
-		if diff := cmp.Diff(testCase.wantErr.Error(), gotErr.Error()); diff != "" {
-			t.Fatalf("unexpected error: -want/+got: %v", diff)
-		}
+		testCase := testCase
```
fchikwekwe: I'm curious about this assignment here, is there a reason to create this variable?

libflux/src/flux/lib.rs
```diff
@@ -286,24 +286,35 @@ pub unsafe extern "C" fn flux_merge_ast_pkgs(
...
+        match (out_pkg_clause, file_clause) {
+            (Some(out_name), Some(in_name)) => {
+                if out_name != in_name {
+                    return Some(Error::from(format!(
+                        r#"file's package clause: "{}" does not match package output package clause: "{}""#,
+                        in_name, out_name
+                    )));
+                }
+            },
+            (None, Some(in_name)) => {
```
jsternberg: Should this and the next merge be allowed? That way we only error if the package clauses conflict.

libflux/go/libflux/parser_test.go
jsternberg: It's a technique for creating a local variable that gets captured. When you use a for loop, the memory it assigns to that variable is reused for each iteration as a way of making the for loop faster. When you use a function inside that has any parallelism, this can result in having each test use only the last value in the array instead of what would normally be expected.

To my knowledge, this is only needed when `t.Parallel()` is used since that causes the `Run` to not be executed immediately. But there's also no harm to doing it whenever you use subtests.

wolffcm: The reason is that `go test` will run some test cases in parallel, if `t.Parallel()` is called.  I guess we are not doing that here.  But if we did, then a goroutine that reads `testCase` from inside the testing `func(t *testing.T) {...}` might see the value of `testCase` change unexpectedly, and lead to non-deterministic behavior.  So, by doing `testCase := testCase`, we create a copy that can be used in the test function safely.

I have just gotten into the habit of doing it, since I've seen it elsewhere.

libflux/src/flux/lib.rs
wolffcm: I guess it depends on what we think the semantics of omitting a package clause are.

I think maybe that if you omit it, you are in the package `main`.  So omitting a package clause in a file would be okay if the other file had no package clause or `package main`, but not anything else.

Thoughts?

fchikwekwe: Where do we establish that the default package is main?

jsternberg: I think we should probably hold off on this idea and the current code is good. I'm not sure the answer to any of the above questions and I think they're really good and interesting ones.

libflux/go/libflux/parser_test.go
fchikwekwe: I figured it had to do with copying the local variable, but was curious as to why. Thanks for the explainers.

libflux/src/flux/lib.rs
wolffcm: Ok, sounds good, I'll just merge this as-is.

@fchikwekwe We have `main` declared as the default package name unceremoniously here:
https://github.com/influxdata/flux/blob/feat/use-algo-w/libflux/src/flux/ast/mod.rs#L20

fchikwekwe: Oh, okay. If that's the case, then maybe in the future we should add a check in this case and the next one to see if the included package is main before throwing an error?

wolffcm: @fchikwekwe Yes, the logic would look something like that.  I think we'd want the error message to reflect whether we are complaining about an explicit `package main` not matching a package clause in another file, or whether the clause was omitted altogether.

PR #2637, opened 2020-03-17
feat(stdlib/experimental): count distinct by tag

kelwang:
close https://github.com/influxdata/idpe/issues/6192

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/count/count.flux
```diff
@@ -0,0 +1,13 @@
...
+import "experimental"
+
+distinctByTag = (tables=<-, measurement, tag) =>
+	tables
+		|> filter(fn: (r) => r._measurement == measurement)
+		|> keep(columns: [tag])
+  		|> limit(n: 1)
+  		|> group()
+  		|> count(column: tag)
+		|> yield()
```
jsteenb2: out of curiosity, why is the yield needed here?

kelwang: the error message told me so, I am not sure why either

jsteenb2: hmm, hopefully @nathanielc will know 🤞

sanderson:
Should this be included as part of the `experimental/aggregate` package rather than the base `experimental` package?

kelwang:
> Should this be included as part of the `experimental/aggregate` package rather than the base `experimental` package?

good idea

nathanielc:
Closing as stale, but would accept a change like this in the future.

PR #2638, opened 2020-03-17
refactor: refactor eval methods to avoid roundtrips

wolffcm:
Also, make FluxCompiler and ASTCompiler use json.RawMessage instead
of *ast.Package, so deserialization can be delayed.

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -165,10 +165,16 @@ type FluxCompiler struct {
 	Query  string          `json:"query"`
 }
 
+func wrapFileJSONInPkg(bs []byte) []byte {
+	return []byte(fmt.Sprintf(
+		`{"type":"Package","package":"main","files":[%s]}`,
+		string(bs)))
+}
+
 func (c FluxCompiler) Compile(ctx context.Context, runtime flux.Runtime) (flux.Program, error) {
 	// Ignore context, it will be provided upon Program Start.
 	if c.Extern != nil {
-		hdl, err := runtime.JSONToHandle(c.Extern)
+		hdl, err := runtime.JSONToHandle(wrapFileJSONInPkg(c.Extern))
 		if err != nil {
 			return nil, err
 		}
@@ -203,7 +209,7 @@ func (c ASTCompiler) Compile(ctx context.Context, runtime flux.Runtime) (flux.Pr
 
 	// Ignore context, it will be provided upon Program Start.
 	if c.Extern != nil {
-		extHdl, err := runtime.JSONToHandle(c.Extern)
+		extHdl, err := runtime.JSONToHandle(wrapFileJSONInPkg(c.Extern))
 		if err != nil {
 			return nil, err
 		}
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -2,12 +2,12 @@ package lang
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"time"
 
 	"github.com/influxdata/flux"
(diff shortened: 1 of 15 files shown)


runtime.go
```diff
@@ -27,6 +33,11 @@ type Runtime interface {
 	LookupBuiltinType(pkg, name string) (semantic.MonoType, error)
 }
 
+type ASTHandle interface {
```
nathanielc: Can you document this interface, I think I understand its usage, but its not completely clear?

runtime.go
```diff
@@ -27,6 +33,11 @@ type Runtime interface {
...
+	ASTHandle()
```
nathanielc: Is this a method just to ensure its intentional to implement the interface? Is it ever called? If not can you document that as its not clear what else the function would do.

runtime/global.go
```diff
@@ -43,13 +44,24 @@ func Prelude() values.Scope {
...
-	return Default.evalHandle(ctx, h, opts...)
+	h, err := parser.ParseToHandle([]byte(flux))
+	if err != nil {
+		return nil, nil, err
+	}
+	return Default.Eval(ctx, h, opts...)
 }
 
 // EvalAST accepts a Flux AST and evaluates it to produce a set of side effects (as a slice of values) and a scope.
 func EvalAST(ctx context.Context, astPkg *ast.Package, opts ...flux.ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
```
nathanielc: Who calls this method? Should callers of this method just use the runtime interface instead? 

What is the reason for not changing this interface to expect an ASTHandle vs the ast.Package?

lang/compiler.go
```diff
@@ -161,43 +161,61 @@ func buildPlan(ctx context.Context, spec *flux.Spec, opts *compileOptions) (*pla
...
 func (c FluxCompiler) CompilerType() flux.CompilerType {
 	return FluxCompilerType
 }
 
-// ASTCompiler implements Compiler by producing a Spec from an AST.
+// ASTCompiler implements Compiler by producing a Program from an AST.
 type ASTCompiler struct {
-	AST *ast.Package `json:"ast"`
-	Now time.Time
+	Extern json.RawMessage `json:"extern,omitempty"`
```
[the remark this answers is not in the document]
wolffcm: In the influxdb repo, we are getting a `http.QueryRequest` and we convert it to a `query.ProxyRequest` that contains an `ASTCompiler`.  We receive the `extern` field and the query AST as separate fields in the query request.  So we would need to either deserialize and merge them, or do some creative string interpolation on the JSON bytes.  I think it's best to leave them separate.

https://github.com/influxdata/influxdb/blob/b1dd986ab3c0f4fc4afb8809f74386793527176b/http/query.go#L267-L271

runtime/global.go
wolffcm: Tasks need to use this method to evaluate an AST, it's related to processing options. They don't need to actually submit a request, but they do need to manipulate and evaluate a Go AST, and look at the resulting scope produced by evaluation.

Would it make sense to rename this `EvalGoAST`?

runtime.go
wolffcm: Yes, I added a comment.  It's needed to avoid a dependency from `flux` on `libflux/go/libflux`.

runtime/global.go
nathanielc: I like the name `EvalAST` as the function signature makes it clear it must be the Go types. Knowing why we still need it is helpful.

PR #2639, opened 2020-03-17
feat: add location information to type error messages

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

This PR adds source location to type error messages. Note there appears to be a lot of additions. Most of them come from `rustfmt` in `nodes.rs` where I have added `SourceLocation` to each type constraint.

I have also updated record and function unification in `types.rs`. The unification process hasn't changed, but I have updated the code to give more descriptive error messages.

### DIFF
--- a/libflux/src/flux/semantic/flatbuffers/types.rs
+++ b/libflux/src/flux/semantic/flatbuffers/types.rs
@@ -5,7 +5,7 @@ use crate::semantic::env::Environment;
 use crate::semantic::flatbuffers::semantic_generated::fbsemantic as fb;
 
 use flatbuffers;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 
 use crate::semantic::fresh::Fresher;
 
@@ -189,8 +189,8 @@ impl From<fb::Prop<'_>> for Option<Property> {
 impl From<fb::Fun<'_>> for Option<Function> {
     fn from(t: fb::Fun) -> Option<Function> {
         let args = t.args()?;
-        let mut req = HashMap::new();
-        let mut opt = HashMap::new();
+        let mut req = BTreeMap::new();
+        let mut opt = BTreeMap::new();
         let mut pipe = None;
         for i in 0..args.len() {
             match args.get(i).into() {
--- a/libflux/src/flux/semantic/fresh.rs
+++ b/libflux/src/flux/semantic/fresh.rs
@@ -67,6 +67,22 @@ impl<T: Hash + Ord + Eq + Fresh, S> Fresh for HashMap<T, S> {
     }
 }
 
+impl<T: Fresh> Fresh for BTreeMap<String, T> {
+    fn fresh(self, f: &mut Fresher, sub: &mut HashMap<Tvar, Tvar>) -> Self {
+        self.into_iter()
+            .map(|(s, t)| (s, t.fresh(f, sub)))
+            .collect::<Self>()
+    }
+}
+
+impl<T: Hash + Ord + Eq + Fresh, S> Fresh for BTreeMap<T, S> {
+    fn fresh(self, f: &mut Fresher, sub: &mut HashMap<Tvar, Tvar>) -> Self {
+        self.into_iter()
+            .map(|(t, s)| (t.fresh(f, sub), s))
+            .collect::<Self>()
+    }
+}
+
 impl<T: Fresh> Fresh for Box<T> {
     fn fresh(self, f: &mut Fresher, sub: &mut HashMap<Tvar, Tvar>) -> Self {
         Box::new((*self).fresh(f, sub))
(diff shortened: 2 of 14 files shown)


libflux/src/flux/semantic/flatbuffers/types.rs
```diff
@@ -189,8 +189,8 @@ impl From<fb::Prop<'_>> for Option<Property> {
 impl From<fb::Fun<'_>> for Option<Function> {
     fn from(t: fb::Fun) -> Option<Function> {
         let args = t.args()?;
-        let mut req = HashMap::new();
-        let mut opt = HashMap::new();
+        let mut req = BTreeMap::new();
+        let mut opt = BTreeMap::new();
```
rockstar: Curious about the choice here. Why switch to `BTreeMap`? I would think that for our purposes, neither data structure really provides much advantage over the other.

jlapacik: If a type error is discovered in one of the arguments of a function, the error message will display the argument with the invalid type. In the case there are multiple invalid argument types, the type checker will return the first one it finds. This of course is dependent on the order in which we iterate over the arguments, which is nondeterministic in the case of a HashMap, deterministic in the case of a BTreeMap.

Changing `HashMap` references to `BTreeMap` was a quick fix to get tests passing. However I think it makes more sense to store argument types in a list in the order they're specified by the user. So I'll probably update this PR accordingly.

libflux/src/flux/semantic/fresh.rs
```diff
@@ -67,6 +67,22 @@ impl<T: Hash + Ord + Eq + Fresh, S> Fresh for HashMap<T, S> {
     }
 }
 
+impl<T: Fresh> Fresh for BTreeMap<String, T> {
```
adrian-thurston: Would be great to get a brief description on the purpose of this ... been seeing this around and still unclear on it.

libflux/src/flux/semantic/fresh.rs
```diff
@@ -67,6 +67,22 @@ impl<T: Hash + Ord + Eq + Fresh, S> Fresh for HashMap<T, S> {
...
+    fn fresh(self, f: &mut Fresher, sub: &mut HashMap<Tvar, Tvar>) -> Self {
+        self.into_iter()
+            .map(|(t, s)| (t.fresh(f, sub), s))
+            .collect::<Self>()
+    }
+}
+
 impl<T: Fresh> Fresh for Box<T> {
     fn fresh(self, f: &mut Fresher, sub: &mut HashMap<Tvar, Tvar>) -> Self {
         Box::new((*self).fresh(f, sub))
```
adrian-thurston: Been a bit confused on the purpose of these, would be great to get a description for my understanding. It seems like many structures are endowed with fresher capabilities.

libflux/src/flux/semantic/parser/mod.rs
```diff
@@ -502,8 +507,8 @@ impl Parser<'_> {
         self.next();
         let mut token = self.next();
 
-        let mut req_args = HashMap::new();
-        let mut opt_args = HashMap::new();
+        let mut req_args = BTreeMap::new();
+        let mut opt_args = BTreeMap::new();
         let mut pipe_arg = None;
         let mut need_comma = false;
         loop {
```
adrian-thurston: It might be good for us to get in the habit of adding explicit type aliases for the generic data structures we use. It can greatly aid in understanding when you can see specifically what the map is for.

libflux/src/flux/semantic/fresh.rs
jlapacik: This is mainly used for testing purposes. Tests involve type checking a flux program and asserting the identifiers in that program are of the correct type. Expected types are usually described using the language we have for type expressions:
```
forall [t0] (x: t0) -> t0
```

For example when type checking the following program:
```
...
f = (x) => x
...
```
type inference might assign the type `forall [t5] (x: t5) -> t5` to `f`. This type should compare as equal to `forall [t0] (x: t0) -> t0`. In order to do this we need to generate two fresh types that are structurally equivalent but have the same quantified type variables. This is what `Fresh` is used for.

jlapacik: Note the types of the stdlib builtins are all specified using type expressions

https://github.com/influxdata/flux/blob/master/libflux/src/flux/semantic/builtins.rs.

Each one these type expressions is parsed and a new equivalent type with fresh type variables is generated before serialization. I believe this is the only place where `Fresh` is invoked for non-testing code.

libflux/src/flux/semantic/parser/mod.rs
adrian-thurston: Realizing now after rebasing that this is the same code I changed in the semantic-btree-hashmap patch so they are aligned, which is good.

PR #2643, opened 2020-03-18
ci: add a libflux-go target to ensure libflux-go will get rebuilt

jsternberg:
The `libflux-go` target will generate a hack file that forces go to
rebuild the libflux wrapper. This target is added as a dependency for
any target that invokes a go build command. This will force go to rebuild.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
Yea I agree, but libflux can be complicated for its dependencies. It technically relies on the stdlib right now so I thought it was best to just always regenerate and we would fix this later.

PR #2644, opened 2020-03-18
feat(stdlib/geo): added shapeData function

sanderson:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/geo/filterRowsPivoted_test.flux
```diff
@@ -0,0 +1,74 @@
+package geo_test
+
+import "experimental/geo"
+import "testing"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+inData = "
+#group,false,false,true,true,false,true,true,true,false,false,false,false,false
```
wolffcm: There are some non-obvious considerations to make when authoring flux tests, especially regarding the input data.  All of the `_test.flux` files are run in the Flux repo using `csv.from` as the data source, but they're also run in the `influxdb` repo by insert the data into a bucket and calling what most people think of as the regular `from()` like you'd use in the web UI.

This test will fail if we run it against influxdb, since it's pivoted here, and our test harness is not smart enough to un-pivot it  before it gets sent to an influxdb bucket. Could I trouble you to un-pivot the input data and then add a call to `v1.fieldsAsCols()` to the query below?

If you want to be extra helpful, you can run your test with influxdb and make sure that it passes.  Assuming you have both influxdb and flux repositories on your filesystem:
```
cd influxdb
go mod edit -replace github.com/influxdata/flux=../flux
go test ./query/... -run TestFluxEndToEnd
```

sanderson: Thanks @wolffcm. I used `pivot` instead of `v1.fieldsAsCols` just to avoid another import.

PR #2647, opened 2020-03-18
fix(stdlib/influxdata/influxdb/monitor): drop _start and _stop before union()

alespour:
This PR attempts to fix minor defect in `monitor.stateChanges()` function.

The columns `_state` and `_stop` should be dropped before `union()`, otherwise the result is two tables for two different `_level` values, and then `difference` result is always 0 and therefore the function always returns empty result.

With this fix, notification rules with "when status changes from x to y" condition work.

`|> drop(columns: ["_start", "_stop"])` is also used in `union()` example in the doc (although without explanation).


nathanielc:
@alespour Can you update/add a test case for this change?

PR #2650, opened 2020-03-19
feat(stdlib/geo): update filterRows to work with pivoted data

sanderson:
Currently `geo.filterRows` will automatically try to pivot data with `toRows` and will not work if the data is already pivoted. I added a check to see if the input tables are pivoted and, if so, skip the call to `toRows`.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


sanderson:
Closing in favor of #2644

PR #2660, opened 2020-03-19
fix(libflux): fix JSON serialization of Rust AST

wolffcm:
In some places we had `tag` attributes that were unneeded.
This lead to duplicate fields on the generated JSON.
I also added a test that roundtrips every kind of AST node.


libflux/src/flux/ast/mod.rs
```diff
@@ -416,7 +416,7 @@ pub struct Block {
 // BadStmt is a placeholder for statements for which no correct statement nodes
 // can be created.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-#[serde(rename = "BadStatement", tag = "type")]
```
jsternberg: Where is this unneeded? If it is being duplicated, it needs to be removed from the other area. I found a few of these last time I looked through here. The tag is needed on this one to ensure deserialization works correctly.

wolffcm: The other area is the `Statement` enum, since `BadStatement` never appears by itself, only as a variant within that enum, I believe that it is safe to remove.

wolffcm: The test also roundtrips everything, including `BadStmt`, so it appears to work correctly.

wolffcm: Ah, maybe I should put the `rename` back in, but I believe that `tag` is unneeded.

jsternberg: Ah thanks for clarifying. Your comment reminded me that what I remembered was the other way around. It's the `Statement` that needs to have the tag not the individual element.

With that, I think you can remove the entire annotation. It seems to be what was done in d1c3353fe339df82253c872ea1e19ba802c1d39c for a similar problem.

wolffcm: I think it still needs to be renamed, since the Go struct is `BadStatement`, and we need to be able to deserialize in Go, or at least we're supposed to be able to.

We didn't have a serialization test for `BadStatement` so I added one.

Issue #2585, opened 2020-03-05
Modify libstd environment serialization to be consistent


jsternberg:
The environment that is serialized as part of libstd will be serialized differently depending on the run because it serializes the environment which is a hash map. When we serialize the environment, we need to do so in a consistent way so that we can produce consistent binaries when doing `cargo build; cargo clean; cargo build` (in release mode).

We can either switch the environment to use a sorted map or we can modify the serialization to sort the values before serialization. The only requirement is that libstd is a consistent binary every time.

PR #2667, opened 2020-03-20
fix: convert HashMap in semantic package to BTreeMap (#2585)

adrian-thurston:
The artifacts of Algorithm W, specifically the final fresher value and the TVar
values in the envrironment, were changing from run-to-run of the algorithm.
This was a result of use of HashMap in Algo W, probably caused by an iteration
embedded in the algorithm. Since the environment is serialized and included in
each build, we lost repeatable builds.

This patch adds type aliases for all hash maps in the semantic package and then
changes the type to BTreeMap. This stabilizes the algorithm and gives us
repeatable builds.

This patch changes all hash maps to btree maps. It is possible that a more
selective switch would suffice, however there is some common generic code
shared among the map types (found in use of the SemanticMap type). That code
would need to be cloned to complete a selective switch.

### DIFF
--- a/libflux/src/flux/semantic/types.rs
+++ b/libflux/src/flux/semantic/types.rs
@@ -7,11 +7,24 @@ use std::{
     fmt,
 };
 
+// For use in generics where the specific type of map is not not mentioned.
+pub type SemanticMap<K, V> = BTreeMap<K, V>;
+pub type SemanticMapIter<'a, K, V> = std::collections::btree_map::Iter<'a, K, V>;
+
 #[derive(Debug, Clone)]
 pub struct PolyType {
     pub vars: Vec<Tvar>,
-    pub cons: HashMap<Tvar, Vec<Kind>>,
+    pub cons: TvarKinds,
     pub expr: MonoType,
+}
+
+pub type PolyTypeMap = SemanticMap<String, PolyType>;
+pub type PolyTypeMapMap = SemanticMap<String, SemanticMap<String, PolyType>>;
+
+#[macro_export]
+/// Alias the maplit literal construction macro so we can specify the type here.
+macro_rules! semantic_map {
+    ( $($x:tt)* ) => ( maplit::btreemap!( $($x)* ) );
 }
 
 impl fmt::Display for PolyType {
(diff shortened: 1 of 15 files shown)


adrian-thurston:
FYI @jlapacik I checked to see if ca037fe81095813abc879c6cdebe570682a4206e resolved the unstable `.data` files issue and to my surprise it didn't. I thought that the real issue was MonoType map, but it's more than that, so I'm pushing this PR forward.

libflux/src/flux/semantic/types.rs
```diff
@@ -7,13 +7,41 @@ use std::{
...
     pub expr: MonoType,
 }
 
+pub type PolyTypeMap = SemanticMap<String, PolyType>;
+pub type PolyTypeMapMap = SemanticMap<String, SemanticMap<String, PolyType>>;
+
+#[macro_export]
+/// This is an redefinition of the maplit::hashmap macro. It allows us to alter
+/// the map type used.
+macro_rules! semantic_map {
```
jlapacik: Can you give a little more detail here? Why do we need to redefine it?

jlapacik:
@adrian-thurston that must mean the issue is with the `PolyTypeMap`

libflux/src/flux/semantic/types.rs
adrian-thurston: Oh I wanted to just do a passthrough to whichever one SemanticMap was aliased to, but couldn't figure it out, so I cloned the whole thing, then forgot about it.

Do you know how to just pass macro arguments to another macro?

adrian-thurston:
> @adrian-thurston that must mean the issue is with the `PolyTypeMap`

Maybe it can be isolated to the one that causes the issue. For these string keys hash map is probably faster.

libflux/src/flux/semantic/types.rs
jlapacik: Oh I see, so we can update the type `SemanticMap` is aliasing without having to refactor the invocations of `semantic_map!`. I missed where you updated the map type in the body of the macro. I'm not quite sure what you mean by passthrough, but this seems like the correct approach to me.

adrian-thurston: Was looking for an effective wrap of the macros, without providing the implementation. I had to read the macro docs more carefully, but I figured it out. It's a nice system, but some hard-to-read syntax at first.
```
macro_rules! aliased_map {
    ( $($x:tt)* ) => ( maplit::hashmap!( $( $x )* ) );
}
```

adrian-thurston: That is saying match any tokens as the macro args and expand them as they are into the hash map! args.

PR #2671, opened 2020-03-23
feat(stdlib/experimental): added alignTime function

### LINKED ISSUE #2669: Function: day 0 time shift with threshold
There are number of time series analysis use cases where there are multiple series which are each shifted in time.  In order to compare them, it is useful to determine what represents time 0 for each series and then normalizing the time dimension.  Having a convenience function within Flux which would perform this time shift would be incredibly powerful.

The current `timeShift()` function adds a fixed amount of time -- and I believe a variable amount of time by series is required.
https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/timeshift/

Here are two poignant examples given what's going on currently:
The spread of COVID-19 is occurring globally, but the start of the spread is occurring in different locations at different times.

Using a specific country tag as a way of defining a series (or even using more specific `s2_cell_id` based on `lat/lon` data) to define a series along with a way to shift each series to a normalized starting point.

Example 1:
The starting point of the series could be based on the first reported case...or first data point in the series.  Take the `first()` from a series, store that `_time` column as a variable, convert it to `ns`, subtract it from each `_time` column for all data points in that series.  This delivers a normalized `_time = 0` for all data points.  But, this needs to be repeated for each series within the data set.

Example 2:
The starting point of the series is based on evaluating a specific conditional.  In this case, we might suggest that there is a minimum threshold for cases which is required before we want to start the comparison.  Rather than taking the `first()` report, we want to evaluate the data with a conditional and use the `first()` data point which evaluates to `TRUE` as `_time = 0` for that series (repeating the process outlined in Example 1).

sanderson:
Resolves #2669 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/experimental.flux
```diff
@@ -21,3 +21,20 @@ builtin to
 
 // An experimental version of join.
 builtin join
+
+// Aligns all tables/series to a common start time
+// By default, it aligns to the _start value
+// If _start doesn't exist, it aligns to 1970-01-01T00:00:00Z UTC
+alignTime = (tables=<-, alignTo=time(v: 0)) => {
+  _tableInfo = tables |> tableFind(fn: (key) => true) |> getRecord(idx: 0)
+  _startTime = if alignTo != time(v: 0) then alignTo else if exists _tableInfo._start then _tableInfo._start else time(v: 0)
```
nathanielc, line 30: Why not instead of computing the _startTime here you do it each time inside the `map` call?

 The way its written it will have to materialize the value of `tables` twice, once to get the startTime and once again to pass through the `stateDuration`, function. Additionally it assumes that the data is grouped by `_start` which may not always be the case.

stdlib/experimental/experimental.flux
```diff
@@ -21,3 +21,20 @@ builtin to
 
 // An experimental version of join.
 builtin join
+
+// Aligns all tables/series to a common start time
```
nathanielc, line 25: What value is aligned? By reading the code I can see its the `_time` column but that is not clear until I read the code.

stdlib/experimental/experimental.flux
sanderson: The `_tableInfo` and `_startTime` variables are there to deterministically set the default time to align to if the `alignTo` parameter isn't set. `_tableInfo` just pulls the first record from the first table to provide an idea of the structure of the data. `_startTime` use that info to:

1. Check if the `alignTo` param is set to anything but zero, if so, then
2. Check if the table contains a `_start` column. If so, align to the `_start` value, if not,
3. Align time to the Unix epoch, `time(v: 0)`.

This also makes it so everything aligns to a single `_start`. I don't want to use the `_start` time in each row because windowed data will have a different `_start` value per table.

> Why not instead of computing the _startTime here you do it each time inside the `map` call?

Readability –  but I can do it directly in the `map` call.



The goal here is to

stdlib/experimental/experimental.flux
```diff
@@ -21,3 +21,27 @@ builtin to
 
 // An experimental version of join.
 builtin join
+
+// Aligns all tables to a common start time by using the same _time value for
+// the first record in each table incrementing all subsequent _time values by
+// using time elapsed between input records.
+// By default, it aligns to the _start value.
```
nathanielc, line 28: So reading the implementation again I am wondering if we need this behavior. It adds ambiguity into the behavior of the function. For example which `_start` value should it use? If the are all the same then it doesn't matter, but if they differ we currently just pick the first record. What is special about the first record? Should the caller sort their data by `_start` first so they are confident the first record is meaningful? 

Thoughts? I see that the test case doesn't rely on this behavior rather instead it uses a specific `alignTo` parameter.

sanderson: It's meant to provide a reasonable default that falls inside of the queried time range. If there are more than one `_start` values, I don't know that it matters which one gets used, just that it doesn't default to the Unix epoch and doesn't force the user to define an `alignTo` time.

sanderson: From a UX perspective, if my queried time range goes back a day and the timestamps returned are from 1970, I might be a little confused.

nathanielc: I guess I don't understand the use case of this function. It seems like it is basically a `timeShift` relative to a specific date. Can you maybe walk through the intended purpose of the function to help me understand?

The reason I am concerned about the `_start` as default is it requires that we run the query twice which is probably unexpected to the user. We run it once just to get the first record, then we run it again to map over the dates. This is a limitation of the way `tableFind` works.

sanderson: The primary use case is visualizing data in a way that aligns all series to the same starting point for direct comparisons over equal amounts of time. For example, here's some COVID-19 data:

image

sanderson: But to your point, in my actual visualization, I've replaced the time dimension with a day counter. Maybe this method of pulling the `_start` time from the data isn't necessary.

image

sanderson: @nathanielc I've been swayed and removed the `_start` default. It simplified the transformation significantly. Once you align time stamps, the significance of the timestamps lessens.

nathanielc: Thanks for sharing the use case as well. The second I saw the graph it made sense to me the application of this function.

PR #2672, opened 2020-03-23
fix(interpreter): bind appropriate interpreter when evaluating functions

jlapacik:
Needed to fix https://github.com/influxdata/influxdb/issues/17065.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jlapacik:
Oops I forgot to remove that. `mutateFunctionScope` was removed with the runtime refactor, so that comment is no longer relevant.

PR #2675, opened 2020-03-23
fix(monitor): reimplement stateChanges function

desa:
Previously `monitor.stateChanges` was not implemented correctly. We did
not detect this due to insufficient testing.

The general strategy of `monitor.stateChanges` is to turn the sequence of `crit -> info -> warn`
into 0s and 1s and then to do a difference and look for records where
that difference is equal to 1. The previous implementation (1) did not
include non-to/from states in the binary sequence it produced (2) did
not correctly detect the state change state.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/monitor/state_changes_test.flux
```diff
@@ -44,13 +28,13 @@ outData = "
 ,,2,000000000000000a,cpu threshold check,warn,statuses,whoa!,cpu,1527018860000000000,2018-05-22T19:54:22Z,threshold,vaaa,vbbb,cpu-total,host.local,7.05
 "
 
-t_state_changes = (table=<-) => table
+t_state_changes_any_to_warn = (table=<-) => table
     |> range(start: -1m)
-    |> v1.fieldsAsCols()
```
jlapacik: Will these tests work on influxdb? Flux just uses csv, but influxdb will actually write the input data to storage using `to`, and I'm not sure `to` can write data that doesn't have a `_field` and `_value` column.

desa: Good question. I know that we do `v1.fieldsAsCols()` in the flux that we generate for thresholds.

https://github.com/influxdata/influxdb/blob/677e836bf4e2d1d3a056c2c7de31bce6b6065df3/notification/check/threshold_test.go#L108-L116
```
data
	|> v1.fieldsAsCols()
	|> monitor.check(
		data: check,
		messageFn: messageFn,
		ok: ok,
		info: info,
		warn: warn,
		crit: crit,
  )
```

and [`write` in `monitor` does](https://github.com/influxdata/flux/blob/6c04c73a47c2d0b6f2c84e584f441aaead82190a/stdlib/influxdata/influxdb/monitor/monitor.flux#L10)
```
option write = (tables=<-) => tables |> experimental.to(bucket: bucket)
```
in the [monitor.check](https://github.com/influxdata/flux/blob/6c04c73a47c2d0b6f2c84e584f441aaead82190a/stdlib/influxdata/influxdb/monitor/monitor.flux#L94-L125)

jlapacik: `experimental.to` expects pivoted data whereas `influxdb.to` does not. The flux end to end tests in influxdb write the input data to storage first before running the test. This is done using a call to `influxdb.to`. That's why I think the call to `v1.fieldsAsCols()` is necessary for these test cases.

jlapacik: I just tested this revision on influxdb and the tests do fail:
```
--- FAIL: TestFluxEndToEnd (24.56s)
    --- FAIL: TestFluxEndToEnd/influxdata/influxdb/monitor (0.69s)
        --- FAIL: TestFluxEndToEnd/influxdata/influxdb/monitor/state_changes_any_to_any (0.06s)
            end_to_end_test.go:174: table has no _field column
        --- FAIL: TestFluxEndToEnd/influxdata/influxdb/monitor/state_changes_info_to_any (0.06s)
            end_to_end_test.go:174: table has no _field column
        --- FAIL: TestFluxEndToEnd/influxdata/influxdb/monitor/state_changes_invalid_any_to_any (0.07s)
            end_to_end_test.go:174: table has no _field column
        --- FAIL: TestFluxEndToEnd/influxdata/influxdb/monitor/state_changes (0.06s)
            end_to_end_test.go:174: table has no _field column
```
Note that error message comes from `to` expecting a `_field` column as part of `inData`. I think you just have to format `inData` the way it was previously and add a call to `v1.fieldsAsCols` before `monitor.stateChanges`. Then it should be good to go.

desa: kk. not sure I totally follow why, but I'll do that.

alespour:
I am convinced the result on `union()` must be sorted by `_source_timestamp` and not `_time`, because `_time` is when `check()` was run and all statuses created by the check have the `_time` value.
The `state_change_test.flux` works with current implementation, if you set `_time` to the same as if it was produced by single `check()` run, the test fails.

desa:
@alespour conceptually, I agree with you. I'm trying to determine if this is actually an issue is practice. Can you reproduce a case when you would expect an alert, but that you do not get one?

alespour:
@desa I guess I see now...  I have a check with custom script, there is no aggregate function so given the scheduling I use, `check` always gets multiple rows so there are multiple statuses with the same `_time`. My check schedule is every 1m, and data arrive every 15s.

As for why it has not been observed in practice, I just checked one of out "classic" checks for a server and have noticed a strange thing... The check is scheduled to run every 1h and the query is configured with 1m aggregation window. However there is only one status in `statuses` in `_monitoring` bucket per hour. So, I have enabled custom check query and I can see the in the script, both `every` item in both `check` and `task` object uses task scheduling interval.
So my theory is that there is this^^^ bug, therefore one check run always produces one value, so in practice now it never happens that there are multiple statuses with the same `_time`.

stdlib/influxdata/influxdb/monitor/monitor.flux
```diff
@@ -20,14 +20,29 @@ from = (start, stop=now(), fn=(r) => true) =>
...
+levelUnknown = "unknown"
+
+_stateChanges = (fromLevel="any", toLevel="any", tables=<-) => {
+    otherLevelFilter = if toLevel == "any" or fromLevel == "any" then (r) => false
+                   else (r) => r._level != toLevel and r._level != fromLevel
+    
+    otherStatuses = tables
+        |> filter(fn: otherLevelFilter)
+        |> map(fn: (r) => ({r with level_value: -10}))
+        |> duplicate(column: "_level", as: "l2")
```
adrian-thurston: I see it is in the old code as well, but why are these `duplicate-drop-rename` sequences here when they appear to be an identity transformation.

adrian-thurston: Ah I see now, it enforces that `_level` exists, throwing an error if not present. Might be nice to ecapsulate these three calls in a function of their own, so as to reduce the code size and also make their purpose clear. 

One other thought ... if column `l2` exists already there is a problem? Maybe not with the schema this function is intended to work with, but if people try to use it elsewhere?

desa: > I see it is in the old code as well, but why are these duplicate-drop-rename sequences here when they appear to be an identity transformation.

Yeah, the big thing we're trying to do here is that we want to get `_level` out of the group key. This was the easiest way we could figure out to do this.

> One other thought ... if column l2 exists already there is a problem? Maybe not with the schema this function is intended to work with, but if people try to use it elsewhere?

That's a really good point. I wonder what a better name would be. Maybe something more obscure like `_____level_temp____`.

Issue #2655, opened 2020-03-19
Restructure libflux crates to avoid dynamic linking issues


nathanielc:
With https://github.com/influxdata/flux/issues/2636 we discovered that because of some Rust tricks when you allocate a Btree in one crate and pass it to another to free that was dynamically linked it creates an invalid free. This means that we should keep allocate and free the Rust structure from the same crate. This seems like a good sane practice anyways.

The current structure of the crates means that Rust objects are allocated in one crate, passed through C ABI and then free via another crate. As a result it becomes difficult to easily audit and track the alloc/free pairs.

The proposal is to restructure the crates so that this passing of pointers between crates is no longer necessary. 

Currently the implementation of the C API is spread across both the `flux` and `libstd` crates. The suggestion is to add a new create that consumes both existing crates and move all the unsafe logic that implements the C API into that single crate. This way the C API is implemented in a singe crate. The alloc/free pairs can easily be audited as they will exist in the same file. 

Here is the new proposed crate structure:

```
libflux/src/
    flux  # this is the "public" crate that directly implements the C API
    core # this is what we currently call the flux crate, it contains the implementation of Algo W
    libstd # this contains the data files of the encoded standard lib
```

With this structure the `core` crate will not contain any unsafe logic or implementation logic of the C API. The `libstd` crate will not contain any logic beyond just returning the std lib data structures. The `flux` crate is the new crate that will tie the other two together and directly implement the C API. 

The `flux` crate should export the types and function from the other two crates thus defining the public Rust API for interacting with Flux.

jsternberg:
I think this will work fine, but it does add another link target. In the future, I would like us to try refactoring this back down to one crate, but I'll write an issue at some point detailing how I think that should be done.

wolffcm:
I like it. Will we need to link against all three crates as static libraries?  If so, the name `libstd` irks me, because the file is named `liblibstd.a`, which seems strange.  Maybe rename it to `stdflux`, so the file ends up as `libstdflux.a`.  Just my $0.02.

jsternberg:
So I understand this better, is it because we're compiling the crates to static libraries which duplicates the btree code into two different implementations?

wolffcm:
@jsternberg We found the invalid free in the valgrind test, which uses dynamic linking.  In that case, yes, the btree static data is duplicated.

Not sure if this would show up in a static build. But in any case our Rust code should work linked either statically or dynamically.

jsternberg:
[...]

PR #2676, opened 2020-03-24
refactor: merge flux and libstd into one exportable crate "flux"

adrian-thurston:
1. Renamed flux to core.
2. Rename libstd to flux.
3. From new flux: pub use core::*, pub use core::ast, etc;
4. Fixed up the Cargo.toml files.
5. Fixed up the use statements needing fixing.
6. Fixed up the Makefile.

refs #2655


Makefile
```diff
@@ -154,7 +154,7 @@ test-valgrind: $(LIBFLUX_MEMTEST_BIN)
 LIBFLUX_MEMTEST_SOURCES=libflux/c/*.c
 $(LIBFLUX_MEMTEST_BIN): libflux $(LIBFLUX_MEMTEST_SOURCES)
 	$(CC) -g -Wall -Werror $(LIBFLUX_MEMTEST_SOURCES) -I./libflux/include \
-		./libflux/target/debug/libflux.a ./libflux/target/debug/liblibstd.a \
```
fchikwekwe: Ahh, this was the change to get rid of the second crate, right?

libflux/src/core/build.rs
```diff
@@ -0,0 +1,38 @@
...
+        .expect("Couldn't write bindings!");
+
+    copy("../../scanner.c", out_path.join("scanner.c")).expect("Could not copy scanner.c");
+
+    // Compile generated scanner
+    cc::Build::new()
+        .include("scanner")
+        .file(out_path.join("scanner.c"))
+        .compile("scanner");
+}
```
fchikwekwe: I honestly don't know much about Bindgen or why this file is necessary. Do you mind explaining a little bit?

libflux/src/core/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+}
+
+/// Frees a previously allocated error.
+///
+/// Note: we use the pattern described here: https://doc.rust-lang.org/std/boxed/index.html#memory-layout
+/// wherein a pointer where ownership is being transferred is modeled as a Box, and if it could be
+/// null, then it's wrapped in an Option.
+#[no_mangle]
+pub extern "C" fn flux_free_error(_err: Option<Box<ErrorHandle>>) {}
```
nathanielc: I'd like to see all of these methods moved into `flux/lib.rs` so that the C API implementation all lives in that single file.

libflux/src/core/build.rs
fchikwekwe: Why was this separated from the other code in the other `build.rs` file? Again just asking out of ignorance.

Makefile
adrian-thurston: Yup, that little patch I sent to the slack channel did that. It showed that the liblibsta.a file was already a combination of the two crates and in terms of linking in C, we already have them combined. This entire patch sort of makes it official, but renames them in the process.

libflux/src/core/build.rs
adrian-thurston: That's the code that generates the bindings for the scanner. The scanner is written in ragel, which is compiled to C. The rust parser needs to call into that and this is the part that makes  the little bit of code that enables the call from Rust to C. I don't believe I changed that part. Unfortunately the patch makes it look like big changes.

If you haven't seen it already, have a look at the rust reference on build scripts. I just learned about it myself. Basically the above code is compiled runs just before cargo tries to build the module, so it can do things like generate bindings and other preparation for building.

https://doc.rust-lang.org/cargo/reference/build-scripts.html

adrian-thurston:
> Looks good. I didn't see anything that looked to strange except for the Bindgen code being moved from one location to another. I didn't really understand that change.

So basically every file that used to be in `libflux/src/flux` is now in `libflux/src/core` and everything that used to be in `libflux/src/liblibstd` is now in `libflux/src/flux`. Unfortunately that makes it look like a huge patch.

adrian-thurston:
> Looks good. I didn't see anything that looked to strange except for the Bindgen code being moved from one location to another. I didn't really understand that change.

Ah, if you run this at the terminal it summarizes the patch and shows the moves. IMO GitHub should be showing this along with the patch.

`git log -p --stat -n 1 origin/refactor/one-public-crate`

libflux/src/core/lib.rs
adrian-thurston: Will do.

adrian-thurston: @nathanielc would you mind to have a look at the new commit?

PR #2681, opened 2020-03-25
feat: added a dynamically linked valgrind test

### LINKED ISSUE #2657: Test both dynamic and static linking in valgrind
Turns out how we link effects the behavior of various Rust/C/Go code interactions (see https://github.com/influxdata/flux/issues/2636). As such we should run the valgrind memory checking tests as both dynamic and statically linked binaries.

adrian-thurston:
Also moved the Makefile rules for the valgrind testing to libflux/c.
Fixes #2657

### DIFF
--- a/libflux/c/Makefile
+++ b/libflux/c/Makefile
@@ -1,0 +1,29 @@
+VALGRIND = valgrind
+VALGRIND_ARGS = --leak-check=full --error-exitcode=1
+
+BIN_STATIC = libflux_memory_tester_static
+BIN_DYNAMIC = libflux_memory_tester_dynamic
+
+SOURCES = *.c
+
+TARGET_DIR = $(PWD)/../target
+INCLUDE_DIR = $(PWD)/../include
+
+test-valgrind: $(BIN_STATIC) $(BIN_DYNAMIC)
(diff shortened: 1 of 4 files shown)


libflux/c/Makefile
```diff
@@ -0,0 +1,29 @@
+VALGRIND = valgrind
+VALGRIND_ARGS = --leak-check=full --error-exitcode=1
+
+BIN_STATIC = libflux_memory_tester_static
+BIN_DYNAMIC = libflux_memory_tester_dynamic
```
fchikwekwe: I'm missing a lot of context for how this Makefile works. Are these files generated in order to be able to run these tests remotely?

adrian-thurston: So these programs are compiled solely for the purpose of running valgrind. We are basically exercising our rust library and seeing if valgrind reports any leaks or memory usage errors. You can run these locally. I've only done it on linux though. I'm not sure if valgrind works on OS X.

adrian-thurston:
@wolffcm ah great that we can use recursive makes ... I was wondering about that because I could recall there being some issue, but couldn't remember what exactly. It's nice to use a recursive call here as there are plenty of vars created.

PR #2691, opened 2020-03-26
fix: check types of parts when evaluating StringExpression

wolffcm:
A string expression of the form `"... ${r.foo} ..."` where `r.foo` either does not exist or has a type other than `string` previously produced a panic. This fix now checks the type and produces an appropriate message.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/dependencies/dependenciestest"
+	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/repl"
 	"github.com/influxdata/flux/runtime"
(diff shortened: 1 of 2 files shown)


interpreter/interpreter_test.go
```diff
@@ -357,10 +373,19 @@ func TestEval(t *testing.T) {
...
+				}
+
+				// We expect an error, so it should be a non-internal Flux error.
+				fluxErr, ok := err.(*errors.Error)
+				if !ok {
+					t.Fatalf("expected error of type %T, got %T", fluxErr, err)
+				} else if fluxErr.Code == codes.Internal {
+					t.Fatalf("expected error to have code other than %s", codes.Internal)
+				}
+				return
```
fchikwekwe: Do we need to check that we get the right error text? Not just the type?

wolffcm: I guess that this test doesn't do that. There *are* other tests that *do* check error messages.  I think there is a balance where we want tests to be meaningful, but not so fragile what we have to update them often.

I considered adding logic to verify the error message, but there are several other test cases that are also expected to error, so I opted to just follow the convention in place for this test.

fchikwekwe: Okay, that makes sense to me. Thanks for clarifying 😊

PR #2693, opened 2020-03-26
feat: experimental.join

yzhang1991:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
@jlapacik 

>Can you explain why it is necessary to promote promql.join from internal to experimental?

Yes, we want to point users at this function as an alternative to the normal `join` function as its more performant. Doing so would be confusing if the user has to import `internal/promql`.

jlapacik:
@nathanielc that makes sense, but I would have expected adding `promql.join` to the experimental package to be a two line addition:
```
import "internal/promql"

join = promql.join
```

nathanielc:
@jlapacik We did it the other way around https://github.com/influxdata/flux/pull/2693/files#diff-d98f93bed47f52743f6637f3773a5066 

Its not all that important which one is the _real_ implementation but it made sense to me to have it live in `experimental`. I also like the fact that we have test cases for both import paths as we want to ensure it continues to work for both packages.

PR #2695, opened 2020-03-27
feat: add findColumn and findRecord functions

adrian-thurston:
Add findColumn and findRecord functions, which are semantically equivalent to

```
tableFind(...) |> getColumn(...)
tableFind(...) |> getRecord(...)
```

Except that they are allowed to not find a table, column or record. In these
cases they return an empty array or an empty object. The tableFind and friends
functions will return an error in these cases.

The rationale for combining these two steps into one is that it eliminates the
need to return something from tableFind which represents "no table found".
Doing that means we need to enhance the type system with a concept of a table
(not a stream) so we can prevent this thing from becomming the input to other
functions. This is a bigger undertaking and can be considered part of the next
generation of flux. Combining the two functions into one gives us the
functionality we desire without dragging in this additional work or incorrectly
exposing a nil table with the type of a stream (which is what tableFind
currently does).

More information in #2730

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -4155,6 +4155,60 @@ r0 = from(bucket:"telegraf/autogen")
 x = r0._field + "--" + r0._measurement
 ```
 
+##### FindColumn
+
+FindColumn extracts the first table from a stream of tables where the group key
+values match a given predicate, and from that table extracts a specified column.
+The column is returned as an array. The function returns an empty array if no
+table is found, or if the column label is not present in the set of columns.
+
+It has the following parameters:
+
+| Name   | Type   | Description                                                                                                   |
+| ----   | ----   | -----------                                                                                                   |
+| fn     | (key: object) -> bool | Fn is a predicate function. The column is extracted from the first table for which fn is true. |
+| column | string                | Column is the name of the column to extract.                                                   |
+
+_NOTE_: make sure that `fn`'s parameter names match the ones specified above (see [why](#Transformations)).
+
+Example:
+
+```
+vs = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> findColumn(fn: (key) => key._field == "usage_idle", column: "_value")
+
+// use values
+x = vs[0] + vs[1]
+```
+
+##### FindRecord
+
+FindRecord extracts the first table from a stream of tables where the group key
+values match a given predicate, and from that table extracts a record at a
+specified index. The record is returned as an object. The function returns an
+empty object if no table is found, or if the index is out of bounds.
+
+It has the following parameters:
+
+| Name | Type | Description                                                                                                     |
+| ---- | ---- | -----------                                                                                                     |
+| fn   | (key: object) -> bool | Fn is a predicate function. The record is extracted from the first table for which fn is true. |
+| idx  | int                   | Idx is the index of the record to extract.                                                     |
+
+Example:
+
+```
+r0 = from(bucket:"telegraf/autogen")
+    |> range(start: -5m)
+    |> filter(fn:(r) => r._measurement == "cpu")
+    |> findRecord(fn: (key) => key._field == "usage_idle", idx: 0)
+
+// use values
+x = r0._field + "--" + r0._measurement
+```
+
 ##### truncateTimeColumn 
 
 Truncates all entries in a given time column to a specified unit.
--- a/libflux/src/core/semantic/builtins.rs
+++ b/libflux/src/core/semantic/builtins.rs
@@ -508,6 +508,20 @@ pub fn builtins() -> Builtins<'static> {
                         idx: int
                     ) -> t0
                 "#,
+                "findColumn" => r#"
+                    forall [t0, t1, t2] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        fn: (key: t1) -> bool,
+                        column: string
+                    ) -> [t2]
+                "#,
+                "findRecord" => r#"
+                    forall [t0, t1] where t0: Row, t1: Row (
+                        <-tables: [t0],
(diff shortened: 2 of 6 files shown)


libflux/src/core/semantic/builtins.rs
```diff
@@ -508,6 +508,20 @@ pub fn builtins() -> Builtins<'static> {
...
+                    forall [t0, t1, t2] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        fn: (key: t1) -> bool,
+                        column: string
+                    ) -> [t2]
+                "#,
+                "findRecord" => r#"
+                    forall [t0, t1] where t0: Row, t1: Row (
+                        <-tables: [t0],
+                        fn: (key: t1) -> bool,
```
jlapacik, line 521: I think this should be `fn: (key: t0) -> bool`

adrian-thurston: Yeah I wasn't sure about that ... I based it off tableFind, which does the same thing.

docs/SPEC.md
```diff
@@ -4155,6 +4155,63 @@ r0 = from(bucket:"telegraf/autogen")
...
 
+##### FindColumn
+
+FindColumn extracts the first table from a stream of tables where the group key
+values match a given predicate, and from that table extracts a specified column.
+The column is returned as an array. The function returns an empty array if no
+table is found, or if the column label is not present in the set of columns.
+
+function 
+function errors.
```
fchikwekwe, line 4166: typo?

libflux/src/core/semantic/builtins.rs
fchikwekwe: Is this the same for both `findColumn` and `findRecord`?

stdlib/universe/table_fns.go
```diff
@@ -299,3 +314,139 @@ func objectFromRow(idx int, cr flux.ColReader) values.Object {
...
+	v, err := arguments.GetRequired(tableFindStreamArg)
+	if err != nil {
+		return nil, err
+	}
+	to, ok := v.(*flux.TableObject)
+	if !ok {
+		return nil, errors.Newf(codes.Invalid, "expected TableObject but instead got %T", v)
+	}
+
+	var fn *execute.TablePredicateFn
```
fchikwekwe, line 336: I noticed that this was passed in as an argument in `tableFind()` but here it is assigned. Just wondering if this changes anything with the way that the table is processed in this call.

libflux/src/core/semantic/builtins.rs
jlapacik: I know this is temporary so I won't block, but I did want to add some thoughts. Feel free to take them or leave them.

First, `t0` is supposed to represent the group key, however that is not enforced in the code. There's nothing preventing a user from referencing a column in their lambda that is not part of the group key. We don't have any tests for this case either which is concerning. Most likely an error is thrown from somewhere in the compiler package. But in some cases the query probably executes just fine.

Second, I see `findColumn` as just a `map`. In fact, if you look at the type signature of `map`, `findColumn` should be equivalent to:
```
// return the _value column as an array
|> ... map(fn: (r) => r._value)
```
Unfortunately `map` is specialized to table streams, but it would be great if the query engine treated arrays and table streams as interchangeable. Because this is not yet the case, we would have to implement `findColumn` separately similar to how we do now. However a major benefit of giving `findColumn` the same type signature as `map` is that now there's no more return type polymorphism and any semantic errors made on the part of the user will be caught and identified by the type checker.

adrian-thurston: @jlapacik @nathanielc I'm understanding that we need to say  that `t1` is the group-key subset of `t0`, but we have no facility to express that.

Is it true that if we do make them the same type we are doing better than the current solution because we are restricting to columns in the table, vs allowing `key` to be anything?

To be honest I don't fully understand the issues here, as I don't have a very deep understanding of the relationship between the query engine and the runtime of flux.

I can say though that `tableFind` has been doing this same thing since it's inception and that is actively used by customers. So either it's safe, or we have subtly buggy code out there. In either case it warrants more investigation I think, at least for my understanding.

adrian-thurston: @fchikwekwe yeah they both do the same thing, use one type for the table and another type for the records of the table. They were both based off `tableFind`.  It certainly seems wrong because they should be bound to each other, but I don't think we have a way to fix it properly right now.

docs/SPEC.md
adrian-thurston: Leftover text scraps, ha. Thank you!

stdlib/universe/table_fns.go
adrian-thurston: There is some code duplication there, which isn't ideal. All three of `tableFindCall`, `findColumnCall`, and `findRecordCall` extract the `fn` arg before passing it to the common worker `tableFind`. I'm not sure if we have a standard for dealing with that. Normally when I'm just pulling args from an environment I will duplicate code and make any workers common shared code, but I'm not sure how others feel about that.

libflux/src/core/semantic/builtins.rs
adrian-thurston: @nathanielc @jlapacik I'm going to merge this on the grounds that it is identical to `tableFind` but I would like to investigate further.

adrian-thurston:
> This was an interesting case and a good opportunity to streamline these functions. Thanks for letting me review.

Thank you for reviewing!

libflux/src/core/semantic/builtins.rs
jlapacik: That's fine. Like I said I'm not going block. It does concern me a little bit that we don't enforce the argument to `fn` as being part of the group key and only the group key. That is why I think it would be better to substitute `t1` with `t0`. But you're right, it's identical to tableFind's current behavior and so it is not introducing any regressions.

PR #2696, opened 2020-03-27
feat: return false if contains is called with an empty set

adrian-thurston:
The contains function would throw an error if called with an empty set. Since
the set can be computed by other functions, which may supply an empty set, it
makes sense to simply return false.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -1214,6 +1214,7 @@ type Arguments interface {
 	GetRequiredBool(name string) (bool, error)
 	GetRequiredFunction(name string) (values.Function, error)
 	GetRequiredArray(name string, t semantic.Nature) (values.Array, error)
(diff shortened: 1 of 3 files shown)


adrian-thurston:
> This change makes sense to me. In your explanation, did you mean that the empty set returned by this function might possibly be used in other functions? Or that other functions might pass an empty set to this function? Or both? Just trying to understand a bit better.

Latter: other functions might pass an empty set to this function. The motivation for adding this comes from a user's desire to use getColumn to compute some set and then filter on rows that contain values in that set. If getColumn returns an empty array then `contains` would throw an error. Much better to return false.
```
filtered_list = base
    |> limit( n: 3 )
    |> tableFind(fn: find)
    |> getColumn(column: "cpu")

filtered_data_1 = base
    |> filter(fn: (r) => ( contains(value: r.cpu, set: filtered_list) )
```

PR #2707, opened 2020-03-31
feat(sql): add Snowflake support

alespour:
This PR adds Snowflake support to `sql` package.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/sql/snowflake.go
+++ b/stdlib/sql/snowflake.go
@@ -182,17 +182,18 @@ func NewSnowflakeRowReader(r *sql.Rows) (execute.RowReader, error) {
 	return reader, nil
 }
 
+var fluxToSnowflake = map[flux.ColType]string{
+	flux.TFloat:  "FLOAT",
+	flux.TInt:    "NUMBER",
+	flux.TString: "TEXT",
+	flux.TBool:   "BOOLEAN",
+	flux.TTime:   "TIMESTAMP_LTZ",
+}
+
 // SnowflakeTranslateColumn translates flux colTypes into their corresponding Snowflake column type
 func SnowflakeColumnTranslateFunc() translationFunc {
-	c := map[string]string{
-		flux.TFloat.String():  "FLOAT",
-		flux.TInt.String():    "NUMBER",
-		flux.TString.String(): "TEXT",
-		flux.TBool.String():   "BOOLEAN",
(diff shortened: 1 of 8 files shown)


stdlib/sql/snowflake.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, err
+	}
+	reader.InitColumnTypes(types)
+
+	return reader, nil
+}
+
+// SnowflakeTranslateColumn translates flux colTypes into their corresponding Snowflake column type
+func SnowflakeColumnTranslateFunc() translationFunc {
+	c := map[string]string{
```
rockstar, line 187: Could you please hoist this value definition up and out of the function itself? It's effectively static, so we should define it up above.

stdlib/sql/snowflake.go
```diff
@@ -0,0 +1,201 @@
...
+		flux.TFloat.String():  "FLOAT",
+		flux.TInt.String():    "NUMBER",
+		flux.TString.String(): "TEXT",
+		flux.TBool.String():   "BOOLEAN",
+		flux.TTime.String():   "TIMESTAMP_LTZ",
+	}
+	return func(f flux.ColType, colName string) (string, error) {
+		s, found := c[f.String()]
```
rockstar, line 195: If we don't use `.String()` here, there's no need to use `.String()` in the above mapping.

stdlib/sql/snowflake.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (m *SnowflakeRowReader) InitColumnNames(names []string) {
+	m.columnNames = names
+}
+
+func (m *SnowflakeRowReader) InitColumnTypes(types []*sql.ColumnType) {
+	fluxTypes := make([]flux.ColType, len(types))
+	for i := 0; i < len(types); i++ {
+		switch types[i].DatabaseTypeName() {
+		case "FIXED", "NUMBER": // FIXED is reported by Snowflake driver
```
rockstar, line 119: Thanks for this comment.

stdlib/sql/snowflake.go
alespour: Absolutely. I copied that from other db support and just adjusted that to proper types.

PR #2717, opened 2020-03-31
refactor(semantic): modify function expression to always contain a block

### LINKED ISSUE #2666: Remove ambiguities in the semantic graph
The semantic graph's purpose is to represent the functional behavior of a Flux script. As such there should not be two ways to represent the same exact script. 

For example syntactically 

```
() => true
() => { return true}
```

these are different but semantically they are the same. We should have a single semantic representation of the above.

jsternberg:
The function expression used to be either able to have a function
expression or a block of expressions. To disambiguate the semantic
graph, this has been changed to only contain a block. This matches with
the rust representation of the semantic graph so there is no ambiguity
between the two anymore.

This also removes `FunctionBlock` since it wasn't necessary anymore and
`FunctionParameters` was moved into the `FunctionExpression` instead.

Fixes #2666.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
I'm making the equivalent change to influxdb first before merging this.

PR #2720, opened 2020-04-02
feat(execute): add random access group lookup

jsternberg:
The random access group lookup is a group lookup that works better with
random access rather than sorted inserts. It creates a key and then
indexes into that key rather than attempting to keep the group lookup
sorted.

This is a partial reapplication of 922f531 by only reapplying the parts
that allowed multiple group lookup implementations but doesn't include
the window changes.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/group_lookup.go
+++ b/execute/group_lookup.go
@@ -1,8 +1,10 @@
 package execute
 
 import (
+	"bytes"
 	"sort"
 
+	"github.com/apache/arrow/go/arrow"
 	"github.com/influxdata/flux"
 )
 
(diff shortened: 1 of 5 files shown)


nathanielc:
@jsternberg  If I remember right the window changes were reverted because they changed the order of data being processed. Wouldn't this group lookup change also do the same? Can we confirm that before we merge?

jsternberg:
@nathanielc the window changes were caused because `infiniteWindowTransformation` did no copies so `aggregateWindow` caused each row to be its own buffer and `derivative` was buggy with multiple buffers. It wasn't related to window itself.

This changes also doesn't modify any existing code.

execute/group_lookup.go
```diff
@@ -295,3 +297,153 @@ func (l *GroupLookup) Range(f func(key flux.GroupKey, value interface{})) {
 		}
 	}
 }
+
+// RandomAccessGroupLookup is a GroupLookup container that is optimized
+// for random access.
```
rockstar, line 302: Looking at the source, I'm not sure what the "random access" part of this is. Is it a function of `map`?

internal/execute/table/builder_cache.go
```diff
@@ -21,6 +21,34 @@ type Builder interface {
 	Release()
 }
 
+// KeyLookup is an interface for storing and retrieving
+// items by their group key.
```
rockstar, line 25: Grateful for this interface and the docs in it. Coulda used it before I started drawing things out yesterday.

execute/group_lookup.go
jsternberg: Yea. It's meant for areas where we don't necessarily know the order of incoming points where it will work consistently no matter the order they get added.

PR #2722, opened 2020-04-02
docs: add commentary on new Flux design

nathanielc:
This PR adds a doc that adds commentary on the new Flux design. The intent of the document is to help seed how this would work in the context of existing Flux features and start a more informed discussion among the Flux team.

### DIFF
--- a/docs/NewFlux_Comentary.md
+++ b/docs/NewFlux_Comentary.md
@@ -1,0 +1,251 @@
+# New Flux
+
+Author: Nathaniel Cook
+
+Recently Paul Dix and I added a document on the vision of a new Flux that is more usable.
+The design there solves several key points of feedback we have received on Flux.
+
+
+That document describes the what and some of the why. This document will discuss the how.
+This document is not meant to be prescriptive rather just my personal commentary on how I see the vision of new Flux could be implemented.
+It is meant to be a seed to start a good team discussion around how  we can implement the new Flux vision.
+
+
+## What is new?
+
+First before we talk about how we would implement the new vision, we should clarify what is new.
+
+### Single Table vs Stream of Tables
+
+The new Flux data model changes to be a single table with a homogeneous schema.
+This is different from the current design which represents data as a stream (unbounded sequence) of tables with possibility differing schemas.
+
+This change is fundamental to the new design, the most common pit fall with current Flux is working with this stream of tables that may or may not be the same structure.
+
+### Influxdb.From is always pivoted
+
+To go along with the single table and homogeneous schema the data returned from the `influxdb.from` function will be _pivoted_.
+Specifically the _\_field_ and _\_value_ structure of the data has been removed and instead column names correspond with the name of the InfluxDB tag or field.
+Like before, no distinction is made between tags and fields as columns in the data model.
+
+A consequence of this change is that the `from` call now also requires the `measurement` parameter as the InfluxDB data model only guarantees a homogeneuos schema within the scope of a measurement.
+
+### Syntax Sugar for From
+
+An additional syntax is proposed that makes querying data from InfluxDB much easier to type.
+
+The syntax is:
+
+```
+@bucket.measurement
+```
+
+The syntax is equivalent to:
+
+```
+influxdb.from(bucket:"bucket", measurement:"measurement")
+```
+
+### Syntax Sugar for Filtering
+
+There is new syntax to make creating filter predicates easier. This syntax has two parts.
+
+1. Declares which columns to retrieve from the database.
+2. Declares a predicate for which data to retrieve from the database.
+
+
+The syntax is:
+
+```
+{location,state, state == "CA"}
+```
+
+That syntax is equivalent to:
+
+```
+filter(fn: (r) => r.state == "CA")
+```
+
+This has the addition of providing explicit expectation of which columns should be returned from the database.
+Currently Flux doesn't know how to describe that behavior and so it has no existing equivalent.
+
+It is intended that this new syntax can occur anywhere within a normal Flux data pipeline.
+
+For example this should also be valid:
+
+```
+influxdb.from(bucket:"bucket", measurement:"measurement"){location, state, state == "CA"}
+```
+
+In the above I have simply chosen to be more explicit about the `from` call instead of using the `@` shorthand notation.
+
+### Aggregates and Selectors
+
+With the change to a single table model, aggregates and selectors are now expressed differently.
+There is a new function `aggregate` that also comes with its own syntax. More on the syntax in a bit.
+
+The `aggregate` function has a `by` parameter which specifies the dimensions on which you would like to perform an aggregate function.
+Similarly the `select` function has a `by` parameter. 
+
+
+These functions now make it explicit how you want to perform an operation and that the grouping of that operation is scoped to just that operation.
+No longer is there a persistent state on the table that indicates how the table (or stream of tables) are grouped.
+
+Similarly the `aggregate` and `select` functions will have parameters for `windowing`, that enable you to define how you would like to group the data in time.
+
+The `group` and `window` functions are not part of this new Flux vision.
+
+### Syntax Sugar for Aggregates
+
+Aggregates also come with their own new syntax. The intent of this syntax is to make it easy to describe which aggregate function to apply to which columns.
+This syntax allows for that definition to be explicit and terse.
+
+The syntax is:
+
+```
+ |> aggregate({min(degrees_bottom), max(degrees_bottom), mean(degrees_bottom)}, by:["location"])
+```
+
+The semantics of this syntax are to apply three different aggregate functions to the same column named `degrees_bottom`.
+The output table will then have three columns named `min_degrees_bottom`, `max_degrees_bottom`, and `mean_degrees_bottom`.
+
+
+>NOTE: Selectors do not need their own syntax as it is easy to specify selection in the existing syntax.
+
+### Column Indexing
+
+A consequence of data being represented as a single table the `tableFind` function is not needed. As such  a user can call `getColumn` directly on a table.
+We can go one step further and use the `.` syntax to mean indexing a table for its column.
+
+For example:
+
+```
+data = @foo.bar{location, state == "CA"}
+
+data.location // a list of locations in the state of CA
+```
+
+
+## The How
+
+Now that we have an overview of the new features syntax, let's discuss how I expect they can be implemented.
+Again this is just to seed the conversation and is by no means a prescription on how it must be done.
+
+
+### Single Table vs Stream of Tables
+
+The current Flux internals use the interpreter to create DAG of promises to data to construct a query spec.
+That spec is then planned etc and passed down to the execution engine.
+
+So long as we can continue to build a query spec from this new Flux API, we can continue to execute the queries.
+My rough plan is that a set of new Flux functions will be added to a `table` package.
+The `table` package will be very similar to the current `universe` package. It will contain all basic functions and new functions as well like `aggregate`.
+
+This new `table` package will need to provide its own mechanism for producing a query spec.
+The current execution engine can handle streams of tables with heterogenous schemas.
+A homogenous schema is a subset of functionality in a heterogeneous schema and as such can run on the execution engine without issue.
+We will still leverage the engines concept of groups to do aggregates etc but those details will be hidden from the user.
+
+As a rough example to illustrate how this would work, the following new Flux:
+
+
+```
+@data.temperatures{location,state,degrees_bottom}
+ |> aggregate({min(degrees_bottom)}, by:["location"])
+```
+
+would generate a spec similar to what this Flux would generate
+
+```
+from(bucket:"data")
+ |> filter(fn:(r) => r._measurement == "temperatures")
+ |> group(columns:["location"])
+ |> min()
+ |> group()
+```
+
+
+### Influxdb.From is always pivoted
+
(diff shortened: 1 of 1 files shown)


docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+### Single Table vs Stream of Tables
+
+The new Flux data model changes to be a single table with a homogeneous schema.
+This is different from the current design which represents data as a stream (unbounded sequence) of tables with possibility differing schemas.
+
+This change is fundamental to the new design, the most common pit fall with current Flux is working with this stream of tables that may or may not be the same structure.
+
+### Influxdb.From is always pivoted
+
+To go along with the single table and homogeneuos schema the data returned from the `influxdb.from` function will be _pivoted_.
```
fchikwekwe, line 27:
```suggestion
To go along with the single table and homogeneous schema the data returned from the `influxdb.from` function will be _pivoted_.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
+# New Flux
+
+Author: Nathaniel Cook
+
+Recently Paul Dix and I added a document on the vision of a new Flux that is more usable.
+The design there solves several key points of feedback we have received on Flux.
+
+
+That document describe the what and some of the why. This document will discuss the how.
```
fchikwekwe, line 9:
```suggestion
That document describes the what and some of the why. This document will discuss the how.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+## What is new?
+
+First before we talk about how we would implement the new vision, we should clarify what is new.
+
+### Single Table vs Stream of Tables
+
+The new Flux data model changes to be a single table with a homogeneous schema.
+This is different from the current design which represents data as a stream (unbounded sequence) of tables with possibility differing schemas.
+
```
fchikwekwe, line 22: This eliminates group keys in favor of a schema? How does this change the way that users are able to group and chart data for comparison?

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+This change is fundamental to the new design, the most common pit fall with current Flux is working with this stream of tables that may or may not be the same structure.
```
fchikwekwe, line 23: This is a current pitfall of Flux? I'm not sure I understand the point made here.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Influxdb.From is always pivoted
+
+To go along with the single table and homogeneuos schema the data returned from the `influxdb.from` function will be _pivoted_.
+Specifically the _\_field_ and _\_value_ structure of the data has been removed and instead column names correspond with the name of the InfluxDB tag or field.
+Like before no distinction is made between tags and fields as columns in the data model.
```
fchikwekwe, line 29:
```suggestion
Like before, no distinction is made between tags and fields as columns in the data model.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Syntax Sugar for From
+
+An additional syntax is proposed that makes querying data from InfluxDB much easier to type.
+
+The syntax is:
+
+```
+@bucket.measurement
+```
```
fchikwekwe, line 41: I really like the simplicity of this syntax. I feel like it will be much for familiar to user's coming from other scripting languages.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+{location,state, state == "CA"}
+```
+
+That syntax is equivalent to:
+
+```
+filter(fn: (r) => r.state == "CA")
+```
+
+With the addition of providing explicit expectation of which columns should be returned from the database.
```
fchikwekwe, line 69:
```suggestion
This has the addition of providing explicit expectation of which columns should be returned from the database.
```
I wasn't sure of the edit here.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+```
+data = @foo.bar{location, state == "CA"}
+
+data.location // a list of locations in the state of CA
+```
+
+
+## The How
+
+Now that we have an overview of the new features syntax lets discuss how I expect they can be implemented.
```
fchikwekwe, line 131:
```suggestion
Now that we have an overview of the new features syntax, let's discuss how I expect they can be implemented.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+Again this is just to seed the conversation and is by no means a prescription on how it must be done.
+
+
+### Single Table vs Stream of Tables
+
+The current Flux internals use the interpreter to create DAG of promises to data to construct a query spec.
+That spec is then planned etc and passed down to the execution engine.
+
+So long as we can continue to build a query spec from this new Flux API we can continue to execute the queries.
```
fchikwekwe, line 140:
```suggestion
So long as we can continue to build a query spec from this new Flux API, we can continue to execute the queries.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+My rough plan is that a set of new Flux functions will be added to a `table` package.
+The `table` package will be very similar to the current `universe` package, it will contain all the basic functions and the new functions as well like `aggregate`.
```
fchikwekwe, line 142:
```suggestion
The `table` package will be very similar to the current `universe` package. It will contain all basic functions and new functions as well like `aggregate`.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+This new `table` package will need to provide its own mechanism for producing a query spec.
+The current execution engine can handle streams of tables with heterogenous schemas.
+A homogenuos schema is a subset of functionality in a heterogeneuos schema and as such can run on the exectution engine without issue.
```
fchikwekwe, line 146:
```suggestion
A homogenous schema is a subset of functionality in a heterogeneous schema and as such can run on the execution engine without issue.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+We will still leverage the engines concept of groups to do aggregates etc but those details will be hidden from the user.
+
+As a rough example to illustrate how this would work the following new Flux:
```
fchikwekwe, line 149:
```suggestion
As a rough example to illustrate how this would work, the following new Flux:
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+The `influxdb.from` function will be updated to accept a `measurement` parameter.
+If it exists the spec will produce the appropriate filter.
+If it doesn't exist the behavior will not change.
+Once we have versioning we can then break the API of the function and require the measurement parameter.
+
+
+### Syntax Sugar
+
+Syntax sugar will be implemented as a kind of macro.
+By macro I mean a source to source transformation before any semantic analysis is performed.
```
fchikwekwe, line 179: Are these `macros` going to be written in pure Flux?

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+Proposed type signature for the aggregate functions.
+
+```
+min : forall [t0] where t0 : Comparable (data: [t0]) -> t0
+max : forall [t0] where t0 : Comparable (data: [t0]) -> t0
+mean : forall [t0] where t0 : Numeric (data: [t0]) -> float
+```
+
+Finally there is one piece missing from this design, the group columns got dropped.
```
fchikwekwe, line 242:
```suggestion
Finally, there is one piece missing from this design. The group columns got dropped.
```

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+In the example you would expect the output data to contain the `location` column, but as I have described the design it would be dropped.
+Obviously we need a fix for that, we could play more macro tricks or something. Looking for ideas on this part of the problem.
+
+
+
+
+## Summary
+
+In summary we see these changes as being additive to the existing language with the addition of some new core features like macros and a new `table` package for these new function types.
```
fchikwekwe, line 251:
```suggestion
In summary, we see these changes as being additive to the existing language with the addition of some new core features like macros and a new `table` package for these new function types.
```

docs/NewFlux_Comentary.md
nathanielc: User's can still group their data, but instead of that grouping being persisted across different functions/transformation the grouping only _lasts_ as long at the function that used it. For example the `aggreate` function will group the data, compute the aggregate of each result and then output a single table with the results from each group. After the call to `aggregate` the grouping is no longer relevant/important.

docs/NewFlux_Comentary.md
nathanielc: The ways this pitfall manifests is users write a Flux script they expect to work for data following a  certain shape. But the data shape isn't clear from just the script text so they often find the data has a different shape than they expected and so their script doesn't work. To make it even harder the shape of the data can change for each table, so while the script works for some of their data it doesn't work for all of it.

So given that a user needs to understand the shape of their data in order to write a good Flux script, the current design can change the shape of the data in unexpected ways. For example this query changes the group key three different times. Can you spot them?

```
from(bucket:"telegraf") 
    |> range(start: -1h)
    |> filter(fn: (r) => r.env == "acc")
    |> group(columns:["region", "env", "_measurement", "_field"])
    |> mean()
    |> map(fn: (r) => ({region: r.region, env: r.env, _measurement: r._measurement, _value: r._value * 2.0}))
    |> keep(columns:["region", "_value"])
    |> sum()
```

The calls to `group`, `map` and `keep` all change the shape of the data because they change the group key.

The end result is the user has to remember what there group key is and they have to follow the flow of the group key through their script in order to know what it is.

docs/NewFlux_Comentary.md
nathanielc: I don't think so, but I don't have a clear idea of how they would be implemented.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,252 @@
...
+
+```
+@bucket.measurement
+```
+
+The syntax is equivalent to:
+
+```
+influxdb.from(bucket:"bucket", measurement:"measurement")
+```
```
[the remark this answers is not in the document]
nathanielc: Yes this is correct. I'd like the syntax to be more generally applicable. Open to ideas there.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,251 @@
...
+@data.temperatures{location,state,degrees_bottom}
+ |> aggregate({min(degrees_bottom)}, by:["location"])
+```
+
+would generate a spec similar to what this Flux would generate
+
+```
+from(bucket:"data")
+ |> filter(fn:(r) => r._measurement == "temperatures")
```
wolffcm, line 161: There is a missing filter predicate `r._field == "degrees_bottom"`, right?  It seems like we would need to know which columns are InfluxDB tags and which are fields statically to make this transformation?

docs/NewFlux_Comentary.md
fchikwekwe: I think the idea of framing this syntactic sugar as macros is potentially powerful. It prefers the versatility of the underlying language while providing a stage for more terseness and innovation. I also thinks this may potentially make this language more open source/community friendly depending on how we implement this.

docs/NewFlux_Comentary.md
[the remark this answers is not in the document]
fchikwekwe: I didn't even pick up on this idea @rockstar, but its an important consideration.

docs/NewFlux_Comentary.md
```diff
@@ -0,0 +1,251 @@
...
+
+As a rough example to illustrate how this would work, the following new Flux:
+
+
+```
+@data.temperatures{location,state,degrees_bottom}
+ |> aggregate({min(degrees_bottom)}, by:["location"])
+```
+
+would generate a spec similar to what this Flux would generate
```
fchikwekwe, line 157:
```suggestion
Would generate a spec similar to what this Flux would generate:
```
Consistent with the ccpitalization and colon above.

docs/NewFlux_Comentary.md
fchikwekwe: I'm confused. Do we need to filter on the field here in order to perform an aggregate function on it?

wolffcm: I guess since the `from` function will produce pivoted data, there will be no `_field` column.  But I still don't see how we specify which fields from a measurement to get from storage.  Is it all of then, by default?

nathanielc: @wolffcm I was thinking we would send the entire list of columns down to storage and let it figure it out compared to the series index. It should be able to work out which columns are tags and which are fields at that point. Then we do the pivot directly as part of the `from` implementation for only the fields that are returned.

Does that sound like it would work?

wolffcm: It seems doable.  I don't think we have any fields vs tags logic in the read path in storage currently, at least for Flux.  It seems like we'll need to change our interface with storage significantly to make this work correctly and efficiently.

PR #2725, opened 2020-04-03
fix(monitor): sort statuses by source timestamp

alespour:
This PR make statuses sorted ~~optional with default~~ by `_source_timestamp`. Sort by _source_timestamp is the natural order of statuses (the `_time` column is the time when check was run and the order may differ from the original timeline).

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/monitor/state_changes_custom_any_to_any_test.flux
```diff
@@ -0,0 +1,54 @@
...
+import "influxdata/influxdb/v1"
+import "testing"
+import "experimental"
+
+option now = () => 2018-05-22T19:54:40Z
+
+option monitor.log = (tables=<-) => tables |> drop(columns:["_start", "_stop"])
+
+// These statuses were produce by custom check query
+inData = "
```
desa, line 13: @jlapacik pointed out that we have to provide tables in an unpivotted format so that these tests will pass when ran against influxdb. Could you switch this to the unpivotted format? Just make the same call you did to get this data, but don't add the `v1.fieldsAsCols()` call.

stdlib/influxdata/influxdb/monitor/monitor.flux
```diff
@@ -12,6 +12,9 @@ option write = (tables=<-) => tables |> experimental.to(bucket: bucket)
 // Log records notification events
 option log = (tables=<-) => tables |> experimental.to(bucket: bucket)
 
+// Reorder sorts statuses before difference computation
+option reorder = (tables=<-) => tables |> sort(columns: ["_source_timestamp"], desc: false)
```
desa, line 16: Can you document why this is important.

nathanielc:
@desa @alespour Can either of you add some context here? I am curious what the use case for this is outside of test cases. 

Will the UI leverage this feature some how? What reasons would I as a user want to change the sort order?

desa:
@nathanielc yeah, definitely. @alespour reached out to me about an issue he hit in the `stateChanges` function. He noticed that it was possible for the `_time` and `_source_timestamp` to drift in such a way that the sort order of the statuses differs (which violated my intuition, I would have expected sort the sort order to remain the same).

nathanielc:
@desa does that mean we just need to change the sort order? Why does it need to be an option that the user can control?

desa:
@nathanielc yeah changing the sort order to always be `_source_timestamp` makes sense.

alespour:
@nathanielc @desa Changed to always sort by `_source_timestamp`.

adrian-thurston:
Hey @alespour, just circling back on this with the intention to merge it and I noticed that the additional test doesn't pass in the context of influxdb. If you're unaware, we run all the tests against an influxdb launcher. The driver is in `influxdb/query/stdlib/testing/`. To make it run I think you just need to eliminate the now option and add a range liike
```
    |> range(start: 2020-01-01T00:00:00Z )
```

alespour:
Thank you @adrian-thurston for the update, I hope I understood what you meant and now the additional test in this PR would pass against influxdb launcher.

adrian-thurston:
Hi @alespour yes that's it! With those changes the test will work both in flux and influxdb. Thanks!

PR #2728, opened 2020-04-03
fix(interpreter): resolve member expressions

jlapacik:
This allows the query planner to be able to push predicates like:

    r.env == v.env

down to storage.


nathanielc:
@jlapacik Glad this is fixed. What was unique about algow that made this a problem compared to master?

jlapacik:
According to the interpreter code, member expressions are not resolved on master. Therefore I would not expect predicates containing them to be pushed down by the query planner. Anecdotal evidence suggested otherwise (for some queries). Unfortunately we do not have the ability test this inside the planner.

See https://github.com/influxdata/flux/issues/2729 for expanding planner tests.

PR #2742, opened 2020-04-09
feat(semantic): add semantic nodes for bad statement and bad expression

### LINKED ISSUE #2708: Produce a broken semantic graph (graph with errors) from a broken parse tree
Given a parse tree with errors in it, attempt, within reason, to produce a semantic graph also with errors. This should involve classifying errors into some number of groups and developing corresponding error nodes for the semantic graph. These should correspond to the kinds of errors we wish to push past in the tools, such as when the user is typing and the program is therefore incomplete, or when the user has an unknown (to them) error.

jsternberg:
Semantic nodes for bad statement and bad expression have been added so
that bad statement and bad expression can be translated into the
semantic graph. They contain the location information and a list of
errors that are associated with the inner statement or expression.

Fixes #2708.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/core/semantic/convert.rs
+++ b/libflux/src/core/semantic/convert.rs
@@ -107,9 +107,7 @@ fn convert_statement(stmt: ast::Statement, fresher: &mut Fresher) -> Result<Stat
         ast::Statement::Variable(s) => Ok(Statement::Variable(Box::new(
             convert_variable_assignment(*s, fresher)?,
         ))),
-        ast::Statement::Bad(_) => {
-            Err("BadStatement is not supported in semantic analysis".to_string())
-        }
+        ast::Statement::Bad(s) => Ok(Statement::Bad(convert_bad_statement(s))),
     }
 }
 
@@ -178,6 +176,17 @@ fn convert_member_assignment(stmt: ast::MemberAssgn, fresher: &mut Fresher) -> R
     })
 }
 
+fn convert_bad_statement(e: ast::BadStmt) -> Box<BadStmt> {
+    let loc = e.base.location;
+    let errors = vec![Error::ASTError(e.text, loc.clone())];
+    let statement = None;
+    Box::new(BadStmt {
+        loc,
+        errors,
+        statement,
+    })
+}
+
 fn convert_expression(expr: ast::Expression, fresher: &mut Fresher) -> Result<Expression> {
     match expr {
         ast::Expression::Function(expr) => Ok(Expression::Function(Box::new(convert_function_expression(*expr, fresher)?))),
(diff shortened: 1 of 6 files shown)


libflux/src/core/semantic/nodes.rs
```diff
@@ -70,6 +71,7 @@ impl fmt::Display for Error {
                 write!(f, "error {}: invalid import path {}", loc, path)
             }
             Error::InvalidReturn(loc) => write!(f, "error {}: return not valid in file block", loc),
+            Error::ASTError(s, loc) => write!(f, "error {}: {}", s, loc),
```
nathanielc, line 74: Should this `write!(f, "error {}: {}", s, loc)` be `write!(f, "error {}: {}", loc, s)`?

libflux/src/core/semantic/nodes.rs
```diff
@@ -88,6 +90,7 @@ pub enum Statement {
     Return(ReturnStmt),
     Test(Box<TestStmt>),
     Builtin(BuiltinStmt),
+    Invalid(Box<BadStmt>),
```
nathanielc, line 93: For consistency can you name the enum `Bad`? I get that its a bit annoying but the consistency with the rest of the variants and naming convention is important.

libflux/src/core/semantic/nodes.rs
```diff
@@ -1786,6 +1831,12 @@ fn infer_literal(
     Ok((env, constraints))
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct BaseNode {
+    pub loc: ast::SourceLocation,
+    pub errors: Vec<Error>,
+}
```
nathanielc, line 1838: Is this unused now?

libflux/src/core/semantic/convert.rs
```diff
@@ -178,6 +176,17 @@ fn convert_member_assignment(stmt: ast::MemberAssgn, fresher: &mut Fresher) -> R
     })
 }
 
+fn convert_bad_statement(e: ast::BadStmt) -> Box<BadStmt> {
+    let loc = e.base.location;
+    let errors = vec![Error::ASTError(e.text, loc.clone())];
+    let statement = None;
```
nathanielc, line 182: Should this match on `convert_statement` like the `convert_bad_expression` does?

jsternberg: That doesn't yet exist yet on the ast statement. I just thought I would add the attribute ahead of time.

libflux/src/core/semantic/nodes.rs
jsternberg: Yes. This was a mistake.

libflux/src/core/semantic/nodes.rs
jsternberg: Sure. Maybe we can consider changing the enum in the AST too in the future.

libflux/src/core/semantic/nodes.rs
jsternberg: Yes.

libflux/src/core/semantic/convert.rs
nathanielc: Ok makes sense. I didn't realize the ast BadStatement didn't embed its own statement

PR #2751, opened 2020-04-14
refactor: remove type annotations from literal expressions

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -333,7 +333,6 @@ func compile(n semantic.Node, subst map[uint64]semantic.MonoType, scope Scope) (
 			parts[i] = e
 		}
 		return &stringExpressionEvaluator{
-			t:     monoType(subst, n.TypeOf()),
 			parts: parts,
 		}, nil
 	case *semantic.TextPart:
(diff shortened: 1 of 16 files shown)


nathanielc:
@jlapacik Why does this fix the issue? It seems like those literal values should have been properly constrained and therefore always have the correct type. 

It seems like a good simplification but I'd like to understand how this fixes the issue.

jlapacik:
The bug was caused when the query planner pushed down part of a predicate. It did not annotate the newly instantiated logical expression with any type.

nathanielc:
ok, so this gets around the issue by ensuring that literals always know their type. What about other kinds of expressions? Will the pushdown logic break any other kinds of expression? Do we need to teach pushdown how to preserve the type info?

jlapacik:
To be clear, this was a good opportunity to remove type annotations from literals. However this issue came directly from logical expressions and the fact that it was possible to instantiate a logical expressions on the go side without assigning it a type. Of course the type should always be bool, so now it is.

Currently the only semantic structures we manipulate/instantiate during planning are logical expressions, so this particular error will not occur again.

PR #2759, opened 2020-04-20
feat: evaluate and store "now" in execution deps for tableFind to use

adrian-thurston:
Added "now" to execution dependencies. These are the dependencies that are
necessary to start an execution pipeline from the evaluation step, as is done
by tableFind.

The "now" option is now evaluated and stored when the interpreter evaluates the
option. It is stored in the execution dependencies. This approach allows the
"now" option to be made available to tableFind during an interpreter
evaluation.

No longer computing "now" in between evaluation and execution. Can just pull
"now" from the execution dependencies. These will either come in from the
arguments, or will be computed during evaluation. We don't need to compute it
again. Because we removed this we can also remove calls to SetNowOptions as it
is no longer necessary to ensure "now" is in the scope. It is no longer looked
up.

We also inject some default execution dependencies for the repl to use, making
tableFind and friends work properly in the repl.


stdlib/universe/table_fns.go
```diff
@@ -79,21 +79,22 @@ func tableFindCall(ctx context.Context, args values.Object) (values.Value, error
 		}
 	}
 
+	if !execdeps.HaveExecutionDependencies(ctx) {
+		return nil, errors.New(codes.Invalid, "do not have an execution context for tableFind, if using the repl, try executing this code on the server using the InfluxDB API")
```
rockstar, line 83: This message feels concerning to me. "If your preferred usage doesn't work, try this other usage." We've learned that there are a lot of people using the REPL, and that that's their preferred way to work with flux. We should absolutely do what we can to make the REPL a first-class flux experience. It looks like the support in the REPL is in this patch, so I'm trying to understand are the conditions whereby this error can occur.

interpreter/interpreter.go
```diff
@@ -177,6 +178,31 @@ func (itrp *Interpreter) doStatement(ctx context.Context, stmt semantic.Statemen
 	return nil, nil
 }
 
+// If the option is "now", evaluate the function and store in the execution
+// dependencies.
+func (irtp *Interpreter) evaluateNowOption(ctx context.Context, name string, init values.Value) {
```
rockstar, line 183: Is there a case where this might become more generic, to handle other options that need to be evaluated at execution? Not really a thing here, just wondering, as I build my internal representation of this patch.

interpreter/interpreter.go
```diff
@@ -177,6 +178,31 @@ func (itrp *Interpreter) doStatement(ctx context.Context, stmt semantic.Statemen
...
+	if name != "now" {
```
rockstar, line 184: We tend to use this string a lot for the same sorts of things. I wonder if it should be a constant somewhere (for our own sake; I assume the compiler interns the string).

stdlib/universe/table_fns.go
adrian-thurston: Oh the patch is actually meant to fix this case, as it supplies the execution dependencies (last point in the PR description) You're right it is a horrible error message. It's the one I wrote when I first made a stab at this and didn't understand the issues, so I just told the user to not use the repl :)

That path used to cause a panic, so it should probably go back to causing a panic. Or actually probably an internal error would be better, as we should always supply the necessary things for tableFind and friends to work.

https://github.com/influxdata/flux/pull/2558

interpreter/interpreter.go
adrian-thurston: It is, but trying to import the right package triggers dependency hell. I was torn ... try to fix that in this patch, or leave it as a literal. What do you think?

interpreter/interpreter.go
adrian-thurston: I'm not sure ... now is the only option I've ever used.

interpreter/interpreter.go
rockstar: I'm satisfied to know you thought about it. I'd prefer not to unravel the sweater for the sake of this patch.

stdlib/universe/table_fns.go
rockstar: Right, the patch itself didn't seem to align with the error message. I think, given your reply, that a panic might be the most appropriate thing; that state should be impossible _because_ of the changes in the patch. It's essentially an assertion on the code.

interpreter/interpreter.go
rockstar: Heh, same.

stdlib/universe/table_fns.go
adrian-thurston: This is what I don't like about dependency injection, it leaves no room to specify that certain dependencies are mandatory. A context class with a new function with a required argument does so much more to communicate what the dependencies are.

interpreter/interpreter.go
```diff
@@ -177,6 +178,31 @@ func (itrp *Interpreter) doStatement(ctx context.Context, stmt semantic.Statemen
...
+	if !execdeps.HaveExecutionDependencies(ctx) {
+		return
+	}
+
+	// Evaluate now.
+	nowTime, err := init.Function().Call(ctx, nil)
+	if err != nil {
+		return
+	}
+	now := nowTime.Time().Time()
```
jlapacik, line 196: The implementation of `now()` does not use the `ctx` variable. And it doesn't look like it has been updated which means that unfortunately `nowTime` will always be the current time of execution. I.e. it will never be overridden.

jlapacik: https://github.com/influxdata/flux/pull/2773 is needed in order for tasks to override the default `now` time outside of flux

adrian-thurston: Ah, I didn't realize there was also a now() function that needs to be updated.

internal/spec/build.go
```diff
@@ -138,23 +139,17 @@ func FromScript(ctx context.Context, now time.Time, script string) (*flux.Spec,
 	}
 	s.Finish()
 
+	deps := execdeps.NewExecutionDependencies(nil, &now, nil)
+	ctx = deps.Inject(ctx)
+
 	s, cctx := opentracing.StartSpanFromContext(ctx, "eval")
-	sideEffects, scope, err := flux.EvalAST(cctx, astPkg, flux.SetNowOption(now))
+	sideEffects, _, err := flux.EvalAST(cctx, astPkg)
```
jlapacik, line 146: To clarify, this change (removing `flux.SetNowOption(now)`) introduced a regression. Before this change, the `now` time that was passed in would be used to update the variable environment. Specifically the option `now` would be assigned a new function value.

After this change, the environment is no longer mutated. Instead the `now` time is injected into the execution context. If the `now` option is assigned as part of a flux program, everything works as before. However if a `now` time is passed via the API, because the `now` option is no longer updated in the environment with that new value, the default implementation of `system.time()` will be used.

At the time of this change `system.time()` just ignored the execution context that it was given. Therefore it would always just return the current system time when it was called. To fix this regression, you either still need to pass in `flux.SetNowOption(now)` to `flux.EvalAST`, or you need `system.time()` to check its context for an updated `now` time.

adrian-thurston: @jlapacik I see now, `SetNowOption` was used not just for computing `now` after script evaluation but also in the implementation of `system.time` and I needed to update that as well.

I actually had searched the code for places where Now was pulled from the environment and evaluated, but I didn't include the standard lib in my search.

adrian-thurston: @jlapacik So the call to `SetNowOption` was overriding the

`option now = system.time` 

in universe.flux?

jlapacik: Yes.

PR #2770, opened 2020-04-21
fix(compiler): apply substitution fully when compiling lambda

jlapacik:
Fixes an issue where `map` would drop a column if the first value in that column was null. The following query exhibited the bug:
```
... |> pivot(rowKey: ["_time"], columnKey: ["_field"], valueColumn: "_value")
    |> map(fn: (r) => ({_time: r._time, _value: r.a / r.b}))
```
Flux cannot infer the type of any expression following pivot. It waits till it gets an input record and calculates the output type from there. The type substitution that was constructed to do this was not applied fully to all types in map's lambda function, resulting in an unknown type which was subsequently dropped by `map`. The type substitution is now being applied fully to all types.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -204,15 +204,23 @@ func findProperty(name string, t semantic.MonoType) (*semantic.RowProperty, bool
 
 // apply applies a substitution to a type.
 // It will ignore any errors when reading a type.
-// This is safe becase we already validated that the function type is a mono type.
+// This is safe becase we already validated that the function type is a monotype.
 func apply(sub map[uint64]semantic.MonoType, props []semantic.PropertyType, t semantic.MonoType) semantic.MonoType {
 	switch t.Kind() {
+	case semantic.Unknown, semantic.Basic:
+		// Basic types do not contain type variables.
+		// As a result there is nothing to substitute.
+		return t
 	case semantic.Var:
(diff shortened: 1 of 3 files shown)


compiler/compiler.go
```diff
@@ -202,14 +202,62 @@ func findProperty(name string, t semantic.MonoType) (*semantic.RowProperty, bool
...
-// monoType ignores any errors when reading the type of a node.
+// apply applies a subsitution to a type.
+// It will ignore any errors when reading a type.
 // This is safe becase we already validated that the function type is a mono type.
-func monoType(subst map[uint64]semantic.MonoType, t semantic.MonoType) semantic.MonoType {
-	tv, err := t.VarNum()
-	if err != nil {
-		return t
+func apply(sub map[uint64]semantic.MonoType, props []semantic.PropertyType, t semantic.MonoType) semantic.MonoType {
+	switch t.Kind() {
```
rockstar, line 209: Is this `switch` statement exhaustive?

rockstar: I'm not sure if this is just common knowledge I should know, if it there should be a comment here to that effect. Thoughts?

jlapacik: No it's a legitimate question. And yes I'll add some clarifying comments.

PR #2773, opened 2020-04-23
fix(stdlib/system): system.time() should check context for override

jlapacik:
`system.time()` did not check its context to see if the `now` time was overridden. This meant tasks as well as any other external system could not override the default `now` time.

### DIFF
--- a/stdlib/system/time.go
+++ b/stdlib/system/time.go
@@ -5,21 +5,27 @@ import (
 	"time"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/lang/execdeps"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
 )
 
-var systemTimeFuncName = "time"
+var systemTimeFunc = values.NewFunction(
+	"time",
+	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Return: semantic.Time,
+	}),
+	func(ctx context.Context, args values.Object) (values.Value, error) {
+		if execdeps.HaveExecutionDependencies(ctx) {
+			if dep := execdeps.GetExecutionDependencies(ctx); dep.Now != nil {
+				return values.NewTime(values.ConvertTime(*dep.Now)), nil
+			}
+		}
+		return values.NewTime(values.ConvertTime(time.Now().UTC())), nil
+	},
+	false,
+)
 
 func init() {
-	flux.RegisterPackageValue("system", systemTimeFuncName, values.NewFunction(
-		systemTimeFuncName,
-		semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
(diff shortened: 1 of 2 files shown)


adrian-thurston:
Thank you for this important update to make #2759 correct.

stdlib/system/time.go
```diff
@@ -5,21 +5,27 @@ import (
...
-var systemTimeFuncName = "time"
+var systemTimeFunc = values.NewFunction(
+	"time",
+	semantic.NewFunctionPolyType(semantic.FunctionPolySignature{
+		Return: semantic.Time,
+	}),
+	func(ctx context.Context, args values.Object) (values.Value, error) {
+		if execdeps.HaveExecutionDependencies(ctx) {
+			if dep := execdeps.GetExecutionDependencies(ctx); dep.Now != nil {
+				return values.NewTime(values.ConvertTime(*dep.Now)), nil
```
wolffcm, line 21: I think I'm missing some context.  I've some questions.

`system.now()` and `universe.now()` are different, right? It's a little murky to me if this is the case and if so how they are different.

My understanding is that `option now = ...` overrides `universe.now()`.  So I guess the deps in the context will override the `system.now()`.  Where specifically is `now` specified in the execution dependencies?

jlapacik: `universe.now` just points to `system.now`

jlapacik: https://github.com/influxdata/flux/blob/7783b894c1b96b0b3d635cfb7bb770b626254eec/stdlib/universe/universe.flux#L9-L10

jlapacik: Note this update is only for overriding `now` via the flux API, not the flux language. When specifying `option now = ...` in a flux program, the default function value that is defined in this package, is replaced with whatever the user specifies.

wolffcm: Ah, ok.  I was confused. The dots are connected for me now. So we set `now` in the context when we process options in the Flux we receive from the client.

adrian-thurston: This has left me a bit confused ... In the commit that triggered this I'm setting now in the execution context from an evaluation of whatever the now option is set to. This was based off of code that followed the evaluation of a script. `now` was evaluated and passed to execution. This was too late for `tableFind` to use, so I moved it to exactly when the option was set.

If `now` defaults to `system.time` and `system.time` is pulling from the execution context we've got a circular dependency there. 

When I was working on this I was under the assumption that now is normally unset and the user may set it. That's why the `SetNowOption` code was there to handle the case that the user has not set it. But now I see it has a default in the standard library.

jlapacik: `system.time()` did not consider the execution context at all. Meaning it would always just default to the current system time. Now it checks the execution dependencies on the context. If `now` has been set, it uses that time, otherwise it defaults to the current system time.

PR #2774, opened 2020-04-23
feat: planner Pattern interface supplies a set of ProcedureKind as root

adrian-thurston:
Modified the planner Pattern interface to supply a set of ProcedureKind roots.
The pattern matcher adds the rule to multiple entries in the root-kind -> []rule
map.

The OneKindPattern was modfied to store a set of roots and was renamed to
UnionKindPattern. It is still used to implement plan.Pat and is now also used
to implement plan.UnionPat.

This change lets us create a pattern that is rooted at a set of ProcedureKind,
allowing us to make a single rule covering many window-aggregate combinations.

### DIFF
--- a/plan/pattern.go
+++ b/plan/pattern.go
@@ -5,7 +5,7 @@ const AnyKind = "*** any procedure kind ***"
 // Pattern represents an operator tree pattern
 // It can match itself against a query plan
 type Pattern interface {
-	Root() ProcedureKind
+	Roots() []ProcedureKind
 	Match(Node) bool
 }
 
@@ -20,8 +20,18 @@ type Pattern interface {
 //  /   \
 // A     B
 func Pat(kind ProcedureKind, predecessors ...Pattern) Pattern {
-	return &OneKindPattern{
-		kind:         kind,
+	return &UnionKindPattern{
+		kinds:        []ProcedureKind{kind},
+		predecessors: predecessors,
+	}
+}
+
+// OneOf matches any plan node from a given set of ProcedureKind and whose
+// predecessors match the given predecessor patterns. This is identical to Pat,
+// except for matching any pattern root from a set of ProcedureKinds.
+func OneOf(kinds []ProcedureKind, predecessors ...Pattern) Pattern {
+	return &UnionKindPattern{
+		kinds:        kinds,
 		predecessors: predecessors,
 	}
 }
@@ -39,8 +49,8 @@ type PhysicalOneKindPattern struct {
 	pattern Pattern
 }
 
-func (p PhysicalOneKindPattern) Root() ProcedureKind {
-	return p.pattern.Root()
+func (p PhysicalOneKindPattern) Roots() []ProcedureKind {
+	return p.pattern.Roots()
 }
 
 func (p PhysicalOneKindPattern) Match(node Node) bool {
@@ -53,22 +63,32 @@ func Any() Pattern {
 	return &AnyPattern{}
 }
 
-// OneKindPattern matches a specified procedure with a predecessor pattern
+// UnionKindPattern matches any one of a set of procedures that have a
+// specified predecessor pattern.
 //
-//                   ProcedureKind
+// For example, UnionKindPattern( { Proc1Kind, Proc2Kind }, { Pat1, Pat2 } )
+// will match either Proc1Kind { Pat1, Pat2 } or Proc2Kind { Pat1, Pat2 }
+//
+//                 [ ProcedureKind ]
 //                 /       |  ...  \
 //        pattern1     pattern2  ... patternK
-type OneKindPattern struct {
-	kind         ProcedureKind
+type UnionKindPattern struct {
+	kinds        []ProcedureKind
 	predecessors []Pattern
 }
 
-func (okp OneKindPattern) Root() ProcedureKind {
-	return okp.kind
+func (okp UnionKindPattern) Roots() []ProcedureKind {
+	return okp.kinds
 }
 
-func (okp OneKindPattern) Match(node Node) bool {
-	if node.Kind() != okp.kind {
+func (okp UnionKindPattern) Match(node Node) bool {
+	found := false
+	for _, kind := range okp.kinds {
+		if node.Kind() == kind {
+			found = true
+		}
+	}
+	if !found {
 		return false
 	}
 
@@ -95,8 +115,8 @@ func (okp OneKindPattern) Match(node Node) bool {
 // AnyPattern describes (and matches) any plan node
 type AnyPattern struct{}
 
-func (AnyPattern) Root() ProcedureKind {
-	return AnyKind
+func (AnyPattern) Roots() []ProcedureKind {
+	return []ProcedureKind{AnyKind}
 }
 
 func (AnyPattern) Match(node Node) bool {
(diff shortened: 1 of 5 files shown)


adrian-thurston:
You can find a rule that uses this here:

https://github.com/influxdata/influxdb/blob/adt-window-agg-rules/query/stdlib/influxdata/influxdb/rules.go#L602

adrian-thurston:
https://github.com/influxdata/idpe/issues/6806

plan/pattern.go
```diff
@@ -53,22 +63,28 @@ func Any() Pattern {
 	return &AnyPattern{}
 }
 
-// OneKindPattern matches a specified procedure with a predecessor pattern
+// UnionKindPattern matches a specified procedure with a predecessor pattern
 //
 //                   ProcedureKind
```
wolffcm, line 68: This comment could be modified to better describe how this pattern works.

jlapacik:
Looks good to me

plan/pattern.go
adrian-thurston: Yeah I have not worked on a planner pattern matcher before so I wanted to know if there might be a better approach.

It has some similarities to a tree pattern matcher I worked on in a network security setting. I also maintained a `map[pattern root kind]  ->  set of rules`. In that case both the patterns and the thing they were tested against were trees and there was only one pass over the trees. Not fixed point because it didn't transform, was just looking for stuff that happened.

adrian-thurston:
Improved the UnionKindPattern comment and added a test case.

plan/pattern.go
wolffcm: Nice. In our case we're actually working with a DAG, so you might've noticed that we expicitly guard against matching patterns to nodes that have more than once successor. E.g., we won't be able to push the filter into the source for this case:
```
data = from(...) |> range(...)
data |> filter(...very selective filter...) |> yield(...)
data |> filter(...another very selective filter...) |> yield(...)
```
However, at some point, it would be nice to be able optimize this to something like:
```
from(...range and very selective filter pushed down...) |> yield(...)
from(...range and another very selective filter pushed down...) |> yield(...)
```
But I think we would need to start modeling cost in the planner to do that effectively. (I'm just mentioning it here so it's on your radar.)

plan/pattern.go
```diff
@@ -20,8 +20,18 @@ type Pattern interface {
...
+	return &UnionKindPattern{
+		kinds:        []ProcedureKind{kind},
+		predecessors: predecessors,
+	}
+}
+
+// Union pat matches any plan node from a given set of ProcedureKind and whose
+// predecessors match the given predecessor patterns. This is idencal to Pat,
+// except for matching any pattern root from a set of ProcedureKinds.
+func UnionPat(kinds []ProcedureKind, predecessors ...Pattern) Pattern {
```
jsternberg, line 32: Nitpick: I like the name `OneOf` better. So I would prefer something like:

```
func OneOf(kinds ...ProcedureKind) Pattern
```

I also don't think you need `predecessors` here. You can do:

```
plan.Pat(universe.FilterKind, plan.OneOf(universe.MaxKind, universe.MinKind), universe.WindowKind)
```

This way we don't have to expect that all patterns will be recursive. There's one pattern for a pattern chain and the other patterns are for specific nodes.

plan/pattern.go
```diff
@@ -20,8 +20,18 @@ type Pattern interface {
...
-	return &OneKindPattern{
-		kind:         kind,
+	return &UnionKindPattern{
+		kinds:        []ProcedureKind{kind},
+		predecessors: predecessors,
+	}
+}
+
+// Union pat matches any plan node from a given set of ProcedureKind and whose
+// predecessors match the given predecessor patterns. This is idencal to Pat,
```
jsternberg, line 30: Typo: idencal. Also, comments should start with the exact name of the function so this should be `UnionPat` instead  of `Union pat`.

plan/pattern.go
```diff
@@ -5,7 +5,7 @@ const AnyKind = "*** any procedure kind ***"
 // Pattern represents an operator tree pattern
 // It can match itself against a query plan
 type Pattern interface {
-	Root() ProcedureKind
+	Roots() []ProcedureKind
```
jsternberg, line 8: This is an interesting way to do this. I like it.

An alternative way to implement a multiple kind pattern was to have this return `AnyKind` and then have match perform an actual check to see if it matches.

plan/pattern.go
adrian-thurston: The `max`/`min` need to be at the root so it has to go the other way around.

plan/pattern.go
adrian-thurston: I was wondering about that ... is it feasible to "split apart" (for lack of a better expression) nodes such as `from` and `range` above when they are found to have multiple successors and it impedes a push down? I guess that's the optimization you mention. So what I'm really asking is ... can you just duplicate any node in the dag in order to satisfy a pushdown rule?

adrian-thurston: With these pattern matching systems, is it normal to just make whatever pattern matching machinery you need in order to accomplish your task? Or is there a standard set of pattern forms that are normally made?

plan/pattern.go
adrian-thurston: Using `OneOf` was actually my first thought because I noticed that elsewhere in the code, but I didn't want it to look too similar to `One` as the struct was already named `OneKindPattern`. Seemed like `OneOfKindPattern` might get missed as something distinct. But yeah, I can definitely change it to keep language consistent.

plan/pattern.go
adrian-thurston: I considered that approach as well as you mentioned it in a standup call the other day, but thought the more that happens in the pattern matcher, IE, the more that is declarative in nature, the better in the long run.

plan/pattern.go
jsternberg: Do you think it would be better to update the `plan.Pat` function to take in a pattern?

There are a lot of ways this can go, but yours might be the best way to change it in a minimal way. I'm going to click approve. I do like the name `OneOf` for a function name, but we can keep `UnionKindPattern` as I really like that name for the type itself.

adrian-thurston: It does though, no?

jsternberg: What I meant was as the first parameter. Right now it takes a `ProcedureKind`.

I was just musing to myself. I think this is a good change to the planner that is iterative and useful.

plan/pattern.go
wolffcm: > can you just duplicate any node in the dag in order to satisfy a pushdown rule?

As long as the semantics are preserved (e.g., the yield nodes produce the same result and we're not duplicating side-effecting nodes like `to`) and the new DAG is better by some metric, memory, CPU, etc, then I don't see why not. 

> With these pattern matching systems, is it normal to just make whatever pattern matching machinery you need in order to accomplish your task?

My sense is yes, that's normal.  I have only worked with pattern matchers that work on trees, so taking into account both predecessors and successors is new territory to me. I wonder if there is research on it.

The first paper (on the Volcano Optimizer) in this chapter of the Red Book might provide some useful background on query planning, if you're interested:
http://www.redbook.io/ch7-queryoptimization.html

plan/pattern.go
adrian-thurston: Oh I see, yeah I'd have to think about that more.

PR #2777, opened 2020-04-24
fix(universe): map panics when overwriting group column

jlapacik:
Fixes a panic in `map` and merges the master branch in `feat/use-algo-w`


stdlib/universe/map.go
```diff
@@ -314,12 +314,16 @@ func groupKeyForObject(i int, cr flux.ColReader, obj values.Object, on map[strin
...
 		v, ok := obj.Get(c.Label)
 		if ok {
 			vs = append(vs, v)
+			cols = append(cols, flux.ColMeta{
+				Label: c.Label,
+				Type:  flux.ColumnType(v.Type()),
+			})
 		} else {
 			vs = append(vs, execute.ValueForRow(cr, i, j))
+			cols = append(cols, c)
```
wolffcm, line 326: My understanding is that this branch is dead code, unless `mergeKey` is true.  `on` would not contain the label otherwise.  Is that consistent with your understanding?

stdlib/universe/map_test.go
```diff
@@ -123,6 +123,45 @@ func TestMap_Process(t *testing.T) {
...
+				},
+				Data: [][]interface{}{
+					{"a", execute.Time(1), 1.0},
+					{"a", execute.Time(2), 2.0},
+					{"a", execute.Time(3), 3.0},
+					{"a", execute.Time(4), 4.0},
+					{"a", execute.Time(5), 5.0},
+				},
+			}},
+		},
```
wolffcm, line 164: Do we need a similar testcase for when `mergeKey` is true?

stdlib/universe/map.go
wolffcm: Wait, in the `mergeKey` case, where the map `fn` is dropping that column, wouldn't we want to omit that column from the group key as well?

jlapacik: Yes that is my understanding. When this function is called, a map of the group key columns are passed in. This is the `on` parameter. My understanding is that only in the case of `mergeKey=true` is it possible that one of the group key columns is not a label on `obj`. However while that label is missing from `obj`, it will be included in the output, so we want to include it in the group key.

wolffcm: Right, that makes sense.

stdlib/universe/map_test.go
jlapacik: Ah, yes I saw test cases for `mergeKey=true`, but I guess it's possible to have `mergeKey=true` and still overwrite it in the lambda. Good catch. I'll add one.

And in this particular case, because the group column is being overwritten, it will not be missing from `obj` and therefore the `if` branch from above will be taken.

jlapacik: Ok test case for `mergeKey=true` added.

PR #2778, opened 2020-04-24
fix(stdlib): run go generate on libflux when go generate is run on stdlib

jsternberg:
A change to the standard library generally involves recompiling libflux.
Instead of requiring the developer to remember that, this changes the
`go generate` step for stdlib to trigger a generate for libflux to force
a rebuild.

This hack will be needed until we consolidate the standard library
either into only Go or only Rust.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jlapacik:
I can't wait till we can do that consolidation.

PR #2792, opened 2020-04-29
fix: copy all fields of WindowProcedureSpec in Copy()

adrian-thurston:
Complete the Copy() function of WindowProcedureSpec. This is necessary for the 
PushDownWindowAggregate rewrite rule tests to pass.


adrian-thurston:
> I wish there was a way to ensure that `Copy` methods are complete in CI, I've seen bugs happen because of omissions like this a couple of times.

We could make some tests using the existing infrastructure? I would guess we just need some plans covering all the different node types, then call copy, then diff. Maybe covering all the nodes is the hard part.

PR #2802, opened 2020-05-04
fix(stdlib): fix concurrent map write in filter transformation

jsternberg:
The filter transformation reused the same function between multiple
table process calls. This shared state meant that multiple tables could
be processed at the same time and use the same state when processing
rows within the tables.

The filter transformation should be fixed to use distinct state for each
table. There's no reason why they need to share state and it is not for
performance. It is an oversight from the original design which doesn't
process tables concurrently. But, that code is more complicated than the
process code so it is easier for us to prevent multiple tables from
being processed in filter. This prevents the race condition by
preventing multiple tables from being processed at the same time.

This also addresses a similar problem in the `internal/gen` package.
The table generation would reuse the same random number generator
between tables and the random number generator isn't thread safe and, if
we added a lock to it to make it thread safe, it wouldn't be consistent
when used with the same seed. We use the same technique to prevent more
than one table from being created at the same time.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/gen/input.go
+++ b/internal/gen/input.go
@@ -82,7 +82,7 @@ type Schema struct {
 
 // Input constructs a TableIterator with randomly generated
 // data according to the Schema.
-func Input(schema Schema) (flux.TableIterator, error) {
+func Input(ctx context.Context, schema Schema) (flux.TableIterator, error) {
 	tags := schema.Tags
 
 	var seed int64
@@ -144,7 +144,9 @@ func Input(schema Schema) (flux.TableIterator, error) {
 	if alloc == nil {
 		alloc = &memory.Allocator{}
 	}
+
 	g := &dataGenerator{
+		Context:   ctx,
 		Period:    values.ConvertDuration(period),
 		NumPoints: numPoints,
 		Nulls:     schema.Nulls,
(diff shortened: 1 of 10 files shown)


internal/gen/input.go
```diff
@@ -425,6 +432,7 @@ func (dg *dataGenerator) Do(f func(tbl flux.Table) error) error {
 			if err := f(tbl); err != nil {
 				return err
 			}
+			<-done
```
wolffcm, line 435: This won't allow the user to cancel a long-running query, right? Is it a lot of trouble to fix it?

stdlib/universe/filter.go
```diff
@@ -194,14 +194,20 @@ func (t *filterTransformation) Process(id execute.DatasetID, tbl flux.Table) err
...
 		// Drop the table.
 		return nil
 	}
-	return t.d.Process(table)
+	if err := t.d.Process(table); err != nil {
+		return err
+	}
+	// Wait until the full table has been processed and read by the next transformation.
+	// Filter can only process one table at a time.
+	<-done
```
wolffcm, line 209: I've the same question as above here.  Don't we want to see if the context was canceled?

internal/gen/input.go
jsternberg: I forgot about that. I'll update this.

PR #2803, opened 2020-05-05
test(planner): toggle window aggregate pushdowns

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jlapacik:
Yeah that's why I brought up adding a pipeline test. A pipeline test will tell is if the operation is getting pushed down or not. I guess this is just an extra measure of safety to ensure we get the same result in both flux and influxdb.

Issue #152, opened 2018-10-19
Implement shorthand time literals


nathanielc:
The SPEC states that `2018-01-01` should parse as midnight on that date. And that `2009-10-15T09:00:00` should parse using the default location.

Currently we require both the time and the time offset to be specified.

aanthony1243:
check with @nathanielc to see if we need to implement both rust + golang parsers or just rust

nathanielc:
This was closed in error

nathanielc:
This was implemented, short date without a time can be parsed assuming a time of UTC midnight.

PR #2804, opened 2020-05-06
refactor: handle parsing errors with BadExpressions over panics

bthesorceror:


bthesorceror:
@adrian-thurston I've added a test around this to illuminate where we were seeing the panic in the LSP.

adrian-thurston:
@bthesorceror Okay thanks. At first I thought this was a regression in the rust parser because the scanner allows for a time literal with no timezone information on the end. But it looks like the go parser does the same ... allows the missing timezone in the scanner and fails during the RFC3339 parse. Perhaps it is to get more detailed error information, other than "scanner failed".

nathanielc:
Yes, the SPEC states that we support those shorter formats https://github.com/influxdata/flux/blob/master/docs/SPEC.md#date-and-time-literals but we have not implemented them yet. See https://github.com/influxdata/flux/issues/152 although it seems that the scanner supports those formats, just not the parser.

So for now we can just improve the error message to say something like shorthand time literals are not implemented yet.

adrian-thurston:
Ah okay, thank you for the background @nathanielc

PR #2805, opened 2020-05-06
feat(stdlib/v1): add fieldKeys and measurementFieldKeys to v1 package

sanderson:
Introduces `fieldKeys()` and `measurementFieldKeys()` into the v1 package for querying lists of fields in a bucket/measurement. These provide graceful equivalents to `SHOW FIELD KEYS` in InfluxQL.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/v1/v1.flux
+++ b/stdlib/influxdata/influxdb/v1/v1.flux
@@ -11,21 +11,6 @@ fieldsAsCols = (tables=<-) =>
     tables
         |> pivot(rowKey:["_time"], columnKey: ["_field"], valueColumn: "_value")
 
-// FieldKeys returns the list of field keys in a given bucket.
-// The return value is always a single table with a single column, "_value".
-fieldKeys = (bucket, predicate=(r) => true, start=-30d) =>
-    from(bucket: bucket)
-      |> range(start: start)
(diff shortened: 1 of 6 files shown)


stdlib/influxdata/influxdb/v1/v1.flux
```diff
@@ -11,6 +11,21 @@ fieldsAsCols = (tables=<-) =>
...
 
+// FieldKeys returns the list of field keys in a given bucket.
+// The return value is always a single table with a single column, "_value".
+fieldKeys = (bucket, predicate=(r) => true, start=-30d) =>
+    from(bucket: bucket)
+      |> range(start: start)
+      |> filter(fn: predicate)
+      |> keep(columns: ["_field"])
+      |> group()
+      |> distinct(column: "_field")
```
jlapacik, line 22: This is just a special case of `tagValues`, right? In other words, could you just define it like this:
```
fieldKeys = (bucket, predicate=(r) => true, start=-30d) =>
    tagValues(bucket: bucket, tag: "_field", predicate: predicate, start: start)
```

sanderson: Good call. Updated.

PR #2836, opened 2020-05-11
chore: update publish workflow to ensure clean publish

bthesorceror:
- fix version number
- remove old build before rebuilding

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,10 +144,13 @@ libflux-go:
 libflux-wasm:
 	cd libflux/src/flux && CC=clang AR=llvm-ar wasm-pack build --scope influxdata --dev
 
-build-wasm:
-	cd libflux/src/flux && CC=clang AR=llvm-ar wasm-pack build --scope influxdata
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -144,7 +144,10 @@ libflux-go:
 libflux-wasm:
 	cd libflux/src/flux && CC=clang AR=llvm-ar wasm-pack build --scope influxdata --dev
 
-build-wasm:
+clean-wasm:
+	rm -rf libflux/src/flux/pkg
+
+build-wasm: clean-wasm
```
wolffcm, line 150: I wonder if the `clean-wasm` should be part of the `publish-wasm` target instead of the `build-wasm` target? That way if you were developing locally and iterating, you would not have to do a clean build each time. I don't know if that's something that anyone does though.

Minor note, it would be good to add these `-wasm` targets to the `.PHONY` target at the end of this Makefile to indicate that they are not files, they're just here to invoke these specific commands.

PR #2838, opened 2020-05-12
fix(stdlib): fix race condition in the filter function

jsternberg:
The filter function had a race condition when processing multiple tables
with multiple buffers. The `RowPredicateFn` kept state internal after
preparing the function which wasn't necessary. This state got overridden
in future calls to filter, but the previous filter was still being
processed in a separate goroutine.

The original fix was to wait for the next transformation to consume the
filter before processing the next filter. Unfortunately, this collided
with the dispatcher which, for simple queries with one root, would only
allocate a single goroutine for executing the query. This meant that
since the `filter()` call never yielded, the downstream transformation
never read the filter which caused a deadlock since filter was waiting
for the downstream transformation.

This fix reverts that one and fixes the real underlying problem which
was the race condition in the row functions. It separates the row
functions into the dynamic function component and the prepared function
and it constructs a new prepared function for each table so there is no
shared state other than the immutable state used to construct the
dynamic function.

For testing, this commit adds a new internal package called
`internal/debug`. It is meant as a landing zone for any transformations
that are useful for debugging queries with some specific behavior. In
this case, the `debug.pass()` function will immediately pass the result
of its parent to the next transformation. This is useful in this case
because the process still takes a slot in the dispatcher for executing,
which was the reason why the version of filter that waited would
deadlock, and it is also useful for breaking planner rules when you
might not want to explicitly disable the planner rule.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
The `internal/debug` package was building off of an idea for a package of debugging utilities that was in this commit: https://github.com/influxdata/flux/commit/6a8ca44a6cca503e6814ee15867ac9339b5b9df5.

I got feedback from @nathanielc that the name "hack" was too general and we arrived on debug as a better name.

stdlib/universe/filter.go
```diff
@@ -345,15 +319,6 @@ func (t *filterTransformation) filter(cr flux.ColReader, record values.Object, i
...
 
-func (t *filterTransformation) wait(ctx context.Context, done <-chan struct{}) error {
-	select {
-	case <-done:
-		return nil
-	case <-ctx.Done():
-		return ctx.Err()
-	}
-}
-
```
fchikwekwe, line 356: I'm still not clear why waiting for a downstream transformation would cause a deadlock: 

> the dispatcher which, for simple queries with one root, would only
allocate a single goroutine for executing the query. This meant that
since the filter() call never yielded...

Was this for cases where there is no transformation after the call to `filter()`?

fchikwekwe, line 356: I'm still not clear why waiting for a downstream transformation would cause a deadlock: 

> ...the dispatcher which, for simple queries with one root, would only
allocate a single goroutine for executing the query. This meant that
since the filter() call never yielded...

Was this for cases where there is no transformation after the call to `filter()`?

execute/row_fn.go
```diff
@@ -286,29 +290,35 @@ func (f *RowMapFn) Eval(ctx context.Context, row int, cr flux.ColReader) (values
...
+	dynamicFn
 }
 
-func NewRowReduceFn(fn *semantic.FunctionExpression, scope compiler.Scope) (*RowReduceFn, error) {
-	r, err := newRowFn(fn, scope)
+func NewRowReduceFn(fn *semantic.FunctionExpression, scope compiler.Scope) *RowReduceFn {
+	return &RowReduceFn{
+		dynamicFn: newDynamicFn(fn, scope),
+	}
+}
```
fchikwekwe, line 300: I'm guessing that not propagating the error here has something to do with no longer keeping the additional state on `dynamicFn`, but I'm having a hard time understanding what. Also, does this just no longer error?

stdlib/universe/state_tracking.go
```diff
@@ -186,14 +183,13 @@ func (t *stateTrackingTransformation) Process(id execute.DatasetID, tbl flux.Tab
...
-	err := execute.AddTableCols(tbl, builder)
-	if err != nil {
+	if err := execute.AddTableCols(tbl, builder); err != nil {
 		return err
 	}
 
 	// Prepare the functions for the column types.
 	cols := tbl.Cols()
-	err = t.fn.Prepare(cols)
+	fn, err := t.fn.Prepare(cols)
```
fchikwekwe, line 192: Oh, I think I see now. You've moved the error collection to `prepare` and taken it off `dynamic`.

stdlib/universe/filter.go
jsternberg: Waiting for the downstream transformation ended up causing a problem because of the following flow.

1. During planning, the planner decided the concurrency of the query was 1. That meant only one transformation could run at a time. I didn't understand the code completely, but I think it decided it was 1 because there was only one root (the yield call). I don't know if the planner refers to the yield as the root or the from as the root, but there was only one of each of those so it set the concurrency to 1.
2. The transformation started. Filter processed a multi-buffer table. It told the downstream transformation to process a table and then waited for it to be processed.
3. Filter held the only open goroutine in the dispatcher so the next transformation never ran. That meant the table was never read and so filter never released the goroutine. I was aware this could happen, but I misread the code and thought the concurrency was always set to 10. There's a number called throughput that is set to 10, but it has nothing to do with the number of goroutines.
4. `filter |> yield` was not affected because yield isn't a real transformation. It creates a result that is returned to the user. That code is not controlled by the dispatcher so it didn't matter that there was no open goroutine.

The changes that were made were:

1. Properly fix the filter function so it is thread safe.
2. Filter could go back to the old code that did not wait for the table to finish being processed. This frees up the goroutine again so that the next transformation can be run.
3. Add `debug.pass()` to the test. This forces an additional transformation that does nothing. I used `debug.pass()` to debug the push down filter and found the same thing useful for this test to ensure that the downstream transformation wasn't blocked by filter. Since `debug.pass()` is a real transformation, it takes up space in the dispatcher and ensures that any other transformation would be able to run too.

execute/row_fn.go
jsternberg: When I was refactoring this I found all of these functions never errored for some reason. If I was going to make a breaking change I may as well fix the signatures at the same time.

stdlib/universe/state_tracking.go
fchikwekwe: Oh, nvm. You're just no longer returning the state from the `dynamicFn` since you're not keeping it there anymore. Forgive me. Big diff.

nathanielc:
The code change looks good to me but I have a higher level question. This is the second time we have changed this section of code to fix a bug. Both of those bugs landed in production before we caught them. Do you have a sense for why we missed those bugs? What testing do we need to ensure we are not regressing those bugs?

Specifically we had a concurrent map panic and a deadlock in the filter code. Do we have tests cases for those two bugs explicitly? I know writing test cases for race condition bugs can be tricky, any thoughts on how we can ensure we got it correct this time?

jsternberg:
> The code change looks good to me but I have a higher level question. This is the second time we have changed this section of code to fix a bug. Both of those bugs landed in production before we caught them. Do you have a sense for why we missed those bugs? What testing do we need to ensure we are not regressing those bugs?

I think so. In one circumstance it is because we don't test multiple buffer transformations very frequently or in a systemic way. Some transformations have this test, but others don't. Another reason is because we don't test very many queries beyond end to end tests, which execute flux through the dispatcher which is where both of these bugs originated from. At the moment, end to end tests are all small datasets that are written in raw csv and so they don't exercise the multiple buffer behavior that caused both of these problems.

> Specifically we had a concurrent map panic and a deadlock in the filter code. Do we have tests cases for those two bugs explicitly? I know writing test cases for race condition bugs can be tricky, any thoughts on how we can ensure we got it correct this time?

There is a specific test case that addresses both problems that happened. If I run the test case with the original code, it fails when used with `-race`. If I run the test with the deadlock, it never completes. But, this isn't a systemic fix and won't catch other related errors in other transformations.

nathanielc:
Thanks for the clarification

>  At the moment, end to end tests are all small datasets that are written in raw csv and so they don't exercise the multiple buffer behavior that caused both of these problems.

Sounds like we need to put together some way to systematically run these tests with multiple buffers. I'll create an issue to explore that further.

One last question, should we feature flag this fix? I am being paranoid over here but want to ask the question anyway so we can have a conversation around it.

jsternberg:
> One last question, should we feature flag this fix? I am being paranoid over here but want to ask the question anyway so we can have a conversation around it.

I am not of the opinion that we need to do that. We have three versions of filter (including this PR). The first potentially causes an unrecoverable panic. The second causes an almost guaranteed deadlock. I am not sure how this third one can be worse than either of those. If the first was a recoverable panic, I might say we should revert to that one and then feature flag the fix. Since it is an unrecoverable panic, I don't think we can revert to that version.

PR #2843, opened 2020-05-13
test: added end-to-end planner tests for group-count and group-sum

adrian-thurston:
Closes influxdata/idpe#7104

### DIFF
--- a/stdlib/planner/group_count_eval_test.flux
+++ b/stdlib/planner/group_count_eval_test.flux
@@ -1,0 +1,58 @@
+package planner_test
+
+import "testing"
+import "planner"
+
+option planner.disablePhysicalRules = ["PushDownGroupAggregateRule"]
+
+option now = () => (2030-01-01T00:00:00Z)
+
+input = "
+#datatype,string,long,dateTime:RFC3339,string,string,string,double
+#group,false,false,false,true,true,true,false
+#default,_result,,,,,,
+,result,table,_time,_measurement,host,_field,_value
+,,0,2018-05-22T19:53:26Z,system,hostA,load1,1.83
+,,0,2018-05-22T19:53:36Z,system,hostA,load1,1.72
+,,0,2018-05-22T19:53:37Z,system,hostA,load1,1.77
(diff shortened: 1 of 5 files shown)


stdlib/planner/group_count_eval_test.flux
```diff
@@ -0,0 +1,55 @@
...
+,,1,2018-05-22T19:53:56Z,system,hostB,load3,1.96
+,,1,2018-05-22T19:54:06Z,system,hostB,load3,1.98
+,,1,2018-05-22T19:54:16Z,system,hostB,load3,1.97
+
+,,2,2018-05-22T19:53:26Z,system,hostC,load5,1.95
+,,2,2018-05-22T19:53:36Z,system,hostC,load5,1.92
+,,2,2018-05-22T19:53:41Z,system,hostC,load5,1.91
+,,2,2018-05-22T19:53:46Z,system,hostC,load5,1.92
+,,2,2018-05-22T19:53:56Z,system,hostC,load5,1.89
+,,2,2018-05-22T19:54:16Z,system,hostC,load5,1.93
```
wolffcm, line 32: It would be good to change up the tags across the same host here, to show that the grouping+aggregating is happening correctly.  For example, you could change the last three of these rows above to `load1` (and also bump the table number).

wolffcm: (ditto for the other _test.flux files)

adrian-thurston:
> I guess that we still don't understand the issue with tests unexpectedly passing in `influxdb`, but that you're going to merge this and figure that part out later?

@wolffcm whatever lack of test case failure after editing the expected output appears to be independent of test case and is related to concurrency and yield. Working on it now.

stdlib/planner/group_count_eval_test.flux
adrian-thurston: @wolffcm made these suggested changes, thanks

wolffcm:
Actually, this won't start failing until the push down rule is enabled, so I'll just merge this.

Issue #2808, opened 2020-05-07
Extra argument in function call produces incorrect message


wolffcm:
This is a case of an error message that is wrong and misleading.
```
import "json"
import "http"
import "slack"
slack.message(
  url: "https://hooks.slack.com/services/...", 
  username: "a", 
  token: "",
  channel: "team", 
  workspace: "", 
  text: "sent from data explorer", 
  iconEmoji: ":smile:", 
  color: "good"
)
```
produces the message
```
type error @5:1-5:238: missing required argument iconEmoji
```

`iconEmoji` is not a required argument, it's not an argument at all.  The message should be 
```
iconEmoji is not an argument to function slack.message
```
or something like that. [Definition of slack.message](https://github.com/influxdata/flux/blob/master/stdlib/slack/slack.flux#L16)

A simple example:
```
> string(v: 10.0, foo: "bar")
Error: type error @1:1-1:28: missing required argument foo
```
Note that if you don't have all the required arguments, the error message is correct:
```
> string(foo: "bar")
Error: type error @1:1-1:19: missing required argument v
```

PR #2850, opened 2020-05-18
refactor: constraints now have an expected type

jlapacik:
Fixes https://github.com/influxdata/flux/issues/2808.

Previously if a function was not passed the required arguments it was expecting an error would be presented of the form:
```
missing required argument ...
```

Constraints now have an expected and an actual type which means we can now distinguish between a function that was passed too few arguments vs too many.

### DIFF
--- a/libflux/src/core/semantic/nodes.rs
+++ b/libflux/src/core/semantic/nodes.rs
@@ -428,15 +428,12 @@ impl OptionStmt {
                 let (env, cons) = stmt.init.infer(env, f)?;
                 let (env, rest) = stmt.member.infer(env, f)?;
 
-                let l = stmt.member.typ.clone();
-                let r = stmt.init.type_of();
-
                 Ok((
                     env,
                     cons + rest
                         + vec![Constraint::Equal {
-                            exp: l,
-                            act: r,
+                            exp: stmt.member.typ.clone(),
+                            act: stmt.init.type_of(),
                             loc: stmt.init.loc().clone(),
                         }]
                         .into(),
--- a/libflux/src/core/semantic/nodes.rs
+++ b/libflux/src/core/semantic/nodes.rs
@@ -33,7 +33,7 @@ use std::vec::Vec;
 // updated type environment and a set of type constraints to be solved.
 pub type Result = std::result::Result<(Environment, Constraints), Error>;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Error {
     Inference(infer::Error),
     UndefinedBuiltin(String, ast::SourceLocation),
@@ -433,7 +433,13 @@ impl OptionStmt {
 
                 Ok((
                     env,
-                    cons + rest + vec![Constraint::Equal(l, r, self.loc.clone())].into(),
+                    cons + rest
+                        + vec![Constraint::Equal {
+                            exp: l,
+                            act: r,
+                            loc: stmt.init.loc().clone(),
+                        }]
+                        .into(),
                 ))
             }
             Assignment::Variable(stmt) => stmt.infer(env, f),
@@ -629,11 +635,11 @@ impl StringExpr {
             if let StringExprPart::Interpolated(ref mut ip) = p {
                 let (e, cons) = ip.expression.infer(env, f)?;
                 constraints.append(&mut Vec::from(cons));
-                constraints.push(Constraint::Equal(
-                    ip.expression.type_of(),
-                    MonoType::String,
-                    ip.expression.loc().clone(),
-                ));
+                constraints.push(Constraint::Equal {
+                    exp: MonoType::String,
+                    act: ip.expression.type_of(),
+                    loc: ip.expression.loc().clone(),
+                });
                 env = e
             }
         }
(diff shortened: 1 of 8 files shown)


libflux/src/core/semantic/types.rs
```diff
@@ -153,16 +155,20 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Error::CannotUnify(t, with) => write!(f, "{} != {}", t, with),
-            Error::CannotConstrain(t, with) => write!(f, "{} is not {}", t, with),
+            Error::CannotUnify(expected, found) => write!(f, "{} != {}", found, expected),
```
nathanielc, line 158: Should we change this error message too?

Instead of saying `int != string` is there better way to communicate what is going on now that we know which type is expected and which one was found?

For example something like `found type 'int' but was expecting 'string'`?

I know its slightly unrelated to the issue being fixed but it all we have to change is this one line know I think we should add it.

libflux/src/core/semantic/types.rs
```diff
@@ -153,16 +155,20 @@ pub enum Error {
...
+            Error::CannotConstrain(found, expected) => write!(f, "{} is not {}", found, expected),
```
nathanielc, line 159: Should we do a similar thing here? Saying `type is not type` is the same as `type != type`. Both communicate a factual statement but do not help the user fix what is broken. 

What is the difference between CannotUnify and CannotConstrain? Is there extra context there that is helpful to the user that we should put in the error text?

libflux/src/core/semantic/types.rs
```diff
@@ -153,16 +155,20 @@ pub enum Error {
...
+            Error::ExtraLabel(a) => write!(f, "found unexpected label {}", a),
             Error::CannotUnifyLabel(a, t, with) => {
                 write!(f, "{} != {} (record label {})", t, with, a)
             }
             Error::MissingArgument(x) => write!(f, "missing required argument {}", x),
+            Error::ExtraArgument(x) => write!(f, "found unexpected argument {}", x),
             Error::CannotUnifyArgument(x, e) => write!(f, "{} (argument {})", e, x),
-            Error::CannotUnifyReturn(t, with) => write!(f, "return type {} != {}", t, with),
+            Error::CannotUnifyReturn(expected, found) => {
+                write!(f, "return type {} != {}", found, expected)
```
nathanielc, line 170: What about here? Should we replace the `!=` with a sentence explanation?

libflux/src/core/semantic/nodes.rs
```diff
@@ -433,7 +433,13 @@ impl OptionStmt {
 
                 Ok((
                     env,
-                    cons + rest + vec![Constraint::Equal(l, r, self.loc.clone())].into(),
+                    cons + rest
+                        + vec![Constraint::Equal {
+                            exp: l,
+                            act: r,
```
nathanielc, line 439: nit, should `l` and `r` be renamed to `exp` and `act` respectively?

libflux/src/core/semantic/nodes.rs
```diff
@@ -941,195 +955,275 @@ impl BinaryExpr {
...
-                    self.left.type_of(),
-                    self.right.type_of(),
-                    self.right.loc().clone(),
-                ),
-                Constraint::Equal(self.left.type_of(), self.typ.clone(), self.loc.clone()),
-                Constraint::Kind(self.typ.clone(), Kind::Addable, self.loc.clone()),
+                Constraint::Equal {
+                    exp: self.left.type_of(),
+                    act: self.right.type_of(),
+                    loc: self.right.loc().clone(),
```
nathanielc, line 961: What makes the left hand of a binary expression expected and the right hand side the actual type?

libflux/src/core/semantic/types.rs
jlapacik: I think this one can stay as it is. This error is returned when a particular kind is expected like `Addable` but a type that is not `Addable` is found instead.

libflux/src/core/semantic/types.rs
jlapacik: Yes identifying which is the expected type and which is the actual type is helpful.

libflux/src/core/semantic/nodes.rs
jlapacik: That decision was arbitrary. As the left hand side is seen first, it seems natural that the left hand type is what was expected.

libflux/src/core/semantic/types.rs
nathanielc: That makes sense, I found this example from a test case:

```
type error @2:14-2:17: string is not Negatable
```

That does read well, thanks for clarifying.

libflux/src/core/semantic/nodes.rs
nathanielc: ok sounds good.

libflux/src/core/semantic/types.rs
jlapacik: How does `expected int but found string` read?

nathanielc: I like it, short and simple. Hopefully with the context of where the error occurs it will lead users in the right direction.

PR #2857, opened 2020-05-22
Feat/telegram

sranka:
This PR adds **telegram** package to flux so that notifications can be sent from flux to [Telegram](https://telegram.org). This PR is a first step in influxdata/influxdb#17899

Notes:
- the implementation and tests follow the concepts in existing "slack" package
- telegram flux offers the same options as in https://docs.influxdata.com/kapacitor/v1.5/event_handlers/telegram/#sidebar
- flux's telegram package will be then used influxdata/influxdb , as the slack implementation does

### Done checklist
- [ ] Test cases written


sranka:
I tested the telegram package using `flux repl`. I used the following steps to setup a test environment:
1. Create a new account an app downloaded from https://telegram.org , a phone number is required. 
1. Create a new bot: https://core.telegram.org/bots#creating-a-new-bot , you will receive an authorization **token** for the API during the registration process.
1. Create a new channel from the application
1. Add the new bot to a channel as an Administrator, the only required permission is to post messages: https://stackoverflow.com/questions/33126743/how-do-i-add-my-bot-to-a-channel
1. Get id of the **channel** to send messages to
   * send a message to the bot in the channel `hello @YOURbot`
   * `curl https://api.telegram.org/bot$token/getUpdates` and look for the `id` the channel

Message sending is described in https://core.telegram.org/bots/api#sendmessage

```bash
curl -X POST  -H 'Content-Type: application/json' \
-d '{"chat_id": "'$channel'", "text": "This is a test"}' \
https://api.telegram.org/bot$token/sendMessage
```

Markdown can be used in the message text, see https://core.telegram.org/bots/api#formatting-options.

There are a few advanced parameters in message sending that have default values. They are documented in stdlib/telegram/telegram.flux

rockstar:
Could you please move this module to use our new "contrib" layout? You can see how to do that [here](https://github.com/influxdata/flux/tree/master/stdlib/contrib).

sranka:
@rockstar so it is now refactored to a new *contrib* layout + I also added a README.md to the new package.

PR #2861, opened 2020-05-22
feat(table): add table function to convert Flux values into a table

nathanielc:
This function can convert a list of records into a table that the execution
engine can process.

This is done by capturing the list of records into a source and building a table
from the records when the source executes.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
I modified the commit in this PR for #3077 and recreated it in #3095. It moves this work to the `experimental/array` package and renames the function to `from(rows:)` because then you can say "array from rows" out loud.

It also modifies the source to use the arrow builders directly and fixes the issue with the source sending tables to multiple transformations.

nathanielc:
Closing in favor of #3095

PR #2867, opened 2020-05-26
feat(query): optimize fill()

### LINKED ISSUE #2823: Optimize memory usage for fill function
The `fill` function needs to have its memory usage optimized in the following ways.

1. [Reuse the columns](https://gist.github.com/jsternberg/c20e6b900baab8863c6693e633dc425f#-arrow-array-references) that aren't referred to by the `column` parameter. The number of rows doesn't change and those columns don't change so we can reuse the arrays from the input.
2. [Stream each input table](https://gist.github.com/jsternberg/c20e6b900baab8863c6693e633dc425f#-stream-tables) to the next transformation because the group key is 1:1.
3. For the column that gets filled, [preallocate the array](https://gist.github.com/jsternberg/c20e6b900baab8863c6693e633dc425f#-preallocate-array-size) before constructing it.
4. Only perform the fill if null values exist in the array. Null values can be checked without iterating over the array by checking the `NullN()` function on the [array](https://godoc.org/github.com/apache/arrow/go/arrow/array#Interface). The array can be retrieved by using [table.Values](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#Values).

### LINKED ISSUE #2828: Switch fill transformation to use the ArrowBuilder
The fill transformation should be switched to use the [ArrowBuilder](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#ArrowBuilder). This is the first step in #2823.

At the moment, the transformation uses the [TableBuilderCache](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/execute?tab=doc#TableBuilderCache) along with [execute.Dataset](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/execute?tab=doc#NewDataset) to provide the builder. This implementation is limited to only use the [ColListTableBuilder](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/execute?tab=doc#ColListTableBuilder) which has slow memory access patterns.

This issue is to convert the fill transformation to use this alternative [Dataset](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#NewDataset), use this alternative [BuilderCache](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#BuilderCache), and use the [ArrowBuilder](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#ArrowBuilder) and the [GetArrowBuilder](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#GetArrowBuilder) utility to retrieve the builder.

```go
// construct transformation and dataset
alloc := &memory.Allocator{}
cache := &table.BuilderCache{
    New: func(key flux.GroupKey) table.Builder {
        return table.NewArrowBuilder(key, alloc)
    },
}
dataset := table.NewDataset(id, cache)

// process individual table
var tbl flux.Table = /* incoming table */
b, created := table.GetArrowBuilder(tbl.Key(), cache)
// insert schema creation code
// use builders from the arrow builder to append values
```

### LINKED ISSUE #2829: Switch the fill transformation to use a passthrough dataset
After #2828 is completed, the transformation should be switched to use [PassthroughDataset](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/execute?tab=doc#PassthroughDataset) instead of [table.NewDataset](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#NewDataset). This will also result in removing the `BuilderCache`. The transformation will still use the `ArrowBuilder` to construct the `flux.Table` and it will pass that table to the `PassthroughDataset` to be sent to the next transformation immediately.

```go
dataset := execute.NewPassthroughDataset(id)
mem := &memory.Allocator{}

// process individual table
var tbl flux.Table = /* incoming table */
b := table.NewArrowBuilder(tbl.Key(), mem)
// insert schema creation code
// use builders from the arrow builder to append values
outTable, err := b.Table()
if err != nil {
    return err
}
return dataset.Process(outTable)
```

### LINKED ISSUE #2830: Switch the fill transformation to use table stream
After #2829 is completed, the fill transformation should be updated to use [table.Stream](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#Stream) to produce the table. It can still construct a single buffer, but the call should be changed from using the [ArrowBuilder](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#ArrowBuilder) to constructing the array builders, constructing the arrays, and then writing them into the [StreamWriter](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#StreamWriter) with [Write](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#StreamWriter.Write).

### LINKED ISSUE #2831: Switch fill transformation to construct a single buffer for each incoming buffer
After #2830 is done, the streamed buffer should be switched to call [Write](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/internal/execute/table?tab=doc#StreamWriter.Write) once for each incoming buffer read from the table. Inside of the call to [Do](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0?tab=doc#Table), new builders should be [constructed for each column](https://pkg.go.dev/github.com/influxdata/flux@v0.67.0/arrow?tab=doc#NewBuilder), [resized to the length of the buffer](https://github.com/apache/arrow/blob/master/go/arrow/array/builder.go#L60), and then have the [values copied into the array](https://github.com/apache/arrow/blob/master/go/arrow/array/builder.go#L65).

```go
dataset := execute.NewPassthroughDataset(id)
mem := &memory.Allocator{}

// process individual table
var tbl flux.Table = /* incoming table */
// insert schema creation code
cols := []flux.ColMeta{...}
outTable, err := table.Stream(tbl.Key(), cols, func(ctx context.Context, w *table.StreamWriter) error {
    // construct builders for the incoming schema
    builders := make([]array.Builder, len(w.Cols()))
    for i, col := range w.Cols() {
        builders[i] = arrow.NewBuilder(col.Type, mem)
    }
    // use builders to append values
    // construct arrays for each column
    arrs := make([]array.Interface, len(builders))
    for i, b := range builders {
        arrs[i] = b.NewArray()
    }
    return w.Write(arrs)
})
if err != nil {
    return err
}
return dataset.Process(outTable)
```

### LINKED ISSUE #2832: Switch fill transformation to reuse arrow arrays
After #2831 is completed, the stream code should be rewritten to reuse the incoming column arrays for columns that aren't modified and, if the column to be filled has no null values, reuse that column too. If the column to be filled has null values, then a new array should be built and used instead.

yzhang1991:
Closes #2823 
Closes #2828 
Closes #2829 
Closes #2830 
Closes #2831 
Closes #2832

Benchmark (n=1M, 6 tags, null%=40%):
```
Before:
886823652 ns/op	  951035324 B/op    1600214 allocs/op

After:
17892792 ns/op    10047908 B/op	    9788 allocs/op

Reduce%:
97.98%            98.94%            99.39%
```

Comparison benchmark:
 - feat/optimize-fill vs. feat/optimize-fill-compare
https://github.com/influxdata/flux/blob/9dda2e3dc7ca8e3c49a4fa9d806e66d73dbd4019/stdlib/universe/fill_test.go#L456


nathanielc:
@yzhang1991 Any ideas on why the allocations improvement was small in comparison to the total time improvement? I would have expected a much larger gain in memory saved as we as one of the major changes was to reuse arrow buffers.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
-				return err
-			}
-		}
-		// Set new value
-		l := cr.Len()
 
-		if l > 0 {
-			if t.spec.UsePrevious {
-				prevNonNull = execute.ValueForRow(cr, 0, idx)
+	table, err := table.StreamWithContext(t.ctx, key, tbl.Cols(), func(ctx context.Context, w *table.StreamWriter) error {
```
fchikwekwe, line 224: Just curious, where was the table constructed before? Wondering if the code in this anonymous function was moved from somewhere.

stdlib/universe/fill.go
```diff
@@ -167,32 +172,37 @@ func createFillTransformation(id execute.DatasetID, mode execute.AccumulationMod
...
+	t, d := NewFillTransformation(a.Context(), s, id, a.Allocator())
 	return t, d, nil
 }
 
 type fillTransformation struct {
-	d     execute.Dataset
-	cache execute.TableBuilderCache
-
-	spec *FillProcedureSpec
+	d     *execute.PassthroughDataset
```
jsternberg, line 180: 👍 This is what I would expect for the fill transformation to change.

stdlib/universe/fill.go
```diff
@@ -167,32 +172,37 @@ func createFillTransformation(id execute.DatasetID, mode execute.AccumulationMod
...
+	ctx   context.Context
+	spec  *FillProcedureSpec
+	alloc *memory.Allocator
 }
 
-func NewFillTransformation(d execute.Dataset, cache execute.TableBuilderCache, spec *FillProcedureSpec) *fillTransformation {
-	return &fillTransformation{
-		d:     d,
-		cache: cache,
+func NewFillTransformation(ctx context.Context, spec *FillProcedureSpec, id execute.DatasetID, alloc *memory.Allocator) (*fillTransformation, execute.Dataset) {
```
jsternberg, line 186: Can you update this signature so it returns `(execute.Transformation, execute.Dataset)`? I prefer not to have public methods that expose private structs.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 		}
 	}
 
-	builder, created := t.cache.TableBuilder(key)
```
jsternberg, line 207: 👍 This is good. The table builder and cache are not necessary.

stdlib/universe/fill.go
jsternberg, line 224: This looks good since the table schema and key shouldn't change at all. No need to allocate memory to reconstruct them.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
-
-			for i := 0; i < l; i++ {
-				v := execute.ValueForRow(cr, i, idx)
-				if v.IsNull() {
-					if err := builder.AppendValue(idx, prevNonNull); err != nil {
-						return err
-					}
+			vs := make([]array.Interface, len(w.Cols()))
+			for j, col := range w.Cols() {
+				if j != idx {
```
jsternberg, line 232: It seems like the original code was structured this way, but we should change it. This is a hanging else clause and so the code should be changed to:

```
if j != idx {
    vs[j] = table.Values(cr, j)
    vs[j].Retain()
    continue
}

// rest of the code
```

yzhang1991:
@nathanielc I am going to tweak the benchmark and add more columns then see how the number changes. Currently, there are just two tags, I think the benefit of reusing unaffected columns' buffers won't be that obvious with this few columns.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
-		}
-		// Set new value
-		l := cr.Len()
 
-		if l > 0 {
-			if t.spec.UsePrevious {
-				prevNonNull = execute.ValueForRow(cr, 0, idx)
+	table, err := table.StreamWithContext(t.ctx, key, tbl.Cols(), func(ctx context.Context, w *table.StreamWriter) error {
+		return tbl.Do(func(cr flux.ColReader) error {
+			l := cr.Len()
```
jsternberg, line 226: Can you encapsulate the logic of this inner function into another function? A lot of our older transformations do a lot of nesting and they were always difficult for me to read for that reason. While we're going through and updating them, we should also focus on refactoring these areas into functions. Maybe something like:

```
func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader, prev *values.Value) {
    // implementation
}
```

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
 				} else {
-					if err := builder.AppendValue(idx, v); err != nil {
-						return err
-					}
 					if t.spec.UsePrevious {
-						prevNonNull = v
+						prevNonNull = execute.ValueForRow(cr, 0, idx)
+					}
+					b := arrow.NewBuilder(col.Type, t.alloc)
+					b.Reserve(l)
```
jsternberg, line 240: The `Reserve` method is a performance trap in Arrow for whatever reason. Use `Resize` instead. This is because `Reserve` will allocate an array that fits `l` elements and it will overallocate to the nearest multiple of 8. This is fine, but it combines with a rather negative behavior when `NewArray()` is called. When `NewArray()` is called, it will attempt to resize the allocated array to the smallest necessary space to fit the array. This results in a call to `Reallocate`. In C implementations, this probably results in a call to `realloc` which is probably quick when making the array size smaller and probably helps with keeping memory performant. In Go, there is no access to `realloc` from pure-Go so the reallocate is implemented in allocating a new array and copying the values which creates further churn on the garbage collector.

In the Go implementation of Arrow, if you know the exact size of the array, always use `Resize()`. There's not a lot of good reasons to use `Reserve()` in the Go implementation.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
+					hasNull := false
+					for i := 0; i < l; i++ {
+						v := execute.ValueForRow(cr, i, idx)
```
jsternberg, line 243: You should avoid `execute.ValueForRow`. It is slow because it calls into an interface, performs a checked type cast, then it will allocate memory to place the retrieve value into an interface to return it which always requires an allocation. Since you are processing a single column, you can write a separate implementation for each column type.

The `filter()` transformation has a good example (https://github.com/influxdata/flux/blob/master/internal/arrowutil/filter.gen.go.tmpl) and so does `pivot()` (https://github.com/influxdata/flux/blob/master/stdlib/universe/pivot.gen.go.tmpl).

For this circumstance specifically, `pivot()` is the better example to mimic.

stdlib/universe/fill.go
```diff
@@ -204,59 +214,63 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
+						} else {
+							if err := arrow.AppendValue(b, v); err != nil {
+								return err
+							}
+							if t.spec.UsePrevious {
+								prevNonNull = v
+							}
+						}
+					}
+					if hasNull {
```
jsternberg, line 258: Instead of iterating to figure out if there are any null values and taking the performance hit of allocating memory, you can perform this check above by using `NullN()` (https://github.com/apache/arrow/blob/master/go/arrow/array/array.go#L33).

```go
arr := table.Values(cr, j)
if arr.NullN() == 0 {
    vs[j] = arr
    vs[j].Retain()
    continue
}
```

yzhang1991:
@nathanielc I have updated the benchmark result

nathanielc:
@yzhang1991 Benchmarks looks good. Make sense that is shows up more when we have more columns.

stdlib/universe/fill.gen.go
```diff
@@ -0,0 +1,221 @@
...
+	case flux.TTime:
+		return t.fillTimeColumn(arr.(*array.Int64))
+
+	default:
+		panic(fmt.Errorf("unsupported array data type: %s", arr.DataType()))
+	}
+}
+
+func (t *fillTransformation) fillIntColumn(arr *array.Int64) array.Interface {
+	var prevValue int64
```
jsternberg, line 38: I think this won't work correctly with multiple buffers. We had this problem with derivative. The previous value has to be kept outside of the `Do` loop so it will need to be passed in.

I'd suggest changing the structure to this:

```
func (t *fillTransformation) fillColumn(typ flux.ColType, arr array.Interface, v *interface{}) {
    // previous code
}

func (t *fillTransformation) fillIntColumn(arr *array.Int64, v *interface{}) array.Interface {
    var prevValue int64
    prevNull := *v == nil
    if !prevNull {
        prevValue = (*v).(int64)
    }
    if t.spec.UsePrevious {
        // Save the previous value for this buffer to use in future buffers.
        defer func() {
            if !prevNull { *v = prevValue }
        }()
    }
}

// not valid code but a rough idea of what it would look like
func (t *fillTransformation) Process(tbl flux.Table) error {
    var fillValue interface{}
    if !t.spec.Value.IsNull() {
        fillValue = /* extract the value from the value */
    }
    tbl.Do(func(cr flux.ColReader) error {
        t.fillColumn(typ, arr, &fillValue)
    })
}
```

I prefer to avoid `values.Value` if possible for this lower level code. I don't think there's any difference if you replaced `interface{}` with `values.Value` so feel free to replace the `interface{}` uses with `values.Value` if you find that easier.

stdlib/universe/fill.gen.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+func (t *fillTransformation) fillIntColumn(arr *array.Int64, fillValue *interface{}) array.Interface {
+	fillValueNull := *fillValue == nil
+	var fillValueInt int64
+	if !fillValueNull {
+		fillValueInt = (*fillValue).(int64)
+	} else {
+		if fillValueNull = arr.IsNull(0); !fillValueNull {
```
jsternberg, line 44: I am not sure there's a circumstance where this is correct. The only thing I can think of is when `UsePrevious` is true and this value gets set lower. I think we can skip this here.

values/values.go
```diff
@@ -148,6 +148,74 @@ var (
 	Null = null{}
 )
 
+func Unwrap(v Value) (interface{}, error) {
```
jsternberg, line 151: Can you remove the error from this and just document that this will panic when called with something that doesn't unwrap? We use that pattern in a lot of other areas so it shouldn't be a problem here.

internal/types.tmpldata
```diff
@@ -12,6 +12,7 @@
     "IsNumeric": true,
     "IsComparable": true,
     "Value": "Value",
+    "FluxValue": "Int",
```
jsternberg, line 15: These entries are no longer needed.

stdlib/universe/fill_test.go
```diff
@@ -405,6 +409,40 @@ func TestFill_Process(t *testing.T) {
...
 			}},
 		},
+		{
+			name: "fill previous multiple buffers",
+			spec: &universe.FillProcedureSpec{
+				DefaultCost: plan.DefaultCost{},
+				Column:      "_value",
+				UsePrevious: true,
+			},
+			data: []flux.Table{&executetest.RowWiseTable{
```
jsternberg, line 419: 👍

stdlib/universe/fill.gen.go
yzhang1991: This is for handling the case when we have not encountered a non-null value in the table yet and the first value in the current buffer is still `null`. Think of an extreme case when all the rows in the specified table are `null` for the designated column.

yzhang1991: I don't know what "set lower" really means.

jsternberg: Later in the function the fill value gets set when it encounters a non-null value. So this code doesn't seem to add anything and this code would only be valid for the `UsePrevious` case anyway.

values/values.go
```diff
@@ -148,6 +148,52 @@ var (
 	Null = null{}
 )
 
+// Extract the primitive value from the Value interface.
+// This will panic when called on some Values that cannot be unwrapped.
```
jsternberg, line 152: I don't see the panic in the code implementation. Can you also specify in the comment which values cannot be unwrapped for the documentation?

PR #2894, opened 2020-06-09
fix(libflux): include a hash of the sources for libflux

jsternberg:
This changes the libflux generate process to generate a hash of the
sources so that libflux is forced to rebuild.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
I'll merge it tomorrow so we have time to talk about it first.

The intention behind this is to encode through the code what sources are expected. The previous method was a hack to avoid doing the work that was involved here. The previous method was to write a file that changed the timestamp every time so that the code would be forced to recompile. This method works for local builds because we can control when we call `go generate` and it forces a rebuild of the cgo package.

This doesn't work for downstream dependencies. Downstream dependencies cannot modify the sources that they use. In the case of a project that uses a vendor directory, you would need a physical change to the source code to trigger a compilation so it relinks the sources. That means the file that gets modified by calling `go generate` needs to be committed to the repository. In order to avoid constantly committing generated files, you would need a way to consistently generate the same file. File hashes of the sources are used so that a consistent output is created.

This means that when we modify libflux in Rust, it will force a change to the Go package that interfaces with it by cgo and force a recompilation. In the context of vendor, it means that when you do `go mod vendor` and the libflux rust sources change, there will be a Go file that the Go build system will look at to trigger a rebuild. And our existing system of checking the generated files will ensure that file is constantly updated.

The reason for including hashes for all of the files rather than a final hash is so that it is easier to debug. If we find ourselves not getting an expected rebuild, we can physically see if the file that was modified is missing from the sources hash rather than guessing if it was included in the hash or not.

PR #2903, opened 2020-06-11
fix(compiler): fix the compiler's return type when with is used

jsternberg:
When one of the arguments for the compiler function was a record and the
record was used with the `with` operator, the compiler function would
learn that the return type extended a record, but not what the record
was.

This fixes the compiler so if the extended record is one of the
arguments, the type that was passed into the compiler is included in the
substitutions for extended records. This allows type inference to be
more accurate in the compiler with a return value like this:

    map(fn: (r) => ({r with _value: r._value * 2.0}))

If the record contains a `_time` column, that previously would not have
been included in the resolved return value. This prevented functions
like `map` from relying on the return value from type inference.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -146,15 +146,19 @@ func substituteTypes(subst map[uint64]semantic.MonoType, inType, in semantic.Mon
 			return err
 		}
 
+		names := make([]string, 0, nproperties)
 		for i := 0; i < nproperties; i++ {
 			lprop, err := inType.RowProperty(i)
 			if err != nil {
 				return err
 			}
 
+			// Record the name of the property in the input type.
+			name := lprop.Name()
+			names = append(names, name)
+
 			// Find the property in the real type if it
 			// exists. If it doesn't exist, then no problem!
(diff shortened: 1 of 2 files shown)


jsternberg:
Map's not currently using this behavior. I actually found this when I was experimenting with a contrib package idea about a month ago. I was planning to try getting those functions merged but I wanted to separate out the bug fixes from the contrib packages so that they weren't tied up with each other.

I figure that we'll want this behavior to work though even outside of my specific need for a contrib package.

The contrib function was an alternative implementation of map and I didn't want to just leave it on my machine forever but, when I worked on it, it was around the time we were trying to get algorithm w merged and I didn't want to add additional work to getting that merged to master.

jsternberg:
Here's a comment in the current `map` function that notes the decision not to use type inference for the algorithm w refactor: https://github.com/influxdata/flux/blob/be8c67e907a2d96c7f9ced8c7660fdffeb152fcc/stdlib/universe/map.go#L146

PR #2906, opened 2020-06-12
feat(contrib/jsternberg): add an alternative aggregate package to contrib

jsternberg:
This contains an alternative aggregate table function that is more
friendly to pivoted data. The method takes a mapping of output columns
and the input column along with the aggregate. All of the aggregates are
defined in pure flux accompanied by some new math functions that operate
directly on array values. As an example:

    import "influxdata/influxdb/v1"
    import "contrib/jsternberg/aggregate"

    from(bucket: "telegraf")
        |> range(start: -5m)
        |> filter(fn: (r) => r._measurement == "cpu" and (r._field == "usage_user" or r._field == "usage_system"))
        |> v1.fieldsAsCols()
        |> aggregate.table(columns: {
            "min_usage_user": aggregate.min(column: "usage_user"),
            "max_usage_user": aggregate.max(column: "usage_user"),
            "mean_usage_user": aggregate.mean(column: "usage_user"),
            "min_usage_system": aggregate.min(column: "usage_system"),
            "max_usage_system": aggregate.max(column: "usage_system"),
            "mean_usage_system": aggregate.mean(column: "usage_system"),
        })

As can be seen, the aggregate will read the data once and then
efficiently compute multiple columns and aggregates using the schema
given. Defining a new aggregate can be done in pure flux. For example,
the `min` function is defined like this:

    import "contrib/jsternberg/math"

    min = (column) => ({
        // Determines which column we operate on.
        column: column,
        // Initializes the state with the first buffer of values (as an array).
        init: (values) => {
            v = math.min(values)
            return v
        },
        // Takes in the previous state and the next buffer of values and returns
        // a new state.
        reduce: (values, state) => {
            v = math.min(values)
            return math.min(values: [state, v])
        },
        // Returns the final result. In some examples, the final state can
        // only be computed after reading all of the data. In others,
        // like min, the state is the final value.
        compute: (state) => state,
    })

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


compiler/compiler.go
```diff
@@ -53,6 +53,8 @@ func Compile(scope Scope, f *semantic.FunctionExpression, in semantic.MonoType)
 			if err := substituteTypes(subst, argT, mtyp); err != nil {
 				return nil, err
 			}
+		} else if !arg.Optional() {
```
jsternberg, line 56: I found this omission when testing this because the complicated nature of the input for columns means that type inference can't really analyze the input. This is because the input for columns is an arbitrary object and the keys aren't statically defined.

This doesn't seem to affect any current code because all of our current functions can have their function specified in advance.

internal/execute/table/stream.go
```diff
@@ -183,7 +183,7 @@ func (s *streamTable) Do(f func(flux.ColReader) error) error {
 	if err := f(s.first); err != nil {
 		s.first.Release()
 		s.first = nil
-		return nil
+		return err
```
jsternberg, line 186: This seems to be a big omission. If the first buffer reports an error, stream seems to just discard it.

PR #2908, opened 2020-06-15
feat(libflux): adding Timeable constraint to time and duration types

ChitlangeSahas:
### Done checklist
- [x] Timeable kind constraint has been added
- [x] Time and Duration satisfy the constraint
- [x] Units tests for the Timeable constraint and Time/Duration types
- [ ] Does NOT include changing any existing functions to use Timeable, those will be separate issues.


build.sh
```diff
@@ -0,0 +1,5 @@
+export PKG_CONFIG=/Users/sahaschitlange/go/bin/pkg-config
```
jsternberg, line 1: This looks like a local script to build flux. Can you remove this? I think you can do `make build` to do the same thing.

test.flux
```diff
@@ -0,0 +1,2 @@
+import "date"
```
jsternberg, line 1: This is a good start but we have a method for writing this kind of test. You might want to look at `stdlib/date/minute_test.flux` as an example for how to write a unit test like this one. Any files that end in `_test.flux` and have a `test` statement inside of them will be run after doing `go generate ./stdlib; go test ./stdlib`.

libflux/src/core/semantic/types.rs
```diff
@@ -432,21 +434,26 @@ impl MonoType {
                 }),
             },
             MonoType::Duration => match with {
-                Kind::Comparable | Kind::Equatable | Kind::Nullable | Kind::Negatable => {
-                    Ok(Substitution::empty())
-                }
+                Kind::Comparable
```
jsternberg, line 437: @jlapacik would be a better person to look at this, but it looks correct to me.

stdlib/date/date.go
```diff
@@ -287,13 +287,28 @@ func init() {
 					return nil, errors.New(codes.Invalid, "missing argument unit")
 				}
 
-				if v.Type().Nature() == semantic.Time && u.Type().Nature() == semantic.Duration {
-					w, err := execute.NewWindow(u.Duration(), u.Duration(), execute.Duration{})
-					if err != nil {
-						return nil, err
+				//print(v.Type())
+				if v.Type().Nature() == semantic.Time || v.Type().Nature() == semantic.Duration && u.Type().Nature() == semantic.Duration {
```
jsternberg, line 291: It might be good if you add this kind of logic to the `values` package. It's a bit difficult to read here. Something like:

```go
package values

func IsTimeable(v Value) bool {
    return v.Type().Nature() == semantic.Time || v.Type().Nature() == semantic.Duration
}
```

I mostly recommend this because I had a difficult time reading this logical condition. It occurred to me that this is testing that the value is a time or duration and the unit is a duration.

build.sh
ChitlangeSahas: Done!

libflux/src/core/semantic/builtins.rs
```diff
@@ -49,7 +49,7 @@ pub fn builtins() -> Builtins<'static> {
                  "millisecond" => "forall [] (t: time) -> int",
                  "microsecond" => "forall [] (t: time) -> int",
                  "nanosecond" => "forall [] (t: time) -> int",
-                 "truncate" => "forall [] (t: time, unit: duration) -> time",
+                 "truncate" => "forall [t0] where t0 : Timeable (t: t0, unit: duration) -> time", // sahas
```
jlapacik, line 52: Can you remove the terminating comment?

libflux/src/core/semantic/types.rs
jlapacik: This is correct.

libflux/src/core/semantic/builtins.rs
jlapacik: Also is there a reason you updated just truncate? I thought the type signatures would be updated all at once along with the implementations.

ChitlangeSahas: @jlapacik I was not sure if I would be able to implement the Kind constraint so I just tried one function to see if it would. Next step of mine is to apply the change to all functions where it is applied.

About the comment, I will remove it. I wanted a flag to search for everywhere I made a change to so I can go back and revert if necessary

nathanielc: @ChitlangeSahas I'd like to see this work done in at least two parts. 1. A change that adds timeable as a new constraints with test etc. 2. A change that adds it to date.truncate and possible the other date functions. I'd even be ok with each date function getting its own PR so we can review smaller changes at a time.

So for this PR I would focus on getting the unit tests in place for the Timeable constraint. 

Does that make sense?

ChitlangeSahas: Yeah for sure, I will change the signature back to what it was.

stdlib/date/date.go
```diff
@@ -287,13 +287,26 @@ func init() {
 					return nil, errors.New(codes.Invalid, "missing argument unit")
 				}
 
-				if v.Type().Nature() == semantic.Time && u.Type().Nature() == semantic.Duration {
-					w, err := execute.NewWindow(u.Duration(), u.Duration(), execute.Duration{})
-					if err != nil {
-						return nil, err
+				if values.IsTimeable(v) && u.Type().Nature() == semantic.Duration {
```
jsternberg, line 290: I think my understanding was that we're moving this section into a separate PR. Is that correct?

test.flux
jsternberg, line 1: This file needs to be removed.

values/values.go
```diff
@@ -294,6 +294,11 @@ func CheckKind(got, exp semantic.Nature) {
 	}
 }
 
+// check if value is Timeable
+func IsTimeable(v Value) bool {
```
jsternberg, line 298: I think this should be part of a different PR, but for general knowledge, it is a Go standard to start all documentation comments with `// NameOfFunction` as it forces the documenter to use the active voice and it makes it easier to grep for documentation comments with `git grep "// NameOfFunction"`.

stdlib/date/date.go
ChitlangeSahas: You are right, my bad. I will just replace that file with the current version from master branch.

test.flux
ChitlangeSahas: Done.

PR #2925, opened 2020-06-18
feat(stdlib/sql): add BigQuery support

alespour:
This PR add BigQuery support to `sql` package.

* It uses forked driver [bonitoo-io/go-sql-bigquery](https://github.com/bonitoo-io/go-sql-bigquery) because the original was incomplete/ready for integration and seems kind of unmaintaned or inactive at least
* The driver name is `bigquery` and uses URL format with `bigquery` scheme as [connection string](https://github.com/bonitoo-io/go-sql-bigquery#connection-string)

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/sql/bigquery.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	reader.InitColumnTypes(types)
+
+	return reader, nil
+}
+
+var fluxToBigQuery = map[flux.ColType]string{
+	flux.TFloat:  "FLOAT64",
+	flux.TInt:    "INT64",
+	flux.TUInt:   "INT64",
```
wolffcm, line 164: Not all uint64 values can be stored as a signed 64-bit number, maybe `NUMERIC` or something would be more appropriate here?

stdlib/sql/sql_test.go
```diff
@@ -322,6 +322,270 @@ func TestSQLiteParsing(t *testing.T) {
...
+			want:       [][]values.Value{{values.NewString("6")}, {values.NewString("1")}, {values.NewString("643")}, {values.NewString("42")}, {values.NewString("1283")}},
+		},
+		{
+			name:       "bools",
+			columnName: "_bools",
+			data:       mockRowsToSQLRows(sqlmock.NewRows([]string{"column"}).AddRow(bool(true)).AddRow(bool(false)).AddRow(bool(true)).AddRow(bool(false)).AddRow(bool(true))),
+			want:       [][]values.Value{{values.NewBool(true)}, {values.NewBool(false)}, {values.NewBool(true)}, {values.NewBool(false)}, {values.NewBool(true)}},
+		},
+		{
+			name:       "datetime",
```
wolffcm, line 425: This is actually the `timestamp` type and not `datetime`, or am I missing something?

go.mod
```diff
@@ -38,7 +40,7 @@ require (
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/pkg/term v0.0.0-20180730021639-bffc007b7fd5 // indirect
-	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
+	github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
```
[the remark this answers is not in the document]
alespour: It was not intentional, I do not know how to make `mod tidy` not update it... :(

go.mod
```diff
@@ -3,13 +3,15 @@ module github.com/influxdata/flux
...
-	cloud.google.com/go v0.43.0
+	cloud.google.com/go v0.57.0
+	cloud.google.com/go/bigtable v1.3.0
 	github.com/Azure/go-autorest/autorest v0.10.1
 	github.com/Azure/go-autorest/autorest/adal v0.8.3
 	github.com/Azure/go-autorest/autorest/azure/auth v0.4.2
 	github.com/DATA-DOG/go-sqlmock v1.4.1
 	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
 	github.com/apache/arrow/go/arrow v0.0.0-20191024131854-af6fa24be0db
+	github.com/bonitoo-io/go-sql-bigquery v0.3.4
```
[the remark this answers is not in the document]
alespour: Since I wasn't getting a response from the maintainer who seem to have abandoned the project, I gave up on hoping on the merge and continued with modifications in the `bonitoo-io` fork. The owner then commented later that he _ had totally forgotten about this project_ and merged the PR, but the fork has diverted more from the original into more pure SQL golang driver unlike the original which also contained some ORM stuff etc. The original project does not seem to be active (no commits since Oct 27, 2019) and maintained (eg. PR was merge despite failing check due to some missing 3rd party lib).

stdlib/sql/bigquery.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (m *BigQueryRowReader) GetNextRow() ([]values.Value, error) {
+	row := make([]values.Value, len(m.columns))
+	for i, column := range m.columns {
+		switch value := column.(type) {
+		case bool, int64, float64, string:
+			row[i] = values.New(value)
+		case civil.Date:
+			row[i] = values.NewString(value.String())
+		case civil.DateTime:
```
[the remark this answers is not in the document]
alespour: So far I have been mapping datetime types without TZ info to Flux strings, leaving formatting as is. You are right that values of such types are somewhat useless in Flux, but without timezone info, they cannot be RFC3339 formatted.  And I guess it be wrong to assume some timezone for datetime values we do not know what they represent, wouldn't it?

I can think of adding a proprietary connection string option (to all DBs supported by Flux that have such types without TZ) that would allow user to specify TZ to be used for these types, and when this would be set, we could map these types to either `time` or at least use RFC3399-formatted string.

go.mod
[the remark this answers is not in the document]
wolffcm: Ah, I guess that maybe one of the new dependencies needed a newer version of it then.

stdlib/sql/bigquery.go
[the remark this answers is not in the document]
wolffcm: Ok, I like the idea of a special connection string option. If it becomes an issue perhaps we can do that. Otherwise, I think this is fine for now.

stdlib/sql/bigquery.go
alespour: Yes `NUMERIC` is appropriate I think. This however requires that `bonitoo-io/go-sql-bigquery` driver also implements `driver.ColumnConverter` (and `driver.NamedValueChecker`) interface like eg. [MySQL driver does](https://github.com/go-sql-driver/mysql/blob/master/statement.go#L142) to support `uint64`  with high bit set, because default Go converter does not and throws `uint64 values with high bit set are not supported` error (https://golang.org/src/database/sql/driver/types.go#L273)

wolffcm: If it is not a lot of trouble, I would say that it's worth doing. It makes a lot more sense to me to map the type to something that can contain all possible values than to occasionally error at runtime for large unsigned numbers.

Wouldn't it be possible to create a `driver.ColumnConverter` that has maps a Flux `uint` column to `NUMERIC` but just calls `driver.DefaultParameterConverter` for the other types?

alespour: I'll have a look at it.

Please just note that it may be possible with BigQuery because we use own forked driver, but eg. with PostgreSQL or SQLServer, `sql.to` with large uint64 values will fail with `uint64 values with high bit set are not supported` error because they rely on default converter. Maybe that's why uint64 is mapped to BIGINT, INT etc for Flux sqlite, postgres etc support because it was considered overflow-safe by the author since it would fail with large uint values with the mentioned error anyway :)

wolffcm: Ok, I'll hold off on merging this to see what you find as far as supporting `uint64`. I think this is good to go otherwise.

alespour: It seems the most simple implementation is just adding a function like 
```
func (c *Conn) CheckNamedValue(nv *driver.NamedValue) (err error) {
	switch nv.Value.(type) {
	case uint64:
		// no conversion
	default:
		nv.Value, err = driver.DefaultParameterConverter.ConvertValue(nv.Value)
	}
	return err
}
```
and properly format the values for `uint64` during BigQuery call preparation (`CAST('...' AS NUMERIC)`). 
All this is in the `go-sql-bigquery` driver.

I would prefer that such specific handling of `uint64` would be controlled by connection string option (eg. `uint64AsNumeric=false|true`), and that it would **not be enabled by default in the driver**.
In Flux, on the other hand, `uint64AsNumeric=true` could be appended to the connection string by default unless  user explicitly specified `uint64AsNumeric=false`.
I've tested the above "no-conversion" code and seems to work (I haven't implemented the connection string support yet). 

The only "issue" is that, in Flux, outgoing mapping is `uint64` -> `NUMERIC` and incoming mapping is `NUMERIC` -> `float64`, so it is asymmetric...

What do you think?

nathanielc: I would prefer explicit and safe mapping between data types. If a BigQuery type cannot (or generic SQL type) cannot support all values that the corresponding Flux type supports I think we should not allow that mapping. 

In other words only map Flux types to SQL types where its possible to represent all values. Converting between types in Flux is simple, so if a user has `uint` data on the Flux side and are confident that its safe to convert to a int ,i.e no high bits set, then they can explicitly convert the column to an `int` column before calling `sql.to`

That said Flux type conversion are not overflow safe, meaning they will silently ignore bad conversions. We can create an issue to make that optional behavior. I would still prefer and explicit conversion over an implicit one on the driver. 

This way the user experience is the first time they try and write `uint` data to sql they get an error and learn they need to explicitly convert the type, giving them and opportunity to validate it is safe to do so. 

That is in contrast to an experience where with simple tests the query works but once they encounter real world data the query breaks with the error about the conversion not being valid.

nathanielc: If this direction means we should reconsider how Postgres and other mapping works then we should do that now as well.

wolffcm: @nathanielc By "now", do you mean in this PR?

I like the idea of disallow sending `uint` data to stores that don't have a corresponding data type. `uint` seems different enough from `NUMERIC` that we shouldn't make the conversion automatically.

However, I don't think we've put much thought into unifying how Flux handles SQL types, so it seems like we may have some breaking changes required. E.g., for Postgres has already supported conversion from Flux `uint` to `BIGINT` for quite a while, and so we will need to start returning an error for that case. Our `sql` package is not experimental, how should we handle breaking changes like that?

nathanielc: No inside this PR but we should file issues now and determine how to prioritize them. I think in this case we can classify the previous behavior as a bug and call out the behavior change specifically in the change log release announcements. 

We should first add the ability to throw an error on overflow for any numeric conversions before making these changes to existing behavior.

wolffcm: @nathanielc Ok, thanks for clarifying. I'll create two issues:
- A bug for the inconsistent behavior that we get when passing `uint` to various SQL stores
- Add ability to throw an error for numeric conversions that would overflow

@alespour For this PR, I think you can just make it so that passing a `uint` column to `sql.to()` where we're sending the data to a BigQuery store will result in an error, with a related test. With that done, I think this will be good to merge.

wolffcm: I have filed issues 
https://github.com/influxdata/flux/issues/3066
https://github.com/influxdata/flux/issues/3067

alespour: Go default `uint64` handling already does just that ie. error is thrown when `uint64` conversion would overflow `int64` (https://golang.org/src/database/sql/driver/types.go#L270):
```
	case reflect.Uint64:
		u64 := rv.Uint()
		if u64 >= 1<<63 {
			return nil, fmt.Errorf("uint64 values with high bit set are not supported")
		}
		return int64(u64), nil
```
Eg. in REPL, user gets `"Error: sql: transaction has already been committed or rolled back: sql: converting argument $11 type: uint64 values with high bit set are not supported"` with **for all databases** except for MySQL.

So is it necessary to have the same check duplicated in `sql.to`?

wolffcm: The difference (if I understand correctly) is that we want the error thrown if the user tries to insert any values at all from a `uint` column into a BigQuery store. Even if all the uint values are less than 2^63, we still want to throw the error. This way, a user developing an application and testing with only small values won't unexpectedly see their application fail in the real world with values larger than 2^63.

It seems to me that the current behavior is only to throw the error if there are uint values that are out of range, do I have that right?

alespour: Ah, I see, no implicit conversion at all. 
Then it is enough just to remove `TUint` mapping for BigQuery, because without this mapping, Flux will throw `BigQuery does not support column type uint` error when preparing CREATE TABLE statement which is always executed before any insert.

alespour: Done.

PR #2926, opened 2020-06-18
feat(contrib): add a faster map function

jsternberg:
This adds a faster map function that is more limited than the map
function in the universe package. This map function is unable to modify
the group key of a table that it processes and doesn't need to perform
regrouping.

The map function takes the same arguments and the same return value as
the universe equivalent. The person calling the function can either
return an object with only the value columns or they can return the
result with the group keys too. This is to allow the `with` operator to
still be used. If the function attempts to return a row that would
change one of the group keys, the `map` function will ignore the change.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/src/core/semantic/builtins.rs
```diff
@@ -872,6 +872,14 @@ pub fn builtins() -> Builtins<'static> {
                 "null" => r#"forall [t0] t0"#,
                 "none" => r#"forall [t0] t0"#,
             },
+            "contrib/jsternberg/rows" => semantic_map! {
+                "map" => r#"
+                    forall [t0, t1, t2, t3] where t0: Row, t1: Row, t2: Row, t3: Row (
+                        <-tables: [t0],
+                        fn: (r: t2) -> t3
+                    ) -> [t1]
+                "#,
```
jlapacik, line 881: This seems like too many type variables. Shouldn't the type signature be the same as the regular map function?

    forall [t0, t1] (<-tables: [t0], fn: (r: t0) -> t1) -> t1

jsternberg: I didn't think about that. Fixed.

jsternberg:
I'm going to merge this now so it ends up in the next release, but likely try and work on some benchmarks next week. The good thing about this function is I think I've eliminated most of the allocation or slowdown problems involving regrouping and inefficient allocations. But, this still performs pretty slowly in my opinion because of issues with executed functions. I think this might end up being a good function to use when working to speed up row functions and the runtime compiler in general.

PR #2927, opened 2020-06-18
fix(execute): check stream equality

### LINKED ISSUE #2893: Verify correctness of table streams regardless of table and column order
Table streams with equivalent data, but differing column and/or table orders should be considered equal. We currently have test helpers that do not correctly check that streams are the same regardless of column/table order.

This is possibly causing flakiness in tests across the flux repo and in our pipeline tests. 

The following functions need to be updated to do this. 

EqualResultIterators: https://github.com/influxdata/flux/blob/745073a0c2558ad023b01aba41172c87fd354c45/execute/executetest/result.go#L85
EqualResult: https://github.com/influxdata/flux/blob/745073a0c2558ad023b01aba41172c87fd354c45/execute/executetest/result.go#L128
NormalizeTables: https://github.com/influxdata/flux/blob/745073a0c2558ad023b01aba41172c87fd354c45/execute/executetest/table.go#L479

Since EqualResult and EqualResultIterators read the entire table stream into memory, we can just updated NormalizeTables to sort the result. 

NormalizeTables needs to be updated to call two functions: one that will sort tables by group key and another that will sort each table's columns in a consistent order.  

NOTE: This might not be efficient for tests that return a big result.

fchikwekwe:
Closes #2893. 

Previously table stream checked with `NormalizeTables()` were only registering as equal when both the tables and columns were in the exact same order. This PR will put tables and columns in a consistent order so that they can be properly checked for equality.

### DIFF
--- a/csv/result_test.go
+++ b/csv/result_test.go
@@ -30,198 +30,21 @@ type TestCase struct {
 	encoderConfig csv.ResultEncoderConfig
 }
 
-var symmetricalTestCases = []TestCase{
-	{
-		name:          "single table",
-		encoderConfig: csv.DefaultEncoderConfig(),
-		encoded: toCRLF(`#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,double
+func TestResultDecoder(t *testing.T) {
+	testCases := []TestCase{
+		{
+			name:          "single table",
+			encoderConfig: csv.DefaultEncoderConfig(),
+			encoded: toCRLF(`#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,double
 #group,false,false,true,true,false,true,true,false
 #default,_result,,,,,,,
 ,result,table,_start,_stop,_time,_measurement,host,_value
 ,,0,2018-04-17T00:00:00Z,2018-04-17T00:05:00Z,2018-04-17T00:00:00Z,cpu,A,42
 ,,0,2018-04-17T00:00:00Z,2018-04-17T00:05:00Z,2018-04-17T00:00:01Z,cpu,A,43
 `),
-		result: &executetest.Result{
-			Nm: "_result",
-			Tbls: []*executetest.Table{{
-				KeyCols: []string{"_start", "_stop", "_measurement", "host"},
-				ColMeta: []flux.ColMeta{
-					{Label: "_start", Type: flux.TTime},
-					{Label: "_stop", Type: flux.TTime},
-					{Label: "_time", Type: flux.TTime},
-					{Label: "_measurement", Type: flux.TString},
-					{Label: "host", Type: flux.TString},
-					{Label: "_value", Type: flux.TFloat},
-				},
-				Data: [][]interface{}{
-					{
-						values.ConvertTime(time.Date(2018, 4, 17, 0, 0, 0, 0, time.UTC)),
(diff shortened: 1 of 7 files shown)


execute/executetest/table.go
```diff
@@ -480,6 +481,88 @@ func NormalizeTables(bs []*Table) {
...
+func sortByGroupKey(tables []*Table) []*Table {
+	if len(tables) < 2 {
+		return tables
+	}
+
+	for _, table := range tables {
+		// somehow this line makes the test pass
+		// no idea why
+		_ = table.Key().String()
+	}
```
fchikwekwe, line 498: I'm not sure why, but removing this for loop causes `TestNormalizeTables()` to fail. Any ideas?

execute/executetest/table.go
```diff
@@ -480,6 +481,88 @@ func NormalizeTables(bs []*Table) {
 	for _, b := range bs {
 		b.Key()
 	}
+	bs = sortByGroupKey(bs)
+}
+
+// sortByGroupKey uses recursive quicksort to sort the tables by group key
+// in alphabetical order
```
fchikwekwe, line 488: Implemented two different versions of quick sort for this PR. I used recursive quick sort for table order, since the tables were easy to reorder independently. And I used iterative quick sort for column order, since columns need to be updated per table, it was easier to keep track of them using a stack.

csv/result_test.go
```diff
@@ -30,198 +30,21 @@ type TestCase struct {
 	encoderConfig csv.ResultEncoderConfig
 }
 
-var symmetricalTestCases = []TestCase{
```
fchikwekwe, line 33: Test cases were not being normalized from this array. Moved them into the respective test functions.

execute/executetest/table.go
```diff
@@ -475,11 +476,58 @@ func (b SortedTables) Swap(i int, j int) {
...
+	}
+}
+
+func sortColumns(table *Table) {
+	// index reference to make sure that sort is consistent for metadata and data
+	indexes := createReferenceArray(table)
+
+	columnsCopy := make([]flux.ColMeta, len(table.ColMeta))
+	for i, value := range indexes {
+		columnsCopy[i] = table.ColMeta[value]
```
rockstar, line 501: Does `copy` not do this for us?

execute/executetest/table.go
```diff
@@ -475,11 +476,58 @@ func (b SortedTables) Swap(i int, j int) {
...
+	}
+	table.ColMeta = columnsCopy
+
+	// don't create a new table.Data if there is no Data
+	if len(table.Data) == 0 {
+		return
+	}
+	valuesCopy := make([][]interface{}, len(table.Data))
```
rockstar, line 509: Do we need to make a full `valuesCopy`? Can we get away with just copying the row as we need it, since we're not re-ordering the rows themselves?

fchikwekwe: We still need the outer array to house the interface array.

execute/executetest/table.go
fchikwekwe: No, the copy doesn't populate the map with the required values. For some reason, the `copy` function doesn't play well with `flux.ColMeta` values.

rockstar: Could you add a comment to that effect? I can see myself seeing that one day and thinking "why is this like this?" and trying to simplify it. Luckily, we'll have a failing test as a result, but the _why_ would be super helpful.

fchikwekwe: Comment added.

PR #2929, opened 2020-06-21
feat(stdlib/date): update date.truncate to use Timeable

### LINKED ISSUE #2922: Update date.truncate to use Timeable
The `date.truncate` Flux function should be updated to use the `Timeable` constraint so that it can accept either time or durations values.

DOD
 - Update type signature of function
 - Update implementation to correctly handle duration values
 - Add tests for new duration behavior

ChitlangeSahas:
Update date.truncate to use Timeable and added logic to handle time
and duration types accordingly.

Done checklist
 - [x] Update type signature of function
 - [x] Update implementation to correctly handle duration values
 - [x] Add tests for new duration behavior

Closes #2922


interpreter/interpreter.go
```diff
@@ -613,6 +613,7 @@ func (itrp *Interpreter) doCall(ctx context.Context, call *semantic.CallExpressi
 	}
 
 	// Call the function
+	ctx = context.WithValue(ctx, "scope", scope)
```
jsternberg, line 616: This seems wrong to me, but I'm not sure what the proper way would be to access options.

If we do end up using context to pass the current scope, we would need to create a specific context key for it and formalize this as the way to access the surrounding scope.

Another way we can do this is by requiring the now time to be passed to the function. I've used this pattern a few times:

```
truncate = (t, unit) => _truncate(t: t, unit: unit, now: now())
builtin _truncate
```

This way the resolution of the `now()` identifier is handled by the interpreter instead of within Go. This is a possible way to deal with this as I don't think this is something we've encountered previously.

I also know @adrian-thurston was working on a way of injecting the now time into the context for `tableFind`. I don't know if he ever merged that but we may want to consider that as a possible solution too.

stdlib/date/truncate_duration_test.flux
```diff
@@ -0,0 +1,43 @@
...
+"
+
+t_duration_truncate = (table=<-) =>
+	(table
+	    |> range(start: 2018-05-22T19:53:00.000000000Z)
+	    |> drop(columns: ["_start", "_stop"])
+		|> map(fn: (r) => ({r with _value: int(v: date.truncate(t: duration(v: r._value), unit: 1s))})))
+
+test _duration_truncate = () =>
+	({input: testing.loadStorage(csv: inData), want: testing.loadMem(csv: outData), fn: t_duration_truncate})
```
jsternberg, line 43: Try to watch out for these kinds of trailing newlines. It's not required and not all editors will insert a trailing newline, but many will so it's generally better to ensure that a file has a trailing newline. For example, vim will insert a trailing newline always and there's no real way to write the file without it so the next person who would modify this file would end up getting a one line diff.

stdlib/date/date.go
```diff
@@ -3,6 +3,7 @@ package date
 import (
 	"context"
 	"fmt"
+	"github.com/influxdata/flux/lang/execdeps"
```
wolffcm, line 6: _nit_ We normally put all the stdlib imports first, followed by a blank line, followed by all the other imports. Each section (stdlib and the other imports) should be sorted alphabetically. Most IDEs get this wrong unfortunately.

stdlib/date/truncate_duration_test.flux
ChitlangeSahas: Noted.

lang/compiler.go
```diff
@@ -391,6 +387,25 @@ func (p *AstProgram) getSpec(ctx context.Context, runtime flux.Runtime, alloc *m
 }
 
 func (p *AstProgram) Start(ctx context.Context, alloc *memory.Allocator) (flux.Query, error) {
+	if p.opts == nil {
```
wolffcm, line 390: I think I have misled you on my advice on how to handle the `now` time stuff. Here's the best way to fix it I think:
- Start by rolling back all the change in this file so far
- Add the following lines below, before calling `p.Program.Start`
```
deps := execdeps.NewExecutionDependencies(alloc, &p.Now, p.Logger)
ctx = deps.Inject(ctx)
```

At first I thought that we would not want to duplicate code needlessly, but it turns out that the value of `Now` may change after `Eval` is called, because it may process an `option now = () => ...` statement. This change is propagated to `p.Now` in `getSpec`, above. So we want to overwrite the old value if it has changed.

ChitlangeSahas: I followed your conversation with Joshua and this makes sense.

lang/compiler.go
```diff
@@ -391,6 +391,8 @@ func (p *AstProgram) getSpec(ctx context.Context, runtime flux.Runtime, alloc *m
 }
 
 func (p *AstProgram) Start(ctx context.Context, alloc *memory.Allocator) (flux.Query, error) {
+	deps := execdeps.NewExecutionDependencies(alloc, &p.Now, p.Logger)
+	ctx = deps.Inject(ctx)
```
wolffcm, line 395: Can you move these lines down to line 411/413? We want to inject the dependencies in the context after the call to `getSpec` so that the correct now time is used.

PR #2940, opened 2020-06-23
feat(stdlib/date): update date.nanosecond to use Timeable

ChitlangeSahas:
Done checklist

 - [x] Update type signature of function
 - [x] Update implementation to correctly handle duration values
 - [x] Add tests for new duration behavior

Closes: #2921


ChitlangeSahas:
Had to close this. The intent of this PR was achieved with #2941

PR #2945, opened 2020-06-26
feat(stdlib/http): adding pathEscape function

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/google/uuid v1.1.1 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e
-	github.com/influxdata/pkg-config v0.2.2
+	github.com/influxdata/pkg-config v0.2.3
```
rockstar, line 26: Why did we have to update `pkg-config` for this branch?

stdlib/http/path_encode_test.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"testing"
+
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/stdlib/http"
+	"github.com/influxdata/flux/values"
+)
+
+// some characters
+// want those escaped
```
rockstar, line 12: I think you can drop these comments, no?

stdlib/http/path_encode.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/runtime"
+	"github.com/influxdata/flux/values"
+)
+
+func init() {
+	runtime.RegisterPackageValue("http", "pathEscape", pathEscapeFunc)
+}
+
+const inputStringArg = "x"
```
rockstar, line 16: Is `x` a good argument name? Could we use something more descriptive?

I realize this is a common pattern across a lot of our codebase, but I think we could do ourselves a favor and start to break from that tradition a bit.

stdlib/http/http_path_encode_endpoint_test.flux
```diff
@@ -0,0 +1,39 @@
+package http_test
```
rockstar, line 1: I wonder if we really need an endpoint test for this. From my understanding of this test, it tests the same code path as the unit test while also using a lot of overhead that doesn't contribute to the test. It might make good example code for how to use `http.pathEscape`, but as a test, it seems unnecessary. Is that off base?

go.mod
yuktarakesh: I realized I was failing the tidy test and ran 'go mod tidy' which is the result of this. Not sure if I should be doing anything else for it.

stdlib/http/path_encode_test.go
yuktarakesh: Yes, will do.

stdlib/http/path_encode.go
yuktarakesh: Makes sense - I think I can rename it to indicate it being an input string.

stdlib/http/http_path_encode_endpoint_test.flux
yuktarakesh: From speaking with Chris, we thought it might be good to see how this test functions with everything else. If this causes bigger issues it might make sense to revisit its necessity.

rockstar: I'm normally a little reticent to have multiple tests testing the same thing (they get hard to remove later), but if you and @wolffcm have talked it over and feel confident this is right, I'm happy for it to merge.

wolffcm: Yeah, my reasoning is that functions executed from inside of a table stream function like `map` are using a slightly different code path, and so it's worth adding tests for that case. For example, the `Context` might have different stuff in it, like we saw with functions that needed to access `now`. 

In general I totally agree that unneeded testing can become painful if left unchecked.

PR #2951, opened 2020-06-30
feat(libflux): apply Timeable constraint to int type

ChitlangeSahas:
Applying Timeable constraint to int type so that functions that use int as a time, can be constrained to Timeable.

### Done checklist

- [x] Timeable kind constraint has been added
- [x] int satisfies the constraint
- [x] Units tests for the Timeable constraint and int type.
- [x] Does NOT include changing any existing functions to use Timeable, those will be separate issues.


sanderson:
How are integers used to calculate a time value? Are they nanoseconds? Relative to epoch time or `now`?

ChitlangeSahas:
@sanderson  My understanding is adding or subtracting nanoseconds from time / `now`. I could be off.

sanderson:
@rockstar Can you confirm? ☝️

jlapacik:
Integers are actually interpreted as nanoseconds relative to epoch.

sanderson:
Thanks @jlapacik

wolffcm:
I am late to this, but this seems weird to me.

Suppose that you forget to add the unit in your duration. It would silently give you a different answer than you expect (likely no data). Also the questions Scott asked will be raised by those reading Flux code:
- they might easily assume that integers are relative to `now`, since durations are too
- the units will be unclear

I get that not allowing `int` to be `Timeable` would mean that we can't make `range` take `Timeable` since that would be a breaking change, but this doesn't seem to move things in the right direction. It means that other functions besides `range` will have the same troubles described above.

wolffcm:
@nathanielc I understand that you argued that integers should be timeable, so I wonder if you could assuage my fear that this change makes Flux less usable, for reasons described in my previous comment. ☝️

nathanielc:
I agree with the arguments that `int` as a Timeable type is not user friendly, the only reason I went forward with it is because it would be a breaking change to remove. Perhaps we should explore the impact of breaking change and just change it? 


I think if a user wants to use an integer as nanoseconds since the epoch as a time value it should be explicit and look like this `time(v: EPOCH_TIMESTAMP)`. 

Should we do an audit of query logs to see if anyone is using this behavior?

PR #2952, opened 2020-06-30
refactor: AstProgram.GetAst()

yzhang1991:
…lize()

Changes needed for the OSS API to extract option reference PR https://github.com/influxdata/influxdb/pull/18796

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -364,8 +364,9 @@ func (p *AstProgram) GetAst() (flux.ASTHandle, error) {
 }
 
 func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flux.Spec, values.Scope, error) {
-	if _, err := p.GetAst(); err != nil {
-		return nil, nil, err
+	ast, astErr := p.GetAst()
+	if astErr != nil {
+		return nil, nil, astErr
 	}
 
 	// The program must inject execution dependencies to make it available to
@@ -374,7 +375,7 @@ func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flu
 	ctx = deps.Inject(ctx)
 	s, cctx := opentracing.StartSpanFromContext(ctx, "eval")
 
-	sideEffects, scope, err := p.Runtime.Eval(cctx, p.Ast, flux.SetNowOption(p.Now))
+	sideEffects, scope, err := p.Runtime.Eval(cctx, ast, flux.SetNowOption(p.Now))
 	if err != nil {
 		return nil, nil, err
 	}
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -349,15 +349,9 @@ type AstProgram struct {
 	Now time.Time
 }
 
-// Prepare the AstProgram for semantic analysis
-func (p *AstProgram) Normalize() error {
-	if p.opts == nil {
-		p.opts = defaultOptions()
-	}
-	if p.Now.IsZero() {
-		p.Now = time.Now()
-	}
-	if p.opts.extern != nil {
+// Merge the extern code into the main package.
+func (p *AstProgram) MergeExtern() error {
+	if p.opts != nil && p.opts.extern != nil {
 		extern := p.opts.extern
 		if err := p.Runtime.MergePackages(extern, p.Ast); err != nil {
 			return err
@@ -369,7 +363,13 @@ func (p *AstProgram) Normalize() error {
 }
 
 func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flux.Spec, values.Scope, error) {
-	if err := p.Normalize(); err != nil {
+	if p.Now.IsZero() {
+		p.Now = time.Now()
+	}
+	if p.opts == nil {
+		p.opts = defaultOptions()
+	}
+	if err := p.MergeExtern(); err != nil {
 		return nil, nil, err
 	}
 	// The program must inject execution dependencies to make it available to
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -169,9 +169,14 @@ type FluxCompiler struct {
 }
 
 func wrapFileJSONInPkg(bs []byte) []byte {
+	source := string(bs)
+	pkgHeader := `{"type":"Package"`
+	if source[0:len(pkgHeader)] == pkgHeader {
+		return bs
+	}
 	return []byte(fmt.Sprintf(
 		`{"type":"Package","package":"main","files":[%s]}`,
-		string(bs)))
+		source))
 }
 
 func IsNonNullJSON(bs json.RawMessage) bool {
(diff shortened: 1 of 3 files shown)


lang/compiler.go
```diff
@@ -344,7 +349,8 @@ type AstProgram struct {
 	Now time.Time
 }
 
-func (p *AstProgram) getSpec(ctx context.Context, runtime flux.Runtime, alloc *memory.Allocator) (*flux.Spec, values.Scope, error) {
+// Prepare the AstProgram for semantic analysis
+func (p *AstProgram) Normalize() error {
```
jsternberg, line 353: Does this function need to be publicly exposed?

lang/compiler.go
```diff
@@ -169,9 +169,14 @@ type FluxCompiler struct {
 }
 
 func wrapFileJSONInPkg(bs []byte) []byte {
+	source := string(bs)
+	pkgHeader := `{"type":"Package"`
+	if source[0:len(pkgHeader)] == pkgHeader {
```
jsternberg, line 174: I'm not sure what's going on here. Can you add a comment? It seems to be checking for a specific string but I'm a bit concerned that the look for an exact match will create problems. There's no requirement that the JSON start with the type.

lang/compiler.go
yzhang1991: Yeah, I want to expose it. It is used here: https://github.com/influxdata/influxdb/pull/18796/files#diff-41a88fd64faa5296ede9fdeb08529d1fR195

The problem is that I need to walk the semantic graph to look for option references and the graph is encapsulated inside the compile function. So I want to get the semantic graph by myself. The `extern` piece of code needs to be merged together in order to let our analyzer to make sense of the option usages.

lang/compiler.go
yzhang1991: When the UI pass in the `extern` JSON, it was always a `File`, to enable it to be merged with the user's query, we used this function to wrap it into a package. But our compiler always generates a `Package`. When I need to generate some `extern` JSON objects in tests using our compiler, it will cause problems here.

There is no requirement that the JSON start with the type, but if it has type and the type is already a package, we do not wrap it again. I feel it is safe to do so.

lang/compiler.go
jsternberg: It looks to me like the primary purpose is to retrieve a merged AST with the externs. Instead of a normalize function, can you add a function to retrieve the merged AST directly?

The main reason is it seems a bit abstract the way that it currently is. When I read the function description, I don't really understand why I would need this. I think it would be more clear if we had a `Get` function that said "returns the normalized ast" rather than a function that changed the state of the compiler.

lang/compiler.go
jsternberg: This still makes me very uncomfortable. Is there a way that we can unwrap the package from the compiler so that we can pass a file in here? Or I wonder if a separate field in the struct would work.

```go
type ASTCompiler struct {
    Extern json.RawMessage `json:"extern,omitempty"`
    ExternPackage flux.ASTHandle `json:"-"`
    AST    json.RawMessage `json:"ast"`
    Now    time.Time
}
```

This way you can pass in the ast handle directly and there's no need to serialize and then deserialize again.

lang/compiler.go
yzhang1991: You have to be in the `AstProgram` so you have the context and access to the `extern`. It may need a clearer name. This is used by the compiler itself to mutate its own state anyway, and it is idempotent.

lang/compiler.go
yzhang1991: I actually wanted to change UI's template for generating the extern API, so they can directly generate a package, not waiting for us to add this hard-coded wrapper: https://github.com/influxdata/influxdb/blob/master/ui/src/variables/utils/buildVarsOption.ts#L9

lang/compiler.go
jsternberg: That makes sense and I think that's fine. It also makes sense that you would want to reuse the compiler for this as the main way that we access the merged state.

I think the current API is a bit confusing. It seems to imply that it needs to be called to make the AST program work or that it makes some kind of non-idempotent change to the AST. When I hear normalize, I hear it making a change to the contents. I also get concerned with functions that change state because they can be hard to keep track of and it forces `Normalize()` to be reentry safe for being called multiple times.

I think it would be a better API if we created a function that directly returned the AST you want from the compiler or if we added a function to directly do what you want. Since we are adding a function to a program produced by the compiler, we could add an `Analyze` method that merges the AST's and produces an analyzed semantic graph. We would still get the benefit of code reuse but it is my opinion that the API would be more self-documenting.

The problem I find with a normalize function is it isn't clear why you would need to normalize the AST or what that means. And if the purpose of the function is to prepare the AST for semantic analysis we can just have a function that performs semantic analysis.

lang/compiler.go
jsternberg: I don't think I'm comfortable with that. This would be a backwards incompatible change and it makes me more uncomfortable to attempt to do inspection of the JSON to see if we are wrapping a package or a file. It's relying on implementation details rather than an established way to perform JSON introspection.

It may be true that the extern should have likely been a package but this is an external API and I would not be comfortable changing this at this point. If we want to change the structure of the AST compiler, I think we need to create a version 2 or add an additional JSON attribute to the existing compiler.

If we want to change that or overload this operator I think that needs to be part of a separate issue and discussion. Adding an additional struct field that takes an AST handle directly mitigates the backwards compatibility aspects and solves the immediate problem if we want to go with the code snippet I have above.

yzhang1991: I handled this problem by tweaking the test code in order to avoid complicating the product code.

lang/compiler.go
yzhang1991: We sometimes need to analyze just an AST without having an `AstProgram` or additional `extern`.
This merge step is very unique to `AstProgram` so I feel still prefer to make it in this way and name it better.

jsternberg: I think you might be misunderstanding my concern. I like the function and I like the implementation. I just want the function to return the merged AST so that we are self-documenting how to use the interface. I want to be able to write this sentence.

> If you want to retrieve the final AST from the AstCompiler, you can use this method and it will merge any externs if they exist or it will return the compiled AST.

When you create a function that mutably changes the object and then retrieve the value form the mutable object, it becomes harder to know when you're supposed to call that function. If I'm an external user, how do I know that I need to call the merge function before I get the AST attribute? That's the reason why I would prefer a function that retrieves the merged AST and let the compiler figure out if it needs to merge it or not.

lang/compiler.go
```diff
@@ -344,29 +344,37 @@ type AstProgram struct {
...
+			return nil, err
 		}
 		p.Ast = extern
 		p.opts.extern = nil
 	}
+	return p.Ast, nil
+}
+
+func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flux.Spec, values.Scope, error) {
+	if _, err := p.GetAst(); err != nil {
```
jsternberg, line 367: You can just use the returned ast here.

PR #2953, opened 2020-07-01
fix(compiler): substitute array element types correctly

jsternberg:
The array element type wouldn't substitute correctly because it compared
only one of the types with itself instead of comparing the two different
types.

The compiler has also been modified to catch when a required argument is
omitted within the go runtime compiler.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -127,7 +127,7 @@ func substituteTypes(subst map[uint64]semantic.MonoType, inType, in semantic.Mon
 			return err
 		}
 
-		rt, err := inType.ElemType()
+		rt, err := in.ElemType()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


jsternberg:
I was working on my own contrib function and I've been attempting to use the type system more to determine output types for go transformations. This specific one passed the arrow array column to a function to return the result and the compiler was telling me that it didn't have a valid type.

It's been a kind of side experimentation and has been useful for finding stuff like this.

jsternberg:
This is the PR for that function: https://github.com/influxdata/flux/pull/2906

I haven't updated the PR. I had an idea for how to refactor it and when I refactored it I tried to use the compiler return type instead of the type of the returned values and it wasn't working correctly.

PR #2956, opened 2020-07-01
fix(execute): appending an array of booleans with null values functions correctly

jsternberg:
When appending an array of boolean values, it was marking the wrong
indices as boolean by marking one after the value that was appended
rather than the one just appended. This resulted in other sections of
the code being wrong because the copied column was not copied correctly.

Related to influxdata/influxdb#18433.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/table_test.go
+++ b/execute/table_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/apache/arrow/go/arrow/array"
+	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/execute"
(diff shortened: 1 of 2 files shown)


execute/table_test.go
```diff
@@ -386,6 +388,206 @@ func TestCopyTable(t *testing.T) {
...
+			}(),
+			want: &executetest.Table{
+				ColMeta: []flux.ColMeta{{
+					Label: "_value",
+					Type:  flux.TInt,
+				}},
+				Data: [][]interface{}{
+					{int64(2)}, {int64(3)}, {nil}, {int64(8)}, {nil}, {int64(6)},
+				},
+			},
```
fchikwekwe, line 419: I'm assuming these arrays have to be homogenous? 

If so, thanks for adding this test coverage. If not, maybe add an array with mixed types?

jsternberg: Yes. They all have to be the same type.

PR #2957, opened 2020-07-01
feat(stdlib/generate/from): update from to use Timeable

ChitlangeSahas:


libflux/src/core/semantic/builtins.rs
```diff
@@ -109,7 +109,7 @@ pub fn builtins() -> Builtins<'static> {
                  "join" => "forall [t0, t1, t2] where t0: Row, t1: Row, t2: Row (left: [t0], right: [t1], fn: (left: t0, right: t1) -> t2) -> [t2]",
             },
             "generate" => semantic_map! {
-                "from" => "forall [] (start: time, stop: time, count: int, fn: (n: int) -> int) -> [{ _start: time | _stop: time | _time: time | _value:int }]",
+                "from" => "forall [t0] where t0: Timeable (start: t0, stop: t0, count: int, fn: (n: int) -> int) -> [{ _start: time | _stop: time | _time: time | _value:int }]",
```
wolffcm, line 112: Unless I'm missing something, it seems like there's a little more work necessary to ensure that durations work with `generate.from`. I think that this part of the code will produce an error if you pass in durations:
https://github.com/influxdata/flux/blob/baaa4eccb94b88998b247637c9ae412e2b0d63ea/stdlib/generate/from.go#L41-L51

I know it's not feasible to write a `_test.flux` that exercises `generate.from`, but you could write a test case similar to this one for `filter`:
https://github.com/influxdata/flux/blob/be8c67e907a2d96c7f9ced8c7660fdffeb152fcc/stdlib/universe/filter_test.go#L27

That would ensure that we handle duration arguments correctly.

ChitlangeSahas: > Unless I'm missing something, it seems like there's a little more work necessary to ensure that durations work with `generate.from`. I think that this part of the code will produce an error if you pass in durations:

@wolffcm I could be misunderstanding that statement, but I thought the `ToQueryTime` took care of `duration` -> `time` when the `start` or `stop` is duration. 🤔 

Screen Shot 2020-07-06 at 8 49 27 AM

seems to work: (?) 

Screen Shot 2020-07-06 at 8 47 19 AM

wolffcm: Ah, ok, so it works, that's great. I didn't know the extra logic was added there. I still think there's value to a unit test, I'm not sure why there wasn't one there in the first place.

stdlib/generate/from_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"testing"
+	"time"
+
+	"github.com/influxdata/flux"
+	_ "github.com/influxdata/flux/builtin"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/querytest"
+	"github.com/influxdata/flux/runtime"
+	_ "github.com/influxdata/flux/stdlib"
```
wolffcm, line 13: I think you only need the `"github.com/influxdata/flux/builtin"` line above, and you can remove this line and the one for `"github.com/influxdata/flux/stdlib/generate"`

PR #2959, opened 2020-07-02
feat(stdlib/universe): added MergeFilterRule to universe.filter

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/universe/filter.go
```diff
@@ -359,3 +360,40 @@ func (RemoveTrivialFilterRule) Rewrite(ctx context.Context, filterNode plan.Node
...
 }
+
+// MergeFiltersRule merges Filter nodes whose body is a single return to create one Filter node.
+type MergeFiltersRule struct{}
+
+func (MergeFiltersRule) Name() string {
+	return "MergeFiltersRule"
+}
+
+// need to think of pattern
```
wolffcm, line 371: I think that this comment and the one below at line 376 aren't needed now.

yuktarakesh: Let me make that change!

wolffcm:
I think that we also want `_test.flux` tests for this change, specifically exercising the planner. There's several examples of this, here's one:
https://github.com/influxdata/flux/blob/master/stdlib/planner/window_count_eval_test.flux
https://github.com/influxdata/flux/blob/master/stdlib/planner/window_count_push_test.flux

These tests are exactly the same, but they have an `option` that will disable a planner rule. This way we can see that the query gets the same result whether the rule is applied or not.

Can you write a similar test for your rule?

stdlib/universe/filter_test.go
```diff
@@ -783,6 +783,56 @@ gen.tables(n: 2048, tags: [{name: "a", cardinality: 10}])
...
+			},
+			After: &plantest.PlanSpec{
+				Nodes: []plan.Node{
+					plan.CreatePhysicalNode("from", from),
+					plan.CreatePhysicalNode("filter0", filter2),
+				},
+				Edges: [][2]int{{0, 1}},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
yuktarakesh: yes!

stdlib/universe/filter.go
```diff
@@ -359,3 +361,42 @@ func (RemoveTrivialFilterRule) Rewrite(ctx context.Context, filterNode plan.Node
...
+		// Not an expression.
+		return filterNode, false, nil
+	}
+	filterSpec2 := filterNode.Predecessors()[0].ProcedureSpec().(*FilterProcedureSpec)
+	bodyExpr2, ok := filterSpec2.Fn.Fn.GetFunctionBodyExpression()
+	if !ok {
+		// Not an expression.
+		return filterNode, false, nil
+	}
+	if filterSpec1.KeepEmptyTables != filterSpec2.KeepEmptyTables && !filterSpec2.KeepEmptyTables {
```
wolffcm, line 390: Can you add a comment the describes the reasoning of the logic here? Just so it's clear if we ever need to revisit this.

PR #2971, opened 2020-07-08
fix(execute/table): appending empty tables to a buffered builder normalizes the schema

jsternberg:
When `AppendBuffer` was used to append buffers to a buffered builder,
empty tables (which produce no buffers) would not affect the schema at
all. This meant that the copies of tables to a buffered builder didn't
work to properly normalize the schema when the incoming table was empty.

This updates the `BufferedBuilder` to include an `AppendTable` which
will correctly handle empty tables.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
@rockstar when I was writing the tests for this, I found it was a bit frustrating to write them out with `executetest.Table` so I wanted to see if I could come up with something better. I can rewrite the tests with `executetest.Table` is preferred, but I thought it might be a good idea to experiment to see if there was an easier way to statically write tables for tests.

PR #2981, opened 2020-07-09
feat(execute/table): add stringify method for a table and a diff utility

jsternberg:
This adds a `Stringify` method to the `internal/execute/table` package
to convert a table into a string that is easily diffable. It also
contains a `DiffIterator` method that will sort and then stringify each
table and then pass it through a line diff instead of `cmp.Diff`.

This should produce more readable results than `cmp.Diff`.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/execute/table/sort.go
+++ b/internal/execute/table/sort.go
@@ -1,0 +1,32 @@
+package table
+
+import (
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+)
+
+// Sort will read a TableIterator and produce another TableIterator
+// where the keys are sorted.
+//
+// This method will buffer all of the data since it needs to ensure
+// all of the tables are read to avoid any deadlocks. Be careful
+// using this method in performance sensitive areas.
+func Sort(tables flux.TableIterator) (flux.TableIterator, error) {
+	groups := execute.NewGroupLookup()
+	if err := tables.Do(func(table flux.Table) error {
+		buffered, err := execute.CopyTable(table)
+		if err != nil {
+			return err
+		}
+		groups.Set(buffered.Key(), buffered)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	var buffered []flux.Table
+	groups.Range(func(_ flux.GroupKey, value interface{}) {
+		buffered = append(buffered, value.(flux.Table))
+	})
+	return Iterator(buffered), nil
+}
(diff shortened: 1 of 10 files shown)


internal/execute/table/sort.go
```diff
@@ -0,0 +1,32 @@
...
+	}); err != nil {
+		return nil, err
+	}
+
+	var buffered []flux.Table
+	groups.Range(func(_ flux.GroupKey, value interface{}) {
+		buffered = append(buffered, value.(flux.Table))
+	})
+	return Iterator(buffered), nil
+}
```
fchikwekwe, line 32: This sorts the keys. Is that used to then sort the stream?

internal/execute/table/stringify.go
```diff
@@ -0,0 +1,151 @@
...
+	var sb strings.Builder
+	stringifyKey(&sb, table)
+	if err := table.Do(func(cr flux.ColReader) error {
+		stringifyRows(&sb, cr)
+		return nil
+	}); err != nil {
+		_, _ = fmt.Fprintf(&sb, "table error: %s\n", err)
+	}
+	return sb.String()
+}
```
fchikwekwe, line 26: This is a cool idea. I wonder what the output looks like.

internal/execute/table/stringify_test.go
```diff
@@ -0,0 +1,48 @@
...
+	got := table.Stringify(in)
+
+	want := `# _field=f0,_measurement=m0,t0=a _time=time,_value=int
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:00Z,_value=6i
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:10Z,_value=7i
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:20Z,_value=4i
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:30Z,_value=12i
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:40Z,_value=3i
+_field=f0,_measurement=m0,t0=a _time=2020-01-01T00:00:50Z,_value=9i
+`
```
fchikwekwe, line 27: ahh, that's why you look at the tests first. lol

internal/execute/table/sort.go
fchikwekwe: I see. The sort function is called within Diff to put the tables in order. Makes sense.

PR #2982, opened 2020-07-09
feat: find_var_type() api

yzhang1991:


libflux/src/flux/lib.rs
```diff
@@ -458,31 +504,75 @@ pub unsafe extern "C" fn flux_analyze_with(
...
+        },
+    );
+    infer_with_env(ast_pkg, f, Some(env))
+        .map(|(_, env, _)| env.lookup(var_name.as_str()).unwrap().expr.clone())
+}
+
+/// wasm version of the flux_find_var_type() API. Instead of returning a flat buffer that contains
+/// the MonoType, it returns a JsValue。
+#[wasm_bindgen]
+pub fn wasm_find_var_type(source: &str, file_name: &str, var_name: &str) -> JsValue {
```
yzhang1991, line 571: @ebb-tide This is relevant to you

libflux/src/core/semantic/env.rs
```diff
@@ -135,4 +135,12 @@ impl Environment {
             None => panic!("cannot pop final stack frame from type environment"),
         }
     }
+
+    // Copy all the variable bindings from another Environment to the current environment.
+    // This does not change the current environment's parent and readwrite flag.
+    pub fn copy_bindings_from(&mut self, other: &Environment) {
```
jlapacik, line 141: What is this used for?

libflux/src/core/semantic/types.rs
```diff
@@ -516,6 +517,15 @@ impl MaxTvar for Tvar {
     }
 }
 
+impl Serialize for Tvar {
```
jlapacik, line 520: Why can't this be derived?

libflux/src/core/tests/analyze_test.rs
```diff
@@ -1,11 +1,86 @@
...
+vint = v.int + 2
+f = (v) => v.shadow
+g = () => v.sweet
+x = g()
+vstr = v.str + "hello"
+"#;
+    let t = find_var_type(source, "v").expect("Should be able to get a MonoType.");
+    assert_eq!(
+        t,
+        MonoType::Row(Box::new(Row::Extension {
```
jlapacik, line 21: A string comparison would be a little easier to decipher here.

libflux/src/core/tests/analyze_test.rs
```diff
@@ -1,11 +1,86 @@
...
+                        v: MonoType::String,
+                    },
+                    tail: MonoType::Var(Tvar(22))
+                })),
+            }))
+        }))
+    );
+
+    let serialized = serde_json::to_string_pretty(&t).unwrap();
+    println!("{}", serialized);
```
jlapacik, line 43: Are these just left over debug statements or did you mean to test the json output?

libflux/src/core/semantic/mod.rs
```diff
@@ -82,3 +90,25 @@ pub fn convert_source(source: &str) -> Result<nodes::Package, Error> {
     )?;
     Ok(nodes::inject_pkg_types(sem_pkg, &sub))
 }
+
+/// Given a Flux source and a variable name, find out the type of that variable in the Flux source code.
+/// A type variable will be automatically generated and injected into the type environment that
+/// will be used in semantic analysis. The Flux source code itself should not contain any definition
+/// for that variable.
+pub fn find_var_type(source: &str, var_name: &str) -> Result<MonoType, Error> {
```
jlapacik, line 98: Should this take an optional type environment so that we don't have to duplicate the same code in the `flux` package?

libflux/src/core/semantic/env.rs
yzhang1991: It is currently used to copy the variable binding we injected over to the prelude environment: https://github.com/influxdata/flux/blob/feat/option-reference-api/libflux/src/flux/lib.rs#L538

I made it a function because I think we might, in the future, need to track the references for more than one variable, and having such function will be useful (maybe overthinking).

libflux/src/core/semantic/types.rs
yzhang1991: It actually can, changed it.

libflux/src/core/semantic/mod.rs
yzhang1991: The only overlap of those two `find_var_type()` functions is injecting type var into an environment. The fresher, env, built-ins, importer, almost every parameter is prepared differently. I don't really like to add many parameters to reuse just one environment.add() logic.

libflux/go/libflux/analyze.go
```diff
@@ -74,6 +77,35 @@ func Analyze(astPkg *ASTPkg) (*SemanticPkg, error) {
 	return p, nil
 }
 
+func FindVarType(astPkg *ASTPkg, varName string) (semantic.MonoType, error) {
+	defer func() {
+		// This is necessary because the ASTPkg returned from the libflux API calls has its finalizer
+		// set with the Go runtime. But the find_var_type() API will consume the AST package during
+		// the conversion from the AST package to the semantic package.
+		// Setting this ptr to nil will prevent a double-free error.
+		astPkg.ptr = nil
```
nathanielc, line 86: Should we change the behavior of `find_var_type` to not consume the AST package? Does it need to become owner of the resource?

libflux/include/influxdata/flux.h
```diff
@@ -113,6 +113,13 @@ struct flux_error_t *flux_analyze_with(struct flux_semantic_analyzer_t *, struct
 // flux_ast_pkg_t* argument, so it should not be reused after calling this function.
 struct flux_error_t *flux_analyze(struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
 
+// Find out the type of a variable referenced in the given Flux AST and return a MonoType flat buffer for it.
+// The Flux AST should not contain any definition for the referenced variable.
+// A type variable for the designated variable will be automatically injected into the type environment
+// that will be used in semantic analysis.
+// The second parameter is the variable identifier string.
+struct flux_error_t *flux_find_var_type(struct flux_ast_pkg_t *, const char *, struct flux_buffer_t *);
```
nathanielc, line 121: If we don't change the behavior of the function should we add a note here about how the AST will be consumed by calls to the method?

libflux/src/flux/lib.rs
```diff
@@ -458,31 +505,77 @@ pub unsafe extern "C" fn flux_analyze_with(
...
+    let (env, sub) = infer_pkg_types(&mut sem_pkg, prelude, &mut f, &imports, &builtin_importer)?;
+    Ok((sem_pkg, env, sub))
+}
+
+/// Given a Flux source and a variable name, find out the type of that variable in the Flux source code.
+/// A type variable will be automatically generated and injected into the type environment that
+/// will be used in semantic analysis. The Flux source code itself should not contain any definition
+/// for that variable.
+/// This version of find_var_type is aware of the prelude and builtins.
+pub fn find_var_type(ast_pkg: ast::Package, var_name: String) -> Result<MonoType, Error> {
```
nathanielc, line 555: I am confused why there are two `find_var_type` functions? When is one or the other used?

libflux/src/core/semantic/mod.rs
jlapacik: Ok. Is there an argument for keeping both functions?

libflux/src/flux/lib.rs
yzhang1991: The one in the `flux` crate is aware of the prelude and the one in `core` is not.
The one in `core` is just for testing purpose

PR #2993, opened 2020-07-13
feat(docs): add types grammar to spec

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


docs/SPEC.md
```diff
@@ -405,10 +405,28 @@ The default value is set using the time zone of the running process.
...
 
 A type defines a set of values and operations on those values.
-Types are never explicitly declared as part of the syntax.
+Types are never explicitly declared as part of the syntax except as part of a [builtin statement](#built-ins).
 Types are always inferred from the usage of the value.
 Type inference follows a Hindley-Milner style inference system.
 
+    polytype = monotype ["where" constraints] .
+    monotype = tvar | basic | array | record | function .
+    tvar     = "A" … "Z" .
```
nathanielc, line 414: do we want to allow for `t#` format types as well in cases where there are lots of type variables? Or only use letters?

docs/SPEC.md
```diff
@@ -405,10 +405,28 @@ The default value is set using the time zone of the running process.
...
+    basic    = "int" | "uint" | "float" | "string" | "bool" | "time" | "duration" | "bytes" | "regexp" .
+    array    = "[" monotype "]" .
+    record   = "{" [properties] "}" | "{" tvar "with" properties "}" .
+    function = "(" [parameters] ")" "->" monotype .
```
nathanielc, line 418: Should this be `=>` instead of `->`?

jlapacik: Yes this is supposed to be `=>`

docs/SPEC.md
jlapacik: I think we should only use letters for now. To need any more than 3 type variables will be extremely rare at best.

docs/SPEC.md
```diff
@@ -1085,11 +1009,29 @@ These preassigned values are defined in the source files for the various built-i
...
+    Parameter  = [ "<-" | "?" ] identifier ":" MonoType .
+
+    Constraints = Constraint { "," Constraint } .
+    Constraint  = Tvar ":" Kinds .
+    Kinds       = identifier { "+" identifier } .
 
 Example
 
-    builtin from : (bucket: string, bucketID: string) -> stream
+    builtin filter : (<-tables: [T], fn: (r: T) -> bool) -> [T]
```
nathanielc, line 1034: do we want to include the `where T : Record` bit to this example?

jlapacik: I think I'll leave it out for now just because the constraints themselves aren't part of the spec.

PR #2998, opened 2020-07-14
docs(overview): update the langauge around tables data frames in the overview doc

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -1,6 +1,6 @@
 # Design Overview
 
-This document provides an overview of the design of the query engine.
+This document provides an overview of the design of the Flux query engine.
 
 ## Concepts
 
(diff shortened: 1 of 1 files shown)


docs/Overview.md
```diff
@@ -10,7 +10,7 @@ There are several different concepts that make up the complete query engine.
     A query is represented as a directed acyclic graph (DAG).
 * Flux - Functional Language for defining a query to execute.
 * Parser - Parses a Flux script and produces a query.
-* Data Frame - A data frame is a matrix of time series data where one dimension is time and the other is series.
+* Table - Data is modeled as a sequence of records in table.
```
jsternberg, line 13: "in table" seems like there is a missing word but I can't identify which one was intended. Also, is the intention that table now represents all data in a single schema (like the tableflux proposal) or are we intending to capture the current state where a table is a sequence of records with a common schema and group key?

docs/Overview.md
```diff
@@ -84,7 +84,7 @@ type Predicate interface {}
 // Storage provides an interface to the storage layer.
 type Storage interface {
     // Read gets data from the underlying storage system and returns a data frame or error state.
-    Read(context.Context, []Predicate, TimeRange, Grouping) (DataFrame, ErrorState)
+    Read(context.Context, []Predicate, TimeRange, Grouping) (Table, ErrorState)
```
jsternberg, line 87: Referring to my previous comment, if we are intending to document the current state of things, then this would return an array of tables.

docs/Overview.md
adrian-thurston: Seems to be a circular definition? Thinking this should be a simple definition such as a "table is a list of rows" ('m sure there is something better floating around, just an example)

PR #2999, opened 2020-07-15
feat(execute/table): expose static table package and table diff functions

jsternberg:
This exposes the static table package to more easily produce test
tables. This is exposed publicly so other projects, such as influxdb,
can use it to test the tables they produce.

It adds a new function to the static table to extend the table by either
adding or overwriting columns. This way it is easier to create a base
table, such as one with common group keys, and then override it for each
table in a table iterator.

This also adds a diff iterator function that will accept a table
iterator, sort it, then pass it to `cmp.Diff` as a way of encapsulating
the diff workflow.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


internal/execute/table/stringify.go
```diff
@@ -1,151 +1,11 @@
...
-		if ts.Nanosecond() > 0 {
-			sb.WriteString(ts.Format(time.RFC3339Nano))
-		} else {
-			sb.WriteString(ts.Format(time.RFC3339))
-		}
-	default:
-		sb.WriteString("!(invalid)")
-	}
+func Stringify(t flux.Table) string {
+	return table.Stringify(t)
```
wolffcm, line 10: Why make the internal function point at the public one? Can't you just update all the callers?

jsternberg: The main reason is so I don't have to move the entire package. If I moved the function, then callers that use both would have to refer to both packages instead of one and they have the same name so you would have to rewrite the package name. This allows existing callers to opt into using only the public api or the full experimental api.

If this were a move of everything I would have just rewritten the import.

PR #3003, opened 2020-07-17
Feat/builtins parser update

ChitlangeSahas:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/ast/mod.rs
```diff
@@ -492,8 +492,131 @@ pub struct BuiltinStmt {
...
+    Int(Int),
+    Uint(Uint),
+    Float(Float),
+    String(StringType),
+    Bool(Bool),
+    Time(Time),
+    Duration(DurationType),
+    Bytes(Bytes),
+    Regexp(Regexp),
+    Invalid,
```
nathanielc, line 510: I'd like to see these names use a consistent pattern. Should we use the `Type` suffix for all of the types?

Meaning like this:

```
pub enum MonoType {
    Tvar(TvarType),
    Int(IntType),
    Uint(UintType),
    Float(FloatType),
    String(StringType),
    Bool(BoolType),
    Time(TimeType),
    Duration(DurationType),
    Bytes(BytesType),
    Regexp(RegexpType),
    Invalid,
}
```

libflux/src/core/ast/mod.rs
```diff
@@ -492,8 +492,131 @@ pub struct BuiltinStmt {
...
+    pub base: BaseNode,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct Regexp {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
+}
```
nathanielc, line 607: Since each of these structs contains only a BaseNode field would it make sense to create a single `BasicType` node like this?

```
pub struct BasicType {
    #[serde(skip_serializing_if = "BaseNode::is_empty")]
    #[serde(default)]
    #[serde(flatten)]
    pub base: BaseNode,
    pub name: Identifier,
}
```

Then the specific basic type is just stored as an identifier on the node? 

There are pros and cons to either pattern curious what others think?

nathanielc: @jlapacik Thoughts :point_up: ?

jlapacik: I think that removes a lot of redundancy. Are you thinking something like this?
```Rust
pub enum MonoType {
    Basic(Basic),
    ...
}

pub struct Basic {
    pub base: BaseNode,
    pub ty: Primitive,
}

pub enum Primitive {
    Int,
    Uint,
    ...
}
```

nathanielc: I think keeping them as Identifiers is best because we don't have any vanilla enum values like the Primitive enum in the AST and it creates some challenges with serialization. For example what would the JSON be for the Primitive enum? Is it a number or a string? A numer would make working with the AST in other languages difficult because you would need know the mapping of number to enum. If its a string thats basically the same thing as an Identifier so I say just use that.

jlapacik: @yuktarakesh @ChitlangeSahas I think @nathanielc's suggestion makes for a good simplification. Especially when it comes time to implement the serialization bits.

nathanielc: One other thought I had here, should we call the AST node NamedType instead of BasicType? That way if we ever have user defined types we can use the same ast node to represent it since its just a type named by its identifier?

PR #3008, opened 2020-07-20
feat(ast/edit): crud options and properties

fchikwekwe:
Closes influxdata/influxdb#17386.

### DIFF
--- a/ast/edit/task_editor.go
+++ b/ast/edit/task_editor.go
@@ -1,0 +1,109 @@
+package edit
+
+import (
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/codes"
+)
+
+// GetOption finds and returns the init for the option's variable assignment
+func GetOption(file *ast.File, name string) (ast.Expression, error) {
+	for _, st := range file.Body {
+		if val, ok := st.(*ast.OptionStatement); ok {
+			assign := val.Assignment
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if va.ID.Name == name {
+					if ok {
+						return va.Init, nil
+					}
+				}
+			}
+		}
+	}
+
+	return nil, &flux.Error{
+		Code: codes.Internal,
+		Msg:  "Option not found",
+	}
+}
+
+// SetOption replaces an existing option's init with the provided init or adds
+// the option if it doesn't exist. The file AST is mutated in place.
+func SetOption(file *ast.File, name string, expr ast.Expression) {
+	// check for the correct file
+	for _, st := range file.Body {
+		if val, ok := st.(*ast.OptionStatement); ok {
+			assign := val.Assignment
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if va.ID.Name == name {
+					// replace the variable assignment's init
+					va.Init = expr
+					return
+				}
+			}
+		}
+	}
+	// option was not found. prepend new option to body
+	file.Body = append([]ast.Statement{&ast.OptionStatement{
+		Assignment: &ast.VariableAssignment{
+			ID:   &ast.Identifier{Name: name},
+			Init: expr,
+		},
+	}}, file.Body...)
+}
+
+// DeleteOption removes an option if it exists. The file AST is mutated in place.
+func DeleteOption(file *ast.File, name string) {
+	for i, st := range file.Body {
+		if val, ok := st.(*ast.OptionStatement); ok {
+			assign := val.Assignment
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if va.ID.Name == name {
+					file.Body = append(file.Body[:i], file.Body[i+1:]...)
+					return
+				}
+			}
+		}
+	}
+}
+
+// GetProperty finds and returns the AST node for the property value.
+func GetProperty(obj *ast.ObjectExpression, key string) (ast.Expression, error) {
+	for _, prop := range obj.Properties {
+		if key == prop.Key.Key() {
+			return prop.Value, nil
+		}
+	}
+	return nil, &flux.Error{
+		Code: codes.Internal,
+		Msg:  "Property not found",
+	}
+}
+
+// SetProperty replaces an existing property definition with the provided object expression or adds
+// the property if it doesn't exist. The object expression AST is mutated in place.
+func SetProperty(obj *ast.ObjectExpression, key string, value ast.Expression) {
+	for _, prop := range obj.Properties {
+		if key == prop.Key.Key() {
+			prop.Value = value
+			return
+		}
+	}
+
+	obj.Properties = append(obj.Properties, &ast.Property{
+		BaseNode: obj.BaseNode,
+		Key:      &ast.Identifier{Name: key},
+		Value:    value,
+	})
+}
+
+// DeleteProperty removes a property from the object expression if it exists.
+// The object expression AST is mutated in place.
+func DeleteProperty(obj *ast.ObjectExpression, key string) {
+	for i, prop := range obj.Properties {
+		if key == prop.Key.Key() {
+			obj.Properties = append(obj.Properties[:i], obj.Properties[i+1:]...)
+			return
+		}
+	}
+}
--- a/ast/edit/task_editor_test.go
+++ b/ast/edit/task_editor_test.go
@@ -1,0 +1,432 @@
+package edit_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/ast/asttest"
+	"github.com/influxdata/flux/ast/edit"
+)
+
+func TestGetOption(t *testing.T) {
+	testCases := []struct {
+		testName string
+		optionID string
+		file     *ast.File
+		want     ast.Expression
+	}{
+		{
+			testName: "test getOption",
+			optionID: "task",
+			file: &ast.File{
+				Name: "foo.flux",
+				Body: []ast.Statement{
+					&ast.OptionStatement{
+						Assignment: &ast.VariableAssignment{
+							ID:   &ast.Identifier{Name: "bar"},
+							Init: nil,
+						},
+					},
+					&ast.OptionStatement{
+						Assignment: &ast.VariableAssignment{
+							ID:   &ast.Identifier{Name: "task"},
+							Init: &ast.BooleanLiteral{Value: false},
+						},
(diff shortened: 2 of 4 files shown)


compiler/compiler_test.go
```diff
@@ -16,11 +16,11 @@ var CmpOptions = semantictest.CmpOptions
...
-		name    string
-		fn      string
-		inType  semantic.MonoType
-		input   values.Object
-		want    values.Value
+		name           string
+		fn             string
+		inType         semantic.MonoType
+		input          values.Object
+		want           values.Value
```
fchikwekwe, line 23: change made by go fmt

ast/edit/task_editor.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+	return &flux.Error{
+		Code: codes.Internal,
+		Msg:  fmt.Sprintf("Could not delete %v, Option not found.", name),
+	}
+}
+
+// GetProperty finds and returns the AST node for the property value.
+func GetProperty(obj *ast.ObjectExpression, key string) (ast.Node, error) {
```
nathanielc, line 97: Should this return an `*ast.Property` since we know it will always be that type?

ast/edit/task_editor.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	return nil, &flux.Error{
+		Code: codes.Internal,
+		Msg:  "Property not found",
+	}
+}
+
+// SetOption replaces an existing property definition with the provided object expression or adds
+// the property if it doesn't exist. The object expression AST is mutated in place.
+func SetProperty(obj *ast.ObjectExpression, key string, value *ast.Expression) error {
```
nathanielc, line 111: Why does this need to be  a `*ast.Expression` instead of a `ast.Expression`?

ast/edit/task_editor.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	return nil, &flux.Error{
+		Code: codes.Internal,
+		Msg:  "Option not found",
+	}
+}
+
+// SetOption replaces an existing option definition with the provided option node or adds
+// the option if it doesn't exist. The pkg AST is mutated in place.
+func SetOption(pkg *ast.Package, name string, id string, opt ast.Node) error {
```
nathanielc, line 37: Should we change the API so that is deals with `ast.File` instead of `ast.Package` that way there is no ambiguity around which file to modify? 

In other words leave the logic of finding the specific file outside of this API.

ast/edit/task_editor.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"fmt"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/codes"
+)
+
+// GetOption finds and returns the AST node for the option value.
+func GetOption(pkg *ast.Package, name string) (ast.Node, error) {
```
nathanielc, line 12: Same question here should we make this API operate on `ast.File` instead?

ast/edit/task_editor.go
fchikwekwe: Yep, that's a good idea.

ast/edit/task_editor.go
fchikwekwe: I just did it that way. I don't think it does.

ast/edit/task_editor.go
fchikwekwe: I'll need to rewrite some of my tests to make this work, but this change makes sense.

fchikwekwe:
Made a few updates after review from @nathanielc. The option api now expects an `ast.File` instead of an `ast.Package`.

nathanielc:
@fchikwekwe Thanks for the updates. Did you have any thoughts on this?

>I see that GetOption and SetOption the API interacts with the OptionStatement node. This means that the consumers of the API will have to construct OptionStatments with the VariableAssignement nodes. Does it make sense instead to interact with the Init value of Option to make it simpler for callers?

fchikwekwe:
@nathanielc Interesting point. Unless I'm missing something, I don't see a big difference between interacting with the `VariableAssignment` node's ID vs the `Init` value. Are you saying that one is easier to construct? From my perspective writing tests for this API it doesn't seem to make a big difference.

nathanielc:
@fchikwekwe Dealing with just the `Init` means you do not have to construct the OptionStatment or the VariableAssignement nodes. Since that would be repeated code that all callers would need to duplicate I thought it would be nice to incorporate it into the API itself.

compiler/compiler_test.go
jlapacik: I feel like I've seen this before. I wonder if this is due to inconsistent go versions among the development team.

ast/edit/task_editor.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/codes"
+)
+
+// GetOption finds and returns the AST node for the option value.
+func GetOption(file *ast.File, name string) (ast.Node, error) {
```
jlapacik, line 12: The typescript API returns an Expression here instead of a Node.

ast/edit/task_editor.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return nil, &flux.Error{
+		Code: codes.Internal,
+		Msg:  "Option not found",
+	}
+}
+
+// SetOption replaces an existing option definition with the provided option node or adds
+// the option if it doesn't exist. The pkg AST is mutated in place.
+func SetOption(file *ast.File, id string, opt ast.Node) {
```
jlapacik, line 36: The typescript API takes an Expression type here instead of a Node type.

ast/edit/task_editor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return &flux.Error{
+		Code: codes.Internal,
+		Msg:  fmt.Sprintf("Could not delete %v, Option not found.", name),
+	}
+}
+
+// GetProperty finds and returns the AST node for the property value.
+func GetProperty(obj *ast.ObjectExpression, key string) (*ast.Property, error) {
```
jlapacik, line 87: The typescript API returns an Expression here instead of a Property.

ast/edit/task_editor.go
nathanielc: @jlapacik is the expression returned the VariableAssignment node or the Init expression?

ast/edit/task_editor.go
nathanielc: For this one is the does the Expression represent the property value?

ast/edit/task_editor.go
jlapacik: It's the `init` expression as VariableAssignment is of type `Statement`

ast/edit/task_editor.go
jlapacik: For the typescript API, as it stands right now, yes.

fchikwekwe: I've updated my api to also return an Expression which represents the property value

ast/edit/task_editor.go
fchikwekwe: Updated this as well.

fchikwekwe:
Checking in. I've updated the API and corresponding tests to more closely match the types used in the Typescript API. Let me know what y'all think.

ast/edit/task_editor.go
```diff
@@ -0,0 +1,120 @@
...
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if va.ID.Name == name {
+					file.Body = append(file.Body[:i], file.Body[i+1:]...)
+					return nil
+				}
+			}
+		}
+	}
+
+	return &flux.Error{
```
nathanielc, line 71: Do we want to error if the option is not found? I think callers will not be interested to know if it existed or not previously only that it has been removed if it did. Thoughts?

ast/edit/task_editor.go
```diff
@@ -0,0 +1,120 @@
...
+// DeleteProperty removes a property from the object expression if it exists.
+// The object expression AST is mutated in place.
+func DeleteProperty(obj *ast.ObjectExpression, key string) error {
+	for i, prop := range obj.Properties {
+		if key == prop.Key.Key() {
+			obj.Properties = append(obj.Properties[:i], obj.Properties[i+1:]...)
+			return nil
+		}
+	}
+	return &flux.Error{
```
nathanielc, line 116: Same question here, should we return an error if the property was not found?

ast/edit/task_editor.go
fchikwekwe: Yeah, I noticed that @jlapacik didn't do this in the typescript version and asked him about it. I thought it might be relevant information, but I guess maybe not. I'll change it.

ast/edit/task_editor_test.go
```diff
@@ -0,0 +1,432 @@
...
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/ast/asttest"
+	"github.com/influxdata/flux/ast/edit"
+)
+
+func TestGetOption(t *testing.T) {
```
brettbuddin, line 13: Could we add a test here that represents the actual structure of a `option task = { ... }` statement (with all required fields)? I'm interested to see a flow of `GetOption` and then `GetProperty` to extract meaningful data from that statement. This could even just go in test `ExampleTaskOptionsExtraction` function 😄

ast/edit/task_editor_test.go
```diff
@@ -0,0 +1,453 @@
...
+
+	expr, err := edit.GetProperty(obj.(*ast.ObjectExpression), "b")
+	if err != nil {
+		t.Fatalf("unexpected error retrieving property: %v", err)
+	}
+
+	if want, got := "6", ast.StringFromLiteral(expr.(*ast.StringLiteral)); want != got {
+		t.Errorf("expected \"6\" but got %s", got)
+	}
+}
```
jlapacik, line 32: @brettbuddin were you looking for something like this ^

brettbuddin: 💯 Thanks, Joshua.

ast/helpers.go
```diff
@@ -0,0 +1,50 @@
+package ast
+
+import (
+	"regexp"
+	"time"
+)
+
+func IntegerLiteralFromValue(v int64) *IntegerLiteral {
```
nathanielc, line 8: While these functions are all super trivial they help document that its expected to create values from literals and vice versa.

PR #3009, opened 2020-07-20
fix: monotype normalize

yzhang1991:


libflux/src/flux/lib.rs
```diff
@@ -603,10 +603,56 @@ mod tests {
...
+                    }
+                    if let Some(p) = &mut f.pipe {
+                        self.normalize(&mut p.v);
+                    }
+                    self.normalize(&mut f.retn);
+                }
+                _ => {}
+            }
+        }
+    }
```
nathanielc, line 654: Can you add tests for the normalizer itself?

libflux/src/flux/lib.rs
```diff
@@ -603,10 +603,56 @@ mod tests {
...
+            Self {
+                tv_map: TvarMap::new(),
+                f: Fresher::default(),
+            }
+        }
+
+        pub fn normalize(&mut self, t: &mut MonoType) {
+            match t {
+                MonoType::Var(tv) => {
+                    let v = self.tv_map.entry(*tv).or_insert(self.f.fresh());
```
nathanielc, line 627: This looks a bit odd. Won';t this mean we call `self.f.fresh` everytime we see a type variable even it is already been used? And that means that in those cases the type variable will be skipped.

Maybe a test to uses the same type variable more than once is needed to validate this behavior. 

I think we want the normalizer to be consistent and not skip any type variable numbers in the sequence.

yzhang1991: Oh good catch, I meant to pass in a closure using `or_insert_with()`

libflux/src/flux/lib.rs
```diff
@@ -603,10 +603,56 @@ mod tests {
     use core::parser::Parser;
     use core::semantic::convert::convert_file;
     use core::semantic::env::Environment;
+    use core::semantic::fresh::Fresher;
     use core::semantic::nodes::infer_file;
-    use core::semantic::types::MonoType;
+    use core::semantic::types::{MonoType, Row, Tvar, TvarMap};
     use core::{ast, semantic};
 
+    pub struct MonoTypeNormalizer {
```
jlapacik, line 611: Can you add ` #[cfg(test)]` to this struct?

yzhang1991: The entire struct is inside a mod marked `#[cfg(test)]`

jlapacik: That's great!

libflux/src/flux/lib.rs
```diff
@@ -603,10 +603,99 @@ mod tests {
...
+                tv_map: TvarMap::new(),
+                f: Fresher::default(),
+            }
+        }
+
+        pub fn normalize(&mut self, t: &mut MonoType) {
+            // This is to avoid using self directly inside a closure,
+            // otherwise it will be captured by that closure and the compiler
+            // will complain that closure requires unique access to `self`
+            let f = &mut self.f;
```
nathanielc, line 628: Can this line be moved to the Var case of the match statement?

PR #3010, opened 2020-07-20
feat(execute/table/static): enhance the static table api

jsternberg:
The static tables have been changed in a few ways. The api has changed
to make it easier to build and combine templates in different ways. The
`static.Table` struct has been turned into something that will implement
the `flux.TableIterator` interface so it can be used to more easily
produce multiple tables instead of being limited to one. This allows it
to be used in areas that produce multiple tables.

The diff function has also been changed to only accept table iterators.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


internal/execute/table/buffered_builder_test.go
```diff
@@ -11,142 +12,159 @@ import (
...
-				"_measurement": static.StringKey("m0"),
-				"_field":       static.StringKey("f0"),
-				"_time":        static.Times("2020-01-01T00:00:40Z", 10, 20),
-				"_value":       static.Ints(3, 1, 9),
-			}},
+			in: static.TableGroup{
+				static.StringKey("_measurement", "m0"),
+				static.StringKey("_field", "f0"),
+				static.Table{
+					static.Times("_time", "2020-01-01T00:00:00Z", 10, 20, 30),
```
wolffcm, line 63: So a time can be expressed either as an RFC339 string or an integer?

wolffcm: Ok, I just read the doc comment for `Times`. The integers are offsets from the string times. I guess if an integer value appears first, it's an offset from the epoch.

execute/table/static/static.go
```diff
@@ -533,15 +595,109 @@ func mustTimeValue(v interface{}, offset int64, unit time.Duration) (t int64, ab
...
+type TableBuilder interface {
+	// Build will construct a set of tables using the
+	// template as input.
+	//
+	// The template is a pointer as a builder is allowed
+	// to modify the template. For implementors, the
+	// template pointer must be non-nil.
+	Build(template *[]Column) []flux.Table
+}
+
```
wolffcm, line 608: What happened to `TableMatrix`? I thought that would be in here somewhere.

jsternberg: It should be in there towards the bottom.

wolffcm: Ah, so it is, I was reviewing when you force pushed and it disappeared on me.

PR #3015, opened 2020-07-21
feat(libflux/core/parser): update mod.rs with parse_record 

ChitlangeSahas:
### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written


libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,57 @@ impl Parser {
             MonoType::Invalid
         }
     }
+
+    #[cfg(test)]
+    fn parse_record(&mut self) -> Record {
+        //( "{" [Properties] "}" ) | ( "{" Tvar "with" Properties "}" )
+        let _ = self.expect(TOK_LBRACE);
```
jlapacik, line 584: I think you'll need this token for constructing the base node.

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,57 @@ impl Parser {
...
+        // can be (Properties) or (Tvar "with" Properties)
+        let t = self.peek();
+        if t.lit.to_uppercase() != (t.lit) {
+            let properties = self.parse_properties();
+            let record = Record {
+                base: self.base_node_from_token(&t),
```
jlapacik, line 590: I think you'll need to use `base_node_from_tokens` here and pass in the tokens for the right and left braces.

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,57 @@ impl Parser {
...
         }
     }
+
+    #[cfg(test)]
+    fn parse_record(&mut self) -> Record {
+        //( "{" [Properties] "}" ) | ( "{" Tvar "with" Properties "}" )
+        let _ = self.expect(TOK_LBRACE);
+        // can be (Properties) or (Tvar "with" Properties)
+        let t = self.peek();
+        if t.lit.to_uppercase() != (t.lit) {
```
jlapacik, line 587: I don't think this will parse `{A:int}` correctly.

libflux/src/core/ast/mod.rs
```diff
@@ -531,7 +531,24 @@ pub struct TypeExpression {
...
     pub monotype: MonoType,
 }
-
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct Record {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
+    pub tvar: Option<MonoType>,
```
jlapacik, line 540: If we're checking to make sure this is a type variable instead of just an identifier, then this should probably be

    pub tvar: Option<TvarType>,

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,57 @@ impl Parser {
...
+        }
+        return Some(properties);
+    }
+    #[cfg(test)]
+    fn parse_property(&mut self) -> PropertyType {
+        let identifier = self.parse_identifier(); // identifier
+        self.expect(TOK_COLON); // :
+        let monotype = self.parse_monotype();
+        PropertyType {
+            base: base_from_monotype(&monotype),
```
jlapacik, line 626: The base node here should extend from the identifier to the monotype.

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,76 @@ impl Parser {
...
     }
+
+    // Record = "{" [ Identifier (Suffix1 | Suffix2) ] "}"
+    // Suffix1 = ":" MonoType { "," Property }
+    // Suffix2 = "with" [Properties]
+    #[cfg(test)]
+    fn parse_record(&mut self) -> Record {
+        let start = self.open(TOK_LBRACE, TOK_RBRACE);
+        let mut properties: Option<Vec<PropertyType>> = None;
+        let property: PropertyType;
```
jlapacik, line 588: No need for this forward declaration.

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,76 @@ impl Parser {
...
+
+        let mut id: Option<Identifier> = None;
+        let _id;
```
jlapacik, line 591: No need for this forward declaration either.

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,76 @@ impl Parser {
...
+        let t = self.peek();
+        if t.tok == TOK_IDENT {
+            // Indentifier
+            _id = self.parse_identifier();
+            // suffix one needs attention
+            let t2 = self.peek();
+            if t2.tok == TOK_COLON {
+                let mut ps = Vec::<PropertyType>::new();
+                self.consume(); // :
```
jlapacik, line 600: Can you use `self.expect` here for edge cases?

libflux/src/core/parser/mod.rs
```diff
@@ -577,6 +577,76 @@ impl Parser {
...
+        }
+        return Some(properties);
+    }
+    #[cfg(test)]
+    fn parse_property(&mut self) -> PropertyType {
+        let identifier = self.parse_identifier(); // identifier
+        self.expect(TOK_COLON); // :
+        let monotype = self.parse_monotype();
+        PropertyType {
+            base: base_from_monotype(&monotype),
```
jlapacik, line 645: Can you change this to `self.base_node_from_others(&identifier.base, &monotype.base)`?

libflux/src/core/parser/tests.rs
```diff
@@ -1466,6 +1466,172 @@ fn test_parse_type_expression_regexp() {
...
+        Record {
+            base: BaseNode {
+                location: loc.get(1, 1, 1, 16),
+                ..BaseNode::default()
+            },
+            tvar: None,
+            properties: Some(vec![
+                PropertyType {
+                    base: BaseNode {
+                        location: loc.get(1, 4, 1, 7),
```
jlapacik, line 1485: Similar to my previous comment. The source location for this property should correspond to the entire property `a:int`, but right now it just refers to the location of the monotype `int`.

libflux/src/core/parser/mod.rs
ChitlangeSahas: Fixed!

libflux/src/core/parser/mod.rs
ChitlangeSahas: I could be wrong about this one, how would I reference it in the else block when it goes out of scope?

libflux/src/core/parser/mod.rs
ChitlangeSahas: ✅

libflux/src/core/parser/mod.rs
ChitlangeSahas: 👍

libflux/src/core/parser/mod.rs
jlapacik: You assign it in the outermost if block, so it's still visible from inside the inner if/else blocks.

ChitlangeSahas: oh very true.

libflux/src/core/parser/mod.rs
jlapacik: Oops, you already check for a `:` before consuming. My bad.

PR #3016, opened 2020-07-21
fix(execute): multiple do calls will fail with the empty table

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -2196,21 +2196,21 @@ type emptyTable struct {
 // NewEmptyTable constructs a new empty table with the given
 // group key and columns.
 func NewEmptyTable(key flux.GroupKey, cols []flux.ColMeta) flux.Table {
-	return emptyTable{
+	return &emptyTable{
 		key:  key,
 		cols: cols,
 	}
 }
 
-func (t emptyTable) Key() flux.GroupKey {
+func (t *emptyTable) Key() flux.GroupKey {
 	return t.key
 }
 
-func (t emptyTable) Cols() []flux.ColMeta {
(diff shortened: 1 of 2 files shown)


adrian-thurston:
Was experimenting with this defect for my own learnings, so I went a bit further and made it into a test case.
```
func TestEmptyTable(t *testing.T) {
    gk := execute.NewGroupKey(
        []flux.ColMeta{
            {Label: "host", Type: flux.TString},
        },
        []values.Value{
            values.NewString("a"),
        }, 
    )

    cols := []flux.ColMeta{
        {Label: "host", Type: flux.TString},
    }   
    
    ec := execute.NewEmptyTable( gk, cols )
    
    var ran int = 0
    df := func(flux.ColReader) error {
        ran += 1
        return nil
    }
    
    err1 := ec.Do(df)
    err2 := ec.Do(df)
    err3 := ec.Do(df)
    
    // Expecting just one call to df().
    if ran == 0 {
        t.Error("df() did not run")
    }
    if ran > 1 {
        t.Error("df() ran more than once")
    }
    if err1 != nil || err2 == nil || err3 == nil {
        t.Error("expected nil, Error, Error from 3 do calls, but got something else" )
    }
}

jsternberg:
I updated with test cases. We have a suite of test cases that we can run against table implementations to make sure that they implement the interface correctly since this interface can be quite difficult to implement. The only issue I had with it originally was that the test checks that the allocator was used, but this specific table implementation never uses the allocator on purpose. So I added a little hack with a comment to indicate that the allocator wasn't intended to be used.

PR #3017, opened 2020-07-22
feat(stdlib/planner): add planner tests for window min and max

jsternberg:
Related to influxdata/influxdb#18635.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
@wolffcm it might not for `mean` but I think it's good anyway. The reason why I added it is because we specifically ran into a bug with selectors when we did first and last so I wanted to cover all of my bases.

PR #3029, opened 2020-07-25
feat(libflux/core/parser): add parse_function in parser/mod.rs. 

ChitlangeSahas:
Add FunctionType and ParameterType struct in ast/mod.rs.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/src/core/ast/mod.rs
```diff
@@ -537,6 +538,49 @@ pub struct ArrayType {
...
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct ParameterType {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
+    pub identifier: Option<Identifier>,
+    pub parameter: MonoType,
+    pub _type: Option<PType>,
+}
```
jlapacik, line 560: Can we make this an enum instead?
```Rust
pub enum ParameterType {
    Required {
        base: BaseNode,
        name: String,
        ty: MonoType,
    },
    Optional {
        base: BaseNode,
        name: String,
        ty: MonoType,
    },
    Pipe {
        base: BaseNode,
        name: Option<String>,
        ty: MonoType,
    },
}
```
This way the type of the parameter - required, optional, or pipe - is encoded as part of the Rust type, and we don't need to keep the original tokens around.

libflux/src/core/ast/mod.rs
```diff
@@ -537,6 +538,49 @@ pub struct ArrayType {
...
+    pub base: BaseNode,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct QuestionMarkSymbol {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
+}
```
jlapacik, line 582: If we make `ParameterType` an enum, these types become redundant and we can remove them.

libflux/src/core/ast/mod.rs
```diff
@@ -537,6 +538,49 @@ pub struct ArrayType {
     pub monotype: MonoType,
 }
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct FunctionType {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
+    pub parameters: Option<Vec<ParameterType>>,
```
jlapacik, line 547: This just hit me right now, but wrapping this vector in an option is redundant. If the function doesn't take any input then this can just be an empty (length zero) vector.

PR #3030, opened 2020-07-27
feat: add query plan to query metadata

yzhang1991:
`Metadata` is a key-value map stored with the query statistics.
It currently contains data for the sources, for example, "influxdb/scanned-bytes" and "influxdb/scanned-values".
The goal here is to add the query plan to query metadata, which is later accessed to generate the `query_log` entry.

Queries with functions like `tableFind()` and `experimental.chain()` will call into the execution engine more than once. The statistics of those additional calls won't be preserved so their query plans are lost.
The solution for it is temporarily caching their metadata in the execution dependency, and we will extract them at the time when the main query is executed. 

In this PR, `flux.Metadata` is moved to a standalone package to address a dependency cycle issue: `influxd -> flux -> flux/interpreter -> flux/lang/execdeps -> flux` 

This PR also adds `Detailer` implementations to several procedure specs. The `Detailer` interface is used to add additional details to the formatted query plan output. For example, the aggregate function that is being pushed down with the `ReadGroup` or the `ReadWindowAggregate` call, the real implementation that is being used for transformations that have dual implementations.

### DIFF
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/metadata"
 	"github.com/influxdata/flux/plan"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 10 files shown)


yzhang1991:
@adrian-thurston Actually, I was thinking about just replacing the `Metadata` with its real underlying type:
```go
type Metadata map[string][]interface{}
```

For this particular use, I think it is good enough, thoughts?

adrian-thurston:
@yzhang1991 I think I tried that, but hit a snag somewhere. I can't remember exactly what that was ... but I think It cascaded to more changes somewhere else I didn't want to touch.

PR #3031, opened 2020-07-27
feat(execute/function): create an alternative for constructing new transformations

jsternberg:
This code tries to simplify creating new transformations by using
reflection to read arguments rather than custom code that looks very
similar. This method should work for a majority of transformations where
creating the operation spec is a simple mapping of types to their Go
equivalents and potentially a bit more.

The idea is that you create a single struct called the
`TransformationSpec`. This is a `ProcedureSpec` that knows how to create
a transformation from itself. The `TransformationSpec` can define and
customize its parameters using struct tags labeled as `flux` or it can
use the defaults as described in the `ReadArgs` documentation. That can
make implementing some transformations more simple. As an example,
`max()` could be implemented by doing:

    type MaxProcedureSpec struct {
        Tables *functions.TableObject `flux:"<-"`
        Column string
    }

    func (s *MaxProcedureSpec) CreateTransformation(id execute.DatasetID, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
        // construct the transformation using the arguments
    }

This work is experimental and incomplete. Please be sure to thoroughly
test the transformation with different arguments when testing. The
intention is to create a foundation that makes creating transformations
easier, less code intense, and easier to document and teach.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/execute/function/args.go
+++ b/internal/execute/function/args.go
@@ -1,0 +1,283 @@
+package function
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"unicode"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/interpreter"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+// Arguments defines the interface for reading an Argument.
+type Arguments interface {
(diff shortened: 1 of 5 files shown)


jsternberg:
Yea, the purpose of this is to hopefully get rid of a lot of the boilerplate for creating a new transformation. That way it hopefully also becomes easier to teach others how to make new transformations without just saying, "Copy and paste an existing transformation."

PR #3033, opened 2020-07-27
fix: data race in metadata

yzhang1991:


adrian-thurston:
@yzhang1991 What's the race here? I was under the impression it was clean WRG to concurrency. Results channel and metadata channel should both be drained and closed at this stage.

adrian-thurston:
@yzhang1991 oh I think I see now. The query results channel can be closed before the metadata channel is drained. The query's `Done` function is never called from `tableFind`. I wonder why that is. The synchronization mechanism is there, but not used.

PR #3044, opened 2020-07-28
feat(ast/edit): get all options and props

fchikwekwe:
Updates options/properties api to be able to blindly return all options and properties.

### DIFF
--- a/ast/edit/task_editor.go
+++ b/ast/edit/task_editor.go
@@ -27,6 +27,22 @@ func GetOption(file *ast.File, name string) (ast.Expression, error) {
 	}
 }
 
+// GetAllOptions returns all options within a file
+func GetAllOptions(file *ast.File) []ast.Expression {
+	var exps []ast.Expression
+	for _, st := range file.Body {
+		if val, ok := st.(*ast.OptionStatement); ok {
+			assign := val.Assignment
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if ok && va != nil {
+					exps = append(exps, va.Init)
+				}
+			}
+		}
+	}
+	return exps
+}
+
 // SetOption replaces an existing option's init with the provided init or adds
 // the option if it doesn't exist. The file AST is mutated in place.
 func SetOption(file *ast.File, name string, expr ast.Expression) {
@@ -78,6 +94,17 @@ func GetProperty(obj *ast.ObjectExpression, key string) (ast.Expression, error) 
 		Code: codes.Internal,
 		Msg:  "Property not found",
 	}
+}
+
+// GetAllProperties returns all properties for a given option
+func GetAllProperties(obj *ast.ObjectExpression) []ast.Expression {
+	var exps []ast.Expression
+	for _, prop := range obj.Properties {
+		if prop.Value != nil {
+			exps = append(exps, prop.Value)
+		}
+	}
+	return exps
 }
 
 // SetProperty replaces an existing property definition with the provided object expression or adds
--- a/ast/edit/task_editor_test.go
+++ b/ast/edit/task_editor_test.go
@@ -5,9 +5,11 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/ast/asttest"
 	"github.com/influxdata/flux/ast/edit"
+	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/parser"
 )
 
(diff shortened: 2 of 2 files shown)


ast/edit/task_editor.go
```diff
@@ -80,6 +102,21 @@ func GetProperty(obj *ast.ObjectExpression, key string) (ast.Expression, error)
 	}
 }
 
+// GetAllProperties returns all properties for a given option
+func GetAllProperties(obj *ast.ObjectExpression) ([]ast.Expression, error) {
```
jsteenb2, line 106: I'm looking at the `ast.Expression` type, which looks like an extension interface around the `ast.Node` type, which looks like:

```go
// Node represents a node in the InfluxDB abstract syntax tree.
type Node interface {
	node()
	Type() string // Type property is a string that contains the variant type of the node
	Location() SourceLocation
	Errs() []Error
	Copy() Node

	// All node must support json marshalling
	json.Marshaler
}
```

how does one pull out the key of the option properties from the returned `[]ast.Expression`?

ast/edit/task_editor.go
```diff
@@ -27,6 +27,28 @@ func GetOption(file *ast.File, name string) (ast.Expression, error) {
...
+		if val, ok := st.(*ast.OptionStatement); ok {
+			assign := val.Assignment
+			if va, ok := assign.(*ast.VariableAssignment); ok {
+				if ok {
+					exps = append(exps, va.Init)
+				}
+			}
+		}
+	}
+	if len(exps) == 0 {
```
rockstar, line 43: What is the case where this (and the other check for `len(...) == 0`) could happen? Is it really an error?

ast/edit/task_editor.go
fchikwekwe: `ast.Expression` types can be type converted to their underlying ast Node type. [Something like this](https://github.com/influxdata/flux/blob/master/ast/edit/task_editor_test.go#L24) on lines 24 and 29. If the type conversion fails, then you're converting it to the wrong ast Node type.

fchikwekwe: [Here's a place where I'm assuming types within the api.](https://github.com/influxdata/flux/blob/master/ast/edit/task_editor.go#L9-L28)

jsteenb2: ah ok. I think I'm following. Do I look at the `Type` method on Node do determine which type assert to use, or a type switch?

ast/edit/task_editor.go
fchikwekwe: This is just a way to check if there are no options or no properties when the caller checks for them. Maybe a bool would be better here than an error, but there needs to be a way to tell them that nothing was found. 

This could certainly happen. If you had a script with no options declared, then the len of the expressions array would be zero.

rockstar: Can we not return an empty list? That's a valid value, isn't it?

ast/edit/task_editor.go
fchikwekwe: yeah, I _think_ you could call the `Type()` method if you don't already know the Node type and then assert. You could switch on the result of the called to `Type()` if you have a predetermined list of possible types.

ast/edit/task_editor.go
jsteenb2: From a consumer's perspective, I'd reason the empty slice would being an adequate signal that options don't exist in the query for GetAllOptions or GetAllProperties 🤔.

fchikwekwe: 🤷🏾‍♀️ I mean yeah its valid. I guess it depends on whether the use would rather check for an error or check the length of the returned array themselves.

fchikwekwe: @jsteenb2 do you have an opinion on this?

fchikwekwe: Oh. Some conversation overlap. That's fine. I can change this.

jsteenb2: If I'm using GetAll, I wouldnt' sue either. I'd just loop over whatever options/properties I am provided. GetAllOptions I dont' have a use case for, but GetAllProperties for an option, I'll use a lot of.

fchikwekwe: yep got it. I'll go ahead and update this.

rockstar: I don't think an error is appropriate if having no options is valid. I think a good indicator is thinking about an error message like "Error: a valid thing occurred" is a good indicator that an error isn't the right thing here.

fchikwekwe: Updated. Not returning an error anymore. Lemme know what y'all think.

ast/edit/task_editor.go
```diff
@@ -80,6 +96,17 @@ func GetProperty(obj *ast.ObjectExpression, key string) (ast.Expression, error)
 	}
 }
 
+// GetAllProperties returns all properties for a given option
+func GetAllProperties(obj *ast.ObjectExpression) []ast.Expression {
```
jsteenb2, line 100: 👌 perfecto

fchikwekwe:
@rockstar @jsteenb2 Turns out that we didn't really need this change since we'll almost always know which option we're targeting within templates. Gonna revert this change for now.

PR #3045, opened 2020-07-28
feat(libflux/ast/tests.rs): add test_json_namedtype to tests.rs.

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/ast/tests.rs
```diff
@@ -395,6 +395,21 @@ fn test_json_expression_statement() {
...
+fn test_json_namedtype() {
+    let n = MonoType::Basic(NamedType {
+        base: BaseNode::default(),
+        name: Identifier {
+            base: BaseNode::default(),
+            name: "int".to_string(),
+        },
+    });
+    let serialized = serde_json::to_string(&n).unwrap();
+    assert_eq!(serialized, r#"{"Basic":{"name":{"name":"int"}}}"#);
```
jlapacik, line 409: I think this should look like:

    {"type":"NamedType","name":{"type":"Identifier","name":"int"}}

In particular, every json element should have a `type` field. In order to do that you'll need to add `#[serde(tag = "type")]` to the MonoType enum. So the MonoType enum should look like this:
```Rust
#[serde(tag = "type")]
pub enum MonoType {
    #[serde(rename = "NamedType")]
    Basic(NamedType),
    ....
```

yuktarakesh: Do I replace #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)] with #[serde(tag = "type")] or keep both?

jlapacik: Keep both. You can add `#[serde(tag = "type")]` right below the first one.

PR #3046, opened 2020-07-28
feat: Go ast types for flux type expression syntax

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


ast/ast.fbs
```diff
@@ -1,5 +1,64 @@
...
+  kind:ParameterKind;
+}
+
+table FunctionType {
+  base_node:BaseNode;
+  parameters:[ParameterType];
+  retn:MonoType;
+}
+
+table Constraint {
```
nathanielc, line 50: Do we want to name this `TypeConstraint`? and maybe rename the Rust struct accordingly? Its a small thing but once we release it we can't easily change it.

ast/ast.go
```diff
@@ -160,6 +160,93 @@ func (e Error) Error() string {
...
+	return "TypeExpression"
+}
+
+type MonoType interface {
+	Node
+	monotype()
+}
+
+type NamedType struct {
+	*Identifier
```
nathanielc, line 179: Why no BaseNode here?

ast/ast.fbs
jlapacik: I agree.

jlapacik:
@nathanielc would you like to give this a review? I've only finished the flatbuffer deserialization bits on the Go side, but it's probably much easier to review in pieces.

PR #3048, opened 2020-07-28
feat(libflux/ast/tests.rs): add test_json_tvartype to tests.rs.

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/ast/tests.rs
```diff
@@ -395,6 +395,18 @@ fn test_json_expression_statement() {
...
     assert_eq!(deserialized, n)
 }
+
+#[test]
+fn test_json_tvartype() {
+    let n = MonoType::Tvar(TvarType {
+        base: BaseNode::default(),
+    });
+    let serialized = serde_json::to_string(&n).unwrap();
+    assert_eq!(serialized, r#"{"type":"TvarType"}"#);
```
jlapacik, line 405: I must have missed this in review, but TvarType should have an identifier just like NamedType. Are you able to update it? The json should look like:

    {"type":"TvarType", "name":{"name": ...}}

yuktarakesh: The test cases should also change in that case right?

jlapacik: Yes.

jlapacik: It looks like there might only be one test case that needs to be updated.

libflux/go/libflux/buildinfo.gen.go
```diff
@@ -24,15 +24,17 @@ var sourceHashes = map[string]string{
 	"libflux/src/core/ast/flatbuffers/tests.rs":                                     "131f266613fd39cfcbc72c41691b68f58a14ee3a994d45f011da70c263f16f26",
 	"libflux/src/core/ast/mod.rs":                                                   "d9d3b7a43942d3e9fbdb4866d30d4f5ed92767c1ce471f2cbfe96959ebce130f",
 	"libflux/src/core/ast/tests.rs":                                                 "c48be4a150643298be2806fc7c3be5d770e6d7d73b80193f0ea45249b4c7aeb9",
+	"libflux/src/core/ast/mod.rs":                                                   "e3e7d18f8074bccb0dfc1acbfdd2cc93891ab76c895e6fd951919307f79a6d36",
+	"libflux/src/core/ast/tests.rs":                                                 "524e07ece164e54f57904cdde18bb8c01033e91faeba64202f477153110ef350",
```
jlapacik, line 28: This file should always have the same number of deletions as additions. The end result should be one line per rust source file. See how you have two lines for each of `libflux/src/core/ast/tests.rs` and `libflux/src/core/ast/mod.rs`. This will result in a `checkgenerate` failure.

PR #3052, opened 2020-07-29
feat(libflux/ast/tests.rs): add test_json_functiontype to tests.rs

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/ast/mod.rs
```diff
@@ -555,18 +556,31 @@ pub struct FunctionType {
...
         name: Identifier,
         ty: MonoType,
     },
+    #[serde(rename = "Pipe")]
     Pipe {
+        #[serde(skip_serializing_if = "BaseNode::is_empty")]
+        #[serde(default)]
+        #[serde(flatten)]
         base: BaseNode,
         name: Option<Identifier>,
```
jlapacik, line 585: Could you add `#[serde(skip_serializing_if = "Option::is_none")]` right above this line so that we don't include a `name` field in the json if this is an un-named pipe parameter?

PR #3076, opened 2020-08-04
feat(libflux/semantic/convert.rs): add convert_polytype

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/core/semantic/convert.rs
+++ b/libflux/src/core/semantic/convert.rs
@@ -3,7 +3,6 @@ use crate::semantic::fresh::Fresher;
 use crate::semantic::nodes::*;
 use crate::semantic::types;
 use crate::semantic::types::MonoType;
-use crate::semantic::types::Kind;
 use crate::semantic::types::MonoTypeMap;
 use crate::semantic::types::SemanticMap;
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


libflux/src/core/semantic/convert.rs
```diff
@@ -233,6 +234,59 @@ fn convert_monotype(
...
+    let mut tvars = HashMap::<String, types::Tvar>::new();
+    let mut expr = convert_monotype(type_expression.monotype, &mut tvars, f)?;
+    let mut vars = Vec::<types::Tvar>::new();
+    let mut cons = SemanticMap::<types::Tvar, Vec<types::Kind>>::new();
+
+    for c in type_expression.constraints {
+        if tvars.contains_key(c.tvar.name.as_str()) {
+            vars.push(match tvars.get(&c.tvar.name) {
+                Some(v) => *v,
+                None => {types::Tvar(0)} // never reached.
```
jlapacik, line 264: This should return an error.
```Rust
None => Err(format!("type variable {} constrained but not used", ...))
```

May I suggest something along the lines of the following (more or less):
```Rust
for c in type_expression.constraints {
    match tvars.remove(c.tvar.name) {
        None => {
            return Err(...);
        }
        Some(tv) => {
            vars.push(tv);
        }
    }
...
```
In other words, just remove the element from the map. If `None` then throw an error.

PR #3088, opened 2020-08-09
added monotype copy functionality using reflect. <stackoverflow>

ChitlangeSahas:
Added Copy functionality to monotype using reflect.


nathanielc:
@ChitlangeSahas I am curious why using reflect here helps. Can you provide an explanation? I see this was merged into another feature branch so I could review there as well, but figured its easier to have a more focused conversation here.

PR #3098, opened 2020-08-11
feat(stdlib/sql): add SAP HANA db support

alespour:
This PR add SAP HANA db support to Flux.

It uses older `v0.14.1` Go driver (Mar 16, 2019), because it is the last version that works with Go 1.12.

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/sql/hdb.go
+++ b/stdlib/sql/hdb.go
@@ -24,7 +24,6 @@ import (
 // * TIMESTAMP does not TZ info stored, but:
 //   - it is "strongly discouraged" to store data in local time zone: https://blogs.sap.com/2018/03/28/trouble-with-time/
 //   - more on timestamps in HDB: https://help.sap.com/viewer/f1b440ded6144a54ada97ff95dac7adf/2.4/en-US/a394f75dcbe64b42b7a887231af8f15f.html
-//   - the driver does is not capable of converting date- or time-formatted strings to time.Time when inserting to DATE or TIME columns
 //   Therefore TIMESTAMP is mapped to TTime and vice-versa here.
 // * the hdb driver is rather strict, eg. does not convert date- or time-formatted string values to time.Time,
 //   or float64 to Decimal on its own and just throws "unsupported conversion" error
--- a/stdlib/sql/hdb.go
+++ b/stdlib/sql/hdb.go
@@ -189,8 +189,8 @@ func HdbColumnTranslateFunc() translationFunc {
 // Template for conditional query by table existence check
 var hdbDoIfTableNotExistsTemplate = `DO
 BEGIN
-	DECLARE SCHEMA_NAME NVARCHAR(%d) = '%s';
-	DECLARE TABLE_NAME NVARCHAR(%d) = '%s';
+    DECLARE SCHEMA_NAME NVARCHAR(%d) = '%s';
+    DECLARE TABLE_NAME NVARCHAR(%d) = '%s';
     DECLARE X_EXISTS INT = 0;
     SELECT COUNT(*) INTO X_EXISTS FROM TABLES %s;
     IF :X_EXISTS = 0
--- a/stdlib/sql/hdb.go
+++ b/stdlib/sql/hdb.go
@@ -88,7 +88,11 @@ func (m *HdbRowReader) GetNextRow() ([]values.Value, error) {
 				}
 				newFloat, _ := (*big.Rat)(&out).Float64()
 				row[i] = values.NewFloat(newFloat)
-			default:
+			default: // flux.TString
+				switch m.sqlTypes[i].DatabaseTypeName() {
+				case "BINARY", "VARBINARY":
+					return nil, errors.Newf(codes.Invalid, "Flux does not support column type %s", m.sqlTypes[i].DatabaseTypeName())
+				}
 				row[i] = values.NewString(string(value))
 			}
 		case nil:
--- a/stdlib/sql/hdb.go
+++ b/stdlib/sql/hdb.go
@@ -213,7 +213,7 @@ func hdbAddIfNotExist(table string, query string) string {
 	var where string
 	var args []interface{}
 	parts := strings.SplitN(strings.ToUpper(table), ".", 2) // schema and table name assumed uppercase in HDB by default (see Notes)
-	if len(parts) == 2 { // fully-qualified table name
+	if len(parts) == 2 {                                    // fully-qualified table name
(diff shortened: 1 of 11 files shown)


stdlib/sql/hdb.go
```diff
@@ -0,0 +1,211 @@
...
+// Adds SAP HANA specific table existence check to CREATE TABLE statement
+func hdbAddIfNotExist(table string, query string) string {
+	var where string
+	parts := strings.SplitN(table, ".", 2)
+	if len(parts) == 2 { // fully-qualified table name
+		where = fmt.Sprintf("WHERE SCHEMA_NAME=UPPER('%s') AND TABLE_NAME=UPPER('%s')", parts[0], parts[1])
+	} else { // table in user default schema
+		where = fmt.Sprintf("WHERE TABLE_NAME=UPPER('%s')", table)
+	}
+	return fmt.Sprintf(hdbDoIfTableNotExistsTemplate, where, query)
```
wolffcm, line 210: I'm a little worried that someone could craft a table name that contains characters like `'; ...` or something and do something nefarious. [example](https://xkcd.com/327/)

Do you have any thoughts on this? For some of the places where we use `%s` I imagine that we could use parameters `?` instead?

wolffcm (changes requested):
Overall, looks great as usual, I just had some questions.

stdlib/sql/hdb.go
```diff
@@ -0,0 +1,211 @@
...
+	flux.TBool:   "BOOLEAN",
+	flux.TTime:   "TIMESTAMP", // not exactly correct (see Notes)
+}
+
+// HdbTranslateColumn translates flux colTypes into their corresponding SAP HANA column type
+func HdbColumnTranslateFunc() translationFunc {
+	return func(f flux.ColType, colName string) (string, error) {
+		s, found := fluxToHdb[f]
+		if !found {
+			return "", errors.Newf(codes.Internal, "SAP HANA does not support column type %s", f.String())
```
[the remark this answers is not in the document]
alespour: I see. I'll fix it here. This can be found in `translationFunc` of all other databases as well.  I've added it to the list of fixes and improvements that should be done to `sql` package that I keep in bonitoo Flux fork https://github.com/bonitoo-io/flux/issues/1

stdlib/sql/hdb.go
```diff
@@ -0,0 +1,211 @@
...
+			case flux.TFloat:
+				var out hdb.Decimal
+				err := out.Scan(value)
+				if err != nil {
+					return nil, err
+				}
+				newFloat, _ := (*big.Rat)(&out).Float64()
+				row[i] = values.NewFloat(newFloat)
+			default:
+				row[i] = values.NewString(string(value))
```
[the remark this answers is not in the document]
alespour: I'm not sure if there is anything else to do, because when I do just this change, I get `Error: panic: unexpected value type bytes` in `sql.from`.

stdlib/sql/hdb.go
alespour: I could not get ? placeholder to work with the SQL DO script (`SQL Error 1287 - identifier must be declared`), but I changed it to use declared variables for schema and table. The resulting query now looks like eg.
```
DO
BEGIN
    DECLARE SCHEMA_NAME NVARCHAR(11) = 'bike_stores';
    DECLARE TABLE_NAME NVARCHAR(5) = 'xcopy';
    DECLARE X_EXISTS INT = 0;
    SELECT COUNT(*) INTO X_EXISTS FROM TABLES WHERE SCHEMA_NAME=UPPER(:SCHEMA_NAME) AND TABLE_NAME=UPPER(:TABLE_NAME);
    IF :X_EXISTS = 0
    THEN
        CREATE TABLE bike_stores.xcopy (NOTE NVARCHAR(5000),...);
    END IF;
END;
```

stdlib/sql/hdb.go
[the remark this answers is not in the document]
wolffcm: Ah, I see. I guess that we do not support `bytes` values inside of tables yet, and that's what why it's not there in `flux.ColType`. I don't think that mapping `BINARY` to `string` is the right thing to do. We want our strings to be valid UTF8, and it's not necessarily the case for `BINARY`. 

So, I think we need to return an error if we encounter a `BINARY` column for now. Can you do that for HDB as part of this PR?

We can potentially fix this later when we have support for `bytes` in Flux tables.

I see that there are several other DBs where we do this, so I guess we'll want to fix them as well as part of a separate issue, which I have filed: https://github.com/influxdata/flux/issues/3209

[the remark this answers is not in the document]
alespour: Done, please see above.

stdlib/sql/hdb.go
wolffcm: I think this is better but it still seems like the `CREATE TABLE` statement allows for an injection.

If I were to do something like `sql.to(table: "foo (i INTEGER); DROP TABLE bar; ...")`, then there's still a problem, if I'm understanding right.

What if you always quoted the identifiers in the CREATE TABLE statement:
```
CREATE TABLE "bike_stores"."xcopy" ("NOTE" NVARCHAR(5000),...);
```
And if any of the identifiers contain a `"`, you can either escape it, or produce a user error. Does that make sense?

alespour: Improved, I hope :) Since by default object names are converted to uppercase by HDB unless quoted, and quotes are used to escape possibly malign SQL code here and not to mark the value as case-sensitive, uppercase is assumed and used for **new** output table name, which I think is conforming and safe.

Table created like bellow then allows a user to insert data both with any of the following statements:

`INSERT INTO bike_stores.orders_copy_q2 ... (note, ...) VALUES (...)`
`INSERT INTO BIKE_STORES.ORDERS_COPY_Q2 ... (NOTE, ...) VALUES (...)`.
`INSERT INTO "BIKE_STORES"."ORDERS_COPY_Q2" ... ("NOTE", ...) VALUES (...)`.

The if-not-exist-then-create now looks like this:

```DO
BEGIN
    DECLARE SCHEMA_NAME NVARCHAR(11) = 'BIKE_STORES';
    DECLARE TABLE_NAME NVARCHAR(14) = 'ORDERS_COPY_Q2';
    DECLARE X_EXISTS INT = 0;
    SELECT COUNT(*) INTO X_EXISTS FROM TABLES WHERE SCHEMA_NAME=ESCAPE_DOUBLE_QUOTES(:SCHEMA_NAME) AND TABLE_NAME=ESCAPE_DOUBLE_QUOTES(:TABLE_NAME);
    IF :X_EXISTS = 0
    THEN
        CREATE TABLE "BIKE_STORES"."ORDERS_COPY_Q2" ("NOTE" NVARCHAR(5000),...);
    END IF;
END;

wolffcm: This is much better. Thanks for the attention to this.

PR #3100, opened 2020-08-11
fix(deps): make socket/sql url test robust

nathanielc:
The socket and sql url validator tests would fail on linux systems
because the message returned from Go depended on the specific DNS
resolver being used. The test case now validates that the error comes
form the correct call to the validation path but does not otherwise care
about the specific wording used to indicate a DNS resolution failure.

Additionally the hostname used is guaranteed to not exist based on RFC
6761.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
As an aside it was odd to me that the tests failing were the socket and sql tests, shouldn't we have this test coverage as part of the PrivateIPValidator unit test suite and then for the socket and sql tests simply validate that the url.Validator was called but not care why it failed?

I didn't change that because its  a bigger change than I have time for ATM but thought I'd at least call it out.

PR #3102, opened 2020-08-11
feat: add nulls parameter to gen.tables()

yzhang1991:


libflux/src/core/semantic/builtins.rs
```diff
@@ -191,7 +191,7 @@ pub fn builtins() -> Builtins<'static> {
                 "#,
             },
             "internal/gen" => semantic_map! {
-                "tables" => "forall [t0] (n: int, tags: [{name: string | cardinality: int}]) -> [{_time: time | _value: float | t0}]",
+                "tables" => "forall [t0] (n: int, ?nulls: float, tags: [{name: string | cardinality: int}]) -> [{_time: time | _value: float | t0}]",
```
jsternberg, line 194: While you're here, I've been meaning to fix this. Can you make tags optional? It's supposed to be optional but the type system has it specified incorrectly.

PR #3104, opened 2020-08-11
fix: calculate distinct key values

jlapacik:
Previously `keyValues` would only output one row per distinct value among all of the `KeyColumns`. Now it will output one row per distinct value per each `KeyColumns` value. The necessary changes included updating the following map from
```Go
var intDistinct = map[int64]bool{}
```
to
```Go
var intDistinct = map[struct {
    string
    int64
}]bool{}
```

Fixes https://github.com/influxdata/influxdb/issues/19134.

### DIFF
--- a/stdlib/universe/key_values.go
+++ b/stdlib/universe/key_values.go
@@ -146,31 +146,42 @@ func (t *keyValuesTransformation) Process(id execute.DatasetID, tbl flux.Table) 
 
 	// TODO: use fn to populate t.spec.keyColumns
 
-	cols := tbl.Cols()
-	i := 0
-	keyColIndex := -1
-	for keyColIndex < 0 && i < len(t.spec.KeyColumns) {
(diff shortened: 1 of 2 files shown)


stdlib/universe/key_values.go
```diff
@@ -195,44 +206,38 @@ func (t *keyValuesTransformation) Process(id execute.DatasetID, tbl flux.Table)
...
 			// Check distinct
-			for j, rowIdx := range keyColIndices {
-				if rowIdx < 0 {
-					continue
-				}
+			for _, c := range keyColumns {
 				switch keyColType {
 				case flux.TBool:
-					vs := cr.Bools(rowIdx)
+					vs := cr.Bools(c.idx)
```
rockstar, line 240: I started reviewing this patch, got to this point, and thought "Am I reviewing generated code?" Kudos to you for manually handling this giant type switch.

jlapacik: No problem.

PR #3105, opened 2020-08-11
feat(universe/group.go): add OrderFilterGroup to group.go as well as order_filter_group_flag_off_test.flux and order_filter_group_flag_on_test.flux

yuktarakesh:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/universe/order_filter_group_flag_on_test.flux
```diff
@@ -0,0 +1,36 @@
...
+,,2,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,2018-05-22T19:53:46Z,system,host.local,load4,1.77
+"
+
+order_fn = (tables=<-) => tables
+    |> range(start: 2018-05-22T19:53:26Z)
+    |> group(columns: ["_time"])
+    |> filter(fn: (r) => r["_field"] == "load4")
+
+test order_evaluate = () =>
+    ({input: testing.loadStorage(csv: input), want: testing.loadMem(csv: output), fn: order_fn})
```
wolffcm, line 36: These tests are great, could I trouble you to add a few more:
- tests similar to this but have `onEmpty: "keep"` in the call to `filter`
- A couple of tests with the following data set:

```
#datatype,string,long,string,string,dateTime:RFC3339,double
#group,false,false,true,true,false,false
#default,_result,0,Wx8gaJ,U2FstZ,,
,result,table,foo,bar,_time,_value

#datatype,string,long,string,string,dateTime:RFC3339,double
#group,false,false,true,true,false,false
#default,_result,1,U37BFq3,KYxg,,
,result,table,foo,bar,_time,_value

#datatype,string,long,string,string,dateTime:RFC3339,double
#group,false,false,true,true,false,false
#default,_result,2,U37BFq3,U2FstZ,,
,result,table,foo,bar,_time,_value

#datatype,string,long,string,string,dateTime:RFC3339,double
#group,false,false,true,true,false,false
#default,_result,3,Wx8gaJ,KYxg,,
,result,table,foo,bar,_time,_value
```
Those are four empty tables. It would be good to run the test with this input and make sure it behaves right when there are empty tables, for both filter with `onEmpty: "keep"` and `onEmpty: "drop"` (this latter one is the default)

Note that there would 8 `_test.flux` files in total, with both `on_test` and `off_test` variants of these four
- normal data set filter with `onEmpty: "keep"`
- normal data set filter with `onEmpty: "drop"`
- empty tables data set filter with `onEmpty: "keep"`
- empty tables data set filter with `onEmpty: "drop"`

I just want us to be certain that it's always safe to transpose group and filter to put the filter first, and I find it hard to reason about in my head.

stdlib/universe/group.go
```diff
@@ -392,3 +393,24 @@ func (r MergeGroupRule) Rewrite(ctx context.Context, lastGroup plan.Node) (plan.
 
 	return merged, true, nil
 }
+
+// `OrderFilterGroup` orders to have 'filter |> group' from 'group |> filter'
+type OrderFilterGroup struct{}
```
wolffcm, line 398: Could you change this name to `GroupFilterTransposeRule`. We have a similar rule that transposes group and window in `influxdb`, so it would be good to follow the convention.

stdlib/universe/order_filter_group_drop_empty_flag_off_test.flux
```diff
@@ -0,0 +1,59 @@
...
+#datatype,string,long,string,string,dateTime:RFC3339,double
+#group,false,false,true,true,false,false
+#default,_result,3,Wx8gaJ,KYxg,,
+,result,table,foo,bar,_time,_value
+"
+
+order_fn = (tables=<-) => tables
+    |> range(start: 2018-05-22T19:53:26Z)
+    |> group(columns: ["_time"])
+    |> filter(fn: (r) => r["_field"] == "load4", onEmpty: "drop")
```
wolffcm, line 56: Could you update this query to be a bit more meaningful? 
- You could group by the column `foo` or `bar` so that two tables are produced
- For the filter, none of the tables have a `_field` set to `"load4"` so that doesn't really make sense. You could instead filter on `r.bar == "Wx8gaJ"` or something like that?

(Note that even if there are no rows in the table, the group key columns still have values.

wolffcm: (this also goes for the other tests)

stdlib/universe/group.go
```diff
@@ -392,3 +393,24 @@ func (r MergeGroupRule) Rewrite(ctx context.Context, lastGroup plan.Node) (plan.
 
 	return merged, true, nil
 }
+
+// `GroupFilterTransposeRule` orders to have 'filter |> group' from 'group |> filter'
+type GroupFilterTransposeRule struct{}
```
wolffcm, line 398: Can you register this rule here? This will ensure that the rule is actually applied for the end to end tests.
https://github.com/influxdata/flux/blob/80120e492229d5fe20b4a68dbf0b13161075ccf6/stdlib/flux_test.go#L21

And also here? This will let the rule run in the Flux REPL.
https://github.com/influxdata/flux/blob/96447025b11f6a19c17cbf95c97bc40562027bd9/cmd/flux/main.go#L15

Issue #3039, opened 2020-07-27
JSON serialization for FunctionType AST node


jlapacik:
DOD:
* tests exist [here](https://github.com/influxdata/flux/blob/master/libflux/src/core/ast/tests.rs) for serializing and deserializing a FunctionType

Issue #3038, opened 2020-07-27
JSON serialization for RecordType AST node


jlapacik:
DOD:
* tests exist [here](https://github.com/influxdata/flux/blob/master/libflux/src/core/ast/tests.rs) for serializing and deserializing a RecordType

Issue #3037, opened 2020-07-27
JSON serialization for ArrayType AST node


jlapacik:
DOD:
* tests exist [here](https://github.com/influxdata/flux/blob/master/libflux/src/core/ast/tests.rs) for serializing and deserializing an ArrayType

Issue #3036, opened 2020-07-27
JSON serialization for TvarType AST node


jlapacik:
DOD:
* tests exist [here](https://github.com/influxdata/flux/blob/master/libflux/src/core/ast/tests.rs) for serializing and deserializing a TvarType

PR #3106, opened 2020-08-11
feat(ast/go/json.go): Add JSON Serialization for AST Go nodes.

ChitlangeSahas:
closes: #3035, 
closes: #3034, 
closes: #3036, 
closes: #3037, 
closes: #3038, 
closes: #3039, 
closes: #3040

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -373,7 +373,7 @@ type ParameterType struct {
 	BaseNode
 	Name *Identifier   `json:"name,omitempty"`
 	Ty   MonoType      `json:"monotype"`
-	Kind ParameterKind `json:"kind"`
+	Kind ParameterKind `json:"-"`
 }
 
 func (c *ParameterType) Copy() Node {
@@ -389,8 +389,8 @@ func (c *ParameterType) Copy() Node {
 	nc.Kind = c.Kind
 	return nc
 }
-func (ParameterType) Type() string {
-	return "ParameterType"
+func (p *ParameterType) Type() string {
+	return string(p.Kind)
 }
 
 // Package represents a complete package source tree
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -232,7 +232,7 @@ func (FunctionType) monotype() {}
 
 type NamedType struct {
 	BaseNode
-	ID *Identifier `json:"id"`
+	ID *Identifier `json:"name"`
 }
 
 func (NamedType) Type() string {
@@ -252,7 +252,7 @@ func (nt *NamedType) Copy() Node {
 
 type TvarType struct {
 	BaseNode
-	ID *Identifier `json:"id"`
+	ID *Identifier `json:"name"`
 }
 
 func (c *TvarType) Copy() Node {
@@ -316,7 +316,7 @@ func (c *RecordType) Copy() Node {
 
 type PropertyType struct {
 	BaseNode
-	Name *Identifier `json:"id"`
+	Name *Identifier `json:"name"`
 	Ty   MonoType    `json:"ty"`
 }
 
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -646,8 +646,8 @@ func (s *OptionStatement) Copy() Node {
 // BuiltinStatement declares a builtin identifier and its type
 type BuiltinStatement struct {
 	BaseNode
-	ID *Identifier    `json:"id"`
-	Ty TypeExpression `json:"ty"`
+	ID *Identifier `json:"id"`
+	//Ty TypeExpression `json:"ty"`
 }
 
 // Type is the abstract type
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -272,7 +272,7 @@ func (TvarType) Type() string {
 
 type ArrayType struct {
 	BaseNode
-	ElementType MonoType `json:"elementType"`
+	ElementType MonoType `json:"element"`
 }
 
 func (ArrayType) Type() string {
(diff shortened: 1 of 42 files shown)


ast/ast.go
```diff
@@ -162,22 +172,56 @@ func (e Error) Error() string {
...
+		return c
+	}
+	nc := new(TypeConstraint)
+	*nc = *c
+	nc.BaseNode = c.BaseNode.Copy()
+
+	nc.Tvar = c.Tvar.Copy().(*Identifier)
+	//nc.Kinds = c.Kinds.Copy().([]*Identifier)
+	for k := range c.Kinds {
+		_ = append(nc.Kinds, c.Kinds[k])
```
jlapacik, line 199: Missed an append. Should be `nc.Kinds = append(nc.Kinds, c.Kinds[k])`

ast/ast.go
```diff
@@ -515,9 +646,8 @@ func (s *OptionStatement) Copy() Node {
 // BuiltinStatement declares a builtin identifier and its type
 type BuiltinStatement struct {
 	BaseNode
-	ID *Identifier `json:"id"`
-	// TODO(nathanielc): Add type expression here
-	// Type TypeExpression
+	ID *Identifier    `json:"id"`
+	Ty TypeExpression `json:"ty"`
```
jlapacik, line 650: Let's not make this change as part of this PR.

ast/json_test.go
```diff
@@ -157,6 +157,308 @@ func TestJSONMarshal(t *testing.T) {
...
+						BaseNode: ast.BaseNode{},
+						ID: &ast.Identifier{
+							BaseNode: ast.BaseNode{},
+							Name:     "int",
+						},
+					},
+				},
+			},
+			// r#"{"type":"ArrayType","element":{"type":"ArrayType","element":{"type":"NamedType","name":{"type":"Identifier","name":"A"}}}}"#
+			want: `{"type":"ArrayType","elementType":{"type":"ArrayType","elementType":{"type":"NamedType","name":{"type":"Identifier","name":"int"}}}}`,
```
jlapacik, line 198: It looks like the json tag needs to be `element` here. The rust json comment is very helpful by the way.

ast/json_test.go
```diff
@@ -157,6 +157,308 @@ func TestJSONMarshal(t *testing.T) {
...
+		},
+		{
+			name: "RecordType_NoTvar_NoProp",
+			node: &ast.RecordType{
+				BaseNode:   ast.BaseNode{},
+				Properties: []*ast.PropertyType{},
+				Tvar:       nil,
+			},
+			//{"type":"RecordType","properties":[]}"#);
+			want: `{"type":"RecordType","properties":[],"tvar":null}`,
```
jlapacik, line 235: Why does `tvar` get added to the json string?

ast/json_test.go
```diff
@@ -157,6 +157,308 @@ func TestJSONMarshal(t *testing.T) {
...
+								BaseNode: ast.BaseNode{},
+								Name:     "int",
+							},
+						},
+					},
+				},
+				Tvar: nil,
+			},
+			// {"type":"RecordType","properties":[{"name":{"name":"A"},"monotype":{"type":"NamedType","name":{"name":"int"}}}]}"#
+			want: `{"type":"RecordType","properties":[{"type":"PropertyType","name":{"type":"Identifier","name":"A"},"ty":{"type":"NamedType","name":{"type":"Identifier","name":"int"}}}],"tvar":null}`,
```
jlapacik, line 260: It looks like `ty` needs to be renamed to `monotype` here

ast/json_test.go
```diff
@@ -157,6 +157,308 @@ func TestJSONMarshal(t *testing.T) {
...
+				},
+				Return: &ast.NamedType{
+					BaseNode: ast.BaseNode{},
+					ID: &ast.Identifier{
+						BaseNode: ast.BaseNode{},
+						Name:     "uint",
+					},
+				},
+			},
+			want: `{"type":"FunctionType","parameters":[{"type":"ParameterType","name":{"type":"Identifier","name":"B"},"ty":{"type":"NamedType","name":{"type":"Identifier","name":"string"}},"kind":0}],"return":{"type":"NamedType","name":{"type":"Identifier","name":"uint"}}}`,
```
jlapacik, line 305: Same here, I think `ty` needs to be renamed to `monotype`

ast/json_test.go
```diff
@@ -157,6 +157,308 @@ func TestJSONMarshal(t *testing.T) {
...
+				},
+				Return: &ast.NamedType{
+					BaseNode: ast.BaseNode{},
+					ID: &ast.Identifier{
+						BaseNode: ast.BaseNode{},
+						Name:     "int",
+					},
+				},
+			},
+			want: `{"type":"FunctionType","parameters":[{"type":"ParameterType","name":{"type":"Identifier","name":"A"},"ty":{"type":"NamedType","name":{"type":"Identifier","name":"int"}},"kind":1}],"return":{"type":"NamedType","name":{"type":"Identifier","name":"int"}}}`,
```
jlapacik, line 335: The parameter type json doesn't quite look like the Rust version. Can you add the rust version as a comment right above this line like you did with some of the others?

ast/json_test.go
yuktarakesh: I get an error saying I need to add something for tvar there, unless there is something I need to change to make it optional?

ast/json_test.go
yuktarakesh: I pushed the code with the rust commented above, since the format for Go is the "kinds" and not the keyword "optional" the syntax is different. Should I change that?

ast/json_test.go
jlapacik: My only concern is if we can unmarshal this string to the appropriate Rust type. As long as we can do that and Rust isn't confused by `"tvar":null` that's all that matters.

Issue #2088, opened 2019-11-01
Use consistent language for Row/Records/Objects


nathanielc:
Currently the concept of a object (a value with a set of key value pairs) has several names:

* row - the name used because it represents a single row of a Flux table
* record - the name used in the literature to discuss a value of this type
* object - comes from javascript and other places

We should consolidate on a single name. My recommendation is to use record. It makes it a different name from the javascript object, which is good since it behaves differently in some ways. It is also a familiar name for an entity within a table as well as a common name when talking about data. 

We need to update the SPEC and code to all use the same name.

nathanielc:
Since creating this issue I have also received feedback that `row` is also confusing as it implies to much similarity with SQL like systems. At this point I am convinced we should use `record` for the name.

sanderson:
We will need to do this in all the user-facing documentation as well.

nathanielc:
https://github.com/influxdata/flux/pull/3111 Does the work to rename the Rust parser and type system code to use Record.

That leaves the Go code, the Flux SPEC and the user-facing docs.

nathanielc:
This PR updates the SPEC https://github.com/influxdata/flux/pull/3112

PR #3111, opened 2020-08-12
refactor: Refactor Row -> Record

ChitlangeSahas:
Fixes part of: #2088


nathanielc:
Also this is only part of the work that needs to be done in https://github.com/influxdata/flux/issues/2088 so we should not close that issue when this PR lands.

Issue #3057, opened 2020-07-30
Add test cases for using tableFind and family inside of a map call.


nathanielc:
I saw a recent use case in the wild where a user was iterating over rows in a table and using table find to perform an aggregate query in a different bucket.

The general form of the query was like this:

```
// Read data about which aggregations to compute
from(bucket: "queue")
    |> range(start: -1h)
    |> filter(...)
    // For each interval produce a row that is the new aggregation result
    |> map(fn: (r) => {
        start = time(v: r._value)
        stop = experimental.addDuration(to: start, d: 1h)
        measurement = r.measurement
        tagX = r.tagX
        agg = from(bucket: "rawdata")
            |> range(start: start, stop: stop)
            |> filter(fn: (r) => r._measurement == measurement and r.tagX == tagX)
            |> // Aggregation function
            |> findRecord(idx:0) // get the aggregation row
        return { r with
            _value: agg._value,
            _time: agg._time,
            _field: "value",
        }
    })
    // Write the table that now contains the updated aggregation results to the aggregation bucket
    |> to(bucket: "aggdata")
```

We should ensure this particular use of tableFind and friends is tested so we do not regress this behavior.

nathanielc:
Notes:

* Ensure that the resource management is the same between the parent execution context and the children execution contexts. I.e. the Allocator etc such that the query as a whole fits in normal constraints for a query

wolffcm:
@nathanielc I'm curious, have you seen a query like this actually work, like with your own eyes?

When I wrote a test case for this there were a couple of issues:
- The `compiler` does not handle pipe arguments correctly, so the call to `findRecord` complains of its required `tables` argument being missing
- The `compiler` produces a `values.Function` that does not implement the `interpreter.Resolver` interface, so when we call `filter` from inside of `map`, I get an error `function is not resolvable`

I've managed to fix both of these on my branch, so I'll open some PRs to fix them.

PR #3113, opened 2020-08-12
refactor(interpreter): relocate resolver code from interpreter to values

wolffcm:
This is a refactor that move the code that implements the `interpreter.Resolver` interface from the Go package `interpreter` to the `values` package.

Relocating this code allowed me to make the `compiler/functionValue` struct also implement the interface. Not relocating the code causes a circular dependency.

This will let use handle queries like this one:
```
csv.from(csv: inData)
  |> range(start: 2020-08-10T00:00:00Z)
  |> filter(fn: (r) => r._measurement == "events" and r._field == "times")
  |> map(fn: (r) => {
    start = time(v: r._value)
    city = r.city
    agg = csv.from(csv: inData)
      |> range(start, stop)
      |> filter(fn: (r) => r._measurement == "city_data" and r._field == "temp" and r.city == city)
      |> mean()
    return {_time: stop, _value: agg._value, _field: "event_temp_mean"}
  })
```
This query would fail because we attempt to resolve the lambda passed to `filter` here:
https://github.com/influxdata/flux/blob/b7d8af7dfbbc59844be41538d2aa0ad0255ca78e/stdlib/universe/filter.go#L67
However, this didn't work because in this example we are evaluating the lambda passed to `map` with the compiler rather than the interpreter, and compiled functions did not implement `Resolver`.  With these changes, the above query can succeed.

This fix is needed to create the test case requested by this issue: #3057


interpreter/interpreter.go
```diff
@@ -892,314 +882,7 @@ func (f function) Scope() values.Scope {
 
 // Resolve rewrites the function resolving any identifiers not listed in the function params.
 func (f function) Resolve() (semantic.Node, error) {
-	n := f.e.Copy()
```
wolffcm, line 895: All of this code has been moved to `values/resolve.go`

values/resolve.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/semantic"
+)
+
+// Resolver represents a value that can resolve itself.
+// Resolving is the action of capturing the scope at function declaration and
+// replacing any identifiers with static values from the scope where possible.
+// TODO(nathanielc): Improve implementations of scope to only preserve values
+// in the scope that are referrenced.
+type Resolver interface {
```
wolffcm, line 15: All of the code in this file was moved from `interpreter/interpreter.go` with only minimal changes so that it can compile.

PR #3114, opened 2020-08-12
refactor(interpreter): expose interpreter's resolver to compiler

wolffcm:
This is a refactor that exposes code that implements the `interpreter.Resolver` interface so that it can be accessed via the `compiler` package. This code allows the `compiler.functionValue` struct also implement the interface.

This will let use handle queries like this one:
```
csv.from(csv: inData)
  |> range(start: 2020-08-10T00:00:00Z)
  |> filter(fn: (r) => r._measurement == "events" and r._field == "times")
  |> map(fn: (r) => {
    start = time(v: r._value)
    city = r.city
    agg = csv.from(csv: inData)
      |> range(start, stop)
      |> filter(fn: (r) => r._measurement == "city_data" and r._field == "temp" and r.city == city)
      |> mean()
    return {_time: stop, _value: agg._value, _field: "event_temp_mean"}
  })
```
This query would previously fail because we attempt to resolve the lambda passed to filter here:
https://github.com/influxdata/flux/blob/b7d8af7dfbbc59844be41538d2aa0ad0255ca78e/stdlib/universe/filter.go#L67
This didn't work because in this example we are evaluating the lambda passed to `map` with the compiler rather than the interpreter, and compiled functions did not implement `Resolver`. With these changes, the above query can succeed.

This fix is needed to create the test case requested by this issue: #3057

### DIFF
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
 )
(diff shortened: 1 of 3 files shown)


compiler/runtime.go
```diff
@@ -572,6 +573,23 @@ type functionValue struct {
 	scope  Scope
 }
 
+// functionValue implements the interpreter.Resolver interface.
+var _ interpreter.Resolver = (*functionValue)(nil)
```
jlapacik, line 577: Is the only purpose of this line of code to document that `functionValue` implements `interpreter.Resolver`? If so, that's awesome.

compiler/runtime_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (m mockImporter) ImportPackageObject(_ string) (*interpreter.Package, error) {
+	panic("unimplemented")
+}
+
+// TestFunctionValue_Resolve just demonstrates that
+// functionValue implements the values.Resolver interface
+func TestFunctionValue_Resolve(t *testing.T) {
+	src1 := `x = 42 y = 100`
+	src2 := `() => x + y`
```
jlapacik, line 27: This is a nice test case. I wish we had more of these.

compiler/runtime.go
wolffcm: Yes, I've seen it elsewhere in our code. This way if somehow this struct stops implementing the interface, the error will occur close to where the fix needs to happen.

jlapacik: That is so useful. I'm definitely going to start doing that.

Issue #3043, opened 2020-07-27
Delete old type expression parser


jlapacik:
Once #3042 is complete, we can delete the [old type expression parser](https://github.com/influxdata/flux/tree/master/libflux/src/core/semantic/parser) as it will be completely unused at this point.

DOD:
* The above directory has been removed from the repo and all flux tests continue to pass

PR #3115, opened 2020-08-13
feat: delete old parser.

ChitlangeSahas:
closes: #3043 

Delete the old parser  🔫 .


nathanielc:
:tada: Deleting code is always more work that it seems and it some of the most important work. Thanks!

PR #3118, opened 2020-08-13
feat(stdlib/universe): check test input schema compatibility w/ influxdb

### LINKED ISSUE #2937: Check flux end to end tests compatibility with influxdb
We also want to check end to end tests for compatibility with InfluxDB on a schedule so that we can update those as needed. 

DOD: 
- add a check for InfluxDB compatibility to our end-to-end tests

Note: 
In the future, we would like to tackle the issues outlined in #1577 by making flux end to end tests automatically compatible with influxDB. 

We've split that issue up into the following issues: 
**the csv.from method should not require a call to range** #2935
**Add table values required by influxDB to flux end to end tests automatically** #2936

fchikwekwe:
Closes #2937


stdlib/universe/die_test.go
```diff
@@ -0,0 +1,45 @@
...
+			if got == nil {
+				t.Fatal("this function should produce an error")
+			}
+
+			if !cmp.Equal(tc.want, got) {
+				t.Fatalf("unexpected result -want/+got\n\n%s\n\n", cmp.Diff(tc.want, got))
+			}
+		})
+	}
+}
```
fchikwekwe, line 45: @rockstar added a unit test here. Lemme know what you think.

fchikwekwe:
@rockstar I agree that we should go back through these tests and update them to match these new requirements. I can bring this up on Thursday during planning.

stdlib/universe/die_test.go
rockstar: I'm not sure what the common pattern for go tests usually is here. It seems awkward to create a list of cases that is one test case long. Maybe just calling `t.Run` directly for your one case, instead of iterating over them, would make things better. Maybe this is idiomatic go, though? I'm definitely not an expert there.

stdlib/universe/die_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func TestDie(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		msg string
+		want error
+	}{
+		{
+			name: "foo",
```
rockstar, line 23: Placeholder text is sure doing a good job here. 😉

stdlib/universe/die_test.go
fchikwekwe: No, its not really necessary here. You're totally right to point that out. Remnant from when I thought I might have more cases to test.

fchikwekwe:
Tests have been fixed and are passing. I'm gonna go ahead and merge this.

PR #3124, opened 2020-08-14
test(stdlib): add a test cases for using table functions inside map

### LINKED ISSUE #3057: Add test cases for using tableFind and family inside of a map call.

wolffcm:
This just adds a couple test cases that use `findColumn` and `findRecord` inside a call to `map`.

~~This will in part help address #3057, but this test won't pass in influxdb, so I will skip it there and add a launcher test.~~
It turns out that this test case works okay in influxdb, so it is sufficient to fix #3057.

### DIFF
--- a/libflux/go/libflux/buildinfo.gen.go
+++ b/libflux/go/libflux/buildinfo.gen.go
@@ -477,6 +477,8 @@ var sourceHashes = map[string]string{
 	"stdlib/universe/string_max_test.flux":                                          "78c162ba2901856e0091f7990018b4d6e31b9fd34f7152002fcca1c2c155abf7",
 	"stdlib/universe/string_sort_test.flux":                                         "3ee25884c1b5b02dd36d153929714dedfced7176d7e2f54ebcdfb5797bd8bbde",
 	"stdlib/universe/sum_test.flux":                                                 "dbc5ae649610f6ffdbe8f9ea3f2b31fc1a2bf1e95bad579a07f39a838c8c5980",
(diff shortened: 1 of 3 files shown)


wolffcm:
@fchikwekwe No, the issue just calls for writing a test case, although I did have to make some product changes (in separate PRs with their own unit tests):
- https://github.com/influxdata/flux/pull/3114
- https://github.com/influxdata/flux/pull/3113

wolffcm:
Also I tried running this test against `influxdb` and it passes, so I will just close #3057 when this is merged.

wolffcm:
I think that Codecov is oriented around unit tests, where it expects the test to exercise code in the same package. The end-to-end test runner is in `stdlib/flux_test.go`, and the new tests don't exercise any code in the `stdlib` package that wasn't already being covered elsewhere I guess.

PR #3129, opened 2020-08-18
feat: add the profiler package

yzhang1991:
This PR introduces a new `Profiler` interface. The responsibility of a profiler is that, given a `flux.Query`, it needs to gather all the statistics within its scope and produce a resultant `flux.Table` which will be returned to the user along with the usual query result tables.

We maintain a global directory for profilers. At this point, we only have one profiler called `FluxStatistics`. `FluxStatistics` returns all the contents in the `flux.Statistics` for a `flux.Query`.

This PR also introduces a `profiler` stdlib package that is similar to the `planner` package.
Users can use this package to control which profiler they wish to turn on.
For example:
```
import "profiler"
option profiler.enabledProfilers = ["FluxStatistics"]

from(bucket: v.bucket)
|> range(start: 2018-05-22T00:00:00Z, stop: 2018-05-22T00:01:00Z)
|> filter(fn: (r) => r._measurement == v.ethan or r._measurement == "cpu")
|> filter(fn: (r) => r.host == "host.local")
|> aggregateWindow(every: 30s, fn: count)
```


execute/profiler.go
```diff
@@ -0,0 +1,64 @@
...
+				Label: "_measurement",
+				Type:  flux.TString,
+			},
+		},
+		[]values.Value{
+			values.NewString("profiler/FluxStatistics"),
+		},
+	)
+	b := NewColListTableBuilder(groupKey, alloc)
+	for _, colName := range []string{"_measurement", "_field", DefaultValueColLabel} {
```
nathanielc, line 45: Do we want to use the `_field/_value` schema or the pivoted schema where each statistic is its own column.

Is see below that all stats are considered strings. I think that makes working with the stats data too cumbersome.

If we used a wide schema with each stat being its own column that would allow for each column to have its own type.

Thoughts?

execute/profiler.go
```diff
@@ -0,0 +1,64 @@
...
+			Label: colName,
+			Type:  flux.TString,
+		}); err != nil {
+			return nil, err
+		}
+	}
+	q.Statistics().Range(func(key string, value string) {
+		b.AppendString(0, "profiler/FluxStatistics")
+		b.AppendString(1, key)
+		b.AppendString(2, value)
```
nathanielc, line 56: Why are all the statistics considered strings? I think most of them are integers while some may be other things.

What if instead of making everything a string we added the `_field` to the group key such that the value

stdlib/profiler/profiler.flux
```diff
@@ -0,0 +1,3 @@
+package profiler
+
+option enabledProfilers = [""]
```
nathanielc, line 3: Why does this need to be a list of empty string? is it so that the type system knows is a list of strings?

adrian-thurston (commented):
Could you give a description of these changes? It's a bit hard to follow what is going from the patch alone.

stdlib/profiler/profiler.flux
yzhang1991: Yes

lang/compiler.go
```diff
@@ -478,41 +514,25 @@ func getPlanOptions(plannerPkg values.Package) (plan.LogicalOption, plan.Physica
...
-func getRules(plannerPkg values.Object, optionName string) ([]string, error) {
-	value, ok := plannerPkg.Get(optionName)
+func getOptionValues(pkg values.Object, optionName string) ([]string, error) {
+	value, ok := pkg.Get(optionName)
 	if !ok {
 		// No value in package.
 		return []string{}, nil
 	}
 
-	// TODO(affo): the rules are arrays of strings as defined in the 'planner' package.
```
yzhang1991, line 499: Our type system can now substitute the removed parts

Issue #2794, opened 2020-04-29
Delete Go Parser


nathanielc:
Now that we have the Rust parser and Formatter we want to delete the Go versions so as to not expose older broken functionality. 

The uses of the Go Parser are already all using the Rust version of the parser. ~The Formatter may have a few uses elsewhere.~

Note we are going to remove the Formatter as part of a different issue.

wolffcm:
There are some functions for parsing durations that need to still exist after deleting the Go parser.

wolffcm:
If there are uses of the Go formatter, look into using the Rust parser instead. If it's difficult file an other issue to do it.

jsternberg:
It's fine for us to modify the formatter to use the Rust formatter, but we need to move the function invocation from the AST package. The `ast` package cannot import the libflux package as it cannot have a C dependency.

So we're going to have to create a new package probably like this:

1. Create a new `github.com/influxdata/flux/ast/format` and `github.com/influxdata/flux/ast/internal/format` package.
2. Copy the current format implementation to the internal package.
3. Make it so that both of the above packages reference the same internal implementation. Mark the current function as deprecated.
4. Perform a release.
5. Modify downstream dependencies to use the new package.
6. Remove the `ast.Format()` method from the `ast` package and the internal package. Move the implementation to the new location and change it to use Rust if it isn't already.

This may seem complex and it may seem easier to just break it and fix the downstream dependencies during the release, but I'd like us to get into the habit of doing this kind of thing when making breaking API changes. Releases shouldn't require refactors to make them work.

fchikwekwe:
Places where the Go parser is still referenced: 

- `ast/flatbuffers_test.go TestRoundTrip`
- `libflux/go/libflux/bench_test.go GoParse`
- `libflux/go/libflux/parser_test.go TestASTPkg_MarshalJSON`

And then of course the parser references itself in some places.

wolffcm:
As far as `ast/flatbuffers_test.go TestRoundTrip` goes, the Go parser is used to create the expected value for an AST that is created like this 
```
parse in Rust -> seralized in Rust to FB -> deserialize from FB in Go
```
This test was intended to show that the AST we get from Rust including flatbuffers serde is nearly same as what we would get from the old Go parser, with only some trivial differences. However we've been using this codepath in production for several months now, so I think we're safe on this front.

We could delete this test and not lose any coverage. This is because we have tests in Rust the demonstrate that the AST was serialized as expected.

[...]

PR #3142, opened 2020-08-24
refactor(parser): Delete go parser and all of its references

scbrickley:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

Helps with #2794


scbrickley:
@nathanielc remind me what needs to happen in the downstream repos? That's where "the dance" comes in right?

nathanielc:
@scbrickley Yes, we want to confirm that both influxdb and idpe repo continue to work with this change. Since this is a big change we want to make sure we confirm that before merging this change. 

The _dance_ is basically checkout all three repos locally and use `go mod replace` directives to point the flux  dependency in those other repos at your local copy of Flux with this change. Then to run tests to ensure its all working.

scbrickley:
Opened PRs in [`idpe`](https://github.com/influxdata/idpe/pull/8420) and [`influxdb`](https://github.com/influxdata/influxdb/pull/19462) to check for compatibility. Looks like everything is green.

PR #3143, opened 2020-08-24
feat(flux): make experimental tracing an attribute of the context

wolffcm:
This turns experimental tracing from a global attribute to one that is set in the context. This will allow `influxdb` (OSS) to turn on experimental tracing in the context (or not) by looking at a feature flag.

This implements part of https://github.com/influxdata/influxdb/issues/16220.

### Done checklist
- [x] Test cases written


wolffcm:
`codecov/patch` is complaining, because the unit test I added is in a different package from the product code changes. However, my test does actually test the new behavior.

PR #3147, opened 2020-08-25
feat: time weighted average

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/universe/universe.flux
```diff
@@ -98,6 +98,11 @@ builtin logarithmicBins
 builtin sleep // sleep is the identity function with the side effect of delaying execution by a specified duration
 builtin die // die returns a fatal error from within a flux script
 
+// Time weighted average where values at the beginning and end of the range are linearly interpolated.
+timeWeightedAvg = (tables=<-, unit) => tables
+    |> integral(unit: unit, interpolate: "linear")
+    |> map(fn: (r) => ({ r with _value: (r._value * float(v: uint(v: unit))) / float(v: int(v: r._stop) - int(v: r._start)) }))
```
rockstar, line 104: When I opened this patch, I expected there to be a chunk of go code. _Very_ happy to see it implemented in pure flux.

stdlib/universe/universe.flux
```diff
@@ -98,6 +98,11 @@ builtin logarithmicBins
 builtin sleep // sleep is the identity function with the side effect of delaying execution by a specified duration
 builtin die // die returns a fatal error from within a flux script
 
+// Time weighted average where values at the beginning and end of the range are linearly interpolated.
```
rockstar, line 101: Was there an explicit decision to put this in universe? I'm starting to be worried that we might be polluting the namespace a lot now.

stdlib/universe/universe.flux
jlapacik: Go code was needed for `integral` https://github.com/influxdata/flux/pull/3146.

stdlib/universe/universe.flux
jlapacik: No explicit decision, I just wasn't sure what package this function should belong to. I agree about polluting the universe package though. I'm open to any suggestions.

Issue #2253, opened 2019-12-06
Remove column parameters from Flux function "range"


wolffcm:
The `range` function has a few seldom-used parameters:
- `timeColumn` defaults to `_time`
- `startColumn` defaults to `_start`
- `stopColumn` defaults to `_stop`

These parameters make it hard to map the input row type to the output row type and break type inference, so the type is
```
                    forall [t0, t1, t2, t3] where t0: Row, t3: Row (
                        <-tables: [t0],
                        start: t1,
                        ?stop: t2,
                        ?timeColumn: string,
                        ?startColumn: string,
                        ?stopColumn: string
                    ) -> [t3]
```

We should remove these parameters so the type can be
```
                    forall [t0, t1, t2] where t0: Row (
                        <-tables: [t0],
                        start: t1,
                        ?stop: t2,
                    ) -> [{_start: time | _stop: time | t0}]
```
Question: what should the behavior be if `_start` and `_stop` already exist?

jlapacik:
Note that record labels are scoped so the input records to `range` may have `_start` and `_stop` columns. In that case the input type might look something like this:

    {_start: time | _stop: time | t0}

And the output type would look like this:

    {_start: time | _stop: time | _start: time | _stop: time | t0}

nathanielc:
I don't think we should remove the arguments. rather we should use the `label` type for them. For now we can treat them as strings. The pattern of specifying the column names as arguments is very common and not unique to `range`. We will need to solve for that pattern. See https://github.com/influxdata/flux/issues/1122

Alternatively we could remove all of theses column arguments and require that the `rename` operation is used for data that doesn't conform to the expected record type. But require is going to need to understand something similar to label types as well. Thoughts?

jlapacik:
We don't have an issue proposing an implementation yet, but any implementation of label types is going to require some form of AST preprocessor. In my opinion adding functionality to an AST preprocessor should be seen as a high barrier to entry as it is a deliberate attempt to get around the type system.

I think `keep` and `drop` are good candidates for label types as they reduce the syntax needed to express the same operation via a call to `map`.

`range` and most other functions are not good candidates in my opinion. Flux is a verbose language by design, however we should try to reduce verbosity where it is not required. I'm all for replacing these column arguments with default labels as part of the type signature. Not only does it allow type inference to correctly infer the types of these functions without the help of an AST preprocessor, but most importantly it simplifies the API for users.

nathanielc:
[...]

PR #3151, opened 2020-08-27
fix: remove time-column args from range function and update type sig

adrian-thurston:
BREAKING CHANGE: Remove timeColumn, startColumn and stopColumn arguments from
the range function and always use the defaults and update the type signature to
return a type based on the input. This eliminates return-type polymorphism from
the range function.

https://github.com/influxdata/flux/issues/2253

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/src/core/semantic/builtins.rs
+++ b/libflux/src/core/semantic/builtins.rs
@@ -629,14 +629,12 @@ pub fn builtins() -> Builtins<'static> {
                 //   https://github.com/influxdata/flux/issues/2243
                 // Also, we should remove the column arguments so we can reuse A in the return type:
                 //   https://github.com/influxdata/flux/issues/2253
+
                 "range" => r#"(
-                        <-tables: [A],
+                        <-tables: [{A with _time: time}],
                         start: B,
-                        ?stop: C,
-                        ?timeColumn: string,
-                        ?startColumn: string,
-                        ?stopColumn: string
-                    ) => [D] where A: Record, D: Record "#,
+                        ?stop: C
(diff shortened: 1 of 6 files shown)


adrian-thurston:
Current status is that I found no occurrence of `timeColumn`, `startColumn` and `stopColumn` args to `range` in queries going back 4 days. This covers 99.993% of queries as some of them fail to parse.

libflux/src/core/semantic/builtins.rs
```diff
@@ -628,11 +628,8 @@ pub fn builtins() -> Builtins<'static> {
                 "range" => r#"(
                         <-tables: [A],
                         start: B,
-                        ?stop: C,
-                        ?timeColumn: string,
-                        ?startColumn: string,
-                        ?stopColumn: string
-                    ) => [D] where A: Record, D: Record "#,
+                        ?stop: C
+                    ) => [A] where A: Record"#,
```
jlapacik, line 632: I think `_start`, `_stop`, and `_time` need to be encoded directly within the type signature

    (<-tables: [{ T with _time: time }], start: A, ?stop: B) => [{ T with _time: time, _start: time, _stop: time }]

adrian-thurston: Ah, that's much tighter. I'll remake the patch tomorrow using the `with _time...` components.

adrian-thurston:
Current status is that these args do not appear in queries going back 12 days from today ( Aug 22, 2020 )

Issue #3155, opened 2020-08-27
Use testing.loadStorage on tests affected by new schema checks


fchikwekwe:
The following tests need to be fixed and need to use testing.loadStorage in flux. 

Right now these tests overload `testing.loadStorage`. This should be removed. 

These tests seem to be affected by the schema check added to `testing.loadStorage`. The timestamp is not being preserved as the right type and causes an error. (e.g. `stop column is not of time type`)
```./stdlib/testing/promql/extrapolatedRate_counter_rate_test.flux
./stdlib/testing/promql/linearRegression_predict_test.flux
./stdlib/testing/promql/extrapolatedRate_norate_test.flux
./stdlib/testing/promql/linearRegression_nopredict_test.flux
./stdlib/testing/promql/extrapolatedRate_nocounter_test.flux```

These tests aren't compatible with the schema check or the call to range added to `testing.loadStorage`. They just produce an incorrect result. 
```./stdlib/universe/group_nulls_test.flux
./stdlib/universe/fill_time_test.flux```

DOD: 
- tests are using the version of  testing.loadStorage located in the testing package
- tests are passing

**Comment by wolffcm** [comment]
I skipped these in OSS also, recently:
https://github.com/influxdata/influxdb/commit/e7cbbaa722303f3e806656233a1c5cbe9b47f320#diff-a446d588141883b1d32802439bf45f49R160-R164

If I'm understanding right, tests that run okay in Flux but have issues in OSS can't have their `loadStorage` options overridden since they will fail with this error in OSS:
```
     --- FAIL: TestFluxEndToEnd/testing/promql/extrapolatedRate_counter_rate (0.02s)
            end_to_end_test.go:226: @8:1-13:2: option "testing.loadStorage" reassigned
```
Do I have that right?

fchikwekwe:
Yes, that's right @wolffcm

PR #3156, opened 2020-08-27
test(stdlib): fix unbounded tests

### LINKED ISSUE #2935: the csv.from method should not require a call to range
The `csv.from()` call does not require a `range()`. It likely should continue to not require a call to `range()`. But, the influxdb source _does_ require this call, causing any end to end test that doesn't have that call to fail. This causes us to have to add the call to `range()` to any end to end tests. 

We have another issue for checking if flux end to end tests will fail in influxdb. However, it would be nice for this call to not be required in the future. 

DOD:
- make it so these tests can run in influxdb without adding the range() call

fchikwekwe:
Closes #2935. 

This PR has four components: 
1. Added a call to range in `./stdlib/testing/testing.flux` to prevent unbounded end-to-end tests.  
2. Pointed as many end-to-end tests at `testing.loadStorage` as possible and documented ones that still cannot using the new schema and range checks (https://github.com/influxdata/flux/issues/3155). 
3. Unskipped some skipped end-to-end tests.
2. Cleaned up some test files.


stdlib/universe/columns_test.flux
```diff
@@ -30,20 +28,21 @@ outData = "
 #group,false,false,true,true,true,true,false
 #default,_result,,,,,,
 ,result,table,_start,_stop,host,name,_value
+,,0,2018-05-20T19:53:26Z,2030-01-01T00:00:00Z,host.local,disk0,_field
```
wolffcm, line 31: Curious why the row order changed here. Is the test not deterministic? Maybe it could use a call to `sort()` to make it so.

fchikwekwe: That's a good note. Thanks for pointing it out. I added a sort so that this test is more consistent.

PR #3167, opened 2020-09-01
feat(execute): support duration for agg window

### LINKED ISSUE #3025: Make window logic reusable
This issue is to review the flux window logic and make sure that it can be reused in the window aggregate read cursors of [influxdb](https://github.com/influxdata/influxdb/tree/master/storage/reads).

It looks like most of the functions for calculating windows live in the [execute package](https://github.com/influxdata/flux/blob/master/execute/window.go). We need to ensure that there are functions for calculating the window start time as well as the window stop time given a timestamp, an every period, and an offset. Once this has been verified/updated, the influxdb read cursors will be able to use these functions to calculate window boundaries, with support for monthly durations coming for free.

fchikwekwe:
Closes #3025. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/window.go
+++ b/execute/window.go
@@ -8,6 +8,10 @@ import (
 	"github.com/influxdata/flux/values"
 )
 
+const (
+	monthInNSecs = 2592000000000000
+)
+
 type Window struct {
 	Every  Duration
 	Period Duration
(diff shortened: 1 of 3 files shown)


execute/window.go
```diff
@@ -127,3 +131,74 @@ func truncateByMonths(t Time, d Duration) Time {
...
+	// subtract the total months from the timestamp
+	nano := int64(timestamp) - (months * monthInNSecs)
+
+	// and calculate nanoseconds for window start
+	modNano := modulo(nano, every.Nanoseconds())
+	offsetNano := modulo(offset.Nanoseconds(), every.Nanoseconds())
+	nsecs := calculateStart(nano, every.Nanoseconds(), modNano, offsetNano)
+
+	// add the start months back to the start nanoseconds
+	return values.Time(nsecs + (months * monthInNSecs))
```
fchikwekwe, line 154: There may be an easier way to get the right types here. Definitely open to changing this if so.

fchikwekwe:
@jlapacik to use which functions?

jlapacik:
`WindowStart` and `WindowStop`. These functions encapsulate the windowing logic that both the flux window function and the influxdb storage operation should use.

PR #3168, opened 2020-09-01
feat(contrib): aggregate.window for an alternative windowing aggregate

jsternberg:
The `contrib/jsternberg/aggregate` package has been extended with an
`aggregate.window`. This takes the same options as `aggregate.table` and
uses the same underlying code but applies it to multiple windows. The
reduce/compute model is still used but applied to window slices instead.

This window function assumes the time is sorted and will fail otherwise.

It does not properly implement creating empty windows and filling them
with values.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/contrib/jsternberg/aggregate/window_test.flux
```diff
@@ -0,0 +1,52 @@
...
+outData = "
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,string,string,string,string,string,string,dateTime:RFC3339,dateTime:RFC3339,double,double,double,double,long
+#group,false,false,true,true,true,true,true,true,true,true,false,false,false,false,false,false,false
+#default,_result,,,,,,,,,,,,,,,,
+,result,table,_start,_stop,_field,_measurement,device,fstype,host,path,start,stop,sum,mean,min,max,count
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,2018-05-22T00:00:00Z,2018-05-22T00:00:30Z,90,30,30,30,3
+,,0,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/,2018-05-22T00:00:30Z,2018-05-22T00:01:00Z,120,40,40,40,3
+,,1,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/tmp,2018-05-22T00:00:00Z,2018-05-22T00:00:30Z,105,35,35,35,3
+,,1,2018-05-22T00:00:00Z,2018-05-22T00:01:00Z,used_percent,disk,disk1s1,apfs,host.local,/tmp,2018-05-22T00:00:30Z,2018-05-22T00:01:00Z,135,45,45,45,3
+"
```
fchikwekwe, line 36: This type of output is such an interesting idea. I think it would work well with the UI for charting and graphing trends.

stdlib/contrib/jsternberg/aggregate/aggregate.flux
```diff
@@ -39,6 +39,32 @@ import "contrib/jsternberg/math"
...
+// Along with the above required option, there are a few additional
+// optional parameters.
+//     time = string
+//         The column name for the time input.
+//         This defaults to _time or time, whichever is earlier in
+//         the list of columns.
+//     period = duration
+//         The length of the interval. This defaults to the
+//         every duration.
+builtin window : (<-tables: [A], ?time: string, every: duration, ?period: duration, columns: C) => [B] where A: Record, B: Record, C: Record
```
fchikwekwe, line 66: I guess if the time column must be in order, then a call to aggregate.window should be preceded by a call to sort.

stdlib/contrib/jsternberg/aggregate/aggregate.flux
```diff
@@ -48,7 +74,7 @@ builtin null : A
 builtin none : A
 
 // define will define an aggregate function.
-define = (init, reduce, compute, fill=null) => (column, fill=fill) => ({
+define = (init, reduce, compute, fill=null) => (column="_value", fill=fill) => ({
```
fchikwekwe, line 77: What does the fill parameter do here?

fchikwekwe (approved):
I can definitely envision some cool use cases for these functions. Good addition!

stdlib/contrib/jsternberg/aggregate/aggregate.flux
jsternberg: It defines which value should fill that column if it is empty. It doesn't work yet with `window()`.

stdlib/contrib/jsternberg/aggregate/aggregate.flux
jsternberg: Yea but it's left up to the user to deal with this. Since the planner doesn't have a way of removing a sort from already sorted data, I didn't want to add it.

I've been thinking about this though and I think I might try to change this to be more relaxed about the time ordering since it's really inconvenient to use this with `group()`.

fchikwekwe: Yeah, I think as long as it is documented clearly, then users can be expected to add their own sort. If a lot of folks run into issues around this, then maybe there can be a future update or something.

PR #3176, opened 2020-09-03
feat: operator profiler

yzhang1991:
This PR:
* Adds an `ExecutionNode` struct to all data sources and transformations to attach a label to an execution node. This label is later used in the operator profiler to identify each node and relate them to the nodes in the plan spec.
* Creates an `operator` profiler. A Goroutine will be started before the execution starts with a channel to accept profiling results from each transformation and data source that are executed, and will generate a result table before the query results are being encoded and returned.
* A new wrapper `opentracing.Span` implementation `OperatorProfilingSpan` and its accompanying utilities so that the entire operator profiling framework perfectly fits into the existing OpenTracing work.


adrian-thurston (approved):
Looks good. Only thing I would change is to not use "_start" and "_stop" in profiler results to indicate the beginning and end of each sample. Usually these columns are part of the group key and indicate the overall range of the data, or the window boundaries, etc.

yzhang1991:
@adrian-thurston They are updated to "Begin" and "End"
They are most likely going to be changed in https://github.com/influxdata/flux/issues/3203

PR #3177, opened 2020-09-03
feat(execute): test support duration for agg window

### LINKED ISSUE #3025: Make window logic reusable

fchikwekwe:
Closes #3025. 

Adding testing to ensure that windows using monthly durations set bounds properly. 

### Done checklist
- [x] Test cases written

### DIFF
--- a/execute/window.go
+++ b/execute/window.go
@@ -18,10 +18,11 @@ type Window struct {
 // and normalizes the offset to a small positive duration.
 // It also validates that the durations are valid when
 // used within a window.
-func NewWindow(every, period, offset Duration) (Window, error) {
-	// Normalize the offset to a small positive duration
-	offset = offset.Normalize(every)
-
+func NewWindow(every, period, offset Duration, months bool) (Window, error) {
+	if !months {
+		// Normalize nanosecond offsets to a small positive duration
+		offset = offset.Normalize(every)
+	}
 	w := Window{
 		Every:  every,
 		Period: period,
--- a/execute/window_test.go
+++ b/execute/window_test.go
@@ -14,11 +14,11 @@ import (
 func TestNewWindow(t *testing.T) {
 	t.Run("normal offset", func(t *testing.T) {
 		want := execute.Window{
-			Every:  values.ConvertDuration(time.Minute),
-			Period: values.ConvertDuration(time.Minute),
-			Offset: values.ConvertDuration(time.Second),
+			Every:  values.ConvertDurationNsecs(time.Minute),
+			Period: values.ConvertDurationNsecs(time.Minute),
+			Offset: values.ConvertDurationNsecs(time.Second),
 		}
-		got := MustWindow(values.ConvertDuration(time.Minute), values.ConvertDuration(time.Minute), values.ConvertDuration(time.Second))
+		got := MustWindow(values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Second), false)
 		if !cmp.Equal(want, got) {
 			t.Errorf("window different; -want/+got:\n%v\n", cmp.Diff(want, got))
 		}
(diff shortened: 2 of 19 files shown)


execute/window_test.go
```diff
@@ -105,21 +105,45 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
...
+			name: "simple months",
+			w: MustWindow(
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(0)),
+			t: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
+			want: execute.Bounds{
+				Start: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
+				Stop:  values.ConvertTime(time.Date(1970, time.June, 1, 0, 0, 0, 0, time.UTC)),
+			},
```
jlapacik, line 127: This is a great test case. What bounds does it output for a time that is before `Jan 1, 1970`? Can you add that test case too?

execute/window_test.go
```diff
@@ -105,21 +105,45 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
...
+			name: "simple months with offset",
+			w: MustWindow(
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5)),
+			t: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
+			want: execute.Bounds{
+				Start: values.ConvertTime(time.Date(1970, time.January, 2, 4, 30, 0, 0, time.UTC)),
+				Stop:  values.ConvertTime(time.Date(1970, time.June, 2, 4, 30, 0, 0, time.UTC)),
+			},
```
jlapacik, line 139: Can you explain why it outputs these bounds? Maybe I don't understand how monthly offsets work, but I would have expected a window with period equal to offset to be the same as a window with that same period and no offset.

fchikwekwe: Whoa, sorry I thought I fixed that. Should be fixed now.

execute/window_test.go
fchikwekwe: Yeah, I updated the offset test to fix this.

execute/window_test.go
```diff
@@ -385,8 +409,8 @@ func TestWindow_GetOverlappingBounds(t *testing.T) {
 	}
 }
 
-func MustWindow(every, period, offset execute.Duration) execute.Window {
-	w, err := execute.NewWindow(every, period, offset)
+func MustWindow(every, period, offset execute.Duration, months bool) execute.Window {
+	w, err := execute.NewWindow(every, period, offset, months)
```
fchikwekwe, line 413: @jlapacik the fix for that bug you saw includes awareness of nanoseconds vs months when creating the window.

execute/window.go
```diff
@@ -18,10 +18,11 @@ type Window struct {
...
 // used within a window.
-func NewWindow(every, period, offset Duration) (Window, error) {
-	// Normalize the offset to a small positive duration
-	offset = offset.Normalize(every)
-
+func NewWindow(every, period, offset Duration, months bool) (Window, error) {
+	if !months {
+		// Normalize nanosecond offsets to a small positive duration
+		offset = offset.Normalize(every)
+	}
```
fchikwekwe, line 25: @jlapacik we need to know if we're working with nanoseconds or not so that we cna know whether to scale durations here or not.

jlapacik: I see so it wasn't a bug in `GetEarliestBounds`? Just in how we were creating the window in the first place? Is it possible to add that test case back in?

fchikwekwe: Sure, I can add another case that's similar to the initial one.

execute/window_test.go
```diff
@@ -105,21 +105,57 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
...
+			name: "months with equivalent offset",
+			w: MustWindow(
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5), true),
+			t: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
+			want: execute.Bounds{
+				Start: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
+				Stop:  values.ConvertTime(time.Date(1970, time.June, 1, 0, 0, 0, 0, time.UTC)),
+			},
```
fchikwekwe, line 151: @jlapacik here's that original test case.

jlapacik (approved):

PR #3184, opened 2020-09-09
fix(stdlib): remove months param

fchikwekwe:
The `NewWindow()` function includes a `months` param which is a bool that tells
whether months are included in a duration or not. This param is confusing and
unnecessary since we already have accessor functions for months and nsecs.

We also need to remove the normalize function and update related code to properly
calculate the offset.

This PR removes the `months` parameter and skips tests that fail because of normalize.

### DIFF
--- a/execute/window_test.go
+++ b/execute/window_test.go
@@ -62,8 +62,7 @@ func TestNewWindow(t *testing.T) {
 		_, gotErr := execute.NewWindow(
 			mustParseDuration("1mo2w"),
 			mustParseDuration("1mo2w"),
-			values.Duration{},
-		false)
+			values.Duration{})
 		if want, got := errAsString(wantErr), errAsString(gotErr); want != got {
 			t.Errorf("window error different; -want/+got:\n%v\n", cmp.Diff(want, got))
 		}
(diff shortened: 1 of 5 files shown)


execute/window_test.go
```diff
@@ -126,30 +124,30 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
...
-				values.ConvertDurationMonths(5),
-				values.ConvertDurationMonths(5),
-				values.ConvertDurationMonths(5), true),
-			t: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
-			want: execute.Bounds{
-				Start: values.ConvertTime(time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC)),
-				Stop:  values.ConvertTime(time.Date(1970, time.June, 1, 0, 0, 0, 0, time.UTC)),
-			},
-		},
+		//{
```
fchikwekwe, line 127: These tests will be unskipped when we fix/remove normalize.

jsternberg (approved):

PR #3187, opened 2020-09-10
fix(scanner): Reset pointer after scanning invalid unicode

### LINKED ISSUE #3160: too many European quotes cause Flux to panic instead of error
a single set of European quotes causes a parse error (as expected), but if there are some number more, it causes a panic.

scbrickley:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

Closes #3160

The C scanner was leaving a pointer in odd places after scanning multi-byte characters not defined in [unicode.rl](https://github.com/influxdata/flux/blob/master/libflux/src/core/scanner/unicode.rl). This PR includes a change that resets the pointer to the beginning of whatever token we just failed to scan whenever the C scanner reports an error.


scbrickley:
Note that the CI test that's failing is doing so because it's regenerating `libflux/scanner.c`, thus overwriting the change necessary to get unit tests to pass. We need to patch Ragel so that this doesn't happen.

libflux/src/core/parser/tests.rs
```diff
@@ -48,6 +48,534 @@ impl<'a> Locator<'a> {
...
+                }))
+            }))],
+            eof: None,
+        },
+    )
+}
+
+#[test]
+fn parse_greedy_quotes_bare() {
+    let mut p = Parser::new(r#"“some string”"#);
```
nathanielc, line 375: This one brings up an interesting case. Should we allow any valid utf8 sequence of characters in a string literal?

According to the SPEC we state that strings can have any character https://github.com/influxdata/flux/blob/master/docs/SPEC.md#string-literals 

Thoughts? Should we create a new issue to fix this?

libflux/src/core/scanner/mod.rs
```diff
@@ -174,6 +201,10 @@ impl Scanner {
             };
             match nc {
                 Some(nc) => {
+                    // It's possible that the C scanner left the data pointer in the middle
+                    // of a character. `unread_skip_whitespace()` resets the pointer to the
+                    // beginning of the character we just failed to scan.
+                    self.unread_skip_whitespace();
```
nathanielc, line 207: Can you explain why whitespace matters here? Why isn't it safe to just unread the last character independent of its class?

nathanielc (commented):
The test are amazing! Thanks for adding so many thorough tests it really helps review. 
I have a few questions below.

libflux/src/core/scanner/mod.rs
scbrickley: Good question! Originally, I just inserted a call to `unread()` and thought that was the end of it, but then [this test](https://github.com/influxdata/flux/blob/0819086f5ae0a1b13f7c3f6692de84d5b4913f16/libflux/src/core/scanner/tests.rs#L1582) was failing. `unread()` resets the scanner to the end of the last token, but if there's a whitespace character between the end of the last token and the start of the new token, a call to `unread()` will leave the pointer in the wrong location, and we will get an error when we try to [add the size of token](https://github.com/influxdata/flux/blob/0819086f5ae0a1b13f7c3f6692de84d5b4913f16/libflux/src/core/scanner/mod.rs#L179) to the pointer.

libflux/src/core/parser/tests.rs
scbrickley: I think it would be tricky to do that without changing the Ragel file. I think this test string is slightly different than what you describe though. In this case, the invalid quotes are wrapping the user-defined string, but are not a part of it. Unless I'm misunderstanding?

I'm not opposed to having any character defined within a string though. Could be useful.

scbrickley:
Not sure exactly what to do about that code coverage report. Do I have to include a test that calls `unread_skip_whitspace()` directly?

libflux/src/core/parser/tests.rs
nathanielc: Lets create an issue around this and then we can groom and see what it would take to do.

libflux/src/core/scanner/mod.rs
nathanielc: My not directly set `self.p` to `token_start` here instead of calling `unread()`?

nathanielc:
@scbrickley Code coverage does not check the Rust code so you can ignore it for this PR

libflux/src/core/scanner/mod.rs
scbrickley: That's... a good point. Hang on. Simpler fix on the way.

libflux/src/core/parser/tests.rs
scbrickley: Issue created: https://github.com/influxdata/flux/issues/3206

nathanielc (approved):
Wonderful! Thanks again for the great tests coverage here.

PR #3189, opened 2020-09-11
fix(contrib): check for an invalid type in rows.map and return an error

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


rockstar (commented):
Can we add a test here, for this case?

PR #3197, opened 2020-09-16
fix: fixed exponent order of precedence to be above multiplicative operators

Bolladeen:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


rogpeppe:
Could this break existing customers' query code? I wonder if it might be worth providing some way for people to vet their existing expressions for whether they've changed meaning (or even generate an error for some transition period - people could work around that by adding explicit brackets)

jsternberg (approved):

Bolladeen:
@rogpeppe, I spoke with Nathaniel and we came to the conclusion that the current implementation is incorrect and that very few if any customers will use this broken behavior in their code. I don’t see the need in such a case for us to  provide them with ways to vet their code. Our current documentation also explicitly states that exponents have higher precedence, so it seems like this is a bug fix and not a change in operator policy.

nathanielc (approved):
This is great! Love the tests

nathanielc:
@rogpeppe Thanks for the concern, we have informed the docs team and the release notes will have a call out to this change for any users that may have been using this in error before.

PR #3208, opened 2020-09-22
fix(compiler): fix compiler type inference with extended records

### LINKED ISSUE #3153: Error "float != string" using filter after experimental.join
This strange error came up from a user's query:
```
Error: type conflict: float != string
```

This is the test case that reproduces the issue against vanilla Flux:
```
import "csv"
import "experimental"

f0csv = "
#group,false,false,false,false,false,false,false,false
#datatype,string,long,string,string,string,dateTime:RFC3339,boolean,double
#default,f0,,,,,,,
,result,table,_field,_measurement,th,_time,tr,_value
,,0,f0,alert,50,2020-08-27T21:40:51.608865237Z,true,115.5
"

f1csv = "
#group,false,false,false,false,false,false,false,false
#datatype,string,long,string,string,string,dateTime:RFC3339,boolean,double
#default,f1,,,,,,,
,result,table,_field,_measurement,th,_time,tr,_value
,,0,f1,alert,50,2020-08-27T21:40:51.608865237Z,true,198
"

f2csv = "
#group,false,false,false,false,false,false,false,false
#datatype,string,long,string,string,string,dateTime:RFC3339,boolean,double
#default,f2,,,,,,,
,result,table,_field,_measurement,th,_time,tr,_value
,,0,f2,alert,50,2020-08-27T21:40:51.608865237Z,true,302
"

f0 = csv.from(csv: f0csv)
f1 = csv.from(csv: f1csv)
f2 = csv.from(csv: f2csv)

join_all1 = experimental.join(left: f0, right: f1,fn: (left, right) => ({left with  t0: left.tr,  v0: left._value,  t1: right.tr,  v1: right._value,}))
join_all2 = experimental.join(left: join_all1, right: f2,fn: (left, right) => ({left with  t2: right.tr,  v2: right._value,}))
join_all = join_all2
cst = join_all
  |> map(fn: (r) => ({r with
    _time: now(),
    _value: if (r.t0 or r.t1 or r.t2) then "c" else "o",
    tr: (r.t0 or r.t1 or r.t2),
    _alert: "notification-test",
    _field: "status",
    _measurement: "statuses",
}))

// cst |> yield(name: "cst")

cst
  |> group(columns: ["_measurement"])
  |> sort(columns: ["_time"])
  |> last()
  |> filter(fn: (r) => (r._value == "crit"))   // uncomment this and query succeeds
```
Note that uncommenting the last line allows the query to succeed. However, the input to `filter` seems to have only one row, and its `_value` column has type `string`, so the error that is produced is perplexing.

wolffcm:
Fixes #3153.

This addresses an issue revealed by a test case like this one:
```
import "experimental/array"

array.from(rows: [{v: 0.1}])
  |> map(fn: (r) => ({r with v: "foo"}))
  |> filter(fn: (r) => r.v != "foo")
```
The output of `array.from` is `{v: float}`, and the output of map is `{v: string | v: float}`. This is strange but it's how records work in our type system. The first item `v: string` is the real type of `v` and the second item is just hanging around and is not generally useful.

The `filter` function invokes the "compiler" which uses the inferred type of the filter's `fn` argument to look for type errors. It ensures that column in the table have the expected type. However in this case `v` is expected to have both `string` and `float` types, which is impossible, and an error `float != string` would occur.

The fix is to ignore fields that have same name after seeing the first one. The first such field will always be the expected actual type of the field.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -72,15 +72,15 @@ func Compile(scope Scope, f *semantic.FunctionExpression, in semantic.MonoType) 
 // inType and mapping any variables to the value in the other record.
 // If the input type is not a type variable, it will check to ensure
 // that the type in the input matches or it will return an error.
-func substituteTypes(subst map[uint64]semantic.MonoType, inType, in semantic.MonoType) error {
+func substituteTypes(subst map[uint64]semantic.MonoType, inferredType, actualType semantic.MonoType) error {
 	// If the input isn't a valid type, then don't consider it as
 	// part of substituting types. We will trust type inference has
 	// the correct type and that we are just handling a null value
 	// which isn't represented in type inference.
-	if in.Nature() == semantic.Invalid {
+	if actualType.Nature() == semantic.Invalid {
 		return nil
-	} else if inType.Kind() == semantic.Var {
-		vn, err := inType.VarNum()
+	} else if inferredType.Kind() == semantic.Var {
+		vn, err := inferredType.VarNum()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


compiler/compiler.go
```diff
@@ -72,15 +72,15 @@ func Compile(scope Scope, f *semantic.FunctionExpression, in semantic.MonoType)
 // inType and mapping any variables to the value in the other record.
 // If the input type is not a type variable, it will check to ensure
 // that the type in the input matches or it will return an error.
-func substituteTypes(subst map[uint64]semantic.MonoType, inType, in semantic.MonoType) error {
+func substituteTypes(subst map[uint64]semantic.MonoType, inferredType, actualType semantic.MonoType) error {
```
wolffcm, line 75: I changed the name of the arguments here just to help make the code clearer.

compiler/compiler.go
```diff
@@ -143,25 +143,30 @@ func substituteTypes(subst map[uint64]semantic.MonoType, inType, in semantic.Mon
...
 				return err
 			}
 
 			// Record the name of the property in the input type.
 			name := lprop.Name()
+			if containsStr(names, name) {
+				// The input type may have the same field twice if the record was
+				// extended with {r with ...}
+				continue
+			}
```
wolffcm, line 164: This is the actual fix for the bug right here.

compiler/compiler.go
fchikwekwe: Thank you for this.

stdlib/universe/map_field_type_change_test.flux
```diff
@@ -0,0 +1,33 @@
...
+,,0,m,f,2018-01-01T00:00:00Z,hello
+"
+
+t_map_field_type_change = (table=<-) =>
+    (table
+        |> range(start: 2018-01-01T00:00:00Z)
+        |> drop(columns: ["_start", "_stop"])
+        |> map(fn: (r) => ({r with _value: 2.0})) // establish _value as a double column in output
+        |> map(fn: (r) => ({r with _value: "hello"})) // convert to a string
+        |> filter(fn: (r) => r._value == "hello") // previously this would produce an error
```
fchikwekwe, line 29: Okay, cool. Super clear test case.

fchikwekwe (approved):
This is a good catch. And thanks for writing super clear test cases.

PR #3222, opened 2020-09-30
refactor: remove builtins file

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


interpreter/interpreter_test.go
```diff
@@ -707,8 +707,8 @@ func TestStack(t *testing.T) {
 			FunctionName: "window",
 			Location: ast.SourceLocation{
 				File:   "universe.flux",
-				Start:  ast.Position{Line: 128, Column: 12},
-				End:    ast.Position{Line: 128, Column: 49},
+				Start:  ast.Position{Line: 135, Column: 12},
+				End:    ast.Position{Line: 135, Column: 49},
```
nathanielc, line 711: why did this change?

jlapacik: This test should probably be refactored. It checks that the source location given by a runtime error is correct. In this case the source location is that of the window function in the standard library. So anytime `universe.flux` is updated this test will fail. I chose to just update the location in the test instead of removing the test.

yzhang1991 (approved):

PR #3228, opened 2020-10-01
feat(influxdb): moves functions from v1 package to schema package

### LINKED ISSUE #2609: Move functions in "influxdata/influxdb/v1" package to part of "influxdata/influxdb" package
The `v1` package name no longer makes sense for the helper functions in the package. They are should be part of the influxdb package.

The `json` and `databases` functions should remain in the `v1` namespace as those truly are specific to v1 concepts.

 For backwards compatibility we should leave the `v1` package in place but implement it in terms of the `influxdb` package. 

Something like this :

```
package influxdb

fieldsAsCols = // implementation goes here
```

```
package v1

import "influxdata/influxdb"

fieldsAsCols = influxdb.fieldsAsCols
```

This way the `v1` package API remains the same while still moving/adding the functions to the correct namespace.

nathanielc:
Originally the intent was that the concepts of measurements, fields and
tags would remain 1.x concepts and so functions for them were placed
into a v1 package. Its clear now that these functions are useful outside
the context of 1.x databases and buckets. As such they have been
promoted to a new `influxdata/influxdb/schema` package.

Fixes #2609 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


wolffcm (commented):
This looks good, should we move the relevant `_test.flux` tests from the `v1` package to `schema` as well? Or copy them over? I guess either would be fine with me.

nathanielc:
@wolffcm That is a good idea, I have moved the test files. I also discovered moving files was a tricky thing to get right with our build steps and created this to make that easier in the future https://github.com/influxdata/flux/pull/3232

wolffcm (approved):

PR #3229, opened 2020-10-01
Calculating a percentage - 1 pass over the data

timhallinflux:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/user_docs/flux-by-example.md
+++ b/docs/user_docs/flux-by-example.md
@@ -155,3 +155,26 @@ csv.from(url: "https://influx-testdata.s3.amazonaws.com/noaa.csv")
 |> map(fn: (r) => ({ r with jsonStr: string(v: json.encode(v: {"location":r.location,"mean":r._value}))}))
 |> map(fn: (r) => ({r with status_code: http.post(url: "http://somehost.com/", headers: {x:"a", y:"b"}, data: bytes(v: r.jsonStr))}))
 ```
+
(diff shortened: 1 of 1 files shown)


docs/user_docs/flux-by-example.md
```diff
@@ -155,3 +155,26 @@ csv.from(url: "https://influx-testdata.s3.amazonaws.com/noaa.csv")
 |> map(fn: (r) => ({ r with jsonStr: string(v: json.encode(v: {"location":r.location,"mean":r._value}))}))
 |> map(fn: (r) => ({r with status_code: http.post(url: "http://somehost.com/", headers: {x:"a", y:"b"}, data: bytes(v: r.jsonStr))}))
 ```
+
+# Calculate a percentage using a defined set with a filter and only querying the data once
```
nathanielc, line 159: This query will query the data twice because of some internal implementation details. Using a `join` can do it in a single pass.

timhallinflux: ooops.  Can you share that example so we document this correctly?  :-)

timhallinflux: I think I figured out what you meant!

timhallinflux: I think I figured out what you were suggesting.  But, please review my change(s) and confirm?

nathanielc (approved):

PR #3230, opened 2020-10-01
fix(quantile): sets default quantile method when not specified

### LINKED ISSUE #2648: Quantile should require the `method` parameter or assume a valid default.
A common mistake with the `quantile` function is to forget the `method` parameter. If you do you get this error message:

```
error calling function “quantile”: unknown method
```

That makes it sound like the `quantile` method does not exist. But what it is trying to say is that the `method` parameter was missing and therefore unknown.

We should update the error from `unknown method` to `invalid quantile method, please check your 
method" parameter` or something along those lines

Additionally it seems that the type signature of `quantile` does not require the `method` parameter but then will error if its not provided. We should either require it at the type level, or we should give it a default value so its not required.  I vote is to default to `estimate_tdigest` so you only need the `method` parameter if you are doing something more specific.

nathanielc:
The quantile method type signature specified that the "method" parameter
was optional but did not apply a default if it was missing. This change
now applies the default.

Fixes #2648

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/quantile_defaults_test.flux
+++ b/stdlib/universe/quantile_defaults_test.flux
@@ -1,0 +1,34 @@
+package universe_test
+
+
+import "testing"
+
+option now = () =>
+	(2030-01-01T00:00:00Z)
+
+inData = "
+#datatype,string,long,string,string,dateTime:RFC3339,double
+#group,false,false,true,true,false,false
+#default,_result,,,,,
+,result,table,_measurement,_field,_time,_value
+,,0,SOYcRk,NC7N,2018-12-18T21:12:45Z,-61.68790887989735
+,,0,SOYcRk,NC7N,2018-12-18T21:12:55Z,-6.3173755351186465
+,,0,SOYcRk,NC7N,2018-12-18T21:13:05Z,-26.049728557657513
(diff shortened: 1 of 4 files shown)


stdlib/universe/quantile_defaults_test.flux
```diff
@@ -0,0 +1,34 @@
...
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,string,string,double
+#group,false,false,true,true,true,true,false
+#default,_result,,,,,,
+,result,table,_start,_stop,_measurement,_field,_value
+,,0,2018-01-01T00:00:00Z,2030-01-01T00:00:00Z,SOYcRk,NC7N,29.50336437998469
+"
+t_quantile = (table=<-) =>
+	(table
+		|> range(start: 2018-01-01T00:00:00Z)
+		|> quantile(q: 0.75))
```
rockstar, line 31: I assume we have tests for `quantile` when we explicitly set a default quantile method?

rockstar (approved):

stdlib/universe/quantile_defaults_test.flux
nathanielc: yes, in fact this is a copy of that exact test case, but with the `method` param removed.

PR #3233, opened 2020-10-02
feat(stdlib/testing): fix field type error in test

fchikwekwe:
This is a PR to update `aggregate_window_median_test.flux` so that it will pass in influxDB. The ultimate plan is to cherry-pick this commit onto a release branch for v0.83.x and then create a patch release of v0.83.x. to use in influxDB.

The change was tested locally and [appears to resolve the issue](https://github.com/influxdata/influxdb/issues/19565) that this test was causing in influxDB.

### DIFF
--- a/stdlib/testing/chronograf/aggregate_window_median_test.flux
+++ b/stdlib/testing/chronograf/aggregate_window_median_test.flux
@@ -21,7 +21,7 @@ inData = "
 ,,1,2018-05-22T00:00:50Z,disk,percentage,disk2s1,apfs,host.local,92.2
 
 
-#datatype,string,long,dateTime:RFC3339,string,string,string,string,string,long
+#datatype,string,long,dateTime:RFC3339,string,string,string,string,string,double
 #group,false,false,false,true,true,true,true,true,false
 #default,_result,,,,,,,,
 ,result,table,_time,_measurement,_field,device,fstype,host,_value
@@ -97,7 +97,6 @@ agg_window_median_fn = (table=<-) => table
     |> filter(fn: (r) => r._measurement == "disk" or r._measurement == "mem")
     |> filter(fn: (r) => r.host == "host.remote")
     |> window(period: 30s)
-    |> toFloat()
     |> median()
     |> group(columns: ["_value", "_time", "_start", "_stop"], mode: "except")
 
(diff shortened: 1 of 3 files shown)


stdlib/testing/chronograf/aggregate_window_median_test.flux
```diff
@@ -84,20 +84,14 @@ outData = "
...
-,,0,2018-05-22T00:00:30Z,2018-05-22T00:01:00Z,percentage,mem,host.remote,35
+
 "
 
 agg_window_median_fn = (table=<-) => table
     |> range(start: 2018-05-22T00:00:00Z, stop: 2018-05-22T00:01:00Z)
     |> filter(fn: (r) => r._measurement == "disk" or r._measurement == "mem")
     |> filter(fn: (r) => r.host == "host.remote")
     |> window(period: 30s)
-    |> toFloat()
```
fchikwekwe, line 100: I felt comfortable removing this because it didn't seem necessary to include different value types in the same aggregation. This is also the only test that I could find where an explicit type conversion is used to "fix" the resulting table in order to aggregate. I'm not sure if that was an intentional feature of this test. 

If the intention is that type conversion functions don't cause field type errors, then additional investigation may be required.

scbrickley (approved):

stdlib/testing/chronograf/aggregate_window_median_test.flux
wolffcm: Yeah, it seems like just an oversight when the test was originally written.

Do you have a sense of why it doesn't work with tsm1?

I see that there are four series with measurement `disk` and field `percentage`, and two of them are longs and two are doubles. I guess tsm1 requires the data types to agree for all series of a field? (tsm2 just requires all the values in a series to have the same type, so it's possible to have two series of the same field with different types, if they just differ by one tag.)

wolffcm (approved):
Looks good.

stdlib/testing/chronograf/aggregate_window_median_test.flux
fchikwekwe: Yeah, this was my conclusion as well. It seems like aggregating on tables of different types isn't allowed anymore. I don't know if this is just a new convention to follow, or if its worth creating an issue around this.

PR #3240, opened 2020-10-02
fix(formatter): add support for comments to builtin statement

### LINKED ISSUE #3239: Comments are lost with builtin statements
Parsing and formatting this code (using Rust):

```
// this is builtin
builtin x : int
```

results in this code

```
builtin x : int
```

nathanielc:
The builtin statement parsing would loose comments and the formatter
would not reproduce them. This fixes both the parser and formatter in
regards to builtin statements.

Fixes #3239

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


Bolladeen (approved):

adrian-thurston (changes requested):
Would be good to capture and format comments in front of the colon, as in this commit 3e9e213d0a792c435a1fa4bfaabeb3dadf35c393 It's unlikely someone would put a comment there, but I think it's better to cover all cases. The `id` and the `type-expression` will be covered already because they reuse other parts of the parser.

nathanielc:
@adrian-thurston Thanks, for the example commit, I have made the changes to preserve any comments on the `colon` token and added appropriate tests.

adrian-thurston (approved):
Looks good, thanks for the update.

PR #3247, opened 2020-10-09
refactor: remove internal fbgen tool

wolffcm:
When I had to write code that would deserialize a flatbuffer
representing a semantic graph into a native Go semantic graph,
I wrote a tool called `fbgen` to generate the Go code to do the
deserializing, since it was so repetitive.

However, now that task is complete, and sometimes we need to make
changes to the FlatBuffers schema for our semantic graph and that causes
fbgen to break the build in a confusing way, which costs precious
developer time.

So this commit just renames semantic/flatbuffers_gen.go and makes it
into a normal non-generated file.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


fchikwekwe (approved):
LGTM

PR #3248, opened 2020-10-09
feat(values): add duration conversion

wolffcm:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


wolffcm:
This was intended to bring the release branch up to date with a commit tagged with `v0.83.3`, but we're not sure if either way of merging will actually retain the same tagged commit. This version is current in OSS, but will be upgraded on Monday, so I'll just close this PR.

PR #3262, opened 2020-10-15
feat(profiler): Aggregate results for operator profiler

### LINKED ISSUE #3203: Aggregate the operator profiler result
1. Move the instantiation of the profiler to be inside the interpreter.Eval(), handled on the fly so the profiler is visible to the tableFind and chain function
2. Aggregate the profiling result as we receive them grouped by type and label. min, max, count, mean, sum

scbrickley:
Closes #3203

Groups profiler results by operation type and label, and aggregates the results into count, min, max, sum, and mean.

Allows explicit times to be passed into `OperatorProfilerSpan.finish()` and `StartSpanFromContext` to make testing easier.

Adds `GetSortedResult()` method to the `Profiler` interface to allow for easier testing without requiring results to be sorted in production.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/profiler.go
+++ b/execute/profiler.go
@@ -15,7 +15,7 @@ import (
 type Profiler interface {
 	Name() string
 	GetResult(q flux.Query, alloc *memory.Allocator) (flux.Table, error)
-	GetSortedResult(q flux.Query, alloc *memory.Allocator, sortKeys []string, desc bool) (flux.Table, error)
+	GetSortedResult(q flux.Query, alloc *memory.Allocator, desc bool, sortKeys ...string) (flux.Table, error)
 }
 
 type CreateProfilerFunc func() Profiler
@@ -53,11 +53,7 @@ type OperatorProfilingSpan struct {
 }
 
 func (t *OperatorProfilingSpan) finish(finishTime time.Time) time.Time {
-	if finishTime.IsZero() {
-		t.Result.Stop = time.Now()
-	} else {
-		t.Result.Stop = finishTime
-	}
+	t.Result.Stop = finishTime
 	if t.profiler != nil && t.profiler.chIn != nil {
 		t.profiler.chIn <- t.Result
 	}
@@ -65,7 +61,7 @@ func (t *OperatorProfilingSpan) finish(finishTime time.Time) time.Time {
 }
 
 func (t *OperatorProfilingSpan) Finish() {
-	finishTime := t.finish(time.Time{})
+	finishTime := t.finish(time.Now())
 	if t.Span != nil {
 		t.Span.FinishWithOptions(opentracing.FinishOptions{
 			FinishTime: finishTime,
@@ -93,7 +89,8 @@ type operatorProfilingResultAggregate struct {
 	resultMean    float64
 }
 
-type operatorProfilerGroupedAggregates = map[string]map[string]operatorProfilingResultAggregate
+type operatorProfilerLabelGroup = map[string]*operatorProfilingResultAggregate
+type operatorProfilerTypeGroup = map[string]operatorProfilerLabelGroup
 
 type OperatorProfiler struct {
 	// Receive the profiling results from the spans.
@@ -107,18 +104,20 @@ func createOperatorProfiler() Profiler {
 		chOut: make(chan operatorProfilingResultAggregate),
 	}
 	go func(p *OperatorProfiler) {
-		aggs := make(operatorProfilerGroupedAggregates)
+		aggs := make(operatorProfilerTypeGroup)
 		for result := range p.chIn {
-			// Because we're using nested maps, we need to check if the inner map
-			// exists before we started manipulating it. If it doesn't, we create one.
 			_, ok := aggs[result.Type]
 			if !ok {
-				aggs[result.Type] = make(map[string]operatorProfilingResultAggregate)
+				aggs[result.Type] = make(operatorProfilerLabelGroup)
+			}
+			_, ok = aggs[result.Type][result.Label]
+			if !ok {
+				aggs[result.Type][result.Label] = &operatorProfilingResultAggregate{}
 			}
 			a := aggs[result.Type][result.Label]
 
 			// Aggregate the results
-			a.resultCount += 1
+			a.resultCount++
 			duration := result.Stop.Sub(result.Start).Nanoseconds()
 			if duration > a.resultMax {
 				a.resultMax = duration
@@ -127,7 +126,6 @@ func createOperatorProfiler() Profiler {
 				a.resultMin = duration
 			}
 			a.resultSum += duration
-			aggs[result.Type][result.Label] = a
 		}
 
 		// Write the aggregated results to chOut, where they'll be
@@ -137,7 +135,7 @@ func createOperatorProfiler() Profiler {
 				agg.resultMean = float64(agg.resultSum) / float64(agg.resultCount)
 				agg.operationType = typ
 				agg.label = label
-				p.chOut <- agg
+				p.chOut <- *agg
 			}
 		}
 		close(p.chOut)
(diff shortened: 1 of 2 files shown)


scbrickley:
Made a pretty big oversight on this and failed to group the results by type & label. Fix for that will be up later today.

execute/profiler_test.go
```diff
@@ -30,38 +31,129 @@ func TestOperatorProfiler_GetResult(t *testing.T) {
...
+		})
+		wg.Done()
+	}
+	for i := 0; i < count; i++ {
+		label := fmt.Sprintf("lab%d", i%2)
+		go fn("type0", label, ctx, 100+i*i)
+		// Waiting between each loop seems to ensure better consistency of row order
+		// for final table. However, this is still not guaranteed, becausethe result
+		// aggregates are stored in a map, and thus order is not 100% guaranteed.
+		time.Sleep(100 * time.Millisecond)
```
scbrickley, line 137: Wanted to draw attention to this. It seems that this was a problem in the old version of the test because of concurrency issues. Now it's a problem because of concurrency issues *and* the fact that we can't guarantee the order of values in a map.

scbrickley: Just to be clear, this shouldn't be a problem in production, but it may cause these two tests to randomly fail every once in a while. Not sure if there's a good way around this.

execute/profiler_test.go
```diff
@@ -18,50 +18,136 @@ import (
...
+		})
+		wg.Done()
+	}
+	for i := 0; i < count; i++ {
+		label := fmt.Sprintf("lab%d", i%2)
+		go fn("type0", label, ctx, 100+i*i)
+		// Waiting between each loop seems to ensure better consistency of row order
+		// for final table. However, this is still not guaranteed, because the result
+		// aggregates are stored in a map, and thus order is not 100% guaranteed.
+		time.Sleep(100 * time.Millisecond)
```
scbrickley, line 131: Wanted to draw attention to this. It seems that this was a problem in the old version of the test because of concurrency issues. Now it's a problem because of concurrency issues *and* the fact that we can't guarantee the order of values in a map.

This shouldn't be a problem in production, but it will cause these two tests to fail every once in a while. Not sure if there's a good way around that.

adrian-thurston (commented):
Sorry I'm a bit late on this review. With regards to the testing, I'm thinking we should collect the results into some local data structure, sort it, then compare that against expected results.  Then we can guarantee the tests won't sporadically fail and can take out the waits.

Also I think we should try to remove the `Sleep` after waiting for the goroutines, but I understand we don't have anything to syncronize on? (See below)

execute/profiler_test.go
```diff
@@ -18,50 +18,136 @@ import (
...
+		1101,
+		1109,
+		2210,
+		1105.0,
+	))
+	wg.Wait()
+	// Wait a bit for the profiling results to be added.
+	// In the query code path this is guaranteed because we only access the result
+	// after the query finishes execution AND its result tables are read and encoded.
+	time.Sleep(100 * time.Millisecond)
```
adrian-thurston, line 81: Is there anything we can do to remove this `Sleep`? Seems like it is ripe for flakiness that causes the test to get skipped sometime in the future.

scbrickley: I'm not sure why that's in there either. @yzhang1991 might have a better idea.

execute/profiler_test.go
scbrickley: This problem has been resolved by: https://github.com/influxdata/flux/pull/3262/commits/3c991325cbb8e015c8561bfc7ebddaff19cd06be

execute/profiler_test.go
scbrickley: Hmm... I think this is no longer necessary with the changes I've made. I'm gonna try the tests without the Sleep calls.

execute/profiler.go
```diff
@@ -52,16 +52,20 @@ type OperatorProfilingSpan struct {
...
+	}
+	if t.profiler != nil && t.profiler.chIn != nil {
+		t.profiler.chIn <- t.Result
 	}
 	return t.Result.Stop
 }
 
 func (t *OperatorProfilingSpan) Finish() {
-	finishTime := t.finish()
+	finishTime := t.finish(time.Time{})
```
yzhang1991, line 68: Why don't we just pass in `time.Now()` here?

execute/profiler.go
```diff
@@ -79,25 +83,66 @@ func (t *OperatorProfilingSpan) FinishWithOptions(opts opentracing.FinishOptions
...
-			p.mu.Lock()
-			p.results = append(p.results, result)
-			p.mu.Unlock()
+		aggs := make(operatorProfilerGroupedAggregates)
+		for result := range p.chIn {
+			// Because we're using nested maps, we need to check if the inner map
+			// exists before we started manipulating it. If it doesn't, we create one.
+			_, ok := aggs[result.Type]
+			if !ok {
+				aggs[result.Type] = make(map[string]operatorProfilingResultAggregate)
```
yzhang1991, line 116: I recommend you create another type alias for `map[string]operatorProfilingResultAggregate`. So both `operatorProfilerGroupedAggregates` and here can get clearer?

execute/profiler.go
```diff
@@ -79,25 +83,66 @@ func (t *OperatorProfilingSpan) FinishWithOptions(opts opentracing.FinishOptions
...
+			a.resultCount += 1
+			duration := result.Stop.Sub(result.Start).Nanoseconds()
+			if duration > a.resultMax {
+				a.resultMax = duration
+			}
+			if duration < a.resultMin || a.resultMin == 0 {
+				a.resultMin = duration
+			}
+			a.resultSum += duration
+			aggs[result.Type][result.Label] = a
```
yzhang1991, line 130: You can let the map store the pointer: `map[string]*operatorProfilingResultAggregate`, so you can directly mutate the data in `a` and don't need to assign it again to the map `aggs[result.Type][result.Label] = a`

execute/profiler.go
```diff
@@ -128,12 +173,81 @@ func (o *OperatorProfiler) GetResult(q flux.Query, alloc *memory.Allocator) (flu
...
+	if err != nil {
+		return nil, err
+	}
+	return tbl, nil
+}
+
+// GetSortedResult is identical to GetResult, except it accept it calls Sort()
+// on the ColListTableBuilder to make testing easier.
+// sortKeys and desc are passed directly into the Sort() call
+func (o *OperatorProfiler) GetSortedResult(q flux.Query, alloc *memory.Allocator, sortKeys []string, desc bool) (flux.Table, error) {
```
yzhang1991, line 230: I think `GetSortedResult` and `GetResult` can share a common helper function to increase code reusability?

execute/profiler_test.go
yzhang1991: It was possible that when you call `GetResult()`, the goroutines have not finished sending the results back.
I think this is still true for this test case, you can test it with a higher `count` and remove the `Sleep` to see if you can still get the right result.

scbrickley: Increasing `count` in the tests still produced the correct result. I think because `span.Finish`, `createProfiler`, and `GetResult` are all syncing on channels, there isn't really a problem with waiting for all the various goroutines to finish anymore. I've removed the `time.Sleep` call for now, but I can put it back in if you feel it's really necessary.

scbrickley:
Addressed some of your feedback @yzhang1991. Good suggestions!

scbrickley:
Looks like there are some conflicts after #3264 merged. I'll see what I can do about resolving them.

scbrickley:
Conflicts resolved :+1:

adrian-thurston (approved):
Looks good. Just gave it a smoke test on a local cluster. All good.

PR #3263, opened 2020-10-16
feat: add series cardinality function to influxdb package

jsternberg:
This adds a `cardinality()` function to the influxdb package to
determine the cardinality of a bucket with a filter for a specific
range. This is a source and can be chained to flux transformations.

To implement this, a dependency interface was created to act as an
influxdb provider. This dependency will cover reading and eventually
writing to influxdb. The interface allows us to define flux
transformations that interact with influxdb both in a remote way, such
as through http calls, and embedded within influxdb itself.

The interface is divided into a provider and a reader. The provider
contains a complete list of the different types of readers which is one
for each one. This provider is documented to be able to change and it is
advised to use the `UnimplementedProvider` as a base to ensure that all
of the methods continue to be implemented.

The readers are divided into one specific method for each method of
reading. If a provider returns a reader interface, it is expected that
the functionality is supported.

This allows us to have different implementations and also to allow an
implementation to make decisions around which type of reader it wants to
use. For example, a direct storage access implementation might fallback
to the http provider when a host is used.

The interface for directly reading through a normal read has been
stubbed out but hasn't been implemented. But, this will allow us to
consolidate the reading code to a single location instead of relying on
the planner and allow more of the flux code to live in the flux
repository instead of influxdb.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/cardinality_test.go
+++ b/stdlib/influxdata/influxdb/cardinality_test.go
@@ -18,32 +18,29 @@ import (
 	"github.com/influxdata/flux/values/valuestest"
 )
 
-func TestFrom_NewQuery(t *testing.T) {
+func TestCardinality_NewQuery(t *testing.T) {
 	tests := []querytest.NewQueryTestCase{
 		{
-			Name:    "from no args",
-			Raw:     `from()`,
+			Name:    "cardinality no args",
+			Raw:     `influxdb.cardinality()`,
 			WantErr: true,
 		},
 		{
-			Name:    "from unexpected arg",
-			Raw:     `from(bucket:"telegraf", chicken:"what is this?")`,
+			Name:    "cardinality unexpected arg",
+			Raw:     `influxdb.cardinality(bucket:"telegraf", chicken:"what is this?")`,
 			WantErr: true,
 		},
 		{
-			Name: "from with database",
-			Raw:  `from(bucket:"mybucket") |> range(start:-4h, stop:-2h) |> sum()`,
+			Name: "cardinality with bucket and range",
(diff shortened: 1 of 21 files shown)


stdlib/influxdata/influxdb/cardinality_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+
+import influxdb "influxdata/influxdb"
+
+influxdb.cardinality(
+	bucket: "telegraf",
+	start: 2020-10-22T09:29:00Z,
+	stop: 2020-10-22T09:30:00Z,
+	predicate: (r) => {
+		return r["_value"] >= 0.0
```
jlapacik, line 298: Nit: could you change this to `r._value`. I know `r["_value"]` is valid flux but I'd like us to use the former throughout the codebase when possible.

stdlib/influxdata/influxdb/cardinality_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+import influxdb "influxdata/influxdb"
+import math "math"
+
+influxdb.cardinality(
+	bucket: "telegraf",
+	start: 2020-10-22T09:29:00Z,
+	stop: 2020-10-22T09:30:00Z,
+	predicate: (r) => {
+		return r["_value"] >= math["pi"]
```
jlapacik, line 361: Same here.

jlapacik (approved):

stdlib/influxdata/influxdb/cardinality_test.go
jsternberg: I mostly did this because it's automated code that generates this and I didn't know what the conditions were to check for a valid identifier. I didn't want to accidentally generate invalid flux if there was a space or something else.

PR #3269, opened 2020-10-19
fix(formatter): make formatter smarter about multiline and parens

nathanielc:
This change make 3 changes to the formatter:
    1. Records that are multiline in the original source remain
       multiline in the formatted output
    2. Parens that have comments are preserved even if not needed
    3. Function bodies are now on the same line without parens if
       possible

These changes in combination mean that where previously the formatter
would output this:

from(bucket: "foo")
    |> range(start: -5m)
    |> filter(fn: (r) =>
        (r._measurement == "m"))

The output is now

from(bucket: "foo")
    |> range(start: -5m)
    |> filter(fn: (r) => r._measurement == "m")

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
FWIW here is a diff of how the new formatter would format universe.flux, there are still some things that could be improved but I think overall its a good improvement.


[universe.diff.txt](https://github.com/influxdata/flux/files/5404901/universe.diff.txt)

adrian-thurston (approved):
Looks good! Much prefer to capture the AST as the user typed it, then in the formatter alter the tree, as you have done here.

PR #3271, opened 2020-10-20
fix(formatter): updates formatter to handle multiline type expressions

nathanielc:
There are two changes as part of this fix:

1. Several of the parser functions for type expressions could not handle
   trailing commas, they have been updated.
2. The formatting functions for type expressions now match the multiline
   behavior of the equivalent literal formatting

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


jlapacik (approved):
Love the trailing commas. Can you add a parser test with trailing commas?

nathanielc:
Thanks, I added the tests

PR #3276, opened 2020-10-23
feat(stdlib/contrib): add ServiceNow support

alespour:
This PR adds support for sending alerts / events to ServiceNow platform.

### Done checklist
- [X] Test cases written


timhallinflux:
@bpalani -- paid feature for cloud?

timhallinflux:
We should put this on hold until we better understand the implications of this: https://github.com/influxdata/feature-requests/issues/202

Let's make sure that we are using the EVENT API...not the Alert API.  It appears that the Alert API is really just there for testing...(and it does create an Event as part of this), but from the customer's perspective I believe they want to see an event land....and then ServiceNow can evaluate if/how they want to create an alert upstream from the event provided from InfluxDB.

alespour:
That is correct, as of now, this handler uses Table API, it has to be changed. Not ready for review.

alespour:
The handler implementation now uses web service (event) API.

stdlib/contrib/bonitoo-io/servicenow/README.md
```diff
@@ -0,0 +1,123 @@
...
+              if r._value < 1.0 then "critical"
+              else if r._value < 5.0 then "warning"
+              else "info",
+          additionalInfo: { "devId": r.dev_id }
+        })
+      )()
+
+## Contact
+
+Provide a way for users to get in touch with you if they have questions or need help using your package. What information you give is up to you, but we encourage providing those below.
```
rockstar, line 118: I'm not sure we need this paragraph, right? It's just part of the template?

rockstar (changes requested):
Thanks for doing this work. In a previous life, I would have loved this specific module, so I'm super excited to see this ecosystem expand. I just had one small change to request, but I'm happy to see this merged once it's complete.

stdlib/contrib/bonitoo-io/servicenow/README.md
alespour: You are absolutely right, that shouldn't be there.

skartikey (commented):
Looks good!

PR #3281, opened 2020-10-26
feat: CreateNumberedPhysicalNode

yzhang1991:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/lang/compiler_test.go
+++ b/lang/compiler_test.go
@@ -474,7 +474,7 @@ func (rule removeCount) Name() string {
 func (rule removeCount) Pattern() plan.Pattern {
 	return plan.Pat(universe.CountKind, plan.Any())
 }
-func (rule removeCount) Rewrite(ctx context.Context, node plan.Node) (plan.Node, bool, error) {
+func (rule removeCount) Rewrite(ctx context.Context, node plan.Node, nextNodeId *int) (plan.Node, bool, error) {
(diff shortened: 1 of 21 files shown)


lang/compiler_test.go
```diff
@@ -474,7 +474,7 @@ func (rule removeCount) Name() string {
 func (rule removeCount) Pattern() plan.Pattern {
 	return plan.Pat(universe.CountKind, plan.Any())
 }
-func (rule removeCount) Rewrite(ctx context.Context, node plan.Node) (plan.Node, bool, error) {
+func (rule removeCount) Rewrite(ctx context.Context, node plan.Node, nextNodeId *int) (plan.Node, bool, error) {
```
rockstar, line 477: This pattern, where we add a new parameter without using it, often indicates to me that we've got a leaky interface or abstraction. I see this pattern _a lot_ in this patch, so wondering you have thoughts about it.

rockstar (commented):
This _looks_ okay, except that I don't have any context for what this accomplishes. It would be really helpful if there was a PR description describing the _why_ of this change, because I'm not understanding the value of this patch in the grand scheme of our work. Maybe it's obvious in the scope of your work, but I haven't gotten to review any of that work.

Should there be tests to go along with this?

PR #3289, opened 2020-10-29
fix: properly truncate timestamps to beginning of window bounds

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/window.go
+++ b/execute/window.go
@@ -106,7 +106,11 @@ func (w Window) GetOverlappingBounds(b Bounds) []Bounds {
 // truncateByNsecs will truncate the time to the given number
 // of nanoseconds.
 func truncateByNsecs(t Time, d Duration) Time {
-	remainder := int64(t) % d.Nanoseconds()
+	dur := d.Nanoseconds()
+	remainder := int64(t) % dur
+	if remainder < 0 {
+		remainder += dur
+	}
 	return t - Time(remainder)
 }
 
--- a/execute/window_test.go
+++ b/execute/window_test.go
@@ -18,7 +18,7 @@ func TestNewWindow(t *testing.T) {
 			Period: values.ConvertDurationNsecs(time.Minute),
 			Offset: values.ConvertDurationNsecs(time.Second),
 		}
-		got := MustWindow(values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Second), false)
+		got := MustWindow(values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Minute), values.ConvertDurationNsecs(time.Second))
 		if !cmp.Equal(want, got) {
 			t.Errorf("window different; -want/+got:\n%v\n", cmp.Diff(want, got))
 		}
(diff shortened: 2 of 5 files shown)


execute/window.go
```diff
@@ -106,7 +106,11 @@ func (w Window) GetOverlappingBounds(b Bounds) []Bounds {
...
 // of nanoseconds.
 func truncateByNsecs(t Time, d Duration) Time {
-	remainder := int64(t) % d.Nanoseconds()
+	dur := d.Nanoseconds()
+	remainder := int64(t) % dur
+	if remainder < 0 {
+		remainder += dur
+	}
 	return t - Time(remainder)
 }
```
jlapacik, line 115: `%` can return a negative value so we need to normalize the remainder as `t - Time(remainder)` expects a non-negative input.

execute/window_test.go
```diff
@@ -208,6 +208,54 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
 				Stop:  execute.Time(5*time.Minute + 30*time.Second),
 			},
 		},
+		{
+			name: "truncate before offset",
+			w: MustWindow(
+				values.ConvertDurationNsecs(5*time.Second),
+				values.ConvertDurationNsecs(5*time.Second),
+				values.ConvertDurationNsecs(2*time.Second), false),
```
nathanielc, line 216: Was refreshing my memory on what MustWindow does and noticed that the bool arg months is not used. Should we remove that as part of this change?

values/time.go
```diff
@@ -212,6 +212,14 @@ func (d Duration) Nanoseconds() int64 { return d.nsecs }
 // It will ensure that the output duration is the smallest positive
 // duration that is the equivalent of the current duration.
 func (d Duration) Normalize(interval Duration) Duration {
+	if d.Months() != 0 && d.Nanoseconds() == 0 &&
+		interval.Months() != 0 && interval.Nanoseconds() == 0 {
+		return Duration{
+			months:   d.Months() % interval.Months(),
```
nathanielc, line 218: Below the nanosecond logic checks for negative offsets. Do we need to do the same here for months?

execute/window_test.go
```diff
@@ -208,6 +208,54 @@ func TestWindow_GetEarliestBounds(t *testing.T) {
...
+			w: MustWindow(
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(5),
+				values.ConvertDurationMonths(2), true),
+			t: mustParseTime("1970-04-01T00:00:00Z"),
+			want: execute.Bounds{
+				Start: mustParseTime("1970-03-01T00:00:00Z"),
+				Stop:  mustParseTime("1970-08-01T00:00:00Z"),
+			},
+		},
```
nathanielc, line 258: Could you add a few more test cases where the offset itself is negative? There seem to be specific code branches for the case where offset is negative and I want to make sure we are testing that code path.

nathanielc (changes requested):
Looks great and the change makes sense. I have requested a few more tests cases to ensure we are covering more edge cases.

execute/window_test.go
jlapacik: I did not notice that, yes I'll remove it. I wish the go compiler would identify unused function params.

values/time.go
jlapacik: I added a check for negative offsets.

execute/window_test.go
jlapacik: We already had one for nanoseconds, I added one for calendar durations.

nathanielc (approved):

PR #3290, opened 2020-10-29
test: reorganize and add more metaquery Flux tests

wolffcm:
This makes some changes to the metaquery tests in `stdlib/influxdata/influxdb/schema`:
- Some of these were failing in `influxdb` because the type of fields was not consistent (like changing from `long` to `double` for a different tag value)
- There was a duplicate test, which I deleted
- I standardized the names of metaqueries to be `show_measurements*`, `show_fields*`, `show_tags*` or `show_tag_keys*`
- I added more tests to ensure we cover fields, measurements and tag values in general for the `tagValues` pushdown, for both the fast and slow path


stdlib/influxdata/influxdb/schema/measurement_field_keys_test.flux
```diff
@@ -1,90 +0,0 @@
-package schema_test
```
wolffcm, line 1: This was the same as `show_fields_with_pred_test.flux`.

yzhang1991 (approved):

PR #3294, opened 2020-10-30
docs(SPEC): add design for formal testing in Flux

nathanielc:
Flux currently has an undocumented `test` statement but it is very
limited in scope. The test must test the result of a Flux query and
cannot test arbitrary expressions.

This update to the Flux spec formalizes testing as a first class concept
to the Flux language.

As can be seen by the example in the SPEC the idea is that multiple
tests can be defined making use of various assertion methods in the
testing package. The testing package will work much like the profiler in
that it will add two results to the output of a query. One result which
is a mapping of test name with a pass/fail status and another which is a
table per test with each assertion results as a row. Assertions will be
identified by their source location.

In order to isolate test runs each test block will be rewritten as if it
were the only test in the file. A file with multiple test blocks will
get exploded into many test files/packages.

Example of how I expect that could work:

    package foo_test
    import "testing"

    answer = 42

    test addition {
        testing.assertEqual(got: 13 + 29, want: answer)
    }

    test multiplication {
        testing.assertEqual(got: 6 * 7, want: answer)
    }

Becomes two separate packages which would look like this:

    //addition.flux
    package main

    import "testing"

    answer = 42

    testing.assertEqual(got: 13 + 29, want: answer)

    //multiplication.flux
    package main

    import "testing"

    answer = 42

    testing.assertEqual(got: 6 * 7, want: answer)

This has the effect that each test is run in complete isolation but can
share common setup code across tests easily.
Additionally options can be specified per test case with this method.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1001,6 +1001,7 @@ A statement controls execution.
 
     Statement = OptionAssignment
               | BuiltinStatement
+              | TestStatement
               | VariableAssignment
               | ReturnStatement
               | ExpressionStatement .
@@ -1113,6 +1114,30 @@ All such values must have a corresponding builtin statement to declare the exist
 Example
 
     builtin filter : (<-tables: [T], fn: (r: T) => bool) => [T]
+
+## Test Statements
+
+A test statment defines a block of code that describes a test.
+A test is a block of code to execute that can make assertions about its behavior.
+
+    TestStatement = "test" identifier  Block
+
+The test statement block is treated as if it were a file block within its own package block.
+Any statements inside the enclosing file block of the test statement are also part of the test package.
+Therefore each test case can be understood to execute in isolation from other tests in the same file.
+
+Example
+
+    import "testing"
+
+    test addition {
+        testing.assertEqual(got: 1 + 1, want: 2)
+    }
+
+    test changeNow {
+        option now = () => 2020-01-01T00:00:00Z
+        testing.assertEqual(got: now(), want: 2020-01-01T00:00:00Z)
+    }
 
 ### Date/Time constants
 


rockstar:
How would we handle the following code:

`test addition { testing.assertEqual(got: 1+1, want: 2) }`

...and then in another file, we have...

`test addition { testing.assertEqual(got: aThing(), want: {}) }`

There's a name conflict that isn't clear to the user, and what I definitely _don't_ want is to have a global namespace of test names.

While this is a conversation about the spec, I think speaking to implementation might be helpful, at least in the case of primitives. I believe we have three basic structures that would be needed for structure:

```
type Test struct {
  Name string
  Script string // ...or AST
  // I would _very_ much like to have a finer grained control of error/failure than just
  // "is there an error here?" A test that errors and a test that fails should be treated
  // differently.
  Err error
}
type TestSuite struct {
  Tests []Test
}
type TestRunner struct {
  TestSuites []TestSuites
}
```

In this case, the `TestRunner` would contain logic for finding tests, constructing `TestSuite`s out of them, setting up the environment (e.g. fixtures), running the test suites, and reporting their results. I see a `TestSuite` as a "namespace" for tests. In the case of the current spec changes, a single file would constitute a `TestSuite`, and would contain the logic that would expand the tests out into a `Test` instance that could be a logical flux script as a string or an AST or something else appropriate.

I think the SPEC should also include some assertions about the `TestRunner`'s format for results, particularly because `flux test` would be the first-party implementation of `TestRunner`. This should probably be some off the shelf format, but I'm not sure I have much of an opinion on what that format is.

Thoughts?

nathanielc:
@rockstar I like the TestSuite abstraction, it helps group tests logically. I concur we do not want a global namespace. Would you use the file name as the TestSuite name? 

Or should we use the Flux `package` name? In that case, the namespace is for the package and so you could create a conflict across multiple test files in the same package. That is how Go works and it hasn't generally been a problem because tests tend to be named after the thing they are testing which gives them generally unique names.

An advantage of using the `package` as the namespace instead of the file name is that on the CLI when you specify which tests you want to run you only have to remember the package name not the file name. Currently in Flux the file name is not used as part of scoping so its not something you normally think about.

Thoughts?

PR #3295, opened 2020-10-30
fix(formatter): add array multiline support to the formatter

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


adrian-thurston (approved):
Looks good. Down the road might be interesting to notice what's in the array and if is a long list of numbers, maybe put X number of items on a line.

PR #3297, opened 2020-11-02
feat: make sure the query plan nodes have unique ids

yzhang1991:
A lot of our planner rules transform the existing plan nodes into completely different ones. For example, `PushDownGroupAggregateRule` collapses `range |> window |> agg` into `ReadWindowAggregate`.

In this process, we usually manually create the new physical nodes without numbering them. So if such planner rules are applied more than once in a query (a query with join or table functions or `chain()`), the plan nodes those rules created will have the same node ID and thus the same dataset ID. This will confuse the execution engine and generate undefined results.

This PR addresses it by introducing a new `CreateUniquePhysicalNode` method that can read the next plan node ID from the context and apply it.

### DIFF
--- a/execute/profiler.go
+++ b/execute/profiler.go
@@ -402,21 +402,25 @@ func (s *QueryProfiler) getTableBuilder(q flux.Query, alloc *memory.Allocator) (
 		stats.TotalAllocated,
 		strings.Join(stats.RuntimeErrors, "\n"),
 	}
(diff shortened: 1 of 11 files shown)


rockstar (approved):
Thanks for the description here. I had _assumed_ something similar, but I'm glad to see it explicit. Also, this patch looks much more narrow than the previous one, and can fit in my brain a lot better. Thanks for doing that for me. Let's go ahead and merge this here good work.

PR #3302, opened 2020-11-03
feat(interval): adds new interval package for consistent window handling

nathanielc:
The interval package is intended to replace the execute.Window and
execute.Bounds types creating a consistent and well encapsulated
implementation of the windowing behavior of Flux with calendar
durations.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/interval/interval_test.flux
```diff
@@ -0,0 +1,76 @@
...
+package main
+
+
+import "experimental"
+import "interval"
+
+experimental.addDuration(d: 1d, to: 2020-01-01T00:00:00Z) == 2020-01-02T00:00:00Z or die(msg: "day addition")
+experimental.addDuration(d: 1mo, to: 2020-01-01T00:00:00Z) == 2020-02-01T00:00:00Z or die(msg: "month addition")
+experimental.addDuration(d: 1mo, to: 2020-01-31T00:00:00Z) == 2020-02-29T00:00:00Z or die(msg: "month addition end of month")
+experimental.addDuration(d: 1mo, to: 2020-02-28T00:00:00Z) == 2020-03-28T00:00:00Z or die(msg: "month addition ??")
```
fchikwekwe, line 40: Is this just to confirm current behavior, or is it meant to emulate expected behavior? Per our conversation, I would expect the answer to be `2020-03-28T00:00:00Z` here.

values/time.go
```diff
@@ -216,6 +216,11 @@ func (d Duration) NanoOnly() bool {
 	return d.Months() == 0 && d.Nanoseconds() != 0
 }
 
+// IsMixed returns true if this duration has a combination of month and nanosecond components.
+func (d Duration) IsMixed() bool {
```
fchikwekwe, line 220: Yep, this was missing.

interval/bounds.go
```diff
@@ -0,0 +1,54 @@
...
+func (b Bounds) Start() values.Time {
+	return b.start
+}
+
+func (b Bounds) Stop() values.Time {
+	return b.stop
+}
+
+func (b Bounds) IsEmpty() bool {
+	return b.start >= b.stop
```
fchikwekwe, line 30: Nice simplification.

interval/window.go
```diff
@@ -0,0 +1,310 @@
...
+
+import (
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/values"
+)
+
+// TODO(nathanielc): Make the epoch a parameter to the window
+// See https://github.com/influxdata/flux/issues/2093
+const epoch = values.Time(0)
```
fchikwekwe, line 11: It seems like calling this concept "epoch" could be a little confusing, since its variable and not the standard epoch that's used in programming.

interval/window.go
```diff
@@ -0,0 +1,310 @@
...
+func (w Window) isValid() error {
+	if w.every.IsZero() {
+		return errors.New(codes.Invalid, "duration used as an interval cannot be zero")
+	}
+	if w.every.IsMixed() {
+		const docURL = "https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/window/#calendar-months-and-years"
+		return errors.New(codes.Invalid, "duration used as an interval cannot mix month and nanosecond units").
+			WithDocURL(docURL)
+	}
+	// TODO(nathanielc): what about negative every is that allowed?
```
fchikwekwe, line 61: Are you asking why its _not_ allowed? Good question.

values/time_test.go
```diff
@@ -123,7 +128,7 @@ func TestTime_Add(t *testing.T) {
 		{
 			t:    "2019-02-28T00:00:00Z",
 			d:    "2mo",
-			want: "2019-04-28T00:00:00Z",
+			want: "2019-04-30T00:00:00Z",
```
fchikwekwe, line 131: yay!

fchikwekwe (approved):

interval/window.go
```diff
@@ -0,0 +1,310 @@
...
+			b = next
+			next = w.NextBounds(next)
+		}
+	}
+	return b
+}
+
+// GetOverlappingBounds returns a slice of bounds that overlaps the input bounds.
+// The returned set of bounds are ordered by decreasing time.
+func (w Window) GetOverlappingBounds(start, stop values.Time) []Bounds {
```
jsternberg, line 115: I think if we're rethinking the API for this I'd like to avoid this function. It's a very convenient function that generally leads a developer working on the standard library down a path that they will regret.

interval/window.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	return Bounds{
+		start: start,
+		stop:  stop,
+		index: index,
+	}
+}
+
+// lastIndex will compute the index of the last bounds to contain t
+func (w Window) lastIndex(t values.Time) int {
```
jsternberg, line 170: It would be helpful if this method contained some kind of caching mechanism for reducing the search space based on repeated queries into the window. Generally, the window will be asked for the first or last boundary for points that are all fairly close to each other. Unfortunately, the most complex calculation is usually the first one as you are trying to index into a value that is, by its nature, potentially `O(n)`. We can speed that up through math and indexing certain values, but I think it will still fundamentally be slow in certain situations.

From my experience, this function will be called at least once for every point. Each time I've written these functions, I've thought about using the previous bounds and just retrieving the next boundary, but there's no guarantee that it won't degenerate into an `O(n)` scan for sparse data.

I'm just spitballing at the moment with this.

interval/bounds.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	MaxTime = math.MaxInt64
+	MinTime = math.MinInt64
+)
+
+type Bounds struct {
+	start values.Time
+	stop  values.Time
+	index int
```
jsternberg, line 18: How is the index being used and can we document its purpose?

interval/window.go
jsternberg: My experience with this is that negative numbers as part of the window gets confusing very quickly. I think it can be ok for a negative period, but a negative every means you are getting the windows in the opposite direction and your incoming data needs to be processed differently. If every is negative, then I think the incoming data needs to be in descending order for certain algorithms to work. These are very easy to overlook and create correctness bugs.

This may be wrong. The windowing is already complicated enough that I'm just not sure what the answer to this is.

jsternberg (commented):
Some initial comments. I like the direction we are heading and I think retrieving the latest bounds instead of the earliest is likely to yield dividends on usability that I hadn't thought of. I think I'd like to see a simplification of the `Bounds` struct and maybe move some of that logic into the window lookup. I'm envisioning the `Window` struct as very similar to the `Location` struct in the `time` package. Something similar to this: https://github.com/golang/go/blob/go1.15.4/src/time/zoneinfo.go#L40.

nathanielc:
Revert end of month duration behavior

interval/bounds.go
fchikwekwe: @jsternberg does this comment help clarify `index` for you?

interval/window_test.go
```diff
@@ -0,0 +1,1044 @@
...
+				{Start: mustTime("2121-12-31T00:00:00Z"), Stop: mustTime("2122-01-31T00:00:00Z")},
+				{Start: mustTime("2122-01-31T00:00:00Z"), Stop: mustTime("2122-02-28T00:00:00Z")},
+				{Start: mustTime("2122-02-28T00:00:00Z"), Stop: mustTime("2122-03-31T00:00:00Z")},
+				{Start: mustTime("2122-03-31T00:00:00Z"), Stop: mustTime("2122-04-30T00:00:00Z")},
+				{Start: mustTime("2122-04-30T00:00:00Z"), Stop: mustTime("2122-05-31T00:00:00Z")},
+				{Start: mustTime("2122-05-31T00:00:00Z"), Stop: mustTime("2122-06-30T00:00:00Z")},
+				{Start: mustTime("2122-06-30T00:00:00Z"), Stop: mustTime("2122-07-31T00:00:00Z")},
+				{Start: mustTime("2122-07-31T00:00:00Z"), Stop: mustTime("2122-08-31T00:00:00Z")},
+				{Start: mustTime("2122-08-31T00:00:00Z"), Stop: mustTime("2122-09-30T00:00:00Z")},
+			},
```
fchikwekwe, line 864: @nathanielc I added this test case....

interval/window_test.go
```diff
@@ -0,0 +1,1044 @@
...
+				{Start: mustTime("2100-07-01T00:00:00Z"), Stop: mustTime("2100-08-01T00:00:00Z")},
+				{Start: mustTime("2100-06-01T00:00:00Z"), Stop: mustTime("2100-07-01T00:00:00Z")},
+				{Start: mustTime("2100-05-01T00:00:00Z"), Stop: mustTime("2100-06-01T00:00:00Z")},
+				{Start: mustTime("2100-04-01T00:00:00Z"), Stop: mustTime("2100-05-01T00:00:00Z")},
+				{Start: mustTime("2100-03-01T00:00:00Z"), Stop: mustTime("2100-04-01T00:00:00Z")},
+				{Start: mustTime("2100-02-01T00:00:00Z"), Stop: mustTime("2100-03-01T00:00:00Z")},
+				{Start: mustTime("2100-01-01T00:00:00Z"), Stop: mustTime("2100-02-01T00:00:00Z")},
+				{Start: mustTime("2099-12-01T00:00:00Z"), Stop: mustTime("2100-01-01T00:00:00Z")},
+			},
+		},
```
fchikwekwe, line 990: @nathanielc ...and this test case to satisfy the request for window cases further away from the epoch. Is this what you were thinking of?

interval/window.go
fchikwekwe: @jsternberg I've moved the todo to the Window struct. I agree that it makes more sense there. As far as allowing epoch to be more customizable, I think we might be able to wait on those changes until we actually start working on the issue to add epoch to windows fully

fchikwekwe:
I added my changes to the latest commit on this PR, but the PR that will be merged is https://github.com/influxdata/flux/pull/3440.

nathanielc:
This was completed in a follow up PR

PR #3304, opened 2020-11-04
refactor: replace builtin package with fluxinit/static

adrian-thurston:
A static initialization is not desirable in the main binaries, as it forces all
paths of code to init, but it is still useful in tests. It allows static
intialization to be performed once for all tests and eliminates the need to
always add the FluxInit call. Added a fluxinit/static package that calls
fluxinit.FluxInit() to replace the builtin package. This hides the nature of
the initialization and makes it clear that it is mandatory initialization code.


builtin/builtin.go
```diff
@@ -1,13 +0,0 @@
-// Package builtin contains all packages related to Flux built-ins are imported and initialized.
-// This should only be imported from main or test packages.
-// It is a mistake to import it from any other package.
```
nathanielc, line 3: Could you add a similar comment to the new `static` package?

nathanielc (approved):
LGTM with a few clarifying comments.

The FluxInit function references the `builtin` package which has now been removed. We should update that comment. 

Also adding a package comment to the `static` package would be helpful.

nathanielc:
Also I expect you will have to create equivalent `static` packages in OSS and IDPE?

builtin/builtin.go
adrian-thurston: Oh yes thanks for catching that.

adrian-thurston:
> Also I expect you will have to create equivalent `static` packages in OSS and IDPE?

Yup, will need that too.

PR #3307, opened 2020-11-05
fix(types): format types using letters instead of numbers

### LINKED ISSUE #3305: Type errors use numeric type variables that are difficult to understand
When you get a type error in Flux, the error message reports the internal type variable numbers in the error message. These numbers are large and distract from what is really trying to be communicated.

For example:
```
my_mean = (tables=<-) => tables
    |> map(fn: (r) => ({ r with meaned: ((tables |> mean() |> first())._value) }))
```
```
 type error @2:63-2:70: expected {_value:t8129 | t8141} but found [t8143]
```

I propose the error message should be like this:

```
 type error @2:63-2:70: expected {_value:A | B} but found [C] 
```

This highlights better the difference between the type, i.e. we expected a record, but instead got a list.

nathanielc:
Previously type errors would always print out the type number of their
corresponding type variable. This was confusing as often those numbers
where in the thousands and didn't really communicate anything other
than if they were different from other type variables in the same
message.

Now type variables are made fresh before reporting the error message
such that they are low numbers and converted to the first ten letters of
the alphabet for display.

Fixes #3305 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/src/core/semantic/types.rs
```diff
@@ -170,23 +170,41 @@ pub enum Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut frsh = Fresher::from(0);
```
jlapacik, line 173: nit: might as well call this `fresh`

jlapacik (approved):
Much more user friendly. Thanks.

PR #3311, opened 2020-11-06
test: added a test case for group-aggregate

adrian-thurston:
Added a test case for group-aggregage where tag key sets are uneven in length
among the input rows that are aggregate/selected in the output grouping.


fchikwekwe (approved):
Thanks for adding this coverage!

PR #3314, opened 2020-11-09
chore(formatter): add a benchmark for the formatter

nathanielc:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/formatter/mod.rs
```diff
@@ -22,8 +22,8 @@ pub fn convert_to_string(file: &File) -> Result<String, String> {
     formatter.output()
 }
 
-pub fn format(contents: String) -> Result<String, String> {
-    let file = parse_string("", contents.as_str());
+pub fn format(contents: &str) -> Result<String, String> {
```
rockstar, line 25: Ah! This is an interesting case that lint couldn't catch.

rockstar (approved):

PR #3315, opened 2020-11-09
feat(formatter): use a tabwriter to vertically align tabstops

nathanielc:
This change adds the ability to vertically align records that appear in
an array. It uses an implementation of the tabwriter from BurntSushi
that implements and _elastic tabstops_ algorithm. Basically it ensures
that tabstops vertically align instead of treating them as fixed width
columns. See http://nickgravgaard.com/elastic-tabstops/, this is the
same method Go uses for its formatter.

I also bench marked this change to the formatter as I would expect it to
slow it down considerably. A benchmark on formatting the everything.flux
file run 14% slower with this change. That seems tolerable for this kind
of change but am open to suggestions to improve it.

This PR is dependent on #3314  to add the benchmark, do NOT merge until then.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/src/core/formatter/mod.rs
+++ b/libflux/src/core/formatter/mod.rs
@@ -3,6 +3,12 @@ use crate::ast::{self, walk::Node, File};
 use crate::parser::parse_string;
 use crate::Error;
 
+use std::io;
+use std::io::Write;
+use std::string::FromUtf8Error;
+
+use tabwriter::{Alignment, IntoInnerError, TabWriter};
+
 use chrono::SecondsFormat;
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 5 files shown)


libflux/src/core/formatter/mod.rs
```diff
@@ -41,28 +50,76 @@ pub struct Formatter {
...
-        if let Some(err) = &self.err {
-            return Err(err.msg.clone());
+    pub fn output(mut self) -> Result<String, Error> {
+        if let Some(err) = self.err {
+            return Err(err);
         }
 
-        Ok(self.builder.clone())
+        self.builder.flush()?;
+        Ok(String::from_utf8(self.builder.into_inner()?)?)
```
rockstar, line 116: While this is a Correct Rust Idiom™, the use of `into_inner` to convert from type to type has always been a little awkward to me. Even acknowledging that rust does it "right" with requiring implementations of `From`, it still feels duck-typey in a way that feels wrong in rust.

Nothing actionable here. I'm just talking out loud.

rockstar (approved):
Was this the driver behind the addition of the formatter benchmark? If so, did we see many changes here?

libflux/src/core/formatter/mod.rs
nathanielc: yeah, this is the _only_ way to do it according to the TabWriter API, but yeah its a bit round about.

nathanielc:
@rockstar Yes, the benchmark was to understand the impact of using the tabwriter. This changes introduces a 14% performance slow down. So its definitely a hit, but it seems worth it to me given the complexity of rereading and writing whitespace to align vertical tabs.

PR #3317, opened 2020-11-10
docs: dictionary type

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -520,6 +520,12 @@ If type inference determines all the properties on a record it is said to be bou
 Not all keys may be known on the type of a record in which case the record is said to be unbounded.
 An unbounded record may contain any property in addition to the properties it is known to contain.
 
+##### Dictionary types
+
+A _dictionary type_ is a collection that associates keys to values.
+Keys must be comparable and of the same type.
+Values must also be of the same type.
+
 ##### Function types
 
 A _function type_ represents a set of all functions with the same argument and result types.
@@ -761,6 +767,7 @@ Literals construct a value.
             | pipe_receive_lit
             | RecordLiteral
             | ArrayLiteral
+            | DictLiteral
             | FunctionLiteral .
 
 ##### Record literals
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -788,6 +795,14 @@ Array literals construct a value with the array type.
     ArrayLiteral   = "[" ExpressionList "]" .
     ExpressionList = [ Expression { "," Expression } ] .
 
+##### Dictionary literals
+
+Dictionary literals construct a value with the dict type.
+
+    DictLiteral     = "[" AssociativeList "]" .
+    AssociativeList = [ Association { "," AssociativeList } ] .
+    Association     = Expression ":" Expression .
```
nathanielc, line 804: Is this ambiguous with array literals? The  `Association = `Expression ":" Expression`  production rule and this one `ExpressionList = [ Expression { "," Expression } ] .` We would need infinite back tracking to wait for either the `:` or the `,`.

docs/SPEC.md
```diff
@@ -520,6 +520,12 @@ If type inference determines all the properties on a record it is said to be bou
 Not all keys may be known on the type of a record in which case the record is said to be unbounded.
 An unbounded record may contain any property in addition to the properties it is known to contain.
 
+##### Dictionary types
+
+A _dictionary type_ is a collection that associates keys to values.
+Keys must be comparable and of the same type.
+Values must also be of the same type.
+
```
nathanielc, line 528: Should we explicitly mention here that dictionary keys can be dynamic, i.e. any expression?

docs/SPEC.md
jlapacik: There's no ambiguity except in the case of an empty array/dict `[]`. However you are correct that it would require infinite backtracking in order to determine the type. I believe our parser is LL(1) so this syntax might not work (assuming we keep keys as arbitrary expressions)

docs/SPEC.md
jlapacik: Yes I mentioned that in the dictionary literal section

docs/SPEC.md
```diff
@@ -788,6 +795,23 @@ Array literals construct a value with the array type.
...
+    Association     = Expression ":" Expression .
+
+The keys can be arbitrary expressions.
+The type system will enforce that all keys are of the same type.
+
+Examples:
+
+    a = "a"
+    c = | a: 1, "b": 2 | // dictionary mapping string values to integers
+    d = | a: 1, 2: 3 | // type error: cannot mix string and integer keys
```
nathanielc, line 813: I like this new syntax, its minimal and clear enough that we have some kind of data type being defined.

nathanielc (approved):

PR #3341, opened 2020-11-18
feat(values): create a Dictionary interface and implementation

### LINKED ISSUE #3318: Add dictionary type to values package
In order to support the new flux dictionary type, we need to add a dictionary type to the `values` package. This dictionary type should be defined as follows
```Go
type Dictionary interface {
    Get(key Value, def Value) Value
    Insert(key Value, v Value) Dictionary
    Remove(key Value) Dictionary
}
```

The underlying implementation should use Ben Johnson's [sorted map](https://github.com/benbjohnson/immutable#sorted-map).
```Go
type dict struct {
    btree *immutable.SortedMap
}

func (d *dict) Get(key Value, def Value) Value {
    v, ok := d.btree.Get(key)
    if !ok {
        return def
    }
    return v.(Value)
}
func (d *dict) Insert(key Value, v Value) Dictionary {
    return &dict{
        btree: d.btree.Set(key, v),
    }
}
func (d *dict) Remove(key Value) Dictionary {
    return &dict{
        btree: d.btree.Delete(key),
    }
}

type intComparer struct {
}

func (intComparer) Compare(a, b interface{}) int {
    if a.(Value).Int() < b.(Value).Int() {
        return -1
    }
    if a.(Value).Int() > b.(Value).Int() {
        return 1
    }
    return 0
}

func NewDict(ty semantic.Nature) Dictionary {
    switch ty {
    case semantic.Int:
        return &dict{immutable.NewSortedMap(intComparer{})}
    ...
    default:
        panic("invalid type")
    }
}
```

We will need to implement comparison functions for the comparable types `int`, `uint`, `float`, `string`, `duration`, and `time`.

Note the `Dictionary` interface itself does not guarantee that all key values are of the same type, nor does it guarantee that all associated values are of the same type. It is assumed that type inference makes this guarantee and therefore the contract proposed by this interface can be a little lax. The simplicity we gain from only managing a single dictionary implementation outweighs the negligible reduction in overhead we would get with more specialized implementations and interfaces.

Also note that this will also require that we update the `Value` interface to include a new `Dict()` method
```Go
type Value interface {
    ...
    Dict() Dictionary
}
```

and that we add this method to all other `Value` implementations.

DOD:
* Function that instantiates a dictionary from a list of records of the form `{key: ..., value: ...}`
* Unit tests in the `values` package

jsternberg:
This adds a Dictionary interface and implementation to the `values`
package. It also updates all of the `values` and usages to acknowledge
that dictionaries exist.

BREAKING CHANGE: The `Value` interface has changed and will require changes to
existing implementations of the interface.

Fixes #3318.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/values/dict.go
+++ b/values/dict.go
@@ -1,0 +1,299 @@
+package values
+
+import (
+	"regexp"
+
+	"github.com/benbjohnson/immutable"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/semantic"
+)
+
+// Dictionary defines the interface for a dictionary.
+//
+// A Dictionary is immutable. Changes to a Dictionary
+// will return a new Dictionary.
+type Dictionary interface {
+	Value
+
+	// Get will retrieve a Value out of the Dictionary.
+	// If the key is not present, the def Value will
+	// be returned instead.
+	Get(key, def Value) Value
+
+	// Insert will insert a Value into the Dictionary
+	// using the key and value. It will return a new
+	// Dictionary with the key/value inserted. If the
+	// key was already in the Dictionary, it will
+	// be replaced.
+	//
+	// Attempting to insert a null value for the key
+	// will return an error.
+	Insert(key, value Value) (Dictionary, error)
+
+	// Remove will remove the key/value pair that
+	// matches with the key. It will return a new
+	// Dictionary with the key/value removed.
+	Remove(key Value) Dictionary
+
+	// Range will iterate over each element in
+	// the Dictionary.
+	Range(func(key, value Value))
+
+	// Len returns the number of elements inside of
+	// Dictionary.
+	Len() int
+}
+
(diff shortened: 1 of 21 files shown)


values/dict.go
```diff
@@ -0,0 +1,312 @@
...
+	})
+	return equal
+}
+
+type (
+	intComparer      struct{}
+	uintComparer     struct{}
+	floatComparer    struct{}
+	stringComparer   struct{}
+	durationComparer struct{}
```
jlapacik, line 167: I thought we discovered that durations aren't comparable?

jsternberg: I think I had already added it since it is possible to compare them. The sorting won't necessarily be meaningful though. I'll remove this section so there isn't unused code.

wolffcm (approved):
Very nice.

jlapacik (approved):

PR #3343, opened 2020-11-19
feat(stdlib): add dict package for interacting with dictionaries

### LINKED ISSUE #3319: Add dict package to stdlib
The dict package should be defined as follows
```
package dict

builtin fromList : (pairs: [{ key: K, value: V }]) => [K: V] where K: Comparable
builtin get : (key: K, from: [K: V], default: V) => V where K: Comparable
builtin insert : (key: K, value: V, dict: [K: V]) => [K: V] where K: Comparable
builtin remove : (key: K, dict: [K: V]) => [K: V] where K: Comparable
```

The implementations should use the dictionary values defined in https://github.com/influxdata/flux/issues/3318. For example, `fromList` should be defined as follows
```Go
func (ctx context.Context, args values.Object) (values.Value, error) {
    pairs, err := args.GetRequiredArray("pairs")
    if err != nil {
        return nil, err
    }

    record, err := pairs.Type().ElemType()
    if err != nil {
        return nil, err
    }

    var keyType semantic.Nature

    for i := 0; i < 2; i++ {
        prop, err := record.RecordProperty(i)
        if err != nil {
            return nil, err
        }
        if prop.Name() != "key" {
            continue
        }
        ty, err := prop.TypeOf()
        if err != nil {
            return nil, err
        }
        keyType = ty.Nature()
    }

    return values.DictFromList(keyType, pairs), nil
}
```

DOD:
* Unit tests in the new `dict` package for the builtin implementations `fromList`, `get`, `insert`, `remove`
* Flux acceptance tests in the dict package showing usage of the dict type

jsternberg:
This adds a dict package for interacting with and creating dictionaries.

Fixes #3319.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/dict/dict.flux
```diff
@@ -0,0 +1,19 @@
+package dict
+
+// fromList will convert an array of key/value pairs
+// into a dictionary.
+builtin fromList : (pairs: [{ key: K, value: V }]) => [K: V] where K: Comparable
+
+// get will retrieve a value from a dictionary. If there is no
+// key in the dictionary, the default value will be returned.
+builtin get : (from: [K: V], key: K, default: V) => V where K: Comparable
```
nathanielc, line 9: I commented on the original design issue but I think being consistent with the name of the `dict` type argument  is important. 

I recommend we rename `from` to `dict` here.

jlapacik: I forgot to update both issues, but I'm also fine if we change this to `dict`.

jlapacik (approved):
Looks great! I might also add or update the flux acceptance test with an `insert` and `remove` call.

jsternberg:
@jlapacik do you have any ideas for how to meaningfully test insert and remove? I am finding it difficult to think of a way to test it in a meaningful way without just doing the same `dict.get` test.

jlapacik:
@jsternberg I would just insert and remove an element or two and then write a test function using `map` that asserts which key/value pairs exist in the dict.

jsternberg:
Ok done.

jlapacik (approved):
Awesome!

PR #3344, opened 2020-11-19
test: benchmark rust scanner

### LINKED ISSUE #3267: porting the scanner to rust
Some preliminary work to port the scanner to rust has been completed. Tests are passing. 

https://github.com/influxdata/flux/commit/9b0eeed0a1cc26e47721d547b9143b46b00b2801

Remaining work includes:

1. Investagate which dependencies can be dropped as a result of moving to pure rust. We may be drawing in additional dependencies for compiling to wasm.

2. Do a performance evaluation to determine if we are slowing down parsing with this move. Using cargo bench to capture the performance characteristics. In this particular case we would be looking for the regression to be no worse than 1.5x. If it is worse, bring it up with the team.

3. Upgrade build dependencies, CI, Makefiles, etc to depend on ragel 7.

fchikwekwe:
Closes #3267. 

The main changes is in `benchmarks.rs`. The other changes are mostly housekeeping. 

This issue measures performance changes after porting the scanner to rust. I added a few benchmarks and tested against some existing ones. We were just looking to not regress performance, but there are some modest gains as seen below. 

|test name|generating "C" code |rust scanner |performance change|
|--|--|-----|---|
|scan_short_text|60.383 ns|44.630 ns| -26.5% | 
|scan_long_text|45.097 ns|43.696 ns| -3.1% |
|replaceAll|456.28 us|383.14 us| -16.0% |
|title_test|442.05 us|363.45 us| -17.8% |
|trim_test|390.74 us|373.56 us| -4.4% |
|toUpper|377.92 us|374.77 us| -0.8% |
|substring|447.97 us|421.40 us| -5.9% |
|toLower|383.53 us|378.75 us| -1.2% |
|replace_test|454.77 us|399.19 us | -12.2% |
|length_test|505.50 us| 382.15 us| -24.4% |
|subset_test|445.94 us| 389.79 us | -12.6% |


Dockerfile_build
```diff
@@ -16,8 +16,11 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 # Install ragel
-ENV RAGEL_VERSION=6.10
-RUN curl http://www.colm.net/files/ragel/ragel-${RAGEL_VERSION}.tar.gz -O && \
+ENV RAGEL_VERSION=7
```
fchikwekwe, line 19: @adrian-thurston do I need to specify exactly which version of ragel we wanna build with here? I'm also not sure if there are any other spots where I need to update the ragel version. I looked at the CI config file and it didnt specify a version of ragel there.

Dockerfile_build
```diff
@@ -15,7 +15,7 @@ RUN apt-get update && \
     autoconf automake autotools-dev libtool xutils-dev valgrind && \
     rm -rf /var/lib/apt/lists/*
 
-# Download, verify, and install ragel version 6.
+# Download, verify, and install ragel version 6. 
 ENV RAGEL_VERSION=6.10
```
scbrickley, line 19: Should this be changed to reference Ragel 7?

fchikwekwe: Both versions are still in the Dockerfile since we're flagging ragel 7 and using 6 by default. I think this change came from a linter or a formatter or something.

fchikwekwe: We're flagging the new ragel because we're flagging the scanner (I believe).

fchikwekwe: I imagine we may remove these lines from the dockerfile once the new scanner is vetted in prod.

fchikwekwe: and the flag removed of course.

scbrickley (approved):
LGTM!

PR #3350, opened 2020-11-23
feat(stdlib): add package to embed flux files in a package

### LINKED ISSUE #3334: Establish a way to ship the flux acceptance tests, executable from downstream consumers

DOD:

jsternberg:
The flux standard library can be retrieved from Go by using the `embed`
package to lookup the text of a file.

This method is a lot faster since it just compresses the files and
embeds it as byte slices inside of the Go code.

Fixes #3334.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
We won't be doing this at the moment. While this is a good example of doing this, it is not necessary at the moment.

PR #3358, opened 2020-11-24
feat: dictionary type unification rule

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/src/core/parser/mod.rs
```diff
@@ -559,7 +559,29 @@ impl Parser {
...
+                        let val = self.parse_monotype();
+                        let end = self.close(TOK_RBRACK);
+                        MonoType::Dict(Box::new(DictType {
+                            base: self.base_node_from_tokens(&start, &end),
+                            key: ty,
+                            val: val,
+                        }))
+                    }
+                }
+            }
```
jlapacik, line 584: This is the block of code that makes it so that we can parse dict types in builtin statements.

libflux/src/core/semantic/types.rs
```diff
@@ -663,6 +671,57 @@ impl Array {
...
+
+impl Dictionary {
+    fn unify(
+        self,
+        actual: Self,
+        cons: &mut TvarKinds,
+        f: &mut Fresher,
+    ) -> Result<Substitution, Error> {
+        let sub = self.key.unify(actual.key, cons, f)?;
+        apply_then_unify(self.val, actual.val, sub, cons, f)
```
jlapacik, line 712: These 2 lines are the crux of the PR. The rest is either boilerplate or generated code.

nathanielc (commented):
LGTM the tests around type unification look good. 

Could you add tests for the formatter/mod.rs changes?

libflux/src/core/parser/mod.rs
```diff
@@ -559,7 +559,29 @@ impl Parser {
         // Tvar | Basic | Array | Record | Function
```
scbrickley, line 559: Small nitpick, but it may be good to update the comment here?

libflux/src/core/semantic/types.rs
```diff
@@ -663,6 +671,57 @@ impl Array {
...
+}
+
+impl MaxTvar for Dictionary {
+    fn max_tvar(&self) -> Tvar {
+        vec![self.key.max_tvar(), self.val.max_tvar()].max_tvar()
+    }
+}
+
+impl Dictionary {
+    fn unify(
```
scbrickley, line 705: I'm unclear on what `unify`ing means in this context. Could you elaborate?

scbrickley (commented):
Looks good to me. I left a few comments/questions, but nothing that needs to hold up the PR outside of what @nathanielc suggested.

libflux/src/core/semantic/types.rs
jlapacik: In the type system, unification is the process of resolving equality constraints, the result of which is a substitution which maps type variables to the monotypes they represent. For example, unification of two primitive types only succeeds (producing an empty substitution) if both types are the same
```
int = int
string = string
int != string
```

Dictionary types are composite types, so two dictionary types unify if and only if both of their components unify
```
[k0:v0] = [k1:v1] iff k0 = k1 and v0 = v1
```

libflux/src/core/ast/flatbuffers/ast_generated.rs
```diff
@@ -1203,6 +1216,269 @@ pub mod fbast {
...
+            }
+        }
+    }
+
+    impl<'a> DictType<'a> {
+        #[inline]
+        pub fn init_from_table(table: flatbuffers::Table<'a>) -> Self {
+            DictType { _tab: table }
+        }
+        #[allow(unused_mut)]
```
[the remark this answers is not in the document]
jlapacik: This code is generated by `flatc`. My guess is that in certain cases this generated variable is unused and because this is rust, it wouldn't compile without this annotation.

libflux/src/core/semantic/types.rs
scbrickley: Awesome. That makes sense. Thanks for the clarification :+1:

nathanielc (approved):

PR #3359, opened 2020-11-24
feat(tests): add a test harness that can read flux files from various sources

### LINKED ISSUE #3334: Establish a way to ship the flux acceptance tests, executable from downstream consumers

jsternberg:
The test harness can be implemented to use different ways of interacting
with a flux provider. Each of these test harnesses can use the same
scaffolding to read tests directly from the filesystem or to read them
from an archive on the filesystem.

Fixes #3334.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/dependencies/tests/tests.go
+++ b/dependencies/tests/tests.go
@@ -1,0 +1,136 @@
+package tests
+
+import (
+	"archive/tar"
+	"archive/zip"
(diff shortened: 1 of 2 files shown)


jsternberg:
Replaced by #3398.

PR #3370, opened 2020-12-01
fix: Fix typos in Dockerfile_build

pierwill:


adrian-thurston (approved):
Thanks for fixing these. I'm getting much too careless lately. Definitely need to improve.

PR #3375, opened 2020-12-04
fix(stdlib/strings): substring method now works on more indices

### LINKED ISSUE #3368: InfluxDB 2.0 OSS: Flux strings.subString() is buggy
Using Flux on InfluxDB OSS Version 2.0.0 (37cc047) strings.substring() does not return the correct substrings.

Taken the example from the [documentation](https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/strings/substring/) a call to `strings.substring(v: "influx", start: 0, end: 3)` should return "infl" (4 characters). That would make sense. In reality it returns "inf" (3 characters).

# Steps to reproduce
```
import "strings"

from(bucket: "olmero")
    |> range(start: -5m)
    |> filter(fn: (r) => 
      (r._measurement == "cpu")
      and (r._field == "usage_idle")
      and (r.cpu == "cpu-total")
    )
    |> group()
    |> limit(n: 1)
    |> keep(columns: ["_value"])
    |> map(fn: (r) => ({
      r with
        aaa: strings.substring(v: "influx", start: 0, end: 0),
        bbb: strings.substring(v: "influx", start: 0, end: 1),
        ccc: strings.substring(v: "influx", start: 0, end: 2),
        ddd: strings.substring(v: "influx", start: 0, end: 3)
    }))
```

# Real result
| table | _value | aaa | bbb | ccc | ddd |
| --- | --- | --- | --- | --- | --- |
| 0 | 98.34475840142755 | i | i | in | inf |

# Expected result
| table | _value | aaa | bbb | ccc | ddd |
| --- | --- | --- | --- | --- | --- |
| 0 | 98.34475840142755 | i | in | inf | infl |

# Further trouble
Furthermore 

`strings.substring(v: "influx", start: 0, end: 13)` 

returns an error (indices out of bounds). That makes using subString in queries on real fields (where values can vary) complex. You need to call 

`strings.substring(v: "influx", start: 0, end: if strings.strlen(v: "influx") < 13 then strings.strlen(v: "influx") else 13)`

to use it in queries (and avoid errors). Would make way more sense if `subStr()` would simply return "influx" in this case.

jsternberg:
The substring method in the strings package now works properly when the
start or end index isn't within the bounds of the existing string. It
truncates these indices to be the start or end of the string. This
reduces the need to check the length of the string to get a substring of
a certain size and makes it work more similar to how the javascript
substring operates.

This also fixes the case where both the start and the end index are
zero. This was erroneously creating a string of 1 character when it
should have been zero.

Fixes #3368.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/strings/strings.go
+++ b/stdlib/strings/strings.go
@@ -7,6 +7,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/runtime"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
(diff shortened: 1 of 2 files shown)


scbrickley (approved):
LGTM - The test cases made this really easy to review

PR #3377, opened 2020-12-04
fix(lang): retrieve now option before running the flux script

jsternberg:
This changes the `AstProgram` to retrieve the now option before it runs
the flux script. The reason is because the now option is placed into the
scope before the script is executed, but it can be reassigned to a
different value during execution. When we tried to retrieve the option
afterwards, it resulted in selecting the local now instead of the one
from the universe package.

We now retrieve the option ahead of time. If the option is redefined
with `option`, then the retrieved option will be modified in a mutable
way that we can access anyway and it avoids accidentally using a similar
variable with the same name.

Fixes influxdata/influxdb#18883.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -428,7 +428,24 @@ func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flu
 
 	s, cctx := opentracing.StartSpanFromContext(ctx, "eval")
 
-	sideEffects, scope, err := p.Runtime.Eval(cctx, ast, &ExecOptsConfig{}, flux.SetNowOption(p.Now))
+	// Set the now option to our own default and capture the option itself
+	// to allow us to find it after the run. A user might overwrite the
+	// now parameter with their own thing so we don't want to allow for
+	// that interference. If `option now` is used to overwrite this,
+	// the inner value pointed to by the option will be modified.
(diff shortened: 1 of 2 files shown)


lang/compiler.go
```diff
@@ -428,18 +428,31 @@ func (p *AstProgram) getSpec(ctx context.Context, alloc *memory.Allocator) (*flu
...
 
-	sideEffects, scope, err := p.Runtime.Eval(cctx, ast, &ExecOptsConfig{}, flux.SetNowOption(p.Now))
+	// Set the now option to our own default and capture the option itself
+	// to allow us to find it after the run. A user might overwrite the
+	// now parameter with their own thing so we don't want to allow for
+	// that interference. If `option now` is used to overwrite this,
+	// the inner value pointed to by the option will be modified.
+	// TODO(jsternberg): Personal note, I don't like how now interacts with
+	// the runtime and flux code in so many places. We should evaluate how
+	// now is used and see if we can improve how now interacts with the system.
```
adrian-thurston, line 438: Agree it needs some cleanup. Since the `tableFind` functions need to have now available in the execution dependencies, I think that should become the authority on `now`.

adrian-thurston (approved):

PR #3384, opened 2020-12-07
fix(compiler): do not allow containers within a record to be null

### LINKED ISSUE #3374: panic calling map() with non-existent column: got "invalid" expected "array
A panic happens when calling `map()` with non-existent column:
> panic: unexpected kind: got "invalid" expected "array"

Flux: v0.96.1

```
data = from(bucket: "_tasks")
  |> range(start: -1m)
  |> filter(fn: (r) => (r["_measurement"] == "runs"))
  |> filter(fn: (r) => (r["_field"] == "logs"))
  |> filter(fn: (r) => (r["status"] == "success"))
  |> last()
  |> map(fn: (r) => ({ r with newColumn: r.abc[0] }))
  |> yield()
```
flux-panic

Note that the panic doesn't happen with `abc[0]`. In that case the error is:
> undefined identifier abc

jsternberg:
When we retrieve a missing parameter from a record, we should not allow
it to return a null value if we know we are supposed to be returning one
of the container types (array, record, or dict). Those three are not
nullable so it is not valid for null to be returned. Instead, an error
is returned instead.

Fixes #3374.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -475,6 +475,52 @@ func TestCompileAndEval(t *testing.T) {
 			}),
 			want: values.NewBool(true),
 		},
+		{
+			name: "null array",
+			fn:   `(r) => r.a[0]`,
+			inType: semantic.NewObjectType([]semantic.PropertyType{
+				{Key: []byte("r"), Value: semantic.NewObjectType([]semantic.PropertyType{
+					{Key: []byte("a"), Value: semantic.NewArrayType(semantic.BasicString)},
+				})},
+			}),
+			input: values.NewObjectWithValues(map[string]values.Value{
+				"r": values.NewObjectWithValues(nil),
+			}),
+			wantEvalErr: true,
+		},
+		{
+			name: "null record",
+			fn:   `(r) => r.a["b"]`,
+			inType: semantic.NewObjectType([]semantic.PropertyType{
+				{Key: []byte("r"), Value: semantic.NewObjectType([]semantic.PropertyType{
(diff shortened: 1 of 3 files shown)


jsternberg:
It should be noted with this PR that Go doesn't have access to any of the constraint parameters from Rust. If it had access to that and there was a constraint on the value to be a basic type, we could catch this in the compile phase instead of eval. Until we do, this is the second best place in my opinion.

jlapacik:
I guess this is another example of why the compiler package should have access to type constraints.

compiler/compiler_test.go
```diff
@@ -475,6 +475,52 @@ func TestCompileAndEval(t *testing.T) {
...
+			}),
+			input: values.NewObjectWithValues(map[string]values.Value{
+				"r": values.NewObjectWithValues(nil),
+			}),
+			wantEvalErr: true,
+		},
+		// TODO(jsternberg): We presently have not implemented dictionary support for
+		// runtime functions. There aren't any builtins that use this functionality,
+		// but when we do, this test will need to be uncommented to ensure that
+		// a null dictionary does not sneak in.
```
rockstar, line 509: I'm usually hesitant to leave commented code in, but this is quite helpful. Do we have a timeframe from when we'll support this? Is there an issue we can add a comment to about uncommenting this?

rockstar (approved):

compiler/compiler_test.go
jsternberg: I'm not sure honestly. I wish I had a better answer than that.

PR #3387, opened 2020-12-09
feat: dictionary literals

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


ast/ast.go
```diff
@@ -1265,6 +1267,43 @@ func (e *ArrayExpression) Copy() Node {
 	return ne
 }
 
+// ArrayExpression is used to create and directly specify the elements of an array object
```
nathanielc, line 1270: nit, update the comment

ast/json.go
```diff
@@ -776,6 +776,76 @@ func (e *ArrayExpression) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+func (e *DictExpression) MarshalJSON() ([]byte, error) {
```
nathanielc, line 779: I am remembering that we define these JSON structures in the swagger doc for InfluxDB. Can you create an issue to update those definitions for these new AST nodes?

libflux/core/src/parser/tests.rs
```diff
@@ -3888,6 +3888,203 @@ fn declare_variable_as_an_empty_array() {
     )
 }
 
+#[test]
+fn parse_empty_dict() {
+    let mut p = Parser::new(r#"[:]"#);
```
nathanielc, line 3893: do we have an existing test that parses an empty array literal?

libflux/core/src/parser/tests.rs
```diff
@@ -3888,6 +3888,203 @@ fn declare_variable_as_an_empty_array() {
...
+                comma: None,
+            }],
+            lbrack: None,
+            rbrack: None,
+        }))
+    )
+}
+
+#[test]
+fn parse_multi_element_dict() {
```
nathanielc, line 3946: Could you add a test were we parse the key as an expression: Something like `[1+1: 2]` or `[a:2]` etc

nathanielc (changes requested):
This is awesome! I made some small suggestions.

libflux/core/src/parser/tests.rs
jlapacik: Yes we do.

ast/json.go
jlapacik: https://github.com/influxdata/influxdb/issues/20299

ast/json.go
```diff
@@ -776,6 +776,76 @@ func (e *ArrayExpression) UnmarshalJSON(data []byte) error {
...
+	raw := struct {
+		Type string `json:"type"`
+		*Alias
+	}{
+		Type:  e.Type(),
+		Alias: (*Alias)(e),
+	}
+	return json.Marshal(raw)
+}
+func (e *DictExpression) UnmarshalJSON(data []byte) error {
```
nathanielc, line 790: Can you also add a JSON unit test?

libflux/core/src/ast/tests.rs
```diff
@@ -1500,6 +1500,56 @@ fn test_json_array_expression() {
...
+                }),
+                comma: None,
+            },
+        ],
+        rbrack: None,
+    }));
+    let serialized = serde_json::to_string(&n).unwrap();
+    assert_eq!(
+        serialized,
+        r#"{"type":"DictExpression","elements":[{"key":{"type":"StringLiteral","value":"a"},"val":{"type":"IntegerLiteral","value":"0"}},{"key":{"type":"StringLiteral","value":"b"},"val":{"type":"IntegerLiteral","value":"1"}},{"key":{"type":"StringLiteral","value":"c"},"val":{"type":"IntegerLiteral","value":"2"}}]}"#
```
nathanielc, line 1548: Can we copy this test case to the Go side as well? Ensure that they both work and interop.

libflux/core/src/formatter/mod.rs
```diff
@@ -652,6 +653,43 @@ impl Formatter {
         self.write_rune(']')
     }
 
+    fn format_dict_expression(&mut self, n: &ast::DictExpr) {
+        let multiline = n.elements.len() > 4 || n.base.is_multiline();
+        self.format_comments(&n.lbrack);
+        self.write_rune('[');
+        if multiline {
+            self.temp_tabstops = true;
```
nathanielc, line 661: Why is temp_tabstops being used here? As I understood it that flag is only used when a comment interrupts a line.

libflux/core/src/formatter/tests.rs
```diff
@@ -138,6 +138,18 @@ a[i]"#,
...
+fn dict() {
+    assert_unchanged(r#"["a": 0, "b": 1]"#);
+    assert_unchanged(
+        r#"[
+    "a": 0,
+    "b": 1,
+    "c": 2,
+]"#,
+    );
+}
```
nathanielc, line 151: Can you add a test case with comments?

nathanielc (changes requested):
Looks great, I gave it a more thorough review this time. A few missing tests but other than that it looks solid.

libflux/core/src/formatter/mod.rs
jlapacik: I didn't gather that when I was reviewing the code. It seems like it's only used when formatting records as far as I could tell. I added it to be consistent with how arrays are formatted.

nathanielc: ah yes I think I have it confused with `temp_indent`. Nevermind this looks good here.

interpreter/interpreter.go
```diff
@@ -515,6 +517,25 @@ func (itrp *Interpreter) doArray(ctx context.Context, a *semantic.ArrayExpressio
 	return values.NewArrayWithBacking(a.TypeOf(), elements), nil
 }
 
+func (itrp *Interpreter) doDict(ctx context.Context, e *semantic.DictExpression, scope values.Scope) (values.Value, error) {
```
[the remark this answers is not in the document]
jlapacik: I added one.

compiler/compiler.go
```diff
@@ -445,6 +471,29 @@ func compile(n semantic.Node, subst map[uint64]semantic.MonoType, scope Scope) (
 			t:     apply(subst, nil, n.TypeOf()),
 			array: elements,
 		}, nil
+	case *semantic.DictExpression:
```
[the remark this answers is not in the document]
jlapacik: I added one.

values/dict.go
```diff
@@ -232,7 +232,7 @@ func dictComparer(dictType semantic.MonoType) immutable.Comparer {
 	case semantic.Time:
 		return timeComparer{}
 	default:
-		panic(errors.Newf(codes.Internal, "invalid key nature: %s", n))
+		return intComparer{}
```
nathanielc, line 235: Whats the reasoning behind this change?

nathanielc (approved):
LGTM I do have one question about a change to the dict comparer but otherwise this is good to go

jsternberg (commented):
There's a function in the interpreter called `resolveValue`. Its intention is to resolve value references that are captured into literal versions so they can be captured. Can you fill in the dictionary implementation? It presently returns unimplemented as the error code and that's because the literal node didn't exist so I couldn't implement it.

values/dict.go
jlapacik: Thanks for catching this. I had meant to revert this.

Issue #3254, opened 2020-10-13
csv.from drops results after the first one


wolffcm:
If I run this script:
```javascript
import "csv"

inData = "
#datatype,string,long,string,string,string,string,double,dateTime:RFC3339
#group,false,false,true,true,true,true,false,false
#default,cpu,,,,,,,
,result,table,_field,_measurement,cpu,host,_value,_time
,,0,usage_guest,cpu,cpu-total,ip-192-168-1-16.ec2.internal,0,2020-10-09T22:18:00Z

#datatype,string,long,string,string,string,string,string,string,string,double,dateTime:RFC3339
#group,false,false,true,true,true,true,true,true,true,false,false
#default,disk,,,,,,,,,,
,result,table,_field,_measurement,device,fstype,host,mode,path,_value,_time
,,3,inodes_free,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4878333294,2020-10-09T22:18:00Z
,,3,inodes_free,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4878333286,2020-10-09T22:19:00Z
,,3,inodes_free,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4878333253.4,2020-10-09T22:19:44.191958Z
,,4,inodes_total,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4882452840,2020-10-09T22:18:00Z
,,4,inodes_total,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4882452840,2020-10-09T22:19:00Z
,,4,inodes_total,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4882452840,2020-10-09T22:19:44.191958Z
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119546,2020-10-09T22:18:00Z
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119554,2020-10-09T22:19:00Z
[...]

PR #3392, opened 2020-12-11
fix(csv): ensure csv.from() only returns 1 result

scbrickley:
Closes: #3254 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/csv/from.go
```diff
@@ -138,21 +139,29 @@ func (c *CSVSource) Run(ctx context.Context) {
...
 		// transformation. Unlike other sources, tables from csv sources
 		// are not read-only. They contain mutable state and therefore
 		// cannot be shared among goroutines.
-		decoder := csv.NewResultDecoder(csv.ResultDecoderConfig{
+		decoder := csv.NewMultiResultDecoder(csv.ResultDecoderConfig{
 			Allocator: c.alloc,
 			Context:   ctx,
 		})
-		result, decodeErr := decoder.Decode(strings.NewReader(c.tx))
+		results, decodeErr := decoder.Decode(ioutil.NopCloser(strings.NewReader(c.tx)))
```
jsternberg, line 153: You need to remember to defer a release to ensure it closes. I don't think it matters for this usage though so just good practice.

stdlib/csv/from_test.go
```diff
@@ -7,16 +7,181 @@ import (
...
 
+func TestFromCSV_ReturnSingleResult(t *testing.T) {
+	tests := []struct {
+		name    string
+		spec    *csv.FromCSVProcedureSpec
+		want    []*executetest.Table
+		wantErr error
+	}{
+		{
+			name: "Happy path",
```
jsternberg, line 30: Standard is either to use `HappyPath` (CamelCase) or `happy path` (lowercase sentence). I usually prefer the latter.

stdlib/csv/from.go
```diff
@@ -169,6 +178,13 @@ func (c *CSVSource) Run(ctx context.Context) {
 		if err != nil {
 			goto FINISH
 		}
+		if results.More() {
+			err = errors.New(
+				codes.FailedPrecondition,
+				"csv.from() should only return 1 result",
```
nathanielc, line 184: nit, I think this error message can read slightly better. Instead of saying it should only return 1 result, perhaps state it can only parse,read one result, this way its clear that if you get this error the fix is to provide only a single result to `csv.from()`.

jsternberg (approved):
Minor nitpicks but otherwise I think this is good to go. Even if the nitpicks aren't addressed I don't think it would cause any issues.

nathanielc (approved):

PR #3398, opened 2020-12-14
feat(tests): add ways to execute tests from tar and zip archives

### LINKED ISSUE #3334: Establish a way to ship the flux acceptance tests, executable from downstream consumers

jsternberg:
The test command can now execute tests from tar or zip archives. It can
also execute tests from individual `.flux` files that are specified on
the command line.

This can be used for downloading a zip archive of the flux source code
from github and then executing the tests without unzipping the file.

Fixes #3334.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/cmd/flux/cmd/test.go
+++ b/cmd/flux/cmd/test.go
@@ -1,9 +1,13 @@
 package cmd
 
 import (
+	"archive/tar"
+	"archive/zip"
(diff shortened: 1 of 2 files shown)


wolffcm (approved):
This looks good to me.

Issue #3400, opened 2020-12-15
allow difference to assume zero start


rogpeppe:
When `difference` is used, the first value is always dropped (either by setting it to null or by dropping the row that it's in).

When using `difference` for calculating total counts as reported by prometheus, the first value is actually significant - it's the total count up until that stage.

How about adding another parameter to `difference`, say `firstZero: bool`? When `firstZero` is true, the current value would start as zero - the first row would always be included and have an unchanged value (`keepFirst` would be ignored), and other values would take the difference from then.
When used in conjunction with `nonNegative: true`, a negative difference would likewise assume a zero start.

Examples:
`difference(firstZero: true)` of `[5, 7, 6, 10]` would produce: `[5, 2, -1, 4]`.

`difference(firstZero: true, nonNegative: true)` of `[5, 7, 6, 10]` would produce: `[5, 2, 6, 4]`.

jlapacik:
I like this proposal, but it seems that `firstZero` should replace `keepFirst` entirely. From my perspective this was the intended behavior of `keepFirst` all along. Perhaps @nathanielc can weigh in here.

nathanielc:
I don't recall the exact reasoning behind `keepFirst` it was introduced in this PR https://github.com/influxdata/flux/pull/1603 maybe we can gain context from there.

rogpeppe:
Aside from backward compatibility considerations, I can still see a use for `keepFirst` - sometimes one might wish to keep some columns around regardless of the status of the resulting difference column.

jlapacik:
My concern is mainly that we'll have two parameters `keepFirst` and `firstZero` that look very similar and so it might be confusing. Also they seem to be mutually exclusive which also isn't very user friendly. It appears that `keepFirst` was added so that `tripleExponentialDerivative` could be expressed in pure flux, but apparently it never was. So it might make more sense to just update the behavior of `keepFirst` since updating the first value from null to non-null likely isn't going to break any queries.

rogpeppe:
> So it might make more sense to just update the behavior of keepFirst since updating the first value from null to non-null likely isn't going to break any queries.

AFAICS that could easily break queries because it would change the numbers in the results - the first value would now appear in the results where before it wouldn't, so the results of a `sum` on the result could be wildly different.

I've been thinking a bit more about my proposal for `firstZero` here and although the proposed semantics fit my current use case well, I'm not sure that they're a good idea in general.

[...]

PR #3402, opened 2020-12-16
chore: stdlib/universe: simplify difference implementation

rogpeppe:
This prepares for implementing https://github.com/influxdata/flux/issues/3400
by regularizing the implementation somewhat.

The `difference` type is now responsible only for returning
values and validity of those values (including dealing with `nonNegative`);
the `Process` method deals with removing the first row.

We rename the `difference.first` field to `difference.valid` (reversing
its sense in the process) because it's currently a little misleading,
because it might be true even when an update method is called
for a record that's not first.

Also reduce indentation somewhat by returning/continuing early
when appropriate.

### DIFF
--- a/stdlib/universe/difference.go
+++ b/stdlib/universe/difference.go
@@ -1,8 +1,6 @@
 package universe
 
 import (
-	"math"
-
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/codes"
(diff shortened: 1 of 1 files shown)


stdlib/universe/difference.go
```diff
@@ -339,72 +317,67 @@ func (t *differenceTransformation) Finish(id execute.DatasetID, err error) {
...
+		d.valid = true
+		return 0, false
 	}
-
-	diff := v - d.pFloatValue
-	d.pFloatValue = v
-
-	return diff, false
+	// Note: the unsigned substraction works correctly even for negative differences
+	// because of two's-complement arithmetic.
```
jlapacik, line 362: Great comment

jlapacik (approved):
Looks great! Very readable.

Issue #3381, opened 2020-12-07
test rust scanner against queries collected from query log


adrian-thurston:
Write a standalone program that invokes the parser twice, once using the existing C based scanner, and again using the new rust-based scanner, and then compares the parse trees for equality. Run this against the current archive of queries collected from the query log to look for discrepancies. The archive is not currently accessible by the team, for the time being, get it from Adrian.

Doing this requires adding an option to the parser for selecting which scanner to use. Such an option would also be used by a feature flagging framework for incrementally enabling the scanner.

nathanielc:
Notes:

* Use Rust's assert_eq! macro to compare the produced ASTs
* Compare AST from old scanner with new scanner
* Add new unit test to parser if a discrepancy is found
* Commit program source to archiver repo (understanding its run once code)

PR #3409, opened 2020-12-21
refactor(parser): allow parser to toggle between scanners

scbrickley:
Helps with #3381

- Add `Scan` trait, implemented by both the C scanner and the Rust scanner
- Change `Parser` definition to use a `Scan` trait object instead of a Scanner struct
- Add `Parser::with_rust()` static method that returns a parser using the new rust scanner
- Add `with_rust` variants of all public functions that create a Parser

There were some additional changes made to appease `clippy`, specifically:

- Using `is_empty()` rather than comparing against an empty string
- Using `_type` in place of `as type` in scanner modules (e.g. `0_u32` instead of `0 as u32`)
- Removing explicit lifetime annotations where they could be elided


adrian-thurston (changes requested):
Overall looks good. Just thinking it might be better to create a scanner trait, have both the C and Rust scanner implement that, then create the one we need based on the `use_rs` var. Having two scanners present and cloning the input strikes me as more costly than necessary.

adrian-thurston (approved):
Looks good!

adrian-thurston (approved):

PR #3421, opened 2021-01-07
feat(cmd): expose test command so it can be used by external libraries

jsternberg:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/cmd/flux/cmd/test.go
+++ b/cmd/flux/cmd/test.go
@@ -28,37 +28,51 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var testCommand = &cobra.Command{
-	Use:   "test",
-	Short: "Run flux tests",
-	Long:  "Run flux tests",
-	Run: func(cmd *cobra.Command, args []string) {
-		fluxinit.FluxInit()
-		runFluxTests()
-	},
+type testFlags struct {
+	testNames []string
+	path      string
+	verbosity int
 }
 
-var testNames []string
-var rootDir string
-var verbosity int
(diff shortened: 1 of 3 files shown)


cmd/flux/cmd/test.go
```diff
@@ -362,3 +344,55 @@ func (t *TestReporter) Summarize(tests []*Test) {
 	}
 	fmt.Printf("\n---\nRan %d tests with %d failure(s)\n", len(tests), failures)
 }
+
+type TestSetupFunc func(ctx context.Context) (TestExecutor, error)
+
+type TestExecutor interface {
```
rockstar, line 350: Yeah, this is the API I had envisioned.

cmd/flux/cmd/test.go
```diff
@@ -362,3 +344,55 @@ func (t *TestReporter) Summarize(tests []*Test) {
...
+			return fmt.Errorf("%s", table.Stringify(tbl))
+		})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (testExecutor) Close() error { return nil }
```
rockstar, line 398: Is this complete necessary?

rockstar (approved):
Thanks for doing this; it's a real good start. Had one small question that might have to do more with idiomatic go (the use of `io.Closer` interface).

cmd/flux/cmd/test.go
jsternberg: I added this for influxdb and idpe since they would open resources and need to close them. But it largely depends if the test command is responsible for standing up infrastructure or not.

PR #3429, opened 2021-01-12
Constrained Record Types

Bolladeen:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/core/src/semantic/nodes.rs
```diff
@@ -157,23 +157,474 @@ impl Expression {
             Expression::Array(e) => e.typ.clone(),
             Expression::Dict(e) => e.typ.clone(),
             Expression::Function(e) => e.typ.clone(),
-            Expression::Logical(_) => MonoType::Bool,
+            Expression::Logical(_) => MonoType::Record(Box::new(types::Record::Extension {
```
jlapacik, line 160: Can we encapsulate these record types in helper functions? Something like
```Rust
fn type_bool() -> MonoType {
    MonoType::Record(...)
}

...

Expression::Logical(_) => type_bool(),
```

libflux/core/src/semantic/nodes.rs
```diff
@@ -1401,7 +1929,19 @@ impl ConditionalExpr {
             + acons
             + Constraints::from(vec![
                 Constraint::Equal {
-                    exp: MonoType::Bool,
+                    exp: MonoType::Record(Box::new(types::Record::Extension {
```
jlapacik, line 1932: We can replace this with a function call `type_bool()`

libflux/core/src/semantic/nodes.rs
```diff
@@ -1523,7 +2087,85 @@ impl IndexExpr {
             + Constraints::from(vec![
                 Constraint::Equal {
                     act: self.index.type_of(),
-                    exp: MonoType::Int,
+                    exp: MonoType::Record(Box::new(types::Record::Extension {
```
jlapacik, line 2090: Similarly we can replace this with a function call `type_int()`

libflux/core/src/semantic/nodes.rs
```diff
@@ -1622,27 +2264,46 @@ impl UnaryExpr {
...
+                exp: MonoType::Record(Box::new(types::Record::Extension {
+                        head: types::Property {
+                            k: "@E".to_string(),
+                            v: MonoType::Var(f.fresh()),
+                        },
+                        tail: MonoType::Var(f.fresh()),
+                    })),
                 loc: self.loc.clone(),
             }),
+			//TODO: Finish adding Unary Operator Constraints
```
jlapacik, line 2288: Can this comment be removed?

libflux/core/src/semantic/tests.rs
```diff
@@ -79,6 +81,113 @@ impl Importer for HashMap<&str, PolyType> {
...
+                    MonoType::Record(rect) => is_monotype(*rect, Some(h.v)),
+                    _ => None,
+                }
+			},
+		}
+    }
+}
+}
+
+fn convert_record_constraints (
```
jlapacik, line 113: What does this function do?

Bolladeen: That function is just a helper function that converts record constraints that are just monotypes back to their corresponding monotype so that the tests pass.

libflux/core/src/semantic/convert.rs
```diff
@@ -242,42 +242,311 @@ fn convert_monotype(
...
-    let mut cons = SemanticMap::<types::Tvar, Vec<types::Kind>>::new();
-
-    for (name, tvar) in tvars {
-        vars.push(tvar);
-        let mut kinds = Vec::<types::Kind>::new();
+    let mut recs = HashMap::<types::Tvar, MonoType>::new();
+    
+    let row_tvar = f.fresh();
+    vars.push(row_tvar);
+    for (name, tvar) in &tvars {
```
jlapacik, line 389: Rather than iterating over the `tvars` map, I might instead iterate over the constraints of the type expression, build a record constraint for each type variable, and then add a new mapping to `recs`.

libflux/core/src/semantic/convert.rs
```diff
@@ -242,42 +242,311 @@ fn convert_monotype(
...
+                                    v: MonoType::Var(new_tvar),
+                                },
+                                tail: rec
+                            }))
+                            }))
+                        },
+                        "Negatable" => {
+                            rec = MonoType::Record(Box::new(types::Record::Extension {
+                                head: types::Property {
+                                    k: "~".to_string(),
```
jlapacik, line 512: How about we change this to `@Neg` instead?

libflux/core/src/semantic/convert.rs
```diff
@@ -242,42 +242,311 @@ fn convert_monotype(
...
+                                tail: MonoType::Record(Box::new(types::Record::Extension {
+                                head: types::Property {
+                                    k: "!=".to_string(),
+                                    v: MonoType::Var(new_tvar),
+                                },
+                                tail: rec
+                            }))
+                            }))
+                        },
+                        "Nullable" => {
```
jlapacik, line 494: Did we say that this can be removed altogether?

libflux/core/src/semantic/convert.rs
```diff
@@ -242,42 +242,311 @@ fn convert_monotype(
...
+                                    k: "-".to_string(),
+                                    v: MonoType::Var(new_tvar),
+                                },
+                                tail: rec
+                            }))
+                        }
+                        "Divisible" => {
+                            rec = MonoType::Record(Box::new(types::Record::Extension {
+                                head: types::Property {
+                                    k: "*".to_string(),
```
jlapacik, line 419: Do we need to include all these labels for `Divisible`? Will we run into any issues if the only label is `/` similar to the `Addable` constraint?

libflux/core/src/semantic/nodes.rs
```diff
@@ -1013,278 +1460,359 @@ impl BinaryExpr {
...
         let cons = match self.operator {
             // The following operators require both sides to be equal.
             ast::Operator::AdditionOperator => Constraints::from(vec![
-                Constraint::Equal {
-                    exp: self.left.type_of(),
-                    act: self.right.type_of(),
-                    loc: self.right.loc().clone(),
-                },
-                Constraint::Equal {
+            	Constraint::Equal {
```
jlapacik, line 1469: We already talked about this, but I just wanted to document it. We either need another equality constraint tying the operand types to the type of the binary expression itself, or alternatively we need to update `Expression::type_of()` to return one of its operand types instead of the `typ` field.

I'm actually leaning towards the latter. The reason being that we could then remove the `typ` field from many of the nodes. This will significantly reduce the size of the semantic graph since instead of storing the type info with each node, we'll be computing it on the fly.

libflux/core/src/semantic/tests.rs
jlapacik, line 113: Once all tests are passing, we could probably remove these functions. I'm not sure what would better for the developer - being explicit in exposing the actual record type representation in a failing test, or hiding that representation and making the diff output simpler.

libflux/core/src/semantic/convert.rs
Bolladeen: Yup, completely agree.

PR #3431, opened 2021-01-13
feat: validate IP addresses from the dialer Control function

adrian-thurston:
The dialer provides a function that gets called after DNS resolution, but
before the network connection is initiated. Use that to validate IP addresses.
This replaces URL-based validation where the IP address is resolved before the
post, and during any redirects. This method prevents attacks where the IP
address is changed between validation and connect. Fixes influxdata/idpe#9480.

### DIFF
--- a/dependencies/http/http_test.go
+++ b/dependencies/http/http_test.go
@@ -3,12 +3,15 @@ package http
 import (
 	"bytes"
 	"io/ioutil"
+	"log"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"strings"
 	"testing"
 
+	"github.com/foxcpp/go-mockdns"
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux/codes"
 	depsUrl "github.com/influxdata/flux/dependencies/url"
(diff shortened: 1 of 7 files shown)


dependencies/url/validator.go
```diff
@@ -59,7 +72,7 @@ func init() {
 	}
 }
 
-//  isPrivateIP reports whether an IP exists in a known private IP space.
+//  IsPrivateIP reports whether an IP exists in a known private IP space.
```
rockstar, line 75: Did you mean to change this?

dependencies/http/http_test.go
```diff
@@ -54,23 +57,88 @@ func TestLimitedDefaultClient(t *testing.T) {
...
 }
 
 func TestInvalidRedirects(t *testing.T) {
+	bad := map[string]mockdns.Zone{
+		"bad01.example.com.": {A: []string{"127.6.6.6"}},
+	}
+
+	// Mock a dns server.
+	dnsLogger := log.New(ioutil.Discard, "mockdns server: ", log.LstdFlags)
+	srv, _ := mockdns.NewServerWithLogger(bad, dnsLogger, false)
```
rockstar, line 130: This is neat. I'd like to take a look at this, as this is how I usually expect mocking to occur, rather than the code generation most mock libraries seem to use.

rockstar (approved):
This is really great work. Thanks for getting to it.

dependencies/url/validator.go
adrian-thurston: Oops, I made it public in an intermediate version, but then changed it back after adding `ValidateIP` to the `Validate` Interface. Then forgot to fix the comment. Will fix.

dependencies/http/http_test.go
adrian-thurston: Yeah dug this up while looking for DNS resolution mocking and it was the only approach I found. Basically starts a DNS server on a random port and then alters the global default `Resolver` struct to connect to that for DNS, instead of the system settings. Maybe a bit heavy but couldn't find a way to hook the DNS lookup in go.

PR #3435, opened 2021-01-14
refactor(scanner): delete C scanner

scbrickley:


libflux/core/src/scanner/mod.rs
```diff
@@ -1,43 +1,47 @@
 #![allow(missing_docs)]
-include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
 pub type CChar = u8;
```
rockstar, line 3: Are we still using this? Can it be deleted?

libflux/core/src/scanner/mod.rs
```diff
@@ -1,43 +1,47 @@
...
-    pe: *const CChar,
-    eof: *const CChar,
-    last_newline: *const CChar,
-    cur_line: u32,
-    checkpoint: *const CChar,
-    checkpoint_line: u32,
-    checkpoint_last_newline: *const CChar,
-    token: TOK,
+    data: Vec<u8>,
+    ps: i32,
```
rockstar, line 20: This makes me raise an eyebrow, as a `CChar` is a `u8`, so we're taking what was a single byte and making it a signed 32-bit integer. I realize that `scan` has different ideas on its interface, but maybe that interface should change.

I'd prefer continuing to use a `u8` if we can get away with it (not sure if we need the type alias). Thoughts?

libflux/core/src/scanner/mod.rs
```diff
@@ -64,19 +68,19 @@ impl Scanner {
     pub fn new(data: CString) -> Self {
```
rockstar, line 68: If we're gonna drop all the `CString` usages in this patch (I had figured we'd defer them, but this is fine) we should also change this to take a `&str`, which simplifies the `Parser` consumption of `Scanner::new` as well.

libflux/core/src/scanner/scanner.rl
```diff
@@ -139,61 +141,60 @@
...
-        break;
-    case 1:
-        cs = flux_en_main_with_regex;
-        break;
-    case 2:
-        cs = flux_en_string_expr;
-        break;
+pub fn scan(
+    data: &[u8],
+    mode: i32,
```
rockstar, line 154: Right, so here's what I mean: the equivalent of an `unsigned char` in C/C++ is `u8` in rust, though there's also `char` as (essentially) an alias that carries semantic meaning.

rockstar (changes requested):
Just a few comments here, acknowledging that some of the build-related pieces we worked through together.

libflux/core/src/scanner/mod.rs
adrian-thurston: When you switch from scanning in C to scanning in Rust, some of the variables you need to populate and make available to the ragel-generated must change in type. C scans using `p` and `pe` as pointers to the input. In rust you define the input array `data`, and then `p` and `pe` become integers indexing the `data` array.

Also to note, the type of the `data` array must correspond to the type of the alphabet defined in the ragel file (see `alphtype`). That ensures that ragel and the host code are in agreement about how to interpret the input data. As things are in the ragel grammar, we must use `u8`, as all the literal tokens are defined in terms of `u8`.

rockstar: @adrian-thurston could we use `&[u8]` as data, rather than a `Vec`? It's essentially immutable, so doesn't need the functionality of `Vec`, and would probably benefit from _requiring_ sequential access, particularly for CPU cache use. What about `&[char]` even, which is equivalent, but carries a semantic value of saying "this should actually be a character we read from the string"?

libflux/core/src/scanner/mod.rs
adrian-thurston: Yeah agree, the parser had turned the input into a C string I think because in prep for the scanner needing it as a C string, but that's no longer the case. So no sense in converting to a C string, unless we need to in order to scan a [u8] ?

rockstar: We can do `"myString".as_bytes()` to get a `[u8]` from a `str`, [as documented here](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes)

libflux/core/src/scanner/mod.rs
rockstar: Please ignore the comment about `char` - rust's `char` is not actually equivalent to `u8` because `char` can be any unicode character (which means more than just one byte).

adrian-thurston: Could use `&[u8]` but not `&[char]` as they won't be interpreted right, because all the ragel code specifies patterns using unsigned single bytes notation (in ragel syntax). If they were all specified as signed, four-byte scalars in ragel, we could use `char`.

My rust sucks now that I've been in go so long ... but I can't seem to make `&[u8]` work because of borrowing, I think we need to copy it in using `[u8]` because the scanner outlives the data that's passed in. But hopefully the copy-in should be optimized away?

adrian-thurston:
```
19 |     data: [u8],
   |           ^^^^ doesn't have a size known at compile-time
```
Maybe this is why I used a vec? It's either `Vec<u8>` or a boxed `&[u8]`?

adrian-thurston: Again, my rust is not super-good right now, but I can only make `Vec<u8>` work, can't copy it into any combination of `Box`, `&`, or `[u8]`.  All result in unfixable compile errors for me.

adrian-thurston (changes requested):
Would recommend removing the file `libflux/scanner.c`.

libflux/core/src/scanner/mod.rs
rockstar: Yeah, there's likely some explicit lifetime work that we'll need to do in order to make it work. It's probably premature optimization at this point.

adrian-thurston (approved):
Looks great!

rockstar (approved):
...I guess I have to vote to remove the "Changes requested"

PR #3436, opened 2021-01-14
feat(values): improve performance for accessing data within a Value

jsternberg:
The `values.Value` interface represents a flux value and it is used by a
large number of runtime functions. The default implementation of the
basic values implements the data access functions, such as `Float()` and
`Int()`, for those respective types.

These access functions were deemed too complex to inline and so
performance was worse for accessing this data.

    values/values.go:69:6: cannot inline value.Float: function too complex: cost 135 exceeds budget 80

One reason is because `semantic.MonoType` required accessing the
flatbuffers data to determine the type nature. Instead of looking up the
type nature through a large function, this lookup has been moved to the
creation of the `semantic.MonoType` so that checking the nature is
accessing a member variable. For basic types, which generally use the
cached `semantic.BasicString`, this means that the type lookup is
incurred once instead of each time the value is accessed.

The other reason is because the panic function, `UnexpectedKind`, was
deemed too hairy by the go inliner. In order to optimize this function,
a new `unexpectedKind` error type was created and used to construct the
error. This `unexpectedKind` error type implements the `errors.As()`
function and can be accessed as a `*flux.Error`.

    err := values.UnexpectedKind(...)
    var ferr *flux.Error
    if errors.As(err, &ferr) {
        // As a flux error.
    }

This is the standard method of casting error types in go since 1.13.

Now, the `Float()` method is simpler and can be inlined.

    values/values.go:69:6: can inline value.Float with cost 31 as: method(value) func() float64 { CheckKind(v.t.Nature(), semantic.Float); return v.v.(float64) }

While the function itself and faster and can be inlined, there is still
a possibility that it will be faster but not inlined. This is because
`values.Value` is an interface and inlining isn't generally possible.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/values/values.go
+++ b/values/values.go
@@ -299,7 +299,7 @@ func NewRegexp(v *regexp.Regexp) Value {
 }
 
 func UnexpectedKind(got, exp semantic.Nature) error {
-	return errors.Newf(codes.Internal, "unexpected kind: got %q expected %q, trace: %s", got, exp, string(debug.Stack()))
+	return unexpectedKind{got: got, want: exp}
 }
 
 // CheckKind panics if got != exp.
@@ -307,6 +307,26 @@ func CheckKind(got, exp semantic.Nature) {
 	if got != exp {
 		panic(UnexpectedKind(got, exp))
 	}
+}
+
+type unexpectedKind struct {
+	got, want semantic.Nature
+}
+
+func (e unexpectedKind) Error() string {
+	// Reuse the Error method from flux.Error by converting
+	// this to a flux.Error type and then calling Error.
+	var err *errors.Error
+	e.As(&err)
+	return err.Error()
+}
+
+func (e unexpectedKind) As(target interface{}) bool {
+	if err, ok := target.(**errors.Error); ok {
+		*err = errors.Newf(codes.Internal, "unexpected kind: got %q expected %q, trace: %s", e.got, e.want, string(debug.Stack()))
+		return true
+	}
+	return false
 }
 
 // IsTimeable checks if value v is Timeable.
--- a/internal/errors/std.go
+++ b/internal/errors/std.go
@@ -6,3 +6,8 @@ import "errors"
 func Is(err, target error) bool {
 	return errors.Is(err, target)
 }
+
+// As is a wrapper around the errors.As function.
+func As(err error, target interface{}) bool {
+	return errors.As(err, target)
+}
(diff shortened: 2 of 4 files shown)


jsternberg:
Benchmark comparison:

```
benchmark                   old ns/op     new ns/op     delta
BenchmarkValue_Str-12       16.8          4.00          -76.19%
BenchmarkValue_Bytes-12     17.4          4.38          -74.83%
BenchmarkValue_Int-12       15.3          3.36          -78.04%
BenchmarkValue_UInt-12      15.5          3.39          -78.13%
BenchmarkValue_Float-12     15.5          3.40          -78.06%
BenchmarkValue_Bool-12      13.4          3.43          -74.40%

benchmark                   old allocs     new allocs     delta
BenchmarkValue_Str-12       0              0              +0.00%
BenchmarkValue_Bytes-12     0              0              +0.00%
BenchmarkValue_Int-12       0              0              +0.00%
BenchmarkValue_UInt-12      0              0              +0.00%
BenchmarkValue_Float-12     0              0              +0.00%
BenchmarkValue_Bool-12      0              0              +0.00%

benchmark                   old bytes     new bytes     delta
BenchmarkValue_Str-12       0             0             +0.00%
BenchmarkValue_Bytes-12     0             0             +0.00%
BenchmarkValue_Int-12       0             0             +0.00%
BenchmarkValue_UInt-12      0             0             +0.00%
BenchmarkValue_Float-12     0             0             +0.00%
BenchmarkValue_Bool-12      0             0             +0.00%
```

values/values.go
```diff
@@ -309,6 +309,24 @@ func CheckKind(got, exp semantic.Nature) {
 	}
 }
 
+type unexpectedKind struct {
+	got, want semantic.Nature
+}
+
+func (e unexpectedKind) Error() string {
+	var err *errors.Error
+	e.As(&err)
```
wolffcm, line 318: Why are you using `As` here instead of just returning the message? I figure you must have a reason for this but I'm not sure what it is.

wolffcm (approved):
I just had the comment on errors. If you want to add a comment to the code explaining it, that might be helpful.  Looks good otherwise, the improvement is nice!

values/values.go
jsternberg: I didn't want to duplicate the code.

PR #3440, opened 2021-01-18
fix(interval): clean up interval package

fchikwekwe:
Original PR with discussion here: https://github.com/influxdata/flux/pull/3302/files

I wanted to drop the commits that were added and then reverted, but thought they might be good to preserve for discussion, hence this new PR.

I also added some benchmarks to guard again performance regressions. It looks, like we actually made some gains on most test cases. 

Benchmarks : 
no_interval_-_simple-36 | 992 ns/op
interval_-_simple-36| 627 ns/op

no_interval_-_simple_with_negative_period-36 | 862 ns/op
interval_-_simple_with_negative_period-36 | 1033 ns/op

no_interval_-_simple_with_offset-36 | 860 ns/op
interval_-_simple_with_offset-36 | 696 ns/op

no_interval_-_simple_with_negative_offset-36 | 920 ns/op
interal_-_simple_with_negative_offset-36 | 680 ns/op

no_interval_-_simple_with_equal_offset_before-36 | 887 ns/op
interval_-_simple_with_equal_offset_before-36 | 648 ns/op

no_interval_-_simple_with_equal_offset_after-36 |  971 ns/op
interval_-_simple_with_equal_offset_after-36 | 680 ns/op

no_interval_-_simple_months-36 | 2808 ns/op
interval_-_simple_months-36 | 2720 ns/op

no_interval_-_simple_months_with_offset-36 |  3700 ns/op
interval_-_simple_months_with_offset-36 | 2385 ns/op


execute/window_test.go
```diff
@@ -1,6 +1,7 @@
 package execute_test
 
 import (
+	"github.com/influxdata/flux/interval"
```
jsternberg, line 4: Order imports.

execute/window_test.go
```diff
@@ -509,3 +510,270 @@ func errAsString(err error) (s string) {
 	}
 	return s
 }
+
```
jsternberg, line 513: `go fmt` (extra newline)

execute/window_test.go
```diff
@@ -509,3 +510,270 @@ func errAsString(err error) (s string) {
...
+				values.ConvertDurationMonths(3),
+				values.ConvertDurationMonths(1)),
+			t: mustTime("1970-01-01T00:00:00Z"),
+			want: testBounds{
+				Start: mustTime("1969-11-01T00:00:00Z"),
+				Stop:  mustTime("1970-02-01T00:00:00Z"),
+			},
+		},
+	}
+	for i := 0; i < 100; i++ {
```
jsternberg, line 765: What's this for loop for?

execute/window_test.go
```diff
@@ -509,3 +510,270 @@ func errAsString(err error) (s string) {
...
+		for _, tc := range testcases {
+			tc := tc
+			b.Run(tc.name, func(b *testing.B) {
+				got := tc.w.GetLatestBounds(tc.t)
```
jsternberg, line 769: Benchmarks need to be of the form:

```
for i := 0; i < b.N; i++ {
    // code to benchmark
}
```

For future reference, it's also not really necessary to check if the values you got were good or not. Test cases are used for that. Benchmarks are mostly used to verify that things go fast. For this case, I think these tests are good since it prevents the optimizer from just removing the entire block and I don't think it will impact the benchmark result.

interval/bounds.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	MaxTime = math.MaxInt64
+	MinTime = math.MinInt64
+)
+
+type Bounds struct {
+	start values.Time
+	stop  values.Time
+	// As additional windows are added to or subtracted from the initial bounds, index keeps track of how many
```
jsternberg, line 18: godoc comments should start with the name of the variable first. So something like:

```
// index keeps track of how many windows have been added or subtracted from the initial bounds etc.
```

A common formatting is to have a newline between the godoc comment and the previous line. Can you also document the other two fields even if they are kind of obvious? Even if it's obvious, it's useful because then there's no ambiguity for things like "is this time boundary inclusive or exclusive?"

fchikwekwe:
@jsternberg tests are passing now. I added the benchmark results to the PR description.

jsternberg (approved):

PR #3450, opened 2021-01-26
refactor(execute/table): break execute dependency in execute/table

jsternberg:
This breaks the dependendency on the execute package from execute/table
which then allows execute/table to be used within execute instead of the
other way around.

This opens up the possibility of using some of the table utilities
inside of the execute package.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


fchikwekwe:
I think this makes sense for the most part. I'm wondering if dataset is a clear enough name for that package though. Otherwise 👍🏾

fchikwekwe (approved):

PR #3453, opened 2021-01-26
feat(stdlib/testing/expect): add testing/expect package for test expectations

### LINKED ISSUE #3336: Add a way to assert planner rules in flux acceptance tests
As described in #3335, downstream test extensions need to be able to assert that pushdown rules were executed. One option is to keep track of planner rules in the execution, and have an `assertPlannerRules`function that inspects the execution

DOD:
  - Flux acceptance tests can use `testing.assertPlannerRules` or a similar API to assert that planner rules were followed

jsternberg:
The `testing/expect` package has been added to the standard library. It
contains functions to mark down when a testcase is expected to perform
some action within the engine and needs to assert that expectation is
fulfilled by the end of the test run.

The function is invoked at the beginning of the testcase and the
expectation is checked after the testcase has been executed.

Fixes #3336.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
This PR will require an accompanying change to the influxdb test runner and any other libraries that may invoke the test runner. The test runners will need to either invoke `testing.Inject()` or include the `testing.Framework` struct as a dependency to function. The testing check needs to be part of the normal flow cycle because otherwise there's nowhere to hook in to check at the end of a query. The test runner can't hook in because the dependency needs to be present where the query is actually run from. In the flux test runner, that's in the same process. In the influxdb test runner, that's in the same process, but within the test launcher and mutable data in a context doesn't pass over the mocked HTTP connection.

stdlib/testing/expect/expect.flux
```diff
@@ -0,0 +1,15 @@
...
+// beginning of a testcase, but it doesn't really
+// matter when they get invoked within the testcase.
+package expect
+
+// planner will cause the present testcase to
+// expect the given planner rules will be invoked
+// exactly as many times as the number given.
+//
+// The key is the name of the planner rule.
+builtin planner : (rules: [string: int]) => {}
```
nathanielc, line 15: Nice use of the dict type :)

stdlib/universe/merge_filter_test.flux
```diff
@@ -0,0 +1,50 @@
...
+    testing.diff(got: result, want: testing.loadMem(csv: output))
+}
+
+testcase merge_filter_flag_off {
+    option planner.disableLogicalRules = ["MergeFiltersRule"]
+    expect.planner(rules: ["MergeFiltersRule": 0])
+
+    result = merge_filter_fn()
+    testing.diff(got: result, want: testing.loadMem(csv: output))
+}
```
nathanielc, line 50: This is wonderful, I love seeing these new testcases. They read well and can test what we actually want to test!

plan/heuristic_planner.go
```diff
@@ -51,8 +53,11 @@ func (p *heuristicPlanner) matchRules(ctx context.Context, node Node) (Node, boo
 			newNode, changed, err := rule.Rewrite(ctx, node)
 			if err != nil {
 				return nil, false, err
+			} else if !changed {
+				continue
```
nathanielc, line 57: This looks like an error, but I could be reading it wrong. By continuing here the `node = newNode` statement is not executed when it would have been in the previous code. Is there a reason that's different now?  If this is indeed an error then why didn't tests fail? Can you add a test case to cover that path if needed?

dependencies/testing/testing.go
```diff
@@ -0,0 +1,98 @@
...
+// This is a no-op if testing dependencies are not present.
+func Check(ctx context.Context) error {
+	if tf, err := getTestingFramework(ctx); err == nil {
+		return tf.Check()
+	}
+	return nil
+}
+
+// ExpectPlannerRule will mark that a planner rule is expected
+// to be executed at least n number of times.
```
nathanielc, line 61: Elsewhere it docs say that the rule is expected to be executed exactly `n` times. The implementation seems to indicate that as well. Can you update this doc to be consistent?

dependencies/testing/testing.go
```diff
@@ -0,0 +1,98 @@
+package testing
```
nathanielc, line 1: Can you add some simple unit tests for the methods in this package? Something along the lines of create a context, call the various methods call Check etc.

nathanielc (changes requested):
Looks really good. I like the design and how it fits into the dependency system so well. 

My favorite part is seeing the Flux acceptance test use the new function. Its beautiful! 

I have a few questions below and a request for some new test cases.

plan/heuristic_planner.go
jsternberg: Hm, my understanding was that if `changed` was false, the rewrite rule did not run and nothing should be done. It seems like all of the planner rules just happened to always return their own selves whenever they returned false rather than nil or anything else.

dependencies/testing/testing.go
jsternberg: Yes. The original implementation was this and then I changed it because otherwise I couldn't test for the absence of a planner rule.

plan/heuristic_planner.go
jsternberg: I wasn't really sure how to change this so I just reverted the behavior to what it was previously.

dependencies/testing/testing.go
jsternberg: Done.

dependencies/testing/testing.go
jsternberg: Done.

nathanielc (approved):
LGTM

PR #3460, opened 2021-01-29
feat(stdlib/universe): add interval package to window transformation

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/interval/bounds.go
+++ b/interval/bounds.go
@@ -22,6 +22,14 @@ type Bounds struct {
 	index int
 }
 
+// NewBounds create a new Bounds given start and stop values
+func NewBounds(start, stop values.Time) Bounds {
+	return Bounds{
+		start: start,
+		stop:  stop,
+	}
+}
+
 func (b Bounds) Start() values.Time {
 	return b.start
 }
@@ -32,6 +40,11 @@ func (b Bounds) Stop() values.Time {
 
 func (b Bounds) IsEmpty() bool {
 	return b.start >= b.stop
+}
+
+// IsZero returns true if the start and stop values are both zero.
+func (b Bounds) IsZero() bool {
+	return b.start == 0 && b.stop == 0
 }
 
 func (b Bounds) String() string {
@@ -55,4 +68,54 @@ func (b Bounds) Length() values.Duration {
 		return values.ConvertDurationNsecs(0)
 	}
 	return b.stop.Sub(b.start)
+}
+
+// Intersect returns the intersection of two bounds.
+// It returns empty bounds if one of the input bounds are empty.
+// TODO: there are several places that implement bounds and related utilities.
+//  consider a central place for them?
+func (b Bounds) Intersect(o Bounds) Bounds {
+	if b.IsEmpty() || o.IsEmpty() || !b.Overlaps(o) {
+		return Bounds{
+			start: b.start,
+			stop:  b.stop,
+		}
+	}
+	i := Bounds{}
+
+	i.start = b.start
+	if o.start > b.start {
+		i.start = o.start
+	}
+
+	i.stop = b.stop
+	if o.stop < b.stop {
+		i.stop = o.stop
+	}
+
+	return i
+}
+
+// Union returns the smallest bounds which contain both input bounds.
+// It returns empty bounds if one of the input bounds are empty.
+func (b Bounds) Union(o Bounds) Bounds {
+	if b.IsEmpty() || o.IsEmpty() {
+		return Bounds{
+			start: values.Time(0),
+			stop:  values.Time(0),
+		}
+	}
+	u := new(Bounds)
+
+	u.start = b.start
+	if o.start < b.start {
+		u.start = o.start
+	}
+
+	u.stop = b.stop
+	if o.stop > b.stop {
+		u.stop = o.stop
+	}
+
+	return *u
 }
--- a/interval/window.go
+++ b/interval/window.go
@@ -55,6 +55,14 @@ func (w Window) IsZero() bool {
 	return w.every.IsZero()
 }
 
+func (w Window) Every() values.Duration {
+	return w.every
+}
+
+func (w Window) Period() values.Duration {
+	return w.period
+}
+
 func (w Window) isValid() error {
 	if w.every.IsZero() {
 		return errors.New(codes.Invalid, "duration used as an interval cannot be zero")
@@ -309,6 +317,3 @@ func isBeforeWithinMonth(a, b values.Time) bool {
 	}
 	return false
 }
-
-//TODO
-// Add tests very far away from the epoch
--- a/stdlib/universe/range.go
+++ b/stdlib/universe/range.go
@@ -96,7 +96,6 @@ func newRangeProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.Proc
 	if spec.TimeColumn == "" {
 		spec.TimeColumn = execute.DefaultTimeColLabel
 	}
-
 	bounds := flux.Bounds{
 		Start: spec.Start,
 		Stop:  spec.Stop,
@@ -110,7 +109,6 @@ func newRangeProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.Proc
 	} else if bounds.IsEmpty() {
 		return nil, errors.New(codes.Invalid, "cannot query an empty range")
 	}
-
 	return &RangeProcedureSpec{
 		Bounds:      bounds,
 		TimeColumn:  spec.TimeColumn,
@@ -122,6 +120,7 @@ func newRangeProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.Proc
 func (s *RangeProcedureSpec) Kind() plan.ProcedureKind {
 	return RangeKind
 }
+
 func (s *RangeProcedureSpec) Copy() plan.ProcedureSpec {
 	ns := new(RangeProcedureSpec)
 	ns.Bounds = s.Bounds
@@ -335,9 +334,11 @@ func (t *rangeTransformation) createRangeGroupKey(inKey flux.GroupKey, startKeyC
 func (t *rangeTransformation) UpdateWatermark(id execute.DatasetID, mark execute.Time) error {
 	return t.d.UpdateWatermark(mark)
 }
+
 func (t *rangeTransformation) UpdateProcessingTime(id execute.DatasetID, pt execute.Time) error {
 	return t.d.UpdateProcessingTime(pt)
 }
+
 func (t *rangeTransformation) Finish(id execute.DatasetID, err error) {
 	t.d.Finish(err)
 }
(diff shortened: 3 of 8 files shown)


interval/window.go
```diff
@@ -304,6 +312,3 @@ func isBeforeWithinMonth(a, b values.Time) bool {
 	}
 	return false
 }
-
-//TODO
-// Add tests very far away from the epoch
```
fchikwekwe, line 309: Done in a previous PR.

stdlib/universe/range.go
```diff
@@ -145,8 +143,15 @@ func createRangeTransformation(id execute.DatasetID, mode execute.AccumulationMo
 	d := execute.NewDataset(id, mode, cache)
 
 	bounds := a.StreamContext().Bounds()
+	e := execute.Bounds{}
+	if bounds != nil {
+		e = execute.Bounds{
+			Start: bounds.Start(),
+			Stop:  bounds.Stop(),
+		}
+	}
```
fchikwekwe, line 152: I added this check because I ran into panics when bounds was nil

stdlib/universe/window_test.go
```diff
@@ -517,12 +514,12 @@ func TestFixedWindow_Process(t *testing.T) {
...
-							{start + 1*execute.Time(time.Minute), start + 2*execute.Time(time.Minute), start + execute.Time(80*time.Second), 8.0},
-							{start + 1*execute.Time(time.Minute), start + 2*execute.Time(time.Minute), start + execute.Time(90*time.Second), 9.0},
-							{start + 1*execute.Time(time.Minute), start + 2*execute.Time(time.Minute), start + execute.Time(100*time.Second), 10.0},
-							{start + 1*execute.Time(time.Minute), start + 2*execute.Time(time.Minute), start + execute.Time(110*time.Second), 11.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(0*time.Second), 0.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(10*time.Second), 1.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(20*time.Second), 2.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(30*time.Second), 3.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(40*time.Second), 4.0},
+							{start, start + 1*values.Time(time.Minute), start + values.Time(50*time.Second), 5.0},
```
fchikwekwe, line 522: Underlapping windows with a period that is smaller than the every now start at the beginning of the window. This was a breaking change, so tests dealing with this needed to be updated.

stdlib/universe/window_test.go
```diff
@@ -568,12 +565,12 @@ func TestFixedWindow_Process(t *testing.T) {
...
```
fchikwekwe, line 573: Same thing here about underlapping windows.

execute/transformation.go
```diff
@@ -87,7 +88,7 @@ func (ts TransformationSet) Finish(id DatasetID, err error) {
 // StreamContext represents necessary context for a single stream of
 // query data.
 type StreamContext interface {
-	Bounds() *Bounds
+	Bounds() *interval.Bounds
```
jsternberg, line 91: This change isn't clear to me. It seems that `interval.Bounds` is intended to be created from the window function and that the `interval.Bounds` itself isn't meaningful outside of it being generated by the windowing function. I fear that using `interval.Bounds` everywhere where we need a start and stop and allowing a person to create a bounds will invalidate some of the math included in the bounds with the index and give a false sense of security. I wonder if we can keep this the same so `execute.Bounds` is used to denote a complete boundary (such as for range) while `interval.Bounds` is used to describe the boundary of an interval. Does that make sense?

interval/bounds.go
```diff
@@ -22,6 +22,24 @@ type Bounds struct {
 	index int
 }
 
+// NewBounds create a new Bounds given start and stop values
+func NewBounds(start, stop values.Time) *Bounds {
```
jsternberg, line 26: See comment above. I fear that this method may not work the way we want it to.

jsternberg (commented):
The windowing portions look fine but I think the PR can be improved by removing the `interval.Bounds` related portions and using the original `execute.Bounds` in those places. I think the two structs potentially serve different purposes and both are useful. `execute.Bounds` describes a hard boundary for the query such as a start and stop while `interval.Bounds` represents a bounds produced by the windowing function. If we confuse those, I fear that we might run into improper use of the libraries since some of the interval logic relies on the index being present.

execute/executor.go
```diff
@@ -44,13 +45,18 @@ func NewExecutor(logger *zap.Logger) Executor {
...
+	bounds        *interval.Bounds
+	executeBounds *Bounds
 }
 
-func (ctx streamContext) Bounds() *Bounds {
+func (ctx streamContext) Bounds() *interval.Bounds {
 	return ctx.bounds
 }
 
+func (ctx streamContext) ExecuteBounds() *Bounds {
```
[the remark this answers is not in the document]
fchikwekwe: sure

interval/bounds.go
```diff
@@ -22,6 +22,24 @@ type Bounds struct {
...
+	return &Bounds{
+		start: start,
+		stop:  stop,
+	}
+}
+
+// SetStart sets the start on a given Bounds
+func (b Bounds) SetStart(start values.Time) {
```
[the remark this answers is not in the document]
fchikwekwe: fair enough. I think I created them when I was still modifying too much code. They're not actually used in flux.

interval/bounds.go
fchikwekwe: I'm guessing you mean that creating a new bounds is akin to using this like in range? The actual use case is mainly for testing purposes. Do you think that's okay or do you still want me to remove it?

interval/bounds.go
```diff
@@ -56,3 +79,53 @@ func (b Bounds) Length() values.Duration {
...
+	if o.stop < b.stop {
+		i.stop = o.stop
+	}
+
+	return i
+}
+
+// Union returns the smallest bounds which contain both input bounds.
+// It returns empty bounds if one of the input bounds are empty.
+func (b *Bounds) Union(o *Bounds) *Bounds {
```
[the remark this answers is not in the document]
fchikwekwe: got it

execute/executor.go
```diff
@@ -44,13 +45,18 @@ func NewExecutor(logger *zap.Logger) Executor {
 }
 
 type streamContext struct {
-	bounds *Bounds
+	bounds        *interval.Bounds
```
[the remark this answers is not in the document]
fchikwekwe: Constructing an interval.Bounds requires a method to create one. I have the NewBounds method that I added for this, but you were saying something about it not working properly. I think this is a case for keeping it if we can.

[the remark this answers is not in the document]
fchikwekwe: Also, if this is the case, then maybe we get rid of the other exposure method mentioned for execute Bounds?

execute/transformation.go
fchikwekwe: Yeah, I guess that makes some sense. A version of bounds that is mutable for creating range bounds and a version of bounds that are created for windowing and immutable. Is that right?

fchikwekwe: I went ahead and took your suggesion here. Only windowing bounds should now be affected.

interpreter/interpreter_test.go
```diff
@@ -858,8 +858,8 @@ func TestStack(t *testing.T) {
 			FunctionName: "window",
 			Location: ast.SourceLocation{
 				File:   "universe.flux",
-				Start:  ast.Position{Line: 135, Column: 12},
-				End:    ast.Position{Line: 135, Column: 49},
+				Start:  ast.Position{Line: 136, Column: 12},
+				End:    ast.Position{Line: 136, Column: 49},
```
fchikwekwe, line 862: I changed the `universe.flux` file, so the desired output changed.

stdlib/universe/window.go
```diff
@@ -181,8 +181,11 @@ func createWindowTransformation(id execute.DatasetID, mode execute.AccumulationM
 	cache := execute.NewTableBuilderCache(a.Allocator())
 	d := execute.NewDataset(id, mode, cache)
 
-	bounds := a.StreamContext().Bounds()
-	if bounds == nil {
+	bounds := execute.Bounds{
+		Start: a.StreamContext().Bounds().Start,
+		Stop:  a.StreamContext().Bounds().Stop,
+	}
+	if bounds == (execute.Bounds{}) {
```
jsternberg, line 188: Does this part need to be changed? I'm a bit uncomfortable with comparing a bounds to the empty struct as a bounds of `[0, 0)` is technically different from nil.

stdlib/universe/universe.flux
```diff
@@ -76,6 +76,7 @@ builtin tripleExponentialDerivative : (<-tables: [{B with _value: A}], n: int) =
 builtin union : (tables: [[A]]) => [A] where A: Record
 builtin unique : (<-tables: [A], ?column: string) => [A] where A: Record
 builtin window : (<-tables: [A], ?every: duration, ?period: duration, ?offset: duration, ?timeColumn: string, ?startColumn: string, ?stopColumn: string, ?createEmpty: bool) => [B] where A: Record, B: Record
+builtin windowInterval : (<-tables: [A], ?every: duration, ?period: duration, ?offset: duration, ?timeColumn: string, ?startColumn: string, ?stopColumn: string, ?createEmpty: bool) => [B] where A: Record, B: Record
```
jsternberg, line 79: I thought that we were introducing this through a planner rule. Is this just for testing or is this a new function?

fchikwekwe: Yes, I needed to have a separate procedure spec for the planner to be aware of the new transformation. It required so much duplication that I decided to temporarily add a new transformation (windowInterval). We can delete the old window transformation and promote this one once the feature flag for interval is rolled out.

fchikwekwe: my WIP for the new feature flag https://github.com/influxdata/idpe/pull/9730

stdlib/universe/window.go
fchikwekwe: I changed it so that the check is performed before the assignment. Does that look better to you?

fchikwekwe: It compares to `nil` now.

stdlib/universe/universe.flux
jsternberg: If the goal is to replace it with a new procedure spec, this isn't required. The function name only needs to be specified if you plan on invoking it from directly within a flux script like `|> windowInterval`. If the plan is just to replace the window transformation with a new procedure spec, then you can omit this.

jsternberg (approved):

stdlib/universe/universe.flux
fchikwekwe: I've removed the unnecesary pieces from the windowInterval transformation so that its not user facing any longer.

PR #3463, opened 2021-01-29
refactor(execute): expose direct access to process message for transformations

jsternberg:
The `ProcessMessage` method can now be directly implemented by
transformations and the adapter code is present for existing
transformations.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/transport.go
+++ b/execute/transport.go
@@ -7,26 +7,38 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/internal/execute/table"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
 )
 
+// Transport is an interface for handling raw messages.
 type Transport interface {
-	Transformation
-	// Finished reports when the Transport has completed and there is no more work to do.
+	// ProcessMessage will process a message in the Transport.
+	ProcessMessage(m Message) error
+}
+
+// AsyncTransport is a Transport that performs its work in a separate goroutine.
+type AsyncTransport interface {
+	Transport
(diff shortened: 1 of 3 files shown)


execute/transport.go
```diff
@@ -263,10 +267,24 @@ type Message interface {
 type MessageType int
 
 const (
+	// RetractTableType is sent when the previous table for
```
wolffcm, line 270: Thank you for all these comments!

wolffcm (commented):
This seems reasonable to me, but I'm not clear on the practical benefit of these changes, can you answer that for me?

jsternberg:
The main purpose was to help with extensibility and composability. At the present moment, if we want to add a new message type to the system, then either a change to the `Transformation` interface or using a `typecast` to something like:

```
type ProcessViewTransformation interface {
    Transformation
    ProcessView(...) error
}
```

One of these two is required because the consecutive transport masquerades as a transformation and needs to forward the message through function calls. This makes the `Transformation` interface very heavy and prone to break. The second version also doesn't work very well when wrapping another type because typecasts can't be forwarded to wrapped types. This simplifies the requirements for a transformation by just saying "If you implement this one method you're good" and then each transformation that implements that method can decide to just ignore messages they don't understand or perform some default actions. This makes delegation and wrapping easier.

In the next commit after this one, I add some new message types that execkit transformations will both produce and handle and the transformation adapter is meant to communicate between these two message handling paradigms. I'll push up that commit so you can see what I mean.

jsternberg:
Here's the follow up commit to this one: https://github.com/influxdata/flux/commit/c23c74a2f7b35bd64956d6106c412d7947bfd761.

wolffcm:
Thinking this over a bit more and looking at the follow up commit that you linked, I have some concerns. It seems like it adds complexity to the query execution layer--an area I already find more than a bit confusing. I suspect that there's some good ideas here, but I think this should be opened up to a broader discussion with the team, and I don't know that we have the bandwidth right now.

@nathanielc What are your thoughts on this, with regards to prioritization or otherwise?

PR #3464, opened 2021-02-01
fix(parser): Break parse loop when bad array item is found

scbrickley:
Closes https://github.com/influxdata/idpe/issues/9592

Exit loop in `parse_array_items_rest()` after parsing a bad array item

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


jlapacik (commented):
I like the fix, but the test seems maybe a little too rigid for testing a parse failure. Maybe just assert that the first statement in the body parses as an array expression?

It just seems like it's very easy to make an update to the parser that would make this test fail due to a BadExpression being constructed differently.

adrian-thurston:
Hey @scbrickley what was the reason we couldn't go for the "unable to make progress" approach? Just matching on a bad expression and breaking from that means we can't recover from a bad expression when possible.

scbrickley:
> Hey @scbrickley what was the reason we couldn't go for the "unable to make progress" approach? Just matching on a bad expression and breaking from that means we can't recover from a bad expression when possible

So I tried keeping track of the byte offset for the tokens and wrapping the return value in a `Expression::Bad`, but other tests would fail if I did that. I worked around it by adding a counter to track how many times we parsed a token with the same byte offset, but that was fragile and produced strange build errors. Something like:
```
BadExpression is not allowed in semantic analysis
```
If I set the counter to a higher treshold, I no longer see those errors, but that solution seemed unwieldy and error prone.

Could you explain what you mean when you say we can't recover from it? My understanding is that we pass the array expression with a bad item up the call stack and it's up to the callers to report the error.

adrian-thurston:
We should be able to get a bad expression, a comma, then continue on parsing. The tree that is returned will contain a bad expression and we can continue on to pick up more of the tree. But considering there are test cases that would need to be adjusted and it's non-trivial, and we need this one out the door, It's better to move forward with the fix you have here.

nathanielc:
@adrian-thurston @scbrickley lets create an issue to update this fix to continue parsing when it finds a bad expression. We can get that issue into next sprint. 

Then we can get this fix out today.

nathanielc (approved):

PR #3471, opened 2021-02-03
feat: getters for a window's every and period

jlapacik:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/interval/window.go
+++ b/interval/window.go
@@ -21,6 +21,11 @@ func init() {
 // See https://github.com/influxdata/flux/issues/2093
 //
 // Window is a description of an infinite set of boundaries in time.
+//
+// Note the properties of this struct should remain private.
+// Furthermore they should not be exposed via public getter methods.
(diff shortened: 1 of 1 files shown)


nathanielc:
Can you explain the reasoning behind exposing the `every` and `period`? The reason they were not exposed initially is so that consumers of the interval code could not do their own duration math, instead they are required to use the provided Bounds API.

jlapacik:
When pushing bare aggregates down to storage, the query planner sets the `every` parameter to the max value. Storage then checks this value to determine which implementation to use (bare vs window).

rockstar (approved):
This code itself looks fine, but I'm interested in the conversation between you and @nathanielc at an architectural level; I can't confidently comment on that.

nathanielc (approved):
Thanks for the clarification. Can you make sure to change the PR title so that the message that makes it into the change log and git logs is accurate?

PR #3482, opened 2021-02-08
feat(stdlib/sql): add Vertica database support

bryanherger:
This pull request adds Vertica database support to Flux using vertica-go v1.1.1 driver. It is based on the PR for SAP HANA added previously.

### Done checklist
- [X] docs/SPEC.md updated
- [X] Test cases written

### DIFF
--- a/stdlib/sql/to.go
+++ b/stdlib/sql/to.go
@@ -263,7 +263,7 @@ func getTranslationFunc(driverName string) (func() translationFunc, error) {
 	switch driverName {
 	case "sqlite3":
 		return SqliteColumnTranslateFunc, nil
-	case "postgres", "sqlmock":
+	case "postgres", "sqlmock", "vertica", "vertigo":
 		return PostgresColumnTranslateFunc, nil
 	case "mysql":
 		return MysqlColumnTranslateFunc, nil
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 	github.com/snowflakedb/gosnowflake v1.3.4
 	github.com/spf13/cobra v0.0.3
 	github.com/uber/athenadriver v1.1.4
+	github.com/vertica/vertica-sql-go v1.1.1
 	go.uber.org/zap v1.14.0
 	golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6
 	golang.org/x/net v0.0.0-20200625001655-4c5254603344
--- a/go.sum
+++ b/go.sum
@@ -259,6 +259,8 @@ github.com/uber-go/tally v3.3.15+incompatible h1:9hLSgNBP28CjIaDmAuRTq9qV+UZY+9P
 github.com/uber-go/tally v3.3.15+incompatible/go.mod h1:YDTIBxdXyOU/sCWilKB4bgyufu1cEi0jdVnRdxvjnmU=
 github.com/uber/athenadriver v1.1.4 h1:k6k0RBeXjR7oZ8NO557MsRw3eX1cc/9B0GNx+W9eHiQ=
 github.com/uber/athenadriver v1.1.4/go.mod h1:tQjho4NzXw55LGfSZEcETuYydpY1vtmixUabHkC1K/E=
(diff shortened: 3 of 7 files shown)


stdlib/sql/vertica.go
```diff
@@ -0,0 +1,170 @@
...
+		case bool, int, uint:
+			row[i] = values.New(int64(col.(int)))
+		case int64, uint64, float64:
+			row[i] = values.New(col)
+		case string:
+			row[i] = values.New(col)
+		case time.Time:
+			row[i] = values.NewTime(values.ConvertTime(col))
+		case []uint8:
+			fmt.Print("hack:", m.columnTypes[i])
```
adrian-thurston, line 54: This got left behind? There shouldn't be any IO here.

Also, it seems that Vertica is based off of postgres, based on the usage of `PostgreColumnTranslateFunc`. But here it appears to be based off of mysql? Sorry not too familiar with Vertica.

stdlib/sql/to.go
```diff
@@ -263,7 +263,7 @@ func getTranslationFunc(driverName string) (func() translationFunc, error) {
 	switch driverName {
 	case "sqlite3":
 		return SqliteColumnTranslateFunc, nil
-	case "postgres", "sqlmock":
+	case "postgres", "sqlmock", "vertica", "vertigo":
```
adrian-thurston, line 266: Is `vertigo` another name for Vertica databases?

bryanherger: An issue from our side is that the driver packages were renamed from "vertigo" to "vertica" at some point, so this is done as a guard for users who might have pulled or cached older GO code.  So yes, it's a deprecated alias.

stdlib/sql/vertica.go
bryanherger: Vertica is based on Postgres, but I based my patch on the SAP HANA patch.  This fmt.Print appears to be an artifact from that.  I'm not sure this code would be reachable since Vertica uses 64-bit integer types.

stdlib/sql/vertica.go
```diff
@@ -0,0 +1,170 @@
...
+			// This works, but you can also just just add the DSN parameter parseTime=true (see line 136)
+			case flux.TTime:
+				t, err := time.Parse(layout, string(col))
+				if err != nil {
+					fmt.Print(err)
+				}
+				row[i] = values.NewTime(values.ConvertTime(t))
+			default:
+				row[i] = values.NewString(string(col))
+			}
```
scbrickley, line 80: My understanding of this code block is that, if we get a slice of bytes, we can't necessarily tell what type it is, so we check the flux column type and attempt to convert to that. I follow the logic there, and the code seems good, but I have some questions I'd like to answer before this merges.

From what I can tell, the code block (including the comment referencing MySQL), was originally added to the `mysql` package, then it got copied and pasted over to the `postgres` package, and I would guess that the same thing happened here.

If this 'hack' was only introduced for the sake of compatibility with MySQL, do we really need it here? Looking at some of the earlier comments on this PR, it's suggested that this code is unreachable when interacting with veritca. I'm not familiar with vertica... can we confirm that's the case? Is it impossible for vertica to have a column of bytes?

If it turns out this code is useful outside of MySQL, then let's keep it and just remove the misleading comments. But if we can confirm that it's unreachable, I think it would be best to simplify here. Maybe just convert it to a string and call it done?

stdlib/sql/vertica.go
```diff
@@ -0,0 +1,170 @@
...
+
+// VerticaTranslateColumn translates flux colTypes into their corresponding Vertica column type
+func VerticaColumnTranslateFunc() translationFunc {
+	c := map[string]string{
+		flux.TFloat.String():  "FLOAT",
+		flux.TInt.String():    "INTEGER",
+		flux.TUInt.String():   "INTEGER",
+		flux.TString.String(): "VARCHAR",
+		flux.TTime.String():   "TIMESTAMP",
+		flux.TBool.String():   "BOOL",
```
scbrickley, line 160: In the `postgres` package, values with the flux type `TInt` and `TUInt` map to the SQL type `BIGINT`, and values with the flux type `TString` map to the SQL type `TEXT`. Just curious why that wouldn't work here. Does vertica not support those types?

PR #3484, opened 2021-02-08
feat(stdlib): added new Window function signature

Bolladeen:
- created new window builtin function that implements new window function signature
  - added to experimental package

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


scbrickley (approved):
LGTM - Although I haven't spent much time in this code myself. Maybe it's worth it to wait for @nathanielc's review?

libflux/flux/benches/benchmarks.rs
```diff
@@ -1070,6 +1070,16 @@ builtin table : (<-tables: [A], columns: C) => [B] where A: Record, B: Record, C
...
+builtin window2 : (
+    <-tables: [{T with _time: time}]
+    ?every: duration,
+    ?period: duration,
+    ?offset: duration,
+    ?createEmpty: bool
+) => [{A with
+    _start: time, 
+    _stop: time,
+    _time: time}]
```
nathanielc, line 1082: @rockstar Thoughts? Should we be trying to keep these benchmarks files up to date when we make changes? Or should we consider them static now?

stdlib/universe/window_interval_2.go
```diff
@@ -0,0 +1,287 @@
...
+type WindowIntervalOpSpec2 struct {
+	Every       flux.Duration `json:"every"`
+	Period      flux.Duration `json:"period"`
+	Offset      flux.Duration `json:"offset"`
+	CreateEmpty bool          `json:"createEmpty"`
+}
+
+func init() {
+	windowSignature := runtime.MustLookupBuiltinType("universe", "window2")
+	runtime.RegisterPackageValue("universe", WindowKind2, flux.MustValue(flux.FunctionValue(WindowKind2, createWindowOpSpec, windowSignature)))
```
nathanielc, line 29: This file contains three types:

An OpSpec, ProcedureSpec and a Transformation, all of those are the same as the window_interval.go code. 

I think you can delete all of those from this file and it will still work. All that is needed is that you register the package value as you do on this line here. Then under the hood they become the same function.

stdlib/universe/window_test.go
```diff
@@ -101,6 +101,20 @@ func TestFixedWindow_PassThrough(t *testing.T) {
 	})
 }
 
+func TestFixedWindow2_PassThrough(t *testing.T) {
```
nathanielc, line 104: Can you add a Flux acceptance test too? The ones that are written in _test.flux files.

nathanielc (changes requested):
Looks good. I think we can simplify and remove a lot of the copied code, see below.

libflux/flux/benches/benchmarks.rs
rockstar: I guess I must've missed when this got added. That's a very big long string that we don't necessarily need in the source. We should change line 23 to be `const LONGER: &'static str = include_str!("path/to/everything.flux")` like we do in other benchmarks to embed it at compile time. I realize it's not strictly part of this PR, but that might be appropriate to do now, since we have to touch it; it'll prevent others from having to make that decision when it needs to be changed next time.

Bolladeen: @rockstar That sounds like a good idea. I'll make that change in this PR.

stdlib/universe/window_interval_2.go
Bolladeen: I double checked, but I could be wrong, but I think those three functions are actually different now that we don't include the time, start, and stop columns.

stdlib/universe/window_test.go
Bolladeen: Yup, good catch.

nathanielc (commented):
This looks great, thanks for making these changes.

I like the name change as well. 

Before we merge I want to have a plan for how we are going to release this and deprecate the other window function. We can chat tomorrow in our team meeting and if that is not enough time we can schedule a working group afterwards.

libflux/flux/benches/everything.flux
```diff
@@ -1047,6 +1047,16 @@ builtin table : (<-tables: [A], columns: C) => [B] where A: Record, B: Record, C
 //         The length of the interval. This defaults to the
 //         every duration.
 builtin window : (<-tables: [A], ?time: string, every: duration, ?period: duration, columns: C) => [B] where A: Record, B: Record, C: Record
+builtin implicitWindow : (
```
rockstar, line 1050: I don't know if we need to add this. We aren't really testing this interface at all. Maybe one day we'll just re-generate `everything.flux` in which case this will get included, but I don't think we should be manually adding things to `everything.flux`.

stdlib/universe/universe.flux
```diff
@@ -76,6 +76,16 @@ builtin tripleExponentialDerivative : (<-tables: [{B with _value: A}], n: int) =
 builtin union : (tables: [[A]]) => [A] where A: Record
 builtin unique : (<-tables: [A], ?column: string) => [A] where A: Record
 builtin window : (<-tables: [A], ?every: duration, ?period: duration, ?offset: duration, ?timeColumn: string, ?startColumn: string, ?stopColumn: string, ?createEmpty: bool) => [B] where A: Record, B: Record
+builtin implicitWindow : (
+    <-tables: [{T with _time: time}]
+    ?every: duration,
+    ?period: duration,
+    ?offset: duration,
+    ?createEmpty: bool
+) => [{A with
```
jlapacik, line 85: This should be `T` so that we can tie the input records to the output records

jlapacik (commented):
Just documenting the outcome of our conversation - we decided to require `_start` and `_stop` columns on the input to `window`. So the updated type signature becomes
```
builtin window: (
    <-tables: [{T with _start: time, _stop: time, _time: time}]
    every: duration,
    ?period: duration,
    ?offset: duration,
    ?createEmpty: bool
    ) => [{T with _start: time, _stop: time, _time: time}]
```

libflux/flux/benches/everything.flux
```diff
@@ -1047,6 +1047,7 @@ builtin table : (<-tables: [A], columns: C) => [B] where A: Record, B: Record, C
 //         The length of the interval. This defaults to the
 //         every duration.
 builtin window : (<-tables: [A], ?time: string, every: duration, ?period: duration, columns: C) => [B] where A: Record, B: Record, C: Record
+
```
nathanielc, line 1050: nit but can you revert this change?

stdlib/experimental/experimental.flux
```diff
@@ -39,3 +39,16 @@ alignTime = (tables=<-, alignTo=time(v: 0)) =>
       ({ r with _time: time(v: (int(v: alignTo ) + r.timeDiff)) })
     )
     |> drop(columns: ["timeDiff"])
+
+
+// An experimental version of window.
+builtin window : (
+  <-tables: [{T with _time: time}]
```
nathanielc, line 46: this should specify _start and _stop as per our chat earlier today.

stdlib/experimental/window.go
```diff
@@ -0,0 +1,14 @@
+package experimental
+
+import (
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/runtime"
+	"github.com/influxdata/flux/stdlib/universe"
+)
+
+const WindowKind = "window"
```
nathanielc, line 9: this constant isn't really needed now, maybe just use the string `"window"` directly below?

stdlib/universe/window.go
```diff
@@ -124,6 +124,92 @@ func createWindowOpSpec(args flux.Arguments, a *flux.Administration) (flux.Opera
 	return spec, nil
 }
 
+func CreateWindowOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
```
nathanielc, line 127: Why was this function added?

nathanielc (commented):
Some comments below

stdlib/universe/window.go
Bolladeen: That function was added because we need to export it to the experimental package and go requires an all uppercase function in order to export.

nathanielc: Can you export the existing function instead of duplicating the function?

stdlib/experimental/experimental.flux
nathanielc: This is still missing _start and _stop.

nathanielc (approved):
Looks great! Thanks for getting this all working!

Issue #1433, opened 2019-06-29
SQL function support for Oracle


gogangipark:
I know Flux implemented postgreSQL, MySQL.
In Enterprise, We need Oracle, MSSQL support

gogangipark:
Do you have any plan to implement oracle/mssql?

timhallinflux:
MS SQL is done and available. https://docs.influxdata.com/influxdb/v2.0/query-data/flux/sql/

I'm going to modify this to focus on Oracle.  We need to have some discussion about how to handle the Oracle driver through the build system.

nathanielc:
@timhallinflux Can you link us some details about the planned Oracle driver that Flux would need to use?

ivankudibal:
* https://github.com/golang/go/wiki/SQLDrivers lists these two candidates:
  * Oracle (uses cgo): https://github.com/godror/godror this C-Go is what we used first, the most active project... support of multiple oracle versions in the cloud can be disadvantage (client time zone file should match the server, otherwise errors with TIMESTAMP WITH TIME ZONE type may occur
(such as "ORA-01805"); or yet not analyzed method of Patching of Timestamp with Time Zone Data Type can be used...?
  * Oracle (pure go): https://github.com/sijms/go-ora has not been released yet, this is what we need to try, too, because the usage in cloud might be way easier

timhallinflux:
Cannot currently be done based on the exposed API/drivers.

PR #3485, opened 2021-02-09
feat(sql): add Oracle database support

alespour:
This PR adds Oracle database support to `sql` package.

It uses [godror](https://github.com/godror/godror) **c-go** driver and supports _Easy Connect_ string format (https://www.orafaq.com/wiki/EZCONNECT).
Examples:
  * `scott/tiger@sales-server:1521/ORCL`
  *  `scott/tiger@//75.184.3.22:1522/xepdb1`

[Oracle Instant Client](https://www.oracle.com/database/technologies/instant-client.html) is required at runtime. _Client version should match the server, or it should be ensured that timezone files match. Otherwise, errors such as "ORA-01805: possible error in date/time operation" may occur when working with columns of data types with time zone information._

Known issues and/or limitations:
  * `BOOLEAN` type does not exist in Oracle. Output mapping uses most(?) recommended approach (`CHAR(1)`).
  * `INTERVAL` types are scanned to `time.Duration`, but there is no corresponding `flux.ColType` (yet?), so these types are mapped converted to string. However, duration string representation in Flux is nothing like in Oracle, so an attempt to write such value directly to interval type column fails with _"ORA-01867: the interval is invalid"_.

### Done checklist
- [x] Test cases written

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/eclipse/paho.mqtt.golang v1.2.0
 	github.com/foxcpp/go-mockdns v0.0.0-20201212160233-ede2f9158d15
 	github.com/go-sql-driver/mysql v1.5.0
+	github.com/godror/godror v0.23.1
 	github.com/gofrs/uuid v3.3.0+incompatible
 	github.com/golang/geo v0.0.0-20190916061304-5b978397cfec
 	github.com/golang/snappy v0.0.0-20180518054509-2e65f85255db // indirect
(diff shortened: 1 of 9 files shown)


timhallinflux:
fixes #1433

wolffcm (changes requested):
Hi Ales, having a runtime dependency is a problem for us. We don't have anything like that currently and it would cause maintenance hassles, especially since we need to cross compile for different platforms. A quick check suggests that the source code is not available for Oracle Instant Client.

Would it be possible to use a pure Go solution like [go-ora](https://github.com/sijms/go-ora) instead?

I can see you've put a significant amount of work into this---sorry to throw a wrench in the works here. I really appreciate the consistent quality of your contributions.

alespour:
@wolffcm I tried `go-ora` driver as underlying driver, I ran into some specific data type issues (I recall timestamps where strangely shifted as if timezone was messed up). I did not invest more time to make it work - I can, gladly.
But what I'm worried more are legal issues. I know that the driver is listed in golang SQL drivers list. But how do these guys know the Oracle protocol, don't they do reverse engineering? Because what I've read, it is private and proprietary...

wolffcm:
There is no driver that will work with Flux here: one has a runtime dependency, and the other has issues with legality. Sadly there are no plans to support Oracle in Flux at this time.

PR #3487, opened 2021-02-10
fix(parser): detect+break infinite loop while parsing array

### LINKED ISSUE #3466: Attempt to recover from errors inside array expressions, while still making progress
In https://github.com/influxdata/flux/pull/3464 we break from the array expression parsing when a bad array entry is found. Prior to that PR we would push onward, however it led to a busy loop, resulting in 100% CPU and an OOM when a parse error was found. 

We should attempt to bring the error recovery back, but without introducing a busy loop. This can be achieved by monitoring progress made through the input. Doing so will break a number of test cases, which will need to be dealt with. https://github.com/influxdata/flux/pull/3464#issuecomment-771780954.

scbrickley:
Closes #3466 

Replaces the fix submitted in #3464 with one that uses a more appropriate exit condition and is less likely to hit false positives.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/core/src/parser/mod.rs
```diff
@@ -1612,14 +1617,15 @@ impl Parser {
                         }
                         _ => None,
                     };
-                    match expression {
-                        Expression::Bad(_) => {
-                            items.push(ArrayItem { expression, comma });
+                    items.push(ArrayItem { expression, comma });
+                    if let Some(tok) = &self.t {
```
adrian-thurston, line 1621: You should use `self.peek()` here, as it will ensure the next token is scanned. This is guaranteed to refer to the first token that has not been consumed by the parser. The `self.t` is simply a buffer var and should not be referenced outside of the `scan`, `peek`, etc functions.

libflux/core/src/parser/mod.rs
```diff
@@ -1597,12 +1600,14 @@ impl Parser {
...
             }
-            _ => {
+            t => {
                 let comma = self.expect(TokenType::Comma);
                 let mut items = vec![ArrayItem {
                     expression: init,
                     comma: self.make_comments(&comma),
                 }];
+                // keep track of the last token's byte offsets
+                let mut last = (t.start_offset, t.end_offset);
```
adrian-thurston, line 1610: It should be sufficient to use only `start_offset`. There shouldn't be a case where only `end_offset` has changed. You can just go `self.peek().start_offset`

adrian-thurston (changes requested):

libflux/core/src/parser/mod.rs
```diff
@@ -1597,12 +1600,14 @@ impl Parser {
...
             }
-            _ => {
+            t => {
                 let comma = self.expect(TokenType::Comma);
                 let mut items = vec![ArrayItem {
                     expression: init,
                     comma: self.make_comments(&comma),
                 }];
+                // keep track of the last token's byte offsets
+                let mut last = t.start_offset;
```
adrian-thurston, line 1610: Have to use `peek` here because the `expect` a few lines above can consume the token.

adrian-thurston (changes requested):

libflux/core/src/parser/mod.rs
scbrickley: Right, missed that one. One sec.

adrian-thurston (approved):
Looks good, handles `["a",1+, hello]` nicely!

PR #3488, opened 2021-02-10
feat(cmd/test): add extends capability to testcase block

### LINKED ISSUE #3447: Implement test extension AST transformation
As detailed in the design issue #3335, we need to modify the AST transformations for `testcase` to work like the following

    package strings_test
 
    testcase strings_trim {
        // strings trim test
    }

would then expand to:

    package main

    strings_trim = () => {
        // strings trim test
    }
    strings_trim()

The AST transformation would inject the call after all other transformations are made. A downstream consumer could then do the following:

    testcase special_string_trim {
        import "strings/strings_test"
        strings_test.string_trim()
    }

...could then be expanded to

    package main

    strings_test = {
        "string_trim": (ast for string trim),
        "other_variable": (ast for others),
    }
    special_string_trim = () {
        strings_test.string_trim()
    }
    special_string_trim()

This would also have to take into account `option` statements, as they'll have to be hoisted to the top scope the same way the `import` statements would be, each in its own context.

jsternberg:
A testcase can now be extended using a relative path to another file.
When used, the file will be read from the filesystem and inserted into
the testcase in a wrapped context. Any testcase blocks within that
import will be wrapped in a function so they can be invoked from the
current testcase.

This has certain restrictions. Recursive extension of testcases isn't
allowed. Options are allowed, but this may cause a conflict if the
option is specified multiple times so this is discouraged for testcases
that will need to be imported from another testcase.

Fixes #3447.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


ast/edit/testcase.go
```diff
@@ -1,60 +1,241 @@
...
+//
+//     import "testing/assert"
+//     math_test = () => {
+//         myVar = 4
+//         test_addition = () => {
+//             assert.equal(want: 2 + 2, got: myVar)
+//             return {}
+//         }
+//         return {myVar, test_addition}
+//     }()
```
nathanielc, line 46: Is this the generated AST being shown? If so what is the input AST that generates this output?

ast/edit/testcase.go
```diff
@@ -1,60 +1,241 @@
...
+//
+// Each import will be prepended to the list of files in the package as its own file.
+//
+// If a testcase includes an import statement, it will be replaced with the given body.
```
nathanielc, line 50: Is this only for recursive imports? Or all imports?

ast/edit/testcase_test.go
```diff
@@ -74,3 +81,134 @@ myVar = 4`
...
+import "testing/assert"
+want = 4
+testcase a {
+	assert.equal(want: want, got: 2 + 2)
+}
+`,
+			"b/b_test.flux": `package b_test
+import "testing/assert"
+testcase b {
+	import "../a/a_test"
```
nathanielc, line 98: What is the reasoning for relative import paths? How will this work in the context of IDPE importing a Flux test? Its not clear to me where on the filesystem a Flux test lives relative to an IDPE test.

libflux/core/src/formatter/mod.rs
```diff
@@ -750,10 +750,36 @@ impl Formatter {
...
+            for stmt in n.body.iter() {
+                self.write_indent();
+                self.format_node(&Node::from_stmt(stmt));
+                self.write_rune(sep);
+            }
+
+            self.unindent();
+            self.write_indent();
+        }
+        self.write_rune('}');
```
nathanielc, line 782: Can you add a test case for the formatting code here?

nathanielc (commented):
Looks really good. I had some questions below

ast/edit/testcase_test.go
```diff
@@ -74,3 +81,134 @@ myVar = 4`
...
+	if want, got := []string{"b"}, names; !cmp.Equal(want, got) {
+		t.Fatalf("unexpected testcase names -want/+got:\n%s", cmp.Diff(want, got))
+	}
+
+	files := make([]string, len(pkgs[0].Files))
+	for i, file := range pkgs[0].Files {
+		files[i] = ast.Format(file)
+	}
+
+	want := []string{
```
rockstar, line 131: Yesssss. Wondered if this test was going to assert on AST data structure or source and glad to see source as the "want." Makes the intent much more clear.

ast/edit/testcase_test.go
```diff
@@ -74,3 +81,134 @@ myVar = 4`
...
+
+a_test.a()
+assert.equal(want: 6, got: 3 + 3)`,
+	}
+	if got := files; !cmp.Equal(want, got) {
+		t.Fatalf("unexpected file contents -want/+got:\n%s", cmp.Diff(want, got))
+	}
+}
+
+type inmemFS struct {
```
rockstar, line 160: Could we expand this out to `inMemoryFilesystem`? I don't thing it buys us anything to abbreviate things here.

rockstar (approved):
This is a big patch, so it took a couple of passes to really get my head around some parts of it, and I'm still not entirely sure I got everything, but I'm happy with what's there, and don't see anything glaring to be concerned about. Happy to see this merged.

ast/edit/testcase.go
rockstar: This is only if you try and import a testcase that itself has a testcase. We discussed this, and while it opens up a can of worms to support it recursively, allowing deeply nested test cases is something we'd rather actively discourage as bad practice.

ast/edit/testcase.go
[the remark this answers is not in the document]
jsternberg: Declare and execute is correct. The advantage of this is there's very little transformation I have to do and closures continue to work. If we transformed it directly to a record, then functions that reference other functions or variables wouldn't be able to reference each other.

jsternberg: It's the generated AST in the case that the above test were imported.

ast/edit/testcase_test.go
jsternberg: The idea that I had was we didn't really have a concept of where the top level was. Tests just live in arbitrary places so we'd have to reference them relative to where they are.

I'm not sure how we'll consolidate the test cases from OSS and other sources. I had not thought of that. I'll try to have a better answer tomorrow.

libflux/core/src/formatter/mod.rs
jsternberg: Done.

ast/edit/testcase.go
jsternberg: It's all imports at the moment. In the normal transformation of a testcase, the import gets extracted to the top level because that's just what testcase transforms do. This puts the syntax in a context where it can import or specify options. My original thought was to translate every testcase as:

```
testcase foo {
    bar()
}

foo = () => {
    bar()
}
foo()
```

This would make the two transformations identical. But, it means that the testcase body can't support options since options cannot be specified outside of the top level. Since a primary reason for this feature is to set planner rules through options, that was a pretty hard requirement.

But, since imports have to be translated this way, we cannot fully support options or imports without making some guesses based on syntax that the AST level really doesn't know. The AST may be able to see `option now = <expr>`, but it cannot make any assumptions about what `now` is because only the interpreter knows what that value really is. The reason we need to translate these into functions is because we don't know which one will be invoked by the testcase when it is imported because that would require the support of the interpreter which we don't want to do at this moment.

My thoughts are that we will need to refactor testcases into a formal part of the interpreter or execution engine when we want to release this feature for external users. In addition to that, we will need to fully define how we intend the package system to be used and formally support user-defined imports.

I think we would be able to make some different choices in the design with those two features implemented.

For one, we wouldn't have to perform any translation of an import into a function and would not have to worry about potentially conflicting identifiers. The package scoping system would prevent that and options can be specified multiple times across different packages. But, options cannot be specified multiple times in the same package which limits us in this circumstance.

Second, if the execution engine/interpreter were executing the test cases, we would not have to perform AST translations which means that we would have more rich knowledge of what an identifier references rather than guessing what an identifier might be referencing. This point is more minor compared to the previous point about user-defined imports.

ast/edit/testcase_test.go
jsternberg: Yea, I get really tired of trying to type out AST data structures so I've gravitated to using the formatter more often. Pretty sure it's going to bite me in the rear whenever we need/want to make changes to the formatter output, but we'll just have to take that since this is a lot easier to read.

ast/edit/testcase_test.go
jsternberg: I'd appreciate some feedback on ways to consolidate the sources, but I thought it might be better to go more into detail about why I figured relative imports were better in this circumstance.

The reason why I figured relative imports were better were for relatively minor reasons, but enough that I figured it was bumpy.

1. The semantics of importing a file vs a package were different. Even if flux only supports one file packages at the moment, the stated semantics are that you import a path and it includes all of the non-test flux files in that directory as part of a single package. This mean that importing a single file didn't make much sense to me and seemed confusing.
2. In light of the above, it was unclear to me if the syntax should be to import `"strings"`, `"strings_test"`, or `"strings/strings_test"`. The first seemed to match the most with what import meant, but it also implied that you would want to import the entire package of test files. The second made no sense because while it referenced the package in what I viewed as a correct way, it also implied every test file and it didn't even point to a filesystem. The third was what was proposed, but it means that this specific import would reference a specific file rather than a package which seemed to confuse the meaning of import. If I were doing this from scratch, my preference would be for test files to have the same package as the main package and to import the main package and have it include the test files. But, I also figured this probably would have issues with scoping since every test uses `inData` as a global variable and those would conflict.
3. It seemed to me that since user-defined imports didn't exist, it would give the wrong impression if I added the concept of an import path. If we had a concept of an import path, then I would have used that to find the files.
4. Building on the lack of an import path, the test runner runs files that it finds and not packages. There's no current method to determine what the root directory is so when you are running the tests for a package, there's no way to know what the root directory is to descend from there. Go used GOPATH and now it uses `go.mod` as a way of defining what the root is. We don't have those so relative paths are really all I had without introducing these concepts.

With the above, it was my opinion that the above concepts were too much for this project and it would torpedo the project. This was a judgment and I think it might be good to evaluate that judgment. I think that might be helpful for when we want to expose this to external users to ask ourselves what we would want to change and make sure that we put a priority on the requirements for that project before we begin working on it.

I considered moving away from using `import` because I thought the semantics might be too confusing, but it was my opinion that adding a new keyword to the language would be worse. Thinking about it, another possibility is to go back to the old syntax and do something like:

```
testcase a with "b" {
}
```

The advantage of this is it would distinguish the importing mechanism completely from testcase so there wouldn't be any confusion about the mechanics. I still think we'd have to do relative paths though since we don't have a concept of a root directory.

jsternberg:
I've updated this to have the syntax:

```
testcase b extends "a_test" {
    // extended test case
}
```

nathanielc (approved):

rockstar (approved):

PR #3490, opened 2021-02-11
chore: execute/executetest: better error messages on failure

rogpeppe:
When `executetest.EqualResultIterators` fails, it's not clear exactly where
the error is, because the error doesn't make it clear how far through the
iteration the failure happens.

Rather than add a load of annotations, instead use a simpler implementation
that just reads all the data from both iterators and uses `cmp.Diff` to show any
differences.

This might not be appropriate for use in a running service (because it uses O(n)
memory), it should be fine in a testing scenario.


rockstar (approved):
> This might not be appropriate for use in a running service (because it uses O(n)
memory), it should be fine in a testing scenario.

I would agree with this sentiment. Thanks for working this out.

PR #3497, opened 2021-02-16
fix(quantile): Track and reduce memory used by tdigest

### LINKED ISSUE #3455: Quantile's use of tdigest can consume lots of untracked memory
Recently we saw that a query using `quantile` was consuming a large amount of memory. 

By default, `quantile` uses a library called `tdigest` to estimate its answer, with a parameter called `compression` exposed to the user that make an accuracy/memory tradeoff. The memory allocated by this library was not tracked using a `memory.Allocator`, so Flux was not able to detect that the query was hitting the memory allocation limit.

We need to track this memory somehow:
- We could make `tdigest` accept an allocator, since we maintain `tdigest`. However that might couple `tdigest` to Flux and we intend it to be more generic
- We could preemptively track memory within the `Process` method somehow, based on how much we think it might use. This might not be accurate though.
- Perhaps there is some other solution

DOD:
- Tdigest exposes an API for creating TDigest structs while delegating the large allocations to its caller. Something like `NewWithBuffers(a, b []Centroid)`. This will allow the Flux code to create a small wrapper around tdigest that can control the allocations and track them in the allocator.
- Update quantile implementation to a single instance of tdigest and use Reset for each call to Process. (Note make sure the `compression` setting is the same across calls)
- Benchmark to show allocations are significantly reduced

scbrickley:
Closes #3455 

~Requires https://github.com/influxdata/tdigest/pull/27 to merge first.~

- Uses `tdigest`'s new `ByteSizeForCompression` function to get the appropriate byte size to pass to the allocator.
- Moves creation of a digest from `QuantileAgg.Copy()` to `createQuantileTransformation` and simply resets the digest every time the aggregate is copied.
- Adds a NewQuantileAgg function for convenience in testing.

***Benchmarks before change:***
```
go test -bench=Quantile$ -run=^Benchmark -benchmem ./stdlib/universe
goos: linux
goarch: amd64
BenchmarkQuantile-24                   6         188348059 ns/op         1513738 B/op        137 allocs/op
PASS
ok      github.com/influxdata/flux/stdlib/universe      4.296s
```

***Benchmarks after change:***
```
go test -bench=Quantile$ -run=^Benchmark -benchmem ./stdlib/universe
goos: linux
goarch: amd64
BenchmarkQuantile-24                   6         184495440 ns/op           29280 B/op          8 allocs/op
PASS
ok      github.com/influxdata/flux/stdlib/universe      4.436s
```

### DIFF
--- a/stdlib/universe/quantile.go
+++ b/stdlib/universe/quantile.go
@@ -226,10 +226,7 @@ func createQuantileTransformation(id execute.DatasetID, mode execute.Accumulatio
 	return t, d, nil
 }
 
-// This function does not follow normal copy semantics, since it modifies
-// the original. For our purposes, this behavior is fine, but it should be
-// noted that `Copy` is a misnomer.
-func (a *QuantileAgg) Copy() *QuantileAgg {
+func (a *QuantileAgg) Recycle() *QuantileAgg {
 	na := new(QuantileAgg)
 	*na = *a
 	na.digest.Reset()
@@ -249,7 +246,7 @@ func (a *QuantileAgg) NewUIntAgg() execute.DoUIntAgg {
 }
 
 func (a *QuantileAgg) NewFloatAgg() execute.DoFloatAgg {
-	return a.Copy()
+	return a.Recycle()
 }
 
 func (a *QuantileAgg) NewStringAgg() execute.DoStringAgg {
--- a/stdlib/universe/quantile.go
+++ b/stdlib/universe/quantile.go
@@ -204,6 +204,14 @@ type QuantileAgg struct {
 	ok     bool
 }
 
+func NewQuantileAgg(q, comp float64) *QuantileAgg {
+	return &QuantileAgg{
+		Quantile:    q,
+		Compression: comp,
+		digest:      tdigest.NewWithCompression(comp),
+	}
+}
+
 func createQuantileTransformation(id execute.DatasetID, mode execute.AccumulationMode, spec plan.ProcedureSpec, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
 	ps, ok := spec.(*TDigestQuantileProcedureSpec)
 	if !ok {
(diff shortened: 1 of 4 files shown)


wolffcm (commented):
I think you need to update to the latest hash of `tdigest` for this to compile:
```sh
go get github.com/influxdata/tdigest
go mod tidy
```
That should do update `go.mod` to use your latest changes.

stdlib/universe/quantile.go
```diff
@@ -212,14 +220,17 @@ func createQuantileTransformation(id execute.DatasetID, mode execute.Accumulatio
...
+	_ = a.Allocator().Account(tdigest.ByteSizeForCompression(agg.Compression))
 	t, d := execute.NewAggregateTransformationAndDataset(id, mode, agg, ps.AggregateConfig, a.Allocator())
 	return t, d, nil
 }
+
 func (a *QuantileAgg) Copy() *QuantileAgg {
 	na := new(QuantileAgg)
 	*na = *a
-	na.digest = tdigest.NewWithCompression(na.Compression)
+	na.digest.Reset()
```
wolffcm, line 232: I think it's probably fine, but this `Copy` method will change state of the original thing being copied, in order to avoid doing a deep copy of `a`. It seems like `Copy` is really a misnomer. That is not something that I would expect you to fix here; but it would be good to call out in a comment that this method does not really have copy semantics as one might reasonably expect.

wolffcm: Also, it would be good to really be sure that instances produced by `QuantileAgg.Copy` are only used serially, since they will all share the same `tdigest` data structure. I believe that this is the case, since the aggregate `Process` method has only one input and will only process one table at a time.

wolffcm (approved):

stdlib/universe/quantile.go
scbrickley: Did a global search in flux, and the only place `NewFloatAgg` (which wraps `Copy`) is [here](https://github.com/influxdata/flux/blob/2c5db92719a994e45de0c9026e6fe638a1d2f1b3/execute/aggregate.go#L108). I think as long as it stays that way, this implementation of Copy should be fine.

But maybe it would be worthwhile to rename it?

wolffcm: If renaming it isn't too troublesome, then I think it's a good idea. If not, then it's okay to skip it.

scbrickley: Done.

PR #3500, opened 2021-02-17
fix: switch from HashMap to BTreeMap in ast/semantic conversion functions

adrian-thurston:
Using HashMap results in a prelude.data and stdlib.data that varies from build to
build, given the same source. This breaks reproducible builds and adds churn to
the deployments process. The variation is caused by iterating the HashMap,
which doesn't guarantee any particular order.

Helps with influxdata/idpe#6331

### DIFF
--- a/libflux/core/src/semantic/convert.rs
+++ b/libflux/core/src/semantic/convert.rs
@@ -5,7 +5,7 @@ use crate::semantic::types;
 use crate::semantic::types::MonoType;
 use crate::semantic::types::MonoTypeMap;
 use crate::semantic::types::SemanticMap;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::result;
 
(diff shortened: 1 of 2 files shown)


fchikwekwe:
This change totally makes sense. Are we concerned about the b tree slowing things down compared to the hashmap?

nathanielc (approved):

fchikwekwe (approved):
Actually, it looks like rust's hashmap algorithm is a little slower than the typical hashing algorithm. In either case, it likely doesnt matter much here.

adrian-thurston:
The btree will be O(log n), but probably for fairly small `n`. It can impact the big-O of the overall algorithm, but in practice it will likely be a small factor on the performance. But we do need to have the stability so we either need the btree, or to stabilize the set by sorting on the keys before iterating. Probably better to just use the btree.

PR #3504, opened 2021-02-17
feat(stdlib): added new Integral function signature

Bolladeen:
This PR adds the Integral function to the experimental package and specifies the timeColumn and value Column on the input table rather than as another function parameter. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


adrian-thurston (approved):
Looks good to me.

stdlib/experimental/experimental.flux
```diff
@@ -49,3 +49,6 @@ builtin window : (
   ?offset: duration,
   ?createEmpty: bool
 ) => [{T with _start: time, _stop: time, _time: time}]
+
+// An experimental version of integral.
+builtin integral : (<-tables: [{T with _time: time, _value: B}], ?unit: duration, ?interpolate: string) => [{T with _time: time, _value: B}]
```
nathanielc, line 54: I'd like to start a discussion around the `_value` column. (The _time columns looks great).

All of our aggregate functions accept a `column` parameter which names the column that should be aggregated. To change those functions we would want to change them to expect `_value` hard coded instead of letting it be dynamic. 

This opens a few questions for me. 

* What about use cases using `fieldsAsCols` do they just not work anymore? 
* Is there a different way we could handle the `_value` column? Maybe this is the time to consider the `label` type?

Thoughts?

nathanielc (commented):
LGTM modulo one discussion question below. If the discussion ends up being long, then lets change this PR to only change the `timeColumn` parameter and leave the `column` parameter.

stdlib/experimental/experimental.flux
jlapacik, line 54: Since `integral` is an aggregate function, should it preserve the `_time` column? Or should it remove `_time`. The current behavior is to remove `_time`.

jlapacik: How does `fieldsAsCols` break with this change?

nathanielc: This use case (which is quite common) breaks:

```
from()
   |> range()
   |> filter()
   |> schema.fieldsAsCols()
   |> mean(column:"usage_idle")
```

After this change the user has to do


```
from()
   |> range()
   |> filter()
   |> schema.fieldsAsCols()
   |> rename(column: "usage_idle", as:"_value")
   |> mean(column:"usage_idle")
```

Generally we expect that `rename` will be more frequent with these kinds of changes but feels like it gets in the way a lot this way.

Basically I want to start a discussion around the ergonomics of using `_value` with aggregates contrasted with naming columns after the fields (i.e. pivoted data).

jlapacik: Personally I like the ergonomics. I don't know that we've formally specified how a label type would work in this case. My understanding is that we would need some type of macro facility in order to get around explicit renaming.

Bolladeen: Good catch, I think defaulting to the current behavior is probably the right move.

nathanielc: After having a chat with @Bolladeen I recommend we move forward with this PR as is. 

My intent above was to have a conversation and document the decision process. Here are my thoughts after our chat:

* Requiring _value for aggregates is the common happy path. 
* Users using fieldsAsCols almost always are after more than one field, in which case the aggregate method are not as useful since they operate on a single field.  Using `map` to combine multiple fields(columns) into a single column would be expected here and provides an easy way to name the resulting column `_value` without needing an explicit `rename` operation. 
* The point of this change is to make the expectations of these functions clear, i.e. what data do I need to call it and what data do I get back. Adding a label type goes against this idea as it preserves the indirection of what columns a function expects (even though it can in theory do so in a type safe manner). To me this says we don't want to move in the direction of label types because we have seen that the indirection they use is confusing to users.

I do have one remaining question but it doesn't need an answer right now:

* How should describe aggregates across multiple columns?

nathanielc (approved):

Issue #2256, opened 2019-12-06
Functions that drop value columns cannot be typed well


wolffcm:
Several functions (notably the aggregate functions) will drop some value columns from their input table when producing their output tables.

This makes it hard to relate the input type of a function to its output type.

We should fix this somehow, perhaps by modeling the group key column specifically for table stream functions, e.g., suppose we modeled the type of `sum` like this:
```
forall [t0, t1, t2] where t0: Row, t1: Numeric, t2: Row, (
    <-tables: [{keyColumns: t0 | _value: t1 | t2}]
) -> [{keyColumns: t0 | _value: t1}]
```

nathanielc:
Another possible solution to this problem is to not drop the columns at all rather let them take on a specific value. This was discussed a bit here https://github.com/influxdata/flux/pull/3530#issuecomment-790936788

I see a few options:

1. Leave the columns but set their value to `null`.
2. Leave the columns and let the implementation chose one of the values from the aggregated rows.

I think I prefer option 1 as it has less chance of being confusing to users since the data will always be null, i.e. representing an unknown value. If you pick a value as in option 2 it might mislead users as being a reasonable value to pick.

PR #3530, opened 2021-03-04
feat(stdlib): added new Aggregate Function signatures

Bolladeen:
Many of the aggregate functions can have better function signatures as shown in the universe.flux file. The main goal for this issue is to create a relationship between the input and output tables as shown in this example from the already updated Integral function.

Original:
`builtin integral : (<-tables: [A], ?unit: duration, ?timeColumn: string, ?column: string, ?interpolate: string) => [B] where A: Record, B: Record
`
New:
`builtin integral : (<-tables: [{T with _time: time, _value: B}], ?unit: duration, ?interpolate: string) => [{T with _value: B}]
`
The following functions can drop a column parameter and set it as default while also seeing meaningful changes to the relationship between the in and out tables.

- Count
- histogramQuantile
- Mean
- Mode
- Quantile
- Skew
- Spread
- Stddev
- Sum

Edit: These are additive changes and all new functions will be added to the experimental package.
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


Bolladeen:
The following are the new function signatures for each of the above functions:

Count:
`builtin count : (<-tables: [{T with _value: A}]) => [{T with _value: int}]
`
histogramQuantile:
`builtin histogramQuantile : (<-tables: [{T with _value: float, le: float}], ?quantile: float, ?minValue: float) => [{T with _value: float}]
`
Mean:
`builtin mean : (<-tables: [{T with _value: float}]) => [{T with _value: float}]
`
Mode:
`builtin mode : (<-tables: [{T with _value: A}]) => [{T with _value: A}]
`
Quantile:
`[builtin quantile : (<-tables: [{T with _value: float}], q: float, ?compression: float, ?method: string) => [{T with _value: float}]
`
Skew:
`builtin skew : (<-tables: [{T with _value: float}]) => [{T with _value: float}]
`
Spread:
`builtin spread : (<-tables:  [{T with _value: A}]) =>  [{T with _value: A}] where A: Numeric
`
Stddev:
`builtin stddev : (<-tables: [{T with _value: float}], ?mode: string) => [{T with _value: float}]
`
Sum:
`builtin sum : (<-tables: [{T with _value: A}]) => [{T with _value: A}] where A: Numeric
`

philomory:
Am I reading this correctly that functions like `count` will no longer work without a `_value` column?

Bolladeen:
> Am I reading this correctly that functions like `count` will no longer work without a `_value` column? E.g, the following code will no longer function as-is?
> 
> ```
> import "influxdata/influxdb/schema"
> 
> from(bucket: "example")
>   |> range(start: -10m)
>   |> filter(fn: (r) =>r._measurement == "example")
>   |> filter(fn: (r) => r._field == "foo" or r._field == "bar")
>   |> schema.fieldsAsCols()
>   |> filter(fn: (r) => r.foo > r.bar)
>   |> count(column: "foo")
> ```
> 
> (Obviously I could just throw in a `rename(columns: {foo: "_value"})`, a `set(key: "_value", value: "ok")`, or use `reduce()` instead of filter-then-count; I'm just trying to get clarity on the proposed changes).

@philomory These new functions (and their signatures) are being added to the experimental package. So for count, the universe package version remains the same, but the new behavior is being added to the count in the experimental package. Let me know if that makes sense.

jlapacik (approved):
These new signatures are great!

philomory:
@Bolladeen Sorry, I understood that these were going in the experimental package, but, presumably they are being trialed there in preparation for becoming the "official" signatures for the corresponding functions in the universe package at some point in the future, yes?

nathanielc (commented):
This change looks great. There is one open question that I have:

What do we do about the group key?

For example counts new signature: `count : (<-tables: [{T with _value: A}]) => [{T with _value: int}]`

This states that the records in are the same as the records out with exception of the value column. But if the input records have columns host, region, _value and the group key is just region, then the output records will have columns region, _value.

This challenge aside the change is great. We need to decide how we want to deal with the group key in these situations

nathanielc:
@philomory Yes that is the idea. We are experimenting with this new signature to become the new "official" way of doing these kinds of transformation functions. 

Our reasoning is that the clarity it gives to what the shape of the data looks like on input and output outweighs the need to rename the columns when that need arises. 

Curious about your thoughts on the change?

philomory:
I guess I don't see the point of it? Like, it seems plenty clear to me when you have the arguments as-is and just have them default to what their hard-coded values would be in the new signatures.

More importantly, I don't think most actual users ever even _see_ a signature like `count : (<-tables: [{T with _value: A}]) => [{T with _value: int}]`; they see this:

image
image

Whereas with the new signatures they'll see something along the lines of:

> `count()`
> NOTE: In order for this function to operate you must have a column in your table called `_value`; if your table does not have a `_value` column, you must rename one of the existing columns to `_value` before passing it to this function; see `rename()` for details.

Mind, the `count()` function is a bit of an odd case; arguably the `count()` function shouldn't care about columns _at all_, it should just count the number of records in each table in the stream without respect to what the columns in those tables are; that's effectively what it does anyway since it doesn't care whether what data is in the selected column or even whether it's just a null; it just balks if the column doesn't exist. But, leaving `count()` aside more-or-less the same argument applies to `sum()` I think.

nathanielc (approved):
Had a chat with @Bolladeen there are still two open questions:

1. How do we handle the group key? For now we will leave the type signature as they are in this PR. The effect being that those columns that get aggregated away are just null. See https://github.com/influxdata/flux/issues/2256
2. Do we like the change since it requires rename more often?

We think its best to try and answer these questions by seeing how these functions can be used in real world examples. 

The plan is to merge these into experimental and answer both of these questions before promoting any of these changes out of experimental.

jlapacik:
@nathanielc thanks for bringing this up. My preference is that we would no longer aggregate columns away. Obviously this would mean that we wouldn't be able to just use the existing implementations. But I think the type signature and the implementation should definitely be consistent.

PR #3531, opened 2021-03-04
Bugfix/fix slack payload for new tokens

philomory:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written
- [ ] Sign [CLA](https://www.influxdata.com/legal/cla/) (if not already signed)

This PR removes the `as_user` param from the JSON payload of slack messages. For messages sent using the `chat.postMessage` API, sent with auth tokens owned by "new" Slack apps, the `as_user` param must not be present in the payload at all or the message will be rejected. For older apps, or messages sent using a webhook rather than `chat.postMessage`, the `as_user` param is permitted; however, previously we hard-coded a value of `false` for this parameter, which is the default value of the parameter if omitted anyway, so there is no harm in omitting it in all cases.


philomory:
It is not at all clear to me what changes I'm expected to make to `docs/SPEC.md` as part of this PR.

stdlib/slack/slack_test.go
```diff
@@ -67,8 +68,16 @@ func NewServer(t *testing.T) *Server {
...
+		b, err := ioutil.ReadAll(r.Body)
+		defer r.Body.Close()
+    if err != nil {
+			t.Error(err)
+		}
+		err = json.Unmarshal(b,&sr.PostData)
+		if err != nil {
+			t.Error(err)
+		}
+		err = json.Unmarshal(b,&sr.RawData)
```
philomory, line 80: This is necessary because I need the test to differentiate between JSON with containing `"as_user":null` and JSON that does not contain an `as_user` field whatsoever; the former will be rejected by Slack's API.

Unfortunately, as far as I am aware, there is no way to distinguish those two JSON documents once the JSON has been decoded into a Go struct; so, for the extra test, I need it decoded as a `map[string]interface{}`, where the distinction between a key being absent vs being null is preserved.

go.mod
```diff
@@ -29,7 +29,7 @@ require (
 	github.com/google/uuid v1.1.1 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e
-	github.com/influxdata/pkg-config v0.2.5
+	github.com/influxdata/pkg-config v0.2.6
```
rockstar, line 32: Could you revert this change in your patch? It's probably one we should make, but it's unrelated to change needed here.

go.sum
```diff
@@ -166,8 +166,8 @@ github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NH
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e h1:/o3vQtpWJhvnIbXley4/jwzzqNeigJK9z+LZcJZ9zfM=
 github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e/go.mod h1:4kt73NQhadE3daL3WhR5EJ/J2ocX0PZzwxQ0gXJ7oFE=
-github.com/influxdata/pkg-config v0.2.5 h1:iC19aXlkUPiwxjxeeKk8TT8S5s3pargNPLgZE/rvOzc=
-github.com/influxdata/pkg-config v0.2.5/go.mod h1:EMS7Ll0S4qkzDk53XS3Z72/egBsPInt+BeRxb0WeSwk=
+github.com/influxdata/pkg-config v0.2.6 h1:GQFKw3m3OmmPMze9n75ZVVtNu4LJ2MJolHbxvg4AAvg=
+github.com/influxdata/pkg-config v0.2.6/go.mod h1:EMS7Ll0S4qkzDk53XS3Z72/egBsPInt+BeRxb0WeSwk=
```
rockstar, line 170: Same with this change: it shouldn't be a part of this patch.

stdlib/slack/slack_test.go
rockstar: Could you summarize that as a comment in the patch itself? Maybe it's not a huge deal, but I'll definitely forget that, and won't quite understand why we're testing what we're testing.

stdlib/slack/slack_test.go
```diff
@@ -232,6 +242,9 @@ data = "
 			if req.PostData.Attachments[0].Color != tc.color {
 				t.Errorf("got color %s, expected %s", req.PostData.Attachments[0].Color, tc.color)
 			}
+			if _, hasKey := req.RawData["as_user"]; hasKey {
```
rockstar, line 245: I would suggest a comment here that this part of the test can be removed one year from today.

rockstar (changes requested):
Thanks for doing this work. It seems a little odd to assert something that has been removed, but the test and the implementation make me comfortable that we can change it easily. I would like to see some documentation in the form of comments that this assertion is temporary and once people stop using that functionality, we can remove the assertion. Other than that, I'm happy to see this patch merged.

stdlib/slack/slack_test.go
philomory: Sure, I can do that

go.mod
philomory: If I remove this change, then the test suite fails to pass in CircleCI; or at least, it did last time. I can undo it and see if it still works.

stdlib/slack/slack_test.go
philomory: I can add that comment, though I should note that this isn't quite the same as just e.g. asserting some random nonsense parameter wasn't included; you can include arbitrary nonsense in your requests and Slack will just ignore it. `as_user` is a specific documented parameter of their API that, if included in your requests, will cause the request to fail (assuming you aren't using certain legacy auth flows).

go.mod
philomory: OK, no idea what was going on previously; tests passed this time with 0.2.5 no problem.

stdlib/slack/slack_test.go
philomory: I added the requested comment, though I added it to the Struct declaration where I define the `RawData` map, rather htan the lines where I populate it. I can move it if you'd prefer it go elsewhere.

rockstar (approved):
Thanks for making the requested changes, and for this patch in general. I suspect it'll help a lot of users going forward.

PR #3532, opened 2021-03-04
fix(csv): add tests and fix the NoHeader config for the csv decoder

nathanielc:
Previously the NoHeader setting for the csv.ResultDecoderConfig had no
tests and did not work. This fixes it and adds tests.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


adrian-thurston (approved):
Good catch!

PR #3534, opened 2021-03-05
feat(csv): add support for parsing csv files without annotations

nathanielc:
This change allows for the parsing of csv files without annotations.
The parsed data produces a single table where all data is of type
"string". All rows are considered to be part of the same group.

Additionally support is added to skip the UTF8 bom

Open questions:

Do we like the `mode` parameter to the csv function? Or should we do something else?

Do we want to leave LazyQuotes enabled for all parsing or explicitly enable that as part of the `raw` mode? LazyQuotes makes Go's csv parser handles quotes in a more forgiving way.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


csv/result.go
```diff
@@ -71,6 +71,10 @@ func NewResultDecoder(c ResultDecoderConfig) *ResultDecoder {
 
 // ResultDecoderConfig are options that can be specified on the ResultDecoder.
 type ResultDecoderConfig struct {
+	// NoAnnotations indicates that the CSV data will not have annotations rows.
+	// Without annotations the decoder will assume every column is of type string and
+	// that all data is in a single table in a single result.
+	NoAnnotations bool
```
wolffcm, line 77: What will the column names be though? I guess I'll see as I read on.

csv/result.go
```diff
@@ -306,17 +323,36 @@ func (sfe *serializedFluxError) Error() string {
...
-			line = extraLine
-			extraLine = nil
-		} else {
-			l, err := r.Read()
+	if c.NoAnnotations {
+		// No annotations means that we are going to treat all rows as part of the same result with exactly one table
+		resultID = "_result"
+		tableID = "0"
+		recordStartIdx = 0
+		line, err := r.Read()
```
wolffcm, line 336: If the lines have differing numbers of fields, do we produce an error, or pad with nulls? I guess I will see as I read on.

csv/result.go
```diff
@@ -1271,3 +1305,25 @@ func NewMultiResultEncoder(c ResultEncoderConfig) flux.MultiResultEncoder {
 		Encoder:   NewResultEncoder(c),
 	}
 }
+
+// bufferedCSVReader allows for unreading a single line of the csv data
+type bufferedCSVReader struct {
```
wolffcm, line 1310: I guess you're using this instead of the `extraLine` that was here before? I like it, this simplifies the logic a lot.

csv/result.go
```diff
@@ -1271,3 +1305,25 @@ func NewMultiResultEncoder(c ResultEncoderConfig) flux.MultiResultEncoder {
...
+		b.line = nil
+		return line, nil
+	}
+	return b.r.Read()
+}
+
+// Unread places the provided line back on the buffer.
+// It is invalid to call unread multiple times without calling Read inbetween.
+func (b *bufferedCSVReader) Unread(line []string) {
+	b.line = line
```
wolffcm, line 1328: I wonder if there should be any guards here to avoid the perception that clients could unread several lines and expect them all to be buffered. Perhaps a second call to `Unread` when `b.line` is already set should produce an error?

csv/result.go
wolffcm: Ah, ok. The column labels are not actually an annotation. They are still required.

csv/result.go
wolffcm: Thinking it over, since we still have a line for the column labels, that sets the number of columns, same as if we had annotations.

stdlib/csv/from.go
```diff
@@ -22,8 +22,14 @@ const FromCSVKind = "fromCSV"
 type FromCSVOpSpec struct {
 	CSV  string `json:"csv"`
 	File string `json:"file"`
+	Mode string `json:"mode"`
 }
 
+const (
+	annotationMode = "annotations"
+	rawMode        = "raw"
```
wolffcm, line 30: I guess I like the name `mode` okay. I can't think of anything better.

stdlib/csv/from.go
```diff
@@ -150,16 +170,31 @@ func (c *CSVSource) Run(ctx context.Context) {
...
 		var data io.ReadCloser
 		data, err = c.getDataStream()
 		if err != nil {
 			goto FINISH
 		}
-		results, decodeErr := decoder.Decode(data)
+		// Many applications will add a UTF BOM (byte order mark) to the beginning of csv files
+		// We expect UTF8 encoded data so the byte order does not matter.
+		// Therefore we skip the BOM if it exists.
+		// See http://www.unicode.org/faq/utf_bom.html#BOM
```
wolffcm, line 191: Ah, ok. I did not know about the BOM, or that UTF8 always has the same byte order.

wolffcm (approved):
Ok, I guess this change is about letting us work with arbitrary CSV files in Flux, perhaps ones generated by some third party source. This certainly seems useful.

Instead of requiring clients to declare the presence of annotations up front, we could instead infer whether or not there are annotations and proceed accordingly. Would that be possible, and if so, would it be desirable?

Thinking it over, I think we want to leave annotations required by default, since it's defined in our docs and is our primary serialization format for query results.

Looks like a nice improvement! Let's make sure to get it documented when this merges.

wolffcm:
As far as lazy quotes, I would be in favor of leaving it on anytime we deserialize from CSV. I can't think of any case where we would want it off.

csv/result.go
nathanielc: Yep, it replaces the need for `extraLine` logic, this struct was added to make it simpler.

csv/result.go
nathanielc: I like that idea, i'll update to return and error and we can check for it.

stdlib/csv/csv_test.flux
```diff
@@ -0,0 +1,54 @@
...
+            {time: 2021-03-12T13:58:59Z, float: 42.69, int: -67, uint: uint(v: 152), bool: false, string: "hello world"},
+            {time: 2021-03-12T13:58:59Z, float: 42.69, int: -67, uint: uint(v: 152), bool: false, string: "hello world"},
+        ],
+    )
+
+    // Using annotations mode so all columns can be specific types
+    result = csv.from(csv: input, mode: "annotations")
+
+    testing.diff(got: result, want: want)
+}
```
nathanielc, line 54: @rockstar  Could you review this test case to make sure I did it right?

nathanielc: Also are these tests part of Flux CI? We created an issue for influxdb and IDPE but did we do it for Flux?

nathanielc:
@wolffcm I have made the changes requested and fixed the bugs/tests. Can you give this another review?

stdlib/csv/flux_test_gen.go
```diff
@@ -0,0 +1,15 @@
...
+
+var FluxTestPackages = []*ast.Package{&ast.Package{
+	BaseNode: ast.BaseNode{
+		Errors: nil,
+		Loc:    nil,
+	},
+	Files:   []*ast.File{},
+	Package: "csv_test",
+	Path:    "csv",
+}}
```
wolffcm, line 15: Why do we have this here? I'm not concerned about it really, just curious.

stdlib/csv/csv_test.flux
wolffcm: I am really liking the new test format.

csv/result_test.go
```diff
@@ -119,6 +119,160 @@ func TestResultDecoder(t *testing.T) {
 				}},
 			},
 		},
+		{
+			name: "single table small buffer",
+			decoderConfig: csv.ResultDecoderConfig{
+				MaxBufferCount: 2,
```
wolffcm, line 125: This is the test case that would have reproduced the bug you fixed, I guess.

wolffcm (approved):
Looks good to me.

stdlib/csv/csv_test.flux
```diff
@@ -0,0 +1,54 @@
+package csv_test
```
rockstar, line 1: Does this duplicate all the tests from the csv package? Can we delete the old ones?

stdlib/csv/csv_test.flux
```diff
@@ -0,0 +1,54 @@
...
+
+
+import "array"
+import "testing"
+import "csv"
+
+testcase from_raw{
```
rockstar, line 8: I'm breaking a very hard rule here about never commenting on whitespace in reviews, but only as an observation because we maintain the formatter. I think there should be a space here. Did this go through a formatter, and it formatted this way, or are we not checking format on flux in CI? I don't think we should change this, but rather file an issue to get lack of flux formatting caught at the CI level. This can be the indicator that it's done.

rockstar (approved):
Yay for new tests. I'm happy to see trees sprouting up where I was envisioning a forest. Thanks for being the guinea pig.

stdlib/csv/csv_test.flux
nathanielc: There were no tests previously in the csv package. These are net new tests. I think we were relying on the fact that almost all tests use csv parsing to work to cover the csv package. Now we have explicit tests

stdlib/csv/csv_test.flux
nathanielc: This did go through the formatter (using the LSP). I'll update to make sure I have the newest version. I agree this needs a space. I'll submit a PR to the formatter if it does not add a space.

nathanielc: Updated my LSP and it now adds a space so all good.

stdlib/csv/csv_test.flux
rockstar: "These net new tests" is the kind of patch I like to review.

rockstar (approved):

Issue #3470, opened 2021-02-03
"Option not found" error classified as internal


gavincabbage:
We're seeing occasional `Option not found` errors when users update tasks and remove the `option task = {...}` block from the script. This error comes from [here](https://github.com/influxdata/flux/blob/master/ast/edit/task_editor.go#L9-L28) where it is classified as an internal error. Since this is a task syntax error, and generally caused by user error, the returned error should be classified as invalid, rather than internal. This will prevent it from being reported as a system degradation by internal monitoring tools. 

System: Cloud 2
Flux Version: `github.com/influxdata/flux v0.103.1-0.20210202003033-17007b9a6de4`

wolffcm:
@gavincabbage I agree that this is a misclassified error, thanks for filing this.

I would think you would want to produce a more helpful error to the tasks user, so I wonder if it would be helpful for us to instead of just returning a generic `invalid` error, to return something to make it easy to robustly detect an `option not found` condition?

```Go
var OptionNotFoundError = &flux.Error{Code: codes.Invalid, Msg: "option not found"}

func GetOption(...) (ast.Expression, error) {
  ...
  return nil, OptionNotFoundError
}
```
Then in client code you could `if err == edit.OptionNotFoundError ...` to handle the error gracefully.

gavincabbage:
Sounds great @wolffcm, I had a similar thought! We can definitely teach the task API to do that if Flux exposes a sentinel error.

skartikey:
Fixed with PR #3539

PR #3539, opened 2021-03-09
"Option not found" error classified as internal #3470

skartikey:
Issue : We're seeing occasional Option not found errors when users update tasks and remove the option task = {...} block from the script.
Since this is a task syntax error, and generally caused by user error, the returned error should be classified as invalid, rather than internal. This will prevent it from being reported as a system degradation by internal monitoring tools.

Fix : Instead of just returning a generic invalid error, return variable 'OptionNotFoundError' to robustly detect an option not found condition to handle the error gracefully.

Test : Added test to check the return OptionNotFoundError variable.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


wolffcm (approved):
Looks good!

PR #3549, opened 2021-03-12
feat(libflux): support formatting the AST from libflux

### LINKED ISSUE #3329: Expose Rust Format function in flux.h
We need to expose the Rust Format method via flux.h so the Go code can get access to that code path.

The format method should accept a Rust AST, which can be acquired either by using the Flux parser or the JSON parse to deserialize the JSON AST into a Rust AST.

The function should fill a flux buffer type with the formatted string contents.

DOD:

  - A new function/set of functions exists for formatting flux in flux.h
  - A new, documented and tested pure rust API exists in the flux crate for handling the formatting
  - A valgrind test exists for the new ffi exports

jsternberg:
Fixes #3329.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


wolffcm (approved):
Looks good!

PR #3555, opened 2021-03-16
test: add flux acceptance tests to ci

rockstar:


Makefile
```diff
@@ -102,14 +103,17 @@ staticcheck:
...
+test: test-go test-rust test-flux
 
 test-go: libflux-go
 	$(GO_TEST) $(GO_TEST_FLAGS) ./...
 
 test-rust:
 	cd libflux && $(CARGO) test $(CARGO_ARGS) && $(CARGO) clippy $(CARGO_ARGS) -- -Dclippy::all
 
+test-flux:
+	$(GO_RUN) ./cmd/flux test
```
nathanielc, line 115: Do we want a `-v` here so we can see which tests failed in the output?

rockstar: Good idea.

nathanielc (approved):

PR #3556, opened 2021-03-17
feat(libflux): added support for all basic types in string interpolation

Bolladeen:
The following PR allows for all basic type (Bool, Int, Uint, Float, Time, Duration) to be acceptable string interpolation parts in the flux language. 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


compiler/compiler.go
```diff
@@ -546,9 +546,14 @@ func compile(n semantic.Node, subst map[uint64]semantic.MonoType, scope Scope) (
 		if err != nil {
 			return nil, err
 		}
-		return &interpolatedEvaluator{
+		val := &interpolatedEvaluator{
```
rockstar, line 549: Is `val` an `evaluator` or a `value`? Is there a better, more detailed name we could give this variable?

libflux/core/src/semantic/nodes.rs
```diff
@@ -608,11 +608,12 @@ impl StringExpr {
...
                 let (e, cons) = ip.expression.infer(env, f)?;
                 constraints.append(&mut Vec::from(cons));
-                constraints.push(Constraint::Equal {
-                    exp: MonoType::String,
+                constraints.push(Constraint::Kind {
+                    exp: Kind::Stringable,
                     act: ip.expression.type_of(),
                     loc: ip.expression.loc().clone(),
                 });
+                //TODO: add Kind::Stringable constraint
```
rockstar, line 616: Is there an open issue for this?

rockstar (approved):
This looks good, thanks for doing it. One seemingly minor change requested, but happy to see this merge.

libflux/core/src/semantic/nodes.rs
Bolladeen: Good catch, it was a mental note I was making and should be deleted.

compiler/compiler_test.go
```diff
@@ -44,12 +44,12 @@ func TestCompileAndEval(t *testing.T) {
 			fn:   `(r) => "n = ${r.n}"`,
```
nathanielc, line 44: Can you add a test case that shows interpolation working for other Stringable types besides string?

docs/SPEC.md
```diff
@@ -325,6 +325,15 @@ Interpolation example:
 
 [IMPL#1775](https://github.com/influxdata/flux/issues/1775) Interpolate arbitrary expressions in string literals
 
+String interpolation now supports all basic type arbitrary expressions  without requiring casting to a string.
```
nathanielc, line 328: nitpit, but can you remove the `now` wording? We want the SPEC to read in the present tense as people reading the SPEC won't necessarily know or remember the history of Flux. 

In the SPEC we list the type classes can you add Stringable to that list?

interpreter/interpreter.go
```diff
@@ -496,6 +496,11 @@ func (itrp *Interpreter) doStringPart(ctx context.Context, part semantic.StringE
 			return nil, errors.Newf(codes.Invalid, "%s: interpolated expression produced a null value",
 				p.Location())
 		} else if v.Type().Nature() != semantic.String {
+			val := values.NewString(fmt.Sprintf("%v", values.Unwrap(v)))
```
nathanielc, line 499: Do we want to rely on Go's formatting of the value here? Or should we be more explicit about how each of the Stringable types is converted to a string? For example using `strconv.FormatInt` etc?

interpreter/interpreter_test.go
```diff
@@ -61,11 +61,13 @@ func TestEval(t *testing.T) {
 			wantErr: any,
 		},
 		{
-			name: "string interpolation field has wrong type",
+			name: "string interpolation non-string type",
 			query: `
 				r = makeRecord(o: {a: "foo", b: 42})
 				"r._value = ${r.b}"`,
```
nathanielc, line 67: Can you update this to have both cases? A case where it errors and a case where it succeeds? For example trying to format `r` in the string would still error, `"r = ${r}"`

nathanielc (commented):
Looks really good, I'd like to see a few more tests. I called out the specific ones below. Other than that this looks ready to go!

docs/SPEC.md
Bolladeen: On a related note, should I talk to docs team about the updated functionality of string interpolation?

interpreter/interpreter.go
Bolladeen: I think either can work. As far as I could tell when testing, the actual output didn't differ for the basic types. I can change it just in case that behavior changes in the future.

docs/SPEC.md
nathanielc: Yes, ping them in their channel with a link to this PR

interpreter/interpreter.go
nathanielc: I think we should be explicit about how we are formatting the values. I don't expect Go's definition to change, but it would be best to be explicit so we are not tied to Go's behaviors.

compiler/compiler_test.go
```diff
@@ -49,7 +49,37 @@ func TestCompileAndEval(t *testing.T) {
...
+			name: "interpolated string expression with duration type",
+			fn:   `(r) => "n = ${r.n}"`,
+			inType: semantic.NewObjectType([]semantic.PropertyType{
+				{Key: []byte("r"), Value: semantic.NewObjectType([]semantic.PropertyType{
+					{Key: []byte("n"), Value: semantic.BasicDuration},
+				})},
+			}),
+			input: values.NewObjectWithValues(map[string]values.Value{
+				"r": values.NewObjectWithValues(map[string]values.Value{
+					"n": values.NewString("1m"),
```
nathanielc, line 67: Why is this is a `values.NewString`? I expected it to be a  `values.NewDuration`

docs/SPEC.md
```diff
@@ -325,6 +325,15 @@ Interpolation example:
 
 [IMPL#1775](https://github.com/influxdata/flux/issues/1775) Interpolate arbitrary expressions in string literals
```
nathanielc, line 326: This line can be removed with this change as it now implements arbitrary interpolation.

docs/SPEC.md
```diff
@@ -325,6 +325,15 @@ Interpolation example:
...
 
+String interpolation supports all basic type arbitrary expressions  without requiring casting to a string. 
+Basic types include integer literals, unsigned integer literals, float literals, boolean literals, time literals and duration literals.
```
nathanielc, line 329: another nit (I am picky about the wording in the SPEC :D ) I prefer to use only concepts that have been defined in the SPEC so that it can be a self contained document as much as possible. We haven't defined what `casting` or `basic types` are so my ask is reword this to use the `Stringable` type constraint since that is now defined below. Something like:

String interpolation expressions must satisfy the Stringable constraint.

interpreter/interpreter.go
```diff
@@ -496,6 +497,21 @@ func (itrp *Interpreter) doStringPart(ctx context.Context, part semantic.StringE
...
+				return values.NewString(strconv.FormatInt(val.(int64), 10)), nil
+			case semantic.UInt:
+				return values.NewString(strconv.FormatUint(val.(uint64), 10)), nil
+			case semantic.Float:
+				return values.NewString(strconv.FormatFloat(val.(float64), 'f', -1, 64)), nil
+			case semantic.Time:
+				return values.NewString(val.(values.Time).String()), nil
+			case semantic.Duration:
+				return values.NewString(val.(values.Duration).String()), nil
+			}
```
nathanielc, line 514: Where is the equivalent logic for the compiler? Maybe they can share the same implementation? This logic only uses types from the `values` package, perhaps a helper method in the values package like `Stringify(v Value) (String, error)` would be useful?

nathanielc (approved):

PR #3557, opened 2021-03-17
fix(csv): add error handling for wrong number of fields for raw csv

nathanielc:
Previously if a data row in a csv file had the wrong number of fields
the error would not be caught and the csv parser would get stuck in an
infinite loop.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/csv/result.go
+++ b/csv/result.go
@@ -601,7 +601,10 @@ func (d *tableDecoder) advance() (bool, error) {
 		}
 		// whatever this line is, it's not part of this table so goto DONE
 		if len(line) != d.meta.NumFields {
-			if len(line) > annotationIdx && line[annotationIdx] == "" {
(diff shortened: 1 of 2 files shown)


csv/result.go
```diff
@@ -601,7 +601,10 @@ func (d *tableDecoder) advance() (bool, error) {
 		}
 		// whatever this line is, it's not part of this table so goto DONE
 		if len(line) != d.meta.NumFields {
-			if len(line) > annotationIdx && line[annotationIdx] == "" {
+			// If this line is another annotation then the current table has no more data.
```
rockstar, line 604: This is a great comment.

rockstar (approved):
Perfect patch. Couldn't ask for anything more. Thanks.

PR #3558, opened 2021-03-18
test(stdlib/planner): add min/max pushdown tests

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/planner/group_max_test.flux
```diff
@@ -0,0 +1,113 @@
...
+    |> window(every: 1d)
+    |> max()
+
+group_max_agg_window = () =>
+    testing.loadStorage(csv: input)
+    |> range(start: 2018-05-22T19:53:26Z)
+    |> group(columns: ["host"])
+    |> aggregateWindow(fn: max, every: 1d)
+
+group_max_agg_window_empty = () =>
```
nathanielc, line 77: What is the reasoning for having these functions defined outside their testcase block? Are they reused somewhere?

stdlib/planner/group_max_test.flux
```diff
@@ -0,0 +1,113 @@
...
+
+out_max_agg_window = "
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,string,double
+#group,false,false,true,true,false,true,true,true,false
+#default,_result,,,,,,,,
+,result,table,_start,_stop,_time,_measurement,host,_field,_value
+,,0,2018-05-22T19:53:26Z,2030-01-01T00:00:00Z,2018-05-23T00:00:00Z,system,host.local,load4,1.77
+"
+
+out_max_agg_window_empty = "
```
nathanielc, line 46: Similar question here, why not put these expected results inside each testcase block?

stdlib/planner/group_max_test.flux
```diff
@@ -0,0 +1,113 @@
+package planner_test
+
+import "testing"
+import "testing/expect"
+import "planner"
+
+option now = () => (2030-01-01T00:00:00Z)
+
+input = "
+#datatype,string,long,dateTime:RFC3339,string,string,string,double
```
nathanielc, line 10: Should we use `array.from` instead of `csv` text for these test cases?

stdlib/planner/group_max_test.flux
```diff
@@ -0,0 +1,113 @@
...
+
+group_max_bare = () =>
+    testing.loadStorage(csv: input)
+    |> range(start: 2018-05-22T19:53:26Z)
+    |> filter(fn: (r) => r["_value"] == 1.77)
+    |> group(columns: ["_field"])
+    |> max()
+
+group_max_window = () =>
+    testing.loadStorage(csv: input)
```
nathanielc, line 64: Hmm I see that `testing.loadStorage` requires that you pass in `csv` text.  What if we added a new function to the testing package like this?

```
option load = (tables=<-) => tables
    |> range(start: 1800-01-01T00:00:00Z, stop: 2200-12-31T11:59:59Z)
    |> map(fn: (r) => ({r with
    _field: if exists r._field then r._field else die(msg: "test input table does not have _field column"),
    _measurement: if exists r._measurement then r._measurement else die(msg: "test input table does not have _measurement column"),
    _time: if exists r._time then r._time else die(msg: "test input table does not have _time column")
    }))
  ```
  
  This would require that we make sure the drivers update the new `load` option in `influxdb` and `idpe`. Thoughts?

nathanielc (commented):
Love these test, I have some feedback on these since they are new want to set a new conventions to start. @rockstar Want to take a look at my comments below as well?

stdlib/planner/group_max_test.flux
fchikwekwe: Yeah this doesn't seem to difficult to achieve. Do we want this issue as a part of the acceptance test PR?

stdlib/planner/group_max_test.flux
fchikwekwe: For this one and the previous question, I was just following traditional flux test convention, but as discussed I'm updating the tests for the new syntax. Hopefully makes them a little bit more readable.

fchikwekwe:
@nathanielc @rockstar @scbrickley I made some updates based on feedback here and a pairing session that some of us had. Take a look and lemme know if there are other changes that might makes these tests more readable.

rockstar (approved):
These changes look good to me. Could you please add the awkwardness in trying to figure out why the test was failing because the stringified table didn't show group keys to #3560? That way, we can track that as an important thing to see in the error output, e.g. "the table data matches, but the group keys don't"

fchikwekwe:
@rockstar yes! that was the issue that I was going to create around this. I'll add that to the acceptance tests epic.

fchikwekwe:
Oh duh, you just linked an existing issue. Yeah I'll comment.

PR #3563, opened 2021-03-19
feat(interpreter/compiler): added support for null values in string interpolation

Bolladeen:
For null values passed into string interpolation, the null value is represented as the `<null>` string when displayed.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -97,7 +97,7 @@ func TestCompileAndEval(t *testing.T) {
 			input: values.NewObjectWithValues(map[string]values.Value{
 				"r": values.NewObjectWithValues(map[string]values.Value{}),
 			}),
-			wantEvalErr: true,
+			want: values.NewString("n = <null>"),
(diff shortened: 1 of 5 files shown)


compiler/runtime.go
```diff
@@ -184,6 +184,9 @@ func (e *interpolatedEvaluator) Eval(ctx context.Context, scope Scope) (values.V
 	if err != nil {
 		return nil, err
 	}
+	if o.IsNull() {
+		return values.NewString("<null>"), nil
+	}
```
nathanielc, line 189: Could you move this logic inside the Stringify function? That way its not duplicated across both compiler and interpreter?

nathanielc (commented):
Looks good, Can you update the SPEC to say that `<null>` is used for null value in string interpolation? I think it currently states that an error occurs.

nathanielc (approved):
Wonderful!

Issue #3479, opened 2021-02-05
Add Overall Equipment Effectiveness (OEE) function to Flux


nathanielc:
See https://www.oee.com/. OEE is a measure of how effective a production process is, i.e how much time the production line had to stop and the quality of produced parts. That site presents two ways to compute `OEE`

```
OEE  = Ideal Cycle Time  * Good Count / Planned Production Time
```
where

* Ideal Cycle Time is the theoretical minimum time to produce one part
* Good Count is the number of parts produced without requiring rework
* Planned Production Time is the time the process was supposed to be running

For example if I expect a plant to be running for 7h and it takes 1s to make one part and the plant makes 18,848  then` OEE = 1s * 18,848 / 7h = 0.748 (74.8%)`.

The other method for computing OEE is as follows:

```
OEE = Availability * Performance * Quality
```
where

* Availability is the ratio of time that the process was actually running with the planned time, `Availability = Run Time / Planned Production Time`
* Performance is a measure of the efficiency of the run time, `Performance = (Ideal Cycle Time * Total Count) / Run Time`, where  `Total Count` is the total number of parts produced
* Quality is the ratio of  good parts produced and total parts produced `Quality = Good Count / Total Count`

Note that these two equations are the same

```
OEE = (Run Time / Planned Production Time) * ((Ideal Cycle Time * Total Count) / Run Time) * (Good Count / Total Count)
OEE  = Ideal Cycle Time  * Good Count / Planned Production Time
```

The idea behind these two different methods is that the break down of the result into availability, performance and quality provides useful feedback into the cause for a change in effectiveness. Presumably a user could store availability, performance and quality over time and track each metric separately while also computing and tracking OEE.

## Flux Package Design

So what should the Flux package look like?

After talking with customers it seems the industry standard is to compute the result using the availability, performance and quality metrics. Therefore I recommend we do not implement anything for the ideal cycle time method to keep the API simple.

Computing OEE from its components is easy, but it seems like the real value will be computing the intermediate values from raw event data.

Based on my reading I recommend a package API along these lines that assumes that the data is collected as counters and state changes.

Sample data will look like this:

| \_time | state   | partCount | badCount |
| ------ | -----   | --------- | -------- |
| 0      | running | 1200      | 10       |
| 1      | running | 1300      | 11       |
| 2      | stopped | 1300      | 11       |
| 3      | running | 1400      | 14       |

where

* \_time is the time when the data was collected
* state is the current state of the process/machine
[...]

PR #3564, opened 2021-03-23
feat(experimental): add OEE function

alespour:
This PR implements OEE function per https://github.com/influxdata/flux/issues/3479.

The package provides `computeAPQ` and pipeline `APQ` functions. They require that production events stream tables have `state` (production state, eg. `"running"`, `"stopped"`) and part events input `partCount` (integer) and `badCount` (integer) columns.

### Done checklist
- [x] Test cases written


alespour:
@nathanielc can you please have a look at this? thank you.

stdlib/experimental/oee/oee.flux
```diff
@@ -0,0 +1,46 @@
...
+    runningState,
+    plannedTime,
+    idealCycleTime
+) => {
+    availability = productionEvents
+        |> events.duration(unit: 1ns, columnName: "runTime")
+        |> filter(fn: (r) => r.state == runningState)
+        |> sum(column: "runTime")
+        |> duplicate(column: "_stop", as: "_time")
+        |> map(fn: (r) => ({ r with availability: float(v: r.runTime) / float(v: int(v: plannedTime)) }))
```
nathanielc, line 18: small nit, you can do the `duplicate` step as part of the `map` step to make it a bit faster. 
```
|> map(fn: (r) => ({ r with _time: r._stop, availability: float(v: r.runTime) / float(v: int(v: plannedTime)) }))
```

stdlib/experimental/oee/oee.flux
```diff
@@ -0,0 +1,46 @@
+package oee
+
+import "contrib/tomhollingworth/events"
+import "experimental"
+
+computeAPQ = (
```
nathanielc, line 6: Can you add a descriptive comment for what each argument is?

stdlib/experimental/oee/oee.flux
```diff
@@ -0,0 +1,46 @@
...
+            quality: (float(v: right.partCount) - float(v:left.badCount)) / float(v: right.partCount)
+    }))
+
+    return experimental.join(left:performance, right: quality, fn: (left, right) => ({left with
+        quality: right.quality,
+        oee: left.availability * left.performance * right.quality
+    }))
+}
+
+APQ = (
```
nathanielc, line 40: Same here can you document how this function should be used in a comment?

nathanielc (commented):
Overall looks great, just a few things to update.

scbrickley (commented):
Apart from the points already made by @nathanielc, nothing here sticks out to me as needing edits.

alespour:
@nathanielc I have checked the linked issue once again and now I'm not sure this PR meets the expectations.

The counts in sample data in #3479 are cumulative, so the `APQ` uses `difference`  to get the number of parts. But, in the proposal of higher level function that takes productions and parts events stream on input, the `partEvents` is described as a stream where each row represents production of a single part. So the current implementation in this PR is wrong I think... depending on whether counts are cumulative or not, the `APQ` function should either use `difference` or `sum`, I guess?

nathanielc:
The counts are cumulative so using difference is the correct approach. I just verified based on the description we got from some potential users that they are tracking cumulative counts. The counts can reset so it should be a `nonNegative` difference, but not a `sum`.

alespour:
So this PR should be ok as it is. Thank you for the clarification.

nathanielc (approved):

PR #3565, opened 2021-03-23
feat(ast/astutil): create astutil package for ast utilities like formatting

### LINKED ISSUE #3474: Expose a new libflux-based API for flux formatting in go
We need a go interface, akin to `ast.Format`, but using the libflux-based formatter instead of the go formatter. This new API should live in the `formatter` package, rather than the `ast` package so that we don't make all of `ast` depend on `libflux`.

To be explicit, the function interface should look like this: `func Format(f ast.File) string {}`.

DOD:
  - A function exists that can be consumed by tasks that handles formatting.
  - Comments are preserved (see #3328) across flatbuffer serialization
  - Tests exist for the new function that demonstrates that we can successfully call into flux.

jsternberg:
The `ast/astutil` package can be used to format the text using the
libflux formatter. It accepts a file and produces a string using the
formatter.

The `ast/astutil` package depends on `libflux` and should only be
imported by packages that expect to have a `libflux` dependency.

Fixes #3474.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -465,6 +465,7 @@ type File struct {
 	Package  *PackageClause       `json:"package"`
 	Imports  []*ImportDeclaration `json:"imports"`
 	Body     []Statement          `json:"body"`
+	Eof      []Comment            `json:"eof,omitempty"`
 }
 
 // Type is the abstract type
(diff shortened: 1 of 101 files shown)


skartikey (approved):
Looks good!

jsternberg:
I approve of the changes made to the original PR. I think this is good to merge now.

PR #3569, opened 2021-03-23
fix(lang): perform testing checks as part of query done

### LINKED ISSUE #3567: Flux testing.Check needs to be called once a query is complete
Currently its possible that testing.Check is called before a query has completed. This introduces a race and flakey tests.

We need to call testing.Check after the query results have been fully consumed (i.e. as part of Query.Done).

DOD:
- testing.Check is called only once the query is complete
- Test case showing there is no race

nathanielc:
Now as part of the query Done method testing checks are verified.
This avoids a race where the query was started but the results had not
been consumed before the checks were verified.

Fixes #3567

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/lang/query.go
+++ b/lang/query.go
@@ -1,15 +1,18 @@
 package lang
 
 import (
+	"context"
 	"sync"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/dependencies/testing"
 	"github.com/influxdata/flux/memory"
 	"github.com/opentracing/opentracing-go"
 )
 
 // query implements the flux.Query interface.
 type query struct {
+	ctx     context.Context
 	results chan flux.Result
 	stats   flux.Statistics
 	alloc   *memory.Allocator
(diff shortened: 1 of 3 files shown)


lang/query.go
```diff
@@ -32,6 +35,9 @@ func (q *query) Done() {
 		q.span.Finish()
 		q.span = nil
 	}
+
+	// If the testing framework was configured, verify all expectations.
+	q.err = testing.Check(q.ctx)
```
rockstar, line 40: Could this overwrite an existing error?

rockstar (approved):
Thanks for getting to this. I had a small question. I'm also curious whether we're likely to see failures in idpe and influxdb because the tests that were previously not working are working now. Do you think that's something we should expect and plan for in our next release?

lang/query.go
nathanielc: Hmm maybe I'll guard against that.

nathanielc:
@rockstar Turns out we don't have any influxdb or idpe tests yet so its not a worry.

rockstar:
> Turns out we don't have any influxdb or idpe tests yet so its not a worry.

Are we not using `expect` anywhere else? Is it new? I thought this functionality already existed in the end-to-end tests.

nathanielc:
> Are we not using expect anywhere else? Is it new? I thought this functionality already existed in the end-to-end tests.

We are but only in the Flux repo, not in any downstream repos. See https://github.com/influxdata/flux/blob/master/stdlib/universe/merge_filter_test.flux#L46 

So if CI is passing here we are in the clear and we were apparently getting lucky previously that it worked.

I have adjusted the logic to check for a previous error and add a test accordingly.

rockstar (approved):

PR #3573, opened 2021-03-24
fix(universe): refactored type conversion code and fixed error messages

Bolladeen:
The PR converts all internal error messages to user messages for all type conversion methods. It also refactors and removes decent chunks of boilerplate code for said type conversion methods.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/universe/typeconv.go
+++ b/stdlib/universe/typeconv.go
@@ -2,7 +2,6 @@ package universe
 
 import (
 	"context"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -17,14 +16,14 @@ import (
 )
 
 func init() {
-	runtime.RegisterPackageValue("universe", "string", &stringConv{})
-	runtime.RegisterPackageValue("universe", "int", &intConv{})
-	runtime.RegisterPackageValue("universe", "uint", &uintConv{})
-	runtime.RegisterPackageValue("universe", "float", &floatConv{})
-	runtime.RegisterPackageValue("universe", "bool", &boolConv{})
-	runtime.RegisterPackageValue("universe", "time", &timeConv{})
-	runtime.RegisterPackageValue("universe", "duration", &durationConv{})
-	runtime.RegisterPackageValue("universe", "bytes", bytes)
+	runtime.RegisterPackageValue("universe", "string", stringConv)
+	runtime.RegisterPackageValue("universe", "int", intConv)
+	runtime.RegisterPackageValue("universe", "uint", uintConv)
+	runtime.RegisterPackageValue("universe", "float", floatConv)
+	runtime.RegisterPackageValue("universe", "bool", boolConv)
+	runtime.RegisterPackageValue("universe", "time", timeConv)
+	runtime.RegisterPackageValue("universe", "duration", durationConv)
+	runtime.RegisterPackageValue("universe", "bytes", byteConv)
 }
 
 var (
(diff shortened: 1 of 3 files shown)


wolffcm (commented):
Can you help contextualize this a bit for me? I see the PR type is `fix` but it's described as a refactoring. Is this related to the incident?

stdlib/universe/typeconv.go
```diff
@@ -44,654 +43,298 @@ const (
...
+			str = v.Time().String()
+		case semantic.Duration:
+			str = v.Duration().String()
+		case semantic.Bytes:
+			var sb strings.Builder
+			var vB = v.Bytes()
+			for len(vB) > 0 {
+				r, size := utf8.DecodeRune(vB)
+				if r == utf8.RuneError && size == 1 {
+					return nil, errors.Newf(codes.Invalid, "invalid utf8 response")
```
nathanielc, line 78: nit, can you make this error message more clear to an end user who doesn't have the context of this code path?

i.e. something like `bytes contained non utf 8 bytes, cannot be converted into a string`

stdlib/universe/typeconv.go
```diff
@@ -44,654 +43,298 @@ const (
...
-	case semantic.Bool:
-		if v.Bool() {
-			i = 1
-		} else {
-			i = 0
+		switch v.Type().Nature() {
+		case semantic.String:
+			n, err := strconv.ParseInt(v.Str(), 10, 64)
+			if err != nil {
+				return nil, errors.Newf(codes.Invalid, "cannot convert string \"%v\" to int due to invalid syntax", v.Str())
```
nathanielc, line 108: Go has a way to quote strings for you in formatting. This line can be 

```
return nil, errors.Newf(codes.Invalid, "cannot convert string %q to int due to invalid syntax", v.Str())
```

Note the `%q` instead of `\"%v\"`.

nathanielc: Oh, I see you are using `%q` below, is there a reason its not being used here?

nathanielc (commented):
Looks really good, just a few comments/questions below

Bolladeen:
> Can you help contextualize this a bit for me? I see the PR type is `fix` but it's described as a refactoring. Is this related to the incident?

This PR is unrelated to the incident and the fix here would be fixing the error messages. Probably should have given it a better name.

stdlib/universe/typeconv.go
Bolladeen: I think it was just an oversight, good catch.

wolffcm (approved):
Thanks for the explanation. This is a nice improvement. Always nice to see more code being deleted than added. :)

stdlib/universe/typeconv_test.go
```diff
@@ -496,7 +496,7 @@ func TestTypeconv_Time(t *testing.T) {
 			name:      "time(error)",
 			v:         "NotATime",
 			want:      values.Time(0),
-			expectErr: errors.New("parsing time \"NotATime\" as \"2006-01-02T15:04:05.999999999Z07:00\": cannot parse \"NotATime\" as \"2006\""),
```
nathanielc, line 499: This older message seems to have important details for how a user could fix their date string. Should we wrap the error we get from ParseTime so that we don't loose this helpful information?

nathanielc (commented):
One small change and this looks good.

stdlib/universe/typeconv_test.go
Bolladeen: Yeah, that makes sense I'll go ahead and make that change.

nathanielc (approved):

PR #3578, opened 2021-03-25
feat(fluxtest): add test modules to the testcase transform

jsternberg:
The testcase transform now accepts a list of test modules with a
filesystem to access that module. The filesystem needs to follow the
`filesystem.Service` interface and attepting to extend a testcase will
lookup a testcase in the corresponding module and load the file
contents.

In order to extend a testcase, the testcase needs to be in a directory
where a `fluxtest.root` file is present and that file will have a name.
The file can then be extended with the following syntax:

    testcase b extends "flux/a_test" {
        a_test.testcase_name()
    }

This assumes that the `fluxtest.root` file has `"flux"` as the test
module name for this example.

The test command has now been updated so `-p` can take multiple paths,
archives, or files. If a file or directory is specified, the test
command will look in parent directories to find module roots. If
multiple module roots are found with the same name, an error will occur.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


cmd/flux/cmd/test_test.go
```diff
@@ -176,9 +180,24 @@ func testGatherFromArchive(t *testing.T, fn gatherFunc, filename string) {
...
+				t.Errorf("unexpected error: %s", err)
+			}
+			_ = f.Close()
+		}
+	}
+
 	// Check the contents of each of these.
 	ctx := filesystem.Inject(context.Background(), fs)
-	for _, index := range []int{1, 3} {
+	for _, index := range []int{2, 4} {
```
rockstar, line 200: Are these magic numbers?

cmd/flux/cmd/test.go
```diff
@@ -358,29 +401,126 @@ func (s systemfs) Close() error {
...
+	s := string(bytes.TrimSpace(name))
+	if len(s) == 0 {
+		return "", errors.New(codes.FailedPrecondition, "test module name must be non-empty")
+	}
+	return s, nil
+}
+
+// findParentTestRoot searches the parents to find a test root.
+// This function only works for the system filesystem and isn't meant
+// to be used with archive filesystems.
```
rockstar, line 500: This is a good comment. Is there a way we can assert in the function itself that it's being run in that context?

rockstar (approved):
This is a good patch, thanks for getting to it. I'm wondering if we can't add some asserts that will prevent us from mis-using some of this tooling (we have a bit of experience there recently). I leave those decisions to you and your judgment.

ast/edit/testcase.go
```diff
@@ -124,10 +127,25 @@ func newTestPackage(ctx context.Context, basePkg *ast.Package, predicate []ast.S
...
-	path := filepath.Clean(filepath.Join(basedir, extends+".flux"))
-	contents, err := filesystem.ReadFile(ctx, path)
+func extendTest(file *ast.File, extends string, modules TestModules) (*ast.File, error) {
+	components := strings.Split(extends, "/")
+	if len(components) <= 1 {
+		return nil, errors.New(codes.Invalid, "testcase extension requires a test module name and at least one other path component")
+	}
+
+	moduleName := components[0]
+	module, ok := modules[moduleName]
```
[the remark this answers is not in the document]
jsternberg: Accessing a nil map is fine in Go. It just always returns false as empty. But you can't assign to a nil map. You have to explicitly `make()` it. That's the reason why `Add()` references a pointer to the map (which is a pointer) and allocates it.

cmd/flux/cmd/test_test.go
jsternberg: Kind of? They're the index for the files to check from the `archiveFiles` array. Instead of repeating the information, I just referenced them by the index. Since I modified that array I had to modify the indexes. I can add a comment for that.

cmd/flux/cmd/test.go
jsternberg: I don't think so.

PR #3588, opened 2021-03-30
fix: Add Comment fields to Go AST structs

skartikey:
Added comments field in Go AST structs, Also changed existing comments field in rust from linklist to vector of strings.

"This is a DRAFT PR for review only"

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/core/src/ast/flatbuffers/monotype.rs
```diff
@@ -40,7 +40,11 @@ fn build_base_node<'a>(
         builder.create_string(&s)
     });
     let errors = Some(builder.create_vector(errors.as_slice()));
-    fb::BaseNode::create(builder, &fb::BaseNodeArgs { loc, errors })
+    let comments = build_vec(base_node.comments, builder, |builder, s| {
```
jsternberg, line 43: If you can, try to avoid using the same variable and then immediately overwriting it. This works fine but it might be better if you just inlined this.

libflux/core/src/ast/mod.rs
```diff
@@ -338,8 +329,8 @@ pub struct BaseNode {
     // use the base node comments when a non-terminal contains just one
     // terminal on the right hand side. This saves us populating the
     // type-specific AST nodes with comment lists when we can avoid it..
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub comments: CommentList,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub comments: Vec<String>,
```
jsternberg, line 333: Can we keep the `Comment` struct and make this a vector of comments? I'd like to keep the struct so that we can add additional information to the node in the future if we want it. For example, I don't know if we want to put position information in with the comments at some point. If we do, then using a `Comment` struct keeps that possibility open. If we use a vector of strings, that becomes impossible.

Something like:

```rust
pub struct Comment {
    text: String,
}

pub struct BaseNode {
    pub comments: Vec<Comment>,
}
```

jsternberg (commented):
This seems to be heading in the correct direction. Just a few minor comments.

libflux/core/src/ast/flatbuffers/monotype.rs
skartikey: changes are done as suggested.

libflux/core/src/ast/mod.rs
skartikey: That has been implemented too.

ast/ast.go
```diff
@@ -70,6 +70,7 @@ type Node interface {
 	Type() string // Type property is a string that contains the variant type of the node
 	Location() SourceLocation
 	Errs() []Error
+	CommentsList() []Comment
```
jsternberg, line 73: Nitpick: `CommentList`.

jsternberg (approved):

ast/ast.go
skartikey: Thanks for pointing that out, I have fixed it.

jsternberg (approved):

PR #3589, opened 2021-03-30
chore: deny rust warnings again

rockstar:
While working on influxdata/flux-lsp#212, I noticed that we
accidentally allowed warnings somewhere along the line. When they are
specifically allowed, the compiler doesn't even tell us about them. This
patch re-enables the denial of warnings (and fixes the fallout) as well
as adds the strict policy to `flux` and `fluxdoc` crates.


Bolladeen (commented):
Would the warnings that are now denied causing the CircleCI test to fail need to also be changed or is this just going to deny rust warnings again?

rockstar:
> Would the warnings that are now denied causing the CircleCI test to fail need to also be changed

Yeah, I guess I also need to fix those too.

Bolladeen (approved):

Issue #3443, opened 2021-01-20
Removing scoped labels


jpacik:
The following is a proposal for how we might remove scoped labels from records.

Currently it is possible to extend a record with a duplicate label like so
```
r = {a: 0, b: 1}
s = {r with b: 2} // {a: 0, b: 1, b: 2}
```

Notice that `s` contains two `b` labels in the example above.

In flux, tables are really just a list of records. However tables are not allowed to have duplicate columns. Therefore I propose we place the same restriction on records and make the above record extension an error that is caught during type assignment.

This will require some updates to type inference.

First, we'll need to make new assertions when typing record extensions. In particular, when typing `{r with a: 1}`, we will need to assert that `r` does not already have a label `a`. I am proposing we use predicates in order to encode this information directly into the record type. To be more precise, record types may now have optional predicates of the form `r \ l`. The way to read this is that `r` is a record type that does not have a label `l`.

Next we will need a new set of unification rules, in addition to our current ones, to account for these new predicates
```
In the rules that follow r and s are arbitrary records and v is a row variable.

1. r \ a = {}          =>  r = {}
2. r \ a = s \ a       =>  r = s
3. r \ a = s \ b       =>  r = v \ b, s = v \ a
4. r \ a = {a: _ | _}  =>  error
5. r \ a = {b: _ | s}  => s = v \ a, r = {b: _ | v}
```

Note the last rule relies on the following record equivalency
```
{b: _ | v \ a} = {b: _ | v} \ a
```

Once these new type constraints and unification rules have been added, the previous program will no longer be valid. Updating a preexisting label on a record will then be a combination of restriction followed by extension for which we can provide shorthand.

PR #3593, opened 2021-03-31
feat: add testing.load for using raw tables

rockstar:
`testing.load` uses raw tables, currently generated by `array.from`, but
not limited to only that method. Additionally, it ports the string tests
to use `testing.load` instead of `testing.loadStorage`.


stdlib/strings/strings_test.flux
```diff
@@ -1,198 +1,302 @@
...
+        _field: "_field",
+        _measurement: "_measurement",
+    }, {
+        _time: 2018-05-22T19:53:36Z,
+        _value: "   leading",
+        _field: "_field",
+        _measurement: "_measurement",
+    }])
+    // XXX: rockstar (31 Mar 2021) - Adding `map` here, otherwise `testing.diff`
+    // gets a type error because it thinks it's missing a _value label.
```
scbrickley, line 23: Why do so many test cases think they're missing a `_value` label? Is this something we need to fix?

rockstar: I was just on the phone with @nathanielc, and it turns out this is a actual problem we have, with a Big Kid Issue™ and everything. https://github.com/influxdata/flux/issues/3443

I'm going to add that reference to the comments, so there is context there.

scbrickley (approved):
LGTM - Sounds like the issues with the value labels are being addressed.

PR #3598, opened 2021-04-01
fix: Add Comment fields to Go AST structs

skartikey:
"This is a DRAFT pull request"

test result: FAILED. 527 passed; 13 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.27s

failures:
    formatter::tests::block
    formatter::tests::builtin
    formatter::tests::comments
    formatter::tests::dict
    formatter::tests::parens
    formatter::tests::temp_indent
    parser::tests::comment
    parser::tests::comment_builtin
    parser::tests::comment_function_body
    parser::tests::optional_query_metadata_preceding_query_text
    parser::tests::unary_expressions_with_too_many_comments
    scanner::tests::test_scan_comments
    scanner::tests::test_scan_offset

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg (commented):
Made a few comments. It seems like you accidentally pushed an extra commit to a fork and are trying to merge it into your branch within the origin repository. We can fix this tomorrow so there aren't two PRs.

adrian-thurston:
Hi Sunil, I think this line here is the source of the missing comments.

https://github.com/influxdata/flux/blob/0a1aa98a03bf84de9b150a522ab75108a25c320c/libflux/core/src/scanner/mod.rs#L216

I think this commit isn't in the PR so I can't comment on it directly. But if you checkout the original code, at that place we are pushing the token (a comment) to the head of the the comments linked list. Once the first non-comment token is found, we then move the contents of the linked list into that token. So in this place you'll need to append to the comments vector that is now in the scanner struct.

PR #3599, opened 2021-04-01
fix(libflux): move derive helper attribute after derive macro

pierwill:
On current nightly, this fixes a warning:

    error: derive helper attribute is used before it is introduced

It goes on to say

    = warning: this was previously accepted by the compiler but is being phased out;
    it will become a hard error in a future release!

This change gets ahead of the issue.

### Note

I tend to always be using nightly... If this kind of change is valuable, perhaps this could be useful as an early warning system?

### DIFF
--- a/libflux/core/src/ast/mod.rs
+++ b/libflux/core/src/ast/mod.rs
@@ -619,8 +619,8 @@ pub enum ParameterType {
     },
 }
 
-#[serde(tag = "type")]
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
 pub struct TypeExpression {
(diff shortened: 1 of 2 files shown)


fchikwekwe (approved):
@pierwill Yeah we've chatted in the past about testing off of nightly, but we decided as a team that adding additional warning infrastructure that we have to keep up might not provide as much as value as it adds work. Still this change looks benign to me.

fchikwekwe:
Also, changing `Move` to lowercase in your commit and PR messages will pass the semantic message check, I believe.

fchikwekwe:
Oops, closed by accident! Sorry about that!

pierwill:
Thanks, @fchikwekwe! 🙏

PR #3606, opened 2021-04-05
fix(json): dictionary encoding

### LINKED ISSUE #3603: JSON encoding fails for dictionary type
Dictionary seems like a great type for optional "schema-less" (unlike records) arguments for interacting with external systems. JSON encoding is common, however in Flux `json.encode` fails for with `Error: error calling function "encode" @1:1-1:27: json: unsupported type: map[interface {}]interface {}`. 

Example:
```
import "json"
attributes = ["region": "EU"]
json.encode(v: attributes)
```

Since Go maps are encoded as objects, and objects keys are strings in JSON, wouldn't it make sense to force keys into strings and use `map[string]interface{}` just like for Flux record type?

alespour:
This PR fixes #3603. 

Since Go maps are encoded as JSON objects where keys are strings, map key is changed from `interface{}` to `string` and keys values are simply formated as strings.

Example:
```
import "json"
attributes = ["region": "EU"]
json.encode(v: attributes)
```

Returns:
```
{"region": "EU"}
```

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/json/encode.go
+++ b/stdlib/json/encode.go
@@ -3,6 +3,7 @@ package json
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/influxdata/flux/codes"
@@ -100,7 +101,7 @@ func convertValue(v values.Value) (interface{}, error) {
 		return nil, errors.New(codes.Invalid, "cannot encode a function value")
 	case semantic.Dictionary:
 		dict := v.Dict()
-		d := make(map[interface{}]interface{}, dict.Len())
+		d := make(map[string]interface{}, dict.Len())
 		var rangeErr error
 		dict.Range(func(k, v values.Value) {
 			if rangeErr != nil {
@@ -116,7 +117,7 @@ func convertValue(v values.Value) (interface{}, error) {
 				rangeErr = err
 				return
 			}
-			d[key] = val
+			d[fmt.Sprintf("%v", key)] = val
 		})
 		if rangeErr != nil {
 			return nil, rangeErr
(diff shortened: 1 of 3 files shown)


alespour:
@jsternberg Do you think this makes sense? Thank you.

stdlib/json/encode.go
```diff
@@ -100,7 +101,7 @@ func convertValue(v values.Value) (interface{}, error) {
 		return nil, errors.New(codes.Invalid, "cannot encode a function value")
 	case semantic.Dictionary:
 		dict := v.Dict()
-		d := make(map[interface{}]interface{}, dict.Len())
+		d := make(map[string]interface{}, dict.Len())
```
jsternberg, line 104: Can you make a comment about how json only allows string keys? I didn't know that before reading this: https://www.json.org/json-en.html.

jsternberg (changes requested):
Can you add a comment to the spec in the docs folder describing this behavior? It seems fine I just want it to get explicitly documented.

alespour:
> Can you add a comment to the spec in the docs folder describing this behavior? It seems fine I just want it to get explicitly documented.

I'm sorry, do you mean somewhere to SPEC.md or Outputs.md? I'm not sure where such comment fits better.

jsternberg:
I mean `docs/SPEC.md`. If this function has an entry there, that should be updated to reflect this change.

alespour:
@jsternberg sorry, this PR fell into oblivion for some time, but now it popped up again in similar context as before. The code updated to current master, `docs/SPEC.md` also updated.

scbrickley:
Closed in favor of https://github.com/influxdata/flux/pull/4550

PR #3609, opened 2021-04-06
feat(stdlib/contrib): add Alerta notifications support

alespour:
This PR add Alerta notification support. Package `contrib/bonitoo-io/alerta` provides `alert` and `endpoint` functions.

### Done checklist
- [x] Test cases written


stdlib/contrib/bonitoo-io/alerta/alerta.flux
```diff
@@ -0,0 +1,94 @@
...
+      severity: severity,
+      service: service,
+      group: group,
+      value: value,
+      text: text,
+      tags: tags,
+      attributes: attributes,
+      origin: origin,
+      type: type,
+      createTime: strings.substring(v: string(v: timestamp), start: 0, end: 23) + "Z" // Alerta supports ISO 8601 date format YYYY-MM-DDThh:mm:ss.sssZ only
```
jsternberg, line 53: Would this benefit from a new function in the date package? Maybe something like `date.format`? This is just a question to guide future development. Since the function doesn't exist, this isn't something that needs to be changed.

jsternberg (changes requested):
Very minor change. Can you compile the latest `flux` binary and run `flux fmt -w stdlib/contrib/bonitoo-io/alerta`? I'll approve this immediately once you have done that.

stdlib/contrib/bonitoo-io/alerta/alerta.flux
alespour: `flux fmt` applied. Indeed, if there was a date formatting function, it would be useful here.

jsternberg (approved):

PR #3623, opened 2021-04-15
feat: embed compiled stdlib instead of compiling at runtime

jsternberg:
This adds a `go generate` step to write compiled semantic packages for
the standard library that can then be loaded by the go runtime to
execute the packages.

The compiled package has all non-test files concatenated together and
then analyzed together. The analyzed semantic graph is then marshaled as
a flatbuffer, gzipped, and written as an `.fc` file which stands for
"flux compiled".

These files are then embedded with the new `embed` package for go 1.16.
At runtime, these files will be accessed, deserialized into the
in-memory structure, and then used for execution instead of analyzing
them on startup. This reduces startup time drastically.

Each package is written to a separate file. This is done so that changes
to a file will affect only the files that were touched. While this is
the intention, this may not actually happen because the rust code's
fresher is dependent on the imported packages. This can likely be fixed
later so that only the modified file changes.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/embed/fs.go
```diff
@@ -0,0 +1,119 @@
...
+	"io/ioutil"
+	"time"
+)
+
+//go:embed stdlib
+var stdlib embed.FS
+
+var FS fs.FS = &gzipFS{fs: stdlib}
+
+type gzipFS struct {
```
nathanielc, line 17: It looks like gzipFS implements several of the interfaces in the `io/fs` package, Can you document which interfaces its implementing and why?

runtime/runtime.go
```diff
@@ -182,20 +184,29 @@ func (r *runtime) Stdlib() interpreter.Importer {
...
+			strings.TrimSuffix(path, ".fc"),
+			"stdlib/",
+		)
+		pkgs[name] = spkg
+		return nil
+	}); err != nil {
+		return err
 	}
 	r.pkgs = pkgs
 	r.astPkgs = nil
```
nathanielc, line 212: According to my reading of the code the `astPkgs` field is no longer read/used for anything. Can you remove it along with the `RegisterPackage` function? The change will break if any code is calling `RegisterPackage` so I think we should remove that function so that we are aware of any of those cases.

internal/cmd/builtin/cmd/generate.go
```diff
@@ -180,16 +180,6 @@ func generate(cmd *cobra.Command, args []string) error {
 func generateFluxASTFile(dir string, pkg *ast.Package) error {
 	file := jen.NewFile(pkg.Package)
 	file.HeaderComment("// DO NOT EDIT: This file is autogenerated via the builtin command.")
-	file.Func().Id("init").Call().Block(
-		jen.Qual("github.com/influxdata/flux/runtime", "RegisterPackage").
```
nathanielc, line 184: Note this is the only place where RegisterPackage was called that I know of. So it should be safe to remove that function implementation and clean up this process.

stdlib/compile.go
```diff
@@ -0,0 +1,125 @@
+// +build ignore
+
+package main
```
nathanielc, line 3: What is this cmd for? Is it just a sanity check that it works? Should we move it to the `cmd` package?

nathanielc (changes requested):
This is amazing! I love the direct its going and that it seems to be much simpler. I have several questions below.

I do have one overarching question: The semantic graph for the stdlib is already embedded into the final binary on the Rust side of things. We expose it in part via the TypeEnvironment here https://github.com/influxdata/flux/blob/master/libflux/include/influxdata/flux.h#L88 Should we use the semantic graph stored on the Rust side instead of embedding it twice?

stdlib/array/flux_gen.go
```diff
@@ -1,357 +1,3 @@
 // DO NOT EDIT: This file is autogenerated via the builtin command.
 
 package array
-
```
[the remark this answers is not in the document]
jsternberg: It's mostly because I didn't want to figure out "does this file have builtins defined or not?" In the future, we can remove the ones that don't have builtins. This ensures that all of the packages are go packages so they can be imported without questions. If they have a builtin, there's code to register the package values that needs to exist in the `init()` functions.

stdlib/embed/fs.go
```diff
@@ -0,0 +1,119 @@
+package embed
```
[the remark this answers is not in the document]
jsternberg: I think that should work.

stdlib/embed/fs.go
jsternberg: Yea, I can document this. Honestly, it's a bit more complicated than I thought it would be and it makes me think I did something wrong. I figure it will be easy to fix once I understand the new API better in the future.

stdlib/compile.go
jsternberg: It's used by `go generate` to create the flatbuffers files. Maybe I should try to integrate this behavior into the `builtin` command instead?

stdlib/array/flux_gen.go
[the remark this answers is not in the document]
nathanielc: Can you share an example where this code needs to exist for a builtin? I would assume that the builtin code would be able to live on its own and doesn't need the package defined in this file. But I am likely missing something.

jsternberg:
> I do have one overarching question: The semantic graph for the stdlib is already embedded into the final binary on the Rust side of things. We expose it in part via the TypeEnvironment here https://github.com/influxdata/flux/blob/master/libflux/include/influxdata/flux.h#L88 Should we use the semantic graph stored on the Rust side instead of embedding it twice?

Rust discards the semantic graph which is why it's embedded here instead. So at the current moment, I can't restore the semantic graph from Rust because it's not in the binary. The type environment is, but it's just a bunch of monotypes that haven't appeared to be that useful in the runtime from my perspective. My original plan was to refactor things to remove the duplication, but I realized that this work was a few hours and the refactor work would probably be done over a month or so and would be better to integrate with the user-defined packages. So this change kicks the can a bit more.

At the same time, this PR very clearly defines exactly what the Go side needs so it should be easier to do that refactor.

jsternberg:
Note: This PR cannot be merged until InfluxDB is safely using Go 1.16. This is because I chose to use the new Go `embed` package instead of using something more conventional like `go-bindata`. The reason for this choice was the `embed` package likely seems to be the new canonical way to do this so we might as well use it for this instead of using `go-bindata` and then changing to the `embed` package later. The upgrade of InfluxDB to use Go 1.16 will probably be some time in the next 2 weeks.

stdlib/array/flux_gen.go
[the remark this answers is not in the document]
jsternberg: It's the opposite. This needs to exist because I was being lazy for packages that have no builtins. When we have a builtin, we will often have code like this because there's Go code associated with the package.

https://github.com/influxdata/flux/blob/master/stdlib/array/from.go#L27-L32

This gets imported so that the `init()` function will run here:

https://github.com/influxdata/flux/blob/master/stdlib/packages.go#L6

If a package does not have a builtin, then there could be no Go files associated with that packages and then you shouldn't import it in that generated package. Since previously everything always had at least the package AST itself, we imported everything since all packages had a Go component. Now, they don't. So instead of figuring out which packages had Go code associated with them and which ones didn't, I just forced everything to be a Go package by putting a dummy file and kept the rest of the builtin generation the same.

This isn't a "it needs to be this way otherwise it won't work" but mostly a "it was this way previously and I didn't want to make the change more complicated."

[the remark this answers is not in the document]
nathanielc: Sounds good thanks for the explanation

nathanielc:
Make sense, thanks for the explanation. I agree Go's embed feature made this very easy to follow and clarifies the code a lot. As for updating InfluxDB to Go 1.16 I'll start that conversation and see where it takes us.

jsternberg:
InfluxDB is now using go 1.16 so we should be free to switch too.

embed/fs.go
```diff
@@ -0,0 +1,138 @@
+// Package embed contains an embedded copy of the flux standard library.
+// It exposes a filesystem implementation that can be used to read from
+// an already compiled copy of the standard library that can then be
+// loaded at runtime as the semantic graph.
+//
+// The compiled versions of the semantic graph are semantic graphs
+// marshaled as flatbuffers and then gzipped. In order to read them,
+// the filesystem reads the gzip file and then deserializes the
+// flatbuffer as a semantic graph.
```
nathanielc, line 9: Is using flatbuffers here a new code path? I don't think it is but wanted to confirm. What I am concerned about is we have a latent bug in how we encode semantic graph as flatbuffers that this change might expose.

stdlib/array/flux_gen.go
```diff
@@ -1,359 +1,3 @@
 // DO NOT EDIT: This file is autogenerated via the builtin command.
 
 package array
```
nathanielc, line 3: Can you add a comment here that basically states: This package is empty but it ensures that each Flux package has a corresponding Go package. This simplifies importing Go packages with or without builtin values ?

nathanielc (changes requested):
LGTM with a few small changes.

embed/fs.go
jsternberg: It's not a new code path. I believe this invokes a rust function to marshal the semantic graph.

nathanielc (approved):
LGTM I would like to see at least one other review before merging as its a large change

stdlib/array/flux_gen.go
```diff
@@ -1,359 +1,7 @@
 // DO NOT EDIT: This file is autogenerated via the builtin command.
+//
+// This file is empty but its existence ensures that each Flux package has a
+// corresponding Go package. This simplifies importing Go packages with or without
+// builtin values.
```
rockstar, line 5: This is a helpful comment.

rockstar (approved):
> I would like to see at least one other review before merging as its a large change

Honestly, the reviewable code change wasn't all that big, but the resulting diffstat is SUPER RADICAL

PR #3628, opened 2021-04-19
chore(arrow): Add telemetry to help diagnose arrow panic

scbrickley:
~~The goal of this patch is to add a `CheckLevelColumns` method to the `Table` interface, so that when we call `processMessage` in the transporter, we can get some handy debugging information about what state a table is in before when it hits the panic described in https://github.com/influxdata/idpe/issues/10155.~~

The goal of this patch is to detect when we have a table with uneven columns, and return a helpful error message that will land in the query log so that we can get some handy debugging information about the state of a table when it hits the panic described in https://github.com/influxdata/idpe/issues/10155.

Since we don't know how often real-world queries result in uneven table columns, it's possible that this will break a lot of customer queries. Ideally, we want to come up with a way to feature flag this so that it will only target customers that we know have been affected by the panic linked above.

DOD for this PR:

- ~~Any type that implements the `Table` interface has a `CheckLevelColumns` method that confirms all of a table's columns have the same number of entries, and returns a descriptive error message if not~~
- We return an error when we call `Table()` on a `ColListTableBuilder` that has uneven columns
- We have tests confirming we can hit this error consistently under expected conditions
- We understand how wide the blast radius is on this change and/or have a way to feature-flag it for specific customers.

EDIT: after speaking with @jsternberg, it was decided that [`ColListTableBuilder.Table()`](https://github.com/influxdata/flux/blob/ba11195ae5fe6b4050906a2c30e12eb95c698695/execute/table.go#L1194-L1210) was a better place to report this error, since we are fairly certain this is the only table builder that is vulnerable to this bug. This has the added bonus of resulting in a much simpler diff.

### DIFF
--- a/execute/table.go
+++ b/execute/table.go
@@ -619,7 +619,6 @@ func (b *ColListTableBuilder) AddCol(c flux.ColMeta) (int, error) {
 }
 
 func (b *ColListTableBuilder) LevelColumns() error {
-
 	for idx, c := range b.colMeta {
 		switch c.Type {
 		case flux.TBool:
--- a/execute/table.go
+++ b/execute/table.go
@@ -1204,7 +1204,7 @@ func (b *ColListTableBuilder) Table() (flux.Table, error) {
 		for i, cb := range b.cols {
 			t.cols[i] = cb.Copy()
 			if cb.Len() != b.nrows {
-				return nil, fmt.Errorf("column %s of type %s has length %d in table of length %d",
+				return nil, errors.Newf(codes.Internal, "column %s of type %s has length %d in table of length %d",
 					b.colMeta[i].Label, b.colMeta[i].Type, cb.Len(), b.nrows,
 				)
 			}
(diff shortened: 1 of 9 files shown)


execute/table.go
```diff
@@ -1204,6 +1203,11 @@ func (b *ColListTableBuilder) Table() (flux.Table, error) {
 		t.cols = make([]column, len(b.cols))
 		for i, cb := range b.cols {
 			t.cols[i] = cb.Copy()
+			if cb.Len() != b.nrows {
+				return nil, fmt.Errorf("column %s of type %s has length %d in table of length %d",
```
jsternberg, line 1207: Use the `github.com/influxdata/flux/internal/errors` package for this and use `codes.Internal`. It's the equivalent but better practice.

execute/table/copy.go
```diff
@@ -1,6 +1,8 @@
 package table
 
-import "github.com/influxdata/flux"
+import (
```
jsternberg, line 3: No problem with this but might as well revert this file since nothing else changed.

jsternberg (approved):
This generally looks good to me. A few nitpicks but this should be good to go as-is or with the changes.

execute/table/copy.go
scbrickley: Ah, I missed this. That was just my formatter plugin. I'll revert the change.

PR #3630, opened 2021-04-19
fix: upate expected output for acceptance test group_no_agg_table

wolffcm:
There seems to be an issue where we don't catch differences
in vanilla flux. I filed this issue for it:
https://github.com/influxdata/flux/issues/3631


scbrickley (approved):
Seems like a simple enough change. Glad we filed an issue to address the bug :+1:

PR #3633, opened 2021-04-19
feat(experimental): add experimental/influxdb and experimental/usage packages

gavincabbage:
This PR adds two new packages:

- `experimental/influxdb` exposes a function `api` designed to simplify calling our API from Flux. This function is implemented in an environment-agnostic fashion in this PR that does not add much value beyond existing functionality, but is intended to be replaced in OSS and Cloud by environment-specific implementations that do add significant value.
- `experimental/usage` exposes functions `from` and `limits`. These functions build on `experimental/influxdb.api` to call the organization usage and limit endpoints, respectively. 

### Done checklist
- [ ] docs/SPEC.md updated (n/a)
- [x] Test cases written


stdlib/experimental/influxdb/influxdb.flux
```diff
@@ -0,0 +1,18 @@
+package influxdb
+
+// api submits an HTTP request to the specified API path.
+// Returns HTTP status code, response headers, and body as a byte array.
+builtin api : (
```
nathanielc, line 5: How is this function different from the `http.post` method? Could it just be a wrapper around that function instead of adding a new builtin? If not why not? Perhaps we could update `http.post` to enable that kind of workflow.

stdlib/experimental/usage/usage.flux
```diff
@@ -0,0 +1,48 @@
...
+import "experimental/json"
+
+// from returns an organization's usage data. The time range to query is
+// bounded by start and stop arguments. Optional orgID, host and token arguments
+// allow cross-org and/or cross-cluster queries. Setting the raw parameter will
+// return raw usage data rather than the downsampled data returned by default.
+//
+// Note that unlike the range function, the stop argument is required here,
+// pending implementation of https://github.com/influxdata/flux/issues/3629.
+from = (start, stop, host="", orgID="{orgID}", token="", raw=false) => {
```
nathanielc, line 14: what does a default of `{orgID}` do? Seems like it would error if not supplied directly?

stdlib/experimental/usage/usage.flux
```diff
@@ -0,0 +1,48 @@
...
+	response = influxdb.api(
+        method: "get",
+		path: "/api/v2/orgs/" + orgID + "/usage",
```
nathanielc, line 17: We probably need a path sanitation method here otherwise the orgID would allow path expressions like `../../`

stdlib/experimental/influxdb/influxdb.flux
```diff
@@ -0,0 +1,18 @@
...
+	?token: string,
+	?body: string,
+	?headers: A,
+	?query: B,
+	?timeout: duration
+) => {
+	statusCode: int,
+	body: bytes,
+	headers: C
+} where A: Record, B: Record, C: Record
```
nathanielc, line 18: I think we should use a `dict` type instead of a record type here as headers and query must be strings.
Also the body should be bytes on input as well as output.

So the type would be 

```
builtin api : (
	method: string,
	path: string,
	?host: string,
	?token: string,
	?body: bytes,
	?headers: [string: string],
	?query: [string: string],
	?timeout: duration
) => {
	statusCode: int,
	body: bytes,
	headers: [string: string],
} 
```

nathanielc (commented):
Love the direction this is going! I have some ideas/questions about the design and foundation we will set for future API access to resources

stdlib/experimental/influxdb/influxdb.flux
gavincabbage: The main difference is that this function is meant to support any request method, not just an HTTP POST. It can be used to create a resource via POST, retrieve that resource with a GET, remove it with a DELETE, etc.

stdlib/experimental/usage/usage.flux
gavincabbage: Organization usage is a cloud-specific concern, so the design of this function began with cloud as the primary focus. The cloud implementation of the `api` function supports transparent interpolation of the organization ID taken off the request authorization into the requested path (that PR can is in draft in the cloud repository). This is intended to allow a customer to access API endpoints that include org or user IDs in the path without having to look up those specific IDs to use this function.

In other words, in the cloud implementation, if the request is authorized as orgID 123, this call

```
influxdb.api(method: "get", path: "/api/v2/orgs/{orgID}")
```

results in an API call to the endpoint "/api/v2/orgs/123" by deriving the caller's organization ID from the authorization on the request context.

Similarly, in the cloud implementation, if no token argument is supplied, the token is taken off the query request itself. The intent of these features is to allow this function to be easy to use for requests to one's local cloud cluster, but _also_ allow for cross cluster requests.

The initial plan was for `api` to be unimplemented outside of cloud, however it occurred to me that while these special auth-based features are cloud-specific, a user could conceivably want use this function from the Flux REPL or an OSS instance to query their cloud usage. Thus I've include a simplified implementation of the cloud function here.

This leads to the awkward situation of the `{orgID}` default not making much sense without that context, however, so I am open to any suggestions or guidance on how best to accomplish my goals here.

stdlib/experimental/usage/usage.flux
gavincabbage: I don't think that's a concern, since any request to the API has to be authorized anyway. If a user abused the org ID parameter in this way, they would not be authorized. Also, using the underlying `influxdb.api` function would avoid any sanitation we attempted, as would `http.get` or `http.post`.

stdlib/experimental/influxdb/influxdb.flux
gavincabbage: I based this implementation off of the `experimental/http.get` implementation, which used a record. I think that implementation might have predated the dictionary type. I can investigate this idea.

gavincabbage:
Thanks for taking a look @nathanielc! I've responded to your questions and linked this PR to the corresponding cloud implementation (for those who have access to that repository). It may provide more context to some of my comments,

stdlib/experimental/influxdb/influxdb.flux
gavincabbage: Implemented this change, thanks again!

rockstar (commented):
Thanks for doing this work. The code itself here looks good. It's straight forward and sticks with our normal conventions.

This review took a while because I'm a little hesitant about this kinda "wildcard" API in the `api` function. There isn't really an abstraction here. Maybe that's by design, so we can support whatever new API comes along, but "be liberal in what you expect" is not always something that I'm comfortable with. I would like to bring it up with our team before approving this work. Nothing needed on your side as of right now.

gavincabbage:
@rockstar Thanks for taking a look. I look forward to hearing what the team thinks.

To give a bit more context, the value from this `api` function really comes into play in a Cloud and OSS implementation. You're right that this generic Flux implementation is not really worthwhile. The ability to easily piggy-back on the query auth etc. to hit our API without providing host, org and token information is valuable in Cloud or OSS, however. It will facilitate things like generic, no-configuration templates, for example. 

I included this implementation to avoid leaving `api` unimplemented in e.g. the Flux REPL or other contexts, but I would have no problem with leaving it unimplemented outside of Cloud to start.

stdlib/experimental/usage/usage.flux
nathanielc: Thanks for the explanation. I think we could move this abstraction higher in the stack so we do not need a custom builtin. 

Here is the idea we discussed as team. Instead of using sentinel values like `{orgID}` that get replaced we could use an option. 

```
package influxdb

option org = ""
```

Then we inject the value of the current org into that option at runtime. 

Then the `usage` function signature can be:

```
package usage

from = (start, stop, host="", orgID=influxdb.org, token="", raw=false)  
```

Thoughts?

stdlib/experimental/usage/usage.flux
nathanielc: What about cases where another bit of Flux code is calling the `usage.from` code? It would allow the caller of `usage.from` to actually call another API endpoint. While the API will still implement its own auth it still is a concern that calling code can break the intent of the function. 

We have the `http.pathEscape` function which can be applied here to make it safe.

stdlib/experimental/influxdb/influxdb.flux
nathanielc: Makes sense. This got me thinking and I designed out an `http.do` function that would be a generic http request function. I think an API like this would work:

```
package http

builtin do : (
    url: string,
    method: string,
    ?headers: [string:string],
    ?timeout: duration,
) => {
    statusCode: int,
    body: bytes,
    headers: [string:string],
}

post = (
    url,
    headers: [:],
    timeout: 0s,
) => do(url:url, method: "POST", headers: headers, timeout: timeout)

get = (
    url,
    headers: [:],
    timeout: 0s,
) => do(url:url, method: "GET", headers: headers, timeout: timeout)
```

nathanielc (changes requested):
Per @rockstar 's comments we discussed as team this morning and have some suggestions that I mention below. 

However as this is experimental we think we should merge this with its current API and then once we have added some of the other features that I discuss below we can come back and refactor the implementation to leverage those features.

Quick summary of my more detailed comments below:

* Implement influxdb.org and influxdb.user options so we do not have to use sentinel values like `{orgID}` and `{userID}`
* Implement influxdb.api in terms of a http.do function

If we did both of those changes I believe that we could implement influxdb.api in pure Flux instead of requiring a custom builtin. This will is advantageous because it means we keep the HTTP code specific to the HTTP package and it shows that our HTTP package is robust enough for a rich application.

Thoughts?

Also @rockstar If I got any of this wrong please chime in.

I would like to see the `pathEscape` change before this merges.

gavincabbage:
@nathanielc Thanks so much for the feedback. I really love the option suggestion for `orgID`. That seems like it has potential to be much more elegant than the string placeholder solution I implemented initially. It also makes me wonder if a basic `influxdb.orgID()` function would be valuable.

> However as this is experimental we think we should merge this with its current API and then once we have added some of the other features that I discuss below we can come back and refactor the implementation to leverage those features.

In the short term, I would like to proceed with the current API as you mentioned, but we should have occasion to revisit `influxdb.api` when we do additional self-monitoring work. Specifically, I think the resource log and some form of customer-accessible query log are on the horizon will have very similar shapes to `usage.from`, giving us the opportunity to implement the suggestions you mentioned. 

For now, I have implemented the `http.pathEscape` change you requested.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+	{
+		if method, err = args.GetRequiredString("method"); err != nil {
+			return nil, err
+		}
+
+		if path, err = args.GetRequiredString("path"); err != nil {
+			return nil, err
+		}
+
+		if host, err = args.GetRequiredString("host"); err != nil {
```
nathanielc, line 74: The type signature above states that `host` is optional not required.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+			return nil, err
+		}
+
+		if token, err = args.GetRequiredString("token"); err != nil {
```
nathanielc, line 78: The type signature above states that `token` is optional not required.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		return nil, err
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return values.NewObjectWithValues(map[string]values.Value{
```
nathanielc, line 160: This should return a `dict` not an object per the type signature above.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+		"statusCode": values.NewInt(int64(resp.StatusCode)),
+		"headers":    headerToObject(resp.Header),
+		"body":       values.NewBytes(b)}), nil
+}
+
+// headerToObject constructs a values.Object from a map of header keys and values.
+func headerToObject(header http.Header) (headerObj values.Object) {
```
nathanielc, line 167: This should build a `dict` not an object, see `values.DictionaryBuilder`.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+		if raw, ok := args.Get("timeout"); !ok {
+			timeout = values.ConvertDurationNsecs(30 * time.Second)
+		} else if raw.Type().Nature() != semantic.Duration {
+			return nil, errors.New("timeout argument must be a duration")
+		} else {
+			timeout = raw.Duration()
+		}
+
+		if q, ok := args.Get("query"); ok {
+			query = q.Dict()
```
nathanielc, line 91: The `Dict` method will panic if `q` is not a Dictionary, can you make an explicit check on the type/nature of `q` before calling `Dict`.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		if h, ok := args.Get("headers"); ok {
+			headers = h.Dict()
```
nathanielc, line 95: Same goes for `h`, we need to explicitly check that `h` is a `dict`

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		if b, ok := args.Get("body"); ok {
+			body = b.Bytes()
```
nathanielc, line 99: Same here, we need to explicitly check that `b` is a `bytes` type/nature.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,175 @@
...
+			return nil, err
+		}
+
+		if token, err = args.GetRequiredString("token"); err != nil {
+			return nil, err
+		}
+
+		if raw, ok := args.Get("timeout"); !ok {
+			timeout = values.ConvertDurationNsecs(30 * time.Second)
+		} else if raw.Type().Nature() != semantic.Duration {
```
nathanielc, line 84: See this line for example of how to check the nature of a value.

stdlib/experimental/influxdb/influxdb_test.go
```diff
@@ -0,0 +1,209 @@
...
+	for _, test := range []struct {
+		name string
+
+		// passed to api func
+		args values.Object
+
+		// expected to be returned from api func call
+		expectedErrorMessage string
+		expectedStatusCode   int64
+		expectedResponseBody []byte
```
nathanielc, line 27: Can you add the expected headers to this test case? We want test coverage over that part of the API as well.

nathanielc (changes requested):
I gave this a complete review now that we have settled on the API design (at least for now).

I found some small things that need to be changed, mostly mismatches in the implementation with the declared builtin type.

gavincabbage:
> I found some small things that need to be changed, mostly mismatches in the implementation with the declared builtin type.

@nathanielc the token and host are required in this implementation, but they are not required in the Cloud (and hypothetical OSS) implementations. Any guidance on how to handle that?

As far as ultimate priorities, I would prefer to leave the Cloud/OSS implementation correctly documented and omit this platform-agnostic implementation entirely, rather than make the token and host required in Cloud.

gavincabbage:
@nathanielc I've addressed your feedback. 

I opted to handle the issue of host and token being required here but not in Cloud and OSS by leaving them optional in the function signature but providing a meaningful error message for the explaining the special case in this implementation:

```
		if host, ok, err = args.GetString("host"); err != nil {
			return nil, err
		} else if !ok {
			return nil, fluxerrors.New(codes.Invalid, `keyword argument "host" is required when executing outside InfluxDB`)
		}
```

Let me know what you think about that. Given we're planning to revisit `influxdb.api` and maybe replace it entirely with `http.do`, this feels like a decent solution for the time being.

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -0,0 +1,192 @@
...
+		"body":       values.NewBytes(b)}), nil
+}
+
+// headerToDict constructs a values.Dictionary from a map of header keys and values.
+func headerToDict(header http.Header) (values.Dictionary, error) {
+	builder := values.NewDictBuilder(semantic.NewDictType(semantic.BasicString, semantic.BasicString))
+	for name, thevalues := range header {
+		for _, onevalue := range thevalues {
+			if err := builder.Insert(values.NewString(name), values.NewString(onevalue)); err != nil {
+				return nil, errors.New("")
```
nathanielc, line 187: Can you return the real error here instead of masking it behind an empty error?

stdlib/experimental/influxdb/influxdb_test.go
```diff
@@ -0,0 +1,238 @@
...
+		// status and responses to be returned from mock api host
+		status   int
+		response []byte
+
+		// method, status, token and paths mock api host expects to receive
+		expectedMethod         string
+		expectedRequestBody    []byte
+		expectedRequestPath    string
+		expectedRequestToken   string
+		expectedRequestHeaders http.Header
```
nathanielc, line 41: These look great, can you also add tests for the expected response headers? Since the `api` type signature returns the headers we should assert that code path is working.

nathanielc (changes requested):
Two small things then this should be good to go

nathanielc (approved):
Thanks!

PR #3634, opened 2021-04-21
fix(execute): wait for the dispatcher to finish before finishing query

jsternberg:
This modifies the execution engine to wait for the dispatcher to finish
before the query is finally closed. It also adds a cancel context to the
execution engine that gets threaded into the dispatcher so the
dispatcher can be interrupted. This gets attached to the default dataset
which will abort processing tables.

### Done checklist
- [x] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -1,6 +1,8 @@
 package execute
 
 import (
+	"context"
+
 	uuid "github.com/gofrs/uuid"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
@@ -18,6 +20,12 @@ type Dataset interface {
 	Finish(error)
 
 	SetTriggerSpec(t plan.TriggerSpec)
+}
+
+// DatasetContext represents a Dataset with a context.Context attached.
+type DatasetContext interface {
+	Dataset
+	WithContext(ctx context.Context)
 }
 
 // DataCache holds all working data for a transformation.
@@ -66,7 +74,8 @@ func DatasetIDFromNodeID(id plan.NodeID) DatasetID {
 }
 
 type dataset struct {
-	id DatasetID
+	ctx context.Context
+	id  DatasetID
 
 	ts      TransformationSet
 	accMode AccumulationMode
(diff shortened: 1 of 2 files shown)


execute/executor.go
```diff
@@ -314,11 +320,15 @@ func (es *executionState) do(ctx context.Context) {
 			es.abort(err)
 		}
 	}()
+
+	go func() {
+		defer close(es.metaCh)
+		wg.Wait()
+	}()
```
nathanielc, line 327: this change makes it so that we don't close the meta channel until the wait group has finished, which means that the dispatcher has stopped?

nathanielc (commented):
As for testing do all acceptance tests still pass?

execute/executor.go
jsternberg: That's my current understanding. And the query waits for the meta channel to close before declaring itself done.

nathanielc (approved):
Thanks for the explanations, LGTM but can you have someone else on the team review as well?

execute/dataset.go
```diff
@@ -115,6 +129,14 @@ func (d *dataset) evalTriggers() (err error) {
 			// Skip the rest once we have encountered an error
 			return
 		}
+
+		select {
+		case <-d.ctx.Done():
```
adrian-thurston, line 134: There is only one event in the select so I think we can just call `Err()` here. If the result is non-nill then return it. 
https://golang.org/pkg/context/#Context

adrian-thurston (changes requested):
I think the change is good. Just concerned now that a panic in a dispatcher goroutine might deadlock a query now. Do we have that a panic in one dispatcher (or source goroutine) triggers context cancellation and that should get noticed in another dispatcher goroutine, so long as it is waiting for data? That should avoid the deadlock.

jsternberg:
@adrian-thurston a panic in the dispatcher should get caught by `recover()` which will then call `setErr` and eventually `abort()`. I'll update the context to use `.Err()` directly instead of a select.

adrian-thurston (approved):

PR #3635, opened 2021-04-22
fix(map): Return error if we map over mismatched types

### LINKED ISSUE #3621: panic for mismatched types in map
Likely version 0.65 did not panic in this way, since we are seeing new panics when testing 1.9.0rc1 . This is currently a blocker for 1.9.0 enterprise release.

Duplication:
```
import "csv"

input = "
#group,false,false,true,true,false,false,true,true,true
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string
#default,_result,,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement,meter
,,0,2017-02-16T20:30:31.713576368Z,2021-02-16T20:30:31.713576368Z,2019-04-11T07:00:00Z,1.4,bank,pge_bill,35632393IN

#group,false,false,true,true,false,false,true,true,true
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,long,string,string,string
#default,_result,,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement,meter
,,1,2017-02-16T20:30:31.713576368Z,2021-02-16T20:30:31.713576368Z,2019-04-11T07:00:00Z,2,bank,pge_bill,35632393IN
"

csv.from(csv: input ) |> map(fn: (r) => ({_field: "2 Weeks Ago", _value: r._value, _time: r._time}))
```

Original complaint from support blocking 1.9.0 rollout: https://github.com/influxdata/EAR/issues/2192#issuecomment-818317603

scbrickley:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

Closes #3621

### DIFF
--- a/stdlib/universe/map.go
+++ b/stdlib/universe/map.go
@@ -184,6 +184,11 @@ func (t *mapTransformation) Process(id execute.DatasetID, tbl flux.Table) error 
 						return errors.Newf(codes.Internal, "could not find value for column %q", c.Label)
 					}
 				}
+				if !v.IsNull() && c.Type.String() != v.Type().Nature().String() {
+					return errors.Newf(codes.Internal, "column %s:%s is not of type %v",
+						c.Label, c.Type, v.Type(),
+					)
+				}
 				if err := builder.AppendValue(j, v); err != nil {
 					return err
 				}
--- a/stdlib/universe/map_test.go
+++ b/stdlib/universe/map_test.go
@@ -964,6 +964,38 @@ f
 			},
 			wantErr: errors.New(`column _value:int is not of type float`),
 		},
+		{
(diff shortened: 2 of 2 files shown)


fchikwekwe (approved):
Sweet! Looks like a straightforward solution to me. I have a question that I don't think should block a merge. Does the test still pass if one of the mismatched types contains a null value? Like a zero or an empty string. That's an edge case that I think might be worth checking for.

fchikwekwe:
I dont think that empty values would affect the final output, which is why I approved, but I thought I'd mention it since it came to mind.

PR #3636, opened 2021-04-23
feat(fluxcli): create the flux fmt subcommand for the Flux CLI

skartikey:
This PR introduces a fmt subcommand that formats a flux script.

Format a Flux script-file:  (flux fmt [-w] <script-file>)
without -w option, the formated script will be printed on stdout


cmd/flux/cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	pkg := parser.ParseSource(string(fromFile))
+	if ast.Check(pkg) > 0 {
+		return fmt.Errorf("unexpected error: %s , filename: %s", ast.GetError(pkg), script)
+	} else if len(pkg.Files) != 1 {
+		return fmt.Errorf("expected one file in the package, got %d", len(pkg.Files))
+	}
+
+	formatStr, err := astutil.Format(pkg.Files[0])
```
nathanielc, line 44: @jsternberg Mentioned in chat that he has seen some odd behavior when going over the Go Rust boundary multiple times using JSON serialization. 

The way the `parser.ParseSource` and `astutil.Format` APIs are being used here we are definitely serializing the AST multiple times.



Digging into the details, my understanding is that this is the process based on the code as written:

parser.ParseSource: Go string ->  Rust AST -> JSON -> Go AST

astutil.Format: Go AST -> JSON -> Rust AST -> Go string

Notice we could skip all of the serialization steps if we just kept the Rust AST and called format on it directly.:

Go string -> Rust AST ->  Rust string -> Go string 


The low level APIs we could use are [libflux.Parse](https://github.com/influxdata/flux/blob/e673e37e7c4b03ef8b98b0721e3e2d862edb275d/libflux/go/libflux/parser.go#L122) and [libflux.Format](https://github.com/influxdata/flux/blob/e673e37e7c4b03ef8b98b0721e3e2d862edb275d/libflux/go/libflux/parser.go#L36)



Two takeaways from this observation:

1. If we switched to using the lower level APIs do the odd parse errors go away?
2. We should fix whatever issues are arising from the JSON serialization. Perhaps we can create a test case that calls parse.ParseSouce and then astutil.Format on our everything.flux file to validate nothing changes.

cmd/flux/cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+	script := args[0]
+
+	fromFile, err := ioutil.ReadFile(script)
+	if err != nil {
+		return err
+	}
+
+	pkg := parser.ParseSource(string(fromFile))
+	if ast.Check(pkg) > 0 {
+		return fmt.Errorf("unexpected error: %s , filename: %s", ast.GetError(pkg), script)
```
wolffcm, line 39: If I could offer some editorial guidance, it would be good to format errors here like this:
```
parse error: foo.flux|10:20-10:21: expected RPAREN
```
Or something like that. Note that errors returned from parsing (I believe) may include the file, line  number and column when they get formatted already. However errors format their source location info is fine, but the name of the file should appear first.

cmd/flux/cmd/fmt.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to format the query: %v, filename: %s", err, script)
+	}
+
+	if writeResultToSource {
+		if string(fromFile) != formatStr {
+			return updateScript(script, formatStr)
+		}
+	} else {
+		fmt.Printf("Formatted script:\n\n%s\n\n", formatStr)
```
wolffcm, line 54: If the formatted file gets written to `stdout`, won't this line interfere with the output? It would be good to write to `stderr` here instead I think.

Also, it might look nice if this output was just one line (without the new lines) since especially in CI we may be formatting many files at a time.

wolffcm (commented):
Have you thought about allowing this command to format several files at once? It could take a directory as an argument or a list of files maybe.

I guess if multiple files were being formatted it doesn't really make sense to write them all to stdout, so maybe `-w` would be required in that case.

Then in CI we might run it like this:
```sh
flux fmt -w stdlib
```

skartikey:
_> Have you thought about allowing this command to format several files at once? It could take a directory as an argument or a list of files maybe.
>_ 
I followed the std gofmt style here, so if there's no -w option is passed as an argument then show the result on stdout.
At the moment, this is how I am calling it in the Makefile to format multiple files together (include/exclude directories can be specified)-
```
STDLIB_SOURCES_FLUX = $(shell find . -name '*.flux' -not -path './colm/*')
fluxfmt:
	@for f in $(STDLIB_SOURCES_FLUX); do ./flux fmt -w $${f} | cut -c 3-; done

```
So we can keep -w option ON in the CI.

_> I guess if multiple files were being formatted it doesn't really make sense to write them all to stdout, so maybe `-w` would be required in that case.
>_ 
On CLI users can decide whether they want the output on stdout or replace the file. So If it doesn't make sense to show formatted script on stdout, I can remove that option completely and make -w required.

_> Then in CI we might run it like this:
> 
> ```shell
> flux fmt -w stdlib
> ```
>_ 
Yes, directory logic can be added to the go code.
So logic would be to search for *.flux files in the directory and format them in the loop.

cmd/flux/cmd/fmt.go
skartikey: Error from the parser doesn't contain the filename, that is why I have added that. 
Sample error (replaced unexpected with parse)- 

parse error: loc 16:47-16:48: invalid statement @16:47-16:48: : , filename: stdlib/array/array.flux

cmd/flux/cmd/fmt.go
skartikey: I will remove that line thanks.

wolffcm (approved):
Just had a minor comment about formatting errors, but this looks good to me.

cmd/flux/cmd/fmt.go
skartikey: Thanks for the suggestion, the low-level APIs did work. I have updated the PR with the changes. 
I will work on point 2 separately as it's not a blocker for the current task.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -155,6 +155,7 @@ fn dict() {
     "c": 2,
 ]"#,
     );
+    assert_unchanged(r#"[: ]"#);
```
nathanielc, line 158: I'd prefer the formatted source to be `[:]` without the space in the middle. Thoughts?

nathanielc (commented):
Looks great!

libflux/flux-core/src/formatter/tests.rs
skartikey: Yes, that can be done. I followed the existing ': space' format, but it's written that way since there's the right element after it, but there's no right element in our case. I will change it.

PR #3639, opened 2021-04-26
docs(libflux): Document items in `semantic` module

pierwill:
Adds to the rustdoc documentation for the `semantic` module.

Also fixes two warnings: `error: this URL is not a hyperlink`.


pierwill:
(The hyperlink warnings were only there because I am an inveterate `nightly` user...)

libflux/flux-core/src/semantic/types.rs
```diff
@@ -494,12 +494,12 @@ impl MonoType {
     }
 }
 
-// Tvar stands for type variable.
-// A type variable holds an unknown type.
+/// A type variable holds an unknown type.
```
rockstar, line 497: This reads a little awkwardly. "A type variable _that_ holds an unknown type" maybe?

libflux/flux-core/src/semantic/types.rs
```diff
@@ -1320,7 +1328,7 @@ mod tests {
     use crate::ast::get_err_type_expression;
     use crate::parser;
     use crate::semantic::convert::convert_polytype;
-    /// Polytype is an util method that returns a PolyType from a string.
+    /// Polytype is a utility method that returns a [`PolyType`] from a string.
```
rockstar, line 1331: It seems awkward to have "Polytype is a..." when the function is `polytype` and we also have a `Polytype`. Thoughts?

rockstar (approved):
Thanks for doing this work. It's exciting to see it come together. I had a couple of small comments on the source itself.

As an aside, I wonder if this patch allows us to remove [this line](https://github.com/influxdata/flux/blob/libflux-docs-semantic-1/libflux/flux-core/src/semantic/mod.rs#L3) entirely, or make that rule more specific. As we add more docs, I'd like to burn the ships behind us so that we can't get into that situation again. No sense in all this work not being appreciated for what it is.

pierwill:
> As an aside, I wonder if this patch allows us to remove this line entirely, or make that rule more specific.

Indeed that's where I'm headed. There are several more items that needs docs in the `semantic` module before we can remove `#[allow(missing_docs)]` (as in https://github.com/influxdata/flux/commit/851f36f0007cd1dc8e4630fed6f37700bc7938f6 for `parser`).

libflux/flux-core/src/semantic/types.rs
pierwill: I think this is alright, since it's only in the test module here. I'll change it to lowercase (`polytype`) in the doc comment on line 1331, though.

libflux/flux-core/src/semantic/types.rs
pierwill: How about:
```
/// `Tvar` stands for *type variable*.
/// A type variable holds an unknown type, before type inference.
```

PR #3640, opened 2021-04-26
fix: Remove codecov job

pierwill:
Remove codecov job.


nathanielc (approved):
Thanks, are you doing an investigation for how we can use codecov safely in the future?

jdstrand (approved):
I'm going to approve since I am a fan of this PR. Thanks! :) (note, I'm not overly familiar with flux builds)

Issue #3592, opened 2021-03-30
Create void transformation that consumes data and does nothing with it


nathanielc:
For testing purposes it is nice to have a transformation that consumes data and does nothing with it, not even pass it down to the next transformation.

Perhaps adding a builtin called `void` to the `debug` package would be a good place for this?

Type:

```
// type signature is that its identify function, except that at runtime no data is ever returned
void : (<-tables: [A]) => [A] where A: Record
// implementation is filter false, use planner rule to use actual void transformation under the hood
void = (tables=<-) => tables |> filter(fn: (r) => false)
```

Possible other names:

* drain
* consume
* sink
* dev/null

DOD

- void Flux function defined in pure Flux
- void planner rule to rewrite void function into void transformation implementation
- void transformation exists
- Tests show is consumes data but does not pass it on
- Flux acceptance tests shows that the planner rule is applied

jsternberg:
There's now a `debug.sink()` in #3643 that can be used for this. I forgot this issue existed and I forgot about the planner rule. I needed it for some performance testing and thought it would just be good to get it merged since I had to do it anyway.

PR #3643, opened 2021-04-27
feat(stdlib): add debug.slurp and debug.sink

jsternberg:
This adds two new debug functions.

`debug.slurp` will read and copy multiple buffers into a single buffer.
This can be used to test the performance impact of multiple buffers or a
single buffer.

`debug.sink` is a transformation that drains all of the data from
incoming tables.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


skartikey (approved):
Looks good!

nathanielc:
@jsternberg Does debug.sink meet the DOD here https://github.com/influxdata/flux/issues/3592 ? If so can you close that issue?

jsternberg:
Oh yes. I forgot that issue existed.

nathanielc:
Thanks! I just took a look at this PR. I don't see any acceptance tests for the new debug functions? Could you add one for debug.sink at least? I get that adding an acceptance test for debug.slurp isn't all that valuable as its an internal change not a semantic change.

PR #3644, opened 2021-04-27
perf(stdlib): optimize window and sort transformations

jsternberg:
This includes optimized versions of window and sort although it does not enable
them.

This optimizes the window transformation to more efficiently organize and
construct tables for different groups. In particular, this transformation
will ensure that the input is sorted by time, constructs the windows, and
then determines the span of rows that go into each window. This allows it
to allocate the correct amount of memory and avoid appending in random
access patterns.

This optimizes the sort transformation to perform a sort on each individual
buffer and then merge those buffers after all of the buffers have been
read. This is in contrast to constructing one huge table and then using
Go's sort from the standard library.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/window2.go
```diff
@@ -0,0 +1,490 @@
...
+	"github.com/influxdata/flux/internal/execute/dataset"
+	"github.com/influxdata/flux/internal/execute/table"
+	"github.com/influxdata/flux/internal/mutable"
+	"github.com/influxdata/flux/interval"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/values"
+)
+
+func init() {
+	plan.RegisterPhysicalRules(WindowSortedInputRule{})
```
adrian-thurston, line 24: Is there a mechanism in place so that this doesn't happen before a group pushdown rewrite and therefore stop it from matching?

stdlib/universe/window2.go
```diff
@@ -0,0 +1,490 @@
...
+// that can cause the data to become unsorted.
+type WindowSortedInputRule struct{}
+
+func (w WindowSortedInputRule) Name() string {
+	return "WindowSortedInputRule"
+}
+
+func (w WindowSortedInputRule) Pattern() plan.Pattern {
+	return plan.Pat(WindowKind,
+		plan.OneOf([]plan.ProcedureKind{SchemaMutationKind, GroupKind}),
```
adrian-thurston, line 450: This needs a `plan.Any()` as predecessor to the group/mutation.
```
return plan.Pat(
        WindowKind,
        plan.OneOf(
            []plan.ProcedureKind{SchemaMutationKind, GroupKind},
            plan.Any()))
```
Tried that out and there is a panic that occurs.
```
ts=2021-05-26T17:02:20.318730Z lvl=info msg="panic during program start" log_id=0UMPqR4G000 ot_trace_id=38359b56d810afcb ot_trace_sampled=false error="panic: Inconsistent plan graph: successor does not have edge back to predecessor" stacktrace="goroutine 182 [running]:\nruntime/debug.Stack
```

jsternberg: Ah, ok. Do you know why the `plan.Any()` is needed?

stdlib/universe/window2.go
jsternberg: There isn't really. I think the cost mechanic is supposed to choose one of the paths, but we've never really used cost. This shouldn't stop a group pushdown from happening though because it will resolve in one of two ways:

1. `ReadRange |> group |> window` -> `ReadGroup |> window`
2. `ReadRange |> group |> window` -> `ReadRange |> group |> sort |> window` -> `ReadGroup -> sort -> window`

We can add another planner rule into influxdb specifically to resolve the other side of this to either always add the sort or to remove the sort if it's not needed. I think it might be better even with the push down but it really depends on if `ReadGroup` returns the data sorted by time or not.

stdlib/universe/window2.go
jsternberg: Fixed either way. When I tested against an influxdb instance with these optimizations enabled, I got this from the profiler:

```
$ influx query -o influxdata --profilers query,operator 'import "internal/debug" from(bucket: "test1") |> range(start: -2d) |> group() |> window(every: 1d)'
Result: _result
Table: keys: [_start, _stop]
                   _start:time                      _stop:time                      _time:time                  _value:float           _field:string     _measurement:string  
------------------------------  ------------------------------  ------------------------------  ----------------------------  ----------------------  ----------------------  
2021-05-26T00:00:00.000000000Z  2021-05-26T19:38:14.589442000Z  2021-05-26T18:58:09.275329000Z                             4                      f0                      m0  
Result: _profiler
Table: keys: [_measurement]
   _measurement:string             Concurrency:int         CompileDuration:int           QueueDuration:int            PlanDuration:int         RequeueDuration:int         ExecuteDuration:int           TotalDuration:int            MaxAllocated:int          TotalAllocated:int    RuntimeErrors:string                                                                                                                                                                       flux/query-plan:string  influxdb/scanned-bytes:int  influxdb/scanned-values:int  
----------------------  --------------------------  --------------------------  --------------------------  --------------------------  --------------------------  --------------------------  --------------------------  --------------------------  --------------------------  ----------------------  -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------  --------------------------  ---------------------------  
        profiler/query                           0                      188762                       16404                           0                           0                     6797899                     7021636                        2624                           0                          digraph {
  ReadGroup6
  // GroupMode: 2, GroupKeys: [], AggregateMethod: ""
  sort4
  window3
  generated_yield

  ReadGroup6 -> sort4
  sort4 -> window3
  window3 -> generated_yield
}

                           0                            0  
Table: keys: [_measurement]
   _measurement:string                      Type:string            Label:string                   Count:int             MinDuration:int             MaxDuration:int             DurationSum:int            MeanDuration:float  
----------------------  -------------------------------  ----------------------  --------------------------  --------------------------  --------------------------  --------------------------  ----------------------------  
     profiler/operator    *universe.sortTransformation2                   sort4                           1                       48018                       48018                       48018                         48018  
     profiler/operator        *influxdb.readGroupSource              ReadGroup6                           1                      236816                      236816                      236816                        236816  
     profiler/operator  *universe.windowTransformation2                 window3                           1                       41243                       41243                       41243                         41243  
```

adrian-thurston: The empty predecessor pattern matches against nodes with  with no predecessors, so you have to put a pattern there.

adrian-thurston: How does that compare to the before time in your setup? In IDPE I'm finding the additional sort between a pushed-down group and the window is slowing down total query time by a factor of 2.

adrian-thurston: I am really feeling the feature flagging pain. Takes so long to get a before and after timing with a new feature disabled/enabled.

jsternberg: I haven't tried the speed difference with a push down. Would it be better if I split out this planner rule? We can then separate this planner rule from the other performance bits since this planner rule wasn't required to get the speed ups we were saw.

jsternberg: I've removed this now. I'd like to revisit it later but I don't think it's needed to make this worth merging.

adrian-thurston:
I'm still struggling to understand why this case is doing more total allocations.

```
from(bucket: "thurston")
        |> range( start: 0 )
        |> filter( fn: (r) => r._measurement == "cpu" )
        |> filter( fn: (r) => r.cpu == "cpu0" or r.cpu == "cpu1" ) // or r.cpu == "cpu2" or r.cpu == "cpu3" )
        |> group()
        |> window( every: 60m )
        |> filter( fn: (r) => (false) )
```
It's doing 41GB total allocated vs 13GB with the old `window`. There are 4M points after the group. This time the speed is slightly worse with the new at 19 seconds vs 16 with the old `window`

I'm not super familiar with the `arrowutil.CopyByIndexTo`. All the excess allocations come from there. Maybe the append is not doing logarithmic resizing? Will investigate that next.

jsternberg:
It's possible that it's for a few reasons.

1. There is a logarithmic resizing in this case. It's because when you do `group |> window` then the data is often partitioned like into the following windows: `[0, 1, 2, 0, 1, 2, 0, 1, 2]` this causes it to append in the worst case `b.Reserve(1)`. This doesn't perform a resize after every call, but it does cause it to use logarithmic resizing.
2. It's difficult to optimize this case. This is because the `group()` also removes these columns from the group key which makes the data "free". There's no guarantee that it won't be different values although we could potentially use the planner to optimize that.

I suspect that if you do `keep(columns: ["_time", "_value"])` it will probably allocate a lot less.

stdlib/universe/sort2.go
```diff
@@ -0,0 +1,370 @@
...
+		return nil
+	}
+
+	cr.Retain()
+	item := &sortTableMergeHeapItem{cr: cr}
+	if !s.isSorted(cr, mh.sortCols) {
+		item.indices = s.sort(cr, mh.sortCols)
+		item.offset = int(item.indices.Value(0))
+	}
+	mh.items = append(mh.items, item)
```
adrian-thurston, line 84: Wondering if it's possible here to check if the first value in `item` is in sorted order when compared against the last value in the last item of `mh.items`. Then we would know that across the entire table we are in sorted order, and the merge can simply be a concatenation of the buffers.

adrian-thurston (approved):

stdlib/universe/sort2.go
jsternberg: I might not understand this completely so I'm going to restate what I think you're saying just in case I misunderstood.

In this section of code, we are merging buffers that have already been sorted. You are wondering if we can speed this up further if we know that all of the values in buffer 0 are less than buffer 1 and then we can append or use the buffer from buffer 0 and then buffer 1.

I think this is a good thing to try, but I suspect we won't get much of a performance improvement from it. The sort buffers don't necessarily have constructed buffers. They have a set of indices for how to sort the array. If the original buffer is sorted, then we have a sorted buffers. This might happen when we have `from |> range |> group`. But in this case, the buffers likely need to be merged so I don't think this matches the scenario you mentioned.

If the original buffers are unsorted, we have to reconstruct the arrays so using the original buffer doesn't help. If the original buffers are sorted, we could potentially benefit by using the original arrays. I think this optimization in this way would require the buffers to be sorted internally and then those buffers to be sorted so something like the output of `from |> range |> filter` would qualify but also it doesn't seem like you would use a sort at that moment. It also seems like we could make the planner smarter to remove the sort in these cases.

Another thing I've noticed is that sometimes copying data ends up being faster for future transformations even if it slows down the current one. The group transformation has this problem. It sometimes makes the buffers too small and it causes later transformations to hit worst case algorithm scenarios and prevents them from benefiting from cache locality. In these cases, it's sometimes better to reconstruct the buffer. But, it's also important that the buffer isn't too large to avoid too large of allocations. We'll have to experiment to find a balance.

I did wonder if doing this would potentially make reconstructing the arrays faster but I suspect that in the scenario where one buffer is completely less than another buffer, `heap.Fix(0)` will be just as fast because it would be a `O(1)` algorithmically. We might end up in a scenario though where instead of performing a comparison for each row we might be able to only perform one comparison. If this is what you mean, then I definitely think we should try that. I don't know a specific query pattern that would construct this data pattern so it might not be worth checking for.

adrian-thurston:
I was able to reproduce what I observed in the above issue comment https://github.com/influxdata/flux/pull/3644#issuecomment-849244611, within the context of the new performance testing framework in perf-tools. In this case the new window would do more total allocations, meanwhile it was faster.

I added to the perf-testing framework the collection of go's memstats HeapSys, which shows the high-water mark of allocations. While the new version of window is doing more total allocations, the high-water mark for memory allocations is actually lower.

This finding closes this particular line of investigation into window2 regression. Instead I will turn to the specific case reported in tools where OOMs where reported.

https://influxdata.slack.com/archives/C89G3JD2R/p1629451389152100

PR #3651, opened 2021-04-28
docs(libflux): Document `fluxcore::semantic::fresh`

pierwill:
Documents all public items in module.


rockstar (changes requested):
I'm really glad to see this trickling in. Thanks for doing this work.

I'd like to see an update to how we specify `#[allow(missing_docs)]` in this patch. Updating the docs is great, but we need CI to be enforcing the existence of documentation.

pierwill:
@rockstar: Here's how I'm approaching this.

I've been removing `#[allow(missing_docs)]` where I can. For instance, all items in the `parser` module are now documented. https://github.com/influxdata/flux/pull/3662 will do the same for `scanner`. `formatter` will not be difficult.

`ast` and `semantic`, especially the latter, will require some doing before we can remove the `#[allow(missing_docs)]` attributes. This PR works toward that goal by addressing the `semantic::fresh` sub-module (see also https://github.com/influxdata/flux/pull/3649). I'm planning on several more PRs to both `semantic` and `ast`.

rockstar (approved):

PR #3662, opened 2021-04-29
docs(libflux): document `fluxcore::scanner` module

pierwill:
Documents all public items in the module, and removes the module-level `#![allow(missing_docs)]` attribute.

Have chosen here not to add a doc comment for each variant of `fluxcore::scanner::TokenType`.


libflux/flux-core/src/scanner/mod.rs
```diff
@@ -34,31 +35,44 @@ pub struct Scanner {
...
 #[display(fmt = "{}", lit)]
 pub struct Token {
+    /// Type of token.
     pub tok: TokenType,
+    /// String representation of token.
     pub lit: String,
+    /// Offset
     pub start_offset: u32,
+    /// Offset
     pub end_offset: u32,
```
pierwill, line 65: What exactly are these offsets? Characters from the start of a file?

libflux/flux-core/src/scanner/mod.rs
```diff
@@ -209,6 +223,7 @@ impl Scanner {
         }
     }
 
+    /// Produces the next comment token from the input.
```
pierwill, line 226: Wasn't sure about this one.

fchikwekwe: This sounds right to me.

libflux/flux-core/src/scanner/mod.rs
fchikwekwe: Yes. Its the start and end of the current token within the context of the file.

fchikwekwe: "Starting location of token" and "ending location of token" or something slightly better worded would probably be more accurate.

fchikwekwe (commented):
If you could change the offset comments, that would be great. Otherwise, this looks good. Thanks for helping document this file!

libflux/flux-core/src/scanner/mod.rs
pierwill: Offset in character units, or tokens?

fchikwekwe: Characters, I'm pretty sure. Not 100% sure if that includes whitespace or not.

fchikwekwe: I think whitespace is left out. If that matters.

pierwill: Ah! Definitely characters. :) Always look at the tests, I guess.

https://github.com/influxdata/flux/blob/8d902fdf517445a31af22065541a30584bcc49ef/libflux/flux-core/src/scanner/tests.rs#L4-L18

fchikwekwe (approved):

fchikwekwe:
Looks good! 🙌🏾

PR #3663, opened 2021-04-29
docs(libflux): allow missing docs in `fluxcore::semantic::flatbuffers`

pierwill:
Allow missing docs in generated code.


wolffcm (approved):
Looks good, thanks for fixing this!

PR #3665, opened 2021-04-29
chore: promote buffered table builder (and deps) from internal

adrian-thurston:
These functions are needed in IDPE to().


internal/execute/table/builder_cache.go
```diff
@@ -1,155 +1,8 @@
...
-func (d *BuilderCache) ExpireTable(key flux.GroupKey) {
-	if d.Tables == nil {
-		return
-	}
-	ts, ok := d.Tables.Delete(key)
-	if ok {
-		ts.(Builder).Release()
-	}
-}
+type Builder = table.Builder
```
jsternberg, line 7: You can combine these into this:

```
type (
    Builder = table.Builder
    BuilderCache = table.BuilderCache
)
```

jsternberg (approved):
This looks good. Do we also have to setup a type alias for `BufferedTable` or do you think we should be good as-is?

PR #3672, opened 2021-04-30
refactor(libflux): move `from` implementation closer to definition

pierwill:
This code seemed a little strange as I was reading.

The implementation of getting a `scanner::Position` from an `ast::Position` should (I think) happen in the `scanner` module.


pierwill:
For instance, it'd be confusing to end up in the `ast` source when look at the definition of `scanner::Position` from an editor on this line: 

https://github.com/influxdata/flux/blob/5fa83c26f731bbd81975b0a0fd5954e9a7411b7c/libflux/flux-core/src/parser/mod.rs#L292

jsternberg:
This move would seem to create a circular dependency between the ast and the scanner modules. While Rust doesn't treat circular dependencies in crates as a problem, I'm not sure we should encourage this. It's fairly normal in Rust to add a trait implementation for newly defined types to already existing types.

If you want to try and make it a bit more clear, you can switch the trait implementation from `From` to `Into` so that it matches with the declaration above it. I'm not sure this is worth doing, because from is the more preferred one to implement (because it implements both), but here's a sample patch just for experimenting purposes:

```
diff --git a/libflux/flux-core/src/ast/mod.rs b/libflux/flux-core/src/ast/mod.rs
index cf0d93d2..2b7a325d 100644
--- a/libflux/flux-core/src/ast/mod.rs
+++ b/libflux/flux-core/src/ast/mod.rs
@@ -47,11 +47,11 @@ impl From<&scanner::Position> for Position {
     }
 }
 
-impl From<&Position> for scanner::Position {
-    fn from(item: &Position) -> Self {
+impl Into<scanner::Position> for &Position {
+    fn into(self) -> scanner::Position {
         scanner::Position {
-            line: item.line,
-            column: item.column,
+            line: self.line,
+            column: self.column,
         }
     }
 }
diff --git a/libflux/flux-core/src/parser/mod.rs b/libflux/flux-core/src/parser/mod.rs
index a2ae8bc5..8792e776 100644
--- a/libflux/flux-core/src/parser/mod.rs
+++ b/libflux/flux-core/src/parser/mod.rs
@@ -6,7 +6,6 @@ use std::str;
 use super::DefaultHasher;
 use crate::ast;
 use crate::ast::*;
-use crate::scanner;
 use crate::scanner::*;
 
 use wasm_bindgen::prelude::*;
@@ -289,8 +288,8 @@ impl Parser {
             return SourceLocation::default();
         }
 
-        let s_off = self.s.offset(&scanner::Position::from(start)) as usize;
-        let e_off = self.s.offset(&scanner::Position::from(end)) as usize;
+        let s_off = self.s.offset(&start.into()) as usize;
+        let e_off = self.s.offset(&end.into()) as usize;
 
         SourceLocation {
             file: Some(self.fname.clone()),
```

You can also keep the current `From` trait implementation and modify the calling site and this works too.

pierwill:
Thanks for the review, @jsternberg! What does a circular dependency mean here?

jsternberg:
It means that b depends on a but a also depends on b. In this case, the `From` implementation in the scanner depends on the ast module and the `From` implementation in the ast depends on the scanner module. Generally, you want it to go one direction.

In Go, they made a choice that circular dependencies between packages weren't allowed. In Rust, circular dependencies are allowed as long as they are within the same crate. While they are allowed, I think we should still maintain a logical dependency graph and try to stick with that.

pierwill:
👍  I like that second patch modifying the caller, since it keeps things a bit more contained within `ast`.

pierwill:
Closing this for now.

PR #3677, opened 2021-04-30
docs(libflux): work on `fluxcore::ast` Rust documentation

pierwill:
This PR does three things:

- by removing [this line](https://github.com/influxdata/flux/blob/976ae6c94e6f96ed7b3025089f4a0fe2f881e3e4/libflux/flux-core/src/ast/mod.rs#L1), it requires that all items in the `ast` module have documentation. Any item that is not documented, for whatever reason, must have the `#[allow(missing_docs)]` attribute;
- adds the `#[allow(missing_docs)]` attribute to many items with missing/incomplete documentation;
- adds new documentation where it had been missing, and makes edits to existing documentation.

This required lots of changes to [ast/mod.rs](https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/ast/mod.rs). Hopefully, the description above should aid the reviewer. Also, do run `cd libflux && cargo doc --open` to peruse!


libflux/flux-core/src/ast/walk/mod.rs
```diff
@@ -266,7 +272,7 @@ where
     walk_rc(v, Rc::new(node));
 }
 
-pub fn walk_rc<'a, T>(v: &T, node: Rc<Node<'a>>)
+fn walk_rc<'a, T>(v: &T, node: Rc<Node<'a>>)
```
pierwill, line 275: Definitely need a second opinion here.

rockstar: I have often thought that we need to audit the public API for both our crates. I'm not sure it's safe to necessarily make these changes in a documentation patch (there are downstream consumers of this library directly). I'd prefer we just mark this `allow(missing_docs)` and leave the changes to the interface as they are.

rockstar (approved):
Would like a small change, but happy to see this merged after that is done.

pierwill:
> Would like a small change, but happy to see this merged after that is done.

Sounds good. I agree here. Making the change now.

PR #3683, opened 2021-05-04
docs(libflux): document `fluxcore::semantic::sub` module

pierwill:

### DIFF
--- a/libflux/flux-core/src/semantic/sub.rs
+++ b/libflux/flux-core/src/semantic/sub.rs
@@ -1,19 +1,18 @@
-#![allow(missing_docs)]
+//! Substitutions during type inference.
 
 use crate::semantic::types::{MonoType, SubstitutionMap, Tvar};
 
-// A substitution defines a function that takes a monotype as input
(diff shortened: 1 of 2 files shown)


libflux/flux-core/src/semantic/sub.rs
```diff
@@ -1,19 +1,18 @@
...
-//
-// Substitutions are idempotent. Given a substitution s and an input
-// type x, we have s(s(x)) = s(x).
-//
+/// A substitution defines a function that takes a monotype as input
+/// and returns a monotype as output. The output type is interpreted
+/// as being equivalent to the input type.
+///
+/// Substitutions are idempotent. Given a substitution *s* and an input
+/// type *x*, we have *s*(*s*(*x*)) = *s*(*x*).
```
pierwill, line 10: These italics are mathematically inspired style. We could use backticks for pseudo-code style instead, if preferred (`s(s(x)) = s(x)`).

adrian-thurston (approved):

PR #3685, opened 2021-05-04
feat(stdlib): add function to fill an empty table with one row

### LINKED ISSUE #3679: Create fill empty transformation to add null rows to empty tables for aggregate window
We should create a fill empty transformation to add null rows to empty tables. The primary purpose of this transformation is to add null rows to empty tables for the purpose of making selectors behave identically to aggregates for `aggregateWindow`. For that reason, I am proposing that the transformation be implemented in a way to signal that it is a support function and not expose it directly to the user.

The fill empty function should be defined as follows:

```
# package experimental/fill
builtin empty : (<-tables: [A]) => [A] where A: Record
```

The transformation should use a passthrough dataset and it has the singular purpose of inserting a row with a null value for any non-group key value. The code is roughly equivalent to:

```go
func (t *fillEmptyTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
    if !tbl.Empty() {
        return t.d.Process(tbl)
    }

    buf := arrow.TableBuffer{
        GroupKey: tbl.Key(),
        Columns:  tbl.Cols(),
        Values:   make([]array.Interface, len(tbl.Cols())),
    }
    for i, col := range buf.Columns {
        b := arrow.NewBuilder(col.Type, t.mem)
        if idx := execute.ColIdx(col.Label, buf.Key().Cols()); idx >= 0 {
            if err := arrow.AppendValue(b, buf.Key().Value(idx)); err != nil {
                return err
            }
        } else {
            b.AppendNull()
        }
        buf.Values[i] = b.NewArray()
    }
    out := table.FromBuffer(&buf)
    return t.d.Process(out)
}
```

This issue should not add `fill.empty()` to the invocation of `aggregateWindow`.

DOD:
* `fill.empty()` transformation exists inside of the `experimental/fill` package.
* There are tests associated with the above behavior. You may use `fill()` as a way of testing the output as testing for null values can be difficult.
* A planner rule to remove multiple invocations of this transformation when used in succession.

jsternberg:
This function will construct a table with exactly one row if the table
is empty. The columns associated with a group key will have their key
value inserted and the columns without a group key associated with it
will be filled with a null value.

Fixes #3679.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/table/fill_test.flux
```diff
@@ -0,0 +1,127 @@
...
+,,1,2021-04-13T09:15:00Z,m0,f0,2.0,b
+,,1,2021-04-13T09:30:00Z,m0,f0,5.0,b
+,,1,2021-04-13T09:35:00Z,m0,f0,6.0,b
+,,1,2021-04-13T09:40:00Z,m0,f0,8.0,b
+"
+loadData = () => testing.loadStorage(csv: inData)
+    |> range(start: 2021-04-13T09:00:00Z, stop: 2021-04-13T10:00:00Z)
+
+testcase window {
+    want = testing.loadMem(
```
rockstar, line 31: I hate to ask this, because it's so new, but it'd be nice if we could just use `array.from` and `testing.load` here. What do you think about making that change?

rockstar (approved):
I'd _like_ to see us use `testing.load` there instead of `loadStorage`, but I get that it kinda blocks you from other work. Feel free to merge either way, but if you feel like stretching, please do.

stdlib/experimental/table/fill_test.flux
jsternberg: Let me give it a try.

PR #3689, opened 2021-05-05
fix(values): add new display API for values

nathanielc:
Prior to this change it was not clear how to display a values.Value into
a string. The intended method was to type assert to the
values.ValueStringer interface and call String().

Since values.ValueStringer was not part of the values.Value interface it
was optional to implement the String() method and it was required that
it was implemented on each implementation of the various values. This
meant that it was easy to forget to add String() support to new value
types (see #3687) and that if anyone wrapped the types in their own
implementation they would loose this ability.

This change instead adds a explicit Display functions to the package
that consume a given value. This way the Display code is clearly marked
and easily tested. When adding a new type it will be clear that Display
has not been implemented for that type and it does not require that each
implementation implement the String() method.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -194,7 +194,8 @@ func (r *REPL) executeLine(t string) error {
 					return err
 				}
 			} else {
-				fmt.Println(se.Value)
+				values.Display(os.Stdout, se.Value)
+				fmt.Println()
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


rockstar (approved):
I like this API a lot more. Thanks for doing this.

rockstar (approved):

Issue #3659, opened 2021-04-29
Formatting: Escape characters are getting removed in auto format, causing test to fail


skartikey:
Example - influxdata/flux/stdlib/contrib/sranka/teams/teams.flux
Before - 

```
    body = "{
\"@type\": \"MessageCard\",
\"@context\": \"http://schema.org/extensions\",
\"title\": ${string(v: json.encode(v:title))},
\"text\": ${string(v: json.encode(v:text))},
\"summary\": ${string(v:json.encode(v:shortSummary))}
}"
```

After - 
```
    body = "{
"@type": "MessageCard",
"@context": "http://schema.org/extensions",
"title": ${string(v: json.encode(v: title))},
"text": ${string(v: json.encode(v: text))},
"summary": ${string(v: json.encode(v: shortSummary))}
}"
```
AOI - https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/formatter/mod.rs#L1056
Go formatter code can be used here as a reference. https://github.com/influxdata/flux/blob/master/ast/format.go#L651

DOD:

Escape chars are preserved
Tests cases for the formatter exist

Issue #3656, opened 2021-04-29
Formatting: else is appearing in the first column after formatting


skartikey:
Example -  stdlib/influxdata/influxdb/monitor/monitor.flux
Before - 
```
tables
        |> map(fn: (r) => ({r with level_value: if r._level == levelCrit then 4
                                                else if r._level == levelWarn then 3
                                                else if r._level == levelInfo then 2
                                                else if r._level == levelOK then 1
                                                else 0}))
```

After formatting- 
```
tables
        |> map(
            fn: (r) => ({r with
                level_value: if r._level == levelCrit then
                    4
else if r._level == levelWarn then
                    3
else if r._level == levelInfo then
                    2
else if r._level == levelOK then
                    1
else
                    0,
            }),
        )
```

nathanielc:
A possible good way to format this code could look like this:

```
return tables
        |> map(fn: (r) => ({r with
            level_value: if r._level == levelCrit then 4
                                else if r._level == levelWarn then 3
                                else if r._level == levelInfo then 2
                                else if r._level == levelOK then 1
                                else 0
        }))
```

PR #3690, opened 2021-05-06
fix(fmt): escape characters are getting removed in auto format

skartikey:
This PR is to fix the issue where escape characters are getting removed in auto-format, causing tests to fail

Fixes: #3659


stdlib/contrib/sranka/teams/teams.flux
```diff
@@ -16,37 +17,46 @@ message = (url, title, text, summary="") => {
...
     // using string body, object cannot be used because '@' is an illegal character in the object property key
-    summary2 = if summary == "" 
-        then text 
-        else summary
-    shortSummary = if strings.strlen(v: summary2) > summaryCutoff 
-        then "${strings.substring(v: summary2, start: 0, end: summaryCutoff)}..."
-        else summary2
+    summary2 = if summary == "" then
+        text
+else
```
scbrickley, line 25: Is this expected behavior for the formatter? I don't think we should be completely un-indenting `else` statements like this.

skartikey: That is correct. we have an issue already created for fixing that. #3656

scbrickley: Ah, perfect. Thanks for the clarification :+1:

scbrickley (approved):
LGTM!

PR #3691, opened 2021-05-06
fix: add 0.0.0.0 to URL validator

pierwill:


dependencies/url/validator.go
```diff
@@ -55,6 +55,7 @@ var privateIPBlocks []*net.IPNet
 
 func init() {
 	for _, cidr := range []string{
+		"0.0.0.0/16",
```
nathanielc, line 58: Why a subnet of `/16` should it be `/32`? From my quick testing only the address `0.0.0.0` not `0.0.0.1` or any other combination with zeros is an alias for localhost. Also according to https://en.wikipedia.org/wiki/0.0.0.0 it seems that 0.0.0.0 itself is the special address not a range of addresses.

pierwill: Ah yes! I suspect you're right and this should be a single address.

dependencies/url/validator.go
```diff
@@ -55,6 +55,7 @@ var privateIPBlocks []*net.IPNet
 
 func init() {
 	for _, cidr := range []string{
+		"0.0.0.0/32",
```
nathanielc, line 58: nit, can you add a comment here stating that linux treats 0.0.0.0 as 127.0.0.1

nathanielc (commented):
LGTM with one tiny change

nathanielc (approved):

PR #3692, opened 2021-05-06
fix(semantic): refactor semantic printing of types

nathanielc:
This change removes that last of the old type expressions and replaces
it with the new type expressions that match the Flux specification.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


semantic/flatbuffers_test.go
```diff
@@ -77,12 +77,12 @@ func TestDeserializeFromFlatBuffer(t *testing.T) {
 		{
 			name:     "simple unary expr",
 			fbFn:     getUnaryOpFlatBuffer,
-			polyType: `forall [] float`,
+			polyType: `float`,
```
rockstar, line 80: Surprised to see this old syntax here, but not _that_ surprised.

PR #3694, opened 2021-05-07
docs(libflux): document `fluxcore::semantic::bootsrap` module

pierwill:

The meaning of this module was not entirely clear to me. Can we say more about what "bootstrap" means here? 🤔

### DIFF
--- a/libflux/flux-core/src/semantic/bootstrap.rs
+++ b/libflux/flux-core/src/semantic/bootstrap.rs
@@ -79,7 +79,7 @@ impl From<&str> for Error {
         }
     }
 }
-/// Infers the types of the standard library returning two [`Importer`]s, one for the prelude
+/// Infers the types of the standard library returning two [`PolyTypeMap`]s, one for the prelude
 /// and one for the standard library, as well as a type variable [`Fresher`].
 #[allow(clippy::type_complexity)]
 pub fn infer_stdlib() -> Result<(PolyTypeMap, PolyTypeMap, Fresher, Vec<String>), Error> {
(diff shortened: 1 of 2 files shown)


pierwill:
The `flux` crate calls `bootstrap::infer_stdlib()` [here](https://github.com/influxdata/flux/blob/dbc8181d5a51bc2b5bcb0e7bb921a575435f7013/libflux/flux/build.rs#L51).

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -223,7 +224,7 @@ fn dependencies<'a>(
     }
 }
 
-// Constructs a polytype, or more specifically a generic row type, from a hash map
+/// Constructs a polytype, or more specifically a generic row type, from a hash map.
```
pierwill, line 227: s/row/record? 🤔

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -223,7 +224,7 @@ fn dependencies<'a>(
...
 pub fn build_polytype(from: PolyTypeMap, f: &mut Fresher) -> Result<PolyType, Error> {
     let (r, cons) = build_row(from, f);
```
pierwill, line 229: Thinking of renaming `build_row` to `build_record` in this PR as well...

jsternberg:
One caveat:

> Example: The flux crate calls bootstrap::infer_stdlib(). Does this make sense with my description of the module as "Flux start-up"?

Bootstrap happens at compile time and doesn't happen at runtime. The bootstrap infers the types and is used to embed the results into the resulting library so they can be accessed more quickly.

PR #3697, opened 2021-05-07
feat(stdlib): switch window implementation to use interval.Window

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/window_test.flux
```diff
@@ -1,46 +1,44 @@
...
-    |> window(every: 1s)
-    |> mean()
-    |> duplicate(column: "_start", as: "_time")
-    |> window(every: inf)
+testcase window_period_gaps {
+    want = array.from(
+        rows: [
+            {_time: 2018-05-22T19:53:00Z, _measurement: "diskio", _field: "io_time", _start: 2018-05-22T19:53:00Z, _stop: 2018-05-22T19:53:05Z, host: "host.local", name: "disk0", _value: 15204688},
+            {_time: 2018-05-22T19:53:03Z, _measurement: "diskio", _field: "io_time", _start: 2018-05-22T19:53:00Z, _stop: 2018-05-22T19:53:05Z, host: "host.local", name: "disk0", _value: 15204894},
+            {_time: 2018-05-22T19:53:12Z, _measurement: "diskio", _field: "io_time", _start: 2018-05-22T19:53:10Z, _stop: 2018-05-22T19:53:15Z, host: "host.local", name: "disk0", _value: 15205499},
```
fchikwekwe, line 27: I wonder if it makes a difference that we're no longer testing floats here?

fchikwekwe (commented):
Looks fine. Just a quick question about the test case values.

fchikwekwe (approved):

stdlib/universe/window_test.flux
jsternberg: This is a new test. I took the existing `window_test.flux`, renamed it to `window_default_test.flux`, and then created this file. I imagine that we'll put all of the window tests into the same file when we switch them over to `flux test` so I just did that here.

PR #3699, opened 2021-05-07
docs(libflux): document `fluxcore::semantic::walk` module

pierwill:


libflux/flux-core/src/semantic/walk/walk_mut.rs
```diff
@@ -183,6 +185,8 @@ impl<'a> NodeMut<'a> {
             _ => None,
         }
     }
+
+    /// TODO
```
scbrickley, line 189: Let's make sure we come back to this at some point.

scbrickley (approved):
One small nitpick, but other than that this looks good.

Issue #3660, opened 2021-04-29
Formatting: bracket around if is getting removed, causing stdlib generate to fail


skartikey:
file - influxdata/flux/stdlib/contrib/bonitoo-io/tickscript/tickscript.flux

Before - 

`message=(r)=>"Deadman Check: ${r._check_name} is: " + (if r.dead then "dead" else "alive"),`

After - 

`message=(r) => "Deadman Check: ${r._check_name} is: " + if r.dead then "dead" else "alive",`

error - 

Executing cargo build   {"dir": "/Users/sunilkartikey/Code/influxdata/flux/libflux", "target": "x86_64-apple-darwin"}
   Compiling flux v0.5.1 (/Users/sunilkartikey/Code/influxdata/flux/libflux/flux)
error: failed to run custom build command for `flux v0.5.1 (/Users/sunilkartikey/Code/influxdata/flux/libflux/flux)`

Caused by:
  process didn't exit successfully: `/Users/sunilkartikey/Code/influxdata/flux/libflux/target/release/build/flux-57eb54aaccfa8638/build-script-build` (exit code: 1)
  --- stderr
  Error: Error { msg: "BadExpression is not supported in semantic analysis" }
Error installing library        {"name": "flux", "error": "exit status 101"}
/Users/sunilkartikey/Code/influxdata/flux/pkg-config.sh: exit status 1
stdlib/gen.go:4: running "go": exit status 2
make: *** [stdlib/packages.go] Error 1

adrian-thurston:
In planning it was mentioned that we could move `if-expressions` down into the `primary-expression`, without requiring surrounding parenthesis. In other words, allow the after expression above to parse correctly.

Doing that creates an ambiguity in the language because of the trailing expression in the `if-expression`. An example of the ambiguity is

```
if 1 then 2 else 3 + 4
```
With if statements down in `primary-expression`, this can parse as either
```
if 1 then 2 else ( 3 + 4 )
( if 1 then 2 else 3 ) + 4

PR #3703, opened 2021-05-10
fix(fmt): bracket around if-expressions is getting removed in the auto formatter causing error

skartikey:
bracket around if-expressions is getting removed causing stdlib generate to result in a bad-expressions error.

fixes: #3660


stdlib/contrib/bonitoo-io/tickscript/tickscript.flux
```diff
@@ -10,98 +11,87 @@ import "universe"
...
-    |> duplicate(column: "_measurement", as: "__value__") // _measurement column is always present
-    |> count(column: "__value__")
-    |> findColumn(fn: (key) => key._measurement == measurement, column: "__value__")
-
-  _tables =
-    if _counts[0] == 1 then // only dummy table is in the concatenated stream
-      _dummy
-        |> drop(columns: ["unknown"])
-        |> limit(n: 0) // need empty table
+deadman = (check, measurement, threshold=0, id=(r) => "${r._check_id}", message=(r) => "Deadman Check: ${r._check_name} is: " + (if r.dead then "dead" else "alive"), topic="", tables=<-) => {
```
wolffcm, line 52: So before your change, was the formatter dropping the parentheses around the if/then/else here?

stdlib/contrib/bonitoo-io/tickscript/tickscript.flux
```diff
@@ -145,14 +135,15 @@ compute = select
...
 // groupBy groups by specified columns.
 // It is a convenience function, it adds _measurement column which is required by monitor.check().
-groupBy = (columns, tables=<-) =>
-  tables
+groupBy = (columns, tables=<-) => tables
     |> group(columns: columns)
-    |> experimental.group(columns: ["_measurement"], mode:"extend") // required by monitor.check
+    |> experimental.group(columns: ["_measurement"], mode: "extend")
 
+// required by monitor.check
```
wolffcm, line 142: For places where this change has been made, it might be tricky to undo it later. We might want to wait until we can properly format end of line comments before committing formatted stdlib files?

stdlib/experimental/geo/geo.flux
```diff
@@ -25,194 +25,200 @@ builtin stLength : (geometry: A, units: {distance: string}) => float where A: Re
...
+else
+        tables
+            |> gridFilter(
+                region: region,
+                minSize: minSize,
+                maxSize: maxSize,
+                level: level,
+                s2cellIDLevel: s2cellIDLevel,
+            )
+            |> toRows()
```
wolffcm, line 183: I'd like to understand better how we decide whether or not to parenthesize the `alternate` of a conditional. will this interpret the `|> toRows()` as being part of the `alternate`, or will `|> toRows()` be interpreted as following both branches of if conditional? I guess it interprets it as the former because `|>` has higher precedence than `if/then/else`. Do I have that right?

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -926,6 +926,10 @@ fn parens() {
     assert_format("(a() |> b()) ^ 3", "a() |> b() ^ 3");
     assert_format("1 ^ (a() |> b())", "1 ^ a() |> b()");
     assert_unchanged("(1 ^ a()) |> b()");
+    assert_format(
+        r#"qux = (r) => "foo: ${r._rab} is: " + (if r.bar then "bar" else "baz")"#,
+        r#"qux = (r) => "foo: ${r._rab} is: " + (if r.bar then "bar" else "baz")"#,
+    );
```
wolffcm, line 932: This could be replaced with `assert_unchanged`, right? Using that would help a bit with readability.

This test is related to how you added precedence for `if/then/else`, right?

Since you've also switched to using `format_child_with_parens` in the formatter for the `test`, `consequent` and `alternate`, is there a test case you could add that shows how the behavior changed as a result?

wolffcm (commented):
I'm concerned that merging this as-is will result in mangling end-of-line comments such that they won't make sense. Should we wait until we fix that before committing the formatted stdlib Flux code?

stdlib/contrib/bonitoo-io/tickscript/tickscript.flux
skartikey: Yes, that is correct. The bracket around 'if/then/else' was getting dropped.
 
so without the current code changes formatter was returning as shown below - 
`deadman = (check, measurement, threshold=0, id=(r) => "${r._check_id}", message=(r) => "Deadman Check: ${r._check_name} is: " + if r.dead then "dead" else "alive", topic="", tables=<-) => {`

libflux/flux-core/src/formatter/tests.rs
skartikey: I have updated it with assert_unchanged.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -1252,6 +1254,7 @@ impl<'a> Operator<'a> {
                 ast::Operator::NotOperator => 7,
                 ast::Operator::ExistsOperator => 7,
                 ast::Operator::InvalidOperator => 0,
+                ast::Operator::ConditionalOperator => 10,
```
nathanielc, line 1257: Where is this used? I don't see an update to the parser to add the ConditionalOperator to any AST node as a result I don't see any code path where the ConditionalOperator precedence could be queried?

nathanielc (commented):
I know this merged already but I had a question after taking a look. See below

libflux/flux-core/src/formatter/mod.rs
skartikey: That must have been an oversight - thank you for noticing.  I was running different tests to learn how consitionalOperator works in the flux parser/formatter and left that code there.  I will remove this now and will create a fresh PR deleting the nonrelated code. Thanks

PR #3705, opened 2021-05-11
chore: runtime: better "builtins not finalized" panic message

rogpeppe:
This panic often happens because the `fluxinit/static` package
has not been imported, leading to developer confusion.
Mention this common cause in the panic message.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -18,6 +18,8 @@ import (
 // required to execute a flux script.
 var Default = &runtime{}
 
+const panicNotFinalized = "builtins not finalized (you may be missing an `import _ \"github.com/influxdata/idpe/fluxinit/static\"` statement)"
+
 // runtime contains the flux runtime for interpreting and
 // executing queries.
 type runtime struct {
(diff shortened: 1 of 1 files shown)


runtime/runtime.go
```diff
@@ -18,6 +18,8 @@ import (
 // required to execute a flux script.
 var Default = &runtime{}
 
+const panicNotFinalized = "builtins not finalized (you may be missing an `import _ \"github.com/influxdata/idpe/fluxinit/static\"` statement)"
```
nathanielc, line 21: I like the change, which package you import depends on which product you are developing in. Maybe change this from 

```
import _ \"github.com/influxdata/idpe/fluxinit/static\"
```

to

```
import _ \"github.com/influxdata/REPO/fluxinit/static\"
```

nathanielc (changes requested):
LGTM with one small change

rockstar (approved):
I made the requested change.

Issue #3559, opened 2021-03-18
Add Flux Fmt step to CI


nathanielc:
This issue has a few parts:

1. Create the `flux fmt` subcommand for the Flux CLI
2. Update the `make fmt` target to also format the Flux source code
3. Add  a step to `etc/checkfmt.sh` to validate all Flux source code is formatted

DOD:

- Demo showing the workflow to use `flux fmt` within the repo.

PR #3706, opened 2021-05-11
feat(fluxcli): add flux fmt step to CI

skartikey:
1) -c (check) option is added to flux fmt cli command
2) updates make fmt target to also format the flux source code
3) added a step to etc/checkfmt.sh to validate all flux source code is
   formatted.
   
   Fixes: #3559

### DIFF
--- a/cmd/flux/cmd/fmt.go
+++ b/cmd/flux/cmd/fmt.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/influxdata/flux/libflux/go/libflux"
 	"github.com/spf13/cobra"
@@ -20,10 +21,14 @@ var fmtCmd = &cobra.Command{
 }
 
 var writeResultToSource bool
+var analyzeCurrentDirectory bool
 
 func init() {
 	rootCmd.AddCommand(fmtCmd)
+	fmtCmd.SilenceUsage = true
+	fmtCmd.SilenceErrors = true
 	fmtCmd.Flags().BoolVarP(&writeResultToSource, "write-result-to-source", "w", false, "write result to (source) file instead of stdout")
+	fmtCmd.Flags().BoolVarP(&analyzeCurrentDirectory, "analyze-current-directory", "c", false, "analyze the current <directory | file> and report if file(s) are not formatted")
 }
 
 func formatFile(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


cmd/flux/cmd/fmt.go
```diff
@@ -64,6 +66,12 @@ func format(script string) error {
 		return fmt.Errorf("failed to format the query: %s, %v", script, err)
 	}
 
+	if analyzeCurrentDirectory {
+		if string(fromFile) != formatStr {
+			return fmt.Errorf("commit includes flux files that are not fluxfmt-ed")
```
rockstar, line 71: Glad  the actual work here was so small.

etc/checkfmt.sh
```diff
@@ -36,4 +36,14 @@ if [[ -n "$FMT_OUT" ]]; then
     HAS_FMT_ERR=1
 fi
 
+cd ..
+export PATH=$PATH:.
+FMT_OUT="$(flux fmt -c stdlib 2> /dev/null;)"
+if [[ -n "$FMT_OUT" ]]; then
```
rockstar, line 42: Acknowledging that the other checks are done this way, could `flux fmt -c ...` just return a non-zero status code on error, and we check the status code?

rockstar (approved):
I'm happy to see this merged, though I had a small question about the behavior of `flux fmt -c`, and maybe it rolls up into more of a "can we refactor checkfmt.sh" conversation. I just wanted to ask the question and see where it goes.

cmd/flux/cmd/fmt.go
nathanielc: nit, but `flux fmt -c` isn't always run in the context of git or checking a specific commit. Could we make this error message more generic? Something like a list of files that are not formatted or just a statement that a unformatted file was found?

etc/checkfmt.sh
nathanielc: Agreed a status code would be best.

skartikey: Thanks for the suggestion. 
Yes, I can do the changes to return non-zero status code for flux-fmt CLI. but the problem is it will require some changes in the `main.go`  and `root.go` for capturing the error return value. As this controls other CLI commands too so changing that for this task would not be ideal. (we can discuss)

Also, rustfmt --check returns file diff (if there are any) and similarly gofmt -d or -l returns file diff/ file list. So we are still maintaining that in flux-fmt where we are showing a message to the user.

On the other note, I realize there's no point returning an error with fmt -c as we are just checking if there's any flux file that is not fluxfmt-ed. So I have disabled our sub-command to return the error and just show the message returned in the root command.

cmd/flux/cmd/fmt.go
skartikey: Yes I realise the same and added that in the code. Thanks.

PR #3707, opened 2021-05-11
feat(lang): Add Jaeger tracing information to profile metadata.

brettbuddin:
This change extracts the Jaeger trace information from the OpenTracing span and adds it to the profile metadata at the following keys:

- `tracing/id`: (string) Jaeger trace ID
- `tracing/sampled`: (boolean) Whether or not the trace was sampled or not.

I'm not happy that I have to bring in the Jaeger dependency here, but it isn't clear to me how to work around it. The interface contract returns a `TraceID` type instead of a `string`, and there isn't a clear path to augmenting/wrapping the context value on the IDPE side since that's all hidden away from us.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


rockstar (commented):
This code itself looks fine. It does seem a little weird to add this dependency to flux, instead of maybe some interface that can be provided to the flux environment. It's not really a secret that we have some first-class downstream consumers, but code specific to one of those consumers living in flux is a little awkward. I'll leave the final word to @jsternberg, as he probably has some insights from experience here and can make suggestions whereas I am mostly whinging from the peanut gallery.

brettbuddin:
@rockstar I agree with you. I'm digging harder to see if there's anything I can do on our end to smooth this over. Ideally it should just be a really stupid interface upgrade that isn't specific to any particular trace ID source.

jsternberg:
I mostly agree with all the agreeing. I think I'd be willing to call it a day though if we were to throw the `jaeger` package itself into an internal package so there's no chance that someone can depend on the package. Then we can try to smooth this out in the future. I'm not sure what we would want though so I'm not sure it's a good idea to commit to a specific interface.

brettbuddin:
@jsternberg That's a good point. I'm fine with going that route until we can assess what is actually needed for an interface. Given that the interface injection would be a pollution of the external contract for this Cloud-specific case. The current change here is opportunistic and doesn't have huge impact. I think the only thing to consider are the various version bumps that occurred when I brought this in—I don't know how often that happens in this project.

brettbuddin:
@jsternberg I moved the package into `internal`. Could you give it another look?

jsternberg (approved):

gavincabbage (approved):

PR #3710, opened 2021-05-11
fix(build): verify authenticity of rustup script

pierwill:
Uses a hard-coded SHA256 hash to verify the authenticity of the script.

This will break whenever the `rustup-init.sh` script is changed. Looking at [the history](https://github.com/rust-lang/rustup/commits/master/rustup-init.sh), that file gets updated once a month or so.


Dockerfile_build
```diff
@@ -119,10 +119,14 @@ ENV HOME=/home/$UNAME
...
 # The latest stable rust version
 ENV RUST_LATEST_VERSION=1.51.0
-
-# Install Rust
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
-    sh -s -- --default-toolchain "$RUST_LATEST_VERSION" -y
+ENV RUSTUP_LATEST_VERSION=1.24.1
+
+# Download, verify, and run rustup script
+ENV OUR_RUSTUP_SHA="427614d934032ae48a218136dc9e2aad30979575fadf1c53940676ceb3ca0561"
```
rockstar, line 125: I had to think about this a bit, to really get to the heart of why this makes me uncomfortable. When rustup gets (validly) updated, this will break our image build, which we might not notice for a bit. The "we might not notice" is certainly an issue, but it's not really the fault of this change. What I am concerned about is that we've now tied our build process to an external change. For no fault of our own, our build process is now broken; that's the kind of change I don't think we should ever stomach. For that reason, I'd prefer to look at other options to do what we want here.

rockstar (changes requested):

Dockerfile_build
pierwill: Thanks, @rockstar. In my thinking, with this new process, the only time rustup should get updated is when we change `RUSTUP_LATEST_VERSION` manually. At that point, we recompute the hash ourselves. If we fail to do so, the patch to update rustup will fail the tests. So our build process should never fail out of thin air because of this change. 

Curious to get your thoughts here. :)

pierwill: Ah! Does the "image build" in your comment refer to another part of the build process, outside of the Dockerfile in this PR?

rockstar: Ah, I think my brain read `RUSTUP_LATEST_VERSION` as `RUST_LATEST_VERSION`, so didn't see the pin. Let me think about this a little more.

For what it's worth, there is no test to fail in this area, which is why I have concerns. If there was more safety netting, I'd certainly be more about sending it hard and seeing what falls over.

rbetts: >  What I am concerned about is that we've now tied our build process to an external change. For no fault of our own, our build process is now broken; that's the kind of change I don't think we should ever stomach. For that reason, I'd prefer to look at other options to do what we want here.

If we are invoking `rustup` in a way that it can change dynamically (with each invocation) and not just when we manually update a dependency, then we already have our build process tied to an external change.  With this PR, though, we can know when an external change is malicious (as in the code-cov example).

pierwill:
> For what it's worth, there is no test to fail in this area, which is why I have concerns. If there was more safety netting, I'd certainly be more about sending it hard and seeing what falls over.

I just pushed a change that adds some bash logic to exit the build with an error message/code if the checksum fails.

rockstar (commented):
Thanks for doing this work. It's certainly valuable and important to get right. The team have opened issue #3721 and plan to take a deeper look at our options and see if we can find something that might allow us more flexibility, though we may end up just using the patch.

PR #3713, opened 2021-05-13
chore: small changes to libflux

rockstar:
This patch makes some small chore-related changes to `libflux`, mostly around the clippy lints. Some upcoming clippy lints were updated, as well as some older clippy overrides removed that were no longer needed. Additionally, an import that was `pub use` was made private.


scbrickley (approved):
Looks good. I especially like that we can collapse those match statements into a single line. Much cleaner :+1:

PR #3718, opened 2021-05-14
feat: to function writes to remote influxdb

### LINKED ISSUE #1259: Using to function with remote influxdb instance fails
I was trying to write my OSS metrics into my cloud instance. When entering in the remote bucket, org, url, and token into the To function, i am unable to write data. My suspicion is that the query needs to use the token in the to function instead of the token that is used to run the query. 

I used an actual token generated from the cloud instance, not the word `TOKEN`.

```
from(bucket: "telegraf")
  |> range(start: -15m, stop: now())
  |> to(bucket: "oss_metrics", 
        org: "russ+cloud2's organization", 
        host: "https://us-west-2-1.aws.cloud2.influxdata.com", 
        token: "TOKEN")
```

image

lesam:
Closes #1259

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


dependencies/influxdb/http.go
```diff
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"github.com/influxdata/influxdb-client-go/v2/api"
```
jsternberg, line 7: I didn't know there was a v2 now. Looks good.

dependencies/influxdb/http.go
```diff
@@ -65,6 +68,25 @@ func (h HttpProvider) SeriesCardinalityReaderFor(ctx context.Context, conf Confi
 	}, nil
 }
 
+func (h HttpProvider) WriterFor(ctx context.Context, conf Config) (PointsWriter, error) {
+	httpClient, err := h.clientFor(ctx, conf)
+	if err != nil {
+		return nil, err
+	}
+	idThenName := func (n NameOrID) string {
```
jsternberg, line 76: Can you make this a global function so it can be called in other methods potentially?

The primary reason this package doesn't use the influxdb go client is because there was a v0 version being used in one project that depends on flux and there was a newer v1 and they had different APIs so we weren't able to use the official API. This package should be using the official API everywhere if it's available and I have a feeling this functionality will be needed for the query methods too.

dependencies/influxdb/provider.go
```diff
@@ -100,6 +106,12 @@ type Reader interface {
 	Read(ctx context.Context, f func(flux.Table) error, mem memory.Allocator) error
 }
 
+// PointsWriter is a write on which points can be written in batches
+type PointsWriter interface {
```
jsternberg, line 110: Can you rename this to `Writer` so it has parity with `Reader`?

dependencies/influxdb/provider.go
```diff
@@ -100,6 +106,12 @@ type Reader interface {
...
+	Write(...protocol.Metric) error
+	Close() error
```
jsternberg, line 112: Use `io.Closer` here.

dependencies/influxdb/service.go
```diff
@@ -0,0 +1,154 @@
+// TODO: copied from "github.com/influxdata/influxdb-client-go/v2/api/http", contribute back changes instead
```
jsternberg, line 1: Can you either put this file into an internal package within this directory or avoid exporting anything in this file? The packages in this directory are particularly susceptible to API changes because they are intended to be implemented by third-parties that are embedding flux. I'd like to avoid creating a contract for a fork of a dependency.

dependencies/influxdb/http.go
```diff
@@ -441,3 +463,28 @@ func (h seriesCardinalityHttpReader) Read(ctx context.Context, f func(flux.Table
 	}
 	return h.Query(ctx, f, &file, h.Bounds.Now, mem)
 }
+
+type httpWriter struct {
+	writer *api.WriteAPIImpl
+}
+
+
+var _ PointsWriter = &httpWriter{}
```
jsternberg, line 472: This should just be:

```
var _ PointsWriter = httpWriter{}
```

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
 package influxdb
 
 import (
+	"context"
```
jsternberg, line 4: Sort the imports.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
...
 // ToKind is the kind for the `to` flux function
 const ToKind = "to"
 
-var ToSignature = runtime.MustLookupBuiltinType("influxdata/influxdb", "to")
-
 func init() {
-	runtime.RegisterPackageValue("influxdata/influxdb", ToKind, flux.MustValue(flux.FunctionValueWithSideEffect(ToKind, nil, ToSignature)))
+	toSignature := runtime.MustLookupBuiltinType("influxdata/influxdb", "to")
+	runtime.RegisterPackageValue("influxdata/influxdb", ToKind, flux.MustValue(flux.FunctionValueWithSideEffect(ToKind, createToOpSpec, toSignature)))
+	flux.RegisterOpSpec(ToKind, func() flux.OperationSpec { return &ToOpSpec{} })
```
jsternberg, line 30: You can remove this line. It doesn't do anything but remains in the codebase in a lot of places because it doesn't cause any harm and nobody has gone back to remove them.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
...
+	t, err := NewToTransformation(a.Context(), d, cache, s, deps)
+	if err != nil {
+		return nil, nil, err
+	}
+	return t, d, nil
+}
+
+// WriterProvider is a type which provides PointsWriter types
+// for a supplied authorizer, org ID and bucket ID
+type WriterProvider interface {
```
jsternberg, line 65: I think this might be unused now that it is in `dependencies/influxdb`.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
...
+	WriterFor(ctx context.Context, org NameOrID, bucket NameOrID) (PointsWriter, error)
+}
+
+// PointsWriter is a write on which points can be written in batches
+type PointsWriter interface {
```
jsternberg, line 70: Same here. I don't think this is needed.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
...
+}
+
+// Kind returns the kind for the ToOpSpec function.
+func (ToOpSpec) Kind() flux.OperationKind {
+	return ToKind
+}
+
+
+// RowMetric is a Metric
+type RowMetric struct {
```
jsternberg, line 599: If you can make these unexported that would be good.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,622 @@
...
+
+func (r RowMetric) TagList() []*lp.Tag {
+	return r.Tags
+}
+
+func (r RowMetric) FieldList() []*lp.Field {
+	return r.Fields
+}
+
+var _ lp.Metric = &RowMetric{}
```
jsternberg, line 622: Newline at the end here.

jsternberg (changes requested):
Looks good. A few style changes and suggestions. Make sure you `go fmt` everything and sort the imports since I saw a few files where that wasn't the case.

Do you just need this to work for the flux library itself or do you also need it to work in influxdb/idpe? This change will work fine to get the flux CLI to work, but we'll need some follow up work to integrate these same changes into influxdb and idpe. I don't think these changes should break the functionality of either of those.

dependencies/influxdb/service.go
lesam: I plan to remove this file before merge - the only difference from the copy is that it takes a `Doer` interface instead of a `http.Client` struct. I think this can be changed upstream before merge.

lesam:
I only need this change for the flux library itself - I intend to pick it up in Kapacitor where we don't overwrite `to`.

I'm not sure how this affects the Kafka `to` that's also in this repo - is there a test I can run for that?

Also, what tests should be added with this PR?

dependencies/influxdb/service.go
lesam: Specifically I need this PR merged upstream: https://github.com/influxdata/influxdb-client-go/pull/256/files

lesam: Actually let's not block this on the other PR - I put in a TODO referencing the upstream PR and moved the service implementation to `internal`

dependencies/influxdb/http.go
```diff
@@ -441,3 +457,59 @@ func (h seriesCardinalityHttpReader) Read(ctx context.Context, f func(flux.Table
...
+	latestError chan error
+}
+
+func NewHttpWriter(writer *api.WriteAPIImpl) (*httpWriter, error) {
+	w := &httpWriter{
+		writer:      writer,
+		errChan:     writer.Errors(),
+		latestError: make(chan error, 1),
+	}
+	go func() {
```
lesam, line 473: The error handling here is a bit ugly, there is an open issue for the client to improve this with a v3 of the client.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,604 @@
...
-
 func init() {
-	runtime.RegisterPackageValue("influxdata/influxdb", ToKind, flux.MustValue(flux.FunctionValueWithSideEffect(ToKind, nil, ToSignature)))
+	toSignature := runtime.MustLookupBuiltinType("influxdata/influxdb", "to")
+	runtime.RegisterPackageValue("influxdata/influxdb", ToKind, flux.MustValue(flux.FunctionValueWithSideEffect(ToKind, createToOpSpec, toSignature)))
+	plan.RegisterProcedureSpecWithSideEffect(ToKind, newToProcedure, ToKind)
+	execute.RegisterTransformation(ToKind, createToTransformation)
+}
+
+const (
```
lesam, line 35: This is mostly stolen from idpe, though idpe uses v0 of `influxdb-client-go`

dependencies/influxdb/http.go
```diff
@@ -441,3 +457,59 @@ func (h seriesCardinalityHttpReader) Read(ctx context.Context, f func(flux.Table
...
 	return h.Query(ctx, f, &file, h.Bounds.Now, mem)
 }
+
+type httpWriter struct {
+	writer      *api.WriteAPIImpl
+	errChan     <-chan error
+	latestError chan error
+}
+
+func NewHttpWriter(writer *api.WriteAPIImpl) (*httpWriter, error) {
```
jsternberg, line 467: Leave this unexported too. I don't want to create an external api that others can rely on.

dependencies/influxdb/provider.go
```diff
@@ -100,6 +107,12 @@ type Reader interface {
 	Read(ctx context.Context, f func(flux.Table) error, mem memory.Allocator) error
 }
 
+// PointsWriter is a write on which points can be written in batches
```
jsternberg, line 110: Godoc comment wasn't updated. Also end it with a period.

dependencies/influxdb/provider.go
```diff
@@ -132,3 +149,10 @@ func (n NameOrID) IsValid() bool {
 func (n NameOrID) IsZero() bool {
 	return n.ID == "" && n.Name == ""
 }
+
+func (n NameOrID) IdOrElseName() string {
```
jsternberg, line 153: Rename to `IdOrName()` and document what this does.

stdlib/influxdata/influxdb/to.go
```diff
@@ -1,15 +1,604 @@
...
+
+	return err
+}
+
+func createToOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
+	if err := a.AddParentFromArgs(args); err != nil {
+		return nil, err
+	}
+
+	_, httpOK, err := args.GetString("url")
```
jsternberg, line 546: I don't think that url or brokers are valid arguments for this to function. I assume this was probably when `to()` was going to be an all-encompassing function that included choosing the correct implementation automatically. Originally, from was going to take multiple arguments and choose which underlying from to call depending on those arguments.

With that, I think you can remove the entirety of the kafka related stuff here and either do this:

```go
s := &ToOpSpec{}
if err := s.ReadArgs(args); err != nil {
    return nil, err
}
return s, nil
```

Or you can remove that `ReadArgs` function altogether and replace the implementation here with it.

dependencies/influxdb/internal/service.go
```diff
@@ -0,0 +1,155 @@
...
+	"context"
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"mime"
+	"net/http"
+	"net/url"
+	"strconv"
+
+	http2 "github.com/influxdata/influxdb-client-go/v2/api/http"
```
jsternberg, line 15: Can you rename this to something like `influxhttp`, `ihttp`, `httpapi`, or `apihttp`? I'd like a name other than `http2` that is a bit more descriptive.

jsternberg (changes requested):
I think we're almost there. A few more changes. I didn't catch the kafka stuff the first time but the kafka stuff can be removed. It's not used and it's not even possible to reach that code.

stdlib/influxdata/influxdb/to.go
lesam: I kept ReadArgs - I think it reads a bit more nicely.

jsternberg (approved):
I think this looks good. The only condition is to please compile this into influxdb and make sure that to still works in it. If everything is correct, the http `to()` won't work in influxdb but the normal `to()` will still work. We'll get them all integrated together but this is the first step.

dependencies/influxdb/http.go
```diff
@@ -441,3 +458,64 @@ func (h seriesCardinalityHttpReader) Read(ctx context.Context, f func(flux.Table
...
+	buf := new(bytes.Buffer)
+	enc := protocol.NewEncoder(buf)
+	for i := range metric {
+		buf.Truncate(0)
+		_, err := enc.Encode(metric[i])
+		if err != nil {
+			h.writer.Flush()
+			return err
+		}
+		h.writer.WriteRecord(strings.TrimRight(buf.String(), "\n"))
```
lesam, line 502: Note this 'TrimRight' to avoid double-writing `\n`

dependencies/influxdb/http_test.go
```diff
@@ -0,0 +1,160 @@
+package influxdb_test
```
lesam, line 1: This is a newly written test for `httpWriter`

lesam:
> I think this looks good. The only condition is to please compile this into influxdb and make sure that to still works in it. If everything is correct, the http `to()` won't work in influxdb but the normal `to()` will still work. We'll get them all integrated together but this is the first step.

There will need to be some influxdb and IDPE changes when this merges - mainly we can't use the same name for `ToOpSpec` in both places. 1.x does this more correctly by using its own name, we will have to update 2.x and likely IDPE similarly.

stdlib/influxdata/influxdb/to_test.go
```diff
@@ -0,0 +1,795 @@
+package influxdb_test
```
lesam, line 1: Stolen from IDPE

mock/provider.go
```diff
@@ -0,0 +1,18 @@
+package mock
+
+import (
+	"context"
+
+	"github.com/influxdata/flux/dependencies/influxdb"
+)
+
+type MockProvider struct {
```
jsternberg, line 9: Can you rename this and the file `InfluxDBProvider` and `influxdb_provider.go` respectively?

jsternberg (approved):
One change to a name but pre-approving since the rest is good. If you make that one change, feel free to merge.

PR #3723, opened 2021-05-17
fix(execute): remove deadlock when an error happens while the dispatcher is stopping

jsternberg:
If the dispatcher had an error, such as a panic, while stopping, it
would attempt to set the error on the dispatcher. The `Stop()` method,
which could be triggered by a previous error, would take the lock and
then wait for the workers to finish.

But a worker could then try to set the error before it exited and
setting an error also attempted to take the lock. This created a
deadlock where stopping required waiting for the worker, but the worker
was waiting for stop to release the lock.

This modifies the dispatcher so that it does not hold the lock while it
is waiting and only holds it while using the shared state.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/dispatcher.go
+++ b/execute/dispatcher.go
@@ -107,14 +107,22 @@ func (d *poolDispatcher) setErr(err error) {
 
 //Stop the dispatcher.
 func (d *poolDispatcher) Stop() error {
+	// Check if this is the first time invoking this method.
+	d.mu.Lock()
+	if !d.closed {
+		// If not, mark the dispatcher as closed and signal to the current
+		// workers that they should stop processing more work.
+		d.closed = true
+		close(d.closing)
+	}
+	d.mu.Unlock()
+
+	// Wait for the existing workers to finish.
+	d.wg.Wait()
+
+	// Grab the error from within a lock.
 	d.mu.Lock()
 	defer d.mu.Unlock()
-	if d.closed {
-		return d.err
-	}
-	d.closed = true
-	close(d.closing)
-	d.wg.Wait()
 	return d.err
 }
 
--- a/execute/dispatcher_test.go
+++ b/execute/dispatcher_test.go
@@ -1,0 +1,38 @@
+package execute
+
+import (
+	"context"
+	"testing"
+
+	"go.uber.org/zap/zaptest"
+)
+
+func TestDispatcher_Stop(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	d := newPoolDispatcher(10, zaptest.NewLogger(t))
+	d.Start(100, ctx)
+
+	for i := 0; i < 100; i++ {
+		d.Schedule(func(ctx context.Context, throughput int) {
+			<-ctx.Done()
+			panic("expected")
+		})
+	}
+	cancel()
+
+	if err := d.Stop(); err == nil {
+		t.Fatal("expected error")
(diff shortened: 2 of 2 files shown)


execute/dispatcher.go
```diff
@@ -107,14 +107,23 @@ func (d *poolDispatcher) setErr(err error) {
 
 //Stop the dispatcher.
 func (d *poolDispatcher) Stop() error {
+	// Check if we are already closed and return the error if so.
 	d.mu.Lock()
-	defer d.mu.Unlock()
 	if d.closed {
```
GeorgeMac, line 112:
```suggestion
	if d.closed {
		d.mu.Unlock()
```

GeorgeMac (changes requested):
One niggly little missing unlock.

execute/dispatcher.go
jsternberg: Fixed and added a test.

GeorgeMac (approved):
Yay!

execute/dispatcher.go
```diff
@@ -107,14 +107,24 @@ func (d *poolDispatcher) setErr(err error) {
...
+		d.mu.Unlock()
 		return d.err
```
nathanielc, line 114: why do we not need to have the lock when reading the error here when we do below?

jsternberg: The `setErr` call can set the error in a different goroutine which means the read needs to be protected too.

nathanielc: This code does not have read protection. It releases the lock before reading the error, should the code be:

```
if d.closed {
   err := d.err
   d.mu.Unlock()
   return err
}
```

jsternberg: Oh that didn't occur to me. It should just be `defer d.mu.Unlock()`. Your code works too.

execute/dispatcher.go
```diff
@@ -107,14 +107,24 @@ func (d *poolDispatcher) setErr(err error) {
...
 	}
+	// If not, mark the dispatcher as closed and signal to the current
+	// workers that they should stop processing more work.
 	d.closed = true
 	close(d.closing)
+	d.mu.Unlock()
+
+	// Wait for the existing workers to finish.
 	d.wg.Wait()
```
nathanielc, line 123: Does waiting have a chance to set the err? Does this mean that only the first caller to Stop is guaranteed to get the real error? Since multiple calls to Stop don't wait for the error to be set do we now risk loosing the error?

nathanielc (changes requested):
I am a bit concerned about this change a few thoughts.

Previously the entire Stop method held the lock such that the entire method was atomic. Now the method is not atomic so its possible that multiple callers are blocked at different stages in the method. 

I think its now possible that a call to Stop doesn't return an error even though an error occurred. 

Why is Stop called multiple times and is it ok that secondary callers could miss the error?

jsternberg:
I see what you mean. For context, `Stop()` only appears to be invoked once but I think it's good to keep the idempotent part not broken so let's keep that.

I've updated that section to check if it has not been closed yet. If it hasn't been closed, it closes the channel to signal the close and marks itself as closed so it doesn't do that again. It then goes back to the normal path where it waits on the wait group (wait groups can be waited on multiple times and they are idempotent) and then it locks again to read the error. This way, multiple stop calls will wait on the workers and return errors.

nathanielc (approved):
Looks great, I like the simplicity of the change.

PR #3724, opened 2021-05-17
fix(formatter): formatting for conditionals

BarWeiner1:
### Done checklist
- [ ] Changed else formatting
- [ ] Test cases written
- [ ] Testing file improved for visualization


nathanielc (commented):
Changes look good but I think you need to rebase your changes off the latest master. Then run `make fmt` and `make generate` again.

stdlib/experimental/geo/geo.flux
```diff
@@ -35,7 +35,7 @@ ST_Length = (geometry, units=units) => stLength(geometry: geometry, units: units
 ST_LineString = (tables=<-) => tables
     |> reduce(
         fn: (r, accumulator) => ({
-            __linestring: accumulator.__linestring + (if accumulator.__count > 0 then ", " else "") + string(v: r.lon) + " " + string(v: r.lat),
+            __linestring: accumulator.__linestring + if accumulator.__count > 0 then ", " else "" + string(v: r.lon) + " " + string(v: r.lat),
```
nathanielc, line 38: I think this change is the one causing all the tests to fail. With the latest changes from master the `()` should not be removed and therefore will not break this code.

nathanielc (approved):
Changes look good to me. @scbrickley Can you take a look as well?

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -7,7 +7,11 @@ use pretty_assertions::assert_eq;
 
 fn assert_unchanged(script: &str) {
     let output = format(script).unwrap();
-    assert_eq!(script, output);
+    assert_eq!(
+        script, output,
+        "\n EXPECTED: \n {} \n OUTPUT: \n {} \n",
```
scbrickley, line 12: Since we're using the `pretty_assertions` version of the `assert_eq!` (which will print diff between `script` and `output` if they're not equal), I don't think we need to have this format string here. In most cases, this would be good practice, but in this particular case I think it's unnecessary, and just adds noise to the test output.

scbrickley (changes requested):
One small nitpick, but looks good otherwise!

libflux/flux-core/src/formatter/tests.rs
BarWeiner1: The output for the diff in assert_eq! is generally very hard to read in its current form. I am fine with leaving it, I just think it helps to see the output when changing the formatter logic.

scbrickley: My concern here is that we're breaking a convention that's pretty well established in `libflux`. All of our other tests use just the diff. It can be hard to read at first, but once you get used to it, it's actually pretty helpful for pointing out small differences that would otherwise be hard to spot in plain text. Plus, this doesn't stop us from printing the diff, it just prints out both.

To be clear, if this change helps you read the output better during development, that's fine. I just don't think we should commit it.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1025,3 +1025,61 @@ fn temp_indent() {
...
+            else if r._level == levelInfo then
+                2
+            else if r._level == levelOK then
+                1
+            else
+                0,
+            foo: bar,
+        //Check indent on this test
+        }),
+    //Seems correct running this test
```
scbrickley, line 1047: Can we delete these two comments? This is another instance of something that's fine during development, but that we don't want to commit.

scbrickley (changes requested):
Sorry I missed this during the previous review. One last change, and then I think this is good to merge.

libflux/flux-core/src/formatter/tests.rs
BarWeiner1: Yes, sure but in that case should we also get rid of comments like this (starting line 723):
Screen Shot 2021-05-18 at 3 01 05 PM

scbrickley: I think those comments are fine because in that case they're flux comments and we're testing to see if comments are preserved when formatting. The comments are a bit tongue-in-cheek, but we can let that slide for now :stuck_out_tongue_closed_eyes:

scbrickley: Oh, apologies. I think I misread your code. I see that your comments are comments in the flux code being tested. If the intention was to test the formatter behavior around comments, then they can stay. Otherwise, I think it's best to remove them.

scbrickley (approved):
LGTM - thanks for doing this! Excited to see the new and improved formatter in action.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -7,11 +7,7 @@ use pretty_assertions::assert_eq;
 
 fn assert_unchanged(script: &str) {
     let output = format(script).unwrap();
-    assert_eq!(
-        script, output,
-        "\n EXPECTED: \n {} \n OUTPUT: \n {} \n",
-        script, output
-    );
+    assert_eq!(script, output);
```
nathanielc: @scbrickley I actually think this case is different. I think we should leave these changes in place. The normal diff output of the Rust assert_eq macro flattens the strings onto a single line using escaped newlines and tabs instead of literals. I.e. it looks like this `a = b\nc = d` which can be hard to read, specially when checking for indentation is correct across multiple lines. Since this test case is about testing the formatting I think it makes sense to print the expected and output as string literals without any escaping like this:

```
a = b
c = d
```

Both outputs are valuable since the formatted output makes it easy to visualize that your change worked or didn't. The escaped output makes it easy to find off-by-one errors with whitespace that otherwise are hard to spot. 

Thoughts?

scbrickley: Good points. I guess I hadn't considered that formatting presented special concerns that we don't have in other areas of `libflux`. I'm for adding it back in if you think that's best.

PR #3727, opened 2021-05-18
feat: to function writes to remote influxdb

lesam:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

This was previously merged and then rolled back.

The supporting changes for cloud and OSS are now in, so this can be merged safely.


dependencies/influxdb/http.go
```diff
@@ -441,3 +458,64 @@ func (h seriesCardinalityHttpReader) Read(ctx context.Context, f func(flux.Table
...
+	}()
+	return w, nil
+}
+
+var _ Writer = &httpWriter{}
+
+// Write sends points asynchronously to the underlying write api.
+// Errors are returned only on a best-effort basis.
+func (h *httpWriter) Write(metric ...protocol.Metric) error {
+	buf := new(bytes.Buffer)
```
jsternberg, line 493: I think this ends up functionally the same because the escape checker finds out that it's the same but the normal way to deal with this is:

```
var buf bytes.Buffer
enc := protocol.NewEncoder(&buf)
```

It's fine to leave it as-is.

dependencies/influxdb/http.go
```diff
@@ -65,6 +70,18 @@ func (h HttpProvider) SeriesCardinalityReaderFor(ctx context.Context, conf Confi
 	}, nil
 }
 
+func (h HttpProvider) WriterFor(ctx context.Context, conf Config) (Writer, error) {
+	httpClient, err := h.clientFor(ctx, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	service := apihttp.NewService(httpClient.Config.Host, "Token "+httpClient.Config.Token, apihttp.DefaultOptions().SetHTTPDoer(httpClient.Client))
```
jsternberg, line 79: Is there a way we can contribute upstream so that this `Token mytoken` thing isn't so easy to mess up?

jsternberg (approved):
Looks good to merge. One style comment and one question about the upstream API.

dependencies/influxdb/http.go
lesam: Possibly - but I think it makes sense the way it is. This is a fairly low level of the client API and it is specifying the authentication header. The `NewClient` top level API handles the translation: https://github.com/influxdata/influxdb-client-go/blob/v2.3.0/client.go#L110

dependencies/influxdb/http.go
lesam: In C++/Rust I would expect the two to be different, but in go once you take the address you need the initial allocation on the heap anyway, unless the go escape checker can validate it doesn't escape (e.g. that NewEncoder doesn't store the pointer on the heap somewhere).

lesam: Noted about the preferred style though, thanks.

Issue #3715, opened 2021-05-13
Odd behaviour when going over the Go Rust boundary multiple times using JSON serialization.


skartikey:
**Issue -**
The error was found while running flux fmt CLI
`./flux fmt -w stdlib/csv/csv_test.flux` 
results in Error: failed to format the query: missing field `id` at line 1 column 29124, Filename: stdlib/csv/csv_test.flux

**List of flux files returning same error -** 
stdlib/planner/group_min_test.flux
stdlib/planner/group_min_max_table_test.flux
stdlib/planner/group_max_test.flux
stdlib/csv/csv_test.flux
stdlib/universe/merge_filter_test.flux
stdlib/universe/aggregate_window__offset_test.flux

**Steps to reproduce the error -** 
create a test case that calls `parser.ParseSource` and then `astutil.Format` to reproduce the error.
we already have a test for that - ast/astutil/format_test.go

**Investigation -** 
Problem is in the `TestCaseStmt` statement, `id` & `block` is missing in the flux file because of which we are getting above mentioned error. 
https://github.com/influxdata/flux/blob/8fa2e3f7486518c4e62e3ffadd7f21057474e3ce/libflux/flux-core/src/ast/mod.rs#L922

Fix could be - 

```
#[derive(Debug, Default, PartialEq, Clone, Serialize, Deserialize)]
#[allow(missing_docs)]
pub struct TestCaseStmt {
    #[serde(skip_serializing_if = "BaseNode::is_empty")]
    #[serde(default)]
    #[serde(flatten)]
    pub base: BaseNode,
    _#[serde(default)]_
    pub id: Identifier,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub extends: Option<StringLit>,
    _#[serde(default)]_
    pub block: Block,
}
```

DOD:
- New test for the missing field `id` error has been added to the tests in ast/astutil/format_test.go

PR #3729, opened 2021-05-19
fix: Odd behaviour when going over the Go Rust boundary multiple times using JSON serialization.

skartikey:
In the `TestCaseStmt` statement, `id` & `block` are missing in the flux file because of which we are getting the following error. 
`format_test.go:206: missing field `id` at line 1 column 3335`
` format_test.go:206: missing field `block` at line 1 column 3335 `

Adding default to struct TestCaseStmt fixes the issue. 

fixes: #3715

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -736,9 +736,9 @@ func (s *TestStatement) Copy() Node {
 // TestCaseStatement declares a Flux test case
 type TestCaseStatement struct {
 	BaseNode
-	ID      *Identifier
-	Extends *StringLiteral `json:",omitempty"`
-	Block   *Block
+	ID      *Identifier    `json:"id"`
+	Extends *StringLiteral `json:"extends,omitempty"`
+	Block   *Block         `json:"block"`
 }
 
 // Type is the abstract type
--- a/ast/astutil/format_test.go
+++ b/ast/astutil/format_test.go
@@ -186,6 +186,25 @@ builtin baz : int`; want != got {
 	}
 }
 
+func TestFormatWithTestCaseStmt(t *testing.T) {
+	src := `testcase my_test { a = 1 }`
+	pkg := parser.ParseSource(src)
+	if ast.Check(pkg) > 0 {
+		t.Fatalf("unexpected error: %s", ast.GetError(pkg))
+	} else if len(pkg.Files) != 1 {
+		t.Fatalf("expected one file in the package, got %d", len(pkg.Files))
+	}
+
+	got, err := astutil.Format(pkg.Files[0])
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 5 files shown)


libflux/flux-core/src/ast/mod.rs
```diff
@@ -919,9 +919,11 @@ pub struct TestCaseStmt {
     #[serde(default)]
     #[serde(flatten)]
     pub base: BaseNode,
+    #[serde(default)]
     pub id: Identifier,
```
nathanielc, line 923: What's the default for an Identifier?

nathanielc (changes requested):
Why do we want a default? Should we instead find out how that data is missing in the first place? Is it because either Rust or Go forgets to serialize the `id` and `block` data into the JSON?

libflux/flux-core/src/ast/mod.rs
```diff
@@ -1483,7 +1485,7 @@ pub struct Property {
 }
 
 /// Identifier represents a name that identifies a unique Node
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, Default, PartialEq, Clone, Serialize, Deserialize)]
```
rockstar, line 1488: I don't think I knew you could derive `Default`. What does it return when you call `Indentifier::default()`?

rockstar (changes requested):
I'm not sure this is the right solution. Based on the description of the problem and this patch, I would assume that there is a serialization/deserialization step somewhere on the rust boundary that isn't properly handling `id` and `block` properly. I think that's probably what we need to look at, and figure out why that data isn't getting passed back and forth properly.

skartikey:
@nathanielc @rockstar I have kept this PR as a draft as this is not the final solution. I did narrow down the problem to `TestCaseStmt` where the actual problem is. As mentioned in the standup today, this PR is to make sure I am going in the right direction with the issue. 

Thanks for the suggestions, I will dig more and find out where exactly we are losing the `id` and the `block`.

rockstar (approved):

nathanielc (changes requested):
Looks great, request for a new test see below

ast/ast.go
```diff
@@ -736,9 +736,9 @@ func (s *TestStatement) Copy() Node {
 // TestCaseStatement declares a Flux test case
 type TestCaseStatement struct {
 	BaseNode
-	ID      *Identifier
-	Extends *StringLiteral `json:",omitempty"`
-	Block   *Block
+	ID      *Identifier    `json:"id"`
+	Extends *StringLiteral `json:"extends,omitempty"`
+	Block   *Block         `json:"block"`
```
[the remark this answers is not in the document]
skartikey: I have added the test.

nathanielc (approved):

PR #3730, opened 2021-05-19
feat(ast/astutil): Expose a function for ast.Package formatting

brettbuddin:
Checks and Notifications produce `*ast.Package` structures instead of `*ast.File`. There are still cases where a function that accepts `*ast.File` is useful, but having one that can handle the package is needed. This exposes that behavior so it can be used directly.

### DIFF
--- a/ast/astutil/format.go
+++ b/ast/astutil/format.go
@@ -7,7 +7,7 @@ import (
 	"github.com/influxdata/flux/runtime"
 )
 
-// Format will format the AST to a string.
+// Format will format an AST File to a string.
 func Format(f *ast.File) (string, error) {
 	pkg := &ast.Package{
 		Files: []*ast.File{f},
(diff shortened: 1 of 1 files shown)


jsternberg:
I'm a bit concerned about the API here. The reason why I changed the format function to only take an `*ast.File` is because formatting a package, which is inherently multiple files, to a single string doesn't make a lot of sense. It seems to me like it's easily possible to call format multiple times if there are multiple files and then the caller can make a decision of how to handle multiple files. It seems to me like `FormatPackage()` has no way of knowing how to deal with multiple files and will just be forced to concatenate them together which constructs a file that can potentially not be parsed. This is highly likely since it will generate multiple package statements into the same file.

Does this make sense and does the alternative mentioned here help with what you are working on?

brettbuddin:
@jsternberg That makes a lot of sense. I was worried I was missing something. It seems to me like we just need to address this problem at the source in IDPE and produce `*ast.File`s instead of `*ast.Package`s. I'll take a look today and see how much work that'll be for us. For now I've done a really hacky thing of copying a portion of this `Format` function into IDPE to deal with the problem. If moving to `*ast.File` becomes an issue, I might need to move forward with that instead of introducing this here and re-address that as a separate pass.

PR #3731, opened 2021-05-19
docs(libflux): remove unneeded `#![allow(missing_docs)]` attributes

pierwill:
Removes the last module-level `#![allow(missing_docs)]` attributes for `fluxcore`.

All undocumented public items are now marked `#[allow(missing_docs)]`. Most of these we've chosen to leave without docs, such as enum variants.


libflux/flux-core/src/semantic/flatbuffers/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 #[allow(clippy::all, missing_docs)]
 pub mod semantic_generated;
+#[allow(missing_docs)]
```
pierwill, line 5: Note: I've added this here so we can remove the `#!...` attribute from `libflux/flux-core/src/semantic/flatbuffers/types.rs`.

rockstar (approved):

libflux/flux-core/src/semantic/flatbuffers/mod.rs
rockstar: It's kind of an unneeded change, yes? Unless I'm missing something, they're equivalent. Six of one, half a dozen of the other. Since you're doing so much heavy lifting on writing docs, I defer to you on how you would like them structured.

pierwill: They're equivalent, yes. The reason I changed this is that I like to be able to do this

```
$ cd flux/libflux/
$ rg "missing_doc" | grep "!"
flux-core/src/lib.rs:#![cfg_attr(feature = "strict", deny(warnings, missing_docs))]
```

and see that our crate is free of `#![allow(missing_docs)]`! 😃

rockstar: Boom, and that's why I felt good about leaving it to you.

PR #3732, opened 2021-05-19
docs(libflux): add docs test for `format`

pierwill:
Mark other documentation tests with `no_run`.

I've added `cargo test --doc` to `test-rust` to the make target. This is run in CI, yes? I'm happy to remove this part if we'd rather have just the docs change for now.


Makefile
```diff
@@ -110,7 +110,9 @@ test-go: libflux-go
 	$(GO_TEST) $(GO_TEST_FLAGS) ./...
 
 test-rust:
-	cd libflux && $(CARGO) test $(CARGO_ARGS) && $(CARGO) clippy $(CARGO_ARGS) -- -Dclippy::all
+	cd libflux && $(CARGO) test $(CARGO_ARGS) && \
+	$(CARGO) test --doc && \
```
pierwill, line 114: From the [cargo book](https://doc.rust-lang.org/cargo/commands/cargo-test.html):

> --doc
> Test only the library's documentation. This cannot be mixed with other target options.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -32,6 +32,15 @@ pub fn convert_to_string(file: &File) -> Result<String, Error> {
 }
 
 /// Format a string of Flux code.
+///
+/// # Example
+///
+/// ```rust
+/// # use fluxcore::formatter::format;
+/// let source = "(r) => r.user ==              \"user1\"";
```
rockstar, line 40: Did the formatter do this?

pierwill: This is intentional so that we can run `fluxcore::formatter::format`on `source` and get the formatted Flux code.

pierwill: The `#` is to hide boilerplate code in the rendered docs. But perhaps the `use` statement would be nice to show here...

rockstar: Yeah, sorry. The reason I asked is that we have a current conversation going on around horizontal alignment, so wondered if we had another case.

rockstar (approved):

PR #3736, opened 2021-05-19
fix: audit and clean up docker build image

### LINKED ISSUE #3721: Rust installation needs more security verification
In our docker image build, we do a `curl | sh`. We want to avoid that, as it has all sorts of security implications. We'd like something that doesn't require us to manually update any time there is a new release of rust and/or rustup. There is prelim work [here](https://github.com/influxdata/flux/pull/3710) but it is prudent to have our team weigh in on a number of other options and explore ways that we can accomplish this.

DOD:
  - rust is installed without using `curl | sh` in the process.

rockstar:
This patch does a number of things in the spirit of cleaning up and
simplifying the build system. It does the following:

  - Use the upstream rust image to get rust bootstrap tooling, including
    `rustc`, `cargo`, and `rustup`. This was the driving force behind
    this patch, and fixes #3721.
  - Removes the superseded ragel 6 install. We originally installed
    ragel 6 and 7 in parallel so we had a migration path. We are now
    fully using ragel 7 for our build, so ragel 6 is no longer needed.
  - Remove support for cross compiling. The InfluxDB edge team has their
    own setup for cross compiling, so we no longer need to maintain the
    tooling for doing cross compiled binaries. This includes changes to
    the goreleaser configs so that releases don't try to build those
    binaries anymore.

### DIFF
--- a/Dockerfile_build
+++ b/Dockerfile_build
@@ -1,37 +1,44 @@
 # This file describes an image that is capable of building Flux.
 
+# "Wait," you ask. "What's going on here?" Rather than handling rustup validation
+# and verification, we can list the rust container as a prior build stage, and
+# then pull in the artifacts we need. There is an added benefit that tagged versions
+# also include minory releases, so 1.2 includes 1.2.1 and so on, for bugfix releases.
+FROM rust:1.52 as RUSTBUILD
+
 FROM golang:1.16
 
 # Install common packages
 RUN apt-get update && \
     apt-get install --no-install-recommends -y \
-    gcc-arm-linux-gnueabihf libc6-dev-armhf-cross \
-    gcc-aarch64-linux-gnu libc6-dev-arm64-cross \
-    libtinfo5 \
-    ruby \
+    openssl libtinfo5 ruby \
     ca-certificates curl file gnupg \
     build-essential cmake \
     libxml2-dev libssl-dev zlib1g-dev \
     autoconf automake autotools-dev libtool xutils-dev valgrind && \
     rm -rf /var/lib/apt/lists/*
 
-# Download, verify, and install ragel version 6. 
-ENV RAGEL_VERSION=6.10
-RUN curl https://www.colm.net/files/thurston.asc | gpg --import - && \
-    curl https://www.colm.net/files/ragel/ragel-${RAGEL_VERSION}.tar.gz -O && \
-    curl https://www.colm.net/files/ragel/ragel-${RAGEL_VERSION}.tar.gz.asc -O && \
-    gpg --verify ragel-${RAGEL_VERSION}.tar.gz.asc ragel-${RAGEL_VERSION}.tar.gz && \
-    tar -xzf ragel-${RAGEL_VERSION}.tar.gz && \
-    cd ragel-${RAGEL_VERSION}/ && \
-    ./configure --prefix=/usr/local && \
-    make && \
-    make install && \
-    cd .. && rm -rf ragel-${RAGEL_VERSION}*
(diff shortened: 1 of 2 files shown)


Dockerfile_build
```diff
@@ -1,48 +1,53 @@
...
 ENV COLM_VERSION=0.14.2
-RUN curl https://www.colm.net/files/colm/colm-${COLM_VERSION}.tar.gz -O && \
+ENV RAGEL7_VERSION=7.0.1
+RUN curl https://www.colm.net/files/thurston.asc | gpg --import - && \
+    curl https://www.colm.net/files/colm/colm-${COLM_VERSION}.tar.gz -O && \
     curl https://www.colm.net/files/colm/colm-${COLM_VERSION}.tar.gz.asc -O && \
     gpg --verify colm-${COLM_VERSION}.tar.gz.asc colm-${COLM_VERSION}.tar.gz && \
     tar -xzf colm-${COLM_VERSION}.tar.gz && \
     cd colm-${COLM_VERSION}/ && \
     ./configure --prefix=/usr/local/ragel7 --disable-manual && \
```
adrian-thurston, line 46: I notice you're still using the `ragel7` prefix. That works for me, but I just want to mention that ragel 7 is intended to be backwards compatible with 6, so it is *okay* to replace 6 with 7 by installing to the same location., which was `/usr/local/`. The unique prefix was just so they could co-exist. But I see cargo is also a unique prefix ...

adrian-thurston (approved):

Dockerfile_build
rockstar: I was hoping you'd say something about that. I did try to take the prefix out, but it was causing build issues that I just didn't want to chase much more at the expense of the good in this patch.

adrian-thurston: Do you remember the nature of the issues? Like building ragel, or building flux?

rockstar: Building ragel, with some interaction with colm, I believe. I didn't look too deeply at it, and it's probably just a pathing issue. It was more that building the image takes ~20 minutes, so poking at it could still take me another hour at its simplest to fix, and I just didn't want to devote it.

adrian-thurston: Oh possibly `colm` wasn't available to build `ragel`, or something like that. Things are all good as they are. Not the worst to segregate the applications.

jsternberg (approved):
Looks good to me. Are we concerned that removing the cross compilation may make it more difficult to find cross compilation issues before they hit Influxdb?

rockstar:
> Looks good to me. Are we concerned that removing the cross compilation may make it more difficult to find cross compilation issues before they hit Influxdb?

I'm not. The edge team really seems to be on top of things there, and the only time we'd see it is on release anyway. We're currently releasing binaries that don't work, but the issue we have currently was fixed by the edge team downstream in InfluxDB.

Dockerfile_build
```diff
@@ -1,48 +1,53 @@
 # This file describes an image that is capable of building Flux.
 
+# "Wait," you ask. "What's going on here?" Rather than handling rustup validation
+# and verification, we can list the rust container as a prior build stage, and
+# then pull in the artifacts we need. There is an added benefit that tagged versions
+# also include minory releases, so 1.2 includes 1.2.1 and so on, for bugfix releases.
+FROM rust:1.52 as RUSTBUILD
```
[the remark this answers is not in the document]
pierwill: Awesome! Out of curiosity, is this the Rust container maintained here? https://github.com/rust-lang/docker-rust

PR #3744, opened 2021-05-20
fix(execute): log when uneven columns are encountered when processing tables

jsternberg:
This produces a log message when a transformation is about to process a
column reader within a table that has uneven columns without aborting
execution.

This change only affects the transformation that receives the table. It
does not affect yields and it does not print the origin transformation
as it would have been difficult to get the logger threaded into the area
where the table itself is produced.

Instead, we wrap the table and validate each column reader as it is
being processed by the downstream.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/execute/transport.go
+++ b/execute/transport.go
@@ -9,9 +9,11 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute/table"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
+	"go.uber.org/zap"
 )
 
 type Transport interface {
@@ -23,6 +25,7 @@ type Transport interface {
 // consecutiveTransport implements Transport by transporting data consecutively to the downstream Transformation.
 type consecutiveTransport struct {
 	dispatcher Dispatcher
+	logger     *zap.Logger
 
 	t        Transformation
 	messages MessageQueue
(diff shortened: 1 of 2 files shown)


scbrickley (approved):
LGTM - Thanks for doing this!

execute/transport.go
```diff
@@ -360,3 +364,71 @@ func (m *finishMsg) Type() MessageType {
...
+func (t *consecutiveTransportTable) Cols() []flux.ColMeta {
+	return t.tbl.Cols()
+}
+
+func (t *consecutiveTransportTable) Do(f func(flux.ColReader) error) error {
+	return t.tbl.Do(func(cr flux.ColReader) error {
+		if err := t.validate(cr); err != nil {
+			logger := t.transport.logger
+			logger.Info("Invalid column reader received from parent",
+				zap.String("source", t.transport.sourceInfo()),
```
scbrickley, line 395: What exactly is included in the sourceInfo here?

jsternberg: It's the line number information of the transformation that processed the bad table. So if you have `distinct() |> max()` and distinct produces a bad table, it will print `max` and the line number.

scbrickley: And this will show up in the query log, so we'll have a way of identifying what query it came from, correct?

jsternberg: It will but we'll have to search through where the query came from. I'm going to try out another change and push it up. Brett added the ability to access the jaeger trace id recently and I think I can probably use the same code path.

jsternberg: I'm double checking with @brettbuddin but I think this will work. According to him in a DM, it should only work in the cloud product so it doesn't work if you try it in InfluxDB OSS.

brettbuddin (approved):

PR #3748, opened 2021-05-20
fix(values): allows IEEE float divide by zero math

### LINKED ISSUE #3506: Float division by zero doesn't work as expected
[According to the specification](https://docs.influxdata.com/influxdb/v2.0/reference/flux/language/types/#numeric-types), floats in Flux are IEEE-754 64-bit floating point numbers. In most ways, they behave as such. However, division by zero is not handled correctly.

Flux's standard library even has the functions `math.NaN()` and `math.mInf()`, for creating not-a-number and positive or negative infinities respectively. But, for some reason, `1.0/0.0`, `-1.0/0.0` and `0.0/0.0` - which should return the same values as `math.mInf(sign: 1)`, `math.mInf(sign: -1)`, and `math.NaN()`, respectively - all instead result in a runtime error, `cannot divide by zero`.

nathanielc:
The IEEE specifies the behavior of divide by zero for floating point
values. Previous to this change we returned an error, which would stop
the entire query, for divide by zeros errors. With this change divide by
zero is allowed for floating point values as per the IEEE spec.

Integer divide by zero still results in an error as there is no
definition that behavior. The behavior of Flux now mirrors the Go
behavior. Go panics on divide by zero for integer types while floating 
point divide by zero follows the spec returning +/-Infinity.

Fixes #3506

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


values/binary_test.go
```diff
@@ -38,6 +39,10 @@ func TestBinaryOperator(t *testing.T) {
 		// float + float
 		{lhs: 4.5, op: "+", rhs: 8.2, want: 12.7},
 		{lhs: 4.5, op: "+", rhs: floatNullValue, want: floatNullValue},
+		{lhs: 1.0, op: "/", rhs: 0.0, want: math.Inf(1)},
```
jsternberg, line 42: How does the CSV decoder handle infinite float values? I suspect it's likely fine since I think it uses the Go standard library to both output and parse the values, but I'd like us to add a test for this and modify the spec to explicitly list this behavior if it isn't already.

jsternberg (changes requested):
Minor documentation comment. I like the change and think this is the correct behavior change, but I just want us to do our due diligence of testing the CSV encoder/decoder before this is considered good to go.

Along with the CSV encoder/decoder, how does this affect the `to()` implementation for InfluxDB?

nathanielc:
OK I have added a bunch of tests:

Tests for `to()` behavior - Writing +/- Inf and NaN are not support by storage
Test for `csv.from()` parsing +/- Inf and NaN works as expected
Tests for the `toFloat()` family of conversion functions, these tests mostly existed already and I just converted them to acceptance tests (i.e. `testcase`)

jsternberg (approved):

Issue #3682, opened 2021-05-04
Formatting: line separation between the blocks are getting removed by the formatter


skartikey:
example - stdlib/experimental/join_right_side_more_cols_test.flux 
Before - 
```
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119554,2020-10-09T22:19:00Z
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119586.6,2020-10-09T22:19:44.191958Z"

outData = "
#group,false,false,true,false,false,false,false,false,false,false,false,false
```
After - 

```
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119554,2020-10-09T22:19:00Z
,,5,inodes_used,disk,disk1s1,apfs,ip-192-168-1-16.ec2.internal,rw,/System/Volumes/Data,4119586.6,2020-10-09T22:19:44.191958Z"
outData = "
#group,false,false,true,false,false,false,false,false,false,false,false,false
```

PR #3749, opened 2021-05-21
fix(formatter): took out erroneous line deletions (Issue #3682)

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


BarWeiner1:
Any reviewers look closely at the changes we made to formatter/tests.rs to make sure the formatting changes are acceptable.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -727,18 +729,21 @@ impl Formatter {
...
-                if cur != prev || starts_with_comment(Node::from_stmt(&stmt)) {
+                let current_location: i32 = stmt.base().location.start.line as i32;
+                //compare the line position of adjacent lines to preserve formatted double new lines
+                let line_gap = current_location - previous_location;
+                // separate different statements with double newline or statements with comments and preserve double newline format
+                if line_gap > 1 || cur != prev || starts_with_comment(Node::from_stmt(&stmt)) {
                     self.write_rune(sep);
                 }
             }
+            previous_location = stmt.base().location.end.line as i32;
```
wolffcm, line 746: I would expect that the rules for formatting a file will be the same as the rules for formatting a block. Can you make them call the same common function, so this logic is not duplicated?

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -226,22 +225,25 @@ impl Formatter {
...
             if i != 0 {
+                let current_location: i32 = stmt.base().location.start.line as i32;
+                //compare the line position of adjacent lines to preserve formatted double new lines
+                let line_gap = current_location - previous_location;
                 self.write_rune(sep);
                 // separate different statements with double newline or statements with comments
-                if cur != prev || starts_with_comment(Node::from_stmt(&stmt)) {
+                if line_gap > 1 || cur != prev || starts_with_comment(Node::from_stmt(&stmt)) {
                     self.write_rune(sep);
                 }
```
wolffcm, line 240: This is the kind of thing that reasonable people may disagree on, i.e. a [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) kind of issue. Personally I like the logic here but I would make it so we don't add a blank line between different kinds of statements. E.g., this looks fine to me:
```js
fun = (a, b) => {
  c = a * b
  return c
}
```
I don't think it helps readability to put a blank line between the two statements in the function body.

This is a different change from the issue you're fixing here, so no need to fix it now.

wolffcm (changes requested):
Hi Bar and Noe,

I like this change, I just think if you can combine the logic for formatting blocks and statements at file-level, it would be helpful. That's the change I'm requesting.

As far as bookkeeping, it's helpful if you can put a link to the issue you're fixing in the header comment for the PR. That way reviews can see the issue you're fixing and get some context.

This isn't passing CI yet because there's a commit that is not "semantic". This can be fixed by doing `git rebase -i` and changing the commit message to something like `chore: ...`, etc.

libflux/flux-core/src/formatter/mod.rs
adrian-thurston: I agree, the `cur != prev` should come out after this change is in. So long as gaps are preserved where they were originally, there is no need to insert whitespace if the statement types differ.

libflux/flux-core/src/formatter/mod.rs
adrian-thurston: Agree here as well, the line separation logic is duplicated and can be made common. In one of them, the first separator is only written in iterations 1 and up, and in the other it is written in all. That can be moved out of the common loop.

Other than that, looks good!

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -727,29 +713,41 @@ impl Formatter {
...
             self.write_rune(sep);
+            self.unindent();
+            self.write_indent()
+        }
+        self.format_comments(&n.rbrace);
+        self.write_rune('}')
+    }
 
+    fn statement_list(&mut self, n: &[Statement]) {
+        // &Vec<Statement>
```
wolffcm, line 730: Can you change the name to `format_statement_list` to be consistent with the other methods that format things?
And also can this comment line be deleted?

wolffcm (changes requested):
I just had a couple minor requests but this looks good!

wolffcm (approved):

PR #3752, opened 2021-05-24
test(stdlib): test range to nsec precision - bare and group cases

### LINKED ISSUE #3668: Flux should have acceptance tests for the exact boundaries of range, especially combined with last()
According to https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/built-in/transformations/range/ :

> Results include records with _time values greater than or equal to the specified start time and less than the specified stop time.

The acceptance tests in flux should verify that this is true (that ranges are implemented to select the correct nanosecond) so that all the implementing storage engines can be confirmed to have the correct behaviour.

Without these tests it is possible for off-by-one bugs to creep in to the various storage engines. I would be especially concerned about the `Range` pushdowns, optimisations for `last`, and windowing pushdowns.

fchikwekwe:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

Adding [the test coverage requested here](https://github.com/influxdata/flux/issues/3668) by @lesam. 

Closes #3668

### DIFF
--- a/stdlib/universe/range_nsecs_test.flux
+++ b/stdlib/universe/range_nsecs_test.flux
@@ -1,0 +1,357 @@
+package universe_test
+
+
+import "array"
+import "testing"
+
+option now = () => 2030-01-01T00:00:00Z
+
+testcase range_nsecs_bare_last {
+    input = array.from(
+        rows: [
+            {_time: 2021-01-01T00:00:01.000000001Z, _field: "foo", _value:   1.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000002Z, _field: "foo", _value:   2.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value:   3.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000024Z, _field: "foo", _value: 11.24, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000009Z, _field: "bar", _value:   4.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000011Z, _field: "bar", _value:   5.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000015Z, _field: "bar", _value:   6.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000022Z, _field: "bar", _value:   1.2, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000031Z, _field: "bar", _value:  0.28, _measurement: "mm"},
+        ],
+    )
+        |> group(columns: ["_field", "_measurement"])
+        |> testing.load()
+    want = array.from(
+        rows: [
+            {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value: 3.0, _measurement: "mm"},
+            {_time: 2021-01-01T00:00:01.000000011Z, _field: "bar", _value: 5.0, _measurement: "mm"},
+        ],
+    )
+        |> group(columns: ["_field", "_measurement"])
+    got = input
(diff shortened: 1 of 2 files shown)


stdlib/universe/range_nsecs_table_test.flux
```diff
@@ -0,0 +1,132 @@
...
+
+,,1,2021-01-01T00:00:01.000000009Z,bar,4.0,mm
+,,1,2021-01-01T00:00:01.000000011Z,bar,5.0,mm
+,,1,2021-01-01T00:00:01.000000015Z,bar,6.0,mm
+,,1,2021-01-01T00:00:01.000000022Z,bar,1.2,mm
+,,1,2021-01-01T00:00:01.000000031Z,bar,0.28,mm
+"
+
+testcase range_nsecs_bare_last {
+    result = testing.loadStorage(csv: inData)
```
rockstar, line 30: Could we refactor this to use `array.from` and `testing.load` rather than `testing.loadStorage`? I think we want to EOL `testing.loadStorage` and it's reliance on csv.

rockstar (commented):
These changes look good technically, though I think we should start moving away from using csv in our tests unless we're specifically testing the csv interface. `testing.load` is well-supported now, so I feel confident asking us to make that change.

stdlib/universe/range_nsecs_table_test.flux
fchikwekwe: Yeah I spent and hour last week trying to use array.from with these tests and switched to csv when i got sufficiently frustrated.

fchikwekwe: I hate writing tests with csv.from as much as the next guy, but when I wrote these tests in array.from initially, I couldn't get them working at all on my own. Maybe we need a guide for the new test suite or something?

fchikwekwe: Its updated now. Lemme know what you think.

fchikwekwe:
@rockstar I added the second half of the tests to this PR. I hope you don't mind reviewing the rest of the cases. :)

stdlib/universe/range_nsecs_test.flux
```diff
@@ -0,0 +1,353 @@
...
+        {_time: 2021-01-01T00:00:01.000000002Z, _field: "foo", _value: 2.0,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value: 3.0,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000024Z, _field: "foo", _value: 11.24,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000009Z, _field: "bar", _value: 4.0,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000011Z, _field: "bar", _value: 5.0,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000015Z, _field: "bar", _value: 6.0,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000022Z, _field: "bar", _value: 1.2,_measurement: "mm"}, 
+        {_time: 2021-01-01T00:00:01.000000031Z, _field: "bar", _value: 0.28,_measurement: "mm"},
+    ],
+ ) |> group(columns: ["_field", "_measurement"])
```
rockstar, line 23: We probably need to add a `|> testing.load()` here. For flux, that's the equivalent of not having it, but in order for this to be an acceptance test, we need to make sure that these tests are operating against the storage parts of influxdb and cloud. Otherwise, it won't tell us if there's an issue in one of those environments.

rockstar (commented):
I'm good with these changes, though I think we do need to add those `|> testing.load()` calls to the `input` parts, otherwise we'll only get flux testing out of this, which isn't really the spirit of these tests.

rockstar (approved):

PR #3754, opened 2021-05-24
fix: staticcheck linter failures

skartikey:
Fixed following static check linter failures -

1) should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)
2) unnecessary use of fmt.Sprintf (S1039)

### DIFF
--- a/plan/physical.go
+++ b/plan/physical.go
@@ -273,7 +273,9 @@ func CreatePhysicalNode(id NodeID, spec PhysicalProcedureSpec) *PhysicalPlanNode
 	}
 }
 
-const NextPlanNodeIDKey = "NextPlanNodeID"
+type nodeIDKey string
+
+const NextPlanNodeIDKey nodeIDKey = "NextPlanNodeID"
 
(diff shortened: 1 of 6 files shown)


plan/physical.go
```diff
@@ -273,7 +273,9 @@ func CreatePhysicalNode(id NodeID, spec PhysicalProcedureSpec) *PhysicalPlanNode
 	}
 }
 
-const NextPlanNodeIDKey = "NextPlanNodeID"
+type nodeIDKey string
+
+const NextPlanNodeIDKey nodeIDKey = "NextPlanNodeID"
```
jsternberg, line 278: While you're here, can you change the casing on this so it is unexported?

`nextPlanNodeIDKey` instead.

jsternberg (approved):

plan/physical.go
skartikey: This is getting imported here https://github.com/influxdata/flux/blob/master/internal/spec/build.go#L49
so I will leave it as it is.

PR #3766, opened 2021-05-27
fix: Trailing dollar in string expression

### LINKED ISSUE #3717: String with interpolation expression and traling dollar
```
regexp.compile(v: "^${v.foo}-bar$")
```

-> 
```
compilation failed: error at @1:10-1:58: expected RPAREN, got EOF
```

Workarounds:

```
dollar = "$"
regexp.compile(v: "^${v.foo}-bar${dollar}")
```

```
regexp.compile(v: "^${v.foo}-bar\x24")
```

This also works for regexps:

```
regexp.compile(v: "^${v.foo}-bar$$")
```

mkmik:
Closes #3717 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux-core/src/scanner/scanner.rl
```diff
@@ -23,6 +23,7 @@ use crate::scanner::*;
         // so we can record the location of the first byte of a token.
         last_newline_before_token = *last_newline;
         cur_line_token_start = *cur_line;
+        cur_line_token_start = *cur_line;
```
mkmik, line 26: it looks like just a whitespace change

nathanielc (commented):
Thanks, we will get this reviewed.

libflux/flux-core/src/scanner/scanner.rl
nathanielc: i think the line has been duplicated see https://github.com/influxdata/flux/blob/d387dc4e2866ee90917af934817f8a3fe1cb7678/libflux/flux-core/src/scanner/scanner.rl#L26 can you remove the duplicate?

mkmik: Oops. Will do when I get back at a keyboard tomorrow.

mkmik: fixed

libflux/flux-core/src/scanner/scanner.rl
```diff
@@ -46,7 +46,9 @@ use crate::scanner::*;
     escaped_char = "\\" ( "n" | "r" | "t" | "\\" | '"' | "${" );
     unicode_value = (any_count_line - [\\$]) | escaped_char;
     byte_value = "\\x" xdigit{2};
-    dollar_value = "$" ( any_count_line - "{" );
+    # when the dollar is in the last position before the closing quotes
+    # we need to match the quote char but not consume it.
+    dollar_value = "$" ( any_count_line - [{"] | "\"" %{ fhold; } );
     string_lit_char = ( unicode_value | byte_value | dollar_value );
     string_lit = '"' string_lit_char* "$"? :> '"';
```
adrian-thurston, line 53: The dollar is already allowed at the end of `string_lit`, so permitting it in `string_lit_char` is redundant. The other thing about this approach is that using an `fhold` in a grammar fragment that is included in other places is risky, because the equivalent state machine is no longer consuming the strings specified by the grammer. It then becomes hard to reason about. The `fhold` trick should be used sparingly, and generally at the top-most level of a grammar.

It's correct to exclude `$"` from dollar value though, so I would leave that there. That would look like:

```
dollar_value = "$" ( any_count_line - [{"] );
```

To get correct handling of `"ab$"` from the parser, instead I would permit dollar at the end of a string in the `string_expr` scanner by adding the following:

```
'$"' => { fhold; te -= 1; tok = TokenType::Text; fbreak; };
```

By doing it this way, the scanner pattern for string expressions and the more detailed string expression scanner are taking the same approach: allow a single dollar at the end. We have to add `te -= 1` here because the text matched is taken from `te` (token end), which has already been derived from `p` at this point. So we need to update it too.

adrian-thurston (changes requested):

libflux/flux-core/src/scanner/scanner.rl
mkmik: makes sense; thanks!

mkmik: I'll take another stab; something is not quite working out

mkmik:
@adrian-thurston , @nathanielc  can somebody own this; I don't have bandwith to apply the review comments (my first attempt sadly didn't work out)

adrian-thurston (approved):

nathanielc (approved):
Thanks @mkmik for the initial work and @adrian-thurston for getting this across the line!

PR #3773, opened 2021-05-28
fix(formatter): fixed multiline collapse in func params

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/contrib/bonitoo-io/alerta/alerta.flux
```diff
@@ -20,7 +20,22 @@ import "strings"
...
+        severity,
+        service=[],
+        group="",
+        value="",
+        text="",
+        tags=[],
+        attributes,
+        origin="InfluxDB",
+        type="",
+        timestamp=now()) => {
```
fchikwekwe, line 38: @BarWeiner1 @scbrickley I'm guessing this is the desired output of this PR? That function params are each put on their own line?

ast/astutil/format_test.go
```diff
@@ -90,8 +90,7 @@ func TestFormatWithCommentsParens(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if want := `
-// comment\n(1 * 1)`; want != got {
+	if want := `// comment\n(1 * 1)`; want != got {
```
fchikwekwe, line 93: Is this a change that was added to the formatter or just a go format change?

stdlib/testing/kapacitor/eval_map_with_test.flux
```diff
@@ -33,4 +33,3 @@ test _map = () => ({input: testing.loadStorage(csv: inData), want: testing.loadM
 // stream 
 //  |eval(lambda: 2 * _value)
 //.as('_newValue')
-
```
fchikwekwe, line 36: Was this a formatter change?

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1169,3 +1175,40 @@ test2 = 2",
...
+        component="",
+        severity,
+        eventClass="",
+        eventClassKey="",
+        collector="",
+        message="") => {
+    body = json.encode(v: payload)
+
+    return http.post(headers: headers, url: url, data: body)
+}"#,
```
fchikwekwe, line 1212: This is great. Do we have any tests that explicitly check that single line funcs are changed to this format? Is that something easy to test?

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -516,16 +519,33 @@ impl Formatter {
 
     fn format_function_expression(&mut self, n: &ast::FunctionExpr) {
         self.format_comments(&n.lparen);
+        let multiline = n.params.len() > 4 && n.base.is_multiline();
```
fchikwekwe, line 522: 4 seems like a good minimum to change this. Have we considered func signature length as well or was that idea rejected?

fchikwekwe (commented):
Questions about testing and just some clarity on things included in the PR. Otherwise this looks good to me.

stdlib/contrib/bonitoo-io/alerta/alerta.flux
BarWeiner1: Yeah this is desired behavior

ast/astutil/format_test.go
BarWeiner1: We made this go format change, we deemed it acceptable behavior as it stands.

stdlib/testing/kapacitor/eval_map_with_test.flux
BarWeiner1: Yeah I assume so, it just stripped an empty line. Seems fine right?

libflux/flux-core/src/formatter/tests.rs
BarWeiner1: I don't think we have an assert format but by assessing the changes in the other flux files I think we can conclude that it works.
Screen Shot 2021-06-01 at 8 16 12 PM

libflux/flux-core/src/formatter/mod.rs
BarWeiner1: I'm not sure, I think 4 is standard throughout the stdlib, we just emulated it.

libflux/flux-core/src/formatter/tests.rs
scbrickley: @BarWeiner1 It would be helpful to have the kind of test that faith is talking about, if we don't already. That way we can be sure that we don't accidentally regress this behavior with future formatter changes.

ast/astutil/format_test.go
fchikwekwe: Cool! if its the go formatter, then it seems fine.

stdlib/testing/kapacitor/eval_map_with_test.flux
fchikwekwe: Hmm, its a little weird to strip out the final newline in a file. Usually linters prefer that there's at least one newline at the end of the file. But its outside the scope of this pr, so i'd disregard.

libflux/flux-core/src/formatter/tests.rs
fchikwekwe: Yeah I'd like to see some tests added here: 1) that checks that a function with 4 or more params gets reformatted and 2) that checks that a function with 3 or fewer params does not get reformatted.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1211,4 +1211,39 @@ fn preserve_multiline_test() {
...
+        fn,
+        as,
+        every,
+        defaultValue,
+        tables=<-) => {
+    _column = column
+    _as = as
+
+    return tables
+        |> aggregateWindow(every: every, fn: fn, column: _column, createEmpty: true)
```
scbrickley, line 1235: We've defined behavior for when `params > 4` and for when `params < 4`. What do we expect to happen here, where `params == 4`?

BarWeiner1: Well aggregate window has 4 params and remains the same so I think our logic for only separating when params > 4 holds.

scbrickley: Great, just checking. You may want to clarify that in the comment on line 1240 (i.e., replace `<` with `<=`)

scbrickley: Or just nix the comments altogether.

BarWeiner1: Ok I'll update that

stdlib/experimental/usage/usage.flux
```diff
@@ -12,7 +12,13 @@ import "http"
...
 // Note that unlike the range function, the stop argument is required here,
 // pending implementation of https://github.com/influxdata/flux/issues/3629.
-from = (start, stop, host="", orgID="", token="", raw=false) => {
+from = (
+        start,
+        stop,
+        host="",
+        orgID="",
+        token="",
+        raw=false) => {
```
scbrickley, line 21: This is just a personal preference, but in this case I would expect the right paren and everything after it to get moved to the next line, and for a comma to be appended to line 21. Maybe that's outside of the scope of this PR though. Something to bring up with the team perhaps?

```suggestion
from = (
        start,
        stop,
        host="",
        orgID="",
        token="",
        raw=false,
) => {
```

scbrickley (commented):
This looks good for the most part. I have one minor stylistic concern about the new formatting changes, and I'd like to get the input of others on the team before we merge this.

stdlib/experimental/usage/usage.flux
scbrickley: @nathanielc I don't know if we should go through the effort to implement this, or if we should just file a separate issue and let this PR merge. Thoughts?

BarWeiner1: This would be a very easy (1 line) logic fix. We should just ask the team what they prefer and I can push the change if needed.

nathanielc: Yes lets change this to add a newline after the last parameter. The  precedent in Flux has been to make all lines in a list the same including the last line, So for example arrays and records format with the last element and key/value pair on their own line and the closing delimiter on its own line. 

```js
a = [
   1,
   2.
   3,
]
b = {
    a: 1,
    b: 2,
    c: 3,
}
````

BarWeiner1: Ok sounds good.

fchikwekwe (approved):
I think this looks fine to me. For future reference, you should try and squash all your commits into as few as possible (usually just one) once your PR is ready for review. Not a huge deal since we squash them when we merge anyway, but that is a process thing that we usually employ on this team. It also makes the history a bit cleaner since you can get rid of unnecessary commit messages. If you need help understanding how to squash commits, feel free to reach out.

fchikwekwe:
In this case, since sean had a commit as well, you'd squash your commits and leave his separate so that you have two commits to merge into master.

scbrickley:
Feel free to squash my commit as well, if you'd like. That one was also just a squash commit that I made when I was helping resolve some git issues.

Issue #3631, opened 2021-04-19
Acceptance tests don't seem to catch different timestamp


wolffcm:
Recently a commit landed on master that had an incorrect timestamp in the expected output for an acceptance test:
https://github.com/influxdata/flux/commit/91fabec6bef70b8714354843930ecee3c779219b#diff-6219efcaaa8418d87ac1f06b26071baf9786e1dd4f9cd6687246fd46b1192ca8R202
The timestamp should have an hour of `00` not `01`.

I plan on fixing this discrepancy in this PR:
https://github.com/influxdata/flux/pull/3630

However, I verified that with the discrepancy in place, the test still passed when I ran it locally, and seems to sometimes pass in CI. It failed here:
https://app.circleci.com/pipelines/github/influxdata/flux/8968/workflows/e7d350b6-f390-43c8-9b0b-032a28c2a87f/jobs/25323
But it passed here:
https://app.circleci.com/pipelines/github/influxdata/flux/8967/workflows/8d2479b9-f5e9-461b-af3f-88bbd4b7cecd/jobs/25318

Steps to reproduce:
- Make the timestamp change described above, that is, make the inverse of this change:
  - https://github.com/influxdata/flux/pull/3630/files#diff-161f1c8e8bb64b2aa78662e9fc81e37bafb18780279040193990be52c768899fL276-L278
- Run the acceptance tests via `make test-flux`
- Test will pass even though expected output is wrong

skartikey:
This looks like a flaky test to me. I tried running the test from CLI and observed that the test fails(expected behavior) if I run it 5-6 times in a row. 

![image](https://user-images.githubusercontent.com/1942366/119687525-310bb500-be3f-11eb-84a6-b7a70b05423b.png)

PR #3774, opened 2021-06-01
fix: acceptance tests don't seem to catch different timestamp

skartikey:
fixes: #3631 

Added following test cases for the infinity comparison with epsilon - 
```
math.Abs(math.Inf(-1) - math.Inf(-1)) > 0.000001) // false
math.Abs(math.Inf(1) - math.Inf(1)) > 0.000001) // false
math.Abs(math.Inf(1) - math.Inf(-1)) > 0.000001) // true
```


stdlib/experimental/oee/apq_test.flux
```diff
@@ -21,7 +21,7 @@ outData = "
 #group,false,false,false,false,false,false,false,false
 #datatype,string,long,dateTime:RFC3339,double,double,double,double,long
 #default,_result,,,,,,,
-,result,table,_time,availability,oee,performance,quality,runningTime
```
jsternberg, line 24: Oh, this must be the reason I saw this test was flaky. I tried to look at it because it would randomly fail, but now it makes sense. It was the same diff bug but we never caught it.

stdlib/testing/diff.go
```diff
@@ -504,8 +504,9 @@ func (t *DiffTransformation) rowEqual(want, got *tableBuffer, i int) bool {
 		switch wantCol.Type {
 		case flux.TFloat:
 			want, got := wantCol.Values.(*array.Float64).Value(i), gotCol.Values.(*array.Float64).Value(i)
-			// want == got is for handling +Inf and -Inf.
-			return want == got || math.Abs(want-got) <= t.epsilon
+			if math.Abs(want-got) > t.epsilon {
```
jsternberg, line 507: This looks different. The original says that `want == got` was for handling `+Inf` and `-Inf`. If we remove that section of code, do we still handle that case?

jsternberg (changes requested):
It seems like one portion of the code adds a new bug while removing the other bug. Can you check if this is the case?

stdlib/testing/diff.go
skartikey: We do have a couple of existing tests for +/-Inf and it's all passing with my change. https://github.com/influxdata/flux/blob/master/stdlib/testing/diff_test.go#L233

But I agree we should have some robust handling of Inf case for all the possibilities. 

FYI, go behavior I tried for Inf - 

```
fmt.Printf("epsilon  = %v\n", math.Abs(math.Inf(1) - math.Inf(2))) // NaN
fmt.Printf("epsilon  = %v\n", math.Inf(1) == math.Inf(2)) // true
fmt.Printf("%v", math.IsInf(math.Inf(+1), 0)) // true
```

I will update the diff with my new checks.

jsternberg (approved):

jsternberg (approved):

PR #3775, opened 2021-06-01
chore(fluxdoc): merge fluxdoc into flux crate

rockstar:
This patch makes the `fluxdoc` binary part of the flux crate, rather
than it living in its own crate. This can serve as a pattern for any
other tools that might be created in rust in the future.


scbrickley:
Found the solution for the wasm compatibility issue here: https://docs.rs/getrandom/0.2.3/getrandom/#webassembly-support

libflux/flux/Cargo.toml
```diff
@@ -13,12 +13,17 @@ default = ["strict"]
 strict = []
 
 [dependencies]
+comrak = "0.10.1"
 flux-core = { path = "../flux-core" }
 flatbuffers = "0.6.0"
+getrandom = { version = "0.2", features = ["js"] }
```
nathanielc, line 19: Why is this dependency added? I don't see it used anywhere.

nathanielc: The other deps I can see were part of the fluxdoc crate so it makes sense why they were added.

libflux/flux/src/bin/fluxdoc.rs
```diff
@@ -0,0 +1,94 @@
...
+    if args.html != Path::new("") {
+        write_home(&args.html)?;
+        return write_html(&args.html, &pkg);
+    }
+    Ok(())
+}
+
+lazy_static! {
+    static ref TEMPLATES: Tera = {
+        let mut tera = match Tera::new("fluxdoc/templates/*.html") {
```
nathanielc, line 51: Looks like this path needs to be updated.

nathanielc (commented):
LGTM with some small changes

libflux/flux/Cargo.toml
nathanielc: Ah I see the WASM compatibility note above, can you add a comment about why the `getrandom` crate is needed? That way its clear to future us.

scbrickley: `getrandom` is a dependency of the tera crate, and it was causing the build to fail because it wasn't compatible with WASM. Adding this line allows us to compile to the `wasm32-unknown-unknown` target, as explained here: https://docs.rs/getrandom/0.2.3/getrandom/#webassembly-support

libflux/flux/src/bin/fluxdoc.rs
scbrickley: Ah, right. It would just be `"flux/templates/*.html"`, correct?

libflux/flux/Cargo.toml
scbrickley: I think I forgot to refresh the page before I responded and didn't see your newer comments, haha. Yes, I'll add a comment.

nathanielc (approved):

PR #3786, opened 2021-06-03
chore(fluxdoc): unfinished fluxdoc spec + doc comment examples

scbrickley:


libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+## Documenting Packages
+
+### Import path
+
+Doc comments for packages should start with the full import path for the package, wrapped in backticks. Nothing else should be included on this line.
+
+_Example:_
+
+```
+// `experimental/geo`
```
jsternberg, line 106: Can you explain the reason for this? My thought is that it would be easy to infer the package name without requiring an explicit doc comment.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+### Description
+
+Include a brief description of the package. This should read as one or more full setences.
+
+_Example:_
+
+```
+// `math`
+//
+// The Flux math package provides basic constants and mathematical functions.
```
jsternberg, line 118: I think `Flux` is probably unnecessary here. I'm curious about different documentation styles, but have you considered the Go one? In Go, the standard is to just do `Package math provides`. It avoids using "The" which is mostly an unnecessary article.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+## Documenting functions
+
+### Name and Description
+
+Doc comments for functions should begin with the function name, wrapped in backticks, followed by a brief description of what the function does or how it is used. The name and description should read as one or more full sentences.
+
+_Example:_
+
+```
+// `join` merges two input streams into a single output stream based on columns
```
jsternberg, line 16: Is the reason for the backticks just for markdown? It seems to me like we could auto-backtick the first word if it matches with the function name and this might read better. It's a minor nitpick so I'm fine with the backtick, but I'm not the biggest fan of adding too much formatting to doc comments.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+// with equal values.
+```
+
+### Extra Details
```
jsternberg, line 20: Can you expand this section with an example? I didn't understand what was meant here.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+Parameters should take the form of a markdown unordered list. Each list item should start with the name of the parameter wrapped in backticks, followed by a brief, one-line description.
+
+The parameters in the list should be ordered exactly as they are in the function signature, and no parameter should be omitted.
+
+Each list item should read as a complete sentence.
+
+_Example:_
+
+```
+// - `tables` is a stream of tables
```
jsternberg, line 39: A bit nitpicky, but can we add punctuation to the end of these? Normally, I don't like punctuation with bullet points. But my thought is I'd like to be able to do this:

```
// - `method` is the method to use when joining.
//
//    The supported methods of joining are `inner` and `outer`.
//     The `inner` method will require both input tables to match their columns based on the `on` parameter. The `outer` method will include all rows and fill in null values if there isn't an accompanying row.
//
// - `on` is a list of column names on which to join.
```

This gets formatted in markdown as:

- `method` is the method to use when joining.

   The supported methods of joining are `inner` and `outer`.
    The `inner` method will require both input tables to match their columns based on the `on` parameter. The `outer` method will include all rows and fill in null values if there isn't an accompanying row.

- `on` is a list of column names on which to join.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+// - `method` is the method to use when joining
+// - `on` is a list of column names on which to join
+```
+
+### Code Examples
+
+Code examples consist of a heading, an optional preamble, and a markdown-formatted code block.
```
jsternberg, line 46: Can you clarify where in the flux file the code examples go?

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+Flux packages often come with one or more predefined constants that users can import and use in their code. Some of these constants are defined with `builtin` statements, which means their actual value is defined in Go code rather than Flux code. To document a such constant, include a line comment directly above its declaration with a string representation of the constant's value.
+
+Because this value will only ever be treated as a string in the docs, writing a description of the constant's value, rather than the literal value, is acceptable if it makes things easier to read or understand.
+
+If a constant is not defined with a builtin statement, the doc comment parser should be able to get its value from the flux code itself, and no documentation is required.
+
+_Example:_
+
+```
+// 0.693147180559945309417232121458176568075500134360255254120680009
```
jsternberg, line 153: I think something like:

```
// ln2 = 0.693147180559945309417232121458176568075500134360255254120680009
builtin ln2 : float
```

I think this might be better. It matches better with the existing doc comment structure and immediately tells you what it's referencing. We can do:

```
// `ln2` = 0.693147180559945309417232121458176568075500134360255254120680009
```

If we decide to stick with backticks for the symbol name.

jsternberg (commented):
This looks like a good first draft. I had a few recommendations for changes to the documentation format.

It might also be good to reformat this file to put each line on a newline. Something we sometimes do with markdown documents is like this:

```
I wrote a line.
I wrote another line.
Now I can select a specific line more easily to comment on.
And diffs won't show that I changed four lines, but just one.
And this will still display the same as if I wrote it on a single line.
```

libflux/fluxdoc/SPEC.md
scbrickley: I just stole that line from the current docs, and didn't really consider setting a standard for the wording. I like this suggestion though.

libflux/fluxdoc/SPEC.md
scbrickley: Hmmm... not sure why I included that, in retrospect. I'll remove it.

libflux/fluxdoc/SPEC.md
scbrickley: I suppose we don't really need the backticks, but it's something that came out of the working group. I think we liked it because it makes it clear that it needs to match the function name exactly, i.e. `aggregateWindow` instead of `AggregateWindow`, which we saw a lot of when looking through comments in stdlib. I suppose we could just spell that out in the spec, though.

nathanielc (changes requested):
I really like this and the precision with which things are specified is excellent. I have some initial feedback. These are all strong opinions loosely held so please push back on any of them. 


On types should the doc structs contain a string representation of the type? Or should they be the actual Rust structures representing the type? 

If we use the real Rust structure it would allow us to interact with them etc, but we do want to convert them to strings eventually for display. I think a first pass of strings is good but wanted to open up the discussion.

libflux/fluxdoc/SPEC.md
scbrickley: Yes, I will do that.

libflux/fluxdoc/SPEC.md
scbrickley: 1. Yes, I can add punctuation as a requirement.
2. Should I add support for the extra notes you've included into the parameter list? I was thinking each list item would be one line maximum, but I like your example.
3. I thought we didn't support outer joins?

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+## Docs as Rust Data Structures
+
+The overarching goal for this project is to be able to parse the new docs into rust structures, which we will then serialize into JSON for broader consumption.
+
+Here's a proprosal for what those rust data structures could look like:
+
+```rust
+enum Doc {
```
nathanielc, line 231: I just had another thought: what if we add source location information to all of these structs? Then we could link back to the github source code from the docs. 

This doesn't need to be in the first implementation but something we could add later.

libflux/fluxdoc/SPEC.md
scbrickley: I say at the top of the file that each section should go in the order presented in the document, but I can make that more explicit. Something like "Code examples should come immediately after the parameter list." Would that work?

libflux/fluxdoc/SPEC.md
scbrickley: I originally included something similar to this. Sorta went back and forth on it. I'll change it back.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+```
+// `join` merges two input streams into a single output stream based on columns
+// with equal values.
+```
+
+### Extra Details
+
+Some function documentation may require explanation beyond what is provided in the `Name and Description` section described above. Any markdown-formatted text between the end of the initial function description and the beginning of the parameter list will be treated as additional explanatory notes, and will be read and stored separately from the initial description in the final data structure.
+
+Separating the main function description from this longer, more detailed explanation allows consumers who only need a condensed version of the docs (e.g. the InfluxDB UI) to do so, while other consumers are free to use the full docs.
```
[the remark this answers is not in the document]
scbrickley: Yes, just an empty line comment, but we could make it more explicit. Something like this maybe?

```
// `join` merges two input streams into a single output stream based on columns
// with equal values.
// 
// ---
// 
// Null values are not considered equal when comparing column
// values. The resulting schema is the union of the input schemas. The resulting
// group key is the union of the input group keys.
```

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+
+This section may be omitted.
+
+### Parameters
+
+Parameters should take the form of a markdown unordered list. Each list item should start with the name of the parameter wrapped in backticks, followed by a brief, one-line description.
```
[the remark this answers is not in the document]
scbrickley: Sounds useful. I'll take look!

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+
+Code examples consist of a heading, an optional preamble, and a markdown-formatted code block.
+
+The heading should be a markdown H1 heading, and should be immediately followed by a blank line comment.
+
+Any text between the heading and the code block will be treated as part of the preamble. The preamble should be separated from the code block by a blank line comment. The preamble may be omitted.
+
+Code blocks should contain valid flux code that readers can copy and paste into the InfluxDB data explorer and run with no modifications. Any sample data used in a code example should come from an `array.from()` call in the example code itself, rather than being represented as a markdown- or html-formatted table.
+
+This section may include any number of code examples.
```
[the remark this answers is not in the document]
scbrickley: Correct.

libflux/fluxdoc/SPEC.md
jsternberg: We don't so don't put that in the actual docs but I couldn't think of anything else that I could fake document.

I'd say we should call out that the extra notes are supported and we should support them. It's a normal thing supported by markdown so we wouldn't need to do anything special.

libflux/fluxdoc/SPEC.md
jsternberg: Yes the extra clarification is good.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Full Example
+
+For some examples of what a complete doc comment looks like for a flux function, see the doc comments for `join()` and `aggregateWindow()` in the `universe` package.
+
+## Documenting Packages
+
+### Import path
+
+Doc comments for packages should start with the full import path for the package, wrapped in backticks. Nothing else should be included on this line.
```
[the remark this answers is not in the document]
scbrickley: @jsternberg asked the same thing, and I'm not sure why I included. I must have added it early on while I was writing this and just overlooked it. I'll remove it.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+}
+
+struct ConstDoc {
+	name: String,
+	value: String,
+}
+
+struct ConstGroupDoc {
+	head: String,
+	desc: String,
```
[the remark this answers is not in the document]
scbrickley: Not every constant will be part of a group, I think. And I don't know if we can guarantee that all constants in a group will have the same type. I think it would be better to just include the type with each individual constant.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Documenting Flux Options
+
+### Name and Description
+
+The first line of a doc comment for a flux option should start with the name of the option wrapped in backticks, followed by a brief explanation of what the option is, or how it is used.
+
+### Extra Details
+
+Additional notes explaining how users can use the option in their code. This section should follow the same formatting rules as the `Extra Details` section for Flux functions described earlier in this document.
```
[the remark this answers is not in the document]
scbrickley: That's something I thought about, but I couldn't think of a good answer. I think if the mark down includes an unordered list, it may be easy for the parser to confuse that for the parameter list, and that could cause problems.

Maybe this can be solved by using a special character or set of characters to delineate the different sections. I think if we do that, we can include any markdown in this section.

That said, for the sake of keeping things organized, I think we should strongly encourage doc writers to keep code blocks in the `Code Examples` section.

libflux/fluxdoc/SPEC.md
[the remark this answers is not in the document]
nathanielc: I think an empty comment line is best, i.e. the first paragraph is the short version, everything else is the longer version.

libflux/fluxdoc/SPEC.md
```diff
@@ -0,0 +1,284 @@
...
+
+For an example of a complete doc comment for a Flux option, see the `enabledProfilers` option in the `profiler` package.
+
+## Documenting Builtin Constants
+
+Flux packages often come with one or more predefined constants that users can import and use in their code. Some of these constants are defined with `builtin` statements, which means their actual value is defined in Go code rather than Flux code. To document a such constant, include a line comment directly above its declaration with a string representation of the constant's value.
+
+Because this value will only ever be treated as a string in the docs, writing a description of the constant's value, rather than the literal value, is acceptable if it makes things easier to read or understand.
+
+If a constant is not defined with a builtin statement, the doc comment parser should be able to get its value from the flux code itself, and no documentation is required.
```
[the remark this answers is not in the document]
scbrickley: We could check the expression type of the value being assigned and differentiate them that way. Would that work?

libflux/fluxdoc/SPEC.md
scbrickley: That would be great. It might be the small push someone needs to start contributing code!

libflux/fluxdoc/SPEC.md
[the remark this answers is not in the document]
scbrickley: Reading this again, I realize that I just suggested exactly what you warned against :facepalm: 

To make sure I understand your suggestion clearly, you're saying that the docs for a package would just contain a list of the constants it provides, but only the names, not the values. Is that correct?

If that's the case, I think I'm fine with that.

libflux/fluxdoc/SPEC.md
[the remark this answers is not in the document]
pierwill: pulldown-cmark is what rustdoc uses: https://github.com/rust-lang/rust/tree/master/src/librustdoc

PR #3787, opened 2021-06-04
fix(formatter): change FunctionLiteral precedence to preserve parens

### LINKED ISSUE #3758: Formatter does not handle inline function calls with if statements correctly
I had the following change in the standard library (the `-` sign):

```diff
 aggregateWindow = (every, fn, offset=0s, column="_value", timeSrc="_stop", timeDst="_time", createEmpty=true, tables=<-) => tables
     |> window(every: every, offset: offset, createEmpty: createEmpty)
     |> fn(column: column)
-    |> ((tables=<-) => if createEmpty then tables |> table.fill() else tables)()
+    |> (tables=<-) => if createEmpty then tables |> table.fill() else tables()
     |> duplicate(column: timeSrc, as: timeDst)
     |> window(every: inf, timeColumn: timeDst)
```

When I ran the formatter on the file, it removed the parenthesis around the if statement and appended the parenthesis to the end. I think this changes the meaning so it is invoking `tables` as a function instead of invoking the inline function.

nathanielc:
Prior to this change parens around a function literal that was
immediately called would be lost during formatting.
This change makes FunctionLiteral literal have an explicit precedence
such that parens are preserved.

Fixes #3758

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


BarWeiner1 (approved):
Now that you updated the buildinfo, I think the changes look logical and fine.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -72,6 +72,21 @@ fn funcs() {
...
+    assert_unchanged("a()");
+    assert_format("(a)()", "a()");
+    // call anonymous function
+    assert_unchanged("((a) => a)()");
+    // pipe anonymous function
+    assert_unchanged("(() => 1) |> f()");
+}
+#[test]
+fn function_expr() {
+    assert_unchanged("() => 1 and () => 2");
```
scbrickley, line 87: Can you clarify what this line is testing? What would the failure state here look like?

scbrickley: Would it just become `=> 1 and => 2` without the fix?

scbrickley (approved):

libflux/flux-core/src/formatter/tests.rs
nathanielc: A failure state would be to add parens around the righthand function literal. This shows that while precedence is lower than call expressions its higher than other expressions like `and` logical expressions.

nathanielc: Technically this is not semantically valid syntax as you can't `and` two function literals, but we can still parse and format the syntax.

PR #3789, opened 2021-06-04
fix(formatter): remove tabstop processing from formatter

### LINKED ISSUE #3646: Formatting: comments shouldn't get indented, tab inside comments should be preserved
The formatter is indenting the whitespace inside the comment. Ideally, it should be preserved. Also, the extra space before the last comment line shouldn't be there. 

**Example -** 

**Before -** 
```
// Equivalent TICKscript query:
// stream 
// 	 |eval(lambda: 2 * _value)
//		.as('_newValue')
```

**After format-** 
```
// Equivalent TICKscript query:
// stream 
//  |eval(lambda: 2 * _value)
 //.as('_newValue')
```

**Findings -** Looks like the tabs in the comments are getting removed by the tabwriter #3315 , spaces are getting preserved though. 

**Sample unit test output -** 

```
Diff < left / right > :
<"\n// \t |eval(lambda: 2 * _value)\n//\t\t.as(\'_newValue\')"
>"\n//  |eval(lambda: 2 * _value)\n //.as(\'_newValue\')"
```

DOD:

 - Add tests that show tabs are preserved in comments and string literals
 - Remove the TabWriter

nathanielc:
Prior to this change literal tabstop would be removed from comments and
string literals.

There wasn't a clear way forward with how to use tabstops without
affecting string literals and comments so we have removed the feature.

Fixes #3646 
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


skartikey (approved):
Looks good!
I like the tab_literals test case confirming that the tab persists.

PR #3793, opened 2021-06-08
feat(stdlib/pagerduty): add exclude parameter to pagerduty.dedupKey()

jsternberg:
Add an `exclude` parameter to the `pagerduty.dedupKey()` call. If
specified, this will ignore the given keys from the group key when
calculating the dedup key. If the columns don't exist, no error is
triggered.

The default for this value is `["_start", "_stop", "_level"]` which are
the typical group keys in from the monitor package that cause problems
for the dedup key.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


adrian-thurston (approved):

jsternberg:
Wow. I apparently forgot to finish writing the short part of my commit message and then I wrote the detail. Going to fix that before merging. Thanks for the approval. Test failures seem to be because I also forgot to call `go generate`.

PR #3797, opened 2021-06-10
fix(csv): adds length check to csv annotation parsing

nathanielc:
Prior to this change csv that had annotations with less than three
fields would cause the csv decoder to panic. Now an appropriate error
message is returned.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/csv/result.go
+++ b/csv/result.go
@@ -384,10 +384,19 @@ func readMetadata(r *bufferedCSVReader, c ResultDecoderConfig) (tableMetadata, e
 			}
 			switch annotation := strings.TrimPrefix(line[annotationIdx], commentPrefix); annotation {
 			case datatypeAnnotation:
+				if defaultRecordStartIdx > len(line) {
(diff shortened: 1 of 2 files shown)


jsternberg (approved):

csv/result.go
```diff
@@ -384,10 +384,19 @@ func readMetadata(r *bufferedCSVReader, c ResultDecoderConfig) (tableMetadata, e
 			}
 			switch annotation := strings.TrimPrefix(line[annotationIdx], commentPrefix); annotation {
 			case datatypeAnnotation:
+				if defaultRecordStartIdx > len(line) {
+					return tableMetadata{}, errors.Wrap(csv.ErrFieldCount, codes.Invalid, "failed to read \"datatype\" annotation")
+				}
 				datatypes = copyLine(line[defaultRecordStartIdx:])
```
nathanielc, line 390: The panic was on this line and the first line of the csv file being read was:

```
#datatype,long
```

PR #3800, opened 2021-06-10
fix(stdlib): pagerduty tests did not include _value column

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/pagerduty/pagerduty_test.flux
+++ b/stdlib/pagerduty/pagerduty_test.flux
@@ -4,24 +4,24 @@ import "array"
 import "pagerduty"
 import "testing"
 
-input = array.from(
+input = () => array.from(
     rows: [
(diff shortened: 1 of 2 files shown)


stdlib/pagerduty/pagerduty_test.flux
```diff
@@ -4,24 +4,24 @@ import "array"
 import "pagerduty"
 import "testing"
 
-input = array.from(
+input = () => array.from(
```
nathanielc, line 7: why the change to making it a function? Does it have any effect on the result?

nathanielc (approved):
LGTM

stdlib/pagerduty/pagerduty_test.flux
jsternberg: Forgot to respond to this. No, I thought it did so I changed it and it didn't have an effect. Then I forgot to put it back. This works nicely anyway though.

Issue #3780, opened 2021-06-03
Implement bootstrap::stdlib_docs method in the semantic::bootstrap module


scbrickley:
In order to build the flux crate with the generated stdlib documentation baked in, we need a function in the `fluxcore::semantic::bootstrap` module, called `stdlib_docs`. The purpose of the function is to walk stdlib and produce doc structures for every identifier in every package.

The function signature for `stdlib_docs` should read...

`pub fn stdlib_docs(lib: PolyTypeMap, files: Vec<String>) -> Result<Vec<DocPackage>, Error>`

...where `lib` and `files` are the values of the same name returned by `infer_stdlib` in [build.rs](https://github.com/influxdata/flux/blob/8018bdce94cb59851550f130e643aa9f1bafd107/libflux/flux/build.rs#L51-L53)

The logic of the function should be as follows:
1. Use the pre-parsed ASTFileMap to extract doc comments from stdlib
2. Parse those doc comments into `DocPackage` and `DocValue` from the [`fluxdoc` crate](https://github.com/influxdata/flux/blob/8018bdce94cb59851550f130e643aa9f1bafd107/libflux/fluxdoc/src/doc.rs) using the data provided in `lib` and `files` which should get passed down as a reference to helper functions.
4. Return a vector of all the docs.

Once this function is available we can use it to finish #3781

In order to do this we will have to migrate the `DocPackage` and `DocValue` structs into `flux-core`

Once the spec for flux doc comments is finalized (see #3770), we can update this function to use the new fluxdoc data structures, as detailed in #3783

PR #3808, opened 2021-06-16
feat(libflux): bootstrap docs methods

BarWeiner1:
Adds the `stdlib_docs` function to the `fluxcore::semantic::bootstrap` module as described in #3780. The function walks through stdlib and produces doc structures for every identifier in the package.
Also adds a docs function in `build.rs` that calls the `stdlib_docs` function to serialize the doc structure into JSON and writes it to a file `docs.json`.


libflux/flux-core/Cargo.toml
```diff
@@ -13,6 +13,7 @@ default = ["strict"]
 strict = []
 
 [dependencies]
+comrak = "0.10.1"
```
rockstar, line 16: I'm not sure that we should be putting this work in `flux-core`. I'd much rather see it in the `flux` crate

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -1,12 +1,14 @@
 //! Flux start-up.
 
+//use std::collections::HashMap;
```
rockstar, line 3: If we don't need this, let's remove it.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -1,12 +1,14 @@
...
 use std::collections::HashSet;
 use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 
 use crate::ast;
 use crate::parser;
+//use crate::parser::Parser;
```
rockstar, line 11: If we aren't using this, let's remove it.

libflux/flux/build.rs
```diff
@@ -3,6 +3,8 @@ extern crate fluxcore;
 use std::{env, fs, io, io::Write, path};
 
 use fluxcore::semantic::bootstrap;
+//use fluxcore::semantic::bootstrap::DocPackage;
```
rockstar, line 6: Same here. If the commented out code is unneeded, let's remove it.

libflux/flux/docs.json
```diff
@@ -0,0 +1 @@
...}], unit:duration, ?timeColumn:string) => [{A with _time:B, _time:time}] where B: Timeable"},{"pkgpath":"universe","name":"kaufmansER","doc":"<p>kaufmansER computes Kaufman's Efficiency Ratios of the <code>_value</code> column</p>\n","typ":"(<-tables:[A], n:int) => [{B with _value:float, _value:float}] where A: Record"},{"pkgpath":"universe","name":"toString","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:string}]"},{"pkgpath":"universe","name":"toInt","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:int}]"},{"pkgpath":"universe","name":"toUInt","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:uint}]"},{"pkgpath":"universe","name":"toFloat","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:float}]"},{"pkgpath":"universe","name":"toBool","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:bool}]"},{"pkgpath":"universe","name":"toTime","doc":"","typ":"(<-tables:[{A with _value:B}]) => [{A with _value:B, _value:time}]"},{"pkgpath":"universe","name":"today","doc":"<p>today returns the now() timestamp truncated to the day unit</p>\n","typ":"() => time"}]}]
```
rockstar, line 1: I don't think we should include any more generated content than is absolutely necessary. Can we just generate this on the fly and add this file to `.gitignore` so we don't commit it? Is there any value in it being in the tree?

libflux/flux/src/lib.rs
```diff
@@ -5,9 +5,8 @@ pub mod wasm;
 extern crate fluxcore;
 extern crate serde_aux;
 
-#[macro_use]
+//#[macro_use]
```
rockstar, line 8: This is now commented out, so I wonder if we ever needed it. I _suspect_ this was a holdout from rust < 2018 edition and can be removed.

libflux/flux/src/lib.rs
```diff
@@ -43,6 +43,11 @@ pub fn imports() -> Option<Environment> {
         .unwrap()
         .into()
 }
+///Todo: make a docs.json and then execute the code below
```
rockstar, line 46: Is this TODO still relevant?

libflux/flux/src/lib.rs
```diff
@@ -1110,4 +1116,19 @@ from(bucket: v.bucket)
...
     }
+
+    #[test]
+    //Testing the CSV package's docs
+    fn ensure_docs() {
+        let d = docs();
+        let want = r#"DocPackage { path: "csv", name: "csv", doc: "<p>CSV provides an API for working with <a href=\"https://github.com/influxdata/flux/blob/master/docs/SPEC.md#csv\">annotated CSV</a> files.</p>\n", values: [DocValue { pkgpath: "csv", name: "from", doc: "<p>From parses an annotated CSV and produces a stream of tables.</p>\n", typ: "(?bucket:string, ?bucketID:string, ?host:string, ?org:string, ?orgID:string, ?token:string) => [{A with _value:B, _time:time, _measurement:string, _field:string}]" }] }"#;
+        let got = format!("{:?}", d[19]); // the csv DocPackage location
+        assert_eq!(want, got);
+        //println!("{:?}", d[20].values);
```
rockstar, line 1127: I think these lines can be removed.

stdlib/pagerduty/pagerduty_test.flux
```diff
@@ -1,5 +1,6 @@
 package pagerduty_test
 
+
```
rockstar, line 3: This looks like we ran the formatter on this file and it made some changes. Can you please revert the changes in this file, as they are unrelated to the patch at hand?

rockstar (changes requested):
This is a good start. Thanks for working on it.

I do have one relatively big concern about the home for this work. I'm not sure I agree that the docs work belongs in the `flux-core` crate, but I'm open to discussions about it.

stdlib/pagerduty/pagerduty_test.flux
BarWeiner1: If it is not formatted according to the current state of the formatter we can't pass CI. It was just changed when we ran make generate and make fmt

rockstar: Can you try merging master? If your existing, unrelated work isn't passing CI because it's missing this change, that means no one is passing CI with their branches without this change, and I just put a patch up yesterday without this change, so it might just need updating.

nathanielc:
@rockstar I know I mentioned briefly that the plan was to include this into the tree. Let me explain why and then we can go from there.

We have a need to expose the docs to the UI in a way that it can access the docs without needing to make extra network calls. So the design we settled on is to bake the docs into the crate and therefore ship the docs as part of the WASM nodejs module that the frontend includes. This way the UI can make a javascript call and get access to the docs which will be up to date and relevant to the current version. 

Adding the json docs might really expand the size of the built crate. For that I have a few ideas. 

- We could gzip the docs JSON bytes. JSON compresses well and this will give us large savings. We can test to see how large things are before and after
- We could move the logic to include the docs into the new WASM crate so they are only included in the WASM nodejs module and not part of the normal binary
- Longer term but we could remove both the docs and the stdlib types from being baked in and provide a way to load those bits at runtime instead of a build time. (We would still want to pre build them but perhaps ship them as a separate binary file instead of baked into the crate build). This requires some design that we haven't started yet and so I don't want it to block the current work.

@BarWeiner1 @rockstar @JustSomeCarbon  Thoughts? Should we implement the first two options?

rockstar:
> So the design we settled on is to bake the docs into the crate and therefore ship the docs as part of the WASM nodejs module that the frontend includes.

To me, this sounds like we should be having the lsp or another downstream consumer generating these docs in whatever format fits them best. Bundling them in the crate doesn't really serve the crate at all, right? I'm less concerned about the size of the crate, and more concerned about trying to create a one-size-fits-most(maybe) in our top-level core crate.

rockstar (approved):

PR #3803, opened 2021-06-11
chore: Add spec for flux doc comments

scbrickley:
Adds a document containing the specification for the new Flux doc comments, as per #3770

There were a lot of comments on the PR with the rough draft of this document. I opened this new PR with the updated version to make any future conversations in the comments easier to follow.

[Original PR here.](https://github.com/influxdata/flux/pull/3786)

PR #3816, opened 2021-06-22
docs: Edit fluxdoc spec

pierwill:
Some suggestions for the fluxdoc spec in https://github.com/influxdata/flux/pull/3803/.


scbrickley (approved):
This all looks good to me. Thanks for taking the time to look this over!

PR #3823, opened 2021-06-24
docs: document date pkg

JustSomeCarbon:
added documentation for `stdlib/date`


stdlib/date/date.flux
```diff
@@ -1,46 +1,395 @@
 package date
```
nathanielc, line 1: Can you add a package doc comment?

nathanielc (approved):

PR #3828, opened 2021-06-24
docs: document system pkg

JustSomeCarbon:
added documentation to `stdlib/system`


stdlib/system/system.flux
```diff
@@ -1,4 +1,13 @@
 package system
 
 
+// time is a function that returns the current system time
```
nathanielc, line 4: nit, missing period.

stdlib/system/system.flux
```diff
@@ -1,4 +1,13 @@
...
+//
+// ## Example
+// ```
+// import "system"
+//
+// data
+//   |> set(key: "processed_at", value: string(v: system.time() ))
```
nathanielc, line 11: this example is not valid as the `data` var is undefined, perhaps a better example would be to use `array.from`

```
import "system"
import "array"

array.from(rows:[{time: system.time()}])
```

nathanielc (approved):
LGTM a few small changes, merge when updated.

BarWeiner1 (approved):

PR #3834, opened 2021-06-28
fix(execute): allow work queue to be resized when work exceeds queue length

jsternberg:
This modifies the dispatcher to make the work queue dynamically increase
when the amount of work exceeds the default size.

Previously, the work queue was implemented with a channel. Channels have
a static size and it was possible for a worker, which reads from the
work queue, to also attempt to add to the work queue and this resulted
in a deadlock.

This updates the work queue to be a dynamically sized ring to avoid a
deadlock when the amount of work exceeds the size of the work queue.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/dispatcher.go
+++ b/execute/dispatcher.go
@@ -26,7 +26,9 @@ type ScheduleFunc func(ctx context.Context, throughput int)
 
 // poolDispatcher implements Dispatcher using a pool of goroutines.
 type poolDispatcher struct {
-	work chan ScheduleFunc
+	work   *ring
+	ready  chan struct{}
+	workMu sync.Mutex
 
 	throughput int
 
@@ -43,7 +45,8 @@ type poolDispatcher struct {
 func newPoolDispatcher(throughput int, logger *zap.Logger) *poolDispatcher {
 	return &poolDispatcher{
 		throughput: throughput,
-		work:       make(chan ScheduleFunc, 100),
+		work:       newRing(100),
+		ready:      make(chan struct{}, 1),
 		closing:    make(chan struct{}),
 		errC:       make(chan error, 1),
 		logger:     logger.With(zap.String("component", "dispatcher")),
@@ -51,9 +54,19 @@ func newPoolDispatcher(throughput int, logger *zap.Logger) *poolDispatcher {
 }
 
 func (d *poolDispatcher) Schedule(fn ScheduleFunc) {
+	d.workMu.Lock()
+	defer d.workMu.Unlock()
+
+	// Schedule the work and then report to the channel that there
+	// is available work to unblock the worker scheduler thread.
+	d.work.Append(fn)
 	select {
-	case d.work <- fn:
-	case <-d.closing:
+	case d.ready <- struct{}{}:
+		// The ready channel should have a buffer of 1.
+		// Work being present is a binary yes or no.
+		// If we say yes multiple times, we only need to read it once
+		// in the outermost run loop.
+	default:
 	}
 }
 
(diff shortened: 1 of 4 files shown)


execute/dispatcher.go
```diff
@@ -136,8 +145,38 @@ func (d *poolDispatcher) run(ctx context.Context) {
...
+		fn(ctx, d.throughput)
+
+		// Check to see if the context was canceled or
+		// the dispatcher was closed.
+		select {
+		case <-ctx.Done():
+			return
+		case <-d.closing:
+			return
+		default:
```
skartikey, line 179: is this to prevents the select statement from blocking? So when no communication can proceed it executes the default case and finishes immediately. But then the outer for loop will keep on running unless one of the select case is satisfied. If this is intended could you please put a log message/ comment in the default case?

execute/ring.go
```diff
@@ -0,0 +1,47 @@
...
+func (r *ring) Len() int {
+	return r.sz
+}
+
+func (r *ring) Append(v interface{}) {
+	if r.sz == len(r.buf) {
+		// Copy the existing buffer.
+		// We copy this in two parts.
+		buf := make([]interface{}, 0, r.sz*2)
+		buf = append(buf, r.buf[r.i:]...)
```
skartikey, line 39: not sure how go behaves here, but manually copying a buffer could slow down the performance? 
Kindly let me know if go behaves otherwise.

execute/dispatcher.go
```diff
@@ -43,17 +45,24 @@ type poolDispatcher struct {
 func newPoolDispatcher(throughput int, logger *zap.Logger) *poolDispatcher {
 	return &poolDispatcher{
 		throughput: throughput,
-		work:       make(chan ScheduleFunc, 100),
+		work:       newRing(100),
```
skartikey, line 48: looks like we are still defaulting the buffer to 100 like it was in the existing code. But when it goes above 100 then we manually increase the buffer size and manually copy the buffer.

skartikey (commented):
My comments are more towards understanding the code :)
Kindly have a look and reply.

execute/dispatcher.go
jsternberg: Yes, that is correct. Sure.

execute/ring.go
jsternberg: It looks like I forgot to finish writing my comment. I don't believe manually copying a buffer slows anything down. These two statements are the same in Go.

```
newBuf := make([]int, len(buf))
copy(newBuf, buf)
```

```
newBuf := make([]int, 0, len(buf))
newBuf = append(newBuf, buf...)
```

Append's implementation is basically this:

```
func append(dst []int, item int) []int {
    if cap(dst) <= len(dst)+1 {
        dst = dst[:len(dst)+1]
    } else {
        cpy := make([]int, len(dst)+1, len(dst)*2)
        copy(cpy, dst)
        dst = cpy
    }
    dst[len(dst)-1] = item
}
```

It's a little more complex because it accepts variable arguments but it expands the capacity of a slice and then sets the value when there is capacity left and it copies the array into a new buffer when the append would exceed the capacity.

```
buf = append(buf, 1)

newBuf = make([]int, len(buf), len(buf)*2)
copy(,

execute/dispatcher.go
jsternberg: Yes. It's a resizable ring buffer.

execute/ring.go
skartikey: that's insightful. Thanks

skartikey (approved):

PR #3838, opened 2021-06-29
fix(stdlib): moving average creates columns with the same length when n is the size of the input

### LINKED ISSUE #3795: Moving average produces tables with the wrong number of rows in some columns for certain incoming tables
For certain incoming tables, moving average produces the wrong number of rows in the `_value` column.

This query works **correctly**.

```
import "array"

array.from(
  rows: [
    {_time: 2021-06-08T09:00:00Z, _value: 2.0},
    {_time: 2021-06-08T10:00:00Z, _value: 3.0},
    {_time: 2021-06-08T11:00:00Z, _value: 4.0},
  ]
)
  |> movingAverage(n: 2)

// Result: _result
// Table: keys: []
//                     _time:time                  _value:float
// ------------------------------  ----------------------------
// 2021-06-08T10:00:00.000000000Z                           2.5
// 2021-06-08T11:00:00.000000000Z                           3.5
```

This query work **incorrectly**.

```
import "array"

array.from(
  rows: [
    {_time: 2021-06-08T09:00:00Z, _value: 2.0},
    {_time: 2021-06-08T10:00:00Z, _value: 3.0},
    {_time: 2021-06-08T11:00:00Z, _value: 4.0},
  ]
)
  |> movingAverage(n: 3)

// Result: _result
// Table: keys: []
//                     _time:time                  _value:float
// ------------------------------  ----------------------------
// 2021-06-08T10:00:00.000000000Z                             3
//                                                          3.5
```

This bug seems to be triggered whenever `n = <number of rows>`.

jsternberg:
The moving average transformation would append an extra value to the
value column if the `n` value was the exact same as the input column
length.

This fixes the transformation so it correctly does not produce that
extra row in only the value column.

Fixes #3795.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/distinct_test.flux
```diff
@@ -1,5 +1,6 @@
 package universe_test
```
jsternberg, line 1: I noticed that this file was not correctly formatted when I merged that PR. For whatever reason, this passed CI. I'm just including this fix in this PR since it isn't a major change. This is just a formatting change.

scbrickley (approved):
LGTM!

PR #3839, opened 2021-06-29
docs: indluxdb/monitor docs

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/monitor/monitor.flux
+++ b/stdlib/influxdata/influxdb/monitor/monitor.flux
@@ -116,6 +116,21 @@ _stateChanges = (fromLevel="any", toLevel="any", tables=<-) => {
         |> experimental.group(mode: "extend", columns: ["_level"])
 }
 
+// Notify will call the endpoint and log the results.
+notify = (tables=<-, endpoint, data) => tables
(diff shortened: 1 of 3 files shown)


stdlib/influxdata/influxdb/monitor/monitor.flux
```diff
@@ -89,48 +154,91 @@ stateChangesOnly = (tables=<-) => {
...
-// StateChanges only operates on data with data where r._level exists.
 stateChanges = (fromLevel="any", toLevel="any", tables=<-) => {
     return if fromLevel == "any" and toLevel == "any" then
         tables |> stateChangesOnly()
     else
         tables |> _stateChanges(fromLevel: fromLevel, toLevel: toLevel)
 }
 
-// Notify will call the endpoint and log the results.
-notify = (tables=<-, endpoint, data) => tables
```
scbrickley, line 104: This is probably what it's complaining about. Looks like we deleted the `notify` function and the `logs` function below. If we add them back in, we should see that test pass again.

JustSomeCarbon (approved):

PR #3843, opened 2021-06-30
docs: InfluxDB docs

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/influxdata/influxdb/influxdb.flux
```diff
@@ -1,6 +1,82 @@
+// The influxdb package provides functions for analyzing InfluxDB metadata.
```
scbrickley, line 1: This is not a big enough deal to hold up this PR, but according to the [spec](https://github.com/influxdata/flux/blob/master/libflux/flux/FLUXDOC.md#headline), the package headline here should read

```
Package influxdb provides functions for analyzing InfluxDB metadata
```

Like, I said, not a huge deal, but this is a convention we should follow going forward.

scbrickley (approved):
Looks good. I had one minor note, but nothing that we need to block on.

PR #3855, opened 2021-07-02
feat(libflux/flux): exposed function to get docs in WASM

BarWeiner1:
### Done checklist
- [ ] exposed function to get JS values for flux documentation given a flux identifier
- [ ] Test cases written

PR #3857, opened 2021-07-06
chore: Run wasm tests as part of CI

scbrickley:
This PR does a few things:

- Change `get_json_documentation` to take a `&str` and return a `JsValue`
- Move `json_docs_test` to a `wasm::tests` module
- Add a `test-wasm` Make target

Note that `json_docs_test` currently fails because `left != right` in the `assert_eq!` call. We will have to fix that before https://github.com/influxdata/flux/pull/3855 can merge


fchikwekwe (approved):
Yeah this looks fine to me

scbrickley:
Hmmm... looks like I've got a bad `wasm-pack` command. Doesn't look like it was actually able to run the tests. Looking into it now.

PR #3860, opened 2021-07-07
test: add group tests for count and sum

wolffcm:
This just adds a few tests for `group` and `count` or `sum` used together, in preparation for adding these pushdowns to OSS.

In TSM1 the `_field` attribute of a series is treated differently than other tag keys (unlike in TSM2) so I wanted to be sure to hit all the different combinations of grouping and filtering on `_field`.


scbrickley (approved):
I don't have much context on the change these are meant to test, but these testcases all seem straightforward enough, and having more coverage can't hurt.

PR #3861, opened 2021-07-08
chore: make checkfmt always runs current formatter

skartikey:
This PR is to fix the error in the CI while running checkfmt at the integration phase.

Error - 
```
./etc/checkfmt.sh: line 41: flux: command not found
CircleCI received exit code 0
```

### DIFF
--- a/etc/checkfmt.sh
+++ b/etc/checkfmt.sh
@@ -37,8 +37,7 @@ if [[ -n "$FMT_OUT" ]]; then
 fi
 
 cd ..
-export PATH=$PATH:.
-FMT_OUT="$(flux fmt -c stdlib)"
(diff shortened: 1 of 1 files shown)


wolffcm (approved):
Looks good, glad you caught this!

etc/checkfmt.sh
```diff
@@ -38,7 +38,7 @@ fi
 
 cd ..
 export PATH=$PATH:.
-FMT_OUT="$(flux fmt -c stdlib)"
+FMT_OUT="$(env GO111MODULE=on go run -tags '' ./cmd/flux/main.go fmt -c stdlib)"
 if [[ -n "$FMT_OUT" ]]; then
```
wolffcm, line 42: You could also add `set -e` to the top of this script to ensure that it returns non-zero if any command returns an error.

adrian-thurston (approved):

PR #3863, opened 2021-07-08
test: fix group_agg_test to yield data and use array.from for input

wolffcm:
This PR removes some of the tests that I added yesterday in #3860 since they took too long to run in `influxdb`.

I also added `|> yield()` to each test query, since extending the test in `influxdb` seems to require it.

Finally I used `array.from()` to create the input data since that is possible to do now, and it's more readable.

### DIFF
--- a/stdlib/planner/group_agg_test.flux
+++ b/stdlib/planner/group_agg_test.flux
@@ -85,6 +85,26 @@ testcase group_one_tag_filter_field_count {
     testing.diff(got, want) |> yield()
 }
 
+testcase group_two_tag_filter_field_count {
+    want = array.from(
+        rows: [
+            {"t0": "t0v0", "t1": "t1v0", "_value": 3},
+            {"t0": "t0v0", "t1": "t1v1", "_value": 3},
+            {"t0": "t0v1", "t1": "t1v0", "_value": 3},
+            {"t0": "t0v1", "t1": "t1v1", "_value": 3},
+        ],
+    )
+        |> group(columns: ["t0", "t1"])
+    got = testing.load(tables: inData)
+        |> range(start: -100y)
+        |> filter(fn: (r) => r._field == "f0")
+        |> group(columns: ["t0", "t1"])
+        |> count()
+        |> drop(columns: ["_start", "_stop"])
+
+    testing.diff(got, want) |> yield()
+}
+
 // Group + sum tests
 testcase group_one_tag_sum {
     want = array.from(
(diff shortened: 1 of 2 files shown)


stdlib/planner/group_agg_test.flux
```diff
@@ -1,338 +1,134 @@
...
+        {_field: "f1", _measurement: "m0", t0: "t0v0", t1: "t1v1", _time: 2021-07-06T23:06:50Z, _value: 3},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:30Z, _value: 3},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:40Z, _value: 2},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:50Z, _value: 1},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:30Z, _value: 1},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:40Z, _value: 0},
+        {_field: "f1", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:50Z, _value: 2},
+    ],
+)
+    |> group(columns: ["_measurement", "_field", "t0", "t1"])
```
fchikwekwe, line 37: This looks like it was probably cumbersome. Thanks for doing it.

stdlib/planner/group_agg_test.flux
```diff
@@ -1,338 +1,134 @@
...
-        {"t0": "t0v1", "_value": 22},
-    ]) |> group(columns: ["t0"])
-    got = testing.loadStorage(csv: input)
-        |> range(start: -100y)
+    want = array.from(
+        rows: [
+            {"t0": "t0v0", "_value": 19},
+            {"t0": "t0v1", "_value": 22},
+        ],
+    )
```
fchikwekwe, line 95: This test case is definitely simpler to read. Do we not need a test case with multiple tags though?

fchikwekwe (commented):
I think this is probably fine. Just questioning not having test cases that group on multiple tags when we were testing that behavior before. I'm wondering if that might allow some regression in the future.

wolffcm:
I added a commit with a couple more tests, grouping with multiple tags for count and sum.

fchikwekwe (approved):
Looks great!

PR #3864, opened 2021-07-08
fix(flux-core): update bootstrap for new structs

BarWeiner1:
### Done checklist
- [ ] bootstrap::stdlib_docs uses the new data structures defined in the fluxdoc module in the readme. This update in structs will allow us to export them in a format that is easily consumable by the docs site.
- [ ] The new structures need only be partially populated. As a transition step place the entire comment string into the headliner field.
- [ ] Test cases written


rockstar (changes requested):
In the future, please add a description to your PRs. I am slightly adjacent to this work, but I'm pretty hazy on the details of the _why_ for this patch. I see "update bootstrap for new structs" and I wonder "Why are there new structs?" "What was wrong with the old ones?" "What do the new structs provide that we didn't get from the old ones?"

I'd like to see the `generate_function_struct` function made into a `FunctionDoc` static function. Feel free to respond and re-request a review.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
     pub msg: String,
 }
 
-/// DocPackage represents the documentation for a package and its sub packages
+/// Doc is an enum that can take the form of any form of flux documentation through polymorphism.
```
rockstar, line 39: "take the form of any form" reads awkwardly. I'd suggest something less formal-sounding, like "The various types of things that can be documented" or something like that. That way, I can look at the enum and go "Oh, hum, comments aren't documented, I guess" (the silliness of comments being documented notwithstanding).

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
     pub name: String,
-    /// the doc
-    pub doc: String,
-    /// the values of the comments
-    pub values: Vec<DocValue>,
+    /// the headline of the package
+    pub headline: String,
+    /// the description of the package
+    pub description: Option<String>,
+    /// the members of the package which are values and functions with their parameters
```
rockstar, line 63: You could say something like "The values and functions of a package" here, and that would tell me all I need.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
+    headline: String,
+    /// the description of the function
+    description: String,
+    /// the parameters of the function
+    parameters: Vec<ParameterDoc>,
+    /// the type of the function
+    flux_type: String,
+}
+
+/// Implementation block for FunctionDoc to house future "FunctionDod" methods
```
rockstar, line 97: I don't think we need this comment. Usually docs that might be put here would be part of the `struct` declaration docs.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
+impl FunctionDoc {
+    /// static method to construct a new FunctionDoc
```
rockstar, line 99: "Construct a new `FunctionDoc` with name, headline, and flux type." might be better here.

rockstar (approved):
Thanks for doing this. I've made a couple of small copy-editing suggestions, but I think this is good work that can be merged when those changes are made.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
-    /// the name of the comments package?
+pub enum Doc {
+    /// Package represents documentation for an entire Flux package.
+    Package(Box<PackageDoc>),
+    /// Value represents documentation for a value exposed from a package.
+    Value(Box<ValueDoc>),
+    /// Builtin represents documentation for a builtin value exposed from a package.
+    Builtin(Box<ValueDoc>),
+    /// Option represents documentation for a option value exposed from a package.
+    Option(Box<ValueDoc>),
```
wolffcm, line 49: `Option` is a standard type in Rust, will `rustc` interpret this as the Rust `Option` type?

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
-    /// the doc
-    pub doc: String,
-    /// the values of the comments
-    pub values: Vec<DocValue>,
+    /// the headline of the package
+    pub headline: String,
+    /// the description of the package
+    pub description: Option<String>,
+    /// the members are the values and funcitons of a package
+    pub members: HashMap<String, Doc>,
```
wolffcm, line 64: By using a hash map here, it implies that the order of the Doc items as they were in the Flux source won't be maintained. Maybe a `Vec` would be more appropriate here? I do not think any package contains so many items that performance is a concern.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -35,32 +36,89 @@ pub struct Error {
...
+#[derive(Debug, Serialize, Deserialize)]
+struct ParameterDoc {
+    /// the name of the parameter
+    name: String,
+    /// the headline of the parameter
+    headline: String,
+    /// the description of the parameter
+    description: Option<String>,
+    /// a boolean indicating if the parameter is required
+    required: bool,
```
wolffcm, line 121: Do we also need a boolean that says whether the parameter is the stream argument?

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -205,12 +266,22 @@ fn generate_values(
...
+                    }
+                    _ => {
+                        let builtin = ValueDoc {
+                            name: name.clone(),
+                            headline: doc,
+                            description: None,
+                            flux_type: typ,
+                        };
+                        members.insert(name.clone(), Doc::Value(Box::new(builtin)));
+                    }
```
wolffcm, line 283: Do we have any builtins that are not functions? If no, then I guess this block would be essentially dead code? But maybe we have non-function builtins, I am really not sure.

libflux/flux-core/src/semantic/bootstrap.rs
BarWeiner1: I don't think so, and it has not impacted the performance of our tests.

libflux/flux-core/src/semantic/bootstrap.rs
BarWeiner1: I think using a hashmap makes it less prone to errors when looping and also the faster lookup times is a plus. Its also more convenient to lookup a Doc by its name rather than by its index.

libflux/flux-core/src/semantic/bootstrap.rs
JustSomeCarbon: From what I am able to tell, and some minor research on rust, I believe in this case the Option keyword is being overloaded, and so does not default to the std::Option type.

scbrickley: In order to avoid confusion, I think it may be worthwhile to change our `Option` type to something like `Opt`. At the very least, it would make the code more readable.

libflux/flux-core/src/semantic/bootstrap.rs
scbrickley: I agree with @BarWeiner1's explanation here. As far as I know, we don't need to keep the docs sorted in any particular order, but I could be wrong. Is there a use case here that I'm not considering @wolffcm?

libflux/flux-core/src/semantic/bootstrap.rs
scbrickley: That might be nice to add in the future, but for now I don't think it's necessary.

libflux/flux-core/src/semantic/bootstrap.rs
scbrickley: I don't know if we have any builtins that aren't functions, but it's technically possible, right?

libflux/flux-core/src/semantic/bootstrap.rs
wolffcm: Say for a site that is generated from the doc comments, we might want to preserve the order of the documentation that was in the Flux source? But we may also want to use alphabetical order or something like that. But it seems clear that we would not want the random order generated by iterating over a hash map.

If we want the items on a generated site to be in alphabetical order, then a hash map is fine, the site generator can take care of the sorting. But if we want to preserve the order of the Flux source, then we would lose the ability to do that. Maybe that's fine though, if this is not a requirement, feel free to disregard.

(I'm not saying that we couldn't also keep a hash map on the side to do fast lookups by name, probably you would want to support both.)

libflux/flux-core/src/semantic/bootstrap.rs
JustSomeCarbon: The change has been made.

libflux/flux-core/src/semantic/bootstrap.rs
scbrickley: That's fair. Maybe it's worth reaching out to the folks who are working on the new docs site. I can take care of that.

libflux/flux-core/src/semantic/bootstrap.rs
wolffcm: Yes, you're right, better to leave it in.

libflux/flux-core/src/semantic/bootstrap.rs
scbrickley: Got some feedback. No interest in preserving the order found in the source code (that's not to say other consumers wouldn't want that at some point in the future), but there was some interest in grouping by type (i.e. All the functions in a package are grouped together, likewise for builtins, options, etc). I think that's a good idea, but it would mean making changes to the spec, and may be better suited for a separate issue. What do you think @wolffcm?

wolffcm: Sounds good to me, thanks for following up.

wolffcm (approved):

PR #3869, opened 2021-07-09
test: add acceptance tests for group with first and last

wolffcm:
This PR adds some tests for `group` with `first` and `last` selectors. These tests will be extended in OSS in order to provide test coverage for push downs.


skartikey (approved):
Looks good!

PR #3874, opened 2021-07-13
build: update `build.rs` and `cgo` comments to be Windows-friendly

danxmoran:
Now that we're publishing Windows builds of OSS, I'd like to be able to update the CI to run tests natively on Windows (via Circle's Windows executor) to flush out any OS-specific problems. This PR updates build-related things that I stumbled across while trying to make a proof-of-concept work.

~Draft until the corresponding `pkg-config` change merges.~


go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/influxdata/influxdb-client-go/v2 v2.3.1-0.20210518120617-5d1fff431040
 	github.com/influxdata/line-protocol v0.0.0-20200327222509-2487e7298839
-	github.com/influxdata/pkg-config v0.2.7
+	github.com/influxdata/pkg-config v0.2.8-0.20210608200742-e5be1a0d54be
```
danxmoran, line 34: See influxdata/pkg-config#35

libflux/go/libflux/analyze.go
```diff
@@ -1,6 +1,7 @@
 package libflux
 
-// #include "influxdata/flux.h"
+// #cgo CFLAGS: -I${SRCDIR}/../../include/influxdata
+// #include <flux.h>
```
danxmoran, line 4: For some reason, `/`s inside of `include` don't work, but `/`s inside of `cgo CFLAGS` _do_ work.

lesam: Empirically windows accepts forward slashes as equivalent to backslashes 'most of the time'. Perhaps header parsing is more pedantic?

rockstar (approved):
I'm going to hit "Approve" here, because this work is good, and obviously fixes an issue we've got. Additionally, I wonder if you'd be okay just adding a (hopefully very shallow) change to just try and build on a Windows Circle instance, just so we don't break this in the future. I'm more concerned about that than what this patch looks like, because no one on the flux team uses Windows, so we won't have _any_ visibility into it otherwise.

danxmoran:
@rockstar job added! Can you take another look?

To test that the job is a good regression-catcher, I made a [new branch](https://github.com/influxdata/flux/tree/dm-windows-build-fail) containing just the changes to `.circleci/config.yml`. The first push of that branch [timed out in CI](https://app.circleci.com/pipelines/github/influxdata/flux/9865/workflows/3e62b4e3-834b-413d-80ad-5cd38e4ad8aa/jobs/28954) because it was missing the [pkg-config fix](https://github.com/influxdata/pkg-config/commit/537acb05005c3096c8b9efa36b24cf8b2720d316#diff-0116498665a516c1a1b9d61b7c676937303c8821292c91f1b285861102394a6aL52-R54) to prevent an infinite loop when searching for the root directory. I pushed a 2nd commit to that branch to fix pkg-config, but leave flux otherwise unchanged; it also [failed CI](https://app.circleci.com/pipelines/github/influxdata/flux/9866/workflows/a576a324-2b34-4731-8aa3-65ae458d8893/jobs/28958) in the `build.rs` because of the hard-coded unix paths there.

PR #3875, opened 2021-07-13
fix(stdlib): Remove the Sleep function

skartikey:
The sleep function was introduced for testing purposes. However, it presents a serious risk to the cloud.
This PR is to remove the sleep function.

Fixes: https://github.com/influxdata/idpe/issues/11188


BarWeiner1 (approved):
If this passes CI it seems like you removed all sleep references effectively. LGTM!

nathanielc:
LGTM can you change the commit message to be `fix(stdlib)` and add `BREAKING CHANGE` to the commit message? That way the change log will pick up this important change. 

Also curious if it was found in the archiver at all?

skartikey:
> LGTM can you change the commit message to be `fix(stdlib)` and add `BREAKING CHANGE` to the commit message? That way the change log will pick up this important change.
> 
> Also curious if it was found in the archiver at all?

I have added BREAKING CHANGE in the body of the commit message. Is that okay or do I have to add it in the header of the commit message?

Here are the results for the sleep scan in archiver. 20 out of every 100,000 query executions contain sleep calls.

```
2021/07/14 23:39:31 running pattern matching
2021/07/14 23:49:43 finishing up
2021/07/14 23:49:43 
        number of queries:                 6132416
        query parse errors:                 754009
        number of executions:           3403406842
        execution parse errors:           27616894
        flux parse errors:                       0
        flux exec parse errors:                  0
        timed executions count:         2735320884
        timed executions sum:  1626668002970520687
2021/07/14 23:49:43 sleep
        query matches:                      599
        execution matches:               542050
        timed matches count:             541458
        timed matches sum:    34394208219063191
        exec representation:        0.000197950
        time representation:        0.021143963
        time over-rep:            106.814422944
```

skartikey:
Cascadeo is the only customer who is using the sleep() function.
_org_id ba4d8fcfba9235ef (563 occurrences)_

There are some influx accounts who are using it too. But quite less in numbers.

PR #3888, opened 2021-07-21
fix(semantic): properly copy record types with no extends parameter

jsternberg:
It seems that the code to copy a monotype did not copy the monotype
properly when the type was a record and it did not extend another
record. This causes compiler substitution to mark itself as extending
`t0` when the original was a fully qualified type with no record
extension.

This may have affected the `map()` call which could possibly heavily
rely on this code or it could have affected some other part of the
compiler infrastructure.

This was found while testing some code for an experiment so I don't have
a reproducer for how this would affect a live system.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/semantic/monotype.go
+++ b/semantic/monotype.go
@@ -794,12 +794,12 @@ func copyMonoType(builder *flatbuffers.Builder, t MonoType) flatbuffers.UOffsetT
 				Value: monoTypeFromFunc(prop.V, prop.VType()),
 			}
 		}
-		extends := record.Extends(nil)
-		var tv uint64
-		if extends != nil {
-			tv = extends.I()
+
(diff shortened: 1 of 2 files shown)


rockstar (approved):
Could we even write a test that would assert this behavior? I imagine the answer is no, which I really don't like, but also understand why that answer might be no.

jsternberg:
That's a good point. We absolutely can write a unit test. I'll do exactly that.

jsternberg:
It seems there was already a test case for the return types, which is what would have been used, and this only affected when constructing a monotype from Go code where the object was nested. I added a test case for that as it could still affect something in an unforeseen way.

PR #3895, opened 2021-07-27
fix(stdlib): Join does not produce columns of equivalent length when …

skartikey:
…combining mismatched schemas

this PR is to fix when join is joining irregular schemas, append nulls in appropriate locations.

The query ran on REPL after my fix - 

```
import "csv"
import "testing"
import "internal/debug"

a = csv.from(csv: "
#datatype,string,long,dateTime:RFC3339,double,string,string,string
#group,false,false,false,false,true,true,true
#default,_result,,,,,,
,result,table,_time,_value,_measurement,_field,host
,,0,2021-01-01T00:00:00Z,0.0,m0,f0,a

#datatype,string,long,dateTime:RFC3339,double,string,string
#group,false,false,false,false,true,true
#default,_result,,,,,
,result,table,_time,_value,_measurement,_field
,,1,2021-01-01T00:01:00Z,0.0,m0,f0
")

b = csv.from(csv: "
#datatype,string,long,dateTime:RFC3339,double,string,string,string
#group,false,false,false,false,true,true,true
#default,_result,,,,,,
,result,table,_time,_value,_measurement,_field,host
,,0,2021-01-01T00:00:00Z,0.0,m0,f0,b
,,0,2021-01-01T00:01:00Z,0.0,m0,f0,b
")

join(tables: {a, b}, on: ["_time"])
```

image

### DIFF
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -856,6 +856,46 @@ func equalJoinkeys(left, right flux.GroupKey) bool {
 	return true
 }
 
+// Find indexes of column names missing between the two tables
+func (c *MergeJoinCache) missingColIdx(lCols, rCols []flux.ColMeta) []int {
+
+	var idxs []int
+
+	if len(lCols) == 0 || len(rCols) == 0 {
+		return idxs
+	}
+
+	trackIdx := make(map[int]struct{})
+
+	for _, lCol := range lCols {
+		column := tableCol{
+			table: c.names[c.leftID],
+			col:   lCol.Label,
+		}
+		newColumn := c.schemaMap[column]
+		newColumnIdx := c.colIndex[newColumn]
+		trackIdx[newColumnIdx] = struct{}{}
+	}
+
+	for _, rCol := range rCols {
+		column := tableCol{
+			table: c.names[c.rightID],
+			col:   rCol.Label,
+		}
+		newColumn := c.schemaMap[column]
+		newColumnIdx := c.colIndex[newColumn]
+		trackIdx[newColumnIdx] = struct{}{}
+	}
+
+	for _, idx := range c.colIndex {
+		if _, ok := trackIdx[idx]; !ok {
+			idxs = append(idxs, idx)
+		}
+	}
+
+	return idxs
+}
+
 func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Table, error) {
 	// Sort input tables
 	left.Sort(c.order, false)
@@ -882,6 +922,9 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
 			return nil, err
 		}
 	}
+
+	// find missing coloumn indexes between two tables
+	missingIdxs := c.missingColIdx(left.Cols(), right.Cols())
 
 	// Perform sort merge join
 	for !leftSet.Empty() && !rightSet.Empty() {
@@ -917,6 +960,11 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
 							_ = builder.AppendValue(newColumnIdx, columnVal)
 						}
 					})
+
+					// append nil for the missing column in the result table
+					for _, mIdx := range missingIdxs {
+						_ = builder.AppendNil(mIdx)
+					}
 				}
 			}
 			leftSet, leftKey = c.advance(leftSet.Stop, left)
--- a/stdlib/universe/join_mismatched_schema_test.flux
+++ b/stdlib/universe/join_mismatched_schema_test.flux
@@ -1,0 +1,60 @@
+package universe_test
+
+
+import "csv"
+import "testing"
+import "internal/debug"
+
+a = csv.from(
+    csv: "
+#datatype,string,long,dateTime:RFC3339,double,string
+#group,false,false,false,false,true
+#default,_result,,,,
+,result,table,_time,_value,key
+,,0,2021-01-01T00:00:00Z,1.0,foo
+,,0,2021-01-01T00:01:00Z,2.0,foo
+
+#datatype,string,long,dateTime:RFC3339,double
+#group,false,false,false,false
+#default,_result,,,
+,result,table,_time,_value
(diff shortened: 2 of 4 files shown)


stdlib/universe/join_test.go
```diff
@@ -720,6 +720,82 @@ func TestMergeJoin_Process(t *testing.T) {
...
+						{Label: "_value_a", Type: flux.TFloat},
+						{Label: "_value_b", Type: flux.TFloat},
+						{Label: "key_a", Type: flux.TString},
+						{Label: "key_b", Type: flux.TString},
+					},
+					KeyCols: []string{"key_b"},
+					Data: [][]interface{}{
+						{execute.Time(1), 1.5, 10.0, nil, nil},
+						{execute.Time(2), 2.5, 20.0, nil, nil},
+					},
```
wolffcm, line 795: Aargh, I find this `join` so hard to reason about. I like `experimental.join` so much better. I *think* this looks like what I would expect. But this answer doesn't seem consistent with the result in the acceptance test?

wolffcm (changes requested):
Very nice work here. This transformation is probably the trickiest one of them all.

I'm requesting that you move the `map` creation outside of performance-sensitive code at least. I'd also like to understand better the result that the acceptance test is getting... seems like something is off there?

stdlib/universe/join.go
```diff
@@ -893,6 +893,8 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
 					leftRecord := left.GetRow(l)
 					rightRecord := right.GetRow(r)
 
+					trackIdx := make(map[int]struct{})
```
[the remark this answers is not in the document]
skartikey: Done!

stdlib/universe/join_mismatched_schema_test.flux
```diff
@@ -0,0 +1,47 @@
...
+        |> debug.slurp()
+
+    want = csv.from(
+        csv: "
+#datatype,string,long,dateTime:RFC3339,double,double,string,string,string,string,string,string
+#group,false,false,false,false,false,true,true,true,true,true,true
+#default,_result,,,,,,,,,,
+,result,table,_time,_value_a,_value_b,_measurement_a,_measurement_b,_field_a,_field_b,host_a,host_b
+,,0,2021-01-01T00:00:00Z,0,0,m0,m0,f0,f0,a,b
+,,0,2021-01-01T00:01:00Z,0,0,m0,m0,f0,f0,,b
```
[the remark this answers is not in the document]
skartikey: I found the issue, I was making missing col host_a as a KeyCol in the result table. This is wrong as the missing col cannot be a KeyCol in the result. 

Example - 

```
table a (KeyCol - "key")
_time   _value    key

table b (KeyCol - "")
_time    _value

Result join (KeyCol - "key_b")
_time    _value_a    _value_b    key_a    key_b
```
previously I was making result table as below 
```
Result join (KeyCol - "key_a", "key_b")
_time    _value_a    _value_b    key_a    key_b
```

stdlib/universe/join_test.go
skartikey: I have replicated the go test in the acceptance test and with less col it's simple to follow through.

stdlib/universe/join.go
```diff
@@ -916,7 +918,20 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
...
+					for _, idx := range c.colIndex {
+						if _, ok := trackIdx[idx]; !ok {
+							_ = builder.AppendNil(idx)
+						}
+					}
+
+					// clear map
+					for k := range trackIdx {
+						delete(trackIdx, k)
+					}
```
wolffcm, line 934: Do you know if the set of columns that need to be padded will be the same for every iteration of the innermost loop? I'm wondering if we could build `trackIdx` before this loop:
https://github.com/influxdata/flux/pull/3895/files#diff-362f933cc9740993880d5db64b38c8776a74dd18d13a458c91801416b1a10f08L890
And then clear it afterwards.

stdlib/universe/join_test.go
```diff
@@ -720,6 +720,158 @@ func TestMergeJoin_Process(t *testing.T) {
...
+						{execute.Time(1), 1.0, 10.0, "foo", "bar"},
+						{execute.Time(2), 2.0, 20.0, "foo", "bar"},
+					},
+				},
+				{
+					ColMeta: []flux.ColMeta{
+						{Label: "_time", Type: flux.TTime},
+						{Label: "_value_a", Type: flux.TFloat},
+						{Label: "_value_b", Type: flux.TFloat},
+						{Label: "key_a", Type: flux.TString},
```
wolffcm, line 788: Why does this table have a `key_a` column? In the input table for `a` there as no `key` column. Is it because the first table in `a` had that column, and all the columns in the output of the join must have the same schema?

If that is the case then I would expect that `key_a` to also be in the group key, since the group key is sort of part of the schema too?

stdlib/universe/join_mismatched_schema_test.flux
```diff
@@ -0,0 +1,60 @@
...
+,result,table,_time,_value_a,_value_b,key_a,key_b
+,,0,2021-01-01T00:00:00Z,1.0,10.0,foo,
+,,0,2021-01-01T00:01:00Z,2.0,20.0,foo,
+
+#datatype,string,long,dateTime:RFC3339,double,double,string,string
+#group,false,false,false,false,false,false,true
+#default,_result,,,,,,
+,result,table,_time,_value_a,_value_b,key_a,key_b
+,,1,2021-01-01T00:00:00Z,1.5,10.0,,
+,,1,2021-01-01T00:01:00Z,2.5,20.0,,
```
wolffcm, line 55: Ok, looking at this output I guess it makes sense. `key_a` is in the output because the column exists in at least one table in `a`. It is not in the group key because that column didn't exist in the part of the rows that were joined to form this output.

wolffcm (commented):
The tests look good, I still have some questions about the best ways to track the columns that need padding.

wolffcm (approved):
Looks great!

Issue #3890, opened 2021-07-23
Replace occurrences of *array.Binary with StringArray


wolffcm:
This comment has some background on the rationale for this work: https://github.com/influxdata/flux/issues/3866#issuecomment-885789877

Based on work did in #3889, replace occurrences of Arrow string arrays of type `*array.Binary` with the type `*StringArray` instead.

This should be a refactor that has no change in behavior. Where appropriate instances of `*StringArray` should be created that just wrap Arrow arrays.

DOD:
- Most places in the code that reference `*array.Binary` are replaced with `*StringArray`
- Code still compiles and all tests pass

PR #3897, opened 2021-07-28
feat(array): create flux array package

### LINKED ISSUE #3889: Create a StringArray struct type
This comment has some background on the rationale for this work: https://github.com/influxdata/flux/issues/3866#issuecomment-885789877

In order to facilitate representing group key columns as a single value plus a length (rather than a fully materialized array), create type that can represent either. Probably it will look something like this:
```go
import "github.com/apache/arrow/go/arrow/array"

type StringArray struct {
	value string
	isNull bool
	length int

	arr *array.Binary
}
```

This type should have all of the methods for of `*array.Binary` and should simply forward the method call into `arr` if it is not `nil`. Otherwise, methods should return values consistent with an array of length `length` with the same `value` in each element.

The methods to be implemented will be those in the Arrow's `array.Interface`:
https://github.com/apache/arrow/blob/42e5726b543135f18db1560278e062b9a7941f5c/go/arrow/array/array.go#L27
And also those that are specific to `*array.Binary`:
https://github.com/apache/arrow/blob/42e5726b543135f18db1560278e062b9a7941f5c/go/arrow/array/binary.go#L43-L102

DOD:
- New struct type `StringArray` exists
- There are unit tests that show it forwards methods to `arr` if it exists, or returns data consistent with a column that has the same value in every element

jsternberg:
The flux array package has been created to be the location where all
flux array implementations are kept.

The majority of the array implementations are type aliases to the
equivalent arrow array implementation and the interfaces and builders
have also been aliased or forwarded to arrow. Others, like string, are
wrappers around arrow implementations.

All of the locations that previously referenced the arrow array
implementations have now been changed to flux array implementations.

The intention behind this is so that we can control the naming and
implementations of the column representations while not having to modify
all the locations where the code is used. This is most useful with
string where we are creating our own more optimized implementation, but
helps with consistency across the different packages.

Fixes #3889 and #3890.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


array/array.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+//go:generate -command tmpl ../gotool.sh github.com/benbjohnson/tmpl
+//go:generate tmpl -data=@numeric.tmpldata -o numeric.gen.go numeric.gen.go.tmpl
+
+// Interface represents an immutable sequence of values.
+type Interface = array.Interface
+
+// Builder provides an interface to build arrow arrays.
+type Builder interface {
```
wolffcm, line 16: Might be good to note here that you are mimicking Arrow's builder interface, albeit without the unexported methods.

array/array.go
```diff
@@ -0,0 +1,253 @@
...
+}
+func (b *BooleanBuilder) NewBooleanArray() *Boolean {
+	return b.b.NewBooleanArray()
+}
+
+type String struct {
+	value  string
+	length int
+	data   *array.Binary
+}
```
wolffcm, line 109: Why don't you need an `isNull` field here?

array/array.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (a *String) DataType() arrow.DataType {
+	return arrow.BinaryTypes.String
+}
+func (a *String) Data() *array.Data {
+	if a.data != nil {
+		return a.data.Data()
+	}
+	return nil
```
wolffcm, line 118: Should this panic instead of returning `nil`?

array/numeric.gen.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+type (
+	Int   = array.Int64
+	Uint  = array.Uint64
+	Float = array.Float64
+)
+
+type IntBuilder struct {
+	b *array.Int64Builder
```
wolffcm, line 21: I guess you could have omitted the field name `b` here and then you would have gotten all of the methods from the embedded builder. Why not just do that?

wolffcm (changes requested):
This looks very nice. I had a question about null values in `String` columns. 

Can you add tests for the `String` column that shows it works for both the constant value and an underlying Arrow array?

array/array.go
jsternberg: I generally thought it was unlikely to provide a lot of utility. In general, group keys aren't null very often if at all. If they do end up being null, the native arrow representation of them isn't very intense. It's a bitmap with the number of elements divided by 8 for the size.

I figured that adding an additional field to the struct would likely end up costing more in terms of memory overall because the situation where a string column is completely filled with null values would be so rare.

If we want to add `isNull` anyway I can modify this. I just figured we'd wait before optimizing that situation.

array/array.go
jsternberg: I'm wondering if we should maybe just remove it? It's primarily used for slicing or it would be used for direct access to arrow buffers if we wanted to perform serialization in the future. It might be worth removing this from the interface and modifying the `Slice()` function to either be part of the interface or smarter about how it slices.

array/numeric.gen.go
jsternberg: I intentionally did this so that we don't automatically expose the embedded builder methods. I'd like to keep our API explicit. When you embed a type, you add a public field to the struct that becomes part of the API and now can't be removed without breaking API compatibility. The lowercase field attribute keeps the fields in this struct unexported and able to be changed for an internal representation.

array/array.go
wolffcm: I would be in favor of removing it.

array/array.go
wolffcm: Ok, fair enough. I agree it's not a common case. We don't just append to a bitmap though, we still need to reserve space for a pointer to the value that does not exist, right?

wolffcm (approved):

PR #3898, opened 2021-07-28
fix(libflux): restore the previous libflux include paths to use pkg-config

jsternberg:
The libflux package was modified to manually specify `CFLAGS` for the
include path. This worked for local builds where the files were located
in the source directory, but didn't work for builds that used flux as a
dependency like influxdb.

This restores the old behavior that uses the `/include` directory as the
base of the include path and includes `"influxdata/flux.h"` instead of
directly including `"flux.h"`. This should restore the previous
behavior.

It is up to `pkg-config` to produce a usable include path.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/go/libflux/README.md
+++ b/libflux/go/libflux/README.md
@@ -1,0 +1,32 @@
+# CGO Build Details
+
+This package is a wrapper interface for the flux library.
+
+The flux library is written in Rust and exposes a C ABI (Application Binary Interface).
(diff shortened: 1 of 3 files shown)


rockstar (approved):
Ah! There was a question about this change, but I had assumed it was tested and had passed. Thanks for getting to this.

PR #3900, opened 2021-07-28
test: test query fields with only a measurement predicate

wbaker85:
While working on https://github.com/influxdata/influxdb/pull/21962, I found that the existing `show_fields_xxx.flux` tests did not include one where there is only a single predicate of `_measurement`. `show_fields_with_pred_test.flux` has a predicate of `_measurement` as well as one for an additional tag, and `show_fields_test.flux` does not have a filter.

This additional test will cover the case of only having a predicate of `_measurement`. It is primarily useful to me since I am planning to tweak the OSS logic related to a query like this.


fchikwekwe (changes requested):
@wbaker85 we've been migrating our existing end to end tests to a newer format that uses the `testcase` keyword instead of the `test` keyword. I don't think we have a guide for this yet, but here's an example https://github.com/influxdata/flux/blob/master/stdlib/universe/pivot_table_test.flux. Happy to pair if you need help converting your test.

fchikwekwe (approved):
Thanks for making that change and for adding this test!

PR #3902, opened 2021-07-29
fix: derive Copy on ast::Position

rockstar:
This patch derives `Copy` on `ast::Position`. `Clone` was likely
implemented because structs that use `Position` as a member attribute
are also `Clone`, but there is no reason a `ast::Position` can't also
just be `Copy` as well, since both members are `Copy`.

`flux-lsp` has a couple of `clone` calls that could easily be copies
instead as a result of the lack of `Copy` here.


skartikey (approved):
LGTM. It makes sense to have a Copy trait for type Position with both members(u32) having Copy.

skartikey (approved):

PR #3903, opened 2021-07-29
chore: implement lsp_types adapters

rockstar:
I'm not sure if this patch starts us in a direction we want, but I'd like
to at least have a conversation about this. In `flux-lsp`, I need these
kinds of conversions for a few things. Unfortunately, because the struct
doesn't originate in my crate, I have to use conversion functions to do
this work. It'd be much nicer if I could just tell the compiler how to
convert between the types.


adrian-thurston (approved):
It's good with me for something like a position struct. Simply common kinds of structs I think are okay to make conversions for. I've done it many times.

PR #3904, opened 2021-07-29
ci: disable the windows build until it uses the same rust version as the linux build

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,52 +84,52 @@ jobs:
     steps:
       - checkout
       - run: make test-valgrind
-  build-windows:
-    machine:
-      image: windows-server-2019-vs2019:stable
-    resource_class: windows.medium
-    shell: bash.exe -eo pipefail
-    steps:
-      - run:
-          name: Install system dependencies
-          command: |
-            choco upgrade golang --version=1.16.6
-
-            choco install \
-              llvm \
-              pkgconfiglite \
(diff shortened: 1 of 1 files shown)


rockstar (approved):

.circleci/config.yml
```diff
@@ -84,52 +84,52 @@ jobs:
...
-            mkdir -p "${GOPATH}/bin"
-            echo 'export PATH="${GOPATH}/bin:${PATH}"' >> $BASH_ENV
-            go install github.com/influxdata/pkg-config
-      - run:
-          name: Generate libflux
-          command: go generate ./libflux/go/libflux
-      - run:
-          name: Build flux
-          command: go build ./...
+#  build-windows:
```
fchikwekwe, line 87: Don't block on this question. Should we add a comment with the criteria for re-enabling this rule? Or an issue link?

fchikwekwe (approved):

.circleci/config.yml
jsternberg: It might be good but I figured that @danxmoran would probably have a PR ready within the day. I just didn't know enough about how to fix it on windows to be capable of updating the Rust version myself.

fchikwekwe: True, no worries then.

PR #3908, opened 2021-07-29
build: re-enable build-windows, pin rust version

danxmoran:
Add an explicit `rustup default` call to the Windows CI logic, to ensure new versions aren't used accidentally.

I couldn't think of a way to declare the `RUST_VERSION` constant in a single place without significantly refactoring the existing docker-based environment setup, so I settled for adding a comment above each spot the version appears referring to the other location.

### DIFF
--- a/Dockerfile_build
+++ b/Dockerfile_build
@@ -4,6 +4,9 @@
 # and verification, we can list the rust container as a prior build stage, and
 # then pull in the artifacts we need. There is an added benefit that tagged versions
 # also include minory releases, so 1.2 includes 1.2.1 and so on, for bugfix releases.
+#
+# NOTE: If you change this, you must also change `RUST_VERSION` in the `build-windows`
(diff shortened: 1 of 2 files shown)


Dockerfile_build
```diff
@@ -4,6 +4,9 @@
 # and verification, we can list the rust container as a prior build stage, and
 # then pull in the artifacts we need. There is an added benefit that tagged versions
 # also include minory releases, so 1.2 includes 1.2.1 and so on, for bugfix releases.
+#
+# NOTE: If you change this, you must also change `RUST_VERSION` in the `build-windows`
+# job in `.circleci/config.yml`.
```
rockstar, line 9: Thanks for this comment. It's the one thing I was looking for when I saw the review request come in.

rockstar (approved):

jsternberg (approved):
Looks good to me.

PR #3913, opened 2021-07-30
feat(execute): expand message interface with message lifetime controls

### LINKED ISSUE #3884: Expand message interface to include message ack and dup
Expand the `Message` interface in the `execute` package to include an ack and dup mechanism so that messages have a lifetime and can release memory associated with the message.

It should add the following methods in addition to the current methods:

```go
type Message interface {
    // Ack is used to acknowledge that the Message was received
    // and terminated. A Message may be passed between various
    // Transport implementations. When the Ack is received,
    // this signals to the Message to release any memory it may
    // have retained. Ack is idempotent.
    Ack()

    // Dup is used to duplicate the Message.
    // This is useful when the Message has to be sent to multiple
    // receivers from a single sender.
    Dup() Message
}
```

The methods will be no-ops for all of the current message types except for `ProcessMsg` where an implementation and tests will be needed.

jsternberg:
The `execute.Message` interface has been expanded with `Ack()` and
`Dup()` methods.

The `Ack()` method is meant to acknowledge that the message has been
received by a `Transport` and the contents of the message can be
released.

The `Dup()` method is meant to copy the inner contents of a message into
a new message. Both messages must be acked separately. The
implementation of `Dup()` may involve incrementing a reference count
instead of a full deep copy.

These lifetimes will allow messages to contain data that has to be
released.

Fixes #3884.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/transport.go
+++ b/execute/transport.go
@@ -263,18 +263,48 @@ func processMessage(ctx context.Context, t Transformation, m Message) (finished 
 	return
 }
 
+// Message is a message sent from one Dataset to another.
 type Message interface {
+	// Type returns the MessageType for this Message.
 	Type() MessageType
+
+	// SrcDatasetID is the DatasetID that produced this Message.
 	SrcDatasetID() DatasetID
+
+	// Ack is used to acknowledge that the Message was received
+	// and terminated. A Message may be passed between various
+	// Transport implementations. When the Ack is received,
+	// this signals to the Message to release any memory it may
+	// have retained.
+	Ack()
+
+	// Dup is used to duplicate the Message.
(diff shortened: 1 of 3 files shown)


jsternberg:
@nathanielc @adrian-thurston along with the normal code review, I'm also looking to get feedback on comments and documentation to make the message interface clear to work with. One of the bits of documentation that won't be part of this PR is for the `ProcessMessage` method which will more clearly specify the lifetime and ownership of the message. That would have to be part of the PR that changes the transport interface.

adrian-thurston (approved):

PR #3914, opened 2021-07-30
feat(array): modify the string array builder to support constant data

### LINKED ISSUE #3891: Update Flux's StringBuilder to produce a constant representation when values are all the same
The first time a value is appended to a StringBuilder, use the constant single-value+length representation for the output.
If the subsequent values are all the same, just increment the length counter. If we ever find a value that is different, then switch to using the generic Arrow builder.

Location of `StringBuilder`:
https://github.com/influxdata/flux/blob/cd7418cd761175fbfe8a7c2af62bdbdeba1f437a/array/array.go#L225

We feel that this change, while pervasive, is safe for production because we have lots of acceptance tests that cover all of our transformations, and that the abstractions provided are not "leaky".

DOD:
- Make it so StringBuilder will produce a single-value+length representation of a column if all the values are the same. Otherwise, just use the standard arrow builder.
- Add unit tests that show creating a column with values that are all the same will produce a compressed column

jsternberg:
Fixes #3891.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -223,56 +223,154 @@ func (a *String) ValueLen(i int) int {
 }
 
 type StringBuilder struct {
-	b *array.BinaryBuilder
+	builder      *array.BinaryBuilder
+	mem          memory.Allocator
+	value        string
+	length       int
+	capacity     int
+	dataCapacity int
+	refCount     int
 }
 
 func NewStringBuilder(mem memory.Allocator) *StringBuilder {
 	return &StringBuilder{
-		b: array.NewBinaryBuilder(mem, arrow.BinaryTypes.String),
+		mem:      mem,
+		refCount: 1,
 	}
 }
+func (b *StringBuilder) init() {
+	if b.builder == nil {
+		if b.refCount <= 0 {
+			return
+		}
+
+		builder := array.NewBinaryBuilder(b.mem, StringType)
+		if capacity := b.Cap(); capacity > 0 {
+			builder.Resize(capacity)
+			dataCapacity := len(b.value) * capacity
+			if dataCapacity < b.dataCapacity {
+				dataCapacity = b.dataCapacity
+			}
+			builder.ReserveData(dataCapacity)
+		}
+		if b.length > 0 {
+			for i := 0; i < b.length; i++ {
+				builder.AppendString(b.value)
+			}
+		}
+
+		for i := 1; i < b.refCount; i++ {
+			builder.Retain()
+		}
+		b.builder = builder
+	}
+}
+func (b *StringBuilder) reset() {
+	b.builder = nil
+	b.length = 0
+	b.capacity = 0
+	b.dataCapacity = 0
+	b.value = ""
+}
 func (b *StringBuilder) Retain() {
-	b.b.Retain()
+	if b.builder != nil {
+		b.builder.Retain()
+		return
+	}
+	b.refCount++
 }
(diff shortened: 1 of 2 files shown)


array/array.go
```diff
@@ -223,56 +223,146 @@ func (a *String) ValueLen(i int) int {
 }
 
 type StringBuilder struct {
-	b *array.BinaryBuilder
+	b            *array.BinaryBuilder
```
rockstar, line 226: (Not really part of this review, but a thing I'd like to bring up) This is a thing I see in go more than I ever did in even C: what is `b`? Why do we prefer the single letter variable name over, say, `builder`?

array/array.go
```diff
@@ -223,56 +223,146 @@ func (a *String) ValueLen(i int) int {
...
+	mem          memory.Allocator
```
rockstar, line 227: Is `allocator` a better name, or does the interface present itself as if it were memory?

array/array.go
```diff
@@ -223,56 +223,146 @@ func (a *String) ValueLen(i int) int {
...
 
 func NewStringBuilder(mem memory.Allocator) *StringBuilder {
 	return &StringBuilder{
-		b: array.NewBinaryBuilder(mem, arrow.BinaryTypes.String),
+		mem:      mem,
+		refCount: 1,
+	}
+}
+func (b *StringBuilder) init() {
+	if b.b == nil {
```
rockstar, line 242: Yeah, this is what I mean: `b` is not a great name, but it's kinda idiomatic to go to same your "self" variable like that. `b.b` on the other hand? I have no idea what that is.

I'm not sure there's anything actionable now, because team inertia, etc, but I think it'd be a net benefit for our team if we started thinking about whether we want to move away from this.

array/array_test.go
```diff
@@ -21,7 +21,7 @@ func TestString(t *testing.T) {
 					b.Append("a")
 				}
 			},
-			sz: 256,
+			sz: 0,
```
rockstar, line 24: A quick read of this test indicates that we're changing the behavior of this api, and so the test needs updating. Is that right?

rockstar (approved):
I kinda got a bit tangential on this review, mostly in an effort to make us more aware of an issue, but no changes are directly needed. I had a small question, but other than that, this is probably good to go.

array/array.go
jsternberg: I think that's just something Go does? I don't really have an objection to changing this to `builder`. I just don't like typing that much.

array/array.go
jsternberg: Generally I've used `mem` for things that directly use the arrow memory allocator because that's what the parameter is called in the arrow library. I find `alloc` to be a bit weird to read and `allocator` too long.

An example for the arrow library: https://pkg.go.dev/github.com/apache/arrow/go/arrow/array#NewBinaryBuilder.

array/array.go
jsternberg: I can just change it to builder. I don't really have a problem with that.

array/array_test.go
jsternberg: Yes. I wrote the test before implementing this. Now that the builder produces the more optimized format, and this optimized format doesn't use arrow buffers at all, it now "allocates nothing". That's not strictly true but that's true for the sake of this.

PR #3922, opened 2021-08-02
feat(array): add optimized repeat function for arrow arrays

### LINKED ISSUE #3906: Update the Repeat utility to create a single-value+length array if possible
Update `Repeat` to use under the hood a new low-level `Repeat` function from package `flux/array`.

Here is `Repeat`:
https://github.com/influxdata/flux/blob/529ab71b66ce695837db21a94be4adb8b9f69926/arrow/repeat.go#L15

When it's creating a string, it should just create a constant string instead of calling the builder over and over.

DOD:
- For strings, `Repeat` works by producing a single-value+length array and does not call builder over and over
- Unit tests for these new methods

jsternberg:
This adds a `Repeat()` method to the `array` package that mimics the
pattern in from the `flux/arrow` package.

This `Repeat()` differs slightly in that it takes an interface as its
input instead of a `values.Value` and it differs in that it composes an
optimized version of string without repeated iterations over the string
builder.

The `arrow.Repeat()` function has been modified to unwrap the value and
make use of the `array.Repeat()` function instead of its own
implementation.

Closes #3906.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


array/builder.go
```diff
@@ -0,0 +1,157 @@
...
+
+type StringBuilder struct {
+	builder      *array.BinaryBuilder
+	mem          memory.Allocator
+	value        string
+	length       int
+	capacity     int
+	dataCapacity int
+	refCount     int
+}
```
wolffcm, line 16: Can you add some comments here that describe why we have a non-generated builder for strings, and when it makes the switch over to the arrow builder?

array/builder.go
```diff
@@ -0,0 +1,157 @@
...
+func NewStringBuilder(mem memory.Allocator) *StringBuilder {
+	return &StringBuilder{
+		mem:      mem,
+		refCount: 1,
+	}
+}
+func (b *StringBuilder) init() {
+	if b.builder == nil {
+		if b.refCount <= 0 {
+			return
```
wolffcm, line 27: When would this happen? Should we panic here instead?

wolffcm (changes requested):
This looks great. Can you just add some tests for the string builder that show when allocated memory increases vs when it doesn't?

array/builder.go
jsternberg: This can happen when the builder is released and then attempted to be used. I'm not sure a panic is worth it though. This shouldn't really come up as generally the builders shouldn't be used with a reference count of zero. If you want me to add a panic though we can.

array/builder.go
jsternberg: Sure.

wolffcm (approved):
Ok, the string builder tests already exist---I missed that.  Looks good!

array/builder.go
wolffcm: It seems like if tracking memory is important enough, we want to know about errors like this. But we also don't want to fail the query if it wouldn't otherwise fail. I'm conflict but I guess I'm fine with leaving it as-is.

Issue #3851, opened 2021-07-01
Populate the function paramter list in the doc structs


nathanielc:
Once #3783 and #3850 are complete we will have doc structs that correctly parse the doc comment markdown for the headliner and the description. The next step is to parse the parameter list in the markdown and add each ParameterDoc struct to the FunctionDoc struct. 

See FLUXDOC.md for the specification of how the parameter list should be formatted in markdown.

Like in #3850 we should use the pulldown-cmark crate to do this parsing. 

DOD:
- The parameter list can be parsed
- Tests exists that show we can correctly parse the parameter list.

PR #3923, opened 2021-08-03
feat(bootstrap.rs): populate function parameter list

JustSomeCarbon:
This pull request will satisfy the requirements and specifications listed in #3851. Function documentation is correctly parsed and ParameterDoc structures are correctly populated for their associated FunctionDoc.

### Done checklist
- [x] The parameter list is being parsed
- [x] Tests exist that show parameter lists are being correctly parsed
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/flux/src/lib.rs
```diff
@@ -1123,17 +1123,31 @@ from(bucket: v.bucket)
...
+        let mut got = String::new();
+        for d in doc {
+            if d.name == "csv" {
+                // println!("GOT CSV DOC: {:?}", d);
```
BarWeiner1, line 1131: lets take out the comments here

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -96,23 +97,24 @@ pub struct FunctionDoc {
...
-        FunctionDoc {
-            name,
-            headline,
-            description,
-            parameters: vec![],
-            flux_type: typ,
-        }
-    }
-}
+// impl FunctionDoc {
```
rockstar, line 100: If we don't need this code, let's go ahead and delete it.

libflux/flux/src/lib.rs
```diff
@@ -1123,17 +1123,31 @@ from(bucket: v.bucket)
...
-        let got = format!("{:?}", d[0]); // the csv DocPackage location
+        let doc = docs();
+        let stat = r#"PackageDoc { name: "array", headline: "Package array provides functions for building tables from flux arrays.", description: None, members: {"from": Function(FunctionDoc { name: "from", headline: "from constructs a table from an array of records. ", description: "Each record in the array is converted into an output row or record. Allrecords must have the same keys and data types. ## Build an arbitrary table\n```\nimport \"array\"\n\nrows = [\n  {foo: \"bar\", baz: 21.2},\n  {foo: \"bar\", baz: 23.8}\n]\n\narray.from(rows: rows)\n```\n\n## Union custom rows with query results\n```\nimport \"influxdata/influxdb/v1\"\nimport \"array\"\n\ntags = v1.tagValues(\n  bucket: \"example-bucket\",\n  tag: \"host\"\n)\n\nwildcard_tag = array.from(rows: [{_value: \"*\"}])\n\nunion(tables: [tags, wildcard_tag])\n```\n\n", parameters: [ParameterDoc { name: "rows", headline: " is the array of records to construct a table with.", description: None, required: false }], flux_type: "" })} }"#;
```
BarWeiner1, line 1142: All the "/n" tags are appearing in the docs site like they are here when we are making a call to docs(), and also in the html file. Do we know what is putting them in the string representation?

BarWeiner1 (approved):
Change looks good overall.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -270,6 +272,115 @@ fn seperate_description(all_comment: &str) -> (String, Option<String>) {
...
+            Event::End(pulldown_cmark::Tag::List(None)) => {
+                if param_flag {
+                    param_flag = false;
+                }
+            }
+            Event::End(_) => {
+                tmp = &mut funcdocs.description;
+            }
+            _ => {
+                println!("Unsupported type found!");
```
rockstar, line 377: There are two things I wonder about here: (1) should this just `panic`? If the matcher isn't already exhaustive, and this is run as part of our build and not something user facing, I'd rather and unsupported type cause a panic we have to look at rather than a log message that can get lost in scroll. (2) What is the unsupported type? Can we add that to the message, whether we just `println` or `panic`? When we experience this, it's likely going to be because we added a new event here, and we'll want to know immediately what we need to do.

libflux/flux/src/lib.rs
JustSomeCarbon: I am not too sure why that would be the case.

libflux/flux/src/lib.rs
JustSomeCarbon: I have removed the unused code.

libflux/flux/src/lib.rs
```diff
@@ -1119,21 +1119,30 @@ from(bucket: v.bucket)
...
+        let mut got = String::new();
+        for d in doc {
+            if d.name == "csv" {
+                got = format!("{:?}", d);
```
rockstar, line 1128: When we assert on the stringified version, what we're testing is the stringifying of the object. Rather than converting to a string, `stat` should be a `PackageDoc` instance that we test against.  That was the real issue with this test.

rockstar (changes requested):
Thanks for doing this work. There's a couple of questions I have, and I think the test needs to go a little further.

libflux/flux-core/src/semantic/bootstrap.rs
JustSomeCarbon: Right now as the function is, It is built to only look for what was necessary for separating the parameter documentation from the rest of the documentation and building everything we need for that. So the function does not take into account a lot of other events that occur; for example, at the end of a line we will have a `soft break` event that signifies there is more text on the next line. I had no use for looking for `soft break` events and so they are registered as the default unrecognized event. There are many other events that I do not have it looking for but are common and occur normally in a standard doc comment. I don't think a `panic` would be the best course of action for these unused events, but I could possibly add what the unrecognized event was in the `println` if that is helpful.

libflux/flux/src/lib.rs
JustSomeCarbon: I understand, I will reconstruct the tests accordingly.

libflux/flux-core/src/semantic/bootstrap.rs
JustSomeCarbon: I have removed the unused code.

BarWeiner1:
I was testing this out with a few packages besides csv like JSON, interpolate, and dict and the members and their params are just not being picked up in the output for some reason

BarWeiner1 (changes requested):
before merging this we need to debug with a few more packages that are just not getting picked up by this

libflux/flux-core/src/semantic/bootstrap.rs
nathanielc: If these are events that we can safely ignore lets add a comment to that fact and then remove the println statement. No need to print information about expected behavior.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -96,23 +97,24 @@ pub struct FunctionDoc {
...
+//     ) -> FunctionDoc {
+//         FunctionDoc {
+//             name,
+//             headline,
+//             description,
+//             parameters,
+//             flux_type: typ,
+//         }
+//     }
+// }
```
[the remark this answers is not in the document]
JustSomeCarbon: I have removed the code as its no longer used.

libflux/flux/src/lib.rs
JustSomeCarbon: The tests for `csv` and `array` have now been updated to test the `PackageDoc` structures.

libflux/flux-core/src/semantic/bootstrap.rs
JustSomeCarbon: The print statements have been removed an replaced with simple comment placeholders.

BarWeiner1 (approved):

nathanielc (approved):
Looks great!

libflux/flux-core/src/semantic/types.rs
```diff
@@ -1011,6 +1011,20 @@ impl Record {
             },
         }
     }
+    pub(crate) fn find_prop(&self, p: &str) -> Option<MonoType> {
+        match self {
+            Record::Empty => None,
+            Record::Extension { head, tail } => {
```
rockstar, line 1017: Wow, this is neat. I didn't know you could do this in a matcher.

libflux/flux/src/lib.rs
```diff
@@ -1119,21 +1122,170 @@ from(bucket: v.bucket)
 
     #[test]
     //CSV package's docs are accessed from their specific index in the vector of DocPackages, ensuring that they are accessible and correct
-    // XXX: rockstar (5 Aug 2021) - This test can be flaky. For example, there was a
-    // failure because another patch was merged that added a new stdlib, which pushed
-    // `csv` from the 19th item to the 20th item.
```
rockstar, line 1124: Awesome, thank you so much for fixing this test.

rockstar (approved):

PR #3924, opened 2021-08-03
feat(execute): simplify the transport interface and add a transformation adapter

### LINKED ISSUE #3886: Simplify Transport interface and add Transformation Adapter
Simplify the Transport interface from:

```go
type Transport interface {
    Transformation
    Finished() <-chan struct{}
}
```

To:

```go
type Transport interface {
    // ProcessMessage will cause this Transport to process the Message.
    // The Message is owned by the receiver and must be either passed to another Transport
    // or acknowledged using Ack after the message has been processed.
    ProcessMessage(m Message) error
}

type AsyncTransport interface {
    Transport
    Finished() <-chan struct{}
}
```

The executor should be updated to use `AsyncTransport` instead of `Transport` and the consecutive transport should be changed to use `Transport` instead of `Transformation`.

In addition to the above changes, a new method and type should be created with the following signature:

```go
// WrapTransformationInTransport will wrap a Transformation into
// a Transport to be used for the execution engine.
func WrapTransformationInTransport(t Transformation, mem memory.Allocator) Transport {
    if tr, ok := t.(Transport); ok {
        return tr
    }
    return ...
}
```

This function should be called in existing areas that take a `Transformation` and need to convert it into a `Transport`. The function should start off by checking if the `Transformation` implements the `Transport` interface and perform no wrapping when this is the case. Otherwise, it should wrap it in an implementation that consumes the messages it knows and invokes the appropriate method on the underlying transformation.

If `ProcessTableChunk` is received, it will retain that table chunk and append it to a list of table chunks with the same group key. When `FlushKey` is received for that group key, the table chunks will be sent to the `Process` method as a `flux.Table`. When `Finish` is received, any group keys that have not been flushed will be sent as if they had been flushed.

The `consecutiveTransport` type should continue to implement both `Transport` and `Transformation`. In the future, we may refactor this, but this reduces the amount of code we need to change.

jsternberg:
The `Transport` interface is now just `ProcessMessage` which can be
implemented directly by a transformation rather than using the
`Transformation` interface.

A transformation adapter has been created to wrap the older
`Transformation` into a `Transport` to support the new message types and
adapt them for the `Transformation` interface.

`consecutiveTransport` has been updated to implement both `Transport`
and `Transformation` to allow it to be used with either a
`Transformation` or a `Transport` and to forward the message to the
appropriate methods.

Closes #3886.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/transport_test.go
+++ b/execute/transport_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/table"
 	"github.com/influxdata/flux/execute/table/static"
+	"github.com/influxdata/flux/mock"
 )
 
 func TestProcessMsg(t *testing.T) {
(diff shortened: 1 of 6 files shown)


scbrickley (approved):
I don't see anything here that stands out to me as needing changes, but I'm not super familiar with this area of the codebase. I think I'd like to see one other reviewer on this, but apart from that, I don't have any specific reason to block this.

execute/transport_test.go
```diff
@@ -162,3 +163,143 @@ func TestProcessChunkMsg(t *testing.T) {
...
+				chunk.Retain()
+				m := execute.NewProcessChunkMsg(chunk)
+				return tr.ProcessMessage(m)
+			}); err != nil {
+				return err
+			}
+
+			if got, want := len(got), processed; got != want {
+				t.Errorf("wrong number of tables processed -want/+got:\n\t- %d\n\t+ %d", want, got)
+			}
```
wolffcm, line 211: I must be missing something here. At this point in the test, we have sent the process chunk message to `tr`, so I would expect the chunk to be copied and appended to `got`, giving it a length of 1. But we will not increment `processed` until line 217, so it will still be zero. How is this test passing?

wolffcm: Ok, I get what my confusion was. The adapter won't actually invoke the legacy `Process` method until the flush key message is sent, because at that point it will know that it has a complete table. So `got` is not appended yet at this point.

execute/transport_test.go
```diff
@@ -162,3 +163,143 @@ func TestProcessChunkMsg(t *testing.T) {
...
+				m := execute.NewProcessChunkMsg(chunk)
+				return tr.ProcessMessage(m)
+			}); err != nil {
+				return err
+			}
+			processed++
+
+			if got, want := len(got), 0; got != want {
+				t.Errorf("wrong number of tables processed -want/+got:\n\t- %d\n\t+ %d", want, got)
+			}
```
wolffcm, line 278: Ok, now I'm confused again. There was no flush key message issued here, so I would not expect `got` to have been appended to yet. How does this test pass?

wolffcm (commented):
I think this looks good, but I have some questions before you merge this.

I had a question about a test in `transport_test.go`. Could you add some comments to the tests to walk readers through what's going on in a little more detail?

Are there any performance implications to this work? I'm thinking of queries where we might have a table stream with lots of little tables, will this add more overhead?

I see you've also had to touch code that creates spans and sets the node ID. Do you have confidence that tracing and/or the useful error messages haven't been lost---we have tests for both of those, correct?

execute/transport_test.go
jsternberg: I think you might have misread this. The expected number is zero because neither flush key nor finish have been invoked. So this passes because the number of tables in `got` is zero and it expects zero.

jsternberg:
> Are there any performance implications to this work? I'm thinking of queries where we might have a table stream with lots of little tables, will this add more overhead?

I think there's a possibility this will add more overhead. I am not sure of the performance implications. When I was testing it as part of the timebox, I didn't see anything major in the actual execution. I did see something when loading the trace in jaeger itself but I think that was mostly because there were more spans to render.

> I see you've also had to touch code that creates spans and sets the node ID. Do you have confidence that tracing and/or the useful error messages haven't been lost---we have tests for both of those, correct?

I have pretty good confidence that all previous tracing and error messages are still intact and the only thing is that now more messages get traced. Previously, I believe only process and maybe finish got traced. Now all messages get traced.

jsternberg:
@wolffcm updated the comments in the test.

wolffcm (approved):
Ok, sounds good to me. Thanks for answering my questions.

PR #3926, opened 2021-08-04
fix(stdlib): fix pivot when no data left

fchikwekwe:
There's a pipeline test for this transformation that was failing because of improper registration and a logic error. This PR fixes those.

### DIFF
--- a/stdlib/universe/pivot.go
+++ b/stdlib/universe/pivot.go
@@ -48,7 +48,7 @@ func init() {
 	execute.RegisterTransformation(PivotKind, createPivotTransformation)
 
 	// optimized pivot
-	execute.RegisterTransformation(SortedPivotKind, createPivotTransformation)
+	execute.RegisterTransformation(SortedPivotKind, createSortedPivotTransformation)
 }
 
 func createPivotOpSpec(args flux.Arguments, a *flux.Administration) (flux.OperationSpec, error) {
@@ -420,6 +420,7 @@ type SortedPivotProcedureSpec struct {
 func (s *SortedPivotProcedureSpec) Kind() plan.ProcedureKind {
 	return SortedPivotKind
 }
+
 func (s *SortedPivotProcedureSpec) Copy() plan.ProcedureSpec {
 	ns := new(SortedPivotProcedureSpec)
 	ns.RowKey = make([]string, len(s.RowKey))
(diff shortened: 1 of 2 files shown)


stdlib/universe/pivot.go
```diff
@@ -667,14 +706,16 @@ func (t *sortedPivotTransformation) Finish(id execute.DatasetID, err error) {
...
+	if t.group != nil {
+		var tbl flux.Table
+		gr := t.group
+		tbl, err = gr.doPivot(t.prevOutputKey, t.alloc)
+		if err != nil {
+			return
+		}
+		if err = t.d.Process(tbl); err != nil {
+			return
+		}
```
fchikwekwe, line 718: This logic error was causing my pipeline test to fail.

stdlib/universe/pivot.go
```diff
@@ -430,6 +457,18 @@ func (s *SortedPivotProcedureSpec) Copy() plan.ProcedureSpec {
...
+	if !ok {
+		return nil, nil, errors.Newf(codes.Internal, "invalid spec type %T", spec)
+	}
+	t, dataset, err := newSortedPivotTransformation(a.Context(), *s, id, a.Allocator())
+	if err != nil {
+		return nil, nil, err
+	}
+	return t, dataset, nil
+}
+
```
fchikwekwe, line 471: necessary to register transformation

stdlib/universe/pivot.go
```diff
@@ -536,7 +549,7 @@ func (t *sortedPivotTransformation) Process(id execute.DatasetID, tbl flux.Table
 		colLabel := cr.Key().LabelValue(colKey)
 
 		if colLabel == nil {
-			// The column colLabel is not part of the group colLabel
+			// The column key is not part of the group key
```
fchikwekwe, line 552: typo from find and replace.

stdlib/universe/pivot.go
wolffcm: My understanding is that `t.group` should only be `nil` when there was never any data. Is there any other way for this to happen?

wolffcm (approved):
Looks good, I just had a question.

fchikwekwe:
@wolffcm you're right. I think this was previously failing because of a different logic error and I didnt walk it back. Fixing.

stdlib/universe/pivot_test.go
```diff
@@ -1490,6 +1490,16 @@ func TestSortedPivot_Process(t *testing.T) {
 				},
 			},
 		},
+		{
```
fchikwekwe, line 1493: @wolffcm you asked for an additional test case with no data. Here it is.

stdlib/universe/pivot.go
fchikwekwe: Yeah, I think you may be right about that. Initially this was tied to another failure that I was saying, so this was not really written to prevent cases where there was never any data, but since you mentioned that this is good validation code for that particular case, I'll leave it it.

wolffcm (approved):
Looks good!

PR #3930, opened 2021-08-04
feat: POC of vectorization for optimizing map

wolffcm:


wolffcm:
This concept is now well-proved.

PR #3931, opened 2021-08-04
chore: remove libflux site

rockstar:
This directory seems to be an integration to work out wasm work more
than two years ago. There's no scaffolding currently to generate a wasm
binary that can be used by this site, and it's a common place for
dependabot to keep putting patches up to update dependencies. Rather
than trying to keep dependencies up to date, I think we should just
delete the whole thing.


BarWeiner1 (commented):
Yeah this seems like it was done for a demo like the one we are building now for the docs capabilities. I don't see why we would need this.

wolffcm (approved):
I'm okay with removing this if it's troublesome.

PR #3934, opened 2021-08-05
feat(execute): add transport-aware dataset

### LINKED ISSUE #3887: Add Dataset that is Transport-aware
A new `Dataset` implementation should be created that is `Transport` aware and can interact directly with a transport. At its simplest, it must implement the `execute.Dataset` interface and it must be capable of recognizing when transformations implement `Transport` directly.

It should roughly look like this at a minimum:

```
type Dataset struct {
	id    DatasetID
	ts    []execute.Transport
	mem   memory.Allocator
}

func NewDataset(id DatasetID, mem memory.Allocator) *Dataset {
	return &Dataset{
		id:    id,
		mem:   mem,
	}
}

func (d *Dataset) AddTransformation(t execute.Transformation) {
	if t, ok := t.(execute.Transport); ok {
		d.ts = append(d.ts, t)
		return
	}
	d.ts = append(d.ts, execute.WrapTransformationInTransport(t, d.mem))
}
```

Transport-aware transformations should implement both the `Transformation` interface and the `Transport` interface.

jsternberg:
The transport-aware dataset, or `TransportDataset`, can be used to send
messages directly to downstream transports. The `TransportDataset`
implements the `Dataset` interface and can be used in any place that
requires a Dataset.

If the downstream transformation implements the `Transport` interface,
the messages will be sent directly. If the downstream transformation
only implements the transformation interface, the transformation will be
wrapped in the transport adapter and that will be used instead.

If there are multiple downstream nodes, messages will be automatically
duplicated.

Closes #3887.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -3,9 +3,11 @@ package execute
 import (
 	"context"
 
+	"github.com/apache/arrow/go/arrow/memory"
 	uuid "github.com/gofrs/uuid"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute/table"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/plan"
 )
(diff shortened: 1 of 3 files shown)


execute/dataset.go
```diff
@@ -234,3 +236,97 @@ func (d *PassthroughDataset) Finish(err error) {
...
+
+// TransportDataset holds data for a specific node and sends
+// messages to downstream nodes using the Transport.
+//
+// This Dataset also implements a shim for execute.Dataset
+// so it can be integrated with the existing execution engine.
+// These methods are stubs and do not do anything.
+type TransportDataset struct {
+	id    DatasetID
+	ts    []Transport
```
rockstar, line 248: Could we call this `transports`? I'm kinda starting to thing we should be better about using better names.

execute/dataset_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/arrow"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/mock"
+)
+
+func TestTransportDataset_Process(t *testing.T) {
```
rockstar, line 18: I so so _so_ prefer these tests to the tests that create an array of inputs and outputs and iterates over them in a single function. I could read these like a narrative, which helped me get my head around what was actually being tested without scrolling back and forth.

rockstar (approved):

execute/dataset_test.go
```diff
@@ -0,0 +1,221 @@
...
+		t.Fatal(err)
+	}
+
+	if !processed {
+		t.Fatal("message was not processed")
+	}
+}
+
+func TestTransportDataset_AddTransformation(t *testing.T) {
+	processed := false
```
skartikey, line 81: the variable name `processed` is used as bool and int both in the same test file. so changing bool vars to `isProcessed ` and ints to `totalProcessed` would be more readable.

skartikey (approved):

PR #3935, opened 2021-08-05
feat(feature): add feature flag library as an internal package

### LINKED ISSUE #3661: Add feature flagger library to flux
It is common in Go programs to introduce new functionality or performance improvements. But, it is also a common problem that changes to code can potentially have unintended consequences or bugs that will sometimes reach customers. In order to control the functionality of the code at runtime, some form of feature flag is used. In its most basic form, a feature flag might just be some configuration option in the server configuration or an environment variable that is set on the deployment.

In some circumstances, more fine-controlled or responsive feature flagging is required. This library is intended to provide this more fine-controlled and responsive feature flagging.

InfluxDB has a feature flagger library associated with it. It is integrated with the command line which allows a person to specify temporary features to be enabled for a specific server. Our cloud version implements its own flagger that integrates with a third-party service so we can dynamically enable or disable feature flags for a running binary and for individual options like a specific organization or a specific user.

This library is extraordinarily useful for rolling out things like query optimizations or changes to the engine where we test them, but there are many circumstances where a query might run in production environments and we can't test them all with unit tests. It's necessary to be able to roll out these changes slowly or enable them for a specific customer when we are attempting to optimize a specific query, but then we can optimize it for all users rather than just the one.

That means we need to be able to define feature flags and expose a way for a library that imports flux to define their own way of providing flag values.

The feature flagging library that is used by influxdb is fairly isolated and is a good candidate for either breaking into a separate library or copying and pasting the code into flux. We can expose the `Flagger` interface in the flux dependencies to allow library integrators to use their own flaggers while defining feature flags internally.

The goal of this is to strip out the areas of the influxdb feature flag library that aren't directly related to flux. Expose the interfaces along with a default implementation inside of the `dependencies` package. Then add a way to generate the feature flags and generate those flags to an `internal/feature` package so that only flux code can consume the feature flags.

jsternberg:
The feature flag library has been adapted into an internal package and
exposed as an injectable dependency to the flux library. This allows an
application to embed flux and customize exposed behavior using the
interface.

Closes #3661.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


internal/pkg/feature/feature.go
```diff
@@ -0,0 +1,38 @@
...
+	FlagValue(ctx context.Context, flag Flag) interface{}
+}
+
+// Inject will inject the Flagger into the context.
+func Inject(ctx context.Context, flagger Flagger) context.Context {
+	return context.WithValue(ctx, flaggerKey, flagger)
+}
+
+// GetFlagger returns the Flagger for this context or the default flagger.
+func GetFlagger(ctx context.Context) Flagger {
```
scbrickley, line 23: If I understand it correctly, this is where downstream consumers can inject their context so flux can use the same flagger. Does that mean we can configure flux feature flags the same way we do with idpe? How would we configure things in OSS? Or are we not concerned about that?

scbrickley (approved):
This looks fine to me. I just had a few clarifying questions.

internal/pkg/feature/feature.go
jsternberg: This is where we retrieve the feature flagger so we can access them within Flux. I think this method isn't exposed to external packages.

A downstream consumer would inject their flagger into the context using `Inject()`.

I've been a bit confused over the best way to configure flux feature flags in idpe. My current thought is just to duplicate the feature flag directly into the `flags.yml` file for each of the products. In the future, we can potentially merge the flags because flux exposes its flags in `dependencies/feature` with the `All()` method.

For OSS, feature flags get configured with a command line parameter called `--feature-flags`. They're exposed, but they are server-wide and there's no way to do the kind of detailed flagging that we can do in cloud.

scbrickley: That all makes sense to me. Thanks for clarifying!

internal/feature/flags.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/influxdata/flux/internal/pkg/feature"
+)
+
+type (
+	Flag       = feature.Flag
+	Flagger    = feature.Flagger
+	StringFlag = feature.StringFlag
+	FloatFlag  = feature.FloatFlag
+	IntFlag    = feature.IntFlag
+	BoolFlag   = feature.BoolFlag
```
fchikwekwe, line 17: What are these? Do we have various types that feature flags can be?

fchikwekwe, line 17: What are these? Do we have various types that feature flags can be?

internal/pkg/feature/README.md
```diff
@@ -0,0 +1,53 @@
...
+In some circumstances, more fine-controlled or responsive feature flagging is required.
+This library is intended to provide this more fine-controlled and responsive feature flagging.
+
+## Flags
+
+A `Flag` is a definition of a feature flag.
+It is represented by a human-readable name, a description of the flag, the programmatic key, and a default value.
+It also contains metadata for whether the flag is intended to be exposed to an end-user, the expected lifetime of the flag (temporary or permanent), and a contact for the person, team, or other entity that added the feature flag.
+
+These flags can be represented in YAML form as the following:
```
fchikwekwe, line 18: great description and summary!

internal/pkg/feature/cmd/feature/main.go
```diff
@@ -0,0 +1,275 @@
...
+	"go/format"
+	"io/ioutil"
+	"os"
+	"strings"
+	"text/template"
+
+	"gopkg.in/yaml.v2"
+)
+
+const tmpl = `// Code generated by the feature package; DO NOT EDIT.
```
fchikwekwe, line 16: Why are we templating this code? There seems to be only one instance of this file.

internal/pkg/feature/cmd/feature/main.go
```diff
@@ -0,0 +1,275 @@
...
+	Description string
+	Key         string
+	Default     interface{}
+	Contact     string
+}
+
+func (f flagConfig) Valid() error {
+	var problems []string
+	if f.Key == "" {
+		problems = append(problems, "missing key")
```
fchikwekwe, line 85: lol...definitely a semantic variable name.

internal/pkg/feature/cmd/feature/main.go
```diff
@@ -0,0 +1,275 @@
...
+		switch t.(type) {
+		case bool:
+			return "BoolFlag"
+		case float64:
+			return "FloatFlag"
+		case int:
+			return "IntFlag"
+		default:
+			return "StringFlag"
+		}
```
fchikwekwe, line 258: Yeah, again, I'm a little confused on these types. How do they fit into the feature flag?

internal/pkg/feature/feature_test.go
```diff
@@ -0,0 +1,151 @@
...
+			},
+			expected: int32(42),
+		},
+		{
+			name: "float happy path",
+			flag: newFlag("test", 0.0),
+			values: map[string]interface{}{
+				"test": 42.42,
+			},
+			expected: 42.42,
```
fchikwekwe, line 39: I'm seeing the various values used here, but conceptually I'm not really seeing how to fit into my existing understanding of feature flags. Practically, what do these values represent?

fchikwekwe (commented):
I wanted to get a better understanding of the basics before doing another review pass on this. Sorry for the delay, but thank you in advance for helping me understand.

internal/pkg/feature/cmd/feature/main.go
jsternberg: This is the program that generates the code at `internal/feature/flags.go`. This file itself isn't used within the Flux library. It's invoked by `go generate`.

internal/pkg/feature/cmd/feature/main.go
jsternberg: This code was copied from the influxdb version.

internal/pkg/feature/cmd/feature/main.go
jsternberg: When we template feature flags, there's a default value and that value is inferred to be a specific type. If we see a string in the default field, we generate a `StringFlag` and this generates methods on the flag that allow us to access the feature flag with a method like `myFlag.String(ctx)` which returns the string.

This is the same with the other types and the others have equivalent names like `Int(ctx)` or `Float(ctx)`. I think booleans are the only ones that are different and use `Enabled(ctx)` instead.

This code is also copied over from the influxdb implementation.

jsternberg: To point out, we very rarely, if I've ever been able to find it, use the types other than booleans. So I understand the confusion. It might be worth it for us to remove them if we find that they're causing confusion. At the moment, I think my preference would be to keep things as close to influxdb as possible.

fchikwekwe (approved):

PR #3950, opened 2021-08-10
fix(execute): transformation adapter does not error when receiving a flush key for a table that is not present

jsternberg:
The transformation adapter would respond to the flush key message by
attempting to create a table for the key. If there was no table builder
associated with the key, it would cause an error instead of just
ignoring the message.

This updates the transformation adapter to ignore flush key messages for
keys that don't correspond to an existing table builder.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/transport.go
+++ b/execute/transport.go
@@ -654,8 +654,8 @@ func (t *transformationTransportAdapter) ProcessMessage(m Message) error {
 
 		// Retrieve the buffered builder for the given key
 		// and send the data to the next transformation.
-		tbl, err := t.cache.Table(m.Key())
-		if err != nil {
+		tbl, ok, err := t.cache.Table(m.Key())
+		if err != nil || !ok {
 			return err
 		}
 		t.cache.ExpireTable(m.Key())
--- a/execute/table/builder_cache.go
+++ b/execute/table/builder_cache.go
@@ -4,8 +4,6 @@ import (
 	"reflect"
 
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/codes"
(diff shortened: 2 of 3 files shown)


execute/transport.go
```diff
@@ -654,8 +654,8 @@ func (t *transformationTransportAdapter) ProcessMessage(m Message) error {
 
 		// Retrieve the buffered builder for the given key
 		// and send the data to the next transformation.
-		tbl, err := t.cache.Table(m.Key())
-		if err != nil {
+		tbl, ok, err := t.cache.Table(m.Key())
+		if err != nil || !ok {
```
wolffcm, line 658: Based on the implementation of `ProcessChunk` wouldn't we want to produce an error? Getting a flush key message for something not in the table cache would seem to indicate an error somewhere. Do you have an example where this might be expected?

wolffcm (commented):
This looks pretty straightforward, but I had a question.

execute/transport.go
jsternberg: I found this when I was working on implementing filter. Filter is a narrow transformation so it passes the flush key message. If filter completely filters the table, it won't produce any chunks for a certain key. But, it still passes the flush key message.

It does seem a bit strange to me that this is possible, but in order for filter to not forward this message, it would have to keep state of every group key it had seen and sent to know whether it should forward the flush key or swallow it. I think the cost of making filter do accounting of all of the group keys it has seen would probably be worse than just dropping a message that should probably not have been passed.

wolffcm (approved):

PR #3951, opened 2021-08-10
feat(contrib): add contrib/bonitoo-io/hex package

### LINKED ISSUE #3817: Hex conversion functions
There has been a flurry of requests in the past couple of weeks about users looking for Flux functions that would convert values to and from Hex.

Is this something that can be done with the existing math function library? or are there additional elements required?

sranka:
Closes #3817 , see package README.md for details.

 
### Done checklist
- [x] Test cases written


libflux/flux/src/lib.rs
```diff
@@ -1118,22 +1118,39 @@ from(bucket: v.bucket)
     }
 
     #[test]
-    //CSV package's docs are accessed from their specific index in the vector of DocPackages, ensuring that they are accessible and correct
-    // XXX: rockstar (5 Aug 2021) - This test can be flaky. For example, there was a
-    // failure because another patch was merged that added a new stdlib, which pushed
-    // `csv` from the 19th item to the 20th item.
```
sranka, line 1124: @rockstar thank you for your comments, the tests in this PR cannot pass for the reasons you explained. An introduction of any "contrib/*" package (in this PR) increments a position of `csv`package doc. I fixed the doc tests to 'find by name' herein. I did not find any place where a specific index `20` is required.

sranka: The flaky test were repaired meanwhile in 5b821eef54c6610d7ca1e321ad702bdba64f7495

stdlib/contrib/bonitoo-io/hex/hex.flux
```diff
@@ -0,0 +1,10 @@
+package hex
+
+builtin int : (v: A) => int
+builtin string : (v: A) => string
+builtin uint : (v: A) => uint
+builtin bytes : (v: A) => bytes
```
nathanielc, line 6: Can the value for `v` be any type? or does it need to be a string? If it needs to be a string can you change these types to `builtin int : (v: stirng) => int`?

stdlib/contrib/bonitoo-io/hex/hex.go
```diff
@@ -0,0 +1,185 @@
...
+		case semantic.Bool:
+			if v.Bool() {
+				i = 1
+			} else {
+				i = 0
+			}
+		case semantic.Time:
+			i = uint64(v.Time())
+		case semantic.Duration:
+			i = uint64(v.Duration().Duration())
```
nathanielc, line 156: What is the use case for the other cases besides string?

If I read the code `hex.int(v: 5)` I am not sure what I would expect as an answer. I think it is confusing that the functions only convert from hex if the input is a string. 

If we removed all the cases except for the string case does that break the use case?

nathanielc (commented):
The code and tests look good. I have some questions about the behavior of the functions when the arguments are not strings. See below.

stdlib/contrib/bonitoo-io/hex/hex.flux
sranka: Yes it can, it behaves the same way as standard `int()`, but a hex string value is expected.

stdlib/contrib/bonitoo-io/hex/hex.go
sranka: It just works on other types ... as the built-in` int()` fn works.

sranka: @nathanielc if you think that I should better restrict the type of `v` param to string, I will. I guess that (from a developer POV) I can't ensure that it is a string, especially when used with a table input (`hex.toInt`,`hex.toUInt`). It then provides a reasonable default behaviour rather than a strict failure.

stdlib/contrib/bonitoo-io/hex/hex.flux
```diff
@@ -0,0 +1,10 @@
+package hex
+
+builtin int : (v: A) => int
```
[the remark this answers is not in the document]
sranka: OK, I changed the flux functions that convert from hexadecimal string to accept only a string type. They report an error if a string value is not on input, which is also controlled by unit tests.

stdlib/contrib/bonitoo-io/hex/hex.go
```diff
@@ -0,0 +1,153 @@
...
+var (
+	convIntType    = runtime.MustLookupBuiltinType(pkgName, "int")
+	convUintType   = runtime.MustLookupBuiltinType(pkgName, "uint")
+	convStringType = runtime.MustLookupBuiltinType(pkgName, "string")
+	convBytesType  = runtime.MustLookupBuiltinType(pkgName, "bytes")
+)
+
+var errMissingArg = errors.Newf(codes.Invalid, "missing argument %q", conversionArg)
+
+var stringConv = values.NewFunction(
```
jsternberg, line 37: A pattern that I like to do when it comes to functions is this.

```
func init() {
    runtime.RegisterPackageValue(pkgName, "string",
        values.NewFunction(
            "string",
            runtime.MustLookupBuiltinType(pkgName, "string"),
            func(ctx context.Context, args values.Object) (values.Value, error) {
                return interpreter.DoFunctionCallContext(String, ctx, args)
            },
        ),
    )
}

func String(ctx context.Context, args interpreter.Arguments) (values.Value, error) {
    // implement as normal
}
```

stdlib/contrib/bonitoo-io/hex/hex.flux
```diff
@@ -0,0 +1,10 @@
+package hex
+
+builtin int : (v: string) => int
+builtin string : (v: A) => string
+builtin uint : (v: string) => uint
+builtin bytes : (v: string) => bytes
+
+toString = (tables=<-) => tables |> map(fn: (r) => ({r with _value: string(v: r._value)}))
```
jsternberg, line 8: I'd like to remove these functions. They might be useful but I'd prefer not to add them until we see this pattern as something that is necessary.

jsternberg (changes requested):
A few changes mostly in terms of formatting. Looks good otherwise.

stdlib/contrib/bonitoo-io/hex/hex.flux
sranka: Since the flux data originate in InfluxDB per https://github.com/influxdata/feature-requests/issues/297, I supposed that it is a good idea to have short-hand functions that work over a table OOTB, I was mimicking the pattern of the built-in functions. They are now removed per your request.

stdlib/contrib/bonitoo-io/hex/hex.go
sranka: I have chosen a wrong existing file to copy the pattern from. It is better your way.

timhallinflux:
@jsternberg bump...

jsternberg (approved):
Leaving a few comments but no action is required. I'll merge this.

nathanielc (approved):

PR #3953, opened 2021-08-10
feat(stdlib): update filter to use narrow transformation

### LINKED ISSUE #3936: Update filter transformation to use NarrowTransformation
Update the `filter()` transformation to use [NarrowTransformation](https://github.com/influxdata/flux/issues/3927).

The `filter()` transformation presently uses `table.Stream()` to stream tables to the downstream. The transformation should be adapted to use `NarrowTransformation` instead. This should involve the following steps:

1. Remove `table.Stream()`.
2. Prepare the function for each table chunk instead of for the entire table.
3. Filter each incoming table using the same filter algorithm.
4. Send the table chunk using the `TransportDataset`.

jsternberg:
This updates filter to use the `NarrowTransformation` transformation
interface instead of the stream table.

Closes #3936.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/filter.go
+++ b/stdlib/universe/filter.go
@@ -16,6 +16,7 @@ import (
 	"github.com/influxdata/flux/internal/arrowutil"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/internal/execute/table"
+	"github.com/influxdata/flux/internal/feature"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
@@ -157,6 +158,13 @@ func NewFilterTransformation(ctx context.Context, spec *FilterProcedureSpec, id 
 		keepEmptyTables: spec.KeepEmptyTables,
 		alloc:           alloc,
 	}
+
+	if feature.NarrowTransformationFilter().Enabled(ctx) {
+		t := &filterTransformationAdapter{
+			filterTransformation: t,
+		}
+		return execute.NewNarrowTransformation(id, t, alloc)
+	}
 	return t, t.d, nil
 }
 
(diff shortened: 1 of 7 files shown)


jsternberg:
I'm leaving this up as a draft, but I am presently leaning towards merging and integrating the [feature flag](https://github.com/influxdata/flux/issues/3661) work before merging this and putting this behind a feature flag. I don't presently believe this change will cause any problems and I think the test cases are good enough, but we might want to exercise caution anyway. This is especially because this will be the first transformation that directly implements the new `Transport` interface.

stdlib/universe/filter.go
```diff
@@ -140,172 +140,179 @@ func (s *FilterProcedureSpec) PlanDetails() string {
...
+// 	}
+// 	return d.Process(chunk)
+// }
+
+func (t *filterTransformation) filterChunk(fn *execute.RowPredicatePreparedFn, chunk table.Chunk, record values.Object, indices []int, mem arrowmem.Allocator) (table.Chunk, bool, error) {
+	buffer := chunk.Buffer()
+	bitset, err := t.filter(fn, &buffer, record, indices, mem)
 	if err != nil {
-		panic(err)
+		return table.Chunk{}, false, err
```
BarWeiner1, line 285: Are we giving the user the ability to see the error message anywhere?

BarWeiner1 (approved):
If this passes the existing tests and doesn't break transport than the code that isn't commented out seems fine to me. Not really sure about waiting for the feature flag issue to merge. Probably best to get others on the team to comment as well.

scbrickley (commented):
So the purpose of this work is to have `filter` operate on chunks instead of whole tables, correct? Does that allow us to operate on multiple chunks in parallel? Just trying to get some understanding of the big picture here.

It looks like what we're doing is taking an existing `filterTransformation` and, if the flag is enabled, wrap it in a `filterTransformationAdapter` which is just a pointer to the original transformation, but that has an alternate `Process` method defined on it. That way, when the flux engine gets this transformation passed to it, it knows to use the narrow method. Do I have that about right? Are there details that I'm missing?

jsternberg:
> So the purpose of this work is to have filter operate on chunks instead of whole tables, correct? Does that allow us to operate on multiple chunks in parallel? Just trying to get some understanding of the big picture here.

This is not the case. Individual nodes in the graph, like the filter transformation node, still can only process on chunk at a time. This just allows downstream nodes in the pipeline to process the chunks sooner rather than waiting for the entire table to be processed by filter.

> It looks like what we're doing is taking an existing filterTransformation and, if the flag is enabled, wrap it in a filterTransformationAdapter which is just a pointer to the original transformation, but that has an alternate Process method defined on it. That way, when the flux engine gets this transformation passed to it, it knows to use the narrow method. Do I have that about right? Are there details that I'm missing?

This is correct. It's set up so that the two share code so we can remove the old one after testing the new one. The new one implements the `NarrowTransformation` interface and uses the transport implementation for that. When we remove the old one, we'll delete the old process method, merge the original and adapter to just have one `filterTransformation`, and remove any unused attributes and methods.

scbrickley (approved):

Issue #3521, opened 2021-03-01
`fill()` function often fails when column is all nulls


philomory:
The `fill()` function exists to replace null values with a given value; however, depending on your source data and the way stuff is grouped, in some tables you may end up with the column being entirely absent, in which case `fill()` throws an error. In order to get reliable behavior, you end up having to use `map(fn: (r) => r with my_column: if exists r.my_column then r.my_column else "foo"}))`, which is _logically_ the same as `fill(column: "my_column", value: "foo")`, but works even if `my_column` happens to be absent entirely for a given table.

As an example of a real query where this becomes an issue, here is a query I wrote as part of a check that would look for (and alert on) task failures:

```flux
from(bucket: "_tasks")
  |> range(start: -10m)
  |> schema.fieldsAsCols()
  |> group(columns: ["taskID"])
  |> fill(column: "errorMessage", value: "(none)")
  |> monitor.check(
    crit: (r) => r.status == "failed",
    messageFn: (r) => (
      if r.status == "failed" then "Task '${r.name}' failed! Error message: ${r.errorMessage}"
      else "Task '${r.name}' succeeded."
    ),
    data: {...},
  )
```

This looks perfectly fine at first glance, but when the check is executed, if any of the tasks had no errors in the past ten minutes, then the table for that task will have no `errorMessage` column, and then `fill()` will fail with `fill column not found: errorMessage`.

nathanielc:
Thanks for this issue, makes a lot of sense for fill to work even if the column doesn't exist.

PR #3955, opened 2021-08-11
fix(stdlib): fill function fails when col to fill is entirely absent

skartikey:
This PR will add a new null column if the fill column is entirely absent and then populate the fill value into it
The test cases have been added for each type

fixes: #3521

### DIFF
--- a/stdlib/universe/fill.go
+++ b/stdlib/universe/fill.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/internal/errors"
@@ -202,10 +203,6 @@ func (t *fillTransformation) RetractTable(id execute.DatasetID, key flux.GroupKe
 
 func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
 	colIdx := execute.ColIdx(t.spec.Column, tbl.Cols())
-	if colIdx < 0 {
-		return errors.Newf(codes.FailedPrecondition, "fill column not found: %s", t.spec.Column)
-	}
-
 	key := tbl.Key()
 	if idx := execute.ColIdx(t.spec.Column, key.Cols()); idx >= 0 {
 		if key.IsNull(idx) {
@@ -223,13 +220,26 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 
 	var fillValue interface{}
 	if !t.spec.UsePrevious {
-		if tbl.Cols()[colIdx].Type != flux.ColumnType(t.spec.Value.Type()) {
+		if colIdx > -1 && tbl.Cols()[colIdx].Type != flux.ColumnType(t.spec.Value.Type()) {
 			return errors.Newf(codes.FailedPrecondition, "fill column type mismatch: %s/%s", tbl.Cols()[colIdx].Type.String(), flux.ColumnType(t.spec.Value.Type()).String())
 		}
 		fillValue = values.Unwrap(t.spec.Value)
 	}
 
-	table, err := table.StreamWithContext(t.ctx, key, tbl.Cols(), func(ctx context.Context, w *table.StreamWriter) error {
+	// In case of missing fill column, add it to the existing columns
+	tableCols := tbl.Cols()
+	if colIdx < 0 {
+		newCols := make([]flux.ColMeta, len(tableCols), len(tableCols)+1)
+		copy(newCols, tableCols)
+		c := flux.ColMeta{
+			Label: t.spec.Column,
+			Type:  flux.ColumnType(t.spec.Value.Type()),
+		}
+		tableCols = append(newCols, c)
+		colIdx = len(tableCols) - 1
+	}
+
+	table, err := table.StreamWithContext(t.ctx, key, tableCols, func(ctx context.Context, w *table.StreamWriter) error {
 		return tbl.Do(func(cr flux.ColReader) error {
 			return t.fillTable(w, cr, colIdx, &fillValue)
 		})
@@ -251,11 +261,14 @@ func (t *fillTransformation) Finish(id execute.DatasetID, err error) {
 }
 
 func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader, colIdx int, fillValue *interface{}) error {
-	if cr.Len() == 0 {
+	crLen := cr.Len()
+	if crLen == 0 {
 		return nil
 	}
 	vs := make([]array.Interface, len(w.Cols()))
-	for i, col := range w.Cols() {
+
+	// Iterate over the existing columns and if column already exist(colIdx matches with i) call fillColumn on it
+	for i, col := range cr.Cols() {
 		arr := table.Values(cr, i)
 		if i != colIdx {
 			vs[i] = arr
@@ -264,5 +277,22 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
 		}
 		vs[i] = t.fillColumn(col.Type, arr, fillValue)
 	}
+
+	// If the fill column is new, create a completely null column and call fillColumn on it
+	if vs[colIdx] == nil {
+		colType := flux.ColumnType(t.spec.Value.Type())
+		arr := t.addNullColumn(colType, crLen)
+		defer arr.Release()
+		vs[colIdx] = t.fillColumn(colType, arr, fillValue)
+	}
 	return w.Write(vs)
+}
+
+func (t *fillTransformation) addNullColumn(typ flux.ColType, l int) array.Interface {
+	builder := arrow.NewBuilder(typ, t.alloc)
+	builder.Resize(l)
+	for i := 0; i < l; i++ {
+		builder.AppendNull()
+	}
+	return builder.NewArray()
 }
--- a/stdlib/universe/fill_test.go
+++ b/stdlib/universe/fill_test.go
@@ -192,6 +192,36 @@ func TestFill_Process(t *testing.T) {
 			}},
 		},
 		{
+			name: "missing bool fill col",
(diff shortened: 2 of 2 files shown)


stdlib/universe/fill.go
```diff
@@ -223,13 +220,22 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
 	}
 
-	table, err := table.StreamWithContext(t.ctx, key, tbl.Cols(), func(ctx context.Context, w *table.StreamWriter) error {
+	tableCols := tbl.Cols()
+	if colIdx < 0 {
+		c := flux.ColMeta{
+			Label: t.spec.Column,
+			Type:  flux.ColumnType(t.spec.Value.Type()),
+		}
+		tableCols = append(tableCols, c)
```
jsternberg, line 235: This can potentially cause a race condition. Imagine a situation where the original columns is defined in this way:

```
cols := make([]flux.ColMeta, 0, 1)
cols = append(cols, flux.ColMeta{Label: "mycol", Type: flux.TString})
```

If we call `append()` later on these columns, the `append()` function will notice that the capacity would be exceeded and it will allocate a new slice with the appropriate capacity. This works perfectly and is the expected pathway for this code.

On the other hand, if we did this:

```
cols := make([]flux.ColMeta, 0, 2)
cols = append(cols, flux.ColMeta{Label: "mycol", Type: flux.TString})
```

This would also look like it works fine at the allocations site. The length is 1 and the capacity is 2. We overallocated but there's no problem with the slice itself.

When we retrieve the slice here and use `append()`, we put the new column inside of the additional section of capacity. If we did this in two different locations, then we would be modifying the same buffer and it would open ourselves up to a race condition.

Instead, we should do this:

```
newCols := make([]flux.ColMeta, len(tableCols), len(tableCols)+1)
copy(newCols, tableCols)
tableCols = append(newCols, c)
```

stdlib/universe/fill.go
```diff
@@ -255,14 +261,32 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
...
 			vs[i].Retain()
-			continue
 		}
-		vs[i] = t.fillColumn(col.Type, arr, fillValue)
 	}
 	return w.Write(vs)
 }
+
+func (t *fillTransformation) addNullColumn(typ flux.ColType, l int) array.Interface {
+	builder := arrow.NewBuilder(typ, memory.DefaultAllocator)
```
jsternberg, line 286: Can you use the allocator associated with the fill transformation? It's probably `t.alloc`.

stdlib/universe/fill.go
```diff
@@ -255,14 +261,32 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
 		return nil
 	}
 	vs := make([]array.Interface, len(w.Cols()))
+	var colArr array.Interface
```
jsternberg, line 264: Can you move this into the for loop? I don't like having variable references outside of the scope that they are used in. I like to try and keep the variable initialization as close and into as narrow of a scope as where it is used to avoid dangling references.

stdlib/universe/fill.go
```diff
@@ -255,14 +261,32 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
...
 	for i, col := range w.Cols() {
-		arr := table.Values(cr, i)
-		if i != colIdx {
-			vs[i] = arr
+
+		isMissingCol := colIdx < 0 && col.Label == t.spec.Column
```
jsternberg, line 267: Can you add comments to this area? The logic here is confusing me and I'm curious if we can simplify it.

jsternberg (changes requested):

stdlib/universe/fill.go
```diff
@@ -255,14 +264,33 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
...
+
 	for i, col := range w.Cols() {
-		arr := table.Values(cr, i)
-		if i != colIdx {
-			vs[i] = arr
+		var colArr array.Interface
+		// if fillCol is missing AND current col from range is same as fillCol
+		isMissingColCase := colIdx < 0 && col.Label == t.spec.Column
+		if isMissingColCase {
+			colArr = t.addNullColumn(flux.ColumnType(t.spec.Value.Type()), cr.Len())
```
jsternberg, line 273: This will end up leaking memory because it doesn't release the created column. I think you would be better of refactoring this section because I don't think trying to get it into this loop is working properly. Maybe something like:

```
// Iterate over the existing columns.
for i, col := range cr.Cols() {
    arr := table.Values(cr, i)
    if i != colIdx {
        vs[i] = arr
        vs[i].Retain()
        continue
    }
    vs[i] = t.fillColumn(col.Type, arr, fillValue)
}

// If the fill column is new, create a completely null column and use fill column on it.
if len(w.Cols()) > len(cr.Cols()) {
    typ := flux.ColumnType(t.spec.Value.Type())
    if type == flux.TInvalid {
        // return error indicating the value type is invalid? not sure if this is needed.
    }
    arr := t.addNullColumn(typ, cr.Len())
    defer arr.Release()
    vs[colIdx] = t.fillColumn(typ, arr, fillValue)
}
```

If you do this, you should initialize `colIdx` to be the index of the newly created column.

I think this method simplifies the code. Instead of checking on each iteration if you have the special case, you keep the existing code workflow and fix the new column through an if statement after the initial iteration. The initial iteration has been changed to only iterate over existing columns rather than requiring it to check if the input column is missing. So you only check for the missing column once instead of on each loop.

jsternberg (changes requested):

stdlib/universe/fill.go
```diff
@@ -223,13 +220,25 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
 
-	table, err := table.StreamWithContext(t.ctx, key, tbl.Cols(), func(ctx context.Context, w *table.StreamWriter) error {
+	// In case of missing fill column, add it to the existing columns
+	tableCols := tbl.Cols()
+	if colIdx < 0 {
+		newCols := make([]flux.ColMeta, len(tableCols), len(tableCols)+1)
+		copy(newCols, tableCols)
+		c := flux.ColMeta{
+			Label: t.spec.Column,
+			Type:  flux.ColumnType(t.spec.Value.Type()),
```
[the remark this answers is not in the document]
skartikey: spec.Value type is already getting checked in the init function here - https://github.com/influxdata/flux/blob/master/stdlib/universe/fill.go#L57
So if the type is Invalid error will be thrown during the init() itself.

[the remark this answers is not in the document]
jsternberg: Perfect. That's good enough. You can discard this comment then.

stdlib/universe/fill.go
skartikey: This looks cleaner and easy to read. Thanks for the suggestion, I will update the PR accordingly.

stdlib/universe/fill.go
```diff
@@ -223,13 +220,25 @@ func (t *fillTransformation) Process(id execute.DatasetID, tbl flux.Table) error
...
+		}
+		tableCols = append(newCols, c)
```
jsternberg, line 238: Right after this line, I would recommend adding this. See my next comment for why I think this ends up being more clear.

```
colIdx = len(tableCols) - 1
```

stdlib/universe/fill.go
```diff
@@ -264,5 +278,23 @@ func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader,
 		}
 		vs[i] = t.fillColumn(col.Type, arr, fillValue)
 	}
+
+	// If the fill column is new, create a completely null column and call fillColoumn on it
+	if wColsLen > crColsLen {
```
jsternberg, line 283: With the change above, instead of this condition, you can now do:

```
if vs[colIdx] == nil {
    // fill in a bunch of nil values
}
```

I think this is more straightforward than checking the number of columns and comparing them.

stdlib/universe/fill.go
```diff
@@ -251,11 +260,16 @@ func (t *fillTransformation) Finish(id execute.DatasetID, err error) {
 }
 
 func (t *fillTransformation) fillTable(w *table.StreamWriter, cr flux.ColReader, colIdx int, fillValue *interface{}) error {
-	if cr.Len() == 0 {
+	crColsLen := len(cr.Cols())
+	wColsLen := len(w.Cols())
+	crLen := cr.Len()
+	if crLen == 0 || crColsLen == 0 || wColsLen == 0 {
```
jsternberg, line 266: I'm not sure why this condition changed. Is there a test case that exercised this?

jsternberg (changes requested):

stdlib/universe/fill.go
```diff
@@ -251,11 +260,16 @@ func (t *fillTransformation) Finish(id execute.DatasetID, err error) {
...
 		return nil
 	}
-	vs := make([]array.Interface, len(w.Cols()))
-	for i, col := range w.Cols() {
+	vs := make([]array.Interface, wColsLen)
+
+	// Iterate over the existing columns and if coloumn already exist(colIdx matches with i) call fillColoumn on it
```
jsternberg, line 271: Typo in this comment. `coloumn` -> `column`.

stdlib/universe/fill.go
skartikey: This is just for an extra safety check, if for any reason read or write cols length is 0, code should return immediately.

jsternberg: I think this will end up being confusing. It's not possible for either of these to be zero so there's no way to test it. Can you revert this condition back? If they were zero, I believe the below code would still work anyway.

skartikey: one of the checks was already there. Should I remove that as well ?
https://github.com/influxdata/flux/blob/1a84fac74554ebb7bd1674c8671ec96049de0055/stdlib/universe/fill.go#L254

jsternberg: No that check is possible and important. It's possible for a table to be empty. It's just not possible for it to have no columns.

jsternberg (approved):

PR #3956, opened 2021-08-12
docs(execute): add documentation for the execution engine and creating transformations

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/README.md
+++ b/execute/README.md
@@ -1,0 +1,464 @@
+# Flux Engine Developer Guide
+
+This guide covers the Flux Engine design and related concepts.
+The engine includes:
+
+* Execution engine
+* Builtin sources and transformations
+* Memory layout and resource management
+
+## Concepts
+
+The following concepts are important to understanding the Flux engine and the operations that can be executed within the engine.
+This guide will reference these concepts with the assumption that the reader has reviewed the following section.
+
+### Tables and Table Chunks
+
+In Flux, a **Table** is a set of ordered rows grouped together by a common **group key**.
+The **group key** is a set of key/value pairs.
+The elements of the group key appear as columns in the table with the element key appearing as the column name and the element value appearing as the value, constant across all rows.
+In practical terms, this means that for any columns that are in the group key, the values in that table will all be the same.
+
+The table will also contain columns that are not referenced by the group key.
+These columns are free to be any value as long as they are all the same type.
+
+For a table, the following invariants should hold:
+
+* All values inside a column that is part of the group key should have the same value.
+* That value should match the group key's value.
+* The values in a column should all be of the same type.
+
+The last of these invariants is **loosely** held.
+This means that while they should be true, individual transformations may not make an effort to enforce them.
+That means that if a transformation would have its performance affected by enforcing the invariant and the invariant does not affect the transformation's correctness, the transformation may not enforce it when producing the output.
+For an example, the `group()` transformation does not enforce the last invariant and relies on downstream transformations to enforce it.
+
+The set of rows contained within a table are processed in chunks.
+A table chunk is a subset of the rows within a table that are in-memory.
+
+The rows within a table are **ordered**.
+When table chunks are sent, the rows within the first are all ordered before the rows in the second.
+Transformations may choose to change the order, or they may choose to have the order be meaningful.
+For example, the `derivative()` function gives meaning to the order of the input rows while `sort()` will rearrange the order.
+
+A table chunk is composed of a list of arrow arrays, each array corresponding to a column of the table.
+
+### Arrow Arrays
+
+[Apache Arrow](https://arrow.apache.org/) is a language-independent columnar format.
+The flux engine utilizes the Arrow library to represent the columnar data inside of table chunks.
+
+### Execution Pipeline
+
+The **execution pipeline** is the set of nodes passing data from a source to the result.
+The execution pipeline is composed of **nodes**.
+A node with no inputs is a **source** which produces data for nodes that are after, or downstream, of itself.
+A node that takes one or more inputs and produces an output is a **transformation**.
+The final node in the pipeline is the **result** which holds the results of the pipeline.
+
+You can think of the execution pipeline in flux code.
+
+    A |> B |> C
+
+This would translate into three nodes where `A` is a source and `B` and `C` are transformations.
+Flux code does not correspond directly with one function being one node.
+Some functions will be combined into a single node while other functions may get rearranged into other nodes or split into multiple nodes.
+It is the responsibility of the planner to convert flux code into an execution pipeline.
+
+### Dataset
+
+A dataset is used to hold node state and manage sending messages to downstream nodes.
+Each transformation is associated with a single dataset.
+
+When thinking about the execution pipeline, the dataset can be thought of as a distinct part of the node.
+
+    A |> B |> B(d) |> C |> C(d)
+
+### Message
+
+A message is a signal that contains associated data that is being sent from one dataset to the downstream nodes.
+Each message has a lifetime where it is created by the sender and then acknowledged by the receiver.
+A message may hold onto memory and will release its reference to that memory after it has been acknowledged.
+The data contained within a message may be retained and used in other ways.
+
+The following messages presently exist:
+
+|Name|Data|Description
+|----|----|-----------
+|ProcessChunk|table.Chunk|A table chunk is ready to be processed
+|FlushKey|group key|Data associated with the given group key should be flushed from the dataset
+|Finish|error (optional)|The upstream dataset will produce no more messages
+
+The following messages exist, but are deprecated and should not be used by future transformations.
+
+|Name|Data|Description
+|----|----|-----------
+|Process|flux.Table|A full table is ready to be processed
+|UpdateWatermark|time|Data older than the given time will not be sent
+|UpdateProcessingTime|time|Marks the present time
+|RetractTable|key|Data associated with the given group key should be retracted
+
+### Memory Types
+
+Data memory references memory used for storing user data.
+This is in contrast to process or program memory.
+Since flux is integrated in co-tenant environments, it needs to handle arbitrary user data and avoid crashing the system or causing a noisy neighbor performance problem.
+Flux does this by separating the concept of process memory and data memory.
+
+Process memory is any memory required to execute code.
+Process memory is anything in Go that would allocate stack or heap memory and is tracked by the garbage collector.
+User code, by necessity, will use process memory.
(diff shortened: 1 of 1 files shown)


wolffcm (approved):
This looks great. Thanks for writing it up. Please do solicit input from the rest of the team--I think this will prove to be very useful.

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+## Concepts
+
+The following concepts are important to understanding the Flux engine and the operations that can be executed within the engine.
+This guide will reference these concepts with the assumption that the reader knows what is contained in this section.
+
+### Tables and Table Chunks
+
+In Flux, a **Table** is a set of ordered rows grouped together by a common **group key**.
+The **group key** is a set of key/value pairs where the value in the column with the same key name matches the value.
+In practical terms, this means that for any columns that are in the group key, the values in that table will all be the same.
```
adrian-thurston, line 19: Suggestion: "The group key is a set of key/value pairs. The elements of the group key appear as columns in the table, with the element key appearing as the column name and the element value appearing as the value, constant across all rows.

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+The table will also contain columns that are not referenced by the group key.
+These columns are free to be any value as long as they are all the same type.
+
+For a table, the following invariants should hold:
+
+* All values inside a column that is part of the group key should have the same value.
+* That value should match the group key's value.
+* A column with a name should have one and only one type.
```
adrian-thurston, line 28: Suggestion: "The values in a column should all be of the same type."

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+The set of rows contained within a table are processed in chunks.
+A table chunk is a subset of the rows within a table that are in-memory.
+
+The rows within a table are **ordered**.
+When table chunks are sent, the rows within the first are all ordered before the rows in the second.
+Transformations may choose to change the order, or they may choose to have the order be meaningful.
+For example, the `derivative()` function gives meaning to the order of the input rows while `sort()` will rearrange the order.
+
+A table chunk is composed of an arrow array for each column.
```
adrian-thurston, line 43: Suggestion: "A table chunk is composed of a list of Arrow arrays, each array corresponding to a column of the table.

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+Preallocate memory by determining the size of a chunk in advance and using `Resize()` to set the capacity of the builder.
+For strings, it is also helpful to preallocate memory for the data using `ReserveData` if this can be easily known.
+String appends are usually the biggest performance sink for efficiency.
+
+### Limit Chunk Sizes
+
+Limit chunk sizes to [table.BufferSize](https://pkg.go.dev/github.com/influxdata/flux/execute/table#pkg-constants).
+The array values in a column are contained in contiguous data.
+When the chunk size gets larger, the memory allocator has to find a spot in memory that fits that large size.
+Larger chunk sizes are generally better for performance, but too large of buffer sizes put too much pressure on the memory allocator and garbage collector.
```
fchikwekwe, line 370: nit/ Maybe: `...but buffer sizes that are too large put too much pressure on the ...`

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+The `table.BufferSize` constant is an agreed upon size to balance these concerns.
+At the moment, this value is the same as the buffer size that comes from the influxdb storage engine.
+If we find another buffer size works better in the future, we can change this one constant.
```
fchikwekwe, line 373: nit/ Maybe `If we find that another buffer size...`

fchikwekwe (approved):
A few editor's notes, but otherwise the guide so far is well put together and understandable. Thanks for putting the work into this.

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+See the **Mutable Data** section for how to handle these circumstances.
+
+## Execution Engine
+
+The execution engine is central to executing flux queries.
+
+### Pipeline Creation
+
+Before the execution engine begins, it's important to understand the steps that happen before execution.
+These steps will not be described in detail here.
```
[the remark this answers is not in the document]
jsternberg: This is honestly a lot better than the current text. Makes the text come across as more helpful rather than the "good luck figure it out yourself sucker" message that the current text accidentally sounds like. I only really realized that when I saw this comment.

execute/README.md
```diff
@@ -0,0 +1,456 @@
...
+It would then finish by reallocating the arrays to the proper size.
+A faster method is to use a bitset to keep track of which rows are filtered and which ones remain.
+We can allocate a bitset that holds the rows in the current table chunk, run the filter on each row, and record whether it passes the filter or not.
+We can then use that bitset to construct each column independently of each other.
+
+### Utilize Slices
+
+Arrow arrays are immutable and have built in support for slicing data.
+A slice keeps the same reference to the underlying data, but limits the view of that data to a smaller section.
+Slices are most useful when the required data is already in order and only a section is needed.
```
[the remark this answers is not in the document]
jsternberg: I'm not sure what you mean. I had a really hard time phrasing this so I could use some help making it more clear. What I'm trying to say here is that slices can be useful when the data naturally slices itself. Something like `limit()` might be a good example where you just need the data and you don't pay much of a cost. You don't release the unused memory as quickly, but sometimes that extra memory doesn't matter in the short term.

I just wanted to caution someone from using slices to get individual rows. Using slices for individual rows usually ends up making stuff slower.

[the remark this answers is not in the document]
fchikwekwe: I would totally say something akin to what you've just said here then. You could even use the limit example to briefly show what you mean and leave out the word `section` entirely. Just a suggestion though. This might be more clear to other folks who have a look.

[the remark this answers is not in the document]
jsternberg: Ok I updated this section.

scbrickley (approved):
Found this very helpful and informative overall. Just had a couple comments.

execute/README.md
```diff
@@ -0,0 +1,458 @@
...
+
+    func (t *MyNarrowTransformation) process(chunk table.Chunk, state *myState, mem memory.Allocator) (table.Chunk, error) {
+        /* transformation-specific logic */
+    }
+
+Some examples where this version of the narrow transformation is used: [derivative()](https://github.com/influxdata/flux/blob/master/stdlib/universe/derivative.go).
+
+### Group Transformation
+
+TODO: Write this section.
```
rockstar, line 278: Are there issues open for this (and the following) `TODO`? There's little more awkward than seeing a doc years old that has `TODO: document this part`. *glares at many of those online rust books*

rockstar (approved):
Sure wish I had this doc when I was onboarding. I'm excited to have something that I'll have to keep referring back to.

execute/README.md
jsternberg: Not at the moment. I figured I would just circle back and get these once the related issues had been done.

execute/README.md
```diff
@@ -0,0 +1,458 @@
...
+
+This would translate into three nodes where `A` is a source and `B` and `C` are transformations.
+Flux code does not correspond directly with one function being one node.
+Some functions will be combined into a single node while other functions may get rearranged into other nodes or split into multiple nodes.
+It is the responsibility of the planner to convert flux code into an execution pipeline.
+
+### Dataset
+
+A dataset is a connector between transformations.
+Each transformation is associated with a single dataset.
```
adrian-thurston, line 71: If it's a connector between transformations, wouldn't it be associated with two datasets? Input and output?

execute/README.md
```diff
@@ -0,0 +1,458 @@
...
+That dataset keeps the state for that transformation, subsequently passing data to downstream transformations.
+
+When thinking about the execution pipeline, the dataset can be thought of as a distinct part of the node.
+
+    A |> B |> B(d) |> C |> C(d)
+
+### Message
+
+A message is a signal that contains associated data that is being sent from one dataset to the downstream nodes.
+Each message has a lifetime where it is created by the sender and then acknowledged by the receiver.
```
adrian-thurston, line 81: Suggestion: acknowledged *and destroyed* by the receiver.

execute/README.md
```diff
@@ -0,0 +1,458 @@
...
+### Pipeline Creation
+
+Before the execution engine begins, it's important to understand the steps that happen before execution.
+This is only a brief outline of those steps.
+
+* Text is parsed into an AST.
+* AST is converted into semantic graph.
+* Type inference runs on the semantic graph and assigns types to all nodes.
+* The interpreter uses a recursive-descent execution to evaluate the semantic nodes.
+* Side effects are gathered from the interpreter execution and side effects that are linked to table objects are collected into a spec.
```
adrian-thurston, line 139: Might be good to define side-effects somewhere. This was always a point I was unclear on when learning flux. The side effects are actually what manifest as data destinations, be it `to()` or output downstream to the user. So the process of determining what the side effects are is actually determining the outputs of the entire pipeline. To me, It is different from the notion of side effects in a functional programming language, which is modification of some global state. So it's not enough to just say they are "side effects".

execute/README.md
jsternberg: Hm, maybe there's a better way to phrase this. The dataset is a connector, but it's more strongly associated with one transformations. It's responsible for storing data and keeping track of where to send messages. I'll try to rewrite this part.

adrian-thurston (approved):
Finished commenting.

skartikey (approved):
Very useful! It gave me a broader view of the Flux Engine.
Just need clarification at 1 place. I have added my comments there.

execute/README.md
```diff
@@ -0,0 +1,457 @@
...
+### Pipeline Creation
+
+Before the execution engine begins, it's important to understand the steps that happen before execution.
+This is only a brief outline of those steps.
+
+* Text is parsed into an AST.
+* AST is converted into semantic graph.
+* Type inference runs on the semantic graph and assigns types to all nodes.
+* The interpreter uses a recursive-descent execution to evaluate the semantic nodes.
+* Side effects, actions that affect state outside the query engine, are gathered from the interpreter execution and side effects that are linked to table objects are collected into a spec.
```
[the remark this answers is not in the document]
jsternberg: They are not. The side effects are metadata attached to a builtin function. Pure flux functions can only have side effects if they inherit it from a pure Go function. Side effects are things like `yield()` and `to()`. `yield()`'s side effect is that it produces output while `to()`'s side effect is that it changes outside state by writing data to the bucket.

I'll update the concepts section to include side effects so this part isn't so confusing.

PR #3960, opened 2021-08-15
fix(FLUXDOC.md): add path to PackageDoc spec

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/flux/FLUXDOC.md
```diff
@@ -354,6 +354,7 @@ enum Doc {
 }
 
 struct PackageDoc {
+       path: String,
 	name: String,
```
scbrickley, line 358: Can we fix the formatting here?

stdlib/contrib/sranka/webexteams/webexteams.flux
```diff
@@ -35,10 +37,7 @@ message = (
 // `token` - string - [Webex API access token](https://developer.webex.com/docs/api/getting-started).
 // The returned factory function accepts a `mapFn` parameter.
 // The `mapFn` must return an object with `roomId`, `text` and `markdown` properties  as defined in the `message` function arguments.
-endpoint = (
-        url = "https://webexapis.com",
-        token,
-) => (mapFn) => (tables=<-) => tables
+endpoint = (url="https://webexapis.com", token) => (mapFn) => (tables=<-) => tables
```
scbrickley, line 40: Were the changes in this file just formatter changes? They don't seem relevant to the rest of the PR.

libflux/flux/FLUXDOC.md
scbrickley: Also, how are we populating the path when parsing? Is this field just going unused for now?

scbrickley (changes requested):

stdlib/contrib/sranka/webexteams/webexteams.flux
BarWeiner1: Yes this was formatting

libflux/flux/FLUXDOC.md
scbrickley: Just realized this is just the example code in the spec. I would prefer if changes to the spec were accompanied by the appropriate changes to the code. Can we include that as part of this PR?

BarWeiner1: What exactly is wrong with the formatting? Check out the most recent parser changes, but basically we are using the PolyTypeMap to pass in the relative path when populating.

scbrickley: Just the indentation. The fields in a struct should all be aligned.

Did we already do the work to include the path in `PackageDoc`? If so, then disregard my last comment.

BarWeiner1: Does this look better, it is rendering very weird on my device

scbrickley (approved):

PR #3962, opened 2021-08-16
feat(execute): modify aggregate transformation to implement transport

### LINKED ISSUE #3929: Implement Transport interface for AggregateTransformation
The existing `NewAggregateTransformation` function should be modified to implement the `Transport` interface directly.

It should demonstrate the following behavior:

* When a table chunk is received, it should load the previous state for the key. If no previous state exists, it should construct new state. It should then pass that chunk to the aggregate.
* When the flush key message is received, a table should be created and sent using the `TransportDataset`.

The `Transformation` interface should continue to be implemented. When `Process()` is called, it should act as if it received each table chunk and then a flush key message for that key. The `Finish()` method should flush any state that has not previously been flushed.

The `NewAggregateTransformation` function should be changed from:

```
func NewAggregateTransformation(d Dataset, c TableBuilderCache, agg Aggregate, config AggregateConfig) *aggregateTransformation { ... }
```

To:

```
func NewAggregateTransformation(id DatasetID, agg Aggregate, config AggregateConfig) (Transformation, Dataset, error) { ... }
```

Invocations of this method should be changed accordingly and any uses of `NewAggregateTransformationAndDataset` should be changed to use the above method instead.

This change should be behind a feature flag.

jsternberg:
This modifies the existing aggregate transformation to now be
`SimpleAggregateTransformation` and creates another
`AggregateTransformation` that is more general.

Both of these implement transport directly so they have support for
table chunks and flush key messages. This should work directly with any
transformations that utilize that interface.

The simple aggregate transformation is unmodified in how it is used.
There is an aggregate config along with an aggregate interface that can
be used to create the aggregate state. This will continue to work for
simple aggregates like min, max, and last.

The new aggregate transformation implements a more general interface
with an `Aggregate` function, which takes table chunks and
creates/manipulates state for a group key, and a `Compute` function
which transforms that state into a table. This can be used with some
more complex aggregates like `distinct()` and `mode()` while also being
useful for some functions that keep state and can only construct
themselves after receiving a flush key message, like `sort()`.

This change does not affect any of the selector implementations.

Closes #3929.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


execute/aggregate.go
```diff
@@ -1,33 +1,159 @@
...
+func NewAggregateTransformation(id DatasetID, t AggregateTransformation, mem memory.Allocator) (Transformation, Dataset, error) {
+	tr := &aggregateTransformation{
+		t: t,
+		d: NewTransportDataset(id, mem),
+	}
+	return tr, tr.d, nil
+}
+
+// ProcessMessage will process the incoming message.
+func (n *aggregateTransformation) ProcessMessage(m Message) error {
```
skartikey, line 54: Please let me know what n stands for here. if it's not obvious could you please change it to some readable var name?

execute/aggregate.go
```diff
@@ -1,33 +1,159 @@
...
+		}); err != nil {
+			return err
+		}
+	}
+	return n.flushKey(tbl.Key())
+}
+
+func (n *aggregateTransformation) processChunk(chunk table.Chunk) error {
+	state, _ := n.d.Lookup(chunk.Key())
+	if ns, ok, err := n.t.Aggregate(chunk, state, n.d.mem); err != nil {
```
skartikey, line 96: same for ns pls. I guess it stands for the new state?

execute/aggregate.go
```diff
@@ -1,33 +1,159 @@
...
 }
 
-var DefaultAggregateConfig = AggregateConfig{
+var DefaultSimpleAggregateConfig = SimpleAggregateConfig{
 	Columns: []string{DefaultValueColLabel},
 }
 
-func (c AggregateConfig) Copy() AggregateConfig {
+func (c SimpleAggregateConfig) Copy() SimpleAggregateConfig {
 	nc := c
```
skartikey, line 157: newConfig and config will make it more readable than nc, c

execute/aggregate.go
```diff
@@ -36,176 +162,157 @@ func (c AggregateConfig) Copy() AggregateConfig {
...
 	}
+	return NewAggregateTransformation(id, tr, mem)
 }
 
-func NewAggregateTransformationAndDataset(id DatasetID, mode AccumulationMode, agg Aggregate, config AggregateConfig, a *memory.Allocator) (*aggregateTransformation, Dataset) {
-	cache := NewTableBuilderCache(a)
-	d := NewDataset(id, mode, cache)
-	return NewAggregateTransformation(d, cache, agg, config), d
+type aggregateState struct {
+	inType flux.ColType
```
skartikey, line 185: making it colInType will make it more readable.

execute/aggregate.go
```diff
@@ -1,33 +1,159 @@
...
 )
 
+// AggregateTransformation implements a transformation that aggregates
+// the results from multiple TableView values and then outputs a Table
+// with the same group key.
+//
+// This is similar to NarrowTransformation that it does not modify the group key,
+// but different because it will only output a table when the key is flushed.
+type AggregateTransformation interface {
+	// Aggregate will process the TableView with the state from the previous
```
skartikey, line 21: sorry, what is `state` here. the value associated with the given key if it exists?

arrow/table_buffer.go
```diff
@@ -83,11 +83,41 @@ func (t *TableBuffer) Validate() error {
 		return nil
 	}
 
-	sz := t.Values[0].Len()
-	for i := 1; i < len(t.Values); i++ {
-		if t.Values[i].Len() != sz {
+	var sz int
```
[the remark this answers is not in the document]
jsternberg: This variable is initialized on the first iteration of this loop and is used in subsequent loops. If I put it in the for loop, it will not keep the size for subsequent loops.

I could initialize this once outside of the for loop instead of inside of the for loop, but then I'd have to add an additional check for nil because it's possible for the first column to be nil.

arrow/table_buffer.go
```diff
@@ -83,11 +83,41 @@ func (t *TableBuffer) Validate() error {
...
 		}
+		if ok := t.checkCol(t.Columns[i].Type, t.Values[i]); !ok {
+			return errors.Newf(codes.Internal, "column %s of type %s is incompatible with data array %T", t.Columns[i].Label, t.Columns[i].Type, t.Values[i])
+		}
 	}
 	return nil
 }
+
+func (t *TableBuffer) checkCol(typ flux.ColType, arr array.Interface) bool {
+	switch typ {
```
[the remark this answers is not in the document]
jsternberg: Null was checked earlier in validate so it shouldn't be needed here.

execute/aggregate.go
jsternberg: An artifact of it originally being a copy/paste job from `narrowTransformation`. Will change.

execute/aggregate.go
jsternberg: This is previously existing code that wasn't modified by this change. I'd prefer not to touch the code if that's ok.

execute/aggregate.go
jsternberg: Yes. It's the state associated with the group key that is being aggregated.

execute/aggregate.go
jsternberg: I find this a bit harder to read personally. I'd like to leave it as is if that's ok from personal preference. I'll add a comment to clarify it though.

execute/dataset.go
```diff
@@ -328,5 +337,6 @@ func (d *TransportDataset) Finish(err error) {
 		err:        err,
 	}
 	_ = d.sendMessage(m)
+	d.cache = nil
```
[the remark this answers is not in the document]
jsternberg: Modified to use the `Clear()` method on the cache.

skartikey (approved):
Looks good to me. Thanks.

Issue #1856, opened 2019-09-12
interface conversion: interface {} is nil, not string 


davidgs:
Submitting the query:

`import "strings"
import "sql"

humidity = from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "environment" and (r._field == "humidity"))
  |> last()
  |> map(fn: (r) => { return  {Sensor_ID: r.Sensor_id} })
  
  query = "SELECT * FROM Sensor_Data WHERE Sensor_ID is ${ humidity.Sensor_ID}"

sql.from(
  driverName: "mysql",
 dataSourceName: "foo:bar@tcp(localhost:3306)/IoTMeta",
 query: query
)`

Results in error: 

> interface conversion: interface {} is nil, not string

I've checked that `humidity.Sensor_ID` is not nil by adding a '|> yield()` before the `query` line, and it is there. (.csv of that attached). 

[2019-09-12-16-13_chronograf_data.csv.zip](https://github.com/influxdata/flux/files/3607280/2019-09-12-16-13_chronograf_data.csv.zip)

russorat:
happening to me as well with the following query:

this query works with the attached data.
```
import "strings"
import "influxdata/influxdb/v1"

from(bucket: "telegraf")
  |> range(start: -365d)
  |> filter(fn: (r) => r._measurement == "usgs")
  |> filter(fn: (r) => r.type == "earthquake" and r.status == "reviewed")
  |> filter(fn: (r) => r._field == "mag" or r._field == "place" or r._field == "latitude" or r._field == "longitude")// or r._field == "tsunami" or r._field == "url")
  |> filter(fn: (r) => exists r.magType)
  |> drop(columns: ["sources","host","properties_alert"])
  |> v1.fieldsAsCols()
```

but adding the line: ` |> filter(fn: (r) => strings.containsStr(v: r.place, substr: "any string"))` produces the following error: `panic: interface conversion: interface {} is nil, not string`
```
import "strings"
import "influxdata/influxdb/v1"

from(bucket: "telegraf")
  |> range(start: -365d)
  |> filter(fn: (r) => r._measurement == "usgs")
  |> filter(fn: (r) => r.type == "earthquake" and r.status == "reviewed")
  |> filter(fn: (r) => r._field == "mag" or r._field == "place" or r._field == "latitude" or r._field == "longitude")// or r._field == "tsunami" or r._field == "url")
[...]

Issue #2036, opened 2019-10-21
missing object properties (place)


russorat:
I have attached the data for the following query which works:

```
import "strings"
import "influxdata/influxdb/v1"
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "usgs")
  |> filter(fn: (r) => r.type == "earthquake" and r.status == "reviewed")
  |> filter(fn: (r) => r._field == "mag" or r._field == "place")
  |> filter(fn: (r) => exists r.magType)
  |> drop(columns: ["sources","host","properties_alert"])
  |> v1.fieldsAsCols()
  |> filter(fn: (r) => exists r.place)
```

then i add the line `  |> map(fn: (r) => ({ r with start_index: strings.index(v: r.place, substr: " of ") }))` and i get the following error: `cannot add type constraint @ 12:14-12:88: type error 0:0-0:0: missing object properties (place)`
```
import "strings"
import "influxdata/influxdb/v1"
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "usgs")
  |> filter(fn: (r) => r.type == "earthquake" and r.status == "reviewed")
  |> filter(fn: (r) => r._field == "mag" or r._field == "place")
  |> filter(fn: (r) => exists r.magType)
  |> drop(columns: ["sources","host","properties_alert"])
  |> v1.fieldsAsCols()
  |> filter(fn: (r) => exists r.place)
  |> map(fn: (r) => ({ r with start_index: strings.index(v: r.place, substr: " of ") }))
```

[2019-10-21-11-32_chronograf_data.csv.zip](https://github.com/influxdata/flux/files/3752376/2019-10-21-11-32_chronograf_data.csv.zip)

affo:
@russorat can you try adding a filter `exists r.place`? Does that fix your issue?

Sorry for the "meaningless" error message.

russorat:
@affo it’s already there

affo:
Yep, you are right. I am sorry.

@jlapacik this is definitely a bug in the `compiler`.

sbengo:
Hi, I'm facing this issue aswell

I have a tag that is not required in all series and I need to do a `pivot` over field and by all available tags.

[...]

PR #3967, opened 2021-08-17
fix(stdlib): interface conversion interface {} is nil, not string

skartikey:
An interface value consists of a "concrete value" and a "dynamic type": [Value, Type]
    
In strings functions, we were checking dynamic type only. So added nil check for concrete value as well
Also, added test using the interface as a function parameters

fixes: https://github.com/influxdata/flux/pull/1856 https://github.com/influxdata/flux/issues/2036


scbrickley (approved):
LGTM!

PR #3984, opened 2021-08-20
feat(libflux): created executable to retrieve all stdlib docs & updated wasm functions

BarWeiner1:
Created functions to add function to retrieve all stdlibs docs and print them out in an executable file for the static flux docs site. Also updated WASM functions to be able to retrieve all documentation at once.
### Done checklist
- [ ] Test cases written


libflux/flux/src/wasm/mod.rs
```diff
@@ -87,6 +87,15 @@ pub fn get_json_documentation(flux_path: &str) -> JsValue {
     doc
 }
 
+/// Gets json docs for all Stdlib
+#[wasm_bindgen]
+pub fn get_all_json() -> JsValue {
```
nathanielc, line 92: Should we name this `get_all_docs()` it will be expected that it is JSON.

libflux/flux/src/bin/flux-dump-docs.rs
```diff
@@ -0,0 +1,6 @@
+use flux::docs;
+
+fn main() {
+    let doc = docs();
```
nathanielc, line 4: I was expecting this to call a `doc_json` function that returns the raw json so that we do not have to use `serde` to serialize the docs again. Am I reading this code wrong?

libflux/flux/Cargo.toml
```diff
@@ -8,6 +8,11 @@ edition = "2018"
 name = "flux"
 crate-type = ["rlib", "staticlib", "cdylib"]
 
+[[bin]]
+name = "flux-dump-docs"
```
nathanielc, line 12: this config looks good. I like the name of the binary as well.

libflux/flux/src/bin/flux-dump-docs.rs
BarWeiner1: No you're right, I ended up scrapping the doc_json function and implementing the conversion the same way we do it in wasm's mod.rs for consistency. I could change it and add the function back, but I thought this would be easier to implement.

nathanielc (approved):

PR #3985, opened 2021-08-20
feat(stdlib): add new InfluxDB sample datasets and sample.alignToNow

sanderson:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/sample/sample.flux
```diff
@@ -82,3 +98,25 @@ list = () => array.from(
...
+//
+// ```
+// import "influxdata/influxdb/sample"
+//
+// sample.data(set: "birdMigration")
+//    |> sample.alignToNow()
+// ```
+
+alignToNow = (tables=<-) => {
+    _lastTime = (tables |> keep(columns: ["_time"]) |> last(column: "_time") |> findRecord(fn: (key) => true, idx: 0))._time
```
nathanielc, line 117: could we store the _lastTime for each dataset so we do not have to compute it again each time? Or does the data change enough that would get out of date often?

stdlib/influxdata/influxdb/sample/sample.flux
```diff
@@ -11,28 +11,44 @@ sets = [
...
+        url: "https://influx-testdata.s3.amazonaws.com/noaa.csv",
+        desc: "Water level observations from two stations reported by the NOAA Center for Operational Oceanographic Products and Services. Contains data between 2019-08-17 and 2019-09-17.",
+        size: "~10.3 MB",
+        type: "static",
     },
     "usgs": {
         url: "https://influx-testdata.s3.amazonaws.com/usgs-earthquake-all-week-annotated.csv",
         desc: "USGS earthquake data from the last week. Contains geotemporal data collected from USGS seismic sensors around the world. Data is updated approximately every 15m.",
         size: "~6 MB",
+        type: "live",
```
nathanielc, line 44: How does type get used?

stdlib/influxdata/influxdb/sample/sample.flux
sanderson: That would save some compute time, but I don't know the best way to inform the function what dataset is actually being used. I guess we could introspect the measurement from the input data and use that to identify the specific set being used. As is, this is just a generic approach that can be used with any input data, not just sample data.

stdlib/influxdata/influxdb/sample/sample.flux
sanderson: The package itself doesn't use the type, but we are beginning to make distinctions between "live" and "static" datasets in the documentation. This is really just for information's sake. The end use for each type is different – live data can be periodically loaded and written to InfluxDB with a task, static data only needs to be loaded and written once or, if not concerned about bandwidth, just loaded at query time.

stdlib/influxdata/influxdb/sample/sample.flux
sanderson: I spec'd this approach out and have some reservations. This is what the measurement introspection approach I took:

```js
alignToNow = (tables=<-) => {
    _measurement = (tables |> findRecord(fn: (key) => true, idx:0))._measurement
    _endDates = [
        "machine": 2021-08-01T23:59:59.99999999Z,
        "airSensor": 2020-08-01T23:59:59.99999999Z
    ]
    _offset = int(v: now()) - int(v: dict.get(dict: _endDates, key: _measurement, default: now()))
    _offsetDuration = duration(v: _offset)

    return tables |> timeShift(duration: _offsetDuration)
}
```

This did prove to be _slightly_ more performant in my testing (5ms faster on average with an 11MB dataset), but there are a few things I don't love about it:

- We have to update the function anytime we add new sample datasets.
- We'd have to update the function with new dataset endDates if we decide the update the datasets (which does happen).
- The function will only work with the predefined datasets, so we lose some flexibility.

Thoughts?

sanderson:
bump

nathanielc (approved):

Issue #3461, opened 2021-01-29
Function argument defaults should not introduce type constraints


philomory:
There are built-in functions that accept multiple types for the same argument; one prominent example is `range()`, whose `start` parameter can accept either a Time, or a Duration. If you write a function in Flux that passes one of its arguments to `range()` as the `start` parameter, your function now _also_ accepts multiple types for that argument. For example:

```
fun = (tables=<-, x) => tables |> range(start: x)

data |> fun(x: -10m) # this is fine
data |> fun(x: 2021-01-28T12:00:00Z) # so is this
```

However, if your function provides a default value for the argument, suddenly your callers are constrained to call your function with a value of the same type as the default:

```
fun = (tables=<-, x=-20m) => range(start: x)

data |> fun(x: -10m) # this is fine
data |> fun(x: 2020-01-28T12:00:00Z) # this is a type error
```

The most obvious place this comes up is `schema.tagValues` (and similar functions); becasue `schema.tagValues` has a default `start` value of `-30d`, it's not possible to pass an absolute time, even though nothing `schema.tagValues` _does_ with the `start` argument requires that it be a duration and not a time. `schema.tagValues` is even [documented](https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/influxdb-schema/tagvalues/#start) as taking either a duration or a time, but in actual fact it cannot.

There are [issues](https://github.com/influxdata/flux/issues/1071) open about this specific case, and the proposed solution seems to be to make duration literals polymorphic, so they can fill in for times. That may be a fine idea (though I'm not sure it would actually solve the problem, given that the types are backwards), but it still leaves the general case unsolved. For example, consider the following (rather silly but perfectly valid) function:

```
fun = (x) => "Hello, ${string(v: x)}"

fun(x: "bob") # returns "Hello, bob"
fun(x: 10) # returns "Hello, 10"
```

This function _also_ breaks if a default value is added:

```
fun = (x="World") => "Hello, ${string(v: x)}"

fun(x: "bob") # returns "Hello, bob"
fun(x: 10) # is now a type error
```

I think this is unreasonably restrictive; adding a default value for a function argument should never make a previously valid function call invalid.

nathanielc:
See #3629 it is a duplicate and has a good write up of the problem as well.

PR #3987, opened 2021-08-23
feat(stdlib/pagerduty): add custom_details field support

alespour:
Extends Flux to partially satisfy https://github.com/influxdata/feature-requests/issues/300.

Since it is not possible to extend existing `sendEvent` and `endpoint` with additional parameters (and/or behavior) and preserve API backward compatibility at the same, they are left unchanged and new pair of functions `sendEventEx` and `endpointEx` are introduced, doing pretty much the same but with additional fields `custom_details` and `component` (which was documented in the code before but not implemented).

There are two reasons why existing functions cannot be extended without breaking backward compatibility. First, record type has no default value (`{}` infers empty record type), second, `exists` operator does not work for records other than retrieved from InfluxDB , so it is unusable in `pagerduty.endpoint` therefore user `mapFn` function must built record with all fields passed to `sendEvent`.

### Done checklist
- [x] Test cases written


ivankudibal:
merge depends on https://github.com/influxdata/flux/issues/3461

nathanielc:
Hi @ivankudibal and @alespour I just added a feature to Flux that should allow you to update this PR in a backwards compatible way. See https://github.com/influxdata/flux/pull/4041 In short you can set the default value to `record.any` instead of `{}` and it will allow any record to be passed in and default to the empty record.

alespour:
Hi @nathanielc, this is very useful. Function `sendEvent` arguments can be safely extended with record type args. 

There is still however compatibility issue with `endpoint` function. User calls it with `mapFn` function argument. This function returns a record whose fields are then used as arguments to `sendEvent` function. We've tried before to use `exists` operator to check if given field exists in the record and if not, provide some default, but it does not work (I think it was Chris Wolff who told me that this operator works only on records returned from InfluxDB).
So, existing user code where `mapFn` does not return the new fields would fail.

I've added `builtin get : (r: A, key: string, default: B) => B where A: Record` method to `record` pkg that could be used to solve this, I've tested and it works. ``. Draft PR #4049.

alespour:
So, if the the new function `record.get` in PR #4049 is acceptable solution, I will complete that PR with tests and make it avail for review. And after it's merged, I could finish this PR.

nathanielc:
> We've tried before to use exists operator to check if given field exists in the record and if not, provide some default, but it does not work 

This should work when using `record.any`. Can you share what you tried and how it failed?

alespour:
If I add a line like
```
   customDetails: if exists obj.customDetails then obj.customDetails else record.any
```
after https://github.com/influxdata/flux/blob/5cd8f8c037857879511306c03e702da81d255741/stdlib/pagerduty/pagerduty.flux#L227

I get `record is missing label customDetails`. `obj` is a record returned by user supplied function, so this would break existing user code.

nathanielc:
@alespour That looks like a good solution. We had a team discussion to see if there are other options, but we agree that `record.get` is the best short term solution.

However we want to add some exceptions to the `record.get` implementation.

* It needs to be experimental and with documentation that we plan to remove/update it once we fix `exists` to work with all records
* It needs to have documentation that states that the key must be a string literal, like you have shown. But uses that would pass in a dynamic string value should be forbidden, unfortunately it is not easy to enforce this in the code.

In short adding `record.get` is a hack  as it essentially introduces reflection into the Flux engine. Once we fix the `exists` behavior we expect that we will be able to update the implementation of `record.get` to not need this psuedo-reflection. 

Thanks for understanding as we work through this difficult challenges in designing a language.

stdlib/pagerduty/pagerduty.flux
```diff
@@ -143,7 +149,16 @@ sendEvent = (
         "Accept": "application/vnd.pagerduty+json;version=2",
         "Content-Type": "application/json",
     }
-    enc = json.encode(v: data)
+    enc = if customDetails == record.any then
```
nathanielc, line 152: why do we need this conditional? Is there an error encoding an empty JSON object as the custom details?

nathanielc (commented):
LGTM I have a one question

PR #3994, opened 2021-08-24
feat(stdlib): add sampledata package with basic sample datasets

sanderson:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/sampledata/sampledata.flux
```diff
@@ -0,0 +1,351 @@
...
+option start = 2021-01-01T00:00:00Z
+option stop = 2021-01-01T00:01:00Z
+
+_numeric = (includeNull=false) => "
+#group,false,false,false,true,false
+#datatype,string,long,dateTime:RFC3339,string,double
+#default,_result,,,,
+,result,table,_time,tag,_value
+,,0,2021-01-01T00:00:00Z,t1,-2.18
+,,0,2021-01-01T00:00:10Z,t1," + (if not includeNull then "10.92" else "") + "
```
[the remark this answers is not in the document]
nathanielc: Makes sense, thanks for explaining the reasoning behind this. I'll take another look

stdlib/sampledata/sampledata.flux
```diff
@@ -0,0 +1,310 @@
...
+//
+// ## Output basic sample data with float values
+//
+// ```
+// import "sampledata"
+//
+// sampledata.float()
+// ```
+// 
+// ## Output data
```
nathanielc, line 81: What is the motivation for including the output in the doc comment directly? Would it be easy enough to just run the example to see the data? Having the data in the comment means that if it changes we have to update the doc. Also it doesn't show what the data looks like with nulls so to see that you would have to just run the function. Thoughts? The data shouldn't change much but still wanted to ask about it.

stdlib/sampledata/sampledata.flux
```diff
@@ -0,0 +1,310 @@
...
+// | tag | _time                | _value |
+// | :-: | :------------------- | -----: |
+// | t2  | 2021-01-01T00:00:00Z |  19.85 |
+// | t2  | 2021-01-01T00:00:10Z |   4.97 |
+// | t2  | 2021-01-01T00:00:20Z |  -3.75 |
+// | t2  | 2021-01-01T00:00:30Z |  19.77 |
+// | t2  | 2021-01-01T00:00:40Z |  13.86 |
+// | t2  | 2021-01-01T00:00:50Z |   1.86 |
+// 
+float = (includeNull=false) => {
```
nathanielc, line 101: Do we consider the specific data returned to be part of the API of the function? I.e. if the data changed would we consider that a breaking change? I think we would have too because otherwise examples could break or any test cases that use the data.

stdlib/sampledata/sampledata.flux
```diff
@@ -0,0 +1,310 @@
+// Package sampledata provides functions that return basic sample datasets.
+package sampledata
+
+
+import "csv"
+
+option start = 2021-01-01T00:00:00Z
+option stop = 2021-01-01T00:01:00Z
```
nathanielc, line 8: These options are not used within this package, what do they do?

nathanielc (commented):
Looks great, a few more questions but no specific changes requested.

stdlib/sampledata/sampledata.flux
wolffcm: So is the idea here that we can have completely self-contained examples in  our docs that the user could copy and paste and run in the UI or CLI (or maybe even in an interactive text box right in the docs themselves)?

wolffcm (commented):
This looks good to me. I just had a question about the planned usage for this package.

stdlib/sampledata/sampledata.flux
sanderson: Yep, that's exactly what they're designed for.

stdlib/sampledata/sampledata.flux
sanderson: Visibility. Just a way to see the sample data whenever the docs are generated from the stdlib.

wolffcm (approved):

stdlib/sampledata/sampledata.flux
nathanielc: What does changing them do? I don't see any use of them within the package so I am confused what they do. Yes a comment would be great as I am not quite understanding their purpose.

wolffcm: I guess they could just be made into regular variables, just by removing the `option` keyword?

sanderson:
@nathanielc I made `start` and `stop` package constants rather than options and added comments explaining why they exist.

stdlib/sampledata/sampledata.flux
```diff
@@ -0,0 +1,193 @@
+// Package sampledata provides functions that return basic sample datasets.
+package sampledata
+
+
+import "csv"
+
+// start and stop constants provide start and stop times to be used with `range`
+// when `_start` and `_stop` columns are required to demonstrate a transformation.
+//
+start = 2021-01-01T00:00:00Z
```
nathanielc, line 10: Does this mean that all sample  data is within the range of start and stop?

nathanielc (approved):
LGTM I just had one clarifying question for the start/stop constants

stdlib/sampledata/sampledata.flux
sanderson: Yes, they do. I updated the comment to clarify.

Issue #3807, opened 2021-06-16
monitor.flux: deadman alerts not generating a state change notification ('when status changes from')


umaplehurst:
Related to: https://github.com/influxdata/influxdb/issues/18769 -- but may not cover all possible causes for the problem(s) that people are seeing

Testing on InfluxDB Cloud I am finding that I can reproduce following:

a) set up a deadman check, with "Schedule Every" set to 1m, and "When values are not reporting for" set to 5m,
b) i) set up a notification rule, "When status changes from" `OK` to `ANY`,
b) ii) set up a notification rule, "When status changes from" `ANY` to `OK`,

A notification is successfully generated for the transition out of the deadman condition (dead host comes back to life) but not into the deadman condition (alive host becomes dead). It seems that one important precondition for causing this bug is setting the deadman alert check "When values are not reporting for" threshold for a longer period of time (say, 5m) than the notification rule Schedule Task Every time (say, 1m).

I did some dissecting of bits of `monitor.flux` in the Query Editor. If my analysis and understanding of how everything works is right, `_source_timestamp` stops ticking as soon as the incoming data feed goes dead; i.e. for the alert checks that subsequently occur after as the host has gone dead, it will keep the same value of the very last successfully received sample from the host, but `_time` will still keep incrementing with the alert check time.

So, if the incoming data has been dead for quite some time, let's say 5m, by the time `_stateChanges` is executed in `monitor.flux` then the last two 1m-apart check samples that it will compare for the state change (e.g. for `OK` --> `CRIT` when doing a deadman check) will not `sort` because both of those samples will have the exact same value of `_source_timestamp` (as the input source has been dead for quite some time now). This means the subsequent `difference`/`filter` logic in `_stateChanges` will not work as expected, as the code expects the rows to be sorted in the correct direction of time flow. And this results in the bug that the notification is never generated.

The simple fix here is to extend the list of columns in the `sort` and to include `_time` as a second sort column after `_source_timestamp`. This way, the rows will always be sorted by `_time` if sorting by `_source_timestamp` doesn't provide an ordering. I tested this on InfluxDB Cloud by manually modifying the default Notification Rule that the UI creates and editing `monitor["stateChanges"]` in that UI-generated code to use my own patched `_stateChanges` function with this `sort` change described. This sorts out the problem for me and `OK` --> `CRIT` (i.e. deadman condition) notifications are now correctly being generated further upwards to Slack.

umaplehurst:
[...]

PR #3996, opened 2021-08-25
fix(stdlib): sort statuses also by _time

alespour:
This PR fixes issue in alerting where, depending on check and rule scheduling and timing configuration, state changes are sometimes not detected and therefore notifications not sent.

The problem is that status of data points may be evaluated multiple times with certain scheduling, for example as reported in #3807, deadman check evaluates status based on data point time and not a metric value. That results in existence of multiple statuses with the same source timestamp but different time (and level) in the monitoring bucket. Then, when an alert rule attempts to detect state change, it gets incorrect sequence of statuses, because sorting by only (the same) source timestamp is insufficient.

Affected functions are `stateChangesOnly` and `_stateChanges` functions in `monitor` package, where before `sort`,
`_level` is dropped from the group key which results in a new series with possibly unpredictable order (by `_time`).

The issue is a side effect of the changes in #2725. Mea culpa.

### Done checklist
- [x] Test cases written


wolffcm (changes requested):
The fix looks good and the changes make sense. Thanks for fixing this!

Can I trouble you to adapt your test to our newer format which uses a Flux keyword `testcase`? We want to be using it moving forward. Here is an example:
https://github.com/influxdata/flux/blob/955bc0303789cce8938d69b55c7e1697962f1d7e/stdlib/universe/window_test.flux#L7

alespour:
@wolffcm Done. Nice feature!

wolffcm (approved):
Look great. Thanks again!

jsternberg:
Fixed using @wolffcm's first suggestion in https://github.com/influxdata/flux/pull/4012.

PR #3998, opened 2021-08-26
fix: delete the go formatter code as it is not longer used

skartikey:
removed ast/format.go and ast/format_test.go
also removed test_rewriter package as its not longer in use


danxmoran:
Kapacitor switching over to the new formatter in influxdb/kapacitor#2618

danxmoran:
The remaining occurrences of `ast.Format` in kapa are from a different `ast` package (for TICKscript)

skartikey:
Thanks @danxmoran, will merge this PR then.

PR #4005, opened 2021-08-30
fix: removed flux wasm crate and moved it to lsp

BarWeiner1:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
Are there stale dependencies in the Cargo.toml as a result of this change? Does this mean that the `flux` repo doesn't need any wasm knowledge?

BarWeiner1:
I am pretty sure the flux repo no longer needs any wasm knowledge, which dependencies from Cargo.toml can we safely remove because of that?

Makefile
```diff
@@ -108,7 +108,6 @@ staticcheck:
 		`go list ./... | grep -v '\/flux\/stdlib\>'`
 	GO111MODULE=on ./gotool.sh honnef.co/go/tools/cmd/staticcheck ./stdlib/...
 
-test: test-go test-rust test-flux test-wasm
```
scbrickley, line 111: We don't want to remove this line entirely. This is what CI uses to run tests. We just want to remove `test-wasm` from the end of this line.

scbrickley (changes requested):
Just one change, but other than that this looks fine.

scbrickley (approved):
LGTM

rockstar (approved):

Issue #3999, opened 2021-08-27
delete packages which are no longer in use


skartikey:
List of packages that need to be removed from the flux repository, as they are no longer in use. 

List - 
internal/refactortests

PR #4006, opened 2021-08-31
fix(internal): delete packages which are no longer in use

skartikey:
List of packages that need to be removed from the flux repository, as they are no longer in use.

List -
internal/refactortests

fixes: #3999


BarWeiner1 (approved):
LGTM!

PR #4009, opened 2021-09-01
fix: added FormatDuration method which can be exported in other repo

skartikey:
Added FormatDuration method which can be exported from other repo.

### DIFF
--- a/parser/strconv.go
+++ b/parser/strconv.go
@@ -1,6 +1,8 @@
 package parser
 
 import (
+	"strconv"
+	"strings"
 	"unicode/utf8"
 
 	"github.com/influxdata/flux/ast"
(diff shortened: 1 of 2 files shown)


parser/strconv.go
```diff
@@ -48,6 +50,19 @@ func ParseDuration(lit string) (*ast.DurationLiteral, error) {
 	}, nil
 }
 
+// FormatDuration will convert DurationLiteral into formatted string representation
+func FormatDuration(n *ast.DurationLiteral) string {
+	builder := strings.Builder{}
+	formatDuration := func(d ast.Duration) {
+		builder.WriteString(strconv.FormatInt(d.Magnitude, 10))
+		builder.WriteString(d.Unit)
```
nathanielc, line 58: Why is this code abstracted through a function? Would it work to simply move the body of this function inside the loop?

nathanielc (changes requested):
LGTM with one question about the implementation.

parser/strconv.go
skartikey: I copied the implementation from the existing ast formatter. 
I agree it doesn't make sense to abstract it through function. I will make a change now. 
Thanks for noticing.

nathanielc (approved):

PR #4010, opened 2021-09-01
feat(contrib): add array manipulation functions

### LINKED ISSUE #3397: Add support for array merging
The goal is to be able to write something like:
```
good = ["foo", "bar"]

from(bucket: "xxx")
  ...
  |> keep(columns: ["_time", "_value"] + good)
```

For more information, see https://community.influxdata.com/t/how-can-one-merge-two-arrays/17171.

alespour:
This PR adds contrib package `array` which provides set of functions that work with arrays. Also fixes #3397.

* `append` - appends value `v` to array `arr`, `v` must be an array of the same type as `arr`

```
import "contrib/bonitoo-io/array"

good = ["foo", "bar"]

from(bucket: "my-bucket")
   |> range(start: -1h)
   |> keep(columns: array.append(arr: ["_time", "_value"], v: good)
```

* `map` - applies provided function to each element and returns new array

```
import "contrib/bonitoo-io/array"

ia = [1, 1]

sa = array.map(arr: a, fn: (x) => string(v: x))
```

Array may be represented as JSON-encoded string for use in Flux tables, because array is not supported column type. Functions that with with such arrays are:

* `fromStr` - decodes JSON-encoded string representing `arr` into array

```
import "contrib/bonitoo-io/array"

s = "[1, 2, 3]"

a = array.fromStr(arr: s)
```

* `appendStr` - appends value `v` to array `arr` represented as JSON-encoded string, `v` must must be an array of the same type as `arr` (in its decoded form)

```
import "contrib/bonitoo-io/array"

from(bucket: "my-bucket")
  |> range(start: -1h)
  |> reduce(
     fn: (r, accumulator) => ({
          sarr: array.appendStr(arr: accumulator.sarr, v: [r._value])
      }),
      identity: {
          sarr: array.emptyStr  // "[]"
      }
  )
  |> map(fn: (r) => ({ r with status:
      http.post(
          url: "http://endpoint:12345/",
          data: json.encode(v: array.fromStr(arr: r.sarr))
      )})
 )
```
The need to aggregate over table into an array was discussed in the [community Slack](https://influxcommunity.slack.com/archives/CH8RV8PK5/p1629355478165400?thread_ts=1629281233.163700&cid=CH8RV8PK5).  Such "array" aggregation is also done by hand `geo.ST_LineString` function.

### Done checklist
- [x] Test cases written


rockstar (approved):

stdlib/contrib/bonitoo-io/array/array.flux
```diff
@@ -0,0 +1,106 @@
...
+// ```
+// import "contrib/bonitoo-io/array"
+//
+// good = ["foo", "bar"]
+//
+// from(bucket: "my-bucket")
+//   |> range(start: -1h)
+//   |> keep(columns: array.append(arr: ["_time", "_value"], v: good)
+// ```
+builtin append : (arr: [A], v: [A]) => [A]
```
nathanielc, line 27: Should we call this `concat` instead of append as its really creating a new array that is the concatenation of two other arrays?

stdlib/contrib/bonitoo-io/array/array.flux
```diff
@@ -0,0 +1,106 @@
...
+// ## Parse records array
+//
+// ```
+// import "contrib/bonitoo-io/array"
+//
+// s = "[{\"n\": 1}, {\"n\": 2}, {\"n\": 3}]"
+//
+// a = array.fromStr(arr: s)
+// ```
+fromStr = (arr) => ejson.parse(data: bytes(v: arr))
```
nathanielc, line 73: Can you explain the user need for these functions? Seems like if you have arrays as JSON strings you would use `json.parse` directly. Why do we want these helper functions in addition?

stdlib/contrib/bonitoo-io/array/array.flux
```diff
@@ -0,0 +1,106 @@
+// Package array provides functions for interacting with arrays.
+package array
+
+
+import ejson "experimental/json"
+import "json"
+
+// append is a function that appends value to array.
```
nathanielc, line 8: This description is misleading as it really append an array of values to another array of values.

nathanielc: Should we also explicitly state that new memory space is allocated for this purpose, meaning this is not an efficient function?

nathanielc (changes requested):

stdlib/contrib/bonitoo-io/array/array.flux
alespour: Done, thank you.

At the first attempt I wanted to implemented as _append_, but then I changed my mind and decided to keep it coherent with immutability of variables in Flux, but left the name of the function unchanged.

stdlib/contrib/bonitoo-io/array/array.flux
alespour: `concatStr` (`appendStr` before) is for use in `reduce`, user cannot aggregate values in an array (in _identity_ record field) because it is not supported Flux column type.
```
import "contrib/bonitoo-io/array"

 from(bucket: "my-bucket")
   |> range(start: -1h)
   |> reduce(
       fn: (r, accumulator) => ({
           sarr: array.concatStr(arr: accumulator.sarr, v: [r._value])
       }),
       identity: {
           sarr: array.emptyStr  // "[]"
       }
   )
   |> map(fn: (r) => ({ r with status:
       http.post(
           url: "http://endpoint:12345/",
           data: json.encode(v: array.fromStr(arr: r.sarr))
       )})
  )
```
`fromStr` is a just helper that decodes such array-string into an array and is used by `concatStr`.

stdlib/contrib/bonitoo-io/array/array.flux
alespour: Fixed, thank you.

stdlib/contrib/bonitoo-io/array/array.flux
```diff
@@ -0,0 +1,107 @@
...
+// ```
+// import "contrib/bonitoo-io/array"
+//
+// s = "[{\"n\": 1}, {\"n\": 2}, {\"n\": 3}]"
+//
+// a = array.fromStr(arr: s)
+// ```
+fromStr = (arr) => ejson.parse(data: bytes(v: arr))
+
+// concatStr joins two arrays. It returns JSON-encoded array represented as a string.
```
noramullen1, line 75:
```suggestion
// concatStr joins two arrays. It returns a JSON-encoded array represented as a string.
```

nathanielc:
I played with this branch today and was able to use it with great success!

 I am concerned that the `concatStr` function will be too slow for most use cases as we are constantly serializing and deserializing the data into a string. Could we maybe remove those functions from this PR and get it merged? Then we could open a subsequent PR to discuss how to achieve the reduce behavior you are looking for.

stdlib/contrib/bonitoo-io/array/array.flux
alespour: Removed `*Str` functions due to perf concerns.

alespour:
@nathanielc I have removed string de/serialization functions and will prepare a separate PR with them.

nathanielc:
@alespour This is great! I'll merge this today. I am going to move it to the `experimental/array` package so that users don't have to import two different array packages. Then we can promote these functions into the official stdlib array package after seeing users consume them in the wild. 

Thanks for these functions they are going to make a big different is the usability of Flux!

nathanielc (approved):

PR #4011, opened 2021-09-01
chore: Revert commit with failing test

scbrickley:
Reverts changes from #3996, which introduced flux test that fails when run using the influxdb test harness.


jsternberg (approved):

wolffcm (approved):
Hmm, I reviewed and merged the PR, and I tested it against influxdb. Did you capture the failure somewhere?

Issue #3993, opened 2021-08-24
[Timebox] POC: semantic graph vectorize transformation


wolffcm:
Implement a proof of concept for a vectorization transformation that transforms semantic graph. It might be implemented as a function with a signature like this:
```rust
fn vectorize(input: FunctionExpr, param_to_vectorize: String) -> Result<FunctionExpr, Error> {
    // ...
}
```
The semantic graph should be transformed such that all uses of the vectorized input parameter (usually `r`) and derived uses will be transformed to do elementwise operations.

For reference every kind of semantic graph expression is listed here:
https://github.com/influxdata/flux/blob/1a84fac74554ebb7bd1674c8671ec96049de0055/libflux/flux-core/src/semantic/nodes.rs#L110-L133

DOD:
- A function as described above is implemented, with some unit test that demonstrate common use cases.
- A short writeup suitable for discussion containing salient points of the approach taken, and what changes to semantic graph (type system, kinds of expressions) are needed
- Each vectorized expressions should have clear semantics that are described in the above writeup
- Determine whether we can vectorize every possible semantic graph (e.g., in a call to `map` or `filter`)
- No need to actually implement compiler support.

wolffcm:
Here's some notes on what I did. All of the code for this is on this branch:
https://github.com/influxdata/flux/tree/feat/poc-vector-monotype

# The POC

For this POC I modified Rust code that works with the semantic graph.

I added a monotype called `Vector` that represents the type of the fields of a record passed to
a vectorized function.
In this context a vector is similar to the Flux `array` type but this type can only be created by a vectorization pass that
has been applied to a semantic graph.
In addition, the `vector` monotype meets any constraint that its element type meets. I.e., `vector(int)` is `Divisable` but `vector(string)` is not.

Then I added a Rust function that transforms a `FunctionExpression` (semantic graph node) and update the types to be consistent with vectorization.
The basic idea of this pass is to transform a function expression (which would be passed to `map`) to go from a type like this:
```
(r: {a: int, b: float: c: T0 | T1}) => {d: int, e: float, f: T2}
```
To something like this:
```
(r: {a: vector(int), b: vector(float), c: vector(T0) | T1}) => {d: vector(int), e: vector(float), f: vector(T2)}
```
That is, the fields of the record would be converted to represent whole columns, and the function body would have its types updated accordingly.

I leveraged type inference to do this:
- First I made a small change to type inference so that it could start with a set of initial constraints provided by the caller
[...]

PR #4022, opened 2021-09-08
Vectorization POC

wolffcm:
See here for discussion of this PR:
https://github.com/influxdata/flux/issues/3993#issuecomment-915646640


wolffcm:
This has moved beyond the POC phase, and some work is even in production.

PR #4026, opened 2021-09-09
fix: add boolean package to prelude

### LINKED ISSUE #4013: Create boolean package
Currently, boolean values can't be used in packages that `universe` depends on because it creates a circular dependency. We should create a new `boolean` package with no dependencies that can be imported into other packages.

nathanielc:
Prior to this change any dependencies of the universe package could not
reference true and false identifiers. Now there is a boolean package
that has no dependencies and is part of the prelude. This allows any
other package to import boolean when necessary.

Fixes #4013 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/boolean/boolean.flux
```diff
@@ -0,0 +1,9 @@
+// Package boolean provides constants for true and false values.
```
jsternberg, line 1: Can you change the boolean package to something like `internal/vars`? I'd like to avoid adding a new package that confuses users to believe it is something other than an internal detail. This package is only needed to resolve a dependency cycle in the prelude for the standard library and should never be imported manually. The internal signifies that this should not be used directly. I like `vars` as the base name because it signifies we can put other builtin variables in here too.

stdlib/universe/universe.flux
```diff
@@ -283,11 +280,11 @@ _sortLimit = (n, desc, columns=["_value"], tables=<-) => tables
 
 // top sorts a table by columns and keeps only the top n records.
 top = (n, columns=["_value"], tables=<-) => tables
-    |> _sortLimit(n: n, columns: columns, desc: true)
```
jsternberg, line 286: These changes should not be needed and an explicit import should not be needed. With your change to add the package to the prelude, I don't believe any changes to `universe` should be needed. I think putting it in the prelude list where it is is a pretty elegant solution.

jsternberg (changes requested):
I'd like to see the package name changes to something that signifies it as an internal package that is a part of the prelude. The exact package name isn't that important, but just putting it in `internal` is my preference. I am not a particular fan of the name `booleans` just because I think it's an awkward name, but some minor pushback and I don't really care.

The other comment is about reverting the `universe.flux` file. None of the changes made to it should be needed if I am understanding this change correctly. If I'm wrong, I'd be curious to know what happens.

stdlib/boolean/boolean.flux
nathanielc: I move the package to `internal/boolean`. I  chose the name `boolean` so its clear that no other values should be added to the package. Since we are trying to avoid any dependencies within the boolean package, because any of those package will not be able to import the boolean package. This way its clear nothing more should be added to this boolean package.

stdlib/universe/universe.flux
nathanielc: Made a small change to the way we infer the prelude and this now works.

jsternberg (approved):

PR #4033, opened 2021-09-10
test: create a test for group performance

wolffcm:
Because of some formatting issues, it looks like more changed than what I did. There is just one new file:
```
stdlib/universe/group_perf_test.flux
```
This file contains a unit test that should reproduce an issue with group performance.


wolffcm:
This branch is now stale.

Issue #3928, opened 2021-08-04
Add GroupTransformation Transport


jsternberg:
Add a `GroupTransformation` interface and implement a `Transport` that uses the interface. Roughly, the code skeleton will be similar to this:

```
// GroupTransformation is a transformation that can modify the group key.
// Other than modifying the group key, it is identical to a NarrowTransformation.
//
// The main difference between this and NarrowTransformation is that
// NarrowTransformation will pass the FlushKeyMsg to the Dataset
// and GroupTransformation will swallow this Message.

var _ Transport = (*groupTransformation)(nil)
var _ Transformation = (*groupTransformation)(nil)

type groupTransformation struct {
    t GroupTransformation
    d *TransportDataset
}

func NewGroupTransformation(id DatasetID, t GroupTransformation, mem memory.Allocator) (Transformation, Dataset, error) {
    return &groupTransformation{
        t: t,
        d: NewTransportDataset(id, mem),
}
```

The group transformation should implement both the `Transport` and `Transformation` interfaces.

* `Process` from `Transformation` should invoke `Process` on each chunk. It should not invoke `FlushKey`.
* `Finish` from `Transformation` should pass on the error using the dataset.
* `ProcessMsg` and `FinishMsg` should invoke the appropriate `Transformation` method in `ProcessMessage`.
* `ProcessChunkMsg` should be handled in `ProcessMessage`. `FlushKeyMsg` should be acknowledged and dropped.
* All other methods and messages can be no-ops.

jsternberg:
Fixed by #4003.

PR #4035, opened 2021-09-13
feat: update group to use new GroupTransformation interface

scbrickley:
This PR adds a feature flag that enables `group` to use chunk-based processing by leveraging the new `GroupTransformation` interface introduced in #3928.

In the process of implementing this, a bug was found in the interaction between `group` and `quantile` when chunk-based processing was enabled for both. This issue has been temporarily fixed by passing an empty context into the `quantile` transformation so that it is not affected by any feature flags, and therefore cannot use chunk-based processing. Issue #4042 has been created to track work for a longer term fix, and a [comment](https://github.com/influxdata/flux/blob/f1cf96439139e6deefb470f2965f21c128501be9/stdlib/universe/quantile.go#L226-L237) has been left in the code to further explain the problem.


stdlib/universe/group.go
```diff
@@ -150,7 +182,7 @@ type groupTransformation struct {
 	keys []string
 }
 
-func NewGroupTransformation(spec *GroupProcedureSpec, id execute.DatasetID, mem *memory.Allocator) (execute.Transformation, execute.Dataset) {
+func NewGroupTransformation(ctx context.Context, spec *GroupProcedureSpec, id execute.DatasetID, mem *memory.Allocator) (execute.Transformation, execute.Dataset) {
```
jsternberg, line 185: Can you change this to return an error too so that the error from `NewGroupTransformation` can be standardized?

stdlib/universe/group.go
```diff
@@ -242,6 +279,55 @@ func (t *groupTransformation) appendTable(ab *table.BufferedBuilder, tbl flux.Ta
 	return ab.AppendTable(tbl)
 }
 
+func (t *groupTransformation) groupChunkByRow(tbl table.Chunk, d *execute.TransportDataset, mem arrowmem.Allocator) error {
```
jsternberg, line 282: I'd like to see us consolidate this with the existing one, but nothing is coming to mind.

jsternberg (approved):
One incredibly minor change that doesn't even affect anything.

PR #4036, opened 2021-09-13
feat(execute): add narrow state transformation transport

### LINKED ISSUE #3933: Add NarrowStateTransformation Transport
Add a `NarrowTransformation` interface and implement a `Transport` that uses the interface. Roughly, the code skeleton will be similar to this:

```
// NarrowStateTransformation is the same as a NarrowTransformation
// except that it retains state between processing buffers.
type NarrowStateTransformation interface {
	// Process will process the table.Chunk. It will pass the previous state or nil to the Process method
        // on each call. If this method returns true for the second argument, the returned interface will be stored
        // as the state for this group key for the next function call.
	Process(chunk table.Chunk, state interface{}, d *Dataset, mem memory.Allocator) (interface{}, bool, error)
}

var _ Transport = (*narrowStateTransformation)(nil)
var _ Transformation = (*narrowStateTransformation)(nil)

func NewNarrowStateTransformation(id DatasetID, t NarrowStateTransformation, mem memory.Allocator) (Transformation, Dataset, error) { ... }
```

The narrow state transformation should implement both the `Transport` and `Transformation` interfaces.

* `Process` from `Transformation` should invoke `Process` on each chunk and then invoke `FlushKey` on the dataset.
* `Finish` from `Transformation` should pass on the error using the dataset.
* `ProcessMsg` and `FinishMsg` should invoke the appropriate `Transformation` method in `ProcessMessage`.
* `ProcessChunkMsg` and `FlushKeyMsg` should be handled in `ProcessMessage`.
* All other methods and messages can be no-ops.

jsternberg:
The narrow state transformation transport implements the transport
interface for narrow transformations where state is maintained between
table chunks.

This also fixes some areas in the tests where double releases were
happening accidentally due to forwarding a table chunk that did not have
a proper retained reference. It also fixes the narrow transformation
test with flush key which erroneously did not test narrow
transformations.

Fixes #3933.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/narrow_state_transformation.go
+++ b/execute/narrow_state_transformation.go
@@ -1,0 +1,101 @@
+package execute
+
+import (
+	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute/table"
+)
+
+// NarrowStateTransformation is the same as a NarrowTransformation
+// except that it retains state between processing buffers.
+type NarrowStateTransformation interface {
+	// Process will process the TableView.
+	Process(chunk table.Chunk, state interface{}, d *TransportDataset, mem memory.Allocator) (interface{}, bool, error)
+}
+
+var _ Transport = (*narrowStateTransformation)(nil)
+var _ Transformation = (*narrowStateTransformation)(nil)
+
+type narrowStateTransformation struct {
+	t NarrowStateTransformation
+	d *TransportDataset
+}
+
+// NewNarrowStateTransformation constructs a Transformation and Dataset
+// using the NarrowStateTransformation implementation.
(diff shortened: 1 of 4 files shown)


scbrickley (commented):
This all seems good, but I'm curious if this is the kind thing we should put behind a feature flag. What do you think?

jsternberg:
This doesn't modify any transformations so this is mostly just infrastructure code. When we modify a transformation, I think a feature flag will be appropriate.

scbrickley:
Ah okay, now I see. Nothing actually implements the new interface yet so it's not being used. My mistake.

scbrickley (approved):
LGTM

PR #4038, opened 2021-09-14
revert: update to rust 1.55 (#4027)

scbrickley:
This reverts commit b899d8bd6a35797d52964c71a21d192ceda40611.

The upgrade to rust 1.55 was causing build errors in the flux release for influxdb. We've confirmed that downgrading to rust 1.54 resolves the issues: https://app.circleci.com/pipelines/github/influxdata/influxdb/24024/workflows/09396d1f-5b38-4c4c-ba55-a62006c9cfe8

### DIFF
--- a/Dockerfile_build
+++ b/Dockerfile_build
@@ -4,7 +4,7 @@
 # and verification, we can list the rust container as a prior build stage, and
 # then pull in the artifacts we need. There is an added benefit that tagged versions
 # also include minor releases, so 1.2 includes 1.2.1 and so on, for bugfix releases.
-FROM rust:1.55 as RUSTBUILD
+FROM rust:1.54 as RUSTBUILD
 
 FROM golang:1.17
 


jsternberg (approved):

rockstar (commented):
I think I would rather see if we can fix the influxdb build issues. What would it take to do that?

jsternberg:
@rockstar my guess is that something changed in rust or in the build toolchain that now `-lm` is required when linking. First, I'd like to confirm this is the case that something changed. Second, we would then just add `-lm` to `pkg-config` as required to build Flux. Or, we would figure out how to modify a compilation option so that libmath isn't used at all.

I, unfortunately, have not been able to identify why `-lm` might suddenly be required with rust 1.55.

rockstar:
Cloud environment builds are working, right? The nightly would indicate yes, but want to confirm. Just influxdb is broken? History would tell me it's related to cross compile. I would rather do the release without passing the tests in influxdb, and then make a branch that shepherds influxdb's update through.

If this is the easiest thing, I'm good with that, but I'd rather deal with the pain while we're experiencing it rather than to kick the can down the road.

jsternberg:
The primary reason seems to be that the cloud environment pinned the rust version to 1.52 to avoid this. So they're not running into it because they aren't using rust 1.55. The issue seems to be related to rust 1.55 and @scbrickley and I were able to reproduce without cross compilation being involved. It _might_ be a Linux only issue though as he is on Linux.

scbrickley:
Created a new issue here to track the work needed to upgrade back to 1.55: https://github.com/influxdata/flux/issues/4039

PR #4040, opened 2021-09-15
fix(array): add null support to optimized repeat function

skartikey:
This change is to add null support to optimized repeat function for arrow arrays
Also did some refactoring of the existing code

This was caught with the bug linked to this issue, aggregateWindow()
function with 'createEmpty: true' was returning empty tables which were
generating rows with null values when piped with group and sort(test
case added)

fixes: https://github.com/influxdata/idpe/issues/11589

### DIFF
--- a/arrow/repeat.go
+++ b/arrow/repeat.go
@@ -2,17 +2,54 @@ package arrow
 
 import (
 	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/values"
 )
 
 // Repeat will construct an arrow array that repeats
 // the value n times.
-func Repeat(v values.Value, n int, mem memory.Allocator) array.Interface {
-	switch v := values.Unwrap(v).(type) {
-	case values.Time:
-		return array.IntRepeat(int64(v), n, mem)
+func Repeat(colType flux.ColType, v values.Value, n int, mem memory.Allocator) array.Interface {
+	switch colType {
+	case flux.TInt:
+		var ival int64
+		if !v.IsNull() {
+			ival = v.Int()
+		}
+		return array.IntRepeat(ival, v.IsNull(), n, mem)
+	case flux.TUInt:
+		var uival uint64
+		if !v.IsNull() {
+			uival = v.UInt()
+		}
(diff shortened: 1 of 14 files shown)


arrow/repeat.go
```diff
@@ -2,17 +2,54 @@ package arrow
...
 // Repeat will construct an arrow array that repeats
 // the value n times.
-func Repeat(v values.Value, n int, mem memory.Allocator) array.Interface {
-	switch v := values.Unwrap(v).(type) {
-	case values.Time:
-		return array.IntRepeat(int64(v), n, mem)
+func Repeat(colTyp flux.ColType, v values.Value, n int, mem memory.Allocator) array.Interface {
+	switch colTyp {
+	case flux.TInt:
+		var IVal int64
```
jsternberg, line 17: Decapitalize these such as `ival`.

arrow/repeat.go
```diff
@@ -2,17 +2,54 @@ package arrow
...
 )
 
 // Repeat will construct an arrow array that repeats
 // the value n times.
-func Repeat(v values.Value, n int, mem memory.Allocator) array.Interface {
-	switch v := values.Unwrap(v).(type) {
-	case values.Time:
-		return array.IntRepeat(int64(v), n, mem)
+func Repeat(colTyp flux.ColType, v values.Value, n int, mem memory.Allocator) array.Interface {
+	switch colTyp {
```
jsternberg, line 15: `colType`

result.go
```diff
@@ -183,6 +183,7 @@ type GroupKey interface {
 	Cols() []ColMeta
 	Values() []values.Value
 
+	Col(j int) ColMeta
```
jsternberg, line 186: I would prefer we didn't modify this interface since it is already quite large. We might end up refactoring this in the future. Can you change the areas where this is used to be `key.Cols()[j]` instead?

stdlib/universe/sort2_test.flux
```diff
@@ -0,0 +1,41 @@
+package universe_test
+
+
+import "csv"
+import "testing"
+
+option now = () => 2030-01-01T00:00:00Z
+
+testcase sort_panic {
```
jsternberg, line 9: Can you rename this to something that describes what is being tested? Maybe something like `sort_with_null_columns`?

jsternberg (approved):
I'd like to see a few changes, but I'm going to approve it because once those changes are made this is good to merge and I don't want to force you to wait until I'm back at the computer tomorrow.

PR #4041, opened 2021-09-15
feat(record): adds any value to record package

nathanielc:
Prior to this change it was not possible to use a default value that had
a record type because default values constrain the type of the function
parameter. See #3461 for a long term fix. In the medium term a valid
solution is to introduce a polymorphic record value that can be used in
these cases. This change adds record.any which is an empty record
but whose type is unbounded allowing for polymorphic values.

The package has been placed into experimental as we expect to either remove this once #3461 has been fixed or if we decide this has other uses to promote it out of experimental. 

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/internal/record/record.flux
```diff
@@ -0,0 +1,5 @@
+package record
+
+
+// unbounded is a record that at a minumum contains no properties but that may contain any additional properties.
+builtin unbounded : A where A: Record
```
nathanielc, line 5: Rename to `any`.

rockstar (approved):

PR #4044, opened 2021-09-16
revert: "feat: embed compiled stdlib instead of compiling at runtime (#3623)"

jsternberg:
This reverts commit 9b9575705d78fe818e2d51b3318c2dd32f483e35.

Because of the way that type inference uses the fresher when analyzing
files, a change to one package would cause any packages that were
lexigraphically lower in the chain to also be rewritten because the
tvars were now different.

As this causes a rewrite of most files when changing flux code, this is
being reverted until type inference is modified to use a normalized
fresher for each file rather than sharing the one from the bootstrap
process.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
Is the result of the current fresher usage that it causes lots of noise in git? If so I _might_ be ok with this change staying in place while we fix the fresher issue in the meantime. The value of this change seems to be much large the the downside. But I may not fully understand the issue at hand.

jsternberg:
Yea, it causes a lot of noise in git. If you make a change to any file in the standard library, it rewrites every file in lexical order. I think the fix should be decently simple, but I haven't thought of downstream implications.

I think it should be possible to analyze every file with a newly initialized fresher and rely on the import system to instantiate generalized functions from imports when they are used. But, I would need to research more first.

nathanielc:
ok lets revert and then figure out a way forward. Can you create an issue for the fresher and we can groom it on Monday?

nathanielc (approved):

PR #4048, opened 2021-09-16
chore(checkfmt): update checkfmt to print all files

nathanielc:
I found that some files in the stdlib were not formatted. This makes a few small fixes/updates to the checkfmt workflow.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


etc/checkfmt.sh
```diff
@@ -12,39 +13,39 @@ for file in $(go list -f '{{$dir := .Dir}}{{range .GoFiles}}{{printf "%s/%s\n" $
...
-    echo 'Commit includes files that are not gofmt-ed' && \
-    echo 'run "make fmt"' && \
-    echo ''
-fi
-
 (
-  cd libflux || exit
-  FMT_OUT="$(cargo fmt --all -- --check)"
-  if [[ -n "$FMT_OUT" ]]; then
+    echo "Checking Rust source files..."
```
jsternberg, line 24: Indentation seems to be a bit weird here.

jsternberg (approved):

etc/checkfmt.sh
nathanielc: This bit was previously wrapped in `()` so I just left that part alone.

jsternberg: The weird part is that `echo` is 4 spaces and `cd` is 2 spaces of indentation.

PR #4049, opened 2021-09-17
feat(experimental/record): add get method

alespour:
This PR adds `get` method, it returns record field or provided default value. Per Nathaniel's [comment](https://github.com/influxdata/flux/pull/3987#issuecomment-926182962), it should be documented as experimental and likely to change in the future:

### Function `get`

Returns record field identified by a `key` (must be a string literal) or `default` value if no such field exists in the record `r`.

This function is experimental and intended to be used as a workaround for limited use of `exists` operator with non-table records. Once the operator applicability is extended, this function will most certainly be changed or even removed.
The `key` parameter must be a string literal.

Example:
```
import "experimental/record"

obj = { 
   ticks: 11
}

record.get(r: obj, key: "ticks", default: 0)
11

record.get(r: obj, key: "delay", default: -1)
-1
```

_Maybe it should even not be mentioned in the official documentation?
It is commented with the above information in the function Flux code as well._

### Done checklist
- [x] Test cases written


alespour:
I had to fix formatting of flux test files unrelated to this PR in commit f073080.

alespour:
Unfortunately I have no idea how to check if `key` value was a string literal, @nathanielc, could you please give me a hint? Thank you.

nathanielc:
@alespour Good question, there isn't currently a way to enforce its a string literal. For now let's just call that out in the docs of the function.

alespour:
@nathanielc Thank you. I've updated the PR comment with the function docs paragraph. Ready for review now I hope.

stdlib/experimental/record/record.flux
```diff
@@ -3,3 +3,8 @@ package record
 
 // any is a record that contains no properties but according to its type may contain any additional properties.
 builtin any : A where A: Record
+
+// get returns record field identified by a key (must be a string literal) or default if no such field exists in the record.
+// This is a temporary solution for `exists` operator limited use with non-table records,
+// and will almost certainly be removed/changed in the future.
+builtin get : (r: A, key: string, default: B) => B where A: Record
```
nathanielc, line 10: Could you add a link to this issue here https://github.com/influxdata/flux/issues/4073 ? That way we can track the progress on changing/removing this in the future.

nathanielc (changes requested):
One small change and this is good to go.

nathanielc (approved):

nathanielc:
I'll merge once the PR has been rebased of most recent master.

alespour:
Done, thank you.

PR #4052, opened 2021-09-20
fix(sql): disallow setting allowAllFiles parameter in MySQL DSNs

skartikey:
By abusing support for MySQL's LOAD DATA LOCAL INFILE statement in the golang MySQL driver we use, an attacker can use Flux's sql.from() to leak the contents of local files on Cloud 2's queryd-pull-external pods.

The golang MySQL driver we use has no specific facility for disabling INFILE support in code. We can, however, reject connection strings that set allowAllFiles=true in stdlib/sql/source_validator.go#L30 since we already parse it into a configuration struct.

fixes: https://github.com/influxdata/idpe/issues/11716

### DIFF
--- a/stdlib/sql/source_validator.go
+++ b/stdlib/sql/source_validator.go
@@ -4,7 +4,7 @@ import (
 	neturl "net/url"
 	"strings"
 
-	"github.com/bonitoo-io/go-sql-bigquery"
+	bigquery "github.com/bonitoo-io/go-sql-bigquery"
 	"github.com/go-sql-driver/mysql"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/dependencies/url"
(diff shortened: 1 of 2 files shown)


stdlib/sql/source_validator.go
```diff
@@ -4,7 +4,7 @@ import (
 	neturl "net/url"
 	"strings"
 
-	"github.com/bonitoo-io/go-sql-bigquery"
+	bigquery "github.com/bonitoo-io/go-sql-bigquery"
```
rockstar, line 7: Did we need this alias? Or is it just a holdout from development?

rockstar (approved):
This looks good, thanks for grabbing it. Just had one small comment, but nothing to block on.

stdlib/sql/source_validator.go
skartikey: This import alias got added by the vscode go plugin, maybe to leverage this to bypass package name conflicts in the future.

PR #4055, opened 2021-09-20
perf: Avoid cloning in Parser::peek

Marwes:
Reduces the runtime of the benchmark on `everything.flux` by ~25%.

Just poking around trying to understand the structure of the project(s), noticed that a lot of time were spent cloning `Token` (which is a rather large and expensive to clone struct).

Seeing some (unrelated) test failures locally but haven't understood why those happen yet

### DIFF
--- a/libflux/flux-core/src/parser/mod.rs
+++ b/libflux/flux-core/src/parser/mod.rs
@@ -74,7 +74,7 @@ impl Parser {
     }
 
     // peek_with_regex is the same as peek, except that the scan step will allow scanning regexp tokens.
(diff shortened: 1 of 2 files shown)


nathanielc:
@Marwes I took a quick look and at first glance the changes look good. When you are ready for a review assign the review to `influxdata/flux-team` that will randomly pick someone on the team to review. 

image

Marwes:
No idea what `make checkgenerate` does and why it fails now :S .

wolffcm (approved):
Nice work, you'll just want to force push so the most recent commit is semantically formatted.

PR #4066, opened 2021-09-22
fix: Propagate all inferred properties to a function argument

### LINKED ISSUE #4051: Type inference does not infer all properties of a record via multiple member expressions
This query causes this panic:

```
from(bucket: "bucket")
    |> range(start: -10m)
    |> filter(fn: (r) => r._measurement == "m")
    |> map(
        fn: (r) => {
            a = r.a
            b = r.b
            c = r._value + 1
            return r
        },
    )
```

Note that `r.a` and `r.b` do not exist on the data. The `r._value` column does exist and it is a float.

Stacktrace and panic

```
panic: unexpected kind: got "float" expected "int", 
trace: goroutine 5107 [running]: runtime/debug.Stack() 
/usr/local/go/src/runtime/debug/stack.go:24 +0x65 github.com/influxdata/flux/values.unexpectedKind.As({0x0, 0x5}, {0x370eb60, 0xc0049bb538})
/idpe/vendor/github.com/influxdata/flux/values/values.go:347 +0x65 github.com/influxdata/flux/values.unexpectedKind.Error({0xc02e148c30, 0x0})
/idpe/vendor/github.com/influxdata/flux/values/values.go:341 +0x2e github.com/influxdata/idpe/pkg/influxerror.(*Error).Error(0xc0053c8840)
/idpe/pkg/influxerror/errors.go:111 +0x19f github.com/influxdata/idpe/query/queue.(*Handler).handleError(0xc004740360, {0x3c1e0b8, 0xc0050d2660}, {0x3bd3880, 0xc0053c8840}, {0x3c1fcb8, 0xc0001211e0})
/idpe/query/queue/worker.go:179 +0x354 github.com/influxdata/idpe/query/queue.(*Handler).Handle(0xc004740360, {0x3c1e0b8, 0xc00510b050}, {0x3c1fcb8, 0xc0001211e0}, 0xc00510c640)
/idpe/query/queue/worker.go:150 +0x9bb github.com/influxdata/idpe/pkg/queue/worker.(*Dispatcher).handle(0xc0001415e0, {0x3c1e0b8, 0xc00510b050}, 0xc0001211e0, 0x12)
/idpe/pkg/queue/worker/redis_dispatcher.go:382 +0x282 github.com/influxdata/idpe/pkg/queue/worker.(*Dispatcher).runOnce(0xc0001415e0, {0x3c1e010, 0xc0019f5780})
/idpe/pkg/queue/worker/redis_dispatcher.go:353 +0xf96 github.com/influxdata/idpe/pkg/queue/worker.(*Dispatcher).stateRun(0xc0001415e0, {0x3c1e010, 0xc0019f5780})
/idpe/pkg/queue/worker/redis_dispatcher.go:238 +0x5d github.com/influxdata/idpe/pkg/queue/worker.(*Dispatcher).Run(0xc0001415e0, {0x3c1e010, 0xc0019f5780})
/idpe/pkg/queue/worker/redis_dispatcher.go:398 +0x186 github.com/influxdata/idpe/internal/cmd/queryd.runQueryd.func4()
/idpe/internal/cmd/queryd/queryd.go:821 +0x2e created by github.com/influxdata/idpe/internal/cmd/queryd.runQueryd
/idpe/internal/cmd/queryd/queryd.go:820 +0x3912   
```

I experimented with this query and almost any change to this query causes the panic to go away. 

For example removing any assignment statement, or changing the `r.a` and `r.b` to be the same column both cause the panic to go away. My guess is we have some kind of collision where type vars are getting mixed up. Perhaps in the flatbuffer serialization or somewhere in type inference.

Marwes:
Each function arguments gets a fresh type variable assigned to them
to which a type gets inferred to in the function body. This then does not
get propagated to the type (and constraint) created for the function
itself as the substitution(s) are not returned to the
`FunctionExpr::infer` call-site. Thus the argument `r` gets assigned
the type variable `X`, then the body is checked and `X` is inferred to
have the field `c`, however the substitution where this is the case only exists
within the binding which uses `c` and is never applied on the constraint created for the function itself.

This is a hack that gets the test passing and I don't know if this is
actually sound, there is probably a better solution for this out there
with some help.

Closes #4051

### Done checklist
- [x] Test cases written


libflux/flux-core/src/semantic/tests.rs
```diff
@@ -3363,6 +3363,28 @@ fn function_default_arguments_and_pipes() {
...
+        f = (r) => {
+            a = r.a
+            b = r.b + 1
+            c = r.c * 1.0
+            return r
+        }
+        x = f(r:r)
+        "#,
+        exp: map![
+            "f" => "(r:{S with a:X, b: int, c: float) => {S with a:X, b: int, c: float}",
```
Marwes, line 3382:
```suggestion
            "f" => "(r:{S with a:X, b: int, c: float}) => {S with a:X, b: int, c: float}",
```
I assume that there is a `}` missing here, but it still seems to parse?

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -867,6 +867,10 @@ impl FunctionExpr {
         }
         // And use it to infer the body.
         let (nenv, bcons) = self.body.infer(nenv, f)?;
+        // Update the argument variable with the types inferred from the body
+        for (k, t) in req.iter_mut() {
+            *t = nenv.lookup(k).unwrap().expr.clone();
+        }
```
wolffcm, line 873: I think you need to do the same with the optional arguments as well as the required ones. I can think of an example where I think this would matter:
```js
apply = (a=(v) => v, b) => {
  f = a
  return f(v: b) + 1
}
```
In this case the type of parameter `a` would be `(v: X) => X` without considering uses in the body, but we need to further constrain the type of `a` to be `(v: int) => int`. Thoughts?

wolffcm (commented):
This looks good, I just had a question about the optional parameters.

Marwes:
This is actually an issue with just 2 bindings as well. Notice that the row extension variables do not match on the argument and return.

```rust
#[test]
fn issue_4051_1() {
    test_infer! {
        env: map![
            "r" => "A where A: Record",
        ],
        src: r#"
        f = (r) => {
            a = r.a
            b = r.b + 1
            return r
        }
        x = f(r:r)
        "#,
        exp: map![
            "f" => "(r:{S with a:X, b: int, c: float}) => {S with a:X, b: int, c: float}",
            "x" => "{S with a: X, b: int, c: float}",
        ],
    }
}
```

```
unexpected types:

want:
        f: (r:{A with c:float, b:int, a:B}) => {A with c:float, b:int, a:B}
        x: {A with c:float, b:int, a:B}

got:
        f: (r:{t16 with a:G, b:int}) => {t15 with a:G, b:int}
        x: {t19 with a:t18, b:int}
```

So the problem happens as soon as rule 5 (unifying two rows with distinct labels) is applied. It creates a new variable, unifies the records and returns the substitution which in the end updates the environment. This is all done correctly except the substitution is the only place that encoded the unification between these two records so as soon as that is discarded after the solve we lose the connection.

As the substitution is applied to the environment we do __carry the result forward__ , it is only the application of rule 5 that is lost. Since all constraints are re-examined again at the top level we might have been able to still compute the correct result, however as that pass will apply rule 5 separately it will create a different, fresh type variable from the variable that were encoded in the environment.

Marwes:
Added optional arguments to the list to update as well.

While this is a hack, I don't see another quick fix for this issue. Even if this causes some other issues in the type system I would guess it probably fixes more things than it breaks so merging this may be good until a proper fix can be made.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -867,6 +867,15 @@ impl FunctionExpr {
         }
         // And use it to infer the body.
         let (nenv, bcons) = self.body.infer(nenv, f)?;
+        // HACK Remove once substitutions are persisted correctly
+        // Update the argument variable with the types inferred from the body
+        for (k, t) in req.iter_mut().chain(opt.iter_mut()) {
+            *t = nenv
+                .lookup(k)
+                .expect("The binding should have been added above")
```
wolffcm, line 875: Does this panic? Unfortunately we cannot panic in this code, since it's invoked from Go, and (iirc) it can cause the server process to exit.

wolffcm:
> I would guess it probably fixes more things than it breaks 

Can you quantify the things that it might break? Is there a possibility that Flux scripts that work now would not after this change?

libflux/flux-core/src/semantic/nodes.rs
Marwes: It could, of course, but only due to bug elsewhere as the binding that it refers to were defined above in this same function.

Marwes:
> Can you quantify the things that it might break?

Not really, just unknown-unknowns.

> Is there a possibility that Flux scripts that work now would not after this change?

Everything should still work, but still, unknown-unknowns.

nathanielc (approved):
After looking at our other options I think we should go ahead with this change. We can possibly revisit later if we decide to change the way we handle substitutions.

libflux/flux-core/src/semantic/nodes.rs
wolffcm: I think we should still avoid uses of `expect` in this code, and return an error. I know it's a pain, but if we never call code that panics (and I don't think we do in our code, save for 3rd party code) then we don't have to worry about taking down the whole process.

wolffcm (approved):

Issue #4030, opened 2021-09-10
Create a `vector` monotype


wolffcm:
In order to facilitate vectorization, we need a new kind of monotype called `vector`. It will be similar to the `array` monotype, but expressions with type `vector` will only be able to be created by a vectorization transformation.

One significant difference between `array` and `vector` is that `vector` should be able to satisfy Kind constraints by deferring to its element type. In other words, `int` satisfied `Addable`, so that means that `vector(int)` should also satisfy addable.

The new monotype would be added here:
https://github.com/influxdata/flux/blob/b899d8bd6a35797d52964c71a21d192ceda40611/libflux/flux-core/src/semantic/types.rs#L269

The new type should be implemented in both Go and Rust, and also there should be corresponding changes to flatbuffers serialization and deserialization to go between the two.

DOD:
- A new `vector` monotype exists in both Go and Rust
- This type can be serialized to and from FlatBuffers
- Unit tests demonstrate that this type behaves as expected, especially with regards to `unify` and `constrain` methods
  - E.g., unit tests should demonstrate that `vector(float)` satisfied kind constraint `Divisible` but that `vector(string)` does not
- There should be unit tests for serialization/deserialization that cross the Go/Rust boundary

PR #4067, opened 2021-09-22
feat(Vectorize): create a vector monotype

skartikey:
Added a new vector monotype in both Go and Rust
    
This type can be serialized to and from FlatBuffers
    
Unit tests were added to demonstrate that this type behaves as expected, especially with regards to unify and constrain methods
E.g., unit tests should demonstrate that vector(float) satisfied kind constraint Divisible but that vector(string) does not
    
Unit tests for serialization/deserialization that cross the Go/Rust boundary
    
fixes: #4030


ast/ast.fbs
```diff
@@ -116,6 +122,7 @@ union Expression {
   UnaryExpression,
   UnsignedIntegerLiteral,
   BadExpression,
+  VectorExpression,
```
wolffcm, line 125: You don't need to add a vector expression to the flatbuffer schema. This work should only be about creating the `vector` type.

ast/ast.fbs
```diff
@@ -374,6 +381,11 @@ table ArrayExpression {
   elements:[WrappedExpression];
 }
 
+table VectorExpression {
+  base_node:BaseNode;
+  elements:[WrappedExpression];
+}
+
```
wolffcm, line 388: Same comment as above. I'll give some explanation:

For arrays, we let users in Flux create an array using `[]`, as in `[1, 2, 3]` would be an `ArrayExpression` that has type `Array(int)`, or in our type syntax, the type would be expressed as `[int]`.

For for vectors, we don't want users to be able to create them. We want a special transformation of semantic graph function expressions to create them automatically for uses of `map` and `filter` in a user's script.

Most often, vectors will be created as part of a functions signature.  So instead of a function accepting a record type where `r` has scalar fields:
- E.g., `map(fn: (r) => ({a: r.a + 1}))`, where the type of `r` is `{a: int}`, in other words, "a record with a field called `a` whose type is `int`"
- The type of `r` will be changed to `{a: vector(int)}`, in other words, "a record with a field called a whose type is `vector` of `int`"

So there's no need to create an actual vector expression, we'll assume that the caller of the function being transformed will create the vector.

(There will be a need for a new expression called `expand` which will be able to create a `vector` out of a scalar value, but that is later work)

ast/ast.fbs
```diff
@@ -24,6 +25,11 @@ table ArrayType {
   element:MonoType;
 }
 
+table VectorType {
+  base_node:BaseNode;
+  element:MonoType;
+}
+
```
wolffcm, line 32: Actually I don't think support for `VectorType` is needed in AST at all.

In AST, we use types only as part of Flux syntax for `builtin` statements, which declare the type of a thing that is actually implemented in Go (not Flux).

For example, on this line everything after `:` is the type:
https://github.com/influxdata/flux/blob/master/stdlib/universe/universe.flux#L15

However, for `vector` so far we don't expect to need to declare any builtins that use a `vector` type.  So you don't need to do anything with AST for this PR. It can be completely about create a new Vector monotype for semantic graph only.

internal/fbsemantic/semantic.fbs
```diff
@@ -255,6 +261,12 @@ table ArrayExpression {
   typ:MonoType;
 }
 
+table VectorExpression {
+  loc:SourceLocation;
+  elements:[WrappedExpression];
+  typ:MonoType;
+}
+
```
wolffcm, line 269: Again here as well, there is no need to create a `VectorExpression` node, same reason as the comment about AST.

libflux/flux-core/src/semantic/flatbuffers/types.rs
```diff
@@ -680,6 +709,10 @@ mod tests {
         test_serde("[A]");
     }
     #[test]
+    fn serde_vector_type() {
+        test_serde("[A]");
```
wolffcm, line 713: This test uses the type parser to create a monotype and roundtrip it. You would need to add support to the parser for vector types to use it. We should talk about whether you should do this, or just test flatbuffers serialization in an ad hoc way.

libflux/flux-core/src/semantic/flatbuffers/types.rs
```diff
@@ -19,6 +19,7 @@ use crate::semantic::types::{
     Record,
     Tvar,
     TvarKinds,
+    Vector,
```
wolffcm, line 22: So far I've been making comments that say, "you don't actually need this in the PR", so I just wanted to say here, that this code is good and it should be here :)

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -130,6 +130,7 @@ pub enum Expression {
     Boolean(BooleanLit),
     DateTime(DateTimeLit),
     Regexp(RegexpLit),
+    Vector(Box<VectorExpr>),
```
wolffcm, line 133: Again here, there is no need to add a new `Vector` expression to the semantic graph nodes.

libflux/flux-core/src/semantic/types.rs
```diff
@@ -294,6 +294,9 @@ pub enum MonoType {
     Record(Box<Record>),
     #[display(fmt = "{}", _0)]
     Fun(Box<Function>),
+    #[display(fmt = "{}", _0)]
+    Vector(Box<Vector>),
+
```
wolffcm, line 299: 👍

libflux/flux-core/src/semantic/types.rs
```diff
@@ -686,6 +695,52 @@ impl Array {
...
+        self.0.unify(with.0, cons, f)
+    }
+
+    fn constrain(self, with: Kind, cons: &mut TvarKinds) -> Result<Substitution, Error> {
+        match with {
+            Kind::Equatable => self.0.constrain(with, cons),
+            _ => Err(Error::CannotConstrain {
+                act: MonoType::Vector(Box::new(self)),
+                exp: with,
+            }),
```
wolffcm, line 735: The definition of `constrain` for `Vector` should be different from `Array`. Where as array only allows the `Equatable` constraint, we want `Vector` to allow any constraints to apply to it's element type.

In other words, if we have `a + b` and both `a` and `b` have type `vector(int)`, we want to be able to say that `vector(int)` is `Addable` because `int` is `Addable`. In the proof of concept I did it like this:
https://github.com/influxdata/flux/blob/349666b85b2fd5a7ffee180002832cacbfb183a1/libflux/flux-core/src/semantic/types.rs#L731

libflux/flux-core/src/semantic/types.rs
```diff
@@ -1469,6 +1524,13 @@ mod tests {
         );
     }
     #[test]
+    fn display_type_vector() {
+        assert_eq!(
+            "[int]",
```
wolffcm, line 1529: This is the same as the display for `array`, it should be different. Maybe `{{int}}` or `v[int]`? I am just throwing ideas out there.

wolffcm (commented):
I think my issue was too vague and as a result there was some confusion. Really, the changes needed are much less that what you've done.

libflux/flux-core/src/semantic/types.rs
skartikey: I have implemented display as v[int]

skartikey:
@wolffcm thank you for all the insightful comments. I have removed all the vectorExpression bits from the code and kept the monotype in semantic only.

libflux/flux-core/src/semantic/flatbuffers/types.rs
skartikey: Sure, let's discuss this when you have some time today. Thanks

wolffcm (approved):
This looks great.

PR #4069, opened 2021-09-22
refactor(aggregate): Allow aggregates to recycle state

### LINKED ISSUE #4042: Refactor quantile to only recycle memory when it calculates a complete result
In order to cut down on memory use, the flux `quantile` transformation was changed so that it recycled memory between results (see #3497 for details).

However, a problematic interaction between `group` and `quantile` manifests when enabling chunk processing for both, resulting in any results from `quantile`, or any transformation that uses it under the hood (e.g. `median`) to produce incorrect results.

This is due to the fact that, with chunk-based processing, a group node can potentially output a chunk from one group, followed immediately by a chunk from another group. In contrast, old group nodes would pass along an entire group before moving on to the next one. So, because `quantile` recycles memory when it gets data from a new group, it sometimes drops important data before it can calculate a complete result.

A temporary fix for this was introduced in #4035, which prevented `quantile` from using chunk-based processing by passing it an empty context, free of any feature flags.

The goal of this issue is to modify the quantile transformation so that it still uses memory efficiently, but can maintain state between different results without dropping important data.

scbrickley:
Closes #4042

This patch adds an interface called `Recyclable`, which marks an aggregate as one that can re-use state in order to reduce memory consumption. This resolves an issue with `quantile` and the new narrow aggregate transport that was causing aggregate states to be reset before a full result could be calculated.

There are still a few concerns I have with this patch.

~~First, I had to squash a bug that resulted in some aggregates reusing state erroneously. I've left a comment in the code with more details, and I could use some help diagnosing the issue and coming up with a cleaner fix.~~ Edit: a slightly better fix (better in the sense that it's easier to understand what the problem is) was added in commit daac803ee3223003a29ce68b8d5f3d06a36a8ef0 (see the `isRecyclable` method)

~~Second, state is still getting recycled for the quantile transformation when `NewFloatAgg()` is called. I'm not sure whether or not this is an issue, but it seems potentially problematic since it's a point where state is recycled outside of the code path where I included my fix. I would like a second opinion on whether or not it needs to change.~~

***Benchmarks for quantile after applying this patch***
*compare with the benchmarks found in #3497*
```
go test -bench=Quantile$ -run=^Benchmark -benchmem ./stdlib/universe
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkQuantile-12    	       7	 147663543 ns/op	   25120 B/op	       7 allocs/op
PASS
ok  	github.com/influxdata/flux/stdlib/universe	1.513s
```

### DIFF
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -104,24 +104,6 @@ func (t *aggregateTransformation) Process(id DatasetID, tbl flux.Table) error {
 
 func (t *aggregateTransformation) processChunk(chunk table.Chunk) error {
 	state, _ := t.d.Lookup(chunk.Key())
-	// XXX: sean (Oct 7, 2021) - I am not entirely sure why, but checking `ok`
-	// here before setting `state` to the return value of `popFreeState()` avoids
-	// some major bugs. Without it, some aggregates produce wrong answers, and tests
-	// start to fail.
-	//
-	// My hunch is that while `multiState` and `aggregateState` implement
-	// `Recyclable`, the underlying aggregate may not. So we append states to
-	// `t.freeStates` (see the `flushKey` function) in all circumstances, even
-	// if the underlying aggregate cannot be recycled. This causes some aggregates
-	// (sum in particular) to re-use state erroneously.
-	//
-	// Still, I don't know why checking if it's recyclable here solves the problem.
-	// If the state wasn't recyclable, t.freeStates should be empty anyway, so this check
-	// shouldn't make a difference.
-	//
-	// Also, if `state` is nil, wouldn't `ok` always be false? In which case, the below
-	// `if` block would never trigger, which means that the fix would not work.
-	// Something strange is going on here.
 	if state == nil {
 		// If no reusable state is available, state will remain nil
 		state = t.popFreeState()
@@ -450,6 +432,11 @@ func (a *aggregateState) Drop() {
 	}
 }
 
+// multiState is an alias for []aggregateState so that we can
+// define methods on it. It is the type that's returned from
+// the group key lookup on the transport dataset inside `aggregateTransformation`.
+//
+// It contains one aggregateState for every column in the table chunk.
 type multiState []aggregateState
 
 func (m multiState) Recycle() {
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -45,16 +45,18 @@ type AggregateTransformation interface {
 }
 
 type aggregateTransformation struct {
-	t AggregateTransformation
-	d *TransportDataset
+	t          AggregateTransformation
+	d          *TransportDataset
+	freeStates []interface{}
 }
 
 // NewAggregateTransformation constructs a Transformation and Dataset
 // using the aggregateTransformation implementation.
 func NewAggregateTransformation(id DatasetID, t AggregateTransformation, mem memory.Allocator) (Transformation, Dataset, error) {
 	tr := &aggregateTransformation{
-		t: t,
-		d: NewTransportDataset(id, mem),
+		t:          t,
+		d:          NewTransportDataset(id, mem),
+		freeStates: make([]interface{}, 10),
 	}
 	return tr, tr.d, nil
 }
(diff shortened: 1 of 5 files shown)


execute/aggregate.go
```diff
@@ -102,6 +104,29 @@ func (t *aggregateTransformation) Process(id DatasetID, tbl flux.Table) error {
...
+	// shouldn't make a difference.
+	//
+	// Also, if `state` is nil, wouldn't `ok` always be false? In which case, the below
+	// `if` block would never trigger, which means that the fix would not work.
+	// Something strange is going on here.
+	_, ok := state.(Recyclable)
+	if ok && state == nil {
+		// If no reusable state is available, state will remain nil
+		state = t.popFreeState()
+	}
```
nathanielc, line 129: Here is how I understand whats going on.

First, in the flushKey function below we only adds states to the freeStates list of its recyclable. 
Second, as your comment clarifies the t.popFreeState() function will return `nil` if there are no free states.

My hunch here is that we are overriding a known good state with `nil` and thus getting wrong answers.

This is how I would expect this code to work:

```go
state, ok := t.d.Lookup(chunk.Key())
if !ok {
    // We did not find a previous state, therefore we should try and resuse a cached one if it exists.
    // If no reusable state is available, state will remain nil.
    // If state remains nil, a new state will be created below.
    state = t.popFreeState()
}
// the rest of the function, no need to check if the state is recyclable here.
```

execute/aggregate.go
```diff
@@ -45,16 +45,18 @@ type AggregateTransformation interface {
...
 
 // NewAggregateTransformation constructs a Transformation and Dataset
 // using the aggregateTransformation implementation.
 func NewAggregateTransformation(id DatasetID, t AggregateTransformation, mem memory.Allocator) (Transformation, Dataset, error) {
 	tr := &aggregateTransformation{
-		t: t,
-		d: NewTransportDataset(id, mem),
+		t:          t,
+		d:          NewTransportDataset(id, mem),
+		freeStates: make([]interface{}, 10),
```
nathanielc, line 59: This might also be causing the bug. When making a new slice with a capacity you need to specify both the initial length and desired capacity of the slice like so `make([]type, length, capacity)`. If you leave off the second value it behaves as if both length and capacity are the same. 

As a result we are creating a slice of states that starts with 10 entries all of which are nil. Since we only add new states if `len() < cap()` we only ever get to use one state slot (since we pop first before pushing).

Using `make([]interface{}, 0, 10)` will create an empty slice with capacity for 10 values.

execute/aggregate.go
```diff
@@ -369,12 +421,53 @@ type aggregateState struct {
...
+	}
+}
+
+func (a *aggregateState) Drop() {
+	if agg, ok := a.agg.(Recyclable); ok {
+		agg.Drop()
+	}
+}
+
+type multiState []aggregateState
```
nathanielc, line 451: Its not clear to me how multiState is used. Can you add a comment explaining its behavior?

execute/aggregate.go
```diff
@@ -369,12 +421,53 @@ type aggregateState struct {
...
+func (m multiState) Drop() {
+	for _, state := range m {
+		state.Drop()
+	}
+}
+
+func (t *simpleAggregateTransformation2) initializeState(chunk table.Chunk, current interface{}) (multiState, error) {
 	if current != nil {
-		return current.([]aggregateState), nil
+		return current.(multiState), nil
```
nathanielc, line 467: What are the multiple states being tracked? I assumed there would be one state per table key. Are there more?

execute/aggregate.go
scbrickley: I could introduce a function that detects whether a multiState contains any non-recyclable items. Running a check on that before adding/removing states may resolve the problem?

scbrickley: For reference, this commit contains an alternate fix: https://github.com/influxdata/flux/pull/4069/commits/daac803ee3223003a29ce68b8d5f3d06a36a8ef0

stdlib/universe/quantile_agg_test.flux
```diff
@@ -0,0 +1,65 @@
...
+        |> group(columns: ["t0"])
+
+    got = inData
+        |> range(start: 2018-05-22T20:00:00Z, stop: 2018-05-22T20:01:00Z)
+        |> group(columns: ["t0"])
+        |> quantile(q: 0.75, method: "estimate_tdigest")
+        |> drop(columns: ["_start", "_stop"])
+
+    testing.diff(want: want, got: got) |> yield()
+}
```
wolffcm, line 50: Does this test case actually reproduce the issue? My understanding was that this bug occurs when there chunks belonging to different groups are interleaved, as in
```
group0-chunk0
group1-chunk0
group0-chunk1
group1-chunk1
```
Do I have that right?

One way to do that would be to shuffle the rows in `array.from` so that values of `t0` are interleaved. Then `group` (I'm pretty sure) will generate chunks from adjacent rows that are in the same group.

stdlib/universe/quantile.go
```diff
@@ -257,8 +248,12 @@ func (a *QuantileAgg) NewUIntAgg() execute.DoUIntAgg {
 	return nil
 }
 
+// Not sure what to do here. NewFloatAgg is called in a number of places,
+// and could cause state to be recycled erroneously. Need to understand more
+// about when/how this is used.
```
wolffcm, line 253: The only uses I can see are here:
https://github.com/influxdata/flux/blob/ac8d8952718cadfe5e1158cf071f8fa1bf8e38dc/execute/aggregate.go#L395
And here:
https://github.com/influxdata/flux/blob/ac8d8952718cadfe5e1158cf071f8fa1bf8e38dc/execute/aggregate.go#L259

One is used from `simpleAggregateTransformation` and the other `simpleAggregateTransformation2`. I am not sure why there are two of them. I'm not sure if this helps.

wolffcm (changes requested):
I think the test should be updated, I made a suggestion on how to do that.

Most of the rest looks good to me, but I don't understand some of the issues going on here (especially re: `NewFloatAgg`) but I would like to. I'd be happy to pair on this.

PR #4071, opened 2021-09-22
fix: make builds reproducible again by ordering package members in docs

### LINKED ISSUE #4070: Flux builds are no longer reproducible due to flux docs embedding
The `docs.json` build artifact contains a list of packages, with a list of members in each package. These are getting written in to `docs.json` in an unpredictable order. It is file-system dependent. This is causing `libflux.a` builds to be non-reproducible. IE, there can be no change to the source code, but `libflux.a` comes out differently on subsequent build runs.
    
This change shows one way to make the build reproducible, by sorting the top level package list when it's constructed, and using serde `serialize_with` on the member hash-map.

https://github.com/influxdata/flux/tree/adt/reproducible-docs
    
These methods are inconsistent. To make them consistent one could either switch the members field from a hash-map to a bst-map, or or one could use `serialize_with` on the top level packages variable.
    
To verify that builds are reproducible use shasum on `libflux.a` and/or `docs.json` in the target directory.

Note that plain text `ptdocs.json` is added just to assist with debugging and verification.

jsternberg:
The individual members for a package were serialized from a hash map in
an unpredictable order. This caused the `libflux.a` file to come out
differently on subsequent build runs.

This changes the `PackageDoc` structure to use a `BTreeMap` instead of a
`HashMap` to ensure the members are serialized consistently.

Fixes #4070.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


adrian-thurston:
Oh I see now, the packages list comes from a map so that is deterministic. The sort on the vector of packages isn't necessary. Just need to make the members deterministic.

libflux/flux-core/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-#![cfg_attr(feature = "strict", deny(warnings, missing_docs))]
```
rockstar, line 1: Why did we remove this line? I see it's added elsewhere, but I'm pretty sure we want this line here in the top `lib.rs`.

jsternberg: I must have missed reverting this.

jsternberg: I saw that this line was removed and I readded it to the wrong place.

Marwes (approved):
LGTM modulo the attribute being in the wrong file

PR #4074, opened 2021-09-23
chore(libflux): move docs logic into its own Rust module

nathanielc:
Prior to this change the code for generating docs from the stdlib was in
a mix of the bootstrap and flux modules. Now there is a doc module
explicitly for logic related to parsing docs from source.

No other changes were made to the code.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


skartikey (approved):
LGTM!

PR #4076, opened 2021-09-24
fix(checkfmt): fmtd non formatted flux files

skartikey:
with a recent fix, https://github.com/influxdata/flux/pull/4048 , started seeing errors in checkfmt workflow. so formatted a few flux files in this PR


fchikwekwe (approved):
I'm reading this as you just running flux fmt on these files. If that's the case, then this LGTM.

PR #4077, opened 2021-09-24
perf: Avoid rebuilding values that had nothing to substitute

Marwes:
Running `build.rs` takes a long time to due to the flux standard library being re-built every time which greatly hinders the edit-compile-test cycle. Adding a benchmark for this shows that applying substitutions is by far the largest contributor to this so I changed the substitution logic to work on references instead and only re-construct the value if it has actually changed (due to a substitution being made). As most (or at least many) values that `apply` is called on have nothing to replace this greatly reduces the amount of reconstruction that needs to be done.

```
infer/bootstrap         time:   [645.93 ms 654.57 ms 665.24 ms]
                        change: [-49.112% -47.045% -44.805%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
```

(I use this same logic for any sort of type replacement in gluon, and I lifted the merge functions from https://github.com/gluon-lang/gluon/blob/master/base/src/merge.rs)


libflux/flux-core/src/semantic/types.rs
```diff
@@ -261,6 +261,9 @@ impl cmp::PartialOrd for Kind {
     }
 }
 
+/// Pointer type used in `MonoType`
+pub type Ptr<T> = Box<T>;
```
Marwes, line 265: This is orthogonal to the main change in this PR but I initially wanted to check how difficult it would be to use `Arc` instead to allow types to be shared instead of duplicated (via clone). However due to `Substitutable` and `Fresh` taking `self` by value that wasn't possible straight away but will instead come in a separate PR (which further reduces the runtime).

Still left it in as it would conflict with the other changes.

wolffcm (approved):
This looks great! I've learned more about idiomatic Rust as well.

This looks good to go, looks like there just an issue with formatting somehow.

PR #4078, opened 2021-09-24
perf: Share MonoType values instead of cloning

Marwes:
This changes the pointer type used in `MonoType` from `Box` (owned) to `Arc` (shared/reference counted) to reduce the cost of cloning `MonoType` in various places. This does prevent us from mutating `MonoType`, forces us to take `MonoType` by reference in functions like `unify` which forces some additional clone calls to be added (however this is often only because we did not clone the value somewhere else).

Based on #4077

(The change in runtime includes  #4077)
```
infer/bootstrap         time:   [347.86 ms 364.11 ms 383.56 ms]
                        change: [-72.265% -70.544% -68.835%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

```


libflux/flux-core/src/semantic/fresh.rs
```diff
@@ -97,38 +149,54 @@ impl Fresh for Tvar {
...
                     head,
                     tail: MonoType::Record(b),
                 } => {
-                    props.entry(head.k).or_insert_with(Vec::new).push(head.v);
-                    self = *b;
+                    props
+                        .entry(head.k.clone())
+                        .or_insert_with(Vec::new)
+                        .push(head.v.clone());
+                    cur = b;
```
fchikwekwe, line 188: Is this formatting automatic with a format command? Looks nice.

libflux/flux-core/src/semantic/sub.rs
```diff
@@ -151,6 +151,18 @@ pub trait Substitutable {
...
+    /// Apply a substitution to a type variable.
+    fn apply_cow(&self, sub: &dyn Substituter) -> Cow<'_, Self>
+    where
+        Self: Clone + Sized,
+    {
+        match self.apply_ref(sub) {
+            Some(t) => Cow::Owned(t),
+            None => Cow::Borrowed(self),
+        }
+    }
```
fchikwekwe, line 164: Sorry, I'm not seeing where this is getting used? Or what its for.

fchikwekwe (approved):
Overall the change appears pretty straightforward. A few questions that I'd clarified before approval.

libflux/flux-core/src/semantic/fresh.rs
Marwes: Yes, I always have format on save on so the code (should) always be formatted.

libflux/flux-core/src/semantic/sub.rs
Marwes: Seems I removed the code that used it. It is a helper that lets you call `apply(_cow)` on a `&` reference and immediately use the result in any method that then only requires a `&` reference.

So if you want to display a type where you have applied all substitutions you can do `typ.apply_cow(&sub).to_string()` instead of  `typ.clone().apply(&sub).to_string()` or `typ.apply_ref(&sub).unwrap_or(&typ).to_string()`

fchikwekwe: Okay. Rust might complain if its not used anywhere, but thanks for that explanation.

PR #4079, opened 2021-09-24
feat(internal/cmd): add an alternate cli

jsternberg:
This adds an alternate CLI that, in my opinion, has a cleaner interface.
If no argument is given, the REPL is used. If an argument is given, the
file is run through the normal query path which would give the
equivalent answer as InfluxDB's query endpoint.

This CLI also supports sending a jaeger trace.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
I need to do some testing through the CLI interface and going through the repl could affect the results. This isn't intended to change the existing one, but I've had this code on my computer for awhile and have used it multiple times so I'd like to just merge it.

rockstar (commented):
The work itself looks fine here. It's a relatively straight-forward patch. However, I'm feeling hesitant to be the one to say "Yes, let's merge this work" because I think there might need to be some official Planning™ and intent behind adding another CLI. @nathanielc and I have talked a number of times about potentially re-working the CLI, using the lesson's learned by the OSS cli re-work, so I'm hesitant to say "Yes, let's ship this" without some of those considerations (which I haven't yet aggregated), I'm afraid we're doing that "multiple standards" thing that xkcd teases.

Thoughts?

jsternberg:
That's fine. I thought I needed something like this for what I'm doing right now but I think something else might work better. I just didn't want to lose this work as I've needed it a few times.

Let's leave this for now.

PR #4081, opened 2021-09-27
fix(experimental/mqtt): add retain option and publish function

### LINKED ISSUE #3825: Add dynamic per-row parameterization to mqtt.to()
To elaborate on this possibly confusing title, the world that uses MQTT as a data transfer protocol uses it as a dynamic pub/sub tool.  By "dynamic" I mean their clients will probably need to be able to publish/subscribe to different topics dynamically.  The way `mqtt.to()` currently works is it writes out all input table data to the same topic.  This may satisfy some use cases but it likely won't satisfy most.

A common use case for MQTT is to have clients (usually devices) publish to their own respective topics...sometimes granular to the point of having their own topics dedicated to each metric/property they publish.  For example:

device01 PUBLISH'es:
`temperature=45.0 <ts>` to topic: `/device01/temperature`
`pressure=10.0 <ts>` to topic: `/device01/temperature`
device02 PUBLISH'es:
`temperature=25.0 <ts>` to topic `/device02/temperature`
`pressure=11.0 <ts>` to topic: `/device02/pressure`

Each device may also `SUBSCRIBE` to topics that are on-going averages of these metrics over time:
device01 SUBSCRIBE to `/processed/device01/temperature/average`
device01 SUBSCRIBE to `/processed/device01/pressure/average`
device02 SUBSCRIBE to `/processed/device02/temperature/average`
device02 SUBSCRIBE to `/processed/device02/pressure/average`

Unless we want a separate Flux Task running for every topic, a single Flux Task should be able to "fan out" writes to all respective topics.

**What I've tried**
Creating my own wrapper function to have `mqtt.to()` called per row of my resulting stream of tables:
```
import "experimental/mqtt"
import "array"
import "strings"

genTopic = (source_name, source_measurement) =>
  strings.joinStr(arr: ["/processed", source_name, source_measurement, "average"], v: "/")
 
toMQTT = (tables=<-, broker) =>
  tables
    |> map(fn: (r) => { 
                        mqtt.to(tables: array.from(rows: [r]), 
                                broker: broker, 
                                topic: genTopic(source_name: r.host, source_measurement: r._measurement)) 
                        return r
                        } )
 
broker = "broker:1883"
 
from(bucket: "mqtt")
  |> filter(fn: (r) => r["_measurement"] == "cpu")
  |> filter(fn: (r) => r["_field"] == "usage_user" or r["_field"] == "usage_system")
  |> filter(fn: (r) => r["cpu"] == "cpu-total")
  |> mean()
  |> toMQTT(broker: broker)
```

Even wrapping in a function to be called synchronously doesn't work, however:
```
runtime error @9:8-14:28: map: cannot compile @ 9:16-14:26: statement does nothing, side effects are not supported by the compiler
```

`topic` may also not be the only parameter that needs to be dynamic.  It's possible `message` will need to be as well.  I haven't tried to parameterize `message` yet so I don't know if work needs to be done there.  However, the case for it being dynamic is that the default `message` value is the the data passed to `mqtt.to()` which will end up being written out as Line Protocol.  Line Protocol will have redundant information with information that will likely already be in `topic` name/s.  For a cleaner `SUBSCRIBE` experience (and less data written overall), a user might want to, say, trim down the `message` output to be just a value and a timestamp.

### LINKED ISSUE #3836: Add `retain` argument to `mqtt.to()` function
[Explanation of this argument](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/)

When publishing to an MQTT broker, the publishing client sends four things:
* Topic
* Payload
* QoS (Quality of Service)
* Retain 

Currently the `mqtt.to()` function accepts the former three  It does not accept the fourth.  By default, not setting the `retain` flag in an MQTT PUBLISH message means it is _not_ retained.  When it is not retained, it's harder to debug and display topics/messages in MQTT broker UIs.

### LINKED ISSUE #3917: [mqtt.to()] Make `timeColumn` optional
The requirement to send a valid line of Line Protcol seems arbitrarily restrictive here:

1) MQTT topics are often comprehensively descriptive and shouldn't require anymore information than the user decides to include.  Example: I have a topic `/things/temperature/average_5m` that is for the average temp of all things for the last 5 min.  The only information I would need in the MQTT payload is the value, itself.  I don't necessarily need a timeColumn, or Tags, or anything else.

2) Payloads should accept JSON, CSV, and any other format we support.

alespour:
This PR 
- adds `retain` parameter to `to` function (closes #3836)
- adds `publish` function (closes #3825 and closes #3917) and removes `message` parameter from `to` function

#### Function `to`

Sends query result to a broker. The result (Flux tables) is encoded as InfluxDB [line protocol](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/). Each table is sent per one MQTT message.

Changes:
* added optional boolean `retain` parameter (default `false`) fo setting MQTT message `retained` flag.
* it is a pipeline function intended to send query result (tables), it makes no sense to have a parameter for sending arbitrary data, therefore `message` parameter is removed

_Other parameters not changed._

Example:
```
from(bucket: "example-bucket")
  |> range(start: -5m)
  |> filter(fn: (r) => r._measurement == "airSensor")
  |> mqtt.to(
    broker: "tcp://localhost:8883",
    topic: "air-sensors",
    clientid: "sensor-12a4",
    tagColumns: ["sensorID"],
  )
```

#### Function `publish`

New function for sending arbitrary data to MQTT. It returns `bool` status if message was sent or not.

Parameters:
- `topic` - target MQTT topic. Mandatory string parameter, cannot be empty.
- `message` - message payload. Mandatory string parameter, cannot be empty.
 - `broker`, `qos`, `retain`, `clientid`, `username`, `password`, `timeout` - same meaning as in `to`

_There are no other pararameters._

Example of standalone use:
```
import "experimental/mqtt"

mqtt.publish(broker: "tcp://localhost:1883", topic: "greetings", message: "hello")
```

Example of per-row use in pipeline:
```
from(bucket: "example-bucket")
  |> range(start: -5m)
  |> filter(fn: (r) => r._measurement == "airSensor")
  |> last()
  |> map(fn: (r) => ({ r with 
    sent: mqtt.publish(broker: "tcp://localhost:1883", topic: "air-sensors/${r.sensorID}", message: string(v: r._value))
  }))
```

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/experimental/mqtt/to_test.go
+++ b/stdlib/experimental/mqtt/to_test.go
@@ -104,9 +104,11 @@ from(bucket:"mybucket") |> mqtt.to(broker: "tcp://iot.eclipse.org:1883", usernam
 	}
 }
 
-const broker = "tcp://mqtt.eclipseprojects.io:1883" // "tcp://iot.eclipse.org:1883" seems not available anymore
-const topic = "test-influxdb"
-const receiveTimeout = 15 * time.Second
+const (
+	broker = "tcp://mqtt.eclipseprojects.io:1883" // "tcp://iot.eclipse.org:1883" seems not available anymore
+	topic = "test-influxdb"
+	receiveTimeout = 15 * time.Second
+)
 
 var runScript = func(script string) error {
 	ctx := flux.NewDefaultDependencies().Inject(context.Background())
--- a/stdlib/experimental/mqtt/to_test.go
+++ b/stdlib/experimental/mqtt/to_test.go
@@ -1,7 +1,7 @@
 package mqtt_test
 
 import (
-	"fmt"
+	"context"
 	"testing"
 	"time"
 
@@ -10,7 +10,10 @@ import (
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
 	_ "github.com/influxdata/flux/fluxinit/static" // We need to init flux for the tests to work.
+	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/querytest"
+	"github.com/influxdata/flux/runtime"
 	"github.com/influxdata/flux/stdlib/experimental/mqtt"
 	"github.com/influxdata/flux/stdlib/influxdata/influxdb"
 )
@@ -18,7 +21,7 @@ import (
 func TestToMQTT_NewQuery(t *testing.T) {
 	tests := []querytest.NewQueryTestCase{
 		{
-			Name: "from with database with range",
+			Name: "from bucket",
 			Raw: `
 import "experimental/mqtt"
 from(bucket:"mybucket") |> mqtt.to(broker: "tcp://iot.eclipse.org:1883", timeout: 0s)`,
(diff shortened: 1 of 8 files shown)


stdlib/experimental/mqtt/to_test.go
```diff
@@ -56,6 +104,64 @@ from(bucket:"mybucket") |> mqtt.to(broker: "tcp://iot.eclipse.org:1883", timeout
 	}
 }
 
+const broker = "tcp://mqtt.eclipseprojects.io:1883" // "tcp://iot.eclipse.org:1883" seems not available anymore
```
skartikey, line 107: minor: all three const declarations can go inside const ( )

skartikey (approved):
LGTM!

stdlib/experimental/mqtt/to_test.go
alespour: Thank you!

alespour:
Great, can you please merge it @skartikey? (I do not have the permissions)

skartikey:
> Great, can you please merge it @skartikey? (I do not have write access)

Done! @alespour

PR #4082, opened 2021-09-27
fix: Use a new Fresher instance for each package

### LINKED ISSUE #4045: Type inference for each package should use its own fresher instance
Type inference will use a fresher to create new tvars when analyzing a file. These tvars are then later used to unify constraints and substitute type variables. At the moment, when we analyze the standard library, we reuse the same fresher for the entire standard library. This means that if we have package `a` and package `b` and we make a change to package `a`, then the tvars in package `b` will also be modified.

This is done out of an abundance of caution to ensure that we aren't accidentally assigning two tvars to different types. On the other hand, when we import a package, it's not really possible to ever use any of the tvars from the imported package. This is because when we invoke a function, we create a new instantiation of the type by refreshing all of the tvars and performing type inference on that new instantiation.

A more concrete example.

If we have the identify function:

```
identity = (x) => x
```

This will generalize into a polymorphic function where `x` is the tvar `t1`. When the function is later invoked, `identity` is instantiated to replace all of the tvars with new ones. At the point of invocation, the function signature changes to `t2 -> t2` and the constraints are maintained. Type inference then runs on `t2` independently.

Because of this, it should be safe for us to analyze each package with its own fresher instance rather than having them share the fresher. A type for a tvar will be valid from within the package and will generally not be mixable with tvars that were inferred from a different package.

This is required for https://github.com/influxdata/flux/pull/3623 to be merged. It was reverted [here](https://github.com/influxdata/flux/pull/4044).

Marwes:
This keeps each package more independent so that a change in one flux
file does not cause the type variables for unrelated files to change.

Unsure if this needs a more targeted test but I changed the basic
inference tests to use their own `Fresher` for each import which should
closely emulate this.

Fixes #4045

### Done checklist
- [x] Test cases written


libflux/flux-core/src/semantic/tests.rs
```diff
@@ -90,33 +90,22 @@ fn infer_types(
...
     // Parse polytype expressions in initial environment.
     let env = parse_map(env);
 
     // Compute the maximum type variable and init fresher
-    let max = if env.max_tvar() > max {
-        env.max_tvar()
-    } else {
-        max
-    };
+    let max = env.max_tvar();
```
jsternberg, line 108: Is this needed anymore? It seems to me like the area below that initializes a fresher should also use `Fresher::default()`.

libflux/flux/src/lib.rs
```diff
@@ -440,13 +437,14 @@ impl SemanticAnalyzer {
             return Err(fluxcore::Error::from(format!("{}", &errs[0])));
         }
 
-        let mut sem_pkg = fluxcore::semantic::convert::convert_with(ast_pkg, &mut self.f)?;
+        let mut fresher = fresher();
```
jsternberg, line 440: This is the weird situation where this code needs to stay the way that it is. This code is special for the REPL and, unlike everywhere else, is reentrant. It's meant to work when a user inputs multiple commands into the REPL. As the REPL is essentially like gradually parsing a single main package, the fresher has to be maintained across invocations.

jsternberg (changes requested):

libflux/flux-core/src/semantic/tests.rs
Marwes: It shouldn't be, however the `imports` test seems to require that `readwrite = true` right now which makes the substitutions also be applied to that top level environment which shouldn't be necessary as there should not be any free variables in it anyway however it appears to preserve the type as `"f" => "(x: A) => A"` else it gets changed to the isomorphic `"f" => "(x: D) => D"`. Not really sure how that works though.

libflux/flux/src/lib.rs
Marwes: True, as the `Environment` persists here, the fresher must also persist. However if applications on `PolyType` took shadowing into account then we wouldn't need to worry about variables being reused.

Optimally, we should only need to persist a `Fresher` throughout a set of mutually recursive recursive bindings. Outside of that all variables should be bound (is that the right term?).

libflux/flux-core/src/semantic/tests.rs
jsternberg: We should see if we can fix that. I think this section of the code is the part that needs to be changed for the embedding semantic graph to work.

Marwes: I believe I have fixed this now, though I did have to add in another "bug" error to guard against variable reuse (i.e a code path I do not believe should ever be taken).

libflux/flux-core/src/semantic/types.rs
```diff
@@ -94,13 +95,19 @@ impl Substitutable for PolyType {
 }
 
 impl MaxTvar for [Tvar] {
-    fn max_tvar(&self) -> Tvar {
-        self.iter().max().cloned().unwrap_or(Tvar(0))
+    fn max_tvar(&self) -> Option<Tvar> {
```
Marwes, line 98: Since `None < Some()` all comparisons still work the same, it is just that we distinguish `0` being max and no variables (otherwise the "bug" path triggers in tests).

libflux/flux-core/src/semantic/tests.rs
Marwes: Seems to be some issues in the go-based tests still unfortunately

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -319,7 +329,12 @@ impl File {
             imports.push(name);
 
             match importer.import(path) {
-                Some(poly) => env.add(name.to_owned(), poly),
+                Some(poly) => {
+                    // Types imported from other pacakges may have the same variables so we need
```
jsternberg, line 333: My understanding is that this shouldn't be necessary because type instantiation will instantiate the variables. Do you think we can talk about this tomorrow so we can run through this together?

jsternberg (changes requested):
Looking good and I think it's almost ready to merge. I had a few questions for myself so I understand this more fully.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -250,6 +250,16 @@ pub fn infer_pkg_types<T>(
 where
     T: Importer,
 {
+    if env.readwrite {
```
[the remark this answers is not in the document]
Marwes: Yes, though I count is a a bug since we are also the callers of this function (since the go tests fail there is still some issues around this so I guess it was good that I tried adding this).

libflux/flux-core/src/semantic/nodes.rs
Marwes: It should. However any definition added here could contain variables that are later generated from the `Fresher`. If that happens then calling `Environment::apply` would be applied on these imported types as well which is wrong.

So say we import `f: forall a . a -> a` (a == Tvar(0), then the fresher also creates `Tvar(0)` and it gets unified with `int`. Once the substitution gets applied then the code as written will substitute `f: forall a . -> a => int -> int`  which is wrong because the `forall` (from the `PolyType`) should make these `Tvar(0)`s distinct.

I think it may be easier/safer to just fix the shadowing issue around `PolyType`s

libflux/flux-core/src/semantic/nodes.rs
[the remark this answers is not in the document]
jsternberg: Do you think you can post the Go error? I'd like to know the error as I've suspected the Go code is wrong for some time, but have never known how it is wrong.

jsternberg (approved):

PR #4083, opened 2021-09-27
fix: Report errors from function parameters

Marwes:
### Done checklist
- [x] Test cases written


libflux/go/libflux/buildinfo.gen.go
```diff
@@ -22,16 +22,16 @@ var sourceHashes = map[string]string{
 	"libflux/flux-core/src/ast/flatbuffers/mod.rs":                                                "00c75dc1da14487953a4a017616fb8a237fe3da437c876f1328532dd7906f015",
 	"libflux/flux-core/src/ast/flatbuffers/monotype.rs":                                           "4e82aae8e7edb8e230302241fe702ece9ae84fc708003c24709c2ea944cfaae8",
 	"libflux/flux-core/src/ast/flatbuffers/tests.rs":                                              "ea85b78049f0c3d79daedcf9e171ccb00d0a596d1a643aa218590c99e6b160bc",
-	"libflux/flux-core/src/ast/mod.rs":                                                            "676e22d24cd7a97525688ed391a980d08b4bee40c5b5067d187f3ae6d66210ca",
```
Marwes, line 25: Does this file actually need to be committed to git? Having to run `make generate` is rather tedious on each PR and it conflicts easily.

jsternberg: Yes, it unfortunately does. This is the file that signals to the Go build system that the Rust code has changed and forces it to rerun the compiler. Without this, the Go build system doesn't know when it should rebuild the Rust code and it can get confused.

libflux/flux-core/src/ast/mod.rs
```diff
@@ -764,7 +764,10 @@ fn get_err_monotype(mt: MonoType) -> String {
 }
 fn get_err_parameters(prs: Vec<ParameterType>) -> String {
     for pr in prs {
-        get_err_parameter(pr);
+        let e = get_err_parameter(pr);
```
jsternberg, line 767: If I'm understanding this change correctly, did this method just not work at all previously? Is there a test for this behavior? This is a pretty good find and I'd like to make sure we keep it fixed.

jsternberg (changes requested):
Just need a test. This looks good. I like the `mem::take` change especially.

nathanielc (approved):

libflux/go/libflux/buildinfo.gen.go
Marwes: But if the file is missing (which would be the case on CI) then it ought to be re-generated which should have the same effect, right? I might be missing something, but isn't the file changing just something to ensure it rebuilds on local builds?

libflux/flux-core/src/parser/tests.rs
```diff
@@ -3147,6 +3147,16 @@ fn test_parse_record_type_tvar_properties() {
     )
 }
 
+#[test]
+fn test_parse_record_unclosed_error() {
```
Marwes, line 3151: I believe this test should check the behaviour? @jsternberg

libflux/go/libflux/buildinfo.gen.go
jsternberg: The Go build system isn't aware of where any Rust files are or what they look like. It only has the capability to do that with C and C++ files that are included with cgo. We use `pkg-config` as an interface to help us coordinate, but Go itself treats it as an outside dependency and doesn't track anything about it. It doesn't even necessarily run `pkg-config` unless it detects that something inside of the package has changed. So the changes to this file inform the Go build system that stuff outside of its control has changed.

libflux/flux-core/src/parser/tests.rs
jsternberg: Oh github collapsed this change so I didn't see it.

jsternberg (approved):

libflux/go/libflux/buildinfo.gen.go
Marwes: `make` is already used though so couldn't we add a step on CI that generates this file before it invokes go?. Or do this from the go build system if it has the capability of running arbitrary scripts.

jsternberg: This possibly looks like an error in the makefile. The makefile should rerun this automatically if it is used. The makefile isn't the only way to build the code and the go build system is really finicky about things, but if you use the go build file, it should do this automatically. It looks like if [this line](https://github.com/influxdata/flux/blob/c7de555f4b1cb66045c5a6c0ee020f017b0aa6be/Makefile#L39) was moved into the `LIBFLUX_GENERATED_TARGETS` section then make would automatically regenerate this file when a Go target was invoked.

nathanielc:
The CI job has these lines:

```
stdlib/influxdata/influxdb/sample/alignToNow_test.flux
stdlib/influxdata/influxdb/sample/sample.flux
stdlib/influxdata/influxdb/schema/show_tag_values_measurement_or_predicate_test.flux
stdlib/universe/derivative_test.flux
found files that are not formatted
```

Those files were recently fixed in master and the CI job as it wasn't reporting errors correctly. 

A rebase to current master will likely fix the errors.

PR #4087, opened 2021-09-27
feat(zoneinfo): copy over location related code from the Go time package

### LINKED ISSUE #4057: Copy tzdata related code from Go standard library to internal package
We will need to copy some of the tzdata related code from the Go standard library to allow our own time zone code to use it.

The `time` package contains code to parse the tzdata files into a Go data structure and code to interact with it. We need access to one of the internal methods on the `Location` struct so we will need to copy out the code related to `Location` so we have access to it for our own time zone implementation.

There is some code for this in the [feat/timezone-poc](https://github.com/influxdata/flux/tree/feat/timezone-poc) branch.

DOD:
- Location struct ported to Flux
- We do not want the embedded tzdata
- We need the function that can parse a tzdata file 
- Unit tests for Location struct and related code (these can come from Go stdlib)
- Copyright headers, make sure we correctly attribute the original code.
- We only need to bring the arch specific code for macOS, Linux, and Windows

jsternberg:
This copies over the location related code from the Go standard library
and removes some of the unused features from it. We remove the platforms
we do not support and remove concepts like `Local` because support for
that doesn't make sense anyway in a server context.

Fixes #4057.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/internal/zoneinfo/LICENSE
+++ b/internal/zoneinfo/LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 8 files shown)


fchikwekwe (approved):
LGTM

PR #4090, opened 2021-09-28
test: Correct the type now that the parser reports errors in arguments

Marwes:
cc #4083 #4066

### DIFF
--- a/libflux/flux-core/src/semantic/tests.rs
+++ b/libflux/flux-core/src/semantic/tests.rs
@@ -3379,7 +3379,7 @@ fn issue_4051() {
         x = f(r:r)
         "#,
         exp: map![
-            "f" => "(r:{S with a:X, b: int, c: float) => {S with a:X, b: int, c: float}",
+            "f" => "(r:{S with a:X, b: int, c: float}) => {S with a:X, b: int, c: float}",
             "x" => "{S with a: X, b: int, c: float}",
         ],
     }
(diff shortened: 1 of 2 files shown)


libflux/flux-core/src/semantic/tests.rs
```diff
@@ -3379,7 +3379,7 @@ fn issue_4051() {
         x = f(r:r)
         "#,
         exp: map![
-            "f" => "(r:{S with a:X, b: int, c: float) => {S with a:X, b: int, c: float}",
+            "f" => "(r:{S with a:X, b: int, c: float}) => {S with a:X, b: int, c: float}",
```
rockstar, line 3382: How did a typo like this get past CI? I would say "Can we add a test for this?" but _this is the test_, so I'm worried we have an issue here.

rockstar (approved):

libflux/flux-core/src/semantic/tests.rs
nathanielc: So a quick google search shows we can tell CircleCI to merge the branch with the parent before tests etc. See https://circleci.com/developer/orbs/orb/circleci/build-tools#usage-merge-with-parent

This could be a simple fix here. We do have an internal system for private repos that runs CI against a rebased commit but that system doesn't work with OSS repos. So perhaps we could try this CircleCI config for a week or so.

PR #4091, opened 2021-09-28
chore(circleci): add step to merge PR with parent

nathanielc:
This will help to prevent in most cases CI failures against master that
arise because two PRs merge that are in conflict with one another and
one PR is based off an older master commit.

This will not solve all cases. This change assumes that the branch's
parent is the master branch, this is true most of the time but not
always. In cases where its not true its still valuable to test but we
loose the safeguard of testing against master.

In other cases where the CI for both PR runs and passes before either PR
merges we would not catch potential conflicts.

Alternatively we could require that branches are up-to-date with master
before merging but that puts a lot of onus on the PR author to
constantly rebase. This change seems like a good balance in that is
should catch most cases at little cost to the development cycle.

See https://github.com/influxdata/flux/pull/4090#discussion_r717668829

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,10 +141,7 @@ jobs:
             # Cargo's built-in support for fetching dependencies from GitHub requires
             # an ssh agent to be set up, which doesn't work on Circle's Windows executors.
             # See https://github.com/rust-lang/cargo/issues/1851#issuecomment-450130685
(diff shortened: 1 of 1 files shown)


nathanielc:
So circleci's merge-with-parent step is failing without any output. Going to close this PR but leave the branch around for reference

PR #4093, opened 2021-09-28
feat(zoneinfo): add methods to convert to and from local clock time

### LINKED ISSUE #4058: Add FromUTC and ToUTC methods to the Location struct
Related to #4057, we will want to add `FromUTC` and `ToUTC` methods to the `Location` struct. These should be modeled off of the ones in the [proof of concept](https://github.com/influxdata/flux/blob/feat/timezone-poc/internal/zoneinfo/zoneinfo.go#L198-L248).

The purpose of these methods is to transpose a timestamp from a location to the equivalent clock time in or from UTC. So a time at 9:00 AM on January 1, 2020 in a location will get transposed to the unix timestamp for 9:00 AM on January 1, 2020 in UTC or going the other direction.

When going from UTC back to the location, there are two edge cases. When there are two suitable times, the time that is in an earlier zone transition should be chosen. When there is no suitable time, the one at the time of the zone transition should be chosen.

There should be tests that demonstrate the above in locations that are both west and east of the prime meridian.

DOD:
- Port `FromUTC` and `ToUTC` methods from POC.
- Tests with location west and east of meridian for each case
- Add unit tests with 100% code coverage
- Create a name for this concept of _fake UTC_, i.e. converting to clock time UTC not the actual UTC time equivalent for the original time.

jsternberg:
This adds methods to the `Location` struct to convert from local clock
time to a UTC version of the same clock time and to convert to the local
clock time from the UTC version of the same clock time.

This will be used by the time zone support to translate a local time to
a version that we can perform time math on and then to convert back from
the time math to the local clock time.

Fixes #4058.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


internal/zoneinfo/zoneinfo_clock_test.go
```diff
@@ -0,0 +1,207 @@
...
+			want:    "2017-10-01T03:00:00+11:00",
+		},
+		{
+			name:    "Australia_East DST end",
+			locname: Australia_East,
+			s:       "2017-04-02T02:30:00Z",
+			want:    "2017-04-02T02:30:00+11:00",
+		},
+		{
+			name:    "American_Samoa day skip morning",
```
scbrickley, line 71: Should we include similar test cases for American Samoa in the `FromLocalClock` tests? Maybe there's a good reason why we didn't, but that reason is not obvious to me.

internal/zoneinfo/zoneinfo_clock.go
```diff
@@ -0,0 +1,57 @@
...
+		case local >= end:
+			_, offset, start, end, _ = l.lookup(end)
+			if local := sec - int64(offset); local < start || local >= end {
+				// The time does not appear to exist. Truncate the interval.
+				return start * int64(time.Second)
+			}
+		}
+	}
+
+	// Inspect the previous time zone transition.
```
scbrickley, line 41: I know we've talked about this in meetings before, but I think I'm still unclear on what the issue is with the go time library that makes this code necessary. I don't doubt the logic here, I think I would just like to understand the problem it's trying to solve in greater detail.

scbrickley (approved):
I think this is fine to merge. I just had a few questions.

internal/zoneinfo/zoneinfo_clock_test.go
jsternberg: It's not really possible unfortunately. Because the time doesn't exist, it's not really clear what we would be testing. On the other hand, going from UTC to the local timezone has to verify the behavior of "what if this time doesn't exist?"

internal/zoneinfo/zoneinfo_clock.go
jsternberg: I'll expand this comment. The main reason is that we're essentially guessing which zone transition, which is a start and end timestamp along with an offset, that applies to the time we want. Since there are possibly multiple times that correspond to a single utc clock time, we need a way to consistently determine which of them we want. We've decided to say "the first clock time that occurs is the correct one" which means that the time we are looking for is in the earlier zone transition. Since the above code guesses the zone transition by looking it up with the utc timestamp (which isn't correct), we need to check to see if there is another zone transition that could also be the correct one. Because we only care about this condition to find the earliest timestamp, we only have to check the previous one. We don't really care if we accidentally chose the first one and there's also an available zone transition in the future because we will never choose it anyway.

PR #4094, opened 2021-09-28
fix: reject usePrevious when column doesn't exist

rockstar:
This patch addresses an issue where one might do
`fill(column: "a-nonexistent-column" usePrevious: true)`. Why would
someone do that, you might ask. In the current bug being chased, it was
because the table was a result of a `union` of two different schemas,
and sometimes the fill happened when the other schema was the first
record in the `union`.


jsternberg:
Can you expand on the bug that prompted this? My understanding is that the described behavior is non-deterministic but should otherwise be allowed and if the user wants it to be deterministic they'll have to add a sort or some other deterministic transformation after the union.

PR #4095, opened 2021-09-29
chore: Avoid the need to run make generate after modifying the rust code

Marwes:
Removes `buildinfo.gen.go` from git so that changes to it do not cause
merge conflicts however it still exists and are generated by `make
generate` so changes in it will cause cgo to run again. The issue still
remaining is to assert that the go code does not use a cached version of
the rust library so I embedded the git revision into the rust library
and added a test to check that it matches to the one seen when running
the go tests. Since `buildinfo.gen.go` will be generated fresh on each
CI call that forces cgo to use the correct version of the rust build.

### Done checklist
- [x] Test cases written


jsternberg:
This won't work because the Go build system won't get notified that the Rust code has changed. The makefile isn't used by the Go build system and has no way to interact with it. This ends up being a bigger problem when Flux is imported from another project such as InfluxDB which is the reason why we have to commit the file. Without the file, there's no way to import the code in another project because of how the Go build system hashes inputs and there's no way to modify the build for dependent projects because the module cache is not writeable.

Marwes:
> This won't work because the Go build system won't get notified that the Rust code has changed.

It will as long as `go generate` still runs before `go build` which is the case on CI at least. Locally we can't guarantee this but we don't check that the generated file iis up to date either so this does not change anything there. It is, as you say an issue when rendering however as it isn't expected that `go generate` is run for dependencies (?).

> There are also situations where the build happens and the git directory isn't available so we can't use any git commands inside of the build for either go or rust.

As the hash is only necessary for testing it could be made optional to let it work in other repositories (which would not verify the hash anyway).

So if this does not work, perhaps we could (ab)use the new [embed directive](https://pkg.go.dev/embed) that go has and embed all the rust and stdlib files that we currently have hashes for. It does force us to restructure the project so that the go project contains all the files it depends on as embed does not support `..` https://github.com/influxdata/flux/tree/take2 . (Maybe we could use symbolic links instead).

jsternberg:
> It will as long as go generate still runs before go build which is the case on CI at least. Locally we can't guarantee this but we don't check that the generated file is up to date either so this does not change anything there. It is, as you say an issue when rendering however as it isn't expected that go generate is run for dependencies (?).

The problem with dependencies is how the go build system determines whether to rebuild. Imagine that I build influxdb and it involves compiling the `github.com/influxdata/flux/libflux/go/libflux` package. We then make changes to the Rust code and make a new release. When influxdb goes to rebuild, it sees that `github.com/influxdata/flux/libflux/go/libflux` hasn't been updated and uses the old build. This means that the Rust build never gets triggered through `pkg-config`. The `go generate` step is not automatically run during the build process and we need it to ensure that the package gets built properly.

> As the hash is only necessary for testing it could be made optional to let it work in other repositories (which would not verify the hash anyway).
>
> So if this does not work, perhaps we could (ab)use the new embed directive that go has and embed all the rust and stdlib files that we currently have hashes for. It does force us to restructure the project so that the go project contains all the files it depends on as embed does not support .. https://github.com/influxdata/flux/tree/take2 . (Maybe we could use symbolic links instead).

This is interesting. `embed` didn't exist when I was originally setting this up. The only problem I can see with `embed` is that, in order to trigger the build for the `libflux` package, we would need to actually embed the source files in the final binary. I don't think there's anyway we could cause the build to retrigger while not also embedding the file. `embed` would solve the other problems and, if it were acceptable for us to embed the sources in the final binary, I think it would satisfy all of the requirements.

jsternberg:
Just to clarify, this problem here:

> We then make changes to the Rust code and make a new release. When influxdb goes to rebuild, it sees that github.com/influxdata/flux/libflux/go/libflux hasn't been updated and uses the old build.

This is only a problem when using `go mod edit` to specify a filesystem path or when using vendor. When pure modules are used, Go retriggers a build no matter what. Unfortunately, we use vendor in our closed source repository so that problem still exists there. I'm not as concerned about `go mod edit` with a filesystem path because we can then assume the developer will run generate when they are making local changes.

Issue #4004, opened 2021-08-30
Running the formatter multiple times produces different results


scbrickley:
Consider the following flux script:

```
import "array"

array.from(rows: [{_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}, {_value: "one", _time: 2021-08-30T00:00:00Z}], foo: false)
```

Obviously, having all of those record literals on one line is unreadable. Running the formatter quickly fixes this ugly code, as expected: 

```
import "array"

array.from(rows: [
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
    {_value: "one", _time: 2021-08-30T00:00:00Z},
], foo: false)
```

However, running the formatter again, without making any edits, further modifies the script:

```
import "array"

array.from(
    rows: [
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
    ],
    foo: false,
)
```

We should fix the formatter so that it produces the same result every time, and so that running it more than once is a no-op.

nathanielc:
A possible way to fix this is to format the arguments (aka childern) into a temporary buffer and then retab the buffer once you determine if that temp buffer contains newlines.

PR #4101, opened 2021-09-30
fix(libflux): make multiline formatting consistent

fchikwekwe:
This is an initial PR for #4004 that fixes the function expression node type. I'll add another PR for all other affected node types. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/flux-core/src/formatter/mod.rs
```diff
@@ -475,8 +475,42 @@ impl Formatter {
...
+            indentation: self.indentation.clone(),
+            clear: self.clear.clone(),
+            temp_indent: self.temp_indent.clone(),
+            err: self.err.clone(),
+            temp_singleline: self.temp_singleline.clone(),
+        };
+
+        let mut sep;
+
+        for (i, property) in (&n.params).iter().enumerate() {
```
Marwes, line 492:
```suggestion
        for (i, property) in n.params.iter().enumerate() {
```

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -475,8 +475,42 @@ impl Formatter {
...
+        let mut temp = Formatter {
+            builder: String::new(),
+            indentation: self.indentation.clone(),
+            clear: self.clear.clone(),
+            temp_indent: self.temp_indent.clone(),
+            err: self.err.clone(),
+            temp_singleline: self.temp_singleline.clone(),
+        };
+
+        let mut sep;
```
Marwes, line 490: This doesn't need to be forward declared, right?

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -475,8 +475,42 @@ impl Formatter {
...
+                temp.format_node(&Node::from_expr(v));
+            };
+
+            // if any child node adds newlines, then parent node will be multiline
+            if temp.builder.contains("\n"){
+                multiline = true;
+            }
+
+            if i != 0 {
+                sep = ",\n";
```
Marwes, line 503: Won't this cause functions to always be laid out onto multiple lines?

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -475,8 +475,42 @@ impl Formatter {
...
+                temp.write_string(sep);
+                if multiline {
+                    temp.write_indent();
```
Marwes, line 506: If only the second, or third etc argument is spread over multiple lines the the first won't get indented I think. Like

```
import "array"
array.from(foo: false,
    rows: [
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
        {_value: "one", _time: 2021-08-30T00:00:00Z},
    ],
)"
```

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1268,3 +1268,77 @@ a",
...
+        {_value: "one", _time: 2021-08-30T00:00:00Z},
+        {_value: "one", _time: 2021-08-30T00:00:00Z},
+        {_value: "one", _time: 2021-08-30T00:00:00Z},
+        {_value: "one", _time: 2021-08-30T00:00:00Z},
+    ],
+    foo: false,
+)"#
+    );
+
+    assert_format(
```
Marwes, line 1317: Something that is formatted should never change (ie format is idempotent) so we could add an `assert_unchanged` check on the formatted ouput in `fn assert_format`.

Marwes (changes requested):

Marwes:
I don't think all my comments need to be fixed (ie assert_unchanged in assert_format may end up causing more failures).

fchikwekwe:
@Marwes yeah, there's a reason I needed `assert_format` for this case. I'll add a detailed explanation once I'm done with current fixes. 

And yes, some of the errors you point out are correct and causing the bug that I'm fixing now that I mentioned this morning. Thanks for the feedback. I'm working on ironing those out.

fchikwekwe:
I can break out the formatting into another PR if this one is too big. Just let me know.

libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -53,7 +53,7 @@ pub fn infer_stdlib_dir(path: &Path) -> Result<(PolyTypeMap, PolyTypeMap, Semant
 fn infer_pre(f: &mut Fresher, ast_packages: &ASTPackageMap) -> Result<(PolyTypeMap, PolyTypeMap)> {
     let mut prelude_map = PolyTypeMap::new();
     let mut imports = PolyTypeMap::new();
-    for name in PRELUDE {
+    for name in PRELUDE.iter() {
```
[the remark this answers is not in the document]
fchikwekwe: Yeah, thanks. I had to update my rustup toolchain. This chain is being reverted.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -115,18 +115,26 @@ impl Formatter {
         self.indentation -= 1;
     }
 
-    fn set_indent(&mut self, i: u32) {
+    fn reindent(&mut self, want_indent: i32) {
```
fchikwekwe, line 118: This file contains the actual formatter changes.

stdlib/slack/slack.flux
```diff
@@ -126,7 +118,7 @@ endpoint = (url=defaultURL, token="") => (mapFn) => (tables=<-) => tables
     |> map(
         fn: (r) => {
             obj = mapFn(r: r)
-
+
```
wolffcm, line 121: This is a small thing, but if we can avoid indenting blank lines that would be good.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -304,28 +344,47 @@ impl Formatter {
...
-        for (i, p) in (&n.parameters).iter().enumerate() {
-            if i != 0 {
+
+        for i in 0..n.parameters.len() {
+            temp_formatters[i].reindent(self.indentation);
+            self.write_string(&temp_formatters[i].builder);
+            if i < n.parameters.len() - 1 {
                 self.write_string(sep);
                 if multiline {
                     self.write_indent();
```
wolffcm, line 383: Regarding indents appearing on blank lines, it seems like we could avoid this if we indented before writing each item in a list rather than after. But thinking it over I think there could be a lot of ripple effects doing it this way and it's not really a big deal.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -372,9 +432,23 @@ impl Formatter {
         }
     }
     fn format_record_type(&mut self, n: &ast::RecordType) {
-        let multiline = n.properties.len() > 4 || n.base.is_multiline();
+        let mut multiline = n.properties.len() > 4;
```
wolffcm, line 435: Maybe this `4` should be defined as a constant in this file? (Assuming that 4 is the value that's used everywhere)

libflux/flux-core/src/formatter/mod.rs
fchikwekwe: Yeah, I think that not indenting blank lines would be better addressed as a follow-up PR just to try and limit what's addressed here.

stdlib/universe/join_panic_test.flux
```diff
@@ -59,6 +59,5 @@ t_join_panic = (table=<-) => {
         |> group()
 }
 
-test _join_panic = () => 
-    // to trigger the panic, switch the testing.loadStorage() csv from `passData` to `failData`
-    ({input: testing.loadStorage(csv: passData), want: testing.loadMem(csv: outData), fn: t_join_panic})
+test _join_panic = () => // to trigger the panic, switch the testing.loadStorage() csv from `passData` to `failData`
+({input: testing.loadStorage(csv: passData), want: testing.loadMem(csv: outData), fn: t_join_panic})
```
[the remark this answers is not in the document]
wolffcm: The issue with the csv is that the strings are multi-line, and we don't want to actually alter the content of the string, so just replacing `\n` with `\n    ` to re-indent will cause the change... it's unclear to me what the best solution for that would be.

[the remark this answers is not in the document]
fchikwekwe: This is what current happens when I add the newline back in. 
```
 EXPECTED: 
 [
    "a": 0,
    //multilinecomment
    // multiline comment
    "b": 1,
] 
 OUTPUT: 
 [
    "a": 0,
    
    //multilinecomment
    // multiline comment
    "b": 1,
] 
```
I'm trying to see if I can check if a newline is necessary before adding one.

stdlib/influxdata/influxdb/schema/show_tag_values_measurement_or_predicate_test.flux
```diff
@@ -71,12 +71,12 @@ testcase show_tag_values_measurement_or_predicate {
 
     want = csv.from(
         csv: "#datatype,string,long,string
-#group,false,false,false
-#default,0,,
-,result,table,_value
-,,0,north
-,,0,south
-",
+    #group,false,false,false
```
Marwes, line 74: I assume the `safe_to_reindent` flag is supposed to prevent this indentation? Do these files need reformatting?

fchikwekwe: oh boy, yeah there seems to still be a bug with csv. Lemme take a look.

ast/astutil/format_test.go
```diff
@@ -279,11 +279,9 @@ func TestFormatWithComments(t *testing.T) {
 
 	if want := `// hi
 // there
-{
```
fchikwekwe, line 282: @Marwes @wolffcm This is the file where I made a change the the expected output. Its a little bit weird, but I think that the final output makes sense for the most part.

Marwes: I'd probably expect that the line comment forced the record onto multiple lines, but the formatting before didn't indent the comment so it (arguably) also weren't consistent.

ast/astutil/format_test.go
```diff
@@ -279,11 +279,9 @@ func TestFormatWithComments(t *testing.T) {
...
-    _time: r._time,
-    io_time: r._value,
-// this is the end
-}
+{_time: r._time, io_time: r._value
+    // this is the end
+    }
```
Marwes, line 284: This `}` should probably not be indented?

Marwes (commented):
Arguably some inconsistencies with comments in strange places but it might not be worth the time to try and fix every place where someone could put a comment. Otherwise LGTM

stdlib/universe/map_shadow_var_test.flux
```diff
@@ -36,7 +36,7 @@ t_shadow_var = (table=<-) => table
     |> map(
         fn: (r) => {
             myVal = "const"
-
+
```
Marwes, line 39: Spurious indentation here (empty lines ought to not have any indentation)

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -1027,12 +1203,23 @@ impl Formatter {
...
         self.write_rune('"');
         let escaped_string = self.escape_string(&n.value);
+
+        // if the string literal contains any newlines, then its not safe to `reindent` the string
+        // literal. Applying `reindent` to string literals with newlines can break them
+        // e.g. csv data
+        if escaped_string.contains("\n") {
+            self.unindent();
+            self.safe_to_reindent = false;
+        }
```
wolffcm, line 1222: This will prevent ever re-indenting a string literal that contains a newline, but it will also prevent reindention of other things that we _want_ to reident. Really, we want to be able to reindent anything except for a multiline string, but that is difficult to do after you've already formatted.

For example, I don't think this would work quite right:
```js
{
  b: 2,
  c: () => {
      // line 1
     a = 0
    return "a 
multi
line
string"    
    }
}
```
Property `c` is multi-line, the record literal would be multi-line, and we should reindent all the properties. But because `c` contains a string literal with a newline we won't, even though ideally we would want to indent the function body. Can you give this example a try and see what happens?

With that said, I think you've solved the most serious issue here, that the formatter wasn't idempotent. So I think it would be fine to include a test case like the one above, and file a bug to address it later (provided I'm correct in my suspicions). I think this case is difficult to solve given the current architecture of the formatter.

fchikwekwe: Yes, you are right. In this case, the first closing brace is not appropriately indented. This is the output: 
```
 EXPECTED: 
 foo = {
    b: 2,
    c: () => {
        // line 1
        a = 0
        return "a
multi
line
string"
    },
} 
 OUTPUT: 
 foo = {
    b: 2,
    c: () => {
    // line 1
    a = 0

    return "a
multi
line
string"
},
} 
```

I do think that in this case it might be easier to isolate the issue of string literals versus other types into another smaller issue in order for that to be dealt with separately. I'll file that before I close this issue.

fchikwekwe:
@Marwes I think that since these are related to weird comments, its a little bit of a regression, but again, I would prefer to consolidate the issue of fixing formatted comments into another smaller issue so it doesn't block this from merging. Lemme know if you think that's a bad idea.

Marwes:
Yeah, I would fix those problems in a different issue/PR.

PR #4104, opened 2021-10-01
feat(stdlib): expose location functionality to the window transformation

### LINKED ISSUE #4102: Add location support to the window function
With #4059, we will have the capability to use location information as part of the interval package. We need to add an option to the window function that allows us to specify the location to use when windowing.

We will add an optional `location` parameter to the window function. The default location should be the `location` option in the universe package. The default value for `location` is `UTC` and `UTC` causes it to not use any location information which is the default functionality. If another string is used, we will load the location information and provide it to the window in the interval package. An invalid location or an empty string are invalid.

In order to implement the option, something like this will probably be needed.

```
option location = "UTC"

builtin _window = (<-tables: [A], ..., location: string) => [B] where A: Record, B: Record

window = (tables=<-, ..., location=location) => tables |> _window(..., location)
```

DOD:
- The `location` parameter exists for `window()`.
- The default value is `option location`.
- Push downs should be disabled if the location parameter is not `UTC`.
- `option location = "UTC"` should be defined in the same package.

jsternberg:
This exposes the location functionality to the window and
aggregateWindow functions. The location can be specified as an optional
parameter to customize the location that the windowing happens in.

By default, the windowing happens in the location specified by the
`location` option defined in package `universe`. The default value for
this is `utc` which is the equivalent of the string `"UTC"` and disables
location functionality entirely.

This also modifies the `window()` transformation to accept `0s` as an
argument to `window()`. When `0s` is used as a parameter to `every` and
`period`, it tells window to use the default value. The default value
for `every` is `period` and the default value for `period` is `every`.
If both are zero, an error is returned.

The `aggregateWindow()` function now also exposes the `period` parameter.

Fixes #4102.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/window.go
```diff
@@ -46,80 +48,74 @@ func CreateWindowOpSpec(args flux.Arguments, a *flux.Administration) (flux.Opera
...
+		d, err := args.GetRequiredDuration("every")
+		if err != nil {
+			return flux.Duration{}, err
 		}
 
 		if d.IsNegative() {
-			return flux.Duration{}, false, errors.New(codes.Invalid, `parameter "every" must be non-negative`)
-		} else if d.IsZero() {
-			return flux.Duration{}, false, errors.New(codes.Invalid, `parameter "every" must be nonzero`)
+			return flux.Duration{}, errors.New(codes.Invalid, `parameter "every" must be non-negative`)
```
fchikwekwe, line 58: Why do we not care if every is zero anymore?

stdlib/universe/window.go
```diff
@@ -46,80 +48,74 @@ func CreateWindowOpSpec(args flux.Arguments, a *flux.Administration) (flux.Opera
...
+	if location, err := args.GetRequiredString("location"); err != nil {
+		return nil, err
+	} else {
+		spec.Location = location
+	}
+
+	if spec.Every.IsZero() && spec.Period.IsZero() {
 		const docURL = "https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/window/"
-		return nil, errors.New(codes.Invalid, `window function requires at least one of "every" or "period" to be set`).
+		return nil, errors.New(codes.Invalid, `window function requires at least one of "every" or "period" to be set and non-zero`).
```
fchikwekwe, line 89: Oh, is this why? Because this check would catch a zerod every  in many cases?

fchikwekwe (approved):
Question about the every parameter. Otherwise I think this change makes sense for the most part.

stdlib/universe/window.go
jsternberg: The window function has a strange behavior. If every is not specified, its default value is period. If period is not specified, its default value is every. It's an error for both of them to be unset.

This error message got changed because now it's valid to say `window(every: 0s, period: 1s)` as that will signal that every should use period as its default. So the error message is changed to also specify that you can't do `window(every: 0s, period: 0s)`.

PR #4114, opened 2021-10-04
feat: Don't introduce constraints for default arguments

### LINKED ISSUE #3461: Function argument defaults should not introduce type constraints

Marwes:
This alters how inference works around default arguments. Instead of immediately inferring the arguments and using the inferred type in the body and function signature, this is instead delayed to after the type of the function has been inferred.

```
f = (a="a", b) => {
     "${string(v: a)}, ${string(v: b)}"
}

// Before the variables have these types in the body
a: string
b: B
// With this PR
a: A
b: B
// And then we add another constraint after generated like there was a call like
f(a: "a", b: ANY) => {
}
// which asserts that the function can still be called with the default arguments
```

Best I can tell, this works for everything except default arguments which contain a function with a pipe argument. For these functions the pipe argument becomes anonymous as it is inferred from the use of `|>` so calling the function with a normal function no longer works (`f: (?arg:(<-:int) => D) => D`) as the named argument has nothing to match with. Unsure how to reconcile this issue.

```rust
#[test]
fn infer_pipe() {
    test_infer! {
        src: r#"
            f = (arg=(x=<-) => x) => 0 |> arg()
            g = () => f(arg: (x) => 5 + x)
        "#,
        exp: map![
            "f" => "(?arg:(<-x:int) => D) => D",
            "g" => "() => int",
        ]
    }
}


// thread 'semantic::tests::infer_pipe' panicked at 'type error @3:23-3:43: missing pipe argument (argument arg)', flux-core/src/semantic/tests.rs:2849:5
```

Based on #4103 

Fixes #3461

### Done checklist
- [ ] Test cases written


nathanielc:
@Marwes  I think this is expected behavior. On master if I run a similar test but without the default value for `arg` I get the same error:

```rust
#[test]
fn infer_pipe() {
    test_infer! {
        src: r#"
            f = (arg) => 0 |> arg()
            g = () => f(arg: (x) => 5 + x)
        "#,
        exp: map![
            "f" => "(arg:(<-:int) => D) => D",
            "g" => "() => int",
        ],
    }
}

// thread 'semantic::tests::function_pipe' panicked at 'type error @3:23-3:43: missing pipe argument (argument arg)'
```

The error message could be improved but I read it to mean that the value for `arg` being provided is missing the pipe argument in its definition since nothing was marked with `=<-`. 

They way this works in my mental model is that pipe is basically a special named argument (i.e. with label `<-`), so if you call a function with a pipe then you must also provide a function value that can be called with a pipe (i.e. the `=<-` ) behavior.

That said if we manage to figure out a way to introduce positional arguments then the behavior of the pipe can be changed from the argument with `<-` as its default to be defined as the first positional argument. I believe then your test case above would pass without error. 

In other words I think this issue is more an issue with the fact that we only support named args and not because of default values.  Thoughts?

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -10,18 +10,18 @@
...
+use crate::semantic::types;
+use crate::semantic::{
+    env::Environment,
+    fresh::Fresher,
+    import::Importer,
+    infer::{Constraint, Constraints},
+    sub::{Substitutable, Substitution},
+    types::{
+        Array, Dictionary, Function, Kind, MonoType, MonoTypeMap, PolyType, PolyTypeMap, Tvar,
+        TvarKinds,
```
rockstar, line 24: This is a good refactor.

rockstar (approved):
I just kept wanting to comment "This is a good refactor" that I decided that it best to just write it here instead for the aggregate.

Marwes:
@nathanielc Yes, the error is (after I figured out what it meant) consistent. However it is a breaking change as something that worked before no longer works so I am not sure how best to address that. Even if we say that the breaking change is ok (because the example only exists in tests perhaps), it is an instance of default arguments where removing the constraint _reduces_ the amount of code that can call the function, all other instances of removing the constraints generated by default arguments strictly increases the amount of code that can call the function (say converting the function from taking a `String` to `A where A: Stringable`.

With the right addition of positional arguments maybe we can avoid this, but I suspect that there isn't a way of adding positional arguments which is both backwards compatible and addresses this problem

Marwes:
Took another look at this and since piped arguments are the only instance where a default argument makes the function more permissive we could detect those instances and keep the default arguments in inference for just those cases. Letting every other use of default argument still use this more permissive inference.

nathanielc:
I used the query archiver to explore how often a function with a default value that is a function with a pipe argument is seen in the wild. No instances of wild queries were found in the archiver. To me this says we can make this breaking change to keep the implementation and type rules consistent.

PR #4117, opened 2021-10-04
chore(test): allow optional passed in cmp.Option parameters in some tests

DStrand1:
This PR brings in a new ability to pass in cmp.Option into the Physical and Logical RuleTestHelper functions, needed in IDPE and InfluxDB to avoid panics caused by unexported protobuf fields. This allows test cases to be fixed in both IDPE and InfluxDB following some `protobuf` changes:

[IDPE example](https://github.com/influxdata/idpe/blob/395bd96e98f27cc1e3f07bb50dd62e0082410a95/query/stdlib/influxdata/influxdb/rules_test.go#L198)
[InfluxDB example](https://github.com/influxdata/influxdb/pull/22606) (in `rules_test.go`, still in a PR)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,5 +58,6 @@ require (
 	golang.org/x/tools v0.1.4
 	gonum.org/v1/gonum v0.8.2
 	google.golang.org/api v0.47.0
+	google.golang.org/protobuf v1.27.1
 	gopkg.in/yaml.v2 v2.3.0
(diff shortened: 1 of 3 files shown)


lesam:
Looks reasonable to me but someone from the flux team should review

benbjohnson (approved):

PR #4119, opened 2021-10-04
feat: embed compiled stdlib instead of compiling at runtime

jsternberg:
This reintroduces commit 9b9575705d78fe818e2d51b3318c2dd32f483e35 which
was reverted in commit 5d538445bbe5d8f9ad6a5cfe7cc5ad3d8fb5e183.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
I tested this out by modifying `math.flux` and adding a new variable to it. When that is done, only the files that import math get their bytecode versions modified. It would be good to remove this too, but this is acceptable in my opinion and was what I thought would happen with the original PR.

nathanielc (approved):

PR #4121, opened 2021-10-05
feat: add timezone package with fixed offset location

### LINKED ISSUE #4116: Abstract locations and fixed offset timezones
We want the ability to represent both fixed offset timezones and location based timezones. This means implementing both  `loadLocation` and  `fixedOffset` functions that return the same type that the `window` function takes for its location argument. 

One way to do this is with record with properties like:

```
{
    location: string,
    offset: duration
}
```

UTC would be `{location: "UTC", offset: 0s}`, and fixed offset -6 would be `{location: "UTC", offset: -6h}` and America/Denver would be `{location:"America/Denver", offset: 0s}`. 

Note sure what happens if you have location that is not fixed and a non-zero offset. Perhaps a better design would solve that inconsistency.

jsternberg:
The timezone package now exists for loading locations and fixed offsets.
The `location` function inside of `timezone` will verify the location
exists so that it doesn't have to be done by the `window()` function.

The `fixed` option creates a location in the `UTC` zone with an
additional offset that is separate from the normal offset. This offset
ends up working in reverse of the typical offset. An option of `-4h`
will set the offset to be at `-04:00` permanently as if it were in UTC.
The practical effect of this is that these two are equivalent:

    window(..., offset: 4h)
    window(..., location: timezone.fixed(offset: -4h))

Fixes #4116.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/timezone/timezone.go
```diff
@@ -0,0 +1,51 @@
...
+	mt := runtime.MustLookupBuiltinType(pkgpath, name)
+	return values.NewFunction(name, mt, func(ctx context.Context, args values.Object) (values.Value, error) {
+		return interpreter.DoFunctionCall(fn, args)
+	}, false)
+}
+
+// registerFunction will create a function from the function definition
+// and register it in this package with the given name.
+func registerFunction(name string, fn function) {
+	runtime.RegisterPackageValue(pkgpath, name, makeFunction(name, fn))
```
nathanielc, line 46: Why do we use `makeFunction` here? Why not just use `values.NewFunction` like this https://github.com/influxdata/flux/blob/a7853f9cc14cfb58c8f18841d5b6e12ac7b94b99/stdlib/json/encode.go#L16 ? 

In other words why do we need the abstraction through interpreter.DoFunctionCall?

stdlib/timezone/timezone.flux
```diff
@@ -0,0 +1,7 @@
+package timezone
+
+
+utc = {zone: "UTC", offset: 0s}
+fixed = (offset) => ({zone: utc.zone, offset: offset})
+
+builtin location : (name: string) => {zone: string, offset: duration}
```
nathanielc, line 7: These names work well. Can you add some comments to each value explaining their purpose etc?

The way I read the names are like this:

```
timezone.location(name: "XXX") // creates a timezone based on a specific location
timezone.fixed(offset: -4h) // creates a timezone that has a constant fixed offset from UTC
```

I like it.

stdlib/timezone/timezone.flux
```diff
@@ -0,0 +1,7 @@
+package timezone
+
+
+utc = {zone: "UTC", offset: 0s}
+fixed = (offset) => ({zone: utc.zone, offset: offset})
```
nathanielc, line 5: Can we should we consider disallowing mixed duration offsets? If we don't want to enforce it in implementation at least a doc comment explaining that offsets should not be mixed.

nathanielc (changes requested):
LGTM with a few small changes

stdlib/timezone/timezone.go
jsternberg: I just copied it from another package. It helps me to not repeat myself since the same variables are used in different parts. I can get rid of it but I just found it easier to copy. Might be good to abstract it out into a package.

stdlib/timezone/timezone.flux
jsternberg: Mixed duration offsets are generally ok from what I understand. We might want to clarify the order of how they get applied though. But, I'm not entirely sure I understand how the application order effects the end result. The location offset gets applied first in the current implementation.

stdlib/timezone/timezone.go
nathanielc: Which package did you copy it from. Perhaps we should simplify both locations so its doesn't get repeated.

jsternberg: Looks like I copied it from `dict`.

internal/execute/function/function.go
```diff
@@ -0,0 +1,30 @@
+package function
```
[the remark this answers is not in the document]
jsternberg: Moved.

stdlib/timezone/timezone.flux
```diff
@@ -0,0 +1,45 @@
...
+// utc is the default location with a completely linear clock
+// and no offset. It is used as the default for location related
+// options.
+utc = {zone: "UTC", offset: 0h}
+
+// fixed is a function that constructs a location with a fixed offset.
+//
+// ## Parameters
+// - `offset` is the fixed duration for the location offset.
+//   This duration is the opposite of the offset for the `window()` function.
```
nathanielc, line 15: I think it would be good to define the offset as the timezone offset from UTC instead of defining it as the opposite of window's offset. 

We can still call out the behavior that it has the opposite behavior of window's offset but I don't think we should define it that way.

nathanielc (approved):

PR #4125, opened 2021-10-07
ci: remove cargo bench from test-bench

jsternberg:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


Marwes:
Why should benchmarks be removed? It is easy that they bit-rot and stop compiling if they aren't tested.

jsternberg:
Oh I had not. I'll close this. I had no idea how to do that.

Issue #4029, opened 2021-09-10
Create a formatter for semantic graph


wolffcm:
We want to be able to write functions in Rust that accept a semantic graph
and modify it in some way.
We'll want to write unit tests that verify a semantic graph looks as expected, including its types.

If we had a way of formatting semantic graph similar to how we can format AST, unit tests
could assert that the formatted representation of a semantic graph contains a particular substring.

For example:
```rust
#[test]
fn test_type_of_plus() {
  let graph = some_function_that_produces_semantic_graph("x + 3.0");
  assert_true(graph.format().contains("x:float +:float 3.0"))
}
```

In this case the semantic graph expression is formatted with the types of all the subexpressions.
I think in general we want to format semantic graph similar to how we format Flux from AST, but there
is some discussion to be had about how to add type information to the output.

DOD:
- A formatter for semantic graph exists
- Unit tests for the semantic graph formatter exist

PR #4128, opened 2021-10-08
feat(Vectorize): create a formatter for semantic graph

skartikey:
This is an initial commit for formatting semantic graph and work is
still in progress.

Here are few working tests
// 1 + 1  =>  1:int + 1:int
// n = 1.0  =>  n = 1.0:float
// x + 3.0  =>  “x:float + 3.0:float”
// s = “string” => s = “string”:string

fixes: #4029

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    let got = format(&sem_pkg).unwrap();
+
+    // want
+    // 1 + 1  =>  1:int + 1:int
+    // n = 1.0  =>  n = 1.0:float
+    // x + 3.0  =>  “x:float + 3.0:float”
+    // s = “string” => s = “string”:string
+
+    let want = "package main\n1:int + 1:int\nn = 1.0:float\nf = n:float + 3.0:float\ns = \"string\":string";
```
wolffcm, line 26: Can you use a raw string with `r#"..."` for `want` so that it looks pretty in the source code? When we write unit tests to assert on the structure of a semantic graph, I'd imagine that's what we'll do.

libflux/flux-core/src/semantic/formatter/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+        for (i, file) in n.files.iter().enumerate() {
+            if i != 0 {
+                self.write_rune('\n');
+                self.write_rune('\n');
+            }
+            self.format_file(file, false)
+        }
+    }
+
+    fn format_monotype(&mut self, n: &MonoType) {
```
wolffcm, line 181: I was wondering if we might be able to re-use some of the code from the AST formatter. I am not sure how easy that would be... since the underlying structures are different. Any thoughts on this?

libflux/flux-core/src/semantic/formatter/tests.rs
wolffcm: Also, like we talked about I think it would be good to consider maybe omitting types where it's obvious (like in string literals). However I think we should always include the type for operators, since their type will always be inferred, as in `1:int +:int 1:int`.

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,118 @@
...
+    )
+    .unwrap();
+
+    let got = format(&sem_pkg).unwrap();
+
+    let want = r#"package main
+a = "Hello, World!":string
+b = 12:int
+c = 18.5:float
+d = 043200000000000false:duration
```
wolffcm, line 25: This formatting could be improved, can you make it so that it shows `12h:duration` instead of this? Maybe you could match what the AST formatter does... or is it that the representation of durations changes somehow between ASt and semantic graph?

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,118 @@
...
+    let got = format(&sem_pkg).unwrap();
+
+    let want = r#"package main
+a = [1:int2:int3:int]
+b = [1.1:float2.2:float3.3:float]
+c = ["1":string"2":string"3":string]
+d = [01000000000false:duration0120000000000false:duration010800000000000false:duration]
+e = [2019-10-31T00:00:00Z:datetime]
+f = [/a/:regex/b/:regex/c/:regex]
+g = [{a: 0:int, b: 0.0:float}{a: 1:int, b: 1.1:float}]"#;
```
wolffcm, line 56: These elements should have commas separating them. Can you also show the type for the array itself, like `[1:int, 2:int]:[int]`?

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let want = r#"package main
+a = "Hello, World!":string
+b = 12:int
+c = 18.5:float
+d = 043200000000000false:duration
+e = 2019-10-31T00:00:00Z:datetime
+f = /server[01]/:regex"#;
+
+    assert_eq!(want, got);
```
Marwes, line 29: I would highly recommend using https://docs.rs/expect-test/1.1.0/expect_test/ for tests like these. Makes it much easier to update the tests if formatting changes in the future and skips the need to write the expected value in the first place.

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+#[test]
+fn format_function_expression() {
+    let script = r#"
+            (a) => a
+            f = (a, b=1) => a + b
+            x = f(a:2)
+            y = f(a: x, b: f(a:x))
+            "#;
```
wolffcm, line 106: You could add a test case for a function with a pipe argument, like:
```
g = (t=<-) => t
```

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    check(
+        script,
+        expect![[r#"
+            package main
+            (a) =>{
+            return a:t17
+            }:(a:t17) => t17
+            f = (a, b=1) =>{
+            return a:int +:int b:int
```
wolffcm, line 116: It would be good to indent the body of the function if possible. If it's not easy to do, it can wait.

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,264 @@
...
+    let script = r#"
+            if 1 == 2 then 5 else 3
+            ans = if 100 > 0 then "yes" else "no"
+            "#;
+
+    check(
+        script,
+        expect![[r#"
+            package main
+            if 1 ==:bool 2 then 5 else 3
```
wolffcm, line 134: I think conditional expressions need their own type annotation that shows they type of the value that will be produced when the expression is evaluated. In this case it would be `int`. It might be formatted like this:
```
(if 1 ==:bool 2 then 5 else 3):int
```
I think there needs to be `()` so it's clear that the type annotation refers to the conditional expression. Does that make sense? Can you think of a better way to display the type?

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -0,0 +1,264 @@
...
+            o = {temp: 30.0, loc: "FL"}
+            t = o.temp
+            "#;
+
+    check(
+        script,
+        expect![[r#"
+            package main
+            o = {temp: 30.0, loc: "FL"}:{temp:float, loc:string}
+            t = o:{temp:float, loc:string}temp:float"#]],
```
wolffcm, line 197: I guess the `.` is missing just before `temp` here.

wolffcm (changes requested):
I had a few comments but you're nearly there.

PR #4129, opened 2021-10-08
WIP Add Analyzer API for fluxdoc

nathanielc:
Prior to this change there were several entry points into the type
inference code. It was not clear which one was best or why you would use
one over the other. Now there is a single entry into the inference logic
via the Analyzer type.
    
Additionally the Analyzer has been implemented with a generic Importer
allowing for either analysis against embedded stdlib data or against a
file system containing the compiled types.
    
An implementation of the Importer against the file system was
implemented along with a `fluxc` command that dumps compiled stdlib into
a directory structure. This structure is similar to how the stdlib is
embedded into Go. In fact the explicit goal is to eventually merge those
two systems into a single system, but that work is left for a later PR.
These new files are also called `.fc` files and instead of only
containing the semantic graph also contain the poly type for the
package. This PR however does not encode the semantic graph into the new
.fc files as that is not yet implemented.
    
In summary these are the changes in the PR:
    
* New Analyzer type
* Removed lots of old public functions that did analysis in different ways
* Add implementation of Importer using the FileSystem
* Updated the bootstrap module to both read source code for the stdlib
      and output compiled stdlib files
* Adds Compiled flatbuffer type for the content stored in compiled .fc
      files. (Note this does not change the Go embed process).
* Updated serialization of semantic graph to be into a partial builder
      instead of always assuming nothing else will be serialized alongside.
* Add a `fluxc` command that dumps compiled stdlib into files using the
      bootstrap module.

All of this works was motivated by the need to make a `fluxdoc` command
that can leverage the stdlib to do type inference but not have the docs
or stdlib embedded. PR coming soon for that work.


libflux/flux-core/src/bin/fluxc.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+impl From<bootstrap::Error> for Error {
+    fn from(err: bootstrap::Error) -> Error {
+        Error { msg: err.msg }
+    }
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
nathanielc: I have made this PR https://github.com/influxdata/flux/pull/4131 to remove uses of custom Error structs or String types for errors with anyhow::Error. I figured this change would be best as a separate PR. I will rebase this PR off that change.

stdlib/csv/csv.flux
```diff
@@ -64,6 +64,7 @@ package csv
 // ```
 //
 // ## Query a raw CSV string
+//
 // ```
 // import "csv"
 //
```
wolffcm, line 70: This package was in the list of exceptions for `fluxdoc`, can it be removed?

libflux/include/influxdata/flux.h
```diff
@@ -91,20 +91,20 @@ void flux_get_env_stdlib(struct flux_buffer_t *);
...
+void flux_free_stateful_analyzer(struct flux_stateful_analyzer_t *);
 
-// flux_analyze_with will analyze the ast snippet using the flux_semantic_analyzer_t and produce
+// flux_analyze_with will analyze the ast snippet using the flux_stateful_analyzer_t and produce
 // a semantic graph for that snippet.
-struct flux_error_t *flux_analyze_with(struct flux_semantic_analyzer_t *, struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
+struct flux_error_t *flux_analyze_with(struct flux_stateful_analyzer_t *, struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
 
 // flux_analyze analyzes the given AST and will populate the second pointer argument with
 // a pointer to the resulting semantic graph.
```
wolffcm, line 110: Do the Valgrind tests need to be updated as well due to this change?

wolffcm: Never mind, I see you got it.

wolffcm (approved):
This looks great to me. A very nice refactoring.

libflux/flux-core/src/ast/mod.rs
```diff
@@ -1695,5 +1709,7 @@ pub struct DateTimeLit {
     pub value: chrono::DateTime<FixedOffset>,
 }
 
+// The tests code exports a few helpers for writing AST related tests.
+// We make it public so other tests can consume those helpers.
 #[cfg(test)]
-mod tests;
+pub mod tests;
```
rockstar, line 1715: Why did we have to `pub` this?

libflux/flux-core/src/ast/tests.rs
```diff
@@ -2131,3 +2131,48 @@ fn test_ast_json_roundtrip() {
...
+pub struct Locator<'a> {
+    source: &'a str,
+    lines: Vec<u32>,
+}
+
+impl<'a> Locator<'a> {
+    /// Create a new Locator for the given source code
+    pub fn new(source: &'a str) -> Self {
+        let mut lines = Vec::new();
+        lines.push(0);
```
rockstar, line 2145: This is to pad the lines, right? So the lines are one-indexed rather than zero-indexed?

libflux/flux-core/src/ast/tests.rs
```diff
@@ -2131,3 +2131,48 @@ fn test_ast_json_roundtrip() {
     };
     assert_eq!(ast, roundtrip_ast);
 }
+
+/// Locator makes constructing an SourceLocation from a string simple.
+pub struct Locator<'a> {
```
rockstar, line 2136: I don't know how you feel about a dependency here, but influxdata/flux-lsp uses [line_col](https://docs.rs/line-col/0.2.1/line_col/) for this and, as I recall, it handles some cases that aren't the happy path, e.g. different line endings, etc.

_gets farther down in the patch_ Ah, I see, it's just moved code. Still, wanted to keep this comment to see what your thoughts were.

libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+    Ok((docs, diagnostics))
+}
+
+// HACK: Any package in this list will not report any errors from documentation diagnostics
+// The intent is that as we get each package passing the documentation diagnostics check
+// that we remove them from the list and do not add any packages to this list.
+// This way we can incrementally improve the documentation while also ensuring we are keeping a
+// high standard going forward.
+const EXECEPTIONS: [&str; 96] = [
```
rockstar, line 148: s/EXECEPTIONS/EXCEPTIONS/

nathanielc: fixed

internal/fbsemantic/semantic.fbs
```diff
@@ -446,4 +446,9 @@ table UnsignedIntegerLiteral {
   value:uint64;
 }
 
+table Compiled {
```
jsternberg, line 449: Instead of adding a new root type, can you add `polytype:PolyType` as an attribute to `table Package`? This should be backwards compatible.

libflux/c/main.c
```diff
@@ -142,10 +142,10 @@ void test_semantic() {
   printf("\n");
 }
 
-void test_semantic_analyzer() {
+void test_stateful_analyzer() {
   printf("Testing semantic analyzer...\n");
 
-  struct flux_semantic_analyzer_t *analyzer = flux_new_semantic_analyzer();
+  struct flux_stateful_analyzer_t *analyzer = flux_new_stateful_analyzer();
```
jsternberg, line 148: Probably a better name.

libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+    Ok((docs, diagnostics))
+}
+
+// HACK: Any package in this list will not report any errors from documentation diagnostics
+// The intent is that as we get each package passing the documentation diagnostics check
+// that we remove them from the list and do not add any packages to this list.
+// This way we can incrementally improve the documentation while also ensuring we are keeping a
+// high standard going forward.
+const EXCEPTIONS: [&str; 96] = [
```
jsternberg, line 148: Can you make an explicit issue for doing this and put it in here as a comment?

libflux/flux-core/src/parser/mod.rs
```diff
@@ -16,9 +16,9 @@ mod strconv;
 ///
 /// Returns a [`File`] with the value of the `name` parameter
 /// as the file name.
-pub fn parse_string(name: &str, s: &str) -> File {
+pub fn parse_string(name: String, s: &str) -> File {
```
jsternberg, line 19: Is there a reason for this change? It seems to me like the old function signature was more correct.

jsternberg (changes requested):
Minor change requested to the flatbuffers spec.

internal/fbsemantic/semantic.fbs
nathanielc: I anticipate we will add more and more information to this Compiled type as we use it more. For example we may need version information. A package has meaning outside the context that is intended to be reused for compilation. I do not want to conflate those concepts.

libflux/flux-core/src/parser/mod.rs
nathanielc: The previous implementation made a clone of the name. My general philosophy in Rust is that if a function needs to own the data it should state it in the function signature. This allows the called to decide best how to make a copy or otherwise control the allocations. Otherwise calling parse_string would always require a new allocation.

internal/fbsemantic/semantic.fbs
jsternberg: I would not interpret the type of the package to be something to maintain compiled Flux code on the filesystem. It seems to me to be directly associated with the package itself. I would agree that the package type is pure code that is compiled which, in my view, includes the types because it includes the type in multiple other places in the flatbuffers.

I would not agree with the assertion that this is metadata that is not required to understand the code. It's an open bug that we don't include this information for the Go code to access which causes the Go code to operate incorrectly in certain circumstances.

For the example of a hash of the source code, I would likely agree that is not something directly associated with the package. I also don't think that's a reason to have the polytype be associated with a container that holds the package instead of the package itself. I don't know if this would be something we even put into flatbuffers though.

> In other words I see the Package type as what is needed for the runtime to execute the code and I see the Compiled type as what is needed to be able to compile and persist the code.

If this is the case, I'm stronger on including the polytype in the package itself. The runtime is in desperate need of the polytype to correctly execute the code. This would give us access to that information and we would be able to make use of this inside of the compiler. This would instantly enable us to begin work on this improvement while also being backwards compatible in a way that doesn't require us to urgently work on this.

jsternberg (approved):
Approving in anticipation of the rename from `Compiled` to `Module`.

PR #4131, opened 2021-10-11
chore(libflux): refactor to use anyhow::Error

nathanielc:
Prior to this change the libflux code has an inconsistent use of String
and custom Error structs that just wrapped strings to represent errors.
Now with this change those instances have been replaced with
anyhow::Error as a simple error type for when we needed a plain error.

We still have 4 uses of custom error structs/enums where we have very
specific information we want to capture on the error. Perhaps we should
update those to use https://github.com/dtolnay/thiserror a companion to
anyhow in the future.

Additionally I noticed that `merge_packages` was duplicated in both the `flux` and `flux-core` crates. I have deleted the one from `flux` moved/updated its tests accordingly.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux/src/lib.rs
```diff
@@ -71,8 +75,13 @@ pub struct ErrorHandle {
...
 }
 
-impl<T: 'static + error::Error> From<T> for Box<ErrorHandle> {
-    fn from(err: T) -> Self {
+impl From<anyhow::Error> for Box<ErrorHandle> {
+    fn from(mut err: anyhow::Error) -> Self {
+        (&mut err).into()
+    }
+}
+impl From<&mut anyhow::Error> for Box<ErrorHandle> {
```
Marwes, line 83: Is the `&mut anyhow::Error` conversion really needed?

nathanielc: Yes because of how the errors are handled at the C FFI layer we end up with a `&mut anyhow::Error` that we need to capture so that Rust can drop it. See https://github.com/influxdata/flux/pull/4131/files#diff-4cd6520d44cd3f8bc7a508423e85e1a011da344bb34a1389a5a0aab459258e40R462

PR #4138, opened 2021-10-13
feat(libflux/fluxdoc): add short mode to fluxdoc dump 

nathanielc:
This change adds a feature to fluxdoc to dump Flux docs without their
descriptions in a so called _short_ mode.

Based off https://github.com/influxdata/flux/pull/4129

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/flux-core/src/semantic/doc.rs
```diff
@@ -559,6 +561,34 @@ fn comments_to_string(comments: &[ast::Comment]) -> String {
     s
 }
 
+/// Shorten removes all long form descriptions from the docs structures leaving only the headlines
+/// and other metadata.
```
rockstar, line 565: Super disappointed we didn't find a way to add "if you're not into that whole brevity thing" into the docs here. 🤪😂

libflux/flux-core/src/semantic/doc.rs
```diff
@@ -559,6 +561,34 @@ fn comments_to_string(comments: &[ast::Comment]) -> String {
...
+pub fn shorten(doc: &mut PackageDoc) {
+    doc.description = None;
+    for (_, m) in doc.members.iter_mut() {
+        remove_desc(m);
+    }
+}
+
+// removes the description from a Doc
+fn remove_desc(doc: &mut Doc) {
```
rockstar, line 574: I'm not sure that we need this function, as it's relatively short, and only called once, from the function right above it. To read this code, it would require pushing to the mental stack, and I'm not sure it provides much value for that. Thoughts?

rockstar (approved):
Just reviewed the last rev, since it was based on a another branch, but this looks good. Just  a couple of comments.

libflux/flux-core/src/semantic/doc.rs
nathanielc: The function is recursive  with the `shorten` function. Its a bit odd to have two function that together make it recursive but this keeps the code much simpler than trying to convert the types into something that can be recursive with a single function. In short the Doc enum owns its data but we need a `&mut` to the data so that we don't have to make copies of it just to remove data.
 
 I'll add a comment to this effect.

libflux/flux-core/src/semantic/doc.rs
nathanielc: Had to look it up :D

PR #4139, opened 2021-10-13
refactor(quantile): Allow quantile to handle multiple states

### LINKED ISSUE #4042: Refactor quantile to only recycle memory when it calculates a complete result

scbrickley:
***This branch was cut from `feat/disposable-transformations`, so #4134 will need to merge before this does.***

Closes #4042

This patch makes `quantile` compatible with the new aggregate state transformation by allowing it to handle states for multiple group keys, while still recycling memory.

Changes included in this patch:
- Move the `TDigest` out of `QuantileAgg` and into a new data structure called `QuantileAggState`, which implements the `ValueFunc`, `DoFloatAgg`, and `FloatValueFunc` interfaces.
- Add a list of re-usable `TDigests` as a field to `QuantileAgg`
- Implement `Disposable` on `QuantileAggState` so that, when the execution engine is done with it, the digest is reset and handed off to `QuantileAgg` to be re-used by future quantile aggregates.
- Implement `Disposable` on `QuantileAgg` to account for memory no longer being used by the quantile transformation.
- Return a `QuantileAggState` out of `NewFloatAgg`

**Benchmarks**
*This does result in slightly degraded performance from the current quantile implementation, but it's still much better than the performance of the original implementation. See benchmarks from https://github.com/influxdata/flux/pull/3497 for comparison.*

```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkQuantile-12    	       6	 183711173 ns/op	  337664 B/op	      13 allocs/op
PASS
ok  	github.com/influxdata/flux/stdlib/universe	1.765s
```

### DIFF
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -320,6 +320,9 @@ func (t *simpleAggregateTransformation) Process(id DatasetID, tbl flux.Table) er
 			if err := builder.AppendNil(bj); err != nil {
 				return err
 			}
+			if vf, ok := vf.(Disposable); ok {
+				vf.Dispose()
+			}
 			continue
 		}
 
@@ -350,6 +353,9 @@ func (t *simpleAggregateTransformation) Process(id DatasetID, tbl flux.Table) er
 			if err := builder.AppendString(bj, v); err != nil {
 				return err
 			}
+		}
+		if vf, ok := vf.(Disposable); ok {
+			vf.Dispose()
 		}
 	}
 


stdlib/universe/quantile.go
```diff
@@ -258,32 +252,62 @@ func (a *QuantileAgg) NewUIntAgg() execute.DoUIntAgg {
...
 
 func (a *QuantileAgg) NewStringAgg() execute.DoStringAgg {
 	return nil
 }
 
-func (a *QuantileAgg) DoFloat(vs *array.Float) {
+func (a *QuantileAgg) Dispose() {
+	for i := 0; i < len(a.freeDigests); i++ {
+		a.mem.Account(tdigest.ByteSizeForCompression(a.Compression) * -1)
+	}
```
wolffcm, line 274: Can you add an assignment here:
```go
a.freeDigests = nil
```
This way Go's garbage collector will see that the slice is unreachable and will be more likely to free the memory sooner. It might not make much of a difference, but it can't hurt.

stdlib/universe/quantile.go
```diff
@@ -258,32 +252,62 @@ func (a *QuantileAgg) NewUIntAgg() execute.DoUIntAgg {
...
-func (a *QuantileAgg) IsNull() bool {
-	return !a.ok
+func (s *QuantileAggState) IsNull() bool {
+	return !s.ok
+}
+
+func (s *QuantileAggState) Dispose() {
+	if len(s.parent.freeDigests) < cap(s.parent.freeDigests) {
+		s.digest.Reset()
+		s.parent.pushFreeDigest(s.digest)
```
wolffcm, line 307: You might consider pushing all of this logic into `QuantileAgg` so that all the code that works with `freeDigests` is in methods of the enclosing struct. I think the code would read better that way.

stdlib/universe/quantile_agg_test.flux
```diff
@@ -0,0 +1,65 @@
...
+        |> group(columns: ["t0"])
+
+    got = inData
+        |> range(start: 2018-05-22T20:00:00Z, stop: 2018-05-22T20:01:00Z)
+        |> group(columns: ["t0"])
+        |> quantile(q: 0.75, method: "estimate_tdigest")
+        |> drop(columns: ["_start", "_stop"])
+
+    testing.diff(want: want, got: got) |> yield()
+}
```
wolffcm, line 50: So does this test actually reproduce the issue that was occurring?

PR #4148, opened 2021-10-13
feat(fluxdoc): updates fluxdoc parser to capture more data

### LINKED ISSUE #4111: Add `is_option` to ValueDoc and FunctionDoc structs
We need to know if a package members, either ValueDoc or FunctionDoc, are options. We should add a boolean `is_option` to the ValueDoc and FunctionDoc structs.

### LINKED ISSUE #4113: Support aribitrary metadata in doc comments
We want to support arbitrary metadata in doc comments. A motivating use case is to mark when a function was introduce or when it was deprecated.

The proposal is to use HTTP Header style metadata as the last lines of the comment.

For example:

```
// Package foo does a thing
//
// introduced: 0.67.0
package foo

// bar is a number
//
// deprecated: 0.89.0
bar = 1

// fiz is a boolean
// 
// key: value
fiz = false
```

This key value metadata must come at the end of the doc comment and will be added to the doc structures accordingly.

nathanielc:
Now the fluxdoc parser:

* Records if a value/function is an option
* Records the source location being documented
* Records any metadata specified in the comments

Fixes #4111
Fixes #4113

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/libflux/flux-core/src/semantic/doc.rs
+++ b/libflux/flux-core/src/semantic/doc.rs
@@ -1,6 +1,8 @@
 //! Generate documentation from source code comments.
 
+use lazy_static::lazy_static;
 use pulldown_cmark::{Event, OffsetIter, Parser as MarkdownParser, Tag};
+use regex::Regex;
 use std::collections::BTreeMap;
 use std::ops::Range;
 
@@ -27,6 +29,9 @@ pub struct Diagnostic {
 /// Diagnostics is a set of diagnostics
 pub type Diagnostics = Vec<Diagnostic>;
 
+/// Metadata is arbitrary key value data associated with documentation.
+pub type Metadata = BTreeMap<String, String>;
+
 /// Doc is an enum that can take the form of the various types of flux documentation structures through polymorphism.
 #[derive(PartialEq, Debug, Serialize, Deserialize)]
 #[serde(tag = "kind")]
@@ -35,8 +40,6 @@ pub enum Doc {
     Package(Box<PackageDoc>),
     /// Value represents documentation for a value exposed from a package.
     Value(Box<ValueDoc>),
-    /// Option represents documentation for a option value exposed from a package.
-    Opt(Box<ValueDoc>),
     /// Function represents documentation for a function value exposed from a package.
     Function(Box<FunctionDoc>),
 }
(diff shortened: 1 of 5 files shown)


libflux/flux-core/src/semantic/doc.rs
```diff
@@ -170,23 +190,23 @@ fn parse_package_values(
     let mut members: BTreeMap<String, Doc> = BTreeMap::new();
     let mut diagnostics: Diagnostics = Vec::new();
     for stmt in &f.body {
-        if let Some((name, comment, loc)) = match stmt {
+        if let Some((name, comment, loc, is_option)) = match stmt {
```
jsternberg, line 193: Can we change `is_option` into an enum? The enum can have an `is_option()` function attached to it that returns a boolean, but I'd like it if this was typed.

jsternberg:
I meant the Rust type checker. It would make it more explicit what the type was representing rather than just a boolean value that was only identifiable by the name. It prevents the "returning 17 booleans from a function" problem. Considering the function already has 3 return values, I'm getting a bit concerned with a 4th. Once the number of return values in the tuple goes above 3 I get very cautious about adding more.

nathanielc:
Ok I understand your concern with the complexity of the Rust code. I am of the opinion it is best as is. The most important aspect to me is how the data is encoded in JSON and therefore consumed by downstream projects as that is the primary purpose of this code. Whether a value is an option is binary, either it is or isn't. Adding a `variable_type` to the ValueDoc struct doesn't improve that. Say in the future we introduce new properties for example public/private status. Its likely those properties are orthogonal to being an option so we would end up with PublicOption, PrivateOption , and PublicNormal etc. Being an option isn't necessarily connected to the visibility of the value. However using a `Visibility` enum for `public` vs `private` would make sense as you may have `protected` or something similar in the future. But for options it truly is a binary property and I don't want to conflate it with other properties of the value. Thoughts?

libflux/flux-core/src/semantic/doc.rs
```diff
@@ -150,13 +167,17 @@ fn parse_file_doc_comments(
     };
     let (members, mut diags) = parse_package_values(file, types)?;
     diagnostics.append(&mut diags);
+    let (description, metadata, mut diags) = parse_metadata(description, &file.base.location);
+    println!("description {:#?}", &description);
```
jsternberg, line 171: This looks like a debugging println.

PR #4149, opened 2021-10-14
perf(execute): Avoid recompiling types for every Process call

Marwes:
When executing `Process` on a table with few rows the time to `Compile`
the function before execution ends up dominating the total execution
time. This is an issue for data produced by `csv.from` which creates one
`Table` per row and might be for other sources as well though I have no
idea how common it would be.

For running the following query on a file with 500k lines this change
reduces the runtime from ~38s to ~16s.

```
import "csv"

csv.from(file: "../indata_dup.csv")
    |> filter(fn: (r) => r.Animal_name == "crow")
    |> count()
```

To make `compileFn` immutable after compilation (at least as far as I
and `make test` understands) I changed it to only hold the parent scope
and only construct the scope used during evaluation when `Eval` is
actually called.

### DIFF
--- a/execute/row_fn.go
+++ b/execute/row_fn.go
@@ -2,6 +2,7 @@ package execute
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
@@ -17,6 +18,15 @@ type dynamicFn struct {
 	scope      compiler.Scope
 	fn         *semantic.FunctionExpression
 	recordName string
+	compiledFn *compiledFn
+}
+
+type compiledFn struct {
+	fn         compiler.Func
+	inType semantic.MonoType
+	recordType semantic.MonoType
+	cols []flux.ColMeta
+	extraTypes map[string]semantic.MonoType
 }
 
 func newDynamicFn(fn *semantic.FunctionExpression, scope compiler.Scope) dynamicFn {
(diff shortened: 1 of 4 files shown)


execute/row_fn.go
```diff
@@ -44,35 +54,45 @@ func (f *dynamicFn) typeof(cols []flux.ColMeta) (semantic.MonoType, error) {
 }
 
 func (f *dynamicFn) prepare(cols []flux.ColMeta, extraTypes map[string]semantic.MonoType) (preparedFn, error) {
-	// Prepare the type of the record column.
-	recordType, err := f.typeof(cols)
-	if err != nil {
-		return preparedFn{}, err
-	}
+	// If the types have not changed we do not need to recompile, just use the cached version
+	if f.compiledFn == nil || !reflect.DeepEqual(f.compiledFn.cols, cols) || !reflect.DeepEqual(f.compiledFn.extraTypes, extraTypes) {
```
nathanielc, line 58: Should we implement an explicit equality measure for these values instead of using reflect.DeepEqual?

execute/row_fn.go
```diff
@@ -17,6 +18,15 @@ type dynamicFn struct {
 	scope      compiler.Scope
 	fn         *semantic.FunctionExpression
 	recordName string
+	compiledFn *compiledFn
```
nathanielc, line 21: Is my reading of the code correct that we have a cache of size 1? Basically in the case where the type is always the same we do not have to recomiple, but in cases where it changes we recompile each time it changes. So worst case is that we recompile everytime (which is what we do now anyway) and best case is we only compile once. LGTM

execute/row_fn.go
Marwes: Didn't want to write it out in case this didn't work. fixed.

nathanielc (approved):

PR #4151, opened 2021-10-15
refactor: Prepare type inference for a persistent Substitution

Marwes:
A couple of refactoring commits (may be easiest to review commit by commit). This changes the type inference code slightly to remove some patterns that rely on `Substitutions` being small and cheap to create or duplicate. This also changes the `fn infer` signature to just need a single object with the state that is needed which makes it easier to add more state (just add another field instead of adding another parameter to each function call and definition).

### DIFF
--- a/libflux/flux-core/src/semantic/nodes.rs
+++ b/libflux/flux-core/src/semantic/nodes.rs
@@ -95,7 +95,7 @@ impl From<infer::Error> for Error {
     }
 }
 
-struct Infer<'a> {
+struct InferState<'a> {
     f: &'a mut Fresher,
     env: Environment,
 }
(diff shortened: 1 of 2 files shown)


libflux/flux-core/src/semantic/types.rs
```diff
@@ -82,15 +84,22 @@ impl PartialEq for PolyType {
...
+                    None
+                } else {
+                    sub.try_apply(var)
+                }
+            })
+            .map(|expr| PolyType {
+                vars: self.vars.clone(),
+                cons: self.cons.clone(),
+                expr,
+            })
```
wolffcm, line 102: The way I understand this, is if the polytype had any quantified variables, we used to remove those variables from the substitution before applying it. 

Now we instead create a substituter that will filter out the quantified variables on the fly as we recurse into the type expression.

Do I have that right?

wolffcm (approved):
Very nice, I just had a comment about the struct name. Overall a big improvement I think.

libflux/flux-core/src/semantic/types.rs
Marwes: Yep, it is the same effect in the end in that we create a "view" into the substitution that lacks those variables and therefore won't replace them

PR #4156, opened 2021-10-18
fix(libflux): make checking for AST and semantic errors configurable for Analyzer

nathanielc:
Prior to this change any consumers of the Analyzer would always get an
error in the face of a AST or semantic check error. Now this behavior is
configurable for use cases that may want to do a best effort approach
like the LSP.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux-core/src/semantic/tests.rs
```diff
@@ -3566,3 +3578,23 @@ fn test_error_messages() {
         err: "error @3:13-3:20: missing required argument b",
     }
 }
+
+#[test]
+fn test_analyzer_skip_checks() {
+    // Test that if we skips checks we do not get an error and that we can analyze
+    // the paritial source code.
```
rockstar, line 3585: s/paritial/partial/

Issue #4098, opened 2021-09-30
Increase function loses data on counter resets


sanderson:
`increase()` is designed to normalize counter metrics (prometheus counters) by accounting for counter resets. With the current implementation, resets are always treated as 0-sum changes, even though there may be increases after the reset and before the next collection interval.

For example:

```js
import "array"

rows = [
  {_time: 2021-01-01T00:00:00Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 4302.0},
  {_time: 2021-01-01T00:00:10Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 4844.0},
  {_time: 2021-01-01T00:00:20Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 5091.0},
  {_time: 2021-01-01T00:00:30Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 13.0},
  {_time: 2021-01-01T00:00:40Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 215.0},
  {_time: 2021-01-01T00:00:50Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 762.0},
  {_time: 2021-01-01T00:01:00Z, _measurement: "prometheus", _field: "http_query_request_bytes", _value: 1108.0},
]

array.from(rows)
  |> increase()
```

Returns:

| _time                | _measurement | _field                   | _value |
| :------------------- | :----------- | :----------------------- | -----: |
| 2021-01-01T00:00:10Z | prometheus   | http_query_request_bytes |    542 |
| 2021-01-01T00:00:20Z | prometheus   | http_query_request_bytes |    789 |
| 2021-01-01T00:00:30Z | prometheus   | http_query_request_bytes |    789 |
| 2021-01-01T00:00:40Z | prometheus   | http_query_request_bytes |    991 |
| 2021-01-01T00:00:50Z | prometheus   | http_query_request_bytes |   1538 |
| 2021-01-01T00:01:00Z | prometheus   | http_query_request_bytes |   1884 |

It does normalize the counter reset, but always to zero. It doesn't account for values greater than 0 on reset. Ideally, it would add values less that the previous value but greater than 0 to the previous value. So the output would actually look like:

| _time                | _measurement | _field                   | _value |
| :------------------- | :----------- | :----------------------- | -----: |
| 2021-01-01T00:00:10Z | prometheus | http_query_request_bytes | 542 |
[...]

PR #4168, opened 2021-10-22
feat(stdlib): new parameters added to difference function to support increase function

skartikey:
**Issue -** 
increase() is designed to normalize counter metrics (Prometheus counters) by accounting for counter resets. With the current implementation, resets are always treated as 0-sum changes, even though there may be increases after the reset and before the next collection interval. It does normalize the counter reset, but always to zero. It doesn't account for values greater than 0 on reset. Ideally, it would add values less than the previous value but greater than 0 to the previous value. 
Kindly see the issue linked to this PR for more details and examples. 

**Implementation -** 
BREAKING CHANGE: 

A new boolean parameter `initialZero: true` has been added to the increase function with functionality as mentioned below

1. **(existing behavior)**
    The first row is never preserved.
    
    **(new behavior)**
    The first row will be kept and have a value of zero.

2. Subtraction rules for numeric types.

    The difference between two non-null values is their algebraic difference; **(same for existing and new)**
    
    **(existing behavior)**
    OR null, if the result is negative and _nonNegative: true_
    
    **(new behavior)**
    OR current value, if the result is negative and the current value is greater than equal to zero and _nonNegative: true_
    OR null, if the result is negative and the current value is less than zero and _nonNegative: true_

fixes: #4098


jsternberg (commented):
I'm not completely sure I understand the new parameters. I'm also a bit confused about why two parameters are needed.

My understanding is that we would add one new parameter. This parameter would specify the seeding behavior when the difference was reset. When the difference is first computed, the first row is used to seed difference. When a reset happens, we would use the initial value to seed difference instead of the next row. The initial parameter would really only be applicable when `nonNegative: true` is set and would otherwise be useless.

This seems to be what `currentValOrZero` might be doing. I am not sure what `initialZero` is for.

It also seems to me like using this new option should be the default for increase and we don't need to add a parameter to increase.

skartikey:
@jsternberg Kindly read my previous [comment](https://github.com/influxdata/flux/pull/4168#issuecomment-952749886) for the changes requested in the issue. 

The difference function has existing param keepFirst which has similar behaviour to initialZero with only difference is that keepFirst stores null in the first row. For the backward compability reason I have created a new one initialZero to keep the existing behaviour of keepFirst. 

My reason for giving the option in the increase is to provide more flexibility to the user. Like someone wants to keep the previous behavior. 

If you have sometime today lets chat and resolve this quickly.

wolffcm (commented):
I think we should have a conversation about this. I'll add an item to the agenda.

I tend to agree that this new behavior is better and should be the default, but as you've noted there is a backwards compatibility problem if we do that.

It seems strange that the new parameter does two things:
- Adds an initial row to the output
- Changes the output when the difference is negative
Two parameters seems like overkill, but I'm also struggling to come up with a parameter name that makes sense.

stdlib/universe/difference.go
```diff
@@ -184,12 +197,12 @@ func (t *differenceTransformation) Process(id execute.DatasetID, tbl flux.Table)
...
 			return err
 		}
-		differences[j] = newDifference(t.nonNegative)
+		differences[j] = newDifference(t.nonNegative, t.initialZero)
 	}
 
 	// We need to drop the first row since its difference is undefined
 	firstIdx := 1
-	if t.keepFirst {
+	if t.keepFirst || t.initialZero {
```
jsternberg, line 205: This seems incorrect. The initial zero does not imply keep first. Initial zero changes the behavior of keep first if it is specified, but otherwise does nothing if keep first is false.

stdlib/universe/difference.go
```diff
@@ -281,7 +302,11 @@ func (t *differenceTransformation) Process(id execute.DatasetID, tbl flux.Table)
 							return err
 						}
 					} else {
-						if err := builder.AppendNil(j); err != nil {
+						if i == 0 && t.initialZero {
+							if err := builder.AppendFloat(j, float64(0)); err != nil {
```
jsternberg, line 306: Nitpick: `float64(0)` is not needed here. As the function takes a float, Go will assume the constant is a float.

stdlib/universe/difference.go
```diff
@@ -344,8 +371,15 @@ func (d *difference) updateInt(v int64, valid bool) (int64, bool) {
 		d.valid = true
 		return 0, false
 	}
-	if diff := v - prev; diff >= 0 || !d.nonNegative {
+	diff := v - prev
+	if diff >= 0 || !d.nonNegative {
 		return diff, true
+	} else if diff < 0 && d.nonNegative && d.initialZero {
```
jsternberg, line 377: The `diff < 0` is unnecessary as it's redundant. Since the previous condition checks using `||`, then diff must be less than zero.

It also seems to me like the line on 381 doesn't make sense to me. In the case that the next value is less than the initial value of zero, I think we should go with the default behavior where we return `0, false` and output a null column.

So I think this condition should be changed to: `d.nonNegative && d.initialZero && v >= 0` and then you can just `return v, true` and fallthrough to the next behavior.

stdlib/universe/difference.go
```diff
@@ -18,6 +18,7 @@ type DifferenceOpSpec struct {
 	NonNegative bool     `json:"nonNegative"`
 	Columns     []string `json:"columns"`
 	KeepFirst   bool     `json:"keepFirst"`
+	InitialZero bool     `json:"initialZero"`
```
jsternberg, line 21: Are we going to have this value be a boolean or a customizable value?

jsternberg (changes requested):

stdlib/universe/difference.go
skartikey: The boolean value would be sufficient if we just want to give an initial value to zero. 
Wondering if giving some other initial value will make sense in our use case?

stdlib/universe/difference.go
skartikey: I have implemented it in such a way that both are independent of each other. If both are true then initialZero will take priority. Kindly check this test case - https://github.com/influxdata/flux/blob/24dbe5474595fb8420990b80c49f8debfa4772e0/stdlib/universe/difference_test.go#L970

Kindly let me know if you want me to make initialZero dependent on keepFirst.

stdlib/universe/difference.go
skartikey: _"The diff < 0 is unnecessary as it's redundant. Since the previous condition checks using ||, then diff must be less than zero."_
Adding `diff < 0` was intentional for the code readability, but if it seems obvious, I will remove that. 

_It also seems to me like the line on 381 doesn't make sense to me. In the case that the next value is less than the initial value of zero, I think we should go with the default behavior where we return 0, false and output a null column._
I have implemented this as per the requirement in the [issue](https://github.com/influxdata/flux/issues/4098). I have used the exact same table as mentioned in the issue in the flux [test](https://github.com/influxdata/flux/blob/24dbe5474595fb8420990b80c49f8debfa4772e0/stdlib/universe/increase_test.flux#L69). I can implement it in the way you suggested. We can discuss it with Scott if required.

stdlib/universe/difference.go
jsternberg: I would prefer if they were independent of each other. The implementation seems to do the opposite. My understanding is that if initial zero is true and keep first is false, the first row will _not_ be materialized. This seems to always materialize the first row as zero whenever initial zero is specified.

As an example, imagine an input of `[20, 30, 50, 30, 40]`. I'd expect the following outputs given the parameters:

```
(keepFirst: false, initialZero: false) -> [10, 20, null, 10]
(keepFirst: true, initialZero: false) -> [null, 10, 20, null, 10]
(keepFirst: false, initialZero: true) -> [10, 20, 30, 10]
(keepFirst: true, initialZero: true) -> [0, 10, 20, 30, 10]
```

stdlib/universe/difference.go
jsternberg: If this is a requirement or the behavior of how prometheus's increase function works, I think I would prefer whatever that output is. I'm not sure I see this requirement in the comment you linked. It seems like @sanderson is talking about the first row while I'm referring to the reset behavior when the reset value is less than zero. As an example of input: `[10, 20, -10, 20]`. Since we're saying the initial value is zero when a reset happens, but the new value is less than zero, but we said non negative is true, we shouldn't produce negative values. But, producing a value of zero seems wrong to me since that's not really the math that was used. It seems safest to me to just output null in this case.

stdlib/universe/difference.go
skartikey: I have updated the code accordingly and added the test cases for all 4 combinations.

stdlib/universe/difference.go
skartikey: That bit was not clear to me either so I have used the query provided in the issue. Below is the excerpt from the query. 
`|> map(fn: (r) => ({ r with diff: if (not exists r.diff and r._value >= 0) then r._value else if (not exists r.diff and r._value < 0) then float(v: 0) else r.diff }))`
As suggested, I have updated the code to output null in this case.

stdlib/universe/difference.go
```diff
@@ -255,7 +272,11 @@ func (t *differenceTransformation) Process(id execute.DatasetID, tbl flux.Table)
 							return err
 						}
 					} else {
-						if err := builder.AppendNil(j); err != nil {
+						if i == 0 && t.keepFirst && t.initialZero {
```
jsternberg, line 275: I think this implementation will end up being wrong when there are multiple buffers. You're going to need to modify difference's `updateInt` and related methods itself to return zero for the initial value.

Consider if there are multiple buffers and each one is 10 rows long. On the 11th row, you'll have the first value from the second buffer. This code will consider that value to be the "first" row and potentially append zero instead of null. The `difference` struct keeps the state about what the value presently is.

stdlib/universe/difference.go
```diff
@@ -344,8 +371,11 @@ func (d *difference) updateInt(v int64, valid bool) (int64, bool) {
 		d.valid = true
```
jsternberg, line 371: I think this is the line where you'll need to add the initial zero logic for the first point. I think you can just change this to:

```
if !d.valid {
    d.valid = true
    return 0, d.initialZero
}
```

If `keepFirst` is false, the true/false value of this doesn't matter. If `keepFirst` is true, then it will see true and append the integer as the return value.

I think it's ok to leave the beginning of this function the way that it is and not change it to include initial zero. I think that's consistent with the current behavior which is that seeing a null will result in a null value being appended.

jsternberg (changes requested):

stdlib/universe/universe.flux
```diff
@@ -238,7 +245,7 @@ aggregateWindow = (
 // a threshold or are reset. In the case of a wrap/reset,
 // we can assume that the absolute delta between two points will be at least their non-negative difference.
 increase = (tables=<-, columns=["_value"]) => tables
-    |> difference(nonNegative: true, columns: columns)
+    |> difference(nonNegative: true, columns: columns, keepFirst: true, initialZero: true)
```
jsternberg, line 248: My understanding was that `increase` doesn't default `keepFirst` to true? I'm ok if we expose it as an option, but I don't know if this should be the default.

stdlib/universe/difference_test.go
```diff
@@ -723,3 +723,242 @@ func TestDifference_Process(t *testing.T) {
...
+			}},
+			want: []*executetest.Table{{
+				ColMeta: []flux.ColMeta{
+					{Label: "_time", Type: flux.TTime},
+					{Label: "a", Type: flux.TFloat},
+					{Label: "b", Type: flux.TFloat},
+					{Label: "c", Type: flux.TFloat},
+				},
+				Data: [][]interface{}{
+					{execute.Time(1), nil, 1.0, nil},
```
jsternberg, line 905: My understanding is that this should have `a = 6.0`? If it's initial zero, then the difference is seeded at zero. We encounter a null so we would emit a null if `keepFirst` was true, but since `keepFirst` is false we swallow it (hence not in the output). Then we see time 1, see a 6, and `6 - 0 = 6`.

stdlib/universe/difference_test.go
```diff
@@ -723,3 +723,242 @@ func TestDifference_Process(t *testing.T) {
...
+					{execute.Time(2), 4.0, 0.0, 2.0},
+					{execute.Time(3), 6.0, 8.0, 2.0},
+					{execute.Time(4), nil, nil, 1.0},
+				},
+			}},
+		},
+		{
+			name: "float with tags and InitialZero",
```
jsternberg, line 913: This looks like the same test case as the one above it. It might be useful to have this same test case with `keepFirst: true`. Was that the intention or is there a difference I'm not seeing?

jsternberg (changes requested):

stdlib/universe/universe.flux
skartikey: That is correct `keepFirst` was not in the default option to increase function. But to achieve the first row with all zero and satisfy all below four conditions in the `difference` function I have to introduce it to the increase function option. 

```
(keepFirst: false, initialZero: false) -> [10, 20, null, 10]
(keepFirst: true, initialZero: false) -> [null, 10, 20, null, 10]
(keepFirst: false, initialZero: true) -> [10, 20, 30, 10]
(keepFirst: true, initialZero: true) -> [0, 10, 20, 30, 10]
```

stdlib/universe/difference_test.go
skartikey: we are not seeding difference to zero for the case `(keepFirst: false, initialZero: true)`. when both of them are true then only 0 value is returned. https://github.com/influxdata/flux/blob/cda99ce8b5c527710b34389b74d853b53c35bea2/stdlib/universe/difference.go#L410

The current behaviour is similar to existing [difference](https://docs.influxdata.com/flux/v0.x/stdlib/universe/difference/) which is 
`Some value v minus null is v minus the last non-null value seen before v; "or null if v is the first non-null value seen".`

jsternberg: That's my point. When initial zero is true, the last non-null value is always zero. That makes me think that, to be consistent, this should output 6.

jsternberg: Never mind on this. I think I got myself confused. We should not seed the first value with zero as that can create a large bump at the beginning. I'll rereview with that in mind.

jsternberg (approved):

Issue #2159, opened 2019-11-14
EPIC: Be able to filter data based on type


jsternberg:
We want the ability to filter data based on the type of the data. This is useful when writing downsample like tasks that operate on a dynamic set of data. For example a user could write this task to downsample all fields in a measurement.

```
from(bucket:"foo")
    |> range(start: -5m)
    |> filter(fn: (r) => r._measurement == "mydata")
    |> mean()
    |> to(bucket:"foo_5m")
 ```
 
 This works well so long as all fields are numeric. As soon as one field is a string or some other numeric type the entire task will fail with an error an no data will be downsampled.
 
 Therefore we need the ability to filter and otherwise interact with the type of the data.
 
 
 ## Proposal 
 
 
This work will have two major parts:

* Add features to Flux to be able to express filters on the type of data
* Add support to the storage layers for filtering based on type of data

For the first part the proposal is to add a `typeof` function to the language that returns a value whose type is `type`. 

```
package types

builtin of : (v:A) => type

// float is the a type for floating point values.
builtin float : type
```

Where `type` is a new type added to Flux that represents a type. Values of type `type` will be equatable. This way users can express the above Flux script safely as:

```
import "types"

from(bucket:"foo")
    |> range(start: -5m)
    |> filter(fn: (r) => r._measurement == "mydata" and types.of(v:r._value) == types.float
    |> mean()
    |> to(bucket:"foo_5m")
 ```
 
### Some open design questions:
 
 * Should we use `types.of` as a function or an operator? Using an operator makes it easier for the compiler/static pass to optimize away any typeof expression that have a known type staticaly
 * Should we expose kind constraints via this type system? Perhaps with a set of functions that can query whether a type satisfies a constraint? For example `types.isNumeric(t: types.of(v: r._value))`, this will allow uint, int and float types to be selected in a filter expression.

sanderson:
If I could upvote this 100 times, I would! Have wanted something like this for a while.

nathanielc:
@sanderson Can you share some concrete use cases for why this would be so helpful? The one use case I can think of is filtering data based on its type. Is that the only one or are there others?

Also FWIW all the work we are doing on Algo W will directly enable this kind of feature in the future.

sanderson:
Filtering by type for downsampling tasks. This way you can use a different aggregate for numeric types vs strings.

```
from(bucket: "telegraf")
  |> range(start: -24h)
  |> filter(fn: (r) => typeOf(v: r._value) != "string")
[...]

PR #4178, opened 2021-10-26
feat: Add is_type to query the runtime type

Marwes:
Proposal for #2159. Only capable of answering true for primitive types.
Other proposals include `isFloat` etc but I went with this as it was
quicker and is easily convertible to that approach

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/is_type.go
```diff
@@ -0,0 +1,73 @@
...
+				// case semantic.Array:
+				// 	result = type_ == "array"
+				// case semantic.Object:
+				// 	result = type_ == "object"
+				// case semantic.Function;
+				// 	result = type_ == "function"
+				// case semantic.Dictionary:
+				// 	result = type_ == "dictionary"
+				// case semantic.Vector:
+				// 	result = type_ == "vector"
```
nathanielc, line 64: Should we remove/update this comment to explain that we have explicitly decided to not support any composite types?

stdlib/universe/universe.flux
```diff
@@ -214,6 +214,11 @@ builtin logarithmicBins : (start: float, factor: float, count: int, ?infinity: b
 // die returns a fatal error from within a flux script
 builtin die : (msg: string) => A
 
+// Returns true if `v` is an object matching `type`.
+// This function can only check against the primitive types:
+// `string`, `bytes`, `int`, `uint`, `float`, `bool`, `time`, `duration`, `regexp`.
+builtin isType : (v: A, type: string) => bool
```
nathanielc, line 220: I like the simplicity of this function. Any reason we would put this into its own package? Generally universe is quite large so I'd prefer we don't add to it but happy to be convinced otherwise.

nathanielc (changes requested):
Looks good some questions below.

Also can you add a Flux acceptance test (aka using the `testcase` statement) for this function?

stdlib/universe/universe.flux
Marwes: I just put it here for simplicity, I don't have an opinion or context towards where it would make the most sense to put this :confused:

Marwes:
> Also can you add a Flux acceptance test (aka using the testcase statement) for this function?

I added one, but I don't see how they are actually executed? The `convert` pass seems to error if it encounters a `testcase` and I haven't been able to find any code that transforms testcases before that.

nathanielc:
Flux acceptance tests are run via the `./cmd/flux test` command in this repo. The `testcase` conversion happens on the Go side of things before we even hand off the AST to Rust.

Marwes:
> Flux acceptance tests are run via the ./cmd/flux test command in this repo. The testcase conversion happens on the Go side of things before we even hand off the AST to Rust.

Why does go have access to the AST? I thought go handed Rust a string, which is parsed, etc and then go gets a semantic graph back?

Marwes:
A thought, should we add another `Kind` for this like `Builtin` or `Primitive`. If we have `isType: (v: A, type: string) => bool where A: Primitive` we know that users can't call it with objects, arrays etc.

Marwes:
I saw some code using `assert.equal` since I found it in some other testcases, however it doesn't seem to actually exist. The only assertion functions that exist only operate on arrays. `[A]`

nathanielc:
> Why does go have access to the AST? I thought go handed Rust a string, which is parsed, etc and then go gets a semantic graph back?

Go originally had access to the AST and we have maintained that access. For example you can pass  AST via the API to run a query instead of just a string allowing rich clients to construct queries safely. 

In the case of this `testcase` code we needed to have Go do the AST manipulation since it needed access to the AST of imports (i.e. other test cases) and Rust code doesn't have filesystem access. 

@rockstar You might have some more context on this :point_up: . 

> A thought, should we add another Kind for this like Builtin or Primitive. If we have isType: (v: A, type: string) => bool where A: Primitive we know that users can't call it with objects, arrays etc.

That's an interesting idea, I like it I am trying to think if it would be used in any other cases. Since I pinged @rockstar already curious about your thoughts on a change to add a new Primitive constraint. 

Also we might want to call it `Basic` instead of Primitive as I think that's what we call it in the code, but we can decide that later if there are strong opinions.

rockstar:
> In the case of this testcase code we needed to have Go do the AST manipulation since it needed access to the AST of imports (i.e. other test cases) and Rust code doesn't have filesystem access.

It was less about filesystem access and more about purposefully making `testcase` an AST-only thing, i.e. it doesn't even have semantic graph data structure representation, so will error if it gets to semantic analysis. Go sees the testcase and transforms each test case into its own file/package.

Marwes:
Ok I found the code in `ast/testcase/testcase.go` was mostly wondering out of confusion :).

Anyway, what can I use for assertions as `assert.equal` does not seem to exist

stdlib/testing/testing.flux
```diff
@@ -342,3 +343,14 @@ benchmark = (case) => {
...
     return tc.input |> tc.fn()
 }
+
+// assertEquals tests whether two values are equal
+//
+// ## Parameters
+// - `got` is the value to test.
+// - `want` is the expected data to test against.
+//
+assertEqualValues = (got, want) => {
```
nathanielc, line 353: I really like this addition, and we can update it once the testing framework is updated to not expect tables as output.

nathanielc (approved):

PR #4182, opened 2021-10-26
fix(join): Compare sorted join keys

### LINKED ISSUE #4150: Failing to join on multiple fields in a join between InfluxDB and SQL data streams
* Full details of your operating system (or distribution):
  * CentOS Linux release 7.9.2009 (Core) (64 bit)
* The version of Flux you are running.
  * `InfluxDB 2.0.8 | Server: e91d418 | Frontend: cb0a87e` (I'm not sure how to get the Flux version from this.)
* Whether you installed it using a pre-built package or built it from source.
  *  Installed from the pre-built OSS RPM package.
* A small test case that demonstrates the issue or steps to reproduce the issue.
  * See the bottom of the issue.

---

I would expect a query that joins an SQL data stream with an InfluxDB data stream to yield results when joining on multiple fields.

I would expect the following query to give me results based on the two data streams queried:

```flux
import "sql"

routerInfo = sql.from(
  driverName: "postgres",
  dataSourceName: "postgres://user:pass@localhost:5432/testdb?sslmode=disable",
  query: "select router, interface, data from interfaces"
)

routerData = from(bucket: "testdata")
|> filter(fn: (r) =>
    r._measurement == "stats" and
    r._field == "value"
)

join(tables: {metric: routerData, info: routerInfo}, on: ["router", "interface"])
```

However, it gives me "No Results".

I'm posting this bug report after having chatted with people on the InfluxData Slack and tried to debug the issue, only to not find any reason that this should not work. I was suggested to post it on here.

The [documentation for joining InfluxDB data with SQL data](https://docs.influxdata.com/influxdb/v2.0/query-data/flux/sql/#join-sql-data-with-data-in-influxdb) doesn't include an example for joining on multiple columns, but the `on` field has to be a list, and the [documentation for `join()`]( https://docs.influxdata.com/flux/v0.x/stdlib/universe/join/#influxdb-cross-measurement-join) _does_ include an example for joining on multiple fields, so I would expect this to work.

---

Steps to reproduce:

1) Insert the following data into PostgreSQL:
```
testdb=# create table interfaces (router text not null, interface text not null, data json);
CREATE TABLE
testdb=# insert into interfaces (router, interface, data) values ('router1', 'port1', '{"customer":"acme corp"}');
INSERT 0 1
testdb=# insert into interfaces (router, interface, data) values ('router1', 'port2', '{"customer":"foo corp"}');
INSERT 0 1
testdb=# insert into interfaces (router, interface, data) values ('router2', 'port2', '{"customer":"foo corp"}');
INSERT 0 1
testdb=# insert into interfaces (router, interface, data) values ('router2', 'port1', '{"customer":"bar corp"}');
INSERT 0 1
```

2) Insert the following test data into InfluxDB
```
$ influx bucket create -n testdata
$ echo -n "stats,router=router1,interface=port1 value=123" | influx write -b testdata
$ echo -n "stats,router=router1,interface=port2 value=1235" | influx write -b testdata
$ echo -n "stats,router=router2,interface=port1 value=855" | influx write -b testdata
$ echo -n "stats,router=router2,interface=port2 value=2432" | influx write -b testdata
```

3) Execute the following flux query
```flux
import "sql"

routerInfo = sql.from(
  driverName: "postgres",
  dataSourceName: "postgres://user:pass@localhost:5432/testdb?sslmode=disable",
  query: "select router, interface, data from interfaces"
)

routerData = from(bucket: "testdata")
|> filter(fn: (r) =>
    r._measurement == "stats" and
    r._field == "value"
)

join(tables: {metric: routerData, info: routerInfo}, on: ["router", "interface"])
```

Which yields the following result:
image

---

###  Details of yield()-ing each query on its own:

`routerInfo |> yield()`:

image

`routerData |> yield()`:

image

`join(tables: {metric: routerData, info: routerInfo}, on: ["interface", "router"])`:

image

scbrickley:
Fixes #4150

~This patch takes advantage of the fact that `execute.groupKey` already has a list of sorted indices that can be used to access its elements in lexicographical order, without altering the underlying contents. It introduces a new `SortedGroupKey` interface, which functions much like the normal `GroupKey` interface, except that its public methods always operate on sorted columns.~

EDIT: This patch adds a new method to the `GroupKey` interface called `EqualTrueNulls`, which should be functionally identical to `Equal`, except that it never counts two nulls as equal. This new method is then used to compare the join keys for `universe.join` (thereby replacing the problematic `equalJoinkeys` function) as a solution to #4150.

The key to reproducing the original bug is to pay attention to the order in which columns are arranged in a table. If the columns are sorted in lexicographical order, there is no problem. If they are not, we get no output.

As an example, consider the following two flux scripts. They are almost identical, but only the first one yields a table.

Script 1:
```
import "array"

lhs = array.from(
    rows: [
        {interface: "port1", router: "router1", _value: 123},
        {interface: "port1", router: "router2", _value: 855},
        {interface: "port2", router: "router1", _value: 1235},
        {interface: "port2", router: "router2", _value: 2432},
    ],
)
    |> group(
        columns: [
            "interface",
            "router",
        ],
    )

rhs = array.from(
    rows: [
        {interface: "port1", router: "router1", data: "acme corp"},
        {interface: "port2", router: "router1", data: "foo corp"},
        {interface: "port2", router: "router2", data: "foo corp"},
        {interface: "port1", router: "router2", data: "bar corp"},
    ],
)

join(tables: {left: lhs, right: rhs}, on: ["router", "interface"])
```

Script 2:
```
import "array"

lhs = array.from(
    rows: [
        {interface: "port1", router: "router1", _value: 123},
        {interface: "port1", router: "router2", _value: 855},
        {interface: "port2", router: "router1", _value: 1235},
        {interface: "port2", router: "router2", _value: 2432},
    ],
)
    |> group(
        columns: [
            "interface",
            "router",
        ],
    )

rhs = array.from(
    rows: [
        {router: "router1", interface: "port1", data: "acme corp"},
        {router: "router1", interface: "port2", data: "foo corp"},
        {router: "router2", interface: "port2", data: "foo corp"},
        {router: "router2", interface: "port1", data: "bar corp"},
    ],
)

join(tables: {left: lhs, right: rhs}, on: ["router", "interface"])
```

The problem was that some parts of the join transformation operated on join keys with sorted columns, and others operated on join keys with unsorted columns. This caused some strange behavior when comparing or checking for equality between two join keys, since `[port1 router1] != [router1 port1]`.

This patch makes both of these scripts produce the same output.

### DIFF
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -748,9 +748,7 @@ func (c *MergeJoinCache) registerKey(id execute.DatasetID, key flux.GroupKey) {
 	switch id {
 
 	case c.leftID:
-
 		c.buffers[c.rightID].iterate(func(groupKey flux.GroupKey) {
-
 			keys := map[execute.DatasetID]flux.GroupKey{
 				c.leftID:  key,
 				c.rightID: groupKey,
@@ -772,9 +770,7 @@ func (c *MergeJoinCache) registerKey(id execute.DatasetID, key flux.GroupKey) {
 		})
 
 	case c.rightID:
-
 		c.buffers[c.leftID].iterate(func(groupKey flux.GroupKey) {
-
 			keys := map[execute.DatasetID]flux.GroupKey{
 				c.leftID:  groupKey,
 				c.rightID: key,
@@ -841,19 +837,6 @@ func (c *MergeJoinCache) buildPostJoinSchema() {
 	for j, column := range c.schema.columns {
 		c.colIndex[column] = j
 	}
-}
-
-// equalJoinKeys compares two keys for equality.
-// Null values are not considered equal when joining (unlike when grouping).
-func equalJoinkeys(left, right flux.GroupKey) bool {
-
-	for j, v := range left.Values() {
-		// value.Equal will return false if both sides are null
-		if !v.Equal(right.Value(j)) {
-			return false
-		}
-	}
-	return true
 }
 
 // Find indexes of column names missing between the two tables
@@ -928,8 +911,7 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
 
 	// Perform sort merge join
 	for !leftSet.Empty() && !rightSet.Empty() {
-		if equalJoinkeys(leftKey, rightKey) {
-
+		if leftKey.EqualTrueNulls(rightKey) {
 			for l := leftSet.Start; l < leftSet.Stop; l++ {
 				for r := rightSet.Start; r < rightSet.Stop; r++ {
 
--- a/internal/execute/groupkey/groupkey.go
+++ b/internal/execute/groupkey/groupkey.go
@@ -82,7 +82,7 @@ func (k *groupKey) ValueTime(j int) values.Time {
 }
 
 func (k *groupKey) Equal(o flux.GroupKey) bool {
-	return groupKeyEqual(k, o)
+	return groupKeyEqual(k, o, false)
 }
 
 func (k *groupKey) Less(o flux.GroupKey) bool {
@@ -100,6 +100,10 @@ func (k *groupKey) String() string {
 	}
 	b.WriteRune('}')
 	return b.String()
+}
+
+func (k *groupKey) EqualTrueNulls(o flux.GroupKey) bool {
+	return groupKeyEqual(k, o, true)
 }
 
 func (k *groupKey) hash64() (h uint64) {
@@ -154,7 +158,7 @@ func (k *groupKey) hash64() (h uint64) {
 	return h
 }
 
-func groupKeyEqual(a *groupKey, other flux.GroupKey) bool {
+func groupKeyEqual(a *groupKey, other flux.GroupKey, truenulls bool) bool {
 	b, ok := other.(*groupKey)
 	if !ok {
 		b = newGroupKey(other.Cols(), other.Values())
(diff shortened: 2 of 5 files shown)


stdlib/universe/join.go
```diff
@@ -845,11 +841,10 @@ func (c *MergeJoinCache) buildPostJoinSchema() {
...
 // equalJoinKeys compares two keys for equality.
 // Null values are not considered equal when joining (unlike when grouping).
-func equalJoinkeys(left, right flux.GroupKey) bool {
-
-	for j, v := range left.Values() {
+func equalJoinkeys(left, right flux.SortedGroupKey) bool {
+	for j, v := range left.SortedValues() {
 		// value.Equal will return false if both sides are null
-		if !v.Equal(right.Value(j)) {
+		if !v.Equal(right.ValueSorted(j)) {
```
wolffcm, line 847: Really glad to hear you got to the bottom of this!

I must be missing something though. I see the interface `GroupKey` has a method `Equal` and in the implementation it does a sorted comparison:
https://github.com/influxdata/flux/blob/ac55424f0d5dcf9bc45efba441ede66ed24f2913/internal/execute/groupkey/groupkey.go#L157
Is there a reason that the standard `GroupKey.Equal` would not work?

scbrickley: Didn't catch this so I tried it out, and got some failing tests. Looks like the key difference between these two methods is that `v.Equal()` will never consider two nulls to be equal, where as `groupKeyEqual` will.

https://github.com/influxdata/flux/blob/ac55424f0d5dcf9bc45efba441ede66ed24f2913/internal/execute/groupkey/groupkey.go#L171-L174

https://github.com/influxdata/flux/blob/ac55424f0d5dcf9bc45efba441ede66ed24f2913/stdlib/universe/join.go#L842-L844

wolffcm: Maybe you could add a method like `EqualNoNulls` to the `GroupKey` interface? (Changing an interface could break downstream repos but I don't think it will.) Both behaviors seem really useful to have.

scbrickley: That seems reasonable to me. Working on it.

wolffcm (approved):
This is still in draft but looks good to go to me. Nice work!

nathanielc (approved):

PR #4194, opened 2021-10-28
feat: Support field access in vectorization

### LINKED ISSUE #4184: Add a vectorization pass to analysis in Rust
As described in this [comment](https://github.com/influxdata/flux/issues/4028#issue-993584462) we should implement a vectorization pass to analysis in Rust that is performed after type inference.

The result of this pass (if successful) will populate the `vectorized` field added in #4183.

For this issue this pass can be very simple. It should walk over the semantic graph for the Flux program and for each `FunctionExpression` it encounters, it should look at the parameters. If it has a single parameter named `r`, it will be a candidate for vectorization.

For each candidate, walk over the body of the function. If the function has this form:
```
(r) => ({r with a: r.a, b: r.b, ...})
```
The `r with` part of the body may be omitted. The point is that the function body should just be projecting columns from its input to its output. The types of the parameters and all the expressions in the function body should be updated to have vector types. That is if `r` has type `{a: int, b: float, c: string | T}` then it should be udpated to `{a: v[int], b: v[float], c: v[string] | T}`. (The `v[...]` indicates a vector type)

DOD:
- A simple vectorization pass is added to analysis
- Unit tests exist for that demonstrate the vectorization pass works as intended

Marwes:
Simplest possible vectorization pass

Closes #4184

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -1641,21 +1822,24 @@ pub struct IdentifierExpr {
...
+        let (t, cons) = infer.lookup(&self.loc, &self.name)?;
         self.typ = t;
         Ok(cons)
     }
     fn apply(mut self, sub: &Substitution) -> Self {
         self.typ = self.typ.apply(sub);
         self
     }
+
+    fn vectorize(&self, env: &VectorizeEnv) -> Result<Self> {
```
skartikey, line 1834: I am not clear on two functions using the same function signature. How does this work in rust?
Also, there are 4 vectorize methods in this file, I can see they are for different purposes here. but if possible could you please rename it to a more specific task each of the methods is doing. something like vectorize_pkg

skartikey (approved):

libflux/flux-core/src/semantic/nodes.rs
Marwes: All of the `vectorize` methods are on different types, so you have `IdentifierExpr:;vectorize`, `Expression::vectorize` etc. I might have done `VectorizeEnv::vectorize_pkg`, `VectorizeEnv::vectorize_identifier` etc instead but `fn infer` and `fn apply` already exist in the same manner as the added `vectorize` methods so I stuck with the same naming scheme.

skartikey: That's alright then.

PR #4196, opened 2021-10-28
fix(compiler): add more error checking for nulls and invalid types

### LINKED ISSUE #4126: Panic using json.parse() referencing array on an object
Given this valid JSON:
```
{
	"name": "Mary",
	"age": 40,
	"cars": {
		"car1": "Ford",
		"car2": "BMW"
	}
}
```

Using `json.parse()` `cars[0]` results in a Panic because it's referencing a non-existent array:

```
import "array"
import "experimental/json"

array.from(rows: [{_measurement: "m0", mytag: "t0", _field: "f0", _value: "{ \"name\":\"John\", \"age\":30, \"cars\": {\"car1\": \"Ford\", \"car2\": \"BMW\" } }", _time: 2021-09-01T00:00:00Z}])
    |> range(start: -100y)
    |> map(fn: (r) => {
    jsonData = json.parse(data: bytes(v: r._value)) return {
        _time: r._time,
        _field: r._field,
        name: jsonData.name,
        age: jsonData.age,
        cars: jsonData.cars[0]
    }
    })
```
Screen Shot 2021-10-07 at 3 38 43 PM

wolffcm:
There were several places in the compiler where we make assumptions about the types of values, or we assume that the value is not null. I added checks for these conditions and associated tests.

In order to get values whose types are not known until runtime I added a new function `makeAny` to the `testutil` stdlib package.

Fixes #4126.


jsternberg (approved):
I'm a little uncomfortable with this in the long term, but I think we need this until the Go runtime has full access to the polytypes and can reject types that would not be valid.

In general, the compiler should not be checking types and should not be verifying the types of values it receives.

PR #4202, opened 2021-11-01
feat: Always continue inference on errors

Marwes:
Whenever we encounter an error we now store it in the error list instead and just continue. With errors like a undefined identifier we still need to supply a type so that the inference can continue so I added the special `Error` type as a placeholder which never causes any further errors to be generated if it is encountered later on. A fresh type variable could almost be used instead but it would cause further errors to be generated for code like

```
x = y
z = x + 1
z2 = x + "a"
```
(`x` would be inferred to an `int` in the `z` binding which would then cause an error in the `z2` binding, however we do not know whether `y` was intended to be an `int` or `string` so we user `Error` instead to ensure errors are not generated).

### Done checklist
- [x] Test cases written


libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -954,21 +962,6 @@ impl FunctionExpr {
         }
         // And use it to infer the body.
         let bcons = self.body.infer(infer)?;
-        // HACK Remove once substitutions are persisted correctly
```
rockstar, line 957: Any code commented with `HACK` that we remove is a-okay in my book.

rockstar (approved):
Bummed I did the release so early, would've loved to have this in the flux-lsp so quickly. Will play with it later.

PR #4203, opened 2021-11-01
fix(fmt): Formatter formats invalid syntax into even worse invalid syntax

skartikey:
Issue - Formatter was formatting an invalid syntax script.

Fix - The fix is to check AST nodes(call ast.check()) and return if any error is found to the user, before proceeding to format the script. 

while fixing it we found that the AST walk doesn't actually traverse BuiltinStmt. So any errors that occur when parsing the type of builtin are just ignored. So to fix it AST walk for the BuiltinStmt has been added.

fixes: https://github.com/influxdata/vsflux/issues/242


libflux/flux-core/src/formatter/mod.rs
```diff
@@ -26,6 +26,10 @@ pub fn convert_to_string(file: &File) -> Result<String> {
 /// ```
 pub fn format(contents: &str) -> Result<String> {
     let file = parse_string("".to_string(), contents);
+    let node = ast::walk::Node::File(&file);
+    if let Err(err) = ast::check::check(node) {
+        panic!("Script parsing failed. {:?}", err);
+    }
```
Marwes, line 32:
```suggestion
    ast::check::check(node)?;
```

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1268,3 +1268,17 @@ a",
     );
     assert_unchanged(r#"a = "a string literal with a tabstop '	'""#);
 }
+
+#[test]
+fn invalid_syntax() {
+    assert_unchanged(
```
Marwes, line 1274: The `assert_unchanged` function will panic on errors. I think you should just call `format` here and check that you get an error.

Marwes (changes requested):

Marwes:
@skartikey Might be good to have someone else review this as one of the commits are mine now.

skartikey:
@Marwes Just did that.

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1231,7 +1231,7 @@ fn preserve_multiline_test() {
     return tables
         |> aggregateWindow(every: every, fn: fn, column: _column, createEmpty: true)
         |> fill(column: _column, value: defaultValue)
-        |> rename(fn: (column) => if column == _column then _as else column)"#,
+        |> rename(fn: (column) => if column == _column then _as else column)}"#,
```
rockstar, line 1234: Was this a purposeful change? How did this ever work?

libflux/flux-core/src/formatter/tests.rs
```diff
@@ -1268,3 +1268,17 @@ a",
...
+
+#[test]
+fn invalid_syntax() {
+    let script = r#"builtin diff : (
+        <-got: [A],
+        want: [A],
+        ?verbose: bool,
+        ?epsilon: float,
+        ?nansEqual: bool,
+        aoeustahoesih
```
rockstar, line 1280: Haha! I know why this random string is here, but it's going to result is some really fun lore in the future. 😄

rockstar (approved):

libflux/flux-core/src/formatter/tests.rs
skartikey: yeah, this was caught after my fix :)  Earlier there was no ast check at all, code was parsing the script without checking for error and calling the formatter on the parsed script. All fixed now.

Marwes (approved):

PR #4208, opened 2021-11-02
feat: Allow type inference to run on invalid ASTs

Marwes:
Allows for more errors to be reported to users simultaneously, thereby
reducing the number of edit/compile cycles they have to go through.


scbrickley (approved):
LGTM - Hope that this allows for some extra stability in the new LSP. Thanks for doing this!

PR #4211, opened 2021-11-02
feat(docs): add inline docs to array, csv, and date packages

sanderson:
This change adds inline fluxdoc-compliant documentation to the `array`, `csv`, and `date` packages.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/date/date.flux
```diff
@@ -1,397 +1,463 @@
...
+//   Use an absolute time, relative duration, or integer.
+//   Durations are relative to `now()`.
 //
-// ## Return the second of a time value
+// ## Examples
+// 
+// ### Return the second of a time value
 //
-// ```
+// ```no_run
```
Marwes, line 20: Why can't these be run?

sanderson: I don't think it's a matter of "can't be run." I think it's that they don't need to be run. I assumed that we'd run examples to generate and capture input and output streams. Where these don't have any streams of tables, they don't need to be run. But that assumption may be incorrect.

@nathanielc Should I have examples with no input/output tables still run?

nathanielc: If an example doesn't have any input or output then we won't capture and data but they can still run. If they get an error that error will be reported. Therefore we don't need to mark them as `no_run` unless we expect the code will error or have another reason for not running.

nathanielc: If you do get an error trying to run any of these examples that you don't expect ping me and I'll take a look. There could easily be a bug in the example runner at this stage in the process.

nathanielc: After writing this comment I thought of a case were an error would not be reported so I fixed it here https://github.com/influxdata/flux/pull/4215

nathanielc (approved):

nathanielc:
The semantic bot is complaining about this commit https://github.com/influxdata/flux/pull/4211/commits/2ccd8fd753990689962795364a4fb45441db898c if you rebase and squash your commits into a single commit that should fix it.

PR #4212, opened 2021-11-03
chore: remove tero dependency

rockstar:
This dep isn't used anymore, and builds a dependency tree that's deeper
than needed, so we should just drop it.

### DIFF
--- a/libflux/Cargo.lock
+++ b/libflux/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


onelson:
I'm still unfamiliar with Go, so I'm wondering what that buildinfo file is doing. What's the story there?

onelson (commented):
I'm still unfamiliar with Go, so I'm wondering what that buildinfo file is doing. What's the story there?

I see that the sources that changed have new hashes calculated in the map. I guess, a better questions would be, is buildinfo a standard part of the cgo tooling or is it something specific to flux? (I can figure this out in time, it'll be fine).

onelson (approved):

onelson:
> I see that the sources that changed have new hashes calculated in the map. I guess, a better questions would be, is buildinfo a standard part of the cgo tooling or is it something specific to flux? (I can figure this out in time, it'll be fine).

Answering my own question here. Looks like this file gets bumped in response to `go generate` (which is mentioned in the _Contributing_ section of the readme). https://go.dev/blog/generate

PR #4213, opened 2021-11-03
refactor: Refactor and group imports

Marwes:
Ran `cargo fmt` with the following `.rustfmt.toml`. I think this looks a
bit nicer, less repetition and it is easier to add more imported items.

```
reorder_imports=true
group_imports="StdExternalCrate"
imports_granularity="Crate"
```


nathanielc:
I like this change two thoughts:

Should we commit that .rustfmt.toml change so that we continue to enforce this style?

Should we doc somewhere how to configure rust-analyzer to easily maintain this style when editing? See https://rust-analyzer.github.io/manual.html#auto-import 

I think this would be the config that matches the toml:

```json
{
  "assist": {
    "importGranularity": "crate",
    "importEnforceGranularity": true,
    "importPrefix": "by_crate",
    "importGroup": true,
  }
}
```

I tried that out locally and seems to be consistent.

Marwes:
The `.rustfmt.toml` ought to be obeyed by rust-analyzer already if it were added.

Personally I try to avoid custom configuration for formatting so i'd be happy to leave it out. These options were just a way to automatically reformat imports to something easier to edit/read, but I don't specifically think they are the perfect way to format them (or that there is a perfect way to format things).

nathanielc (approved):

PR #4215, opened 2021-11-03
fix(fluxdoc): adds handling for errors executing examples

nathanielc:
Prior to this change errors in executing examples were ignored.
Now they are reported correct when returned from the flux execute CLI or
encoded into the CSV response.

Additionally the Makefile was not testing code behind a feature flag now
it does.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


sanderson (approved):
LGTM

PR #4220, opened 2021-11-04
feat(fluxdoc): Allow reporting unlimited diagnostics

pierwill:
User can pass `--limit=0` to display all errors.

### DIFF
--- a/libflux/flux-core/src/bin/fluxdoc.rs
+++ b/libflux/flux-core/src/bin/fluxdoc.rs
@@ -49,7 +49,7 @@ enum FluxDoc {
         dir: PathBuf,
         /// Limit the number of diagnostics to report. Default 10. 0 means no limit.
         #[structopt(short, long)]
-        limit: Option<isize>,
+        limit: Option<i64>,
         /// Honor the exception list.
         #[structopt(long)]
         allow_exceptions: bool,
@@ -156,7 +156,7 @@ fn dump(
 fn lint(
     stdlib_dir: Option<&Path>,
     dir: &Path,
-    limit: Option<isize>,
+    limit: Option<i64>,
     allow_exceptions: bool,
 ) -> Result<()> {
     let stdlib_dir = match stdlib_dir {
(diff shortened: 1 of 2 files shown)


libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -174,13 +175,16 @@ fn lint(
     };
     let (_, mut diagnostics) = parse_docs(stdlib_dir, dir, exceptions)?;
     if !diagnostics.is_empty() {
-        let rest = diagnostics.len() as i64 - limit as i64;
+        let rest = match limit {
+            usize::MAX => 0,
+            _ => diagnostics.len() - limit,
+        };
         println!("Found {} diagnostics", diagnostics.len());
```
pierwill, line 182: This prevents overflows in the arithmetic.

libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -174,13 +175,16 @@ fn lint(
     };
     let (_, mut diagnostics) = parse_docs(stdlib_dir, dir, exceptions)?;
     if !diagnostics.is_empty() {
-        let rest = diagnostics.len() as i64 - limit as i64;
+        let rest = match limit {
```
nathanielc, line 178: Seems odd that this logic is needed. I expected rest to be a signed integer thus allowing it to become negative if the limit was greater than the actual length of the list, thus keeping the logic below for `if rest > 0` simple as well. Can we still explicitly use a signed integer for rest? Maybe add a comment that we are using a signed integer so that we rest can be negative?

nathanielc (changes requested):
Thanks for the improvement, one small change requested.

libflux/flux-core/src/bin/fluxdoc.rs
pierwill: I understood it like this: if the limit is unlimited, then the remaining lints to print will always be 0!

I'd happy to change it, though. 😇

nathanielc (approved):
Looks great thanks for taking the time to figure out a good size for the integer. 

FWIW if  limit is i64::MAX and usize is 32bits then we end up with u32::MAX as the value for the truncate which is exactly what we want. 

See https://doc.rust-lang.org/rust-by-example/types/cast.html in short i64::MAX has all bits set to 1 so when casting to a smaller size integer Rust keeps the least significant bits which are all 1s which ends up being u32::MAX. Not that we expect this code to run on  a 32bit system but good to know its safe.

PR #4223, opened 2021-11-04
fix(libflux): Fix docs errors when running `cargo doc`

pierwill:
Also:

- Require documentation for `flux` module, and document two missing items
- Use `cargo check` to check for missing docs in `libflux` target


Makefile
```diff
@@ -70,6 +70,7 @@ internal/feature/flags.go: internal/feature/flags.yml
 	$(GO_GENERATE) ./internal/feature
 
 libflux: $(LIBFLUX_GENERATED_TARGETS)
+	cd libflux && $(CARGO) check # this checks for missing docs
```
nathanielc, line 73: I think this would be better as part of the test-rust target instead of the build. That way you can still build the code while working to get docs/tests passing. Thoughts?

pierwill: Makes sense!

libflux/flux/src/lib.rs
```diff
@@ -1,4 +1,6 @@
-#![cfg_attr(feature = "strict", deny(warnings))]
+#![cfg_attr(feature = "strict", deny(warnings, missing_docs))]
+
+//! Serialization from Flux.
```
pierwill, line 3: Maybe

```suggestion
//! Serialization of Flux code.
```
?

Does this module do other work?

pierwill: Although maybe this is okay as is...

Makefile
```diff
@@ -119,6 +119,7 @@ test-go: libflux-go
 
 test-rust:
 	cd libflux && $(CARGO) test $(CARGO_ARGS) && \
+	$(CARGO) check # this checks for missing docs
 	$(CARGO) test --doc && \
 	$(CARGO) clippy $(CARGO_ARGS) -- -Dclippy::all
```
nathanielc, line 124: How do `cargo check`,  `cargo test --doc`, and `cargo clippy` differ?

nathanielc (commented):
Looks like a good change but I don't know enough about implications here pinging @rockstar  for review as well.

libflux/flux/src/lib.rs
wolffcm: I was about to comment to this effect. This module provides the public facing API for the Go runtime including formatting, parsing and stdlib-aware analysis.

Makefile
pierwill: `clippy` checks for some common errors and unidiomatic code. `test --doc` tests the examples within the docs. because we have `deny(missing_docs)` set, `check` will emit warnings for missing docs. (`cargo docs` would too, but we don't actually need to build the docs.)

wolffcm (commented):
This looks good to me but I'd also like to see what Paul thinks.

Also, thanks for volunteering this work, it's much appreciated.

Makefile
```diff
@@ -119,6 +119,7 @@ test-go: libflux-go
 
 test-rust:
 	cd libflux && $(CARGO) test $(CARGO_ARGS) --all-features && \
+	$(CARGO) check && \
```
Marwes, line 122: `cargo test` which is called on the line above here should cover everything that `cargo check` does already

pierwill: This does work. Thanks @Marwes!

rockstar (approved):

PR #4226, opened 2021-11-05
feat(contrib): make some Alerta parameters and fields optional

alespour:
This PR 
1. adds missing README
2. uses `experimental.record` package to make
    - `attributes` parameter (of record type) in `alerta.alert` function optional
    - the following fields in record returned by `mapFn` in `alerta.endpoint` function optional (their default value is the same as in `alerta.alert` function then):
        - service
        - group
        - value
        - text
        - tags
        - attributes
        - type
        - timestamp

### Done checklist
- [X] Test cases written


timhallinflux:
@alespour can you resolve the conflicts?

skartikey (approved):
Looks good!

Issue #4231, opened 2021-11-08
Parse error on map - expected RPAREN, got EOF(1)


dav2dg:
Hi,

I am getting an error when I use the **map** line in the following query
> from(bucket: "myBucket")
>   |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
>   |> filter(fn: (r) => r["_measurement"] == "myMeasure")
>   |> filter(fn: (r) => r["_field"] == "myField")
>   |> map(fn: (r) => ({ r with _value: if r["_measurement"] == "myMeasure" and r["_field"] == "myField" then r._value * 2}) )
>   |> aggregateWindow(every: 5s, fn: mean, createEmpty: false)
>   |> yield(name: "mean")
> 

whereas the following works

> from(bucket: "myBucket")
>   |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
>   |> filter(fn: (r) => r["_measurement"] == "myMeasure")
>   |> filter(fn: (r) => r["_field"] == "myField")
>   |> aggregateWindow(every: 5s, fn: mean, createEmpty: false)
>   |> yield(name: "mean")
> 

To reproduce just paste the first query into the script editor

I have tried docker  influxdb v2.0.4-alpine, 2.0.9-alpine, 2.1.0-alpine
It seems it may be related to **FLUX** issue #1801  and #1884 which were fixed in Flux 0.59.4 from Sept 2019

What did I do wrong????

**docker-compose.yml:**

`version: '3.3'

services:
  influxdb:
    image: influxdb:2.1.0-alpine
    ports:
      - '8086:8086'
    volumes:
      - influxdb-storage:/var/lib/influxdb
    environment:
      - DOCKER_INFLUXDB_INIT_MODE=setup
      - DOCKER_INFLUXDB_INIT_USERNAME=admin
      - DOCKER_INFLUXDB_INIT_PASSWORD=theadminpassword
      - DOCKER_INFLUXDB_INIT_ORG=my-Org
      - DOCKER_INFLUXDB_INIT_BUCKET=myBucket

volumes:
    influxdb-storage:
`

Marwes:
You are missing the `else <EXPR>` part of the `if` expression.

```
|> map(fn: (r) => ({ r with _value: if r["_measurement"] == "myMeasure" and r["_field"] == "myField" then r._value * 2 else SOMETHING}) )
```

Marwes:
Looking at the error emitted, it definitely ought to be better though.

PR #4232, opened 2021-11-08
fix: Reduce the amount of extra parse errors generated

Marwes:
Saw some room for improvement due to #4231. Due to `fn expect` consuming
tokens until it found the correct one it is really easy for the parser
to end up trying to match the rest of the source code, producing an
error for each token, obfuscating the actual cause for the error. It is
possible that this will cause worse errors in other instances (where we
would eventually find the token we expected) but there are not such
tests atm at least.

If we want to improve the error reporting further I would suggest either
using [LALRPOP](https://github.com/lalrpop/lalrpop) and its error
reporting or try to mold this hand-written parser to something closer to
[rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/tree/master/crates/parser)
.

https://rust-analyzer.github.io/blog/2020/09/16/challeging-LR-parsing.html


nathanielc (approved):
I like this fix. It makes a good incremental step towards better error messages and certainly improves this case.

PR #4254, opened 2021-11-10
feat: Add the capability of reading options from the Context

### LINKED ISSUE #4195: Expose options to Flux transformations
There should be a convenient way (maybe the `Context` or maybe `Administrator`) to access any options set in `option` statements in the Flux script.

Specifically create an option for vectorization and `map`.

DOD:
- An option exists in the `internal` package to enable or disable vectorization of `map` to allow easy A/B comparisons of performance
-  New tests are added

Marwes:
Duplicates the package map stored in `importer` but `context.Context` seems to be the only object passed through everywhere at the moment.

Closes #4195

### Done checklist
- [x] Test cases written


stdlib/universe/map.go
```diff
@@ -144,6 +145,12 @@ func (t *mapTransformation) Process(id execute.DatasetID, tbl flux.Table) error
 		return err
 	}
 
+	if vectorize, ok := interpreter.GetOption(t.ctx, "internal/debug", "vectorize"); ok && vectorize.Type().Nature() == semantic.Bool {
+		if vectorize.Bool() {
+			fmt.Printf("Vectorization enabled! (But not yet implemented)\n")
```
wolffcm, line 150: This must've helped you test it out, but seems like you could remove it now.

stdlib/internal/debug/debug_test.flux
```diff
@@ -44,3 +44,13 @@ testcase sink {
...
+
+testcase get_option {
+    got = debug.getOption(pkg: "internal/debug", name: "vectorize")
+    want = false
+
+    testing.diff(
+            got: array.from(rows: [{ v: got }]),
+            want: array.from(rows: [{ v: want }])
+        ) |> yield()
+}
```
wolffcm, line 56: Can you add some more tests here, with the option set to true? And also test that the option is visible inside of the `compile` package, by calling `getOption` from inside of `map`.

Example:
```js
option debug.vectorize = true

...

got = array.from(rows: [{}]) |> map(fn: (r) => ({getOption: getOption(pkg: "internal/debug", name: "vectorize")}))

```

interpreter/interpreter.go
```diff
@@ -136,7 +136,36 @@ func (itrp *Interpreter) doPackageClause(pkg *semantic.PackageClause) error {
...
+
+const packagesKey key = iota
+
+type Packages map[string]*Package
+
+func (p Packages) Inject(ctx context.Context) context.Context {
+	return context.WithValue(ctx, packagesKey, p)
+}
+
+func GetPackages(ctx context.Context) Packages {
```
wolffcm, line 149: This is only used from inside this package, it would be good to change the name to `getPackages` to avoid exposing it.

interpreter/interpreter.go
```diff
@@ -136,7 +136,36 @@ func (itrp *Interpreter) doPackageClause(pkg *semantic.PackageClause) error {
...
+	if packages == nil {
+		return values.InvalidValue, false
+	}
+	if pkg, ok := packages[pkg]; ok {
+		return pkg.object.Get(option)
+	}
+	return values.InvalidValue, false
+}
+
+func (itrp *Interpreter) doImport(dec *semantic.ImportDeclaration, ctx context.Context, scope values.Scope, importer Importer) error {
```
wolffcm, line 168: _nit_ The context is typically the first parameter, as mentioned on Go's stdlib docs:
https://pkg.go.dev/context@go1.17.3#pkg-overview

(We seem to break the rules outlined there in several places though)

wolffcm (changes requested):
This looks good, I just had some minor requested changes.

For what it's worth, I think that the `ctx` is a reasonable place to store options. Options are always defined at the top of the user's script, so they are really "request-scoped values", which is what the context is for.

stdlib/internal/debug/debug_test.flux
Marwes: fixed, though `array.from(rows: [{}])` doesn't actually create any rows over which to iterate over :/ . Couldn't see an easy way to fix it to emit an empty row either.

```
package array_test

import "array"
import "testing"

testcase from_empty_record {
    want = array.from(rows: [{ v: 123 }])
    got = array.from(rows: [{}]) |> map(fn: (r) => ({ v: 123 }))

    testing.diff(got, want) |> yield()
}
```

stdlib/internal/debug/debug_test.flux
```diff
@@ -44,3 +44,32 @@ testcase sink {
...
+
+    got = debug.getOption(pkg: "internal/debug", name: "vectorize")
+    want = true
+
+    testing.diff(got: array.from(rows: [{v: got}]), want: array.from(rows: [{v: want}]))
+        |> yield()
+}
+
+testcase get_option_in_map {
+    option debug.vectorize = true
```
wolffcm, line 67: I did not know that you could use an option statement like this. Normally it has to be at the top of the file. I guess the AST-manipulation for running test does this. This is a nice surprise.

wolffcm (approved):
Nice work, this will be very useful.

PR #4255, opened 2021-11-10
fix: Propagate the element type through array constructors

### LINKED ISSUE #3858: Static type for array.from() causing strange output from findColumn
The following query yields: `cannot represent the type t11447 as column data`

```
  getValue = (b, me, f) => {
    r = from(bucket: b)
    |> range(start: -99d)
    |> filter(fn: (r) => r["_measurement"] == me)
    |> group(columns: ["_field"])
    |> first()
    |> aggregateWindow(every: 1s, fn: last, createEmpty: false)
    |> yield(name: "last")
    |> findColumn(
            fn: (key) => (key._field == f),
            column: "_value"
    )
    return r[0]
    }
scalar =  getValue(b:"noaa",me:"average_temperature",f:"degrees" )

array.from(rows: [{_time: now(), _value: scalar}])
|>yield(name:"test value")

from(bucket: "noaa")
	|> range(start: -30d)
	|> filter(fn: (r) => (r["_field"] == "degrees"))
	|> set(key: "_measurement", value: "scaledMeasurements")
    |> limit(n: 5)
	|> map(fn: (r) =>
		({r with _value: scalar}))
  |> yield(name:"complete")
```

Output:
```
#group,false,false,true,true,false,false,true,false,false
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string
#default,last,,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement,location
,,0,2021-03-30T16:00:05.180856426Z,2021-07-07T16:00:05.180856426Z,2021-05-14T21:11:38Z,74,degrees,average_temperature,santa_monica
#group,true,true
#datatype,string,string
#default,,
,error,reference
,cannot represent the type t11454 as column data,
```

Solution 
```
 return float(v:r[0])
```

Marwes:
Before it was just using the statically determined type which causes functions like `array.from` to fail as they would see the type variable from the generic function instead of the actual type.

Fixes #3858

### Done checklist
- [x] Test cases written

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -517,7 +517,13 @@ func (itrp *Interpreter) doArray(ctx context.Context, a *semantic.ArrayExpressio
 		}
 		elements[i] = v
 	}
-	return values.NewArrayWithBacking(a.TypeOf(), elements), nil
+	arrayType := semantic.MonoType{}
+	if len(elements) > 0 {
+		arrayType = semantic.NewArrayType(elements[0].Type())
+	} else {
+		arrayType = a.TypeOf()
+	}
+	return values.NewArrayWithBacking(arrayType, elements), nil
 }
 
 func (itrp *Interpreter) doDict(ctx context.Context, e *semantic.DictExpression, scope values.Scope) (values.Value, error) {
--- a/libflux/flux-core/src/parser/tests/errors.rs
+++ b/libflux/flux-core/src/parser/tests/errors.rs
@@ -891,9 +891,10 @@ map(fn: (r) => ({ r with _value: if true and false then 1}) )
 "#,
     );
     let parsed = p.parse_file("".to_string());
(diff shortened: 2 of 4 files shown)


stdlib/array/flux_test_gen.go
```diff
@@ -3077,2878 +3077,6 @@ var FluxTestPackages = []*ast.Package{&ast.Package{
 				Name: "array_test",
 			},
 		},
-	}, &ast.File{
```
Marwes, line 3080: This doesn't seem right  I just added to the file:S

wolffcm (approved):
This looks fine, but I wonder if the code that was producing the message `cannot represent the type t11447 as column data` was looking in the wrong place for the type. I guess this will be part of the discussion next week about "runtime values should not know their type".

Marwes:
Arguably, yes. But if we decided that it should not be looking at the `MonoType` stored in the value as it currently does, we ought to remove the stored `MonoType` entirely since it can't be reasonably correct.

PR #4262, opened 2021-11-11
Bump rust toolchain 1.56

onelson:
Following on from #4259 this bumps the rust toolchain up to the current latest version (1.56) and satisfies the new lints that come with it.


onelson:
I noticed the iox repo added a step to their CI to print out the versions of the rust tooling. Wondering if this is something we'd like to see here. Could be a good diagnostic signal for times like now where the specified toolchain is different from the docker image.

Marwes (approved):

PR #4263, opened 2021-11-11
chore(libflux): require documentation for docs module

pierwill:
For good measure. 😇


rockstar (commented):
I'm not sure we need this patch, as the `deny` from the `lib.rs` [here](https://github.com/influxdata/flux/blob/docs-deny-missing/libflux/flux-core/src/lib.rs#L1) should get us the functionality, so it'd be redundant. Am I missing it here?

pierwill:
Ah! You're right! `lib.rs` covers it. Thanks!

Issue #4264, opened 2021-11-11
EPIC: Diagnostics Enhancements


onelson:
Tracker issue for the broad topic of improving error reporting in Flux.

onelson:
Had a meeting today to discuss the scope of this effort. Some of it seems actionable now, some over a longer period, and finally there's some "pie in the sky" thought about ongoing efforts to promote clarity in the diagnostic feedback.

## Seems actionable now (if prioritized)

### More illustrative feedback

This is along the lines of  #4270 where we might aim to use rust-like 

### More granular feedback about type mismatches

Currently it sounds like when flux sees a type mismatch between fields on records that cannot be reconciled the feedback reports the mismatch in terms of 2 differing record types.

If we can report these violations in terms of the fields, it might be easier for users to debug.

> Ideally, we'd say _"`A.foo` has different type than `B.foo`"_ instead of _"`A` and `B` are incompatible types"_.

### Search-friendly jargon

We talked about syntax being tricky to do searches for, and how promoting terms/jargon in both the docs and error messages could be a way to make concepts more searchable.

> Another way to think about it is to have recommendations for pronouncing/verbalizing concepts.
> Ex: is there a good way to verbalize `[A]`?

### Style Guide/Glossary

> NB This is largely **my pitch** since I don't know much about Flux yet. I don't know that there are inconsistencies, but I think it'd be easier for a novice like me to avoid introducing them if a style guide or glossary existed!

Today, probably the closest thing to a glossary I could find is the [flux spec].

Ideally we'd have some sort of internal guidance to help align terminology between error messages and documentation to ensure we _give things a name and only one name_. From a technical writing perspective having consistency in the language can reduce ambiguity.

Once we've got language guidance and a strict list of terminology we could try to stress (color codes in terminal output, etc) those terms when they appear in error messages, or link to docs.

Separately, look at reworking language around array/stream after the type system knows the difference (ref #4275)

cc @sanderson - this might be a nice thing to collaborate on with the Docs team (if we decide to prioritize it).


## Actionable over time

### Error Codes

The thought here is once we standardize the language we use for error messages, we may find it easier to identify buckets/classifications that can have codes associated with them.

_Error Codes_ are nice since they give users anchors to search for on the web, and we can include a table in the language docs that has more detailed explanations of what went wrong. This can possibly help the error messages themselves be a little more terse, moving the longer-form text to the docs.


## Ongoing efforts

[...]

Issue #1380, opened 2019-06-17
Include source location info for all user-facing error messages


wolffcm:
Outside of internal errors, we should be able to provide source location info (both start and end locations from the Flux program) for any error that we return to the user.  (Even for internal errors, source location would be quite useful, e.g, if storage was down the source location info could indicate the call to `from()` that had the issue)

This would help make error messages much more useful, and also open the door to cool web UI features like being able to highlight problematic parts of Flux text just by hovering over the error message.

We could either indicate the source locations by embedding them in a message, or more programmatically by defining a special struct for Flux errors with fields for source location:
```go
package flux 
type interface Error {
  Error() string
  SourceLocation() *ast.SourceLocation
}
```
See also #1364

PR #4270, opened 2021-11-15
feat: Allow errors to be formatted via codespan

Marwes:
Codespan adds in highlighted source to error messages which gives more direct feedback to users as they do not need to trace out the line and column numbers themselves.

cc #4264 #1380


libflux/flux-core/src/semantic/tests.rs
```diff
@@ -3655,16 +3692,27 @@ error @3:17-3:18: undefined identifier y",
...
             isType(v: {}, type: "record")
             isType(v: [], type: "array")
         "#,
-        err: "error @2:13-2:42: {} is not Basic (argument v)
+        err: expect_test::expect![[r#"
+            error: 
+              ┌─ :2:13
+              │
+            2 │             isType(v: {}, type: "record")
+              │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ {} is not Basic (argument v)
```
Marwes, line 3708: Due to how constraints are generated for call expressions we unfortunately do not get more precise information atm, should be fixable though.

wolffcm: The formatting is really nice. I wonder about how errors are rendered in the UI though, is there a way to get the more concise message?

onelson: If it's not possible to present a multi-line error through the UI, I'd bet we could branch the reporting based on a config option.

Marwes: Yep, we can just branch on it and omit the source display. codespan has some support for alternate displays as well https://docs.rs/codespan-reporting/0.11.1/codespan_reporting/term/enum.DisplayStyle.html

Marwes:
Screenshot from 2021-12-03 14-06-37

Rebased and updated the repl/cmd line tool to use the codespan format with coloring.

scbrickley:
This patch is not actually changing the way errors are reported, it's just introducing the capability to use the codespan formatting, right?

If that's the case, this looks fine to me. We can figure out how to (as @onelson said) branch this so that we don't break things in the UI.

libflux/include/influxdata/flux.h
```diff
@@ -104,7 +107,7 @@ void flux_free_stateful_analyzer(struct flux_stateful_analyzer_t *);
 
 // flux_analyze_with will analyze the ast snippet using the flux_stateful_analyzer_t and produce
 // a semantic graph for that snippet.
-struct flux_error_t *flux_analyze_with(struct flux_stateful_analyzer_t *, struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
+struct flux_error_t *flux_analyze_with(struct flux_stateful_analyzer_t *, const char * src, struct flux_ast_pkg_t *, struct flux_semantic_pkg_t **);
```
nathanielc, line 110: I see the valgrind tests are failing in CI, my guess is they need to be updated for this change and a new tests added for the new flux_error_print function.

libflux/include/influxdata/flux.h
```diff
@@ -27,6 +27,9 @@ void flux_free_error(struct flux_error_t *);
 // becomes invalid when the error is freed.
 const char *flux_error_str(struct flux_error_t *);
 
+// Prints the flux error to stdout
+const char *flux_error_print(struct flux_error_t *);
```
nathanielc, line 31: This returns a char* while the Rust implementation doesn't return anything and the Go code doesn't expect a return. Can you update this signature to match?

Do we want to print to stdout from Rust? Or should we make this method a `flux_error_format` instead that builds a string of the contents and let the caller decide to print to stdout. This would make UI errors more possible for example.

nathanielc (changes requested):
This looks great can't wait to use this. Added some comments below mostly about the changes to the C API.

Marwes:
> This patch is not actually changing the way errors are reported, it's just introducing the capability to use the codespan formatting, right?

No, You have to explicitly call the added formatting methods to get codespan formatting.

libflux/include/influxdata/flux.h
Marwes: > Do we want to print to stdout from Rust? Or should we make this method a flux_error_format instead that builds a string of the contents and let the caller decide to print to stdout. This would make UI errors more possible for example.

Building a string and returning it only works for for ANSI compatible terminals so this seemed more general to me. It also auto detects if coloring is supported.

nathanielc (approved):

PR #4272, opened 2021-11-15
fix(stdlib): `length()` should not panic given a table stream

onelson:
Avoid panicking when an array-like table stream is passed to the `length()` function. Instead emit a Flux error.

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/universe/length_test.go
+++ b/stdlib/universe/length_test.go
@@ -2,10 +2,10 @@ package universe_test
 
 import (
 	"context"
-	"github.com/influxdata/flux/runtime"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


stdlib/universe/length_test.go
```diff
@@ -2,6 +2,7 @@ package universe_test
 
 import (
 	"context"
+	"github.com/influxdata/flux/runtime"
```
jsternberg, line 5: Order the imports so standard library imports are in one group and all others are in one underneath it. Make sure they're sorted within the group.

jsternberg (approved):
One code hygiene feedback but otherwise good to merge.

PR #4274, opened 2021-11-16
feat: Make error messages more consistent

Marwes:
I noticed that we lack location information on the errors in the `convert` module so I added it through the `Located` struct. I also refactored the other errors to also use `Located` so that the location information is displayed consistently across all errors (you can see some small alterations to the tested error messages).

Split out and expanded from https://github.com/influxdata/flux/pull/4270 .

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux/src/lib.rs
```diff
@@ -809,12 +809,11 @@ from(bucket: v.bucket)
         let ast = crate::parser::parse_string("test".to_string(), "x = 3 + / 10 - \"");
         let ast = Box::into_raw(Box::new(ast.into()));
         let errh = unsafe { flux_ast_get_error(ast) };
-        assert_eq!(
-            "error at test@1:9-1:10: invalid expression: invalid token for primary expression: DIV
 
-error at test@1:16-1:17: got unexpected token in string expression test@1:17-1:17: EOF",
-            errh.unwrap().err.into_string().unwrap()
-        );
+        expect_test::expect![[r#"
```
rockstar, line 813: I'm not sure if I'm a fan of this library. I feel like it's actually obscuring away what I want. It's more compact, but when I look at the test, I want to know exactly what I'm trying to do.

rockstar (commented):
I think I have real concerns about the value of `expect_test`. I feel like it adds some indirection where I'm not sure there needs to be/should be. Can you help me understand the reasoning behind its use? What do you find it helps you with?

libflux/flux/src/lib.rs
Marwes: https://docs.rs/expect-test/1.1.0/expect_test/ The benefit is that when the error message changes we can just check that the diff looks ok then run `env UPDATE_EXPECT=1 cargo test` and it will automatically update the expected message. For tests like this where the formatting is subject to change and a small change in formatting may lead to many tests changing in a similar way it is really convenient.

rockstar: Yeah, okay, that can certainly be helpful. We have a number of places that have some weird breakages for unrelated reasons, e.g. adding things to the stdlib. Maybe some documentation here would be helpful, but I can't really think of a place where it wouldn't just get lost. Thoughts?

rockstar: Looks like the test tells you how to update it when it fails. Rad.

rockstar (approved):

libflux/flux/src/lib.rs
Marwes: When a test fails on a `expect_test::expect![[]]` macro the error will show how to update the test, which is probably the most direct documentation we can get.

```

---- semantic::formatter::tests::format_function_expression stdout ----


error: expect test failed
   --> flux-core/src/semantic/formatter/tests.rs:114:9

You can update all `expect![[]]` tests by running:

    env UPDATE_EXPECT=1 cargo test

To update a single test, place the cursor on `expect` token and use `run` feature of rust-analyzer.

```

PR #4277, opened 2021-11-17
feat: Provide more exact error information on function calls

Marwes:
Experiment building on what I noticed here https://github.com/influxdata/flux/pull/4270#discussion_r749416169 . Alters `unify` on `Function` such that locations may be injected on the `actual` side and provided for the errors.

Not entirely convinced this is a good idea as it is a bit complicated (and hacky perhaps), but it does fully reuse the existing `unify` function which is at least nice.


wolffcm (approved):
Looks good to me, nice improvement to the locations.

PR #4279, opened 2021-11-18
chore: update to rust 1.56

skartikey:
This patch updates the required rust to 1.56.
This update did not require any code changes in libflux, so should be safe.


wolffcm (approved):
Looks good to me. If I remember correctly, CI won't actually use this change until after this is merged.

Issue #4199, opened 2021-10-28
Add global identifiers to semantic graph


nathanielc:
Semantic graph needs the ability to globally identify builtin functions so that we can find and rewrite them as need.

The motivating case for this change is so that we can rewrite the `is_type` call from https://github.com/influxdata/flux/issues/2159 to a storage level predicate. Same thing would want to be done for the `contains` function. 

If we do not have global identifiers we cannot be sure the a call to `is_type` or `contains` are not to other functions that shadow those names.

Notes for implementation:

- Add global ids to semantic IdentifierExpression as a new field
- IDs will come from the importer in cases of an exported value from a package
- IDs will be generated in cases where the value is local i.e. didn't come from an import 
- Preserve the existing name unmodified for errors and the existing runtime will continue to use it

DOD:

- Semantic graph has a mechanism to globally identify values, this includes all identifiers not just builtin or exported identifiers from packages.
- Flatbuffer serialization is update to communicate any changes in the semantic graph to the Go side
- Tests exist

PR #4283, opened 2021-11-19
refactor: Add Symbol newtype

Marwes:
By adding a new type on "symbols" we allow ourselves to store additional data beside the current literal string value taken from the source (needed for #4199). It also allows for further abstractions such as interning and improved performance in lookups (string comparisons are relatively expensive).


jsternberg (approved):

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -1771,6 +1771,53 @@ impl Property {
     }
 }
 
+// #[allow(missing_docs)]
+// pub type Symbol = String;
+
+#[allow(missing_docs)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Hash, Clone)]
+pub struct Symbol {
```
nathanielc, line 1779: Does Symbol represent the globally unique identifier?  Or does is represent the locally scoped identifier?

Marwes: I'd expect it to be able to represent either, this is just getting the renaming out of the way.

PR #4285, opened 2021-11-19
fix(fluxdoc): validate examples as part of CI lint

nathanielc:
Prior to this change the examples were only run during Flux release.
Now the examples are run during fluxdoc lint to ensure they are
accurate.

Additionally several typos that were missed as a result have been fixed.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


sanderson (approved):
Awesome!

PR #4288, opened 2021-11-22
chore: add CODEOWNERS

nathanielc:
Automatically request review from appropriate teams based on code changed.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+# Automatically request review from a Flux team member for all changes
+/ @influxdata/flux-team
+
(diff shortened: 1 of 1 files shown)


wolffcm (approved):

nathanielc:
There are two layers to CODEOWNERS. First if an owner is specified then the review is automatically requested. Second we can optionally set up branch protections to require approvals from owners. For now I have not enabled any requirements of approval. 

As the standard process is to get an approval from all reviewers I don't feel the need to enforce that explicitly yet.

Maybe once we have a few days with this enabled we can revisit.

PR #4293, opened 2021-11-22
feat(compiler): introduce refactored compiler package

jsternberg:
This introduces a refactored compiler package to `internal/compiler`.
The refactored compiler is designed to take full advantage of the
information from the type system to create an optimized pathway for
runtime execution.

Some important features of this new compiler:
- Preallocates registers for instructions.
- Executes instructions from a main loop instead of recursive descent.

The compiler compiles a function expression into a set of instructions
that operate on an infinite register VM in the mold of an SSA
instruction set. The registers are preallocated with memory for the
given type. Access to identifiers is done through register indices
instead of by name as these are computed at compilation time rather than
at runtime. Similarly, record indices are computed at compile time
instead of determined at runtime.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


jsternberg:
Some benchmarks for this on stuff that's presently implemented and working.

```
benchmark                                                              old ns/op     new ns/op     delta
BenchmarkEval/interpolated_string_expression-12                        534           195           -63.43%
BenchmarkEval/interpolated_string_expression_with_int-12               587           234           -60.25%
BenchmarkEval/interpolated_string_expression_with_duration_type-12     743           343           -53.90%
BenchmarkEval/interpolated_string_expression_error-12                  594           284           -52.14%
BenchmarkEval/interpolated_string_expression_null-12                   553           285           -48.47%
BenchmarkEval/simple_ident_return-12                                   280           50.2          -82.05%
BenchmarkEval/call_filter_function_with_index_expression-12            451           112           -75.17%
BenchmarkEval/conditional-12                                           367           114           -68.82%
BenchmarkEval/unary_logical_operator_-_not-12                          395           124           -68.62%
BenchmarkEval/unary_operator-12                                        405           119           -70.49%
BenchmarkEval/filter_with_member_expression-12                         413           106           -74.26%
BenchmarkEval/regex_literal_filter-12                                  460           176           -61.76%
BenchmarkEval/block_statement_with_conditional-12                      611           141           -77.00%
BenchmarkEval/array_literal-12                                         581           213           -63.31%
BenchmarkEval/record_literal-12                                        2756          95.7          -96.53%
BenchmarkEval/record_with-12                                           4014          114           -97.15%
BenchmarkEval/array_access-12                                          372           96.1          -74.20%
BenchmarkEval/array_access_out_of_bounds_low-12                        676           327           -51.62%
BenchmarkEval/array_access_out_of_bounds_high-12                       582           290           -50.22%
BenchmarkEval/logical_expression-12                                    344           110           -67.88%
BenchmarkEval/two_null_values_are_not_equal-12                         350           94.0          -73.14%

benchmark                                                              old allocs     new allocs     delta
BenchmarkEval/interpolated_string_expression-12                        11             4              -63.64%
BenchmarkEval/interpolated_string_expression_with_int-12               12             5              -58.33%
BenchmarkEval/interpolated_string_expression_with_duration_type-12     16             8              -50.00%
BenchmarkEval/interpolated_string_expression_error-12                  11             5              -54.55%
BenchmarkEval/interpolated_string_expression_null-12                   10             5              -50.00%
BenchmarkEval/simple_ident_return-12                                   5              1              -80.00%
BenchmarkEval/call_filter_function_with_index_expression-12            7              1              -85.71%
BenchmarkEval/conditional-12                                           5              1              -80.00%
BenchmarkEval/unary_logical_operator_-_not-12                          5              1              -80.00%
BenchmarkEval/unary_operator-12                                        6              1              -83.33%
BenchmarkEval/filter_with_member_expression-12                         7              1              -85.71%
BenchmarkEval/regex_literal_filter-12                                  6              1              -83.33%
BenchmarkEval/block_statement_with_conditional-12                      8              1              -87.50%
BenchmarkEval/array_literal-12                                         13             3              -76.92%
BenchmarkEval/record_literal-12                                        44             1              -97.73%
BenchmarkEval/record_with-12                                           59             1              -98.31%
BenchmarkEval/array_access-12                                          6              1              -83.33%
BenchmarkEval/array_access_out_of_bounds_low-12                        11             4              -63.64%
BenchmarkEval/array_access_out_of_bounds_high-12                       8              3              -62.50%
BenchmarkEval/logical_expression-12                                    5              1              -80.00%
BenchmarkEval/two_null_values_are_not_equal-12                         5              1              -80.00%

benchmark                                                              old bytes     new bytes     delta
BenchmarkEval/interpolated_string_expression-12                        608           152           -75.00%
BenchmarkEval/interpolated_string_expression_with_int-12               672           168           -75.00%
BenchmarkEval/interpolated_string_expression_with_duration_type-12     736           208           -71.74%
BenchmarkEval/interpolated_string_expression_error-12                  632           232           -63.29%
BenchmarkEval/interpolated_string_expression_null-12                   608           248           -59.21%
BenchmarkEval/simple_ident_return-12                                   424           32            -92.45%
BenchmarkEval/call_filter_function_with_index_expression-12            552           160           -71.01%
BenchmarkEval/conditional-12                                           424           128           -69.81%
BenchmarkEval/unary_logical_operator_-_not-12                          424           160           -62.26%
BenchmarkEval/unary_operator-12                                        488           160           -67.21%
BenchmarkEval/filter_with_member_expression-12                         504           128           -74.60%
BenchmarkEval/regex_literal_filter-12                                  488           96            -80.33%
BenchmarkEval/block_statement_with_conditional-12                      616           192           -68.83%
BenchmarkEval/array_literal-12                                         496           304           -38.71%
BenchmarkEval/record_literal-12                                        1408          96            -93.18%
BenchmarkEval/record_with-12                                           2528          128           -94.94%
BenchmarkEval/array_access-12                                          488           96            -80.33%
BenchmarkEval/array_access_out_of_bounds_low-12                        680           248           -63.53%
BenchmarkEval/array_access_out_of_bounds_high-12                       600           208           -65.33%
BenchmarkEval/logical_expression-12                                    424           128           -69.81%
BenchmarkEval/two_null_values_are_not_equal-12                         424           96            -77.36%
```

In particular, the record literal and record with ones are the ones I am looking at most.

Some areas that I haven't yet finished.

1. When records are created and used, the creation location and use location, especially when they are copied (like constructing an array literal), can have compatible types but the properties can be in a different order. Because this is the case and this version uses the type system to directly access values by index, we have to take more care with this and resolve the location. An example of this is `[{a: 1, b: "foo"}, {b: "bar", a: 2}]` which is quite nasty.
2. Dynamic records either from builtins are not yet implemented, but I pretty clearly understand how they'll work.
3. Null values are also pretty well understood but haven't been implemented.
4. Function calls have not been implemented. The previous implementation would compile the function on each call. I could do something similar, but I am not considering that complete for this refactor.

Marwes (commented):
There seems to be two distinct changes here, first it changes (adds) the proposed `Value` type, second it refactors the evaluators into an iterative process. While this directly makes it harder to review (as pointed out), it also makes it impossible to know from where any performance improvements comes from.

As for the changes itself, the value type is probably the best we can do in go for a dynamic value type so no issue there.

The evaluator change I am a little skeptical about, changing the lookups to be array based instead of dictionary based and moving to an iterative approach should give performance improvements so that seems like a good idea. However it still sticks with dynamic dispatch over `Evaluator` which is likely slower than going the extra step towards a full byte code interpreter. So it ends up in a sorta halfway space which I don't feel fully able to judge (yet).

Personally I am used to a byte code interpreter as they are generally faster than tree interpreting (less pointer chasing mainly). However there are some impressive results with tree interpreting via GraalVM/Truffle so I wouldn't entirely dismiss trees either https://www.graalvm.org/reference-manual/compiler/

PR #4299, opened 2021-11-24
chore: remove redundant env vars from circle config

wolffcm:
In #4298, `GOPATH` is set in the build dockerfile so that the logged in user has permissions to the Go cache.

In this PR, I'm removing the now redundant `GOPATH` and `GOCACHE` environment variables from the CircleCI config, since they are already set by `Dockerfile_build`. (By default, the Go cache seems to be put under `GOPATH`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,9 +5,7 @@ jobs:
       - image: quay.io/influxdb/flux-build:latest
     resource_class: large
     environment:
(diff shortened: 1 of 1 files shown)


wolffcm:
I think I need to wait for the new `flux-build` image to be created and then run CI again.

Marwes (approved):

Marwes (approved):

PR #4304, opened 2021-11-29
feat: Provide which package exported a symbol

### LINKED ISSUE #4199: Add global identifiers to semantic graph

Marwes:
This ensures that the symbols in the semantic graph knows which package that they originate from. A bit hacky as each package is only exported as a single `PolyType` and that means the fields of that type (such as `types.isType`) would describe the `Symbol`. However due to how intertwined the go interpreter is with the precise serialization of these types I could not change the record fields to hold symbols describing the package. So instead this symbol is manually constructed when we detect that the symbol is imported.

I think it is still fine to review and merge this if someone has the time, but I will try to do some refactoring tomorrow to alleviate this issue so if you want to wait with reviews, there may be a slightly better (though larger) PR tomorrow.

Closes #4199

### Done checklist
- [x] Test cases written


Marwes:
Got the refactoring in, we now re-use the symbols created for each package when they are referenced in another package which is one step closer to making symbols less string-like (doing variable lookups by string is expensive compared to just comparing identity/integers). This is done by doing another separation in the environment/import code, the code now distinguishes a single `PackageExports` from multiple `Packages` (calling it `PackageExports` so it does not conflict with `semantic::Package`, but maybe it should have its own module scope instead of a suffix :shrug: ).

Marwes:
I ended up removing some flatbuffer tests completely as I would need to re-implement the symbol scoping in `convert.rs` in ast to semantic conversion implemented in go which doesn't feel great. The removed tests essentially checked that `ast -> rust semantic convert` gave the same result as `ast -> go semantic convert` gives the same result which I don't think is very necessary as the go conversion only exists for these tests. But if the tests are needed I can try to update the go conversion or try to come up some roughly equivalent tests at least.

libflux/flux-core/src/semantic/mod.rs
```diff
@@ -228,17 +320,14 @@ impl<'env, I: import::Importer> Analyzer<'env, I> {
...
-                    .collect(),
-            },
+        self.env.enter_scope();
+        let env = match nodes::infer_package(&mut sem_pkg, &mut self.env, sub, &mut self.importer) {
+            Ok(()) => {
+                let env = self.env.exit_scope();
+                PackageExports::try_from(env.values)?
+            }
             Err(err) => {
+                self.env.exit_scope();
```
wolffcm, line 330: I might be confused, but can the addition of `enter_scope` and `exit_scope` affect how the `flux repl` works? I am unclear on the details, but somehow it's possible for users to add assignment statements to the topmost scope incrementally so they are visible in subsequent input.

There are no unit tests for the REPL so if it broke somehow I don't think we'd catch it in CI.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -1555,11 +1563,19 @@ impl MemberExpr {
     //
     fn infer(&mut self, infer: &mut InferState<'_, '_>) -> Result {
         self.object.infer(infer)?;
-        let t = self.object.type_of();
+        let t = self.object.type_of().apply(infer.sub);
+
+        if let Expression::Identifier(object) = &self.object {
+            if let Some(package_name) = infer.imports.get(&object.name) {
+                if let Some(property) = infer.importer.symbol(package_name, &self.property) {
+                    self.property = property;
```
wolffcm, line 1571: I am a little hazy in what's going on here, but is this saying that if we find a member expression like `foo.bar` and we have seen an `import "foo"` in the program that we annotate the property with the package `foo`?

Will this work right in places where we've shadowed the import package with an assignment? Looking back at the test cases you wrote in `semantic/tests.rs` I can see it works right but I'm unclear why.

libflux/flux-core/src/semantic/convert.rs
```diff
@@ -1,8 +1,9 @@
 //! Various conversions from AST nodes to their associated
 //! types in the semantic graph.
 
-use std::{collections::BTreeMap, fmt, rc::Rc};
+use std::{collections::BTreeMap, fmt, sync::Arc};
```
wolffcm, line 4: Curious why you changed to `Arc` here, do you expect that something mutable will be accessed concurrently by multiple requests?

wolffcm (changes requested):
I had lots of questions for my own benefit. This is more than I can get my head around at the moment, but it seems reasonable.

There's some code that skips some tests, that should definitely be removed before merging.

libflux/flux-core/src/doc/example.rs
```diff
@@ -29,10 +29,12 @@ pub fn evaluate_package_examples(
     for example in docs.examples.iter_mut() {
         evaluate_example(example, executor)
             .with_context(|| format!("executing example for package {}", path))?;
+        eprintln!("{} ... OK", path);
     }
     for (name, doc) in docs.members.iter_mut() {
         evaluate_doc_examples(doc, executor)
             .with_context(|| format!("executing example for {}.{}", path, name))?;
+        eprintln!("{}.{} ... OK", path, name);
```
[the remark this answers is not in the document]
Marwes: Yes, I thought these tests were hanging as they didn't report any progress. So some printing here is nice.

libflux/flux-core/src/semantic/convert.rs
Marwes: I think `Symbol` ended up inside an `anyhow::Error` so it had to be `Send` due to that.

libflux/flux-core/src/semantic/nodes.rs
Marwes: Yes, because `infer.imports`  is keyed by the `Symbol` for the `import` statement we only replace it if it comes from the import.

libflux/flux-core/src/semantic/mod.rs
Marwes: There are tests that run multiple eval calls where variables from an earlier scope are available in later ones which I assumed were the REPL ones, or at least close to it so we should be good there.

All new bindings (and only the new bindings) are now returned now and the `StatefulAnalyzer` updates its environment with those after `analyze_ast` is called which gives the same behaviour (or close enough).

libflux/flux-core/src/semantic/tests.rs
```diff
@@ -3733,3 +3714,58 @@ fn missing_return() {
...
+    let src = r#"
+            import "types"
+            x = types.isType(v: 1, type: "int")
+
+            foo = () => (1)
+            foo()
+
+            types = { isType: (v, type) => 1 }
+            y = types.isType(v: 1, type: "int")
+        "#;
```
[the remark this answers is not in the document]
Marwes: No it can't find any of those cases. If we want to cover those we should add another pass which propagates `Symbol` (and perhaps other information).

libflux/flux-core/src/semantic/fresh.rs
```diff
@@ -66,6 +67,44 @@ impl<T: Fresh> Fresh for Option<T> {
...
+    fn fresh_ref(&self, f: &mut Fresher, sub: &mut TvarMap) -> Option<Self> {
+        merge_collect(
+            &mut (),
+            self,
+            |_, (k, v)| v.fresh_ref(f, sub).map(|v| (k.clone(), v)),
+            |_, (k, v)| (k.clone(), v.clone()),
+        )
+    }
+}
+
```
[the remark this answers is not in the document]
Marwes: There is also a `impl<T: Hash + Ord + Eq + Fresh + Clone, S: Clone> Fresh for SemanticMap<T, S>`  which has `Fresh` implemented on which conflicted, but I can add `Fresh for String/Symbol/Label` instead and tweak that to fix it.

semantic/symbol_test.go
```diff
@@ -0,0 +1,39 @@
...
+	tcs := []struct {
+		name    string
+		fluxSrc string
+		err     error
+	}{
+		{
+			name: "isType",
+			fluxSrc: `
+				import "types"
+				x = types.isType(v: 1, type: "string")
```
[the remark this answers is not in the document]
Marwes: Added it on the rust side

semantic/monotype.go
```diff
@@ -334,10 +334,12 @@ func (mt MonoType) SortedProperties() ([]*RecordProperty, error) {
 		}
 	}
 	sort.Slice(ps, func(i, j int) bool {
-		if ps[i].Name() == ps[j].Name() {
+		l := NewSymbol(ps[i].Name()).Name()
+		r := NewSymbol(ps[j].Name()).Name()
```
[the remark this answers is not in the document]
Marwes: It used to be, but no longer. Should I remove this maybe?

internal/fbsemantic/semantic.fbs
```diff
@@ -4,6 +4,15 @@ table Fresher {
...
 }
 
+table Packages {
+  packages:[PackageExports];
+}
+
+table PackageExports {
+  id:string;
+  package:TypeEnvironment;
+}
```
nathanielc, line 14: I like these changes to make explicit types for exports from a package and a set of packages. They make the intent of the code clear.

nathanielc (approved):
Looks good to me. Is the `name@package` syntax exposed anywhere? Seems like its always an implementation detail of the Symbol type but I wasn't sure.

wolffcm (approved):

Marwes:
> Looks good to me. Is the name@package syntax exposed anywhere? Seems like its always an implementation detail of the Symbol type but I wasn't sure.

Yes, it is only an implementation detail. I had trouble getting the flatbuffer representation working for a proper struct for some reason so I just left it as a string :(

PR #4309, opened 2021-11-30
refactor: Separate a string based environment out of Environment

Marwes:
`Environment` was used for a wide range of `String => PolyType` mapping, this is despite `Environment` being a relatively complicated type (handling both scoping and readonly maps). As I want to store additional information on a package/import level so that symbols may identify from which package they come from this change allows that extra information to be added to the new `ExportEnvironment` type, leaving `Environment` untouched.

Fixes some potential issues around `Environment` use where only the current `Environment` values where used and the potential parents where ignored.

Refactors the stdlib compilation done in `build.rs` so that each package is only compiled once which reduces the runtime by ~30%


libflux/flux-core/src/semantic/env.rs
```diff
@@ -162,26 +189,6 @@ impl Environment {
...
-    ///
-    /// # Panics
-    ///
-    /// It is invalid to call `pop` on a type environment with only one stack
-    /// frame. This will result in a panic.
-    pub fn pop(mut self) -> Environment {
-        self.exit_scope();
-        self
-    }
-
```
wolffcm, line 184: This was just dead code I guess?

libflux/flux-core/src/semantic/mod.rs
```diff
@@ -51,9 +51,64 @@ pub enum Error {
     Inference(#[from] nodes::Error),
 }
 
+/// An environment of values that are available outside of a package
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct ExportEnvironment {
```
wolffcm, line 56: I wonder if this should be called `ImportEnvironment`?

wolffcm (commented):
This looks good to me, I just had some questions, mostly for my own understanding.

libflux/flux-core/src/doc/mod.rs
```diff
@@ -11,15 +11,15 @@ use std::{
 
 use anyhow::{bail, Result};
 use derive_more::Display;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
```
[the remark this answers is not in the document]
Marwes: `once_cell` is newer and takes advantage of `const fn` instead of macros which I think is prefereable. The `once_cell` API is on track of being merged into rusts standard library as well so it is easier to swap out once that is stabilized https://github.com/rust-lang/rust/issues/74465

libflux/flux-core/src/semantic/env.rs
```diff
@@ -127,8 +143,10 @@ impl Environment {
     pub fn lookup(&self, v: &Symbol) -> Option<&PolyType> {
         if let Some(t) = self.values.get(v) {
             Some(t)
-        } else if let Some(env) = &self.parent {
-            env.lookup(v)
+        } else if let Some(t) = self.external.as_ref().and_then(|env| env.lookup(v)) {
+            Some(t)
```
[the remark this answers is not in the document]
Marwes: I wish it were that easy, but unfortunately `Environment` has been used rather creatively.

* The REPL incrementally adds items to it for each line
* semantic tests (ab)use the readonly and scoping properties to get just the newly added items
* `fn find_var_type` explicitly injects a dummy variable into `Environment` and expects type inference to affect that variable to figure out the type.

So `ExternalEnvironment/external` still reflects this to some extent, hopefully we can formalize this a bit better later.

> So the symbols exposed by the external environment are those that come from import statements, do I have that right?

Specifically for `import` statements, these are only extracted from the separate `importer` struct and then added to the local `Environment` type.

libflux/flux-core/src/semantic/env.rs
Marwes: Yep, I noticed while temporarily changing `Environment` to `pub(crate)`. I think we should eventually keep it internal to the semantic module.

libflux/flux-core/src/semantic/mod.rs
Marwes: Maybe, this is not a great name, though I also use it to "export" the public API after a semantic graph has been analyzed.

wolffcm (approved):
I like the direction this is going, thanks for answering my questions.

PR #4310, opened 2021-12-01
fix(join): Return error if column not found in schema

scbrickley:
This patch guards against a panic triggered by a customer query that passed data with an inconsistent schema into one side of a join call.

To summarize the issue, one of the input streams had a table with a column that none of the other tables had, and the schema for the output was built using one of those other tables. Later in the process, join attempted to look up the index of a column that didn't exist in the schema map, found nothing, and got a default index of `0`. The code did not bother to check whether or not the map lookup was successful, and so it tried to append a value to whatever column was at index `0`, found that it was the wrong type, and panicked.

This patch adds checks to those problematic map lookups, and returns an internal error before the panic is triggered. It also changes the signature of the `ForEach` and `ForEachWithContext` methods in order to more easily propagate that error up the call stack.

This patch is really just a bandaid to avoid panicking. Ideally, when we find a column that is not recognized by the join schema, we would add that column to it, and any time `join` processes a row that does not have that column, it would simply set the row's value for that column to `null` in the final table. This is a longer term fix and may require a greater effort to refactor join, so a separate issue to track that work has been created (See #4315).


sanderson (approved):
Good from a docs perspective.

stdlib/universe/join.go
```diff
@@ -918,30 +930,55 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
...
-						newColumnIdx := c.colIndex[newColumn]
+						newColumn, ok := c.schemaMap[column]
+						if !ok {
+							println("left Column not found in schema")
+							err = errors.Newf(codes.Internal, "column '%s' not found in join schema", columnName)
+							return
+						}
+						newColumnIdx, ok := c.colIndex[newColumn]
+						if !ok {
+							println("left index not found in map")
```
wolffcm, line 947: You'll want to remove this and the other print statements before you merge.

stdlib/universe/join.go
```diff
@@ -604,13 +609,18 @@ func (c *MergeJoinCache) ForEach(f func(flux.GroupKey)) {
...
 // ForEachWithContext iterates over each table in the output stream
-func (c *MergeJoinCache) ForEachWithContext(f func(flux.GroupKey, execute.Trigger, execute.TableContext)) {
+func (c *MergeJoinCache) ForEachWithContext(f func(flux.GroupKey, execute.Trigger, execute.TableContext)) error {
 	trigger := execute.NewTriggerFromSpec(c.triggerSpec)
 
+	var err error
 	c.postJoinKeys.Range(func(key flux.GroupKey, value interface{}) {
+		if err != nil {
+			return
+		}
```
wolffcm, line 623: You could also avoid having to do this error check if this lamdba could return an error, since `Range` could do the check and terminate iteration.

stdlib/universe/join.go
```diff
@@ -918,30 +930,55 @@ func (c *MergeJoinCache) join(left, right *execute.ColListTableBuilder) (flux.Ta
...
+						if !ok {
+							println("left Column not found in schema")
+							err = errors.Newf(codes.Internal, "column '%s' not found in join schema", columnName)
+							return
+						}
+						newColumnIdx, ok := c.colIndex[newColumn]
+						if !ok {
+							println("left index not found in map")
+							err = errors.Newf(codes.Internal, "could not find index for column '%s' in column index map", columnName)
+						}
```
wolffcm, line 949: I wonder if an internal error is what we really want here. Users don't get to see internal error messages, so they won't know how to fix their query when they hit this. It's a little weird to make it a user error since the user is not really at fault here. I wish there was a `not yet implemented` message.

wolffcm (changes requested):
I had some comments. I think that the error handling would become much more intuitive if the function argument to `ForEach` could return an error on its own. Maybe this would require lots more changes, I'm not really sure. But the current design seems prone to errors getting ignored.

stdlib/universe/join.go
scbrickley: I agree that it's a weird situation. I originally had this as `codes.Invalid` thinking that we would want to show an error to the user. After discussing it with @nathanielc though, we came to the conclusion that, because the user isn't at fault here, logging an internal error and letting the table be discarded was the best approach until we can resolve #4315. I'm open to having my mind changed on that though.

wolffcm: Thanks for the rational, I think that makes sense.

wolffcm (approved):
This looks good to me.

PR #4314, opened 2021-12-02
feat: add location/message methods to semantic::Error

nathanielc:
Prior to these message you could not extract the location information
separate from the error message for semantic Errors. Now you can get the
location information explicitly allowing for passing rich diagnostics
in the LSP.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/flux-core/src/semantic/mod.rs
+++ b/libflux/flux-core/src/semantic/mod.rs
@@ -26,29 +26,69 @@ pub mod flatbuffers;
 
 use thiserror::Error;
 
-use crate::{ast, errors::Errors, parser, semantic::types::PolyType};
+use crate::{
+    ast,
+    errors::{Errors, Located},
+    parser,
+    semantic::types::PolyType,
+};
 
-/// Error represents any any error that can occur during any step of the type analysis process.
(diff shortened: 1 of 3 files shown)


Marwes (approved):
Is `semantic::Error` not exported in the public API? Otherwise you could write these matches on the LSP side instead.

Alternatively, something that may be more consistent we could what `nodes::Error` does for `infer::Error` and rewrap the `Location` in `semantic::Error`.

```rust
impl From<infer::Error> for Error {
    fn from(err: infer::Error) -> Self {
        Located {
            location: err.loc,
            error: ErrorKind::Inference(err.err),
        }
    }
}
```

So this would change `semantic::Error` from

```rust
enum Error {
    InvalidAST(Located<ast::check::ErrorKind>)
}
```

into

```rust
pub type Error = Located<ErrorKind>;

pub enum ErrorKind {
    InvalidAST(ast::check::ErrorKind),
}

impl From<> ... 
```

Leaving `location` and `error` easily accessible on the top level error again.

nathanielc:
The change to make `semantic::Error` a `Located` error made the most sense. I have updated the PR.

Issue #4306, opened 2021-11-29
refactor: Library-ify the acceptance test runner


onelson:
Today, the REPL and test harness have two distinct implementations for "running a flux program to completion." Each has some inherent complexity in how it unwinds the graph of side-effects.

As far as I can tell there's no equivalent code in the Flux library itself that can be used for Go-based unit tests.

DOD:

- Move code out of `cmd/flux/cmd/test.go` and into lib modules.
- Refactor as needed to allow for use via Go unit tests.
- ~~*stretch goal* refactor such that REPL can share some common code (this may not be possible, but take a look?)~~ not today

onelson:
- look at similarities between test harness and queryd
- repl may need a big overhaul (do not tackle unification yet)

onelson:
The aim should be to move the test runner part into lib that the `flux test` subcommand can run on a flux program, but leave all the logic about processing individual `testcase` blocks into their own flux programs in the binary.

PR #4317, opened 2021-12-03
test: Add support for RDBMS integration testing

onelson:
With some refactors planned in the `sql` package, we didn't feel comfortable making changes without getting at least some baseline test coverage that runs against live database engines.

The test plan applied in each case is to:

- pre-load some data in each database
- write a new record (via a flux acceptance test)
- verify that we can read both the original seed data and the newly added row (via additional acceptance tests)

Effectively this is a "smoke test" to check that the pipes between flux and the RDBMS are unobstructed.

## How to run these tests

```
make test-flux-integration
```

> The integration tests are filtered out by default when you run `make test-flux` and are explicitly selected when you `make test-flux-integration`

These tests require the following:

- docker to be installed and running
- sqlite3 (the cli tool)

---

We have a long list of supported drivers, and while many are covered by this PR it's not exhaustive.

## Covered Drivers

- mysql <https://hub.docker.com/_/mysql> (additionally, <https://hub.docker.com/_/mariadb>)
- postgres <https://hub.docker.com/_/postgres>
- sqlserver <https://hub.docker.com/_/microsoft-mssql-server>
- vertica <https://hub.docker.com/r/vertica/vertica-ce>
- sqlite3 (no docker; relies on sqlite3 being present on the system, and writes a db file to `/tmp`)

## Pending Drivers

- hdb <https://hub.docker.com/_/sap-hana-express-edition>, <https://www.sap.com/products/hana.html>
- awsathena <https://aws.amazon.com/athena/?nc=sn&loc=0&whats-new-cards.sort-by=item.additionalFields.postDateTime&whats-new-cards.sort-order=desc>
- bigquery <https://cloud.google.com/bigquery>
- snowflake <https://www.snowflake.com/>

I was just not able to get the hdb container available on docker hub to run, despite giving it a good amount of time/effort. I left some notes in the `etc/spawn-db.sh` script about the issues but the TL;DR is *the docker image wants a bunch of kernel tuning/special treatment and also I simply could not get it to run.*

The remaining items all look like they are cloud-hosted stores which would require a bit of a mess if we ever want to run this stuff in CI. Possibly we'd need shared creds and something to help provision storage on demand; I'm not sure what that work would look like.

## Unexpected Work

Note that there's a bugfix smuggled in with this PR in 372db06dd5440e5949d5aa6b060d6dfc4b01306c. During the addition of the test for the Vertica driver, it was discovered that data read from `int` columns was being silently converted to `nil`. We might want to look at a way to make this happen less silently. I'm not sure what the correct approach is here, but I think just about any of the database drivers we leverage today could end up giving us `int` or `uint` values based on the pre-existing schema, so this fix may be more narrow than we really need.

Additionally, it was discovered that the mssql driver (which can be selected using the name `sqlserver` or `mssql`) changes behavior for how it handles parameter binding. It turns out we rely on the behavior provided by the `sqlserver` spelling. We should look to remove mention of the other spelling from our docs to help steer folks towards the happy path. 

## Future Work

#4306 should help give us a way to run these tests in a parameterized fashion while also making the entire setup a little less brittle. 
Today we rely on hard-coded port numbers, named containers, and most importantly the lifecycle of each database is managed "out of band" by this shell script. The flux testcases themselves are comically similar and could be parameterized (trivially) if we were able to feed in the particulars for each case.
By leveraging subtests in Go, we could let docker self-assign ports for each service, the containers could be anonymous (no explicit names) and so on ([proof of concept](https://github.com/influxdata/flux/commit/3ce32aed173c05b89b16b0d5adad304dc8b20aa5)).

Another benefit to moving this work to Go unit tests is the test filtering could be done by build tags (instead of the skip/test lists specified in the Makefile). If we don't elect to dig into the work for #4306 near-term, it might be good to add a way to filter/select flux acceptance tests based on wildcards, globs, patterns, etc.

### Done checklist

- [ ] ~~docs/SPEC.md updated~~
- [x] Test cases written


sanderson (approved):
LGTM from a docs perspective.

onelson:
I'm taking a look at the CI - my `make generate` wasn't good somehow.

onelson:
Eugh, I'm really not sure what's happening here. After doing `make generate` I start getting errors related to the generated files. Even `make fmt` starts failing.

```
flux on  test/db-integration-testing [$?] via 🐹 v1.17.2 
❯ make build
go generate -tags '' ./libflux/go/libflux
env GO111MODULE=on go build -tags '' ./...
stdlib/packages.go:76:2: found packages sql (awsathena.go) and integration (flux_test_gen.go) in /home/onelson/Projects/flux/stdlib/sql
stdlib/packages.go:76:2: found packages sql (awsathena.go) and integration (flux_test_gen.go) in /home/onelson/Projects/flux/stdlib/sql
make: *** [Makefile:76: build] Error 1

flux on  test/db-integration-testing [$?] via 🐹 v1.17.2 
❯ make fmt
go fmt ./...
stdlib/packages.go:76:2: found packages sql (awsathena.go) and integration (flux_test_gen.go) in /home/onelson/Projects/flux/stdlib/sql
make: *** [Makefile:87: fmt-go] Error 1
```

stdlib/sql/integration_test.flux
```diff
@@ -0,0 +1,191 @@
+package integration_test
```
Marwes, line 1:
```suggestion
package sql_test
```

Which should fix the CI failure you are seeing. There can only be one package and and test package per directory :shrug:

onelson: Thank you @Marwes :pray:

stdlib/sql/vertica.go
```diff
@@ -42,7 +42,11 @@ func (m *VerticaRowReader) GetNextRow() ([]values.Value, error) {
 	row := make([]values.Value, len(m.columns))
 	for i, col := range m.columns {
 		switch col := col.(type) {
-		case bool, int, uint, int64, uint64, float64, string:
+		case int:
+			row[i] = values.NewInt(int64(col))
+		case uint:
+			row[i] = values.NewUInt(uint64(col))
+		case bool, int64, uint64, float64, string:
 			row[i] = values.New(col)
```
scbrickley, line 50: Nit: Should we add support for `int` and `uint` in the `values.New()` function? That would allow us to to handle all of these types in one case. Maybe there's a good reason why we don't do that already, but I'm not sure what it is.

scbrickley (approved):
LGTM! I had one small comment/suggestion, but it's not important enough to hold up the PR.

I do wonder though... is this the kind of thing we should run in CI? Or are we fine with just running this on developer machines whenever someone makes a change to the sql package?

stdlib/sql/vertica.go
onelson: I had the same idea, but @jsternberg encouraged the change be limited to the vertica-related code (at least for now).

It does seem to me like widening values that are too narrow should be safe enough in all cases. It also seems like any database driver could give 32-bit numbers based on the pre-existing schema, depending on how the DDL was authored.

Maybe Jonathan can speak to his rationale.

We did talk about perhaps updating `values.New()` to error instead of what amounts to swallowing and converting to a nil, but that sort of change could have wide impact and should come with some time to evaluate the size first.

jsternberg: I'm not sure it would necessarily be bad, but there was a similar section in the influxql code where we had a type switch on an interface and chose a point type for it. We originally had `uint` get translated to integers and it was a bit complicated when we wanted to add support for unsigned types because it was a breaking change in that API. The reason why I was thinking to keep it constrained and avoid any casts is just to avoid any potential ambiguities in that core part of the API.

onelson:
> is this the kind of thing we should run in CI?

I think so, but the setup is currently sufficiently hacky and brittle (ex: hard-coded port numbers and docker container names) that we'd probably want to invest a little more dev time into the workflow first. #4306 could help a lot with the brittle aspects.

Even so, I think the plan would be to try and run these sorts of tests either periodically or based on certain paths in the source tree changing. That should help minimize the time and cpu/memory cost required to run these.

If I remember @nathanielc suggested there be a separate ticket to look at the CI concerns.

stdlib/sql/vertica.go
onelson: Briefly touched on this with Nathaniel just now, and another aspect is how we might like to consider `values.New()` as a convenience only appropriate for "stuff that was already vetted as Flux values." The implication being any use in places like the `sql` package would be "illegal" and we should always be using the constructors for specific Value variants in this case.

I don't know what it'd take to make `values.New()` private to only things that _should be free to use it_, whoever that might be. Alternatively, maybe we add a docstring to coach the usage. :shrug:

Issue #4261, opened 2021-11-11
String literals cause a few weird side effects 


fchikwekwe:
We added a [`reindentation` method ](https://github.com/influxdata/flux/blob/bfb487c13c062dd90faa12940cd24c0c090970a3/libflux/flux-core/src/formatter/mod.rs#L124)to take care of indenting child notes based on their parent node's indentation level in the formatter. When we see a string literal, we force it not to reindent using a [`safe_to_reindent` attribute](https://github.com/influxdata/flux/blob/bfb487c13c062dd90faa12940cd24c0c090970a3/libflux/flux-core/src/formatter/mod.rs#L72). Parent nodes are also marked as unsafe for any further reindentation. This causes some weird indentation outcomes in some cases. 

Here is an example from @wolffcm with a [link to the discussion](https://github.com/influxdata/flux/pull/4101#discussion_r740390333): 
```
 EXPECTED: 
 foo = {
    b: 2,
    c: () => {
        // line 1
        a = 0
        return "a
multi
line
string"
    },
} 
 OUTPUT: 
 foo = {
    b: 2,
    c: () => {
    // line 1
    a = 0

    return "a
multi
line
string"
},
} 
```

After the string literal, we can see that the closing bracket for the function body is improperly indented. 

Here are some tests that were skipped that might be affected by this. Once the formatting is fixed, we should unskip the tests that are now working. 
https://github.com/influxdata/flux/blob/bfb487c13c062dd90faa12940cd24c0c090970a3/libflux/flux-core/src/formatter/tests.rs#L1014-L1599

fchikwekwe:
This is a newer case that appears to still be indenting despite being a string literal.     ```
|> map(
        fn: (r) => ({r with
            newColumn: sendSlackMessage(
                text: "*Earthquake Alert*
M *${string(v: r.mag)}* — *${string(v: r.place)}*",
            ),
        }),
    )
```
turns into:
```
    |> map(
        fn: (r) => ({r with
            newColumn: sendSlackMessage(
                text: "*Earthquake Alert*
            M *${string(v: r.mag)}* — *${string(v: r.place)}*",
            ),
        }),
    )
```
I'd double check that we are using a flux version with the new formatter in it, but this seems like a bug.

PR #4321, opened 2021-12-06
fix: Don't alter the indentation of interpolated strings

Marwes:
Fixes one instance where formatting would alter a string literal (maybe the only instance?). #4261 also mentions an issue with surrounding code getting wrongly indented, however fixing this seems more complicated as the formatter works with just plain (formatted) `String` values, which do not have the information necessary to fix them after the fact.

cc #4261

### Done checklist
- [x] Test cases written

### DIFF
--- a/libflux/flux-core/src/formatter/mod.rs
+++ b/libflux/flux-core/src/formatter/mod.rs
@@ -703,6 +703,12 @@ impl Formatter {
 
     fn format_text_part(&mut self, n: &ast::TextPart) {
         let escaped_string = self.escape_string(&n.value);
+
+        if escaped_string.contains('\n') {
+            self.unindent();
+            self.safe_to_reindent = false;
+        }
+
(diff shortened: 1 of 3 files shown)


libflux/flux-core/src/formatter/mod.rs
```diff
@@ -703,6 +703,12 @@ impl Formatter {
 
     fn format_text_part(&mut self, n: &ast::TextPart) {
         let escaped_string = self.escape_string(&n.value);
+
+        if escaped_string.contains('\n') {
```
Marwes, line 707: Just copy pasted from the string literal format function

Marwes:
@nathanielc  This fixes a bug causing behavior change after formatting whereas #4322 is a start towards fixing the formatting tests that got ignored. Let me know if I should jump on another issue instead.

PR #4322, opened 2021-12-06
refactor: Refactor the AST pretty printer to use pretty.rs

### LINKED ISSUE #4261: String literals cause a few weird side effects

Marwes:
Refactors the pretty printer to build `Doc` values from https://github.com/Marwes/pretty.rs instead of formatting directly into a `String`. By building an intermediate structure which is annotated with information on where we can insert line breaks we no longer need to "reindent" `String` values without knowledge of what the `String` actually contains. Instead we (effectively) "reindent" the `Doc` value which precisely describes where and when it is sound to do so.

Work in progress, but this is a pretty mechanical process of just changing each `fn format_*` call to return `Doc` and append them via the added `docs!` macro, followed by springling with some [group](https://docs.rs/pretty/latest/pretty/struct.DocBuilder.html#method.group) annotations to describe where line breaks may be.

Should fix #4261 and in general make formatting more consistent (by being closer a declarative instead of a procedural approach, saying what formatting should be, rather than how to get there).


nathanielc (commented):
This looks great, I took a look at the API for `pretty` and seems like exactly what we need. As I understand it allows you to create a chain of docs where each doc is made up of text and control markers (i.e. whitespace and nesting). These markers directly represent the structure of how the whitespace should be formatted when rendering the doc, thus allowing for an implementation to understand a max line width and appropriately wrap lines etc without having to understand the contents of what is being formatted beyond whitespace.

When it comes to dynamically rendering expressions as multiline or single line would we use this https://docs.rs/pretty/latest/pretty/struct.DocBuilder.html#method.flat_alt method to achieve that? I.e. provide both alternative structures and then let the render step decide which one to use?

Marwes:
> When it comes to dynamically rendering expressions as multiline or single line would we use this https://docs.rs/pretty/latest/pretty/struct.DocBuilder.html#method.flat_alt method to achieve that? I.e. provide both alternative structures and then let the render step decide which one to use?

Yes, for instance rendering a trailing comma when rendering on multiple lines `[1, 2]` vs

```
[
    1,
    2,
]
```
```
arena.text(",").flat_alt(arena.nil())
```

Got formatting done for the entire AST today but have a bit of work left to correct the formatting.

Marwes:
Formatting should now be good enough, it passes all tests with some alterations, most which should be for the better, some (mostly around) comments for the worse (low priority) and a few which are probably worse, but not much so (can be fixed later).

I have made a few alterations to `pretty.rs` so I will cut a new release for that tomorrow, and do some refactoring to give better names to some methods and de-duplicate a bit but otherwise this is good enough, at least for reviewing.

stdlib/array/from_group_test.flux
```diff
@@ -18,53 +19,55 @@ data = "
...
-            t0: "tagvalue",
-            _time: 2018-12-19T22:14:20Z,
-            _value: true,
-        },
-    ],
-)
-pass = (tables=<-) => tables
-    |> group(columns: ["_measurement", "_field", "t0"])
+input =
+    array.from(
```
nathanielc, line 23: I would have expected `input = array.from(` to fit on a single line. Is the reason it is not formatted unto a single line because the `hang_doc` logic doesn't have explicit support for CallExpressions? 

Not worried about the change, but trying to understand why a newline was needed in this case.

nathanielc: I am going to play around with the branch to see if I can make this change, mostly just to make sure I grasp the new framework.

onelson:
Big diff - didn't get through it all, so I might just not have gotten to the relevant parts. Still, wanted to ask about the outstanding issues (I think you mentioned edge cases around comments in certain positions?)

Separately, I wonder if the maintenance burden on the tests could be reduced by leveraging something for snapshotting. I have seen https://crates.io/crates/insta but never had the opportunity to try it. Maybe there's something out there that could work.

nathanielc (approved):
This is great! I played with it a bit to make sure I understood. Attached a  patch that makes `CallExpressions` hang such that we reduce extra nesting. 
[hang_call.patch.txt](https://github.com/influxdata/flux/files/7707110/hang_call.patch.txt) Does that look like I understand the API?

My patch isn't complete and its expected to be applied. I just wanted to make sure I was roughly understanding how to use the new framework. FWIW My patch introduces extra nesting and so doesn't pass tests.

Marwes:
@nathanielc Yeah, that is the gist of it. The nesting should be removed from the "body" which should fix the extra nesting you are seeing (it is applied automatically when it is constructed with the `union` variants). I think I skipped call expressions because there were some instances where it didn't look great but we can revisit that later.

Marwes:
> Separately, I wonder if the maintenance burden on the tests could be reduced by leveraging something for snapshotting. I have seen https://crates.io/crates/insta but never had the opportunity to try it. Maybe there's something out there that could work.

@onelson Yes, we use `expect_test` in other places for this, however I didn't want to change the tests at the same time as this change. There are a lot of tests which check that the input does not change which may be a bit weird with `expect_test` (the expected value needs to be the input as well) but we can try.

PR #4325, opened 2021-12-07
fix: Mitigate risk of SQL injection in table/column identifiers

onelson:
While any values written to databases via `sql.to()` will be escaped
appropriately via parameter binding, table and column identifiers would not.

It's debatable if we should consider these identifiers a security risk
since the developer (i.e. the author of the flux program) should really
take steps to ensure they can trust their own inputs, especially when it
comes to identifiers.

Still, we can take some naive steps in `sql.to()` to catch elements of
SQL injection techniques by checking for `;` and comments. If these
patterns are found in table or column names, `sql.to()` will now give an
error.

The big downside is this may error even if the identifier was properly
(manually) escaped by the caller. Better to err on the side of caution,
I suppose.

### Done checklist
- [ ] ~~docs/SPEC.md updated~~
- [x] Test cases written


onelson:
Closed in favor of #4328

Issue #4260, opened 2021-11-11
"date" package is not using the option location


dsilletti:
I am using date.truncate() to query the data starting from the beginning of the month or the beginning of the day, in combination with option location with timezone.

The date.truncate() is still working in UTC. 

e.g.

```
import "date"
import "timezone"

option location = timezone.location(name: "Europe/Rome")

from(bucket: "hour")
    |> range(start: date.truncate(t: -2mo, unit: 1mo))

```

the result is starting from midnight UTC and not using the Europe/Rome timezone.

I have the same issue using other functions in the date package, such as:

- date.hour(t: r._time)
- date.monthDay(t: r._time)
- date.weekDay(t: r._time)
- date.month(t: r._time)

the return value is in UTC and not according the option location

nathanielc:
DOD:

- Add support for `location` option to each of the date.* function where appropriate
- Add support for `location` option to any duration math functions i.e. experimental.addDuration experimental.subDuration

dsilletti:
any update on this? thanks

skartikey:
@dsilletti The PR is up and being reviewed by @jsternberg

OdneOksavik:
Hi. Will this thread be updated when this fix is implemented in the latest stable release? If so I will subscribe to it, we have now added some hours to fix this issue for ourselves.

nathanielc:
This issue was fixed in Flux version 0.149.0 see https://github.com/influxdata/flux/releases/tag/v0.149.0 

As for the next stable release of InfluxDB it will contain this fix, we have already merged this fix into that project. However this issue will not get an update when that release occurs. 

FWIW you can determine the Flux version you are using with this query https://github.com/influxdata/flux/blob/master/stdlib/runtime/runtime.flux#L12-L16

OdneOksavik:
Thank you @nathanielc ! I can now see that we are running v.0.139.0.
Then I make a note of updating queries when we update InfluxDB.

PR #4327, opened 2021-12-07
feat(date): Adding location option support for date package

skartikey:
1) Added support for location option to each of the date.* function where timezone makes sense.
2) Added support for location option to duration math functions, experimental.addDuration experimental.subDuration
3) Refactored the existing code as there were duplication in the existing code.
4) Go & acceptance test added in date package

fixes: #4260

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


skartikey:
Remaining work - 
1) Add more go and acceptance test
2) Add support for location option to any duration math functions i.e. experimental.addDuration experimental.subDuration

stdlib/date/hour_duration_test.flux
```diff
@@ -29,9 +37,31 @@ outData = "
...
+
+    testing.diff(got: got, want: want)
+}
+
+testcase hour_duration_location {
+    got = csv.from(csv: inData)
+        |> range(start: 2018-01-01T00:00:00Z)
+        |> map(fn: (r) => ({r with _value: date.hour(t: duration(v: r._value), location: {zone: "America/Los_Angeles", offset: 0h})}))
+
+    want = csv.from(
```
scbrickley, line 51: It might be worth it to convert these to `array.from` calls, just for the sake of readability.

scbrickley (approved):
Looks good to me. Just had one suggestion, but I think this is okay to merge as is.

stdlib/date/hour_duration_test.flux
```diff
@@ -29,9 +37,31 @@ outData = "
...
+",
+    )
+
+    testing.diff(got: got, want: want)
+}
+
+testcase hour_duration_location {
+    got = csv.from(csv: inData)
+        |> range(start: 2018-01-01T00:00:00Z)
+        |> map(fn: (r) => ({r with _value: date.hour(t: duration(v: r._value), location: {zone: "America/Los_Angeles", offset: 0h})}))
```
jsternberg, line 49: We shouldn't have any tests that directly call this with `{zone, offset}`. Please use the functions in the `timezone` package as that's the intended use. If someone reads the tests, we don't want them attempting to use the internal representation directly.

You can also likely use `option location = timezone.location(name: "America/Los_Angeles")` as part of the test. I think that as long as it is in the testcase block you can use it in different locations. I am not completely certain of that though.

internal/locationutil/location.go
```diff
@@ -0,0 +1,52 @@
+package locationutil
```
jsternberg, line 1: I've been looking at this package and I don't think this is what we want to create a clean API. I'm suspecting that it would be easier if, instead of trying to duplicate window code here, we just implemented the logic directly inside of the date package.

It seems to me the core logic is inside of `ToUTCTime` so let's just take that method, let it take a location name, an offset, and the original time, and then have it return a `values.Time` that corresponds to the appropriate translated time. Something like:

```
func getTimeInLocation(t values.Time, location string, offset values.Duration) (values.Time, error)
```

The function can load the location by the name, apply the offset, then return the translated time. We'll avoid this package and then just use the one function we need rather than try to create a general pathway through two areas that aren't very similar beyond using the same struct attributes.

I understand this function will likely be needed in both `date` and `experimental`. We can also put this function in a third `internal/date` package which can be used by both.

internal/locationutil/location.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return LocationZone{}, err
+	}
+	loc.Offset = l.Offset
+	return loc, nil
+}
+
+func (l Location) ToUTCTime(t values.Time) values.Value {
+	// load location
+	LocationZone, err := l.Load()
```
jsternberg, line 36: It is really important we do not throw away this error. I think I get that this code is meant to be reused within the date package, but that means this code should be inside of the date package rather than a util package. Under no scenario should it ever throw away the error and return a valid value.

This function also doesn't follow the typical Go naming conventions. Local variables are intended to be named with `mixedCaps`. This helps differentiate them from exported functions within the same package.

internal/locationutil/location.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	loc, err := LoadLocation(name)
+	if err != nil {
+		return LocationZone{}, err
+	}
+	loc.Offset = l.Offset
+	return loc, nil
+}
+
+func (l Location) ToUTCTime(t values.Time) values.Value {
```
jsternberg, line 34: `ToUTCTime` isn't a good name as it doesn't describe what this function does. We've avoided using the name `UTC` in the timezone code as to not confuse it with `UTC` as a concept.

jsternberg (changes requested):
Related to the comment about the location name and offset together, we don't intend to expose that functionality in an easily accessible way to end users. We'll support the capability in the code just because it doesn't really add a lot of cost, but it's not a workflow we formally support.

For using the location option, the `timezone` package utilities should be the ones used to create the location record that gets passed in. The tests should, if possible, demonstrate that the option works properly with them.

interval/window.go
```diff
@@ -402,3 +402,10 @@ func (l Location) Equal(other Location) bool {
 	}
 	return l.zone.String() == other.zone.String() && l.Offset == other.Offset
 }
+
+func (l Location) Zone() *zoneinfo.Location {
```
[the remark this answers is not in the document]
skartikey: This is an outdated code.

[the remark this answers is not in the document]
jsternberg: Ah I must have made this comment when I started an older review and it accidentally included it when I did the newer one. Discard this. I thought it was posted in a previous review but I could only see it on my own screen so I couldn't tell it wasn't published.

internal/locationutil/location.go
skartikey: implemented!

nathanielc:
@jsternberg 

Given

> but it's not a workflow we formally support.

Should we explicitly forbid the use of location and offset together? Seems likely, given that @skartikey thought it was expected, that users will try to use both together.

jsternberg (approved):

stdlib/experimental/experimental.flux
```diff
@@ -12,11 +12,15 @@
...
 
+// builtin _addDuration used by addDuration
+builtin _addDuration : (d: duration, to: T, location: {zone: string, offset: duration}) => time where T: Timeable
+
 // addDuration adds a duration to a time value and returns the resulting time value.
 //
 // ## Parameters
 // - d: Duration to add.
 // - to: Time to add the duration to.
+// - location: Location to use for the time value.
```
pierwill, line 23: Make sure we're using consistent capitalization here and other parameters.

pierwill (approved):
👍

2opremio:
@skartikey I would had expected `date.truncate()` to also get location support. Why wasn't that the case?

As it it right now, `truncate()` does the truncation in UTC, e.g:

````
option location = timezone.location(name: "Europe/Madrid")

date.truncate(t: 2019-06-03T13:59:01Z, unit: 1d) => 2019-06-03T00:00:00Z
date.truncate(t: 2019-06-03T13:59:01Z, unit: 1mo) => 2019-06-01T00:00:00Z
date.truncate(t: 2019-06-03T13:59:01Z, unit: 1d) => 2019-01-01T00:00:00Z
```

However, I would had expected to take into account the current location. For instance, in "Europe/Madrid" "2019-06-03T00:00:00" and "2019-06-01T00:00:00" both lands in CEST (+02:00) but the beginning (2019-01-01) is at CET (+01:00). Thus I would expect the following location-aware behavior:

```
option location = timezone.location(name: "Europe/Madrid")

date.truncate(t: 2019-06-03T13:59:01Z, unit: 1d) => 2019-06-02T22:00:00Z
date.truncate(t: 2019-06-03T13:59:01Z, unit: 1mo) => 2019-05-31T22:00:00Z
date.truncate(t: 2019-06-03T13:59:01Z, unit: 1d) => 2018-12-31T23:00:00Z
```

I am happy to create a separate ticket if you think that makes sense.

2opremio:
To elaborate further, this is useful in scenarios similar to https://chrispnerd.blogspot.com/2021/12/i-have-application-that-feeds-periodic.html but in which you want to find precise location-aware boundaries.

Without a way to do location-aware truncation location-aware windowing can create "leftover" windows due to the UTC vs local-zone mismatches.

Related: https://github.com/influxdata/flux/pull/4341 
CC: @nathanielc

skartikey:
@2opremio Thank you for suggesting to make truncate to be location-aware. I have opened an [issue](https://github.com/influxdata/flux/issues/4656) for the team to discuss this.

PR #4328, opened 2021-12-07
fix(stdlib/sql): quote db identifiers

onelson:
The aim here is to ensure all _identifiers_ (ex: table/column names) being added to our SQL statements are _quoted and escaped_ to help mitigate risk of SQL injection.

Quoting and escaping happens in a couple places:

- in `sql.to` itself where we generate-then-exec SQL statements
- in each driver's "translate column" function

The translate column function's job is to produce appropriate DDL for use in a `CREATE TABLE` statement and as such, the content needs to be formatted appropriately to contain characters that would otherwise be invalid (not necessarily malicious).

Table names are not technically a part of the attack surface for Flux since it is specified as a string argument to `sql.to` putting it into the "trusted input" category; a string present in the flux program source.
Column names are _technically a risk_ since they are derived from the field names on incoming records from arbitrary sources.

The database integration tests have been updated to purposefully require quoting as a way to exercise the new code changes (the table name and new column has a space in it). 

These changes have been _manually tested_ against BigQuery and Snowflake. _A write-up on the process for manually testing will land in the internal docs soon._

### Prior quoted identifiers

Folks who used `sql.to` prior to this change and happened to need to quote their identifiers themselves will need to remove _their quoting_ to avoid their quotes being escaped and included in the identifier name by _our quoting_.

### ~~Snowflake's~~ Identifier Casing

Snowflake has a quirk when it comes to identifiers. If the identifier is a bare
word (ie, _not quoted_) then it'll be automatically uppercased.
This means bare word identifiers, regardless of how they are cased, match since
they are all normalized to uppercase. 

Since we're aiming to quote everything we generate in Flux (previously this was
not the case), this means callers will need to match the case of the objects in
their database.

In the case of a _new table_ being named in `sql.to`, flux will issue SQL to
create the table based on the incoming record type. The identifiers in the DDL
will (now) be quoted in this situation and so their casing will be preserved.
Assuming the identifier wasn't _already_ uppercased, this technically adds a new
requirement to now _always quote the identifier_ when it appears in queries.

I suspect it is what it is - if we need to quote all identifiers to mitigate
risk of injection, then Snowflake users will just need to be aware of the quirk
and take it into account.

> SAP HANA has a similar behavior, but continues to use the established practice of forcing identifiers to UPPERCASE (which comes with some problems of its own, noted via comments in the diff).

The big breaking change in this diff comes from the behavior changes we see in case sensitivity, and this isn't limited to Snowflake alone. Almost all of the supported DB engines (except perhaps Vertica) normalize bare word identifiers to lower case. A few normalize to upper case. Regardless of which way they normalize, it's still an opportunity for a casing mismatch once the identifiers are being quoted.

### Done checklist
- [x] Test cases written


stdlib/sql/to.go
```diff
@@ -284,37 +298,91 @@ func getTranslationFunc(driverName string) (func() translationFunc, error) {
...
+		return nil, nil, nil, err
+	}
 
 	labels := make(map[string]idxType, len(cols))
 	var questionMarks, newSQLTableCols []string
 	for i, col := range cols {
-		labels[col.Label] = idxType{Idx: i, Type: col.Type}
+
+		quotedColumName := quoteIdent(col.Label)
+		labels[quotedColumName] = idxType{Idx: i, Type: col.Type}
```
onelson, line 372: Using the quoted form of the identifiers here is what caused the existing tests to require an update. I think ideally, the quoting would happen as close to the string formatting as possible, but I elected to quote earlier (here) to make the mapping here less error prone.

This is to say, I found myself unsure when I should be using quoted or unquoted labels in the rest of the func body. Changing to an "always quoted" approach simplified the flow (for me).

onelson:
Prior to this diff, we had code in place to handle _table and column_ identifiers for hdb.

I originally missed that column names were being escaped at all! This was happening as a part of the `translateColumn` implementation for hdb, and I didn't notice. I may move to refactor the rest of the engines to follow this practice, though it pushes the quote/escape farther from the string formatting which I noted before as being desirable. When the escaping happens right at the point where the string is formatted, the safety of it becomes more self-evident. The way I originally missed these identifiers were being quoted in the first place sort of supports this notion.

At time of writing, my changes to double-quote column identifiers was "stacked" on top of the original work. I note that the `hdbEscapeName` will trim outer double quotes, meaning the inputs for these tests should pass through `hdbEscapeName` unchanged, or at most with the case changed (`"name"` becomes `"NAME"`, ex: https://go.dev/play/p/6EROFIHedH8).

Everything appeared to work fine if we explicitly match case (which was part of what the original work hoped to avoid).

Since _work was already in place_ to address the SQL injection risk for hdb, I decided to flip back to `master` to see if manual testing hdb _with identifiers that require quoting_ (as is the new standard introduced by this PR).

In testing it didn't appear to be working as expected. Using hdb with `sql.from` in a REPL on `master` works well but `sql.to` has problems. 

```
> array.from(rows: [{name: "Tom", age: 20 }]) |> sql.to(driverName:"hdb", dataSourceName: dsn, table: "pet info")
Result: toSQL1
Error: runtime error @1:46-1:110: to: SQL Error 288 - cannot use duplicate table name: "SYSTEM"."(DO statement)": line 9 col 9 (at pos 276): PET INFO: line 1 col 14 (at pos 13)
```

I suspect there's an issue with the "create table if not exists" proc. hdb doesn't support the "if not exists" part so a custom proc was written to fill the gap.

For now, my next steps will be to smoke out the defect that prevents this usage from working for hdb using the codepath currently in `master`. This will either work or it'll break casing-wise. My assumption is it should work (lower-case automatically becomes uppercase) so long as the table DDL included uppercase or bare column names in the first place.

---

As an aside, my feeling is we should remove the case transformation we have for hdb identifiers in favor of using the supplied inputs as-is, only quoted and escaped. My rationale is with the current behavior, it seems impossible to write to tables that were created with quoted identifiers that happen to _not be uppercase_. This behavior is surprising. By automatically uppercasing, we would require users to alter their schema to conform with this assumption (if they aren't already conforming).

Original discussion on the quoting strategy is here: #3098

onelson:
Adding proper quoting to the table name in spots where it was not quoted and removing the "stacked" quote/escape I introduced with this PR shows that when creating the table with an upper cased ident, `"PET INFO"`, in the DDL caused the "create table if not exists" to incorrectly run (and fail) when `sql.to` gets `"pet info"` as the table parameter.

This means the upper casing and quoting is working at least!

```
failures:

        integration_hdb_write_to...fail: runtime error @56:12-56:94: to: SQL Error 288 - cannot use duplicate table name: "SYSTEM"."(DO statement)": line 9 col 9 (at pos 273): PET INFO: line 1 col 14 (at pos 13)
```

With the DDL using the lower cased `"pet info"` ident as it was originally, we see a different failure. In this situation I'm not exactly sure where this is going wrong:

```
failures:

        integration_hdb_write_to...fail: sql: transaction has already been committed or rolled back: runtime error @56:12-56:94: to: SQL Error 259 - invalid table name:  Could not find table/view PET INFO in schema SYSTEM: line 1 col 13 (at pos 12)
```

More research required.

stdlib/sql/hdb.go
```diff
@@ -209,22 +209,32 @@ END;
...
 // hdbAddIfNotExist adds SAP HANA specific table existence check to CREATE TABLE statement.
 func hdbAddIfNotExist(table string, query string) string {
-	var where string
+	var where, schema, tbl string
 	var args []interface{}
-	parts := strings.SplitN(strings.ToUpper(table), ".", 2) // schema and table name assumed uppercase in HDB by default (see Notes)
-	if len(parts) == 2 {                                    // fully-qualified table name
+	// Schema and table name assumed uppercase in HDB by default (see Notes)
+	// XXX: since we are currently forcing identifiers to be UPPER CASE.
+	//  Shadowing the table param ensures we use the UPPER CASE form regardless
```
onelson, line 216: In the original impl, the `CREATE TABLE` DDL would include the input string as-is if the schema wasn't specified (a bug that surfaced when we started quoting all identifiers).

stdlib/sql/sql_test.flux
```diff
@@ -13,42 +14,58 @@ pgDsn = "postgresql://postgres@127.0.0.1:5432/postgres?sslmode=disable"
...
+            dataSourceName: hdbDsn,
+            // n.b. we must explicitly UPPER CASE the table name here to match the DDL.
+            query: "SELECT name, age, \"FAV FOOD\" FROM \"PET INFO\" WHERE seeded = true",
+        )
 
-    testing.diff(got: got, want: want)
+    testing.diff(
+        got: got,
+        want:
+            union(tables: [array.from(rows: [STANLEY]) |> debug.opaque(), array.from(rows: [LUCY]) |> debug.opaque()])
```
onelson, line 45: I had originally wanted to have all the "want" unions to be defined at the head of the flux script, but the test harness started reporting "0 of 0 tests" when I did this.

I tried wrapping them in functions but got the same result.

onelson:
> More research required.

There was a bug in the hdb helper that generates the "create table if not exists" where the table identifier would be uppercased in one situation or otherwise it would be used as-is. For the test input, this meant we had a table created during the seed phase (with an uppercase table name) and a 2nd table being created with a lowercase name. Confusing! :woozy_face: 

In addition to this, there are 2 main places where identifiers need to be quoted:
- in `sql.to` where we generate the queries to run
-  _and also_ in each driver's "column translation" function

I initially didn't cover the translate functions, but the need for this became apparent as I refactored and tidied up a bit more.

jstirnaman (approved):
LGTM. @sanderson, in case you're interested.

stdlib/sql/to_privates_test.go
```diff
@@ -129,8 +132,8 @@ func TestSqliteTranslation(t *testing.T) {
 			t.Log(cmp.Diff(nil, err))
 			t.Fail()
 		}
-		if !cmp.Equal(columnLabel+" "+dbTypeString, v) {
-			t.Log(cmp.Diff(columnLabel+" "+dbTypeString, v))
+		if !cmp.Equal(quote(columnLabel)+" "+dbTypeString, v) {
+			t.Log(cmp.Diff(quote(columnLabel)+" "+dbTypeString, v))
 			t.Fail()
```
wolffcm, line 137: I prefer a more concise style that I've seen in our Go repositories:
```Go
if diff := cmp.Diff(quote(columnLabel)+" "+dbTypeString, v); diff != "" {
  t.Errorf("quoted ID does not match: (-want/+got):\n%s", diff)
}
```
But this is a matter of personal style.

stdlib/sql/to_privates_test.go
```diff
@@ -117,6 +115,11 @@ func TestSqliteTranslation(t *testing.T) {
 		"DATETIME": flux.TTime,
 	}
 	columnLabel := "apples"
+	quote, err := getQuoteIdentFunc("sqlite3")
+	if !cmp.Equal(nil, err) {
+		t.Log(cmp.Diff(nil, err))
+		t.Fail()
+	}
```
wolffcm, line 122: It would be totally acceptable (and I think easier to read) to do this here instead:
```Go
if err != nil {
  t.Fail(err)
}
```

stdlib/sql/sql_test.flux
wolffcm: That seems strange, and doesn't fit with my understanding of how tests are run. I would be curious to know the test case that gets that behavior.

stdlib/sql/sql_test.flux
```diff
@@ -13,42 +14,74 @@ pgDsn = "postgresql://postgres@127.0.0.1:5432/postgres?sslmode=disable"
...
+        |> yield()
+}
 
-    testing.diff(got: got, want: want)
+testcase integration_hdb_injection {
+    EVIL
+        |> sql.to(driverName: "hdb", dataSourceName: hdbDsn, table: "injection attempt", batchSize: 1)
+        |> filter(fn: (r) => false)
         |> yield()
 }
```
wolffcm, line 78: Is this a test case with no `diff`? How would this fail?

wolffcm: I guess you just want to show that the test case succeeds with no errors, even if you don't verify the data write to the database?

wolffcm (commented):
Overall this looks very nice. I just had some questions about the tests and some subjective style items.

I saw the "Caveats" section in the description. Are there any other cases where things will not work or work differently now? 

For example, if we didn't quote a table name when doing a `CREATE IF NOT EXISTS` before, then will we create a separate, new table when this change is deployed? 

It seems like there might be more ripple effects than what you've described. Most likely they are fine, I would just like to understand what they are and be sure we're not missing any.

stdlib/sql/sql_test.flux
onelson: I can try and work up a minimal repro case. I'd sort of gotten used to bumping into these edges and just dismissing as me doing something odd, but if that's not the case I can open a separate issue for it.

stdlib/sql/sql_test.flux
onelson: This is something that makes these tests fragile imo. I tried to explain these with a block comment here:

https://github.com/influxdata/flux/pull/4328/files#diff-9a8520c5dc40773f27979e736664f3082b8257ba14b495f2c616319eb8ed6884R33-R37

Since _writing_ to a database with `sql.to` is not something we can verify the outcome of within a single test, we rely on others after the fact to bare witness to this. Subsequent tests will fail if this test case succeeds but our injection mitigation doesn't hold.

The test _can fail_ if the client code in `sql.to` produces an error, such as not being able to connect to the database, or not having sufficient `GRANT`s to create the table if needed. For our purpose, yes. We just want to know _something was sent without error_.

onelson:
> I saw the "Caveats" section in the description. Are there any other cases where things will not work or work differently now?
>
> For example, if we didn't quote a table name when doing a CREATE IF NOT EXISTS before, then will we create a separate, new table when this change is deployed?
>
> It seems like there might be more ripple effects than what you've described. Most likely they are fine, I would just like to understand what they are and be sure we're not missing any.

Let me see if I can be exhaustive without getting into the weeds.

Where the heft of the change is a move from case insensitive to case sensitive, the places where this will cause problems are where the input records (which inform the `CREATE TABLE` DDL) differ in casing to _schemas that __existed prior__ to this change_.

This is to say, if the table name from prior runs didn't normalize to the same name case-wise before and after the change, Flux will try to recreate the table.
The same goes for column names.

I called out Snowflake specifically since it's in the minority in how it normalizes bare word identifiers (casing UP), but now that I give it another thought, this problem exists for almost all the supported engines. It just so happens most engines normalize by casing down instead of up. _SAP HANA_ also cases UP, but we had code in place to try and force things to be upper case for this engine so SAP HANA was already sort of broken but in a more complicated way to explain (I tried to explain in code comments).

For example, in postgres:

```
-- bare word idents are normalized to lowercase
create table Abc (n int); -- table name is actually "abc" not "Abc"
create table "AbC" (n int);
create table "aBc" (n int);
insert into abc (n) values (111);
insert into Abc (n) values (222);
insert into "AbC" (n) values (333);
insert into "aBc" (n) values (444);

-- each returns rows from the first table: 111, 222
select * from abc;
select * from "abc";
select * from AbC;

-- returns 333
select * from "AbC";

-- returns 444
select * from "aBc";
```
I think Vertica is the odd one out in that it's case insensitive always, no matter what.


**Another situation** where we'd see breakage would be for users who are writing to tables that required quoted identifiers all along. They would have had to quote the identifiers themselves, and when this change merges our code will escape their quotes which would mean the identifiers would no longer match the established schema.


With all this, the chance of this impacting users is hard to anticipate. It depends greatly on the specifics of what the usage looked like.

stdlib/sql/sql_test.flux
onelson: So, whatever was happening before is no longer happening today. I suspect there was a syntax error in the flux script someplace and I was mistaken about the cause. I'm pushing a rev that hoists these unions up to the top of the script. :clap:

wolffcm (approved):

onelson:
Looks like we're going to go here. Holding the PR until tomorrow as discussed during today's standup.

jay-influx (approved):
Looks great! This covers or documents all the edge cases I could think of, including the Snowflake casing semantics changes.

Issue #4186, opened 2021-10-26
Add vectorization support to Flux `map` function


wolffcm:
With #4184 and #4185 complete, it should be possible to execute very simple vectorized functions from `map` in Flux.

The work of this issue will be to execute lambdas using the `compiler` package, passing the table chunk as input, by creating an instance of `values.Value` that is the table chunk as an object with vector fields. An example of this is here:
https://github.com/influxdata/flux/pull/3930/files#diff-ca52e5e5693fff638788b406cc69092210f978d482989588503142cf525b95f4R288

When the function has completed executing, the map transformation should convert the output value into an instance of `flux.Table`.

Add a new feature flag that will make `map` execute the vectorized version of a function if it exists.

DOD:
-  New feature flag added
- `map` uses vectorized function if it exists and feature flag is enabled
- There should be existing acceptance tests that demonstrate that this all works end-to-end. They might already exist. If they don't, add them

PR #4329, opened 2021-12-08
feat(stdlib): create a version of map that is columnar

jsternberg:
The columnar version of map will run map over an entire table chunk
rather than running it and then appending per row. This version of map
should be more efficient in the normal cases and be more amenable to
vectorized functions.

Related to #4186.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/universe/map2.go
```diff
@@ -0,0 +1,276 @@
...
+		builders[i] = arrow.NewBuilder(col.Type, mem)
+	}
+	return builders
+}
+
+// execute will execute the function on each row of the table chunk.
+// It will return the new columns after the function was executed.
+// These columns are not organized by group keys and may require regrouping
+// if a group key column no longer contains consistent values.
+func (m *mapTransformation2) execute(chunk table.Chunk, fn *execute.RowMapPreparedFn, mem memory.Allocator) ([]flux.ColMeta, []array.Interface, error) {
```
jsternberg, line 72: This should be the only function that has to get modified for the vectorized version. The return value is exactly what we need and the parts around this all rely on the same output, but this function would need to be updated to also execute the vectorized version.

wolffcm: So if the function that we have here has a type like `(r: {a: v[int], b: v[int]}) => {c: v[int]}`, could we expect that the above code in `createSchema` will just work?

Also, can `createSchema` be a stand-alone function rather than a method?

jsternberg: It kind of depends on the type that gets returned. We may have to teach create schema to look at the element type if the function is vectorized.

It would be possible for `createSchema` to be a standalone function, but I usually try to scope these by making them method calls so I don't have to worry about name conflicts and using overly generic names.

rockstar (approved):

stdlib/universe/map2.go
```diff
@@ -0,0 +1,524 @@
...
+func newMapTransformation2(ctx context.Context, id execute.DatasetID, spec *MapProcedureSpec, mem memory.Allocator) (execute.Transformation, execute.Dataset, error) {
+	var fn mapFunc
+	// if spec.Fn.Fn.Vectorized != nil {
+	// 	fn = &mapVectorFunc{
+	// 		fn: execute.NewVectorMapFn(
+	// 			spec.Fn.Fn.Vectorized,
+	// 			compiler.ToScope(spec.Fn.Scope),
+	// 		),
+	// 	}
+	// } else {
```
scbrickley, line 39: This seems like an important check. Is there a reason we're skipping it?

execute/row_fn.go
```diff
@@ -206,7 +214,7 @@ func NewTablePredicateFn(fn *semantic.FunctionExpression, scope compiler.Scope)
 }
 
 func (f *TablePredicateFn) Prepare(tbl flux.Table) (*TablePredicatePreparedFn, error) {
-	fn, err := f.prepare(tbl.Key().Cols(), nil)
+	fn, err := f.prepare(tbl.Key().Cols(), nil, false)
```
scbrickley, line 217: I'm confused about why we're passing bool literals as arguments to `prepare`. I thought we would be checking whether the function expression has a non-nil value for the `Vectorized` field, and setting the argument for `vectorized` based on that check.

scbrickley (commented):
Overall this looks good. I had a few questions that I'd like to discuss before this merges.

stdlib/universe/map2.go
jsternberg: Whoops.

execute/row_fn.go
jsternberg: This is after the check for vectorization. We've now either passed the vectorized function or not. It's mostly because, if it's vectorized, then the input columns need to be wrapped in vector types. Rather than expose that code to the map transformation, I chose to just add a boolean to this private helper struct.

scbrickley (approved):

wolffcm (approved):
I had some comments and questions, but nothing to keep you from merging. 

Really nice to see this come together!

stdlib/universe/meta_query_keys_test.flux
```diff
@@ -81,6 +81,7 @@ t_meta_query_keys = (table=<-) => {
             |> range(start: 2018-05-22T19:53:26Z)
             |> filter(fn: (r) => r._measurement == "cpu")
             |> keys()
+            |> sort()
```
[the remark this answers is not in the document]
jsternberg: The old map ordered the columns by using the sorted properties. This isn't necessarily correct. It's really supposed to be the order that the type system tells us. The new map used what the type system tells us about the order. These tests produce values depending on the column order which, in my opinion, should be treated as non-deterministic since the only reason why the order differed was because a map was inserted.

stdlib/universe/map2.gen.go
```diff
@@ -0,0 +1,129 @@
...
+	// All values are non-null so check if they are all the same.
+	v := arr.Value(0)
+	for i, n := 1, arr.Len(); i < n; i++ {
+		if arr.Value(i) != v {
+			return false
+		}
+	}
+	return true
+
+}
```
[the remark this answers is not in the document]
jsternberg: They could, but this is a very large package and I didn't want to add free functions to it. Maybe I should refactor this to `arrowutil` so it's in a common location?

execute/vector_fn.go
```diff
@@ -0,0 +1,85 @@
...
+
+	res, err := f.fn.Eval(ctx, f.args)
+	if err != nil {
+		return nil, err
+	}
+
+	// Map the return object to the expected order from type inference.
+	// The compiler should have done this by itself, but it doesn't at the moment.
+	// When the compiler gets refactored so it returns records in the same order
+	// as type inference, we can remove this and just do a copy by index.
```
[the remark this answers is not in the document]
jsternberg: This would probably be too difficult. I can make an issue, but my thought is that it's not really worth doing except as part of a larger overhaul of the compiler runtime.

stdlib/universe/map2.gen.go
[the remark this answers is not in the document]
jsternberg: Done.

PR #4335, opened 2021-12-09
fix: ensure arrays are not table streams before calling `Len()`

onelson:
`TableObject` currently implements the `Array` interface but all the
methods panic, explaining they are unsupported.

This change aims to avoid panicking and instead give an error message
suited to each case.

A new interface is introduced, `values.TableObject,` which allows type
assertions to be made without importing `flux` (to get access to
`flux.TableObject`). This is a tactic to avoid cyclical imports. For
cases where the module already had access to `flux.TableObject`, it's
used directly. For most other cases, `values` was already imported
making the interface available. For other cases, `values.TableObject`
is favored to avoid more contention on `flux`.

Each place where we make a type assertion, there's a link to #4343 which hopes to help track down all these checks for removal later. They should become redundant once we're typing the difference between array and table stream more strongly, preventing one of either type to "sneak through" in situations where it isn't appropriate.

New tests have been added for many, but not all, of the new branches. The omitted ones fall into a few categories:

- side effects that would otherwise look to contact some remote service
- private functions that can't be tested because cyclical imports would need to be introduced to do so
- I couldn't figure out how to build a `TableObject` suited for the situation (nil defaults technically leave these "invalid" for many purposes)

---

Earlier patches targeted specific callsites that appeared in logs:

- <https://github.com/influxdata/flux/pull/4272>
- <https://github.com/influxdata/flux/pull/4276>

### Done checklist
- [x] Test cases written

### DIFF
--- a/values/values.go
+++ b/values/values.go
@@ -20,6 +20,7 @@ type Typer interface {
 // TableObject serves as sort of a "marker trait" to allow us to check if a
 // value is a flux.TableObject without having to import the flux package which
 // in many cases will cause a cyclical import.
+// XXX: remove when array/stream are different types <https://github.com/influxdata/flux/issues/4343>
 type TableObject interface {
 	TableObject()
 }
@@ -188,7 +189,6 @@ func Unwrap(v Value) interface{} {
 		return v.Regexp()
 	case semantic.Array:
 		arr := v.Array()
-		// FIXME: needs a test
 		if _, ok := arr.(TableObject); ok {
 			panic(errors.New(codes.Invalid, "cannot unwrap a table stream"))
 		}
--- a/values/values.go
+++ b/values/values.go
@@ -17,6 +17,15 @@ type Typer interface {
 	Type() semantic.MonoType
 }
 
+// ITableObject serves as sort of a "marker trait" to allow us to check if a
+// value is a TableObject without having to import TableObject which would be a
+// cyclical import.
+// Identical purpose to the interface in the interpreter package, but sadly we
+// can't import it here because of yet another potential cycle.
+type ITableObject interface {
+	TableObject()
+}
+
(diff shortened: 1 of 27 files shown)


values/values.go
```diff
@@ -17,6 +17,15 @@ type Typer interface {
 	Type() semantic.MonoType
 }
 
+// ITableObject serves as sort of a "marker trait" to allow us to check if a
+// value is a TableObject without having to import TableObject which would be a
+// cyclical import.
+// Identical purpose to the interface in the interpreter package, but sadly we
+// can't import it here because of yet another potential cycle.
+type ITableObject interface {
```
jsternberg, line 25: Can we name this `TableObject`? I'm not a big fan of the `I` prefix for interfaces and that's not really used in Go.

Is it also possible to have the method be unexported? I don't know if that's possible. If it's not, this is good. If it is possible, I'd prefer this method to be unexported.

jsternberg (commented):
In general, I think this looks correct. One minor feedback for naming and API related stuff.

values/values.go
onelson: Happy to rename it. I was _sort of_ concerned that it'd be confusing to reuse the struct's name here, but I concede that the name I picked doesn't help clarify much beyond "this is different than `TableObject`".

I think I started out with the method being private but the implementation over on the struct didn't seem to satisfy the interface in that case. I'll try it again just to be sure.

jsternberg: > I think I started out with the method being private but the implementation over on the struct didn't seem to satisfy the interface in that case. I'll try it again just to be sure.

Ah I suspected that might be the case. If that is, then it's fine if the method is exported.

Marwes (approved):

PR #4336, opened 2021-12-09
feat(docs): add inline documentation to experimental packages

sanderson:

- experimental/http
- experimental/influxdb
- experimental/json
- experimental/mqtt
- experimental/oee

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -305,11 +305,6 @@ const EXCEPTIONS: &[&str] = &[
     "contrib/sranka",
     "contrib/sranka/teams",
     "contrib/sranka/webexteams",
-    "experimental/http",
-    "experimental/influxdb",
-    "experimental/json",
-    "experimental/mqtt",
-    "experimental/oee",
```
wolffcm, line 312: 🎉

stdlib/experimental/influxdb/influxdb.flux
```diff
@@ -1,8 +1,63 @@
...
+// - **body**: HTTP response body (bytes). 
+// 
+// ## Parameters
+// - method: HTTP request method.
+// - path: InfluxDB API path.
+// - host: InfluxDB host URL _(Required when executed outside of InfluxDB)_.
+//   Default is `""`.
+// - token: [InfluxDB API token](https://docs.influxdata.com/influxdb/cloud/security/tokens/)
+//   _(Required when executed outside of InfluxDB)_.
+//   Default is `""`.
```
wolffcm, line 25: Is this one not required all the time? I guess we don't require a token for running `to()` for example, so maybe not---just double checking.

stdlib/experimental/http/http.flux
```diff
@@ -1,6 +1,34 @@
...
+// ## Response format
+// `http.get()` returns a record with the following properties:
+// 
+// - **statusCode**: HTTP status code returned by the GET request (int).
+// - **body**: HTTP response body (bytes). 
+// - **headers**: HTTP response headers (record).
+// 
+// ## Parameters
+// - url: URL to send the GET request to.
+// - headers: Headers to include with the GET request.
```
wolffcm, line 21: Will it be clear to the reader that headers is an object with like `{Authorization: "foo", Connection: "keep-alive"}`? Might be good to include an example. (Similar comment for the other functions that have a parameter like this)

stdlib/experimental/influxdb/influxdb.flux
```diff
@@ -1,8 +1,63 @@
...
+// - headers: HTTP request headers.
```
wolffcm, line 26: I just noticed that `headers` is a `dict` here and a `record` above... I wish we were consistent.

stdlib/experimental/json/json.flux
```diff
@@ -1,7 +1,52 @@
...
+//                 a: jsonData.a,
+//                 b: jsonData.b,
+//                 c: jsonData.c,
+//             }
+//         },
+// >     )
+// ```
+// 
+// tags: type-conversions
+//
```
wolffcm, line 51: You can run this one, right?

stdlib/experimental/oee/oee.flux
```diff
@@ -1,17 +1,94 @@
...
-// computeAPQ computes availability, performance, quality and overall equipment effectiveness (oee).
-// productionEvents - a stream of start/stop events for the production process. Each row contains
-//   a _time and state that indicates start and stop events.
-// partEvents - a stream of part counts. Each row contains cumulative counts where column partCount
-//   represents total number of produced parts and badCount number of parts that did not meet quality standards.
-// runningState - production event or state value that corresponds to equipment running state
-// plannedTime - total time that equipment is expected to produce
-// idealCycleTime - theoretical minimum time to produce one part
+// computeAPQ computes availability, performance, and quality (APQ)
+// and overall equipment effectiveness (OEE) using two separate input streams—**production events**
```
wolffcm, line 12: I guess we can't use an em-dash in our docs, I wish we could.

wolffcm (approved):
This looks great.  I had some comments, but I'll leave any changes up to your discretion. Happy to have another look if you decide to make changes.

stdlib/experimental/http/http.flux
sanderson: Added headers to the example below.

stdlib/experimental/influxdb/influxdb.flux
sanderson: So I don't know exactly how this works, but I just tested it in Cloud, and host and token weren't required. 🤷🏻‍♂️ 

I assume it's similar to `to()` where InfluxDB actually uses a different implementation whenever it's executed within the context of InfluxDB.

stdlib/experimental/influxdb/influxdb.flux
sanderson: Yeah, the `experimental/http` package was added before support for dictionaries. This package was added after. I'm all for consistency as well.

stdlib/experimental/json/json.flux
sanderson: Yep, nice catch. I'll remove the `no_run` tag.

stdlib/experimental/oee/oee.flux
sanderson: Replace with a colon.

PR #4338, opened 2021-12-10
fix(stdlib): change the default epsilon for testing.diff to 1e-6

jsternberg:
The default epsilon was too precise for our test cases. Changing the
epsilon to `1e-6` from `1e-9` to allow for the tests to pass on both
amd64 and arm64.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
LGTM :shipit:

PR #4340, opened 2021-12-10
feat(docs): add inline documentation to experimental packages

sanderson:

- experimental/prometheus
- experimental/query
- experimental/record
- experimental/table
- experimental/usage

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/prometheus/prometheus.flux
```diff
@@ -1,3 +1,9 @@
+// Package prometheus provide tools for working with
```
pierwill, line 1:
```suggestion
// Package prometheus provides tools for working with
```

pierwill (approved):

PR #4341, opened 2021-12-10
feat: add timeable support to (add/sub)Duration

nathanielc:
Prior to this change users had to workaround the fact that a timeable
was difficult to manipulate. For example v.timeRangeStart from the UI.
See
https://chrispnerd.blogspot.com/2021/12/i-have-application-that-feeds-periodic.html
for a good explanation.

Now users can use addDuraiton and subDuration directly.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


Marwes (changes requested):
This seems a like an X/Y problem to me, the underlying issue is that one script is used in two different contexts and that script happens to use `subDuration`.

Having a function to explicitly normalize the input between these two environments or not trying to use one script for both may be a less impactful change (else we may need to have all `time` taking functions accept `Timeable`).

nathanielc:
@Marwes We do use this logic of a Timeable elsewhere namely in the `date` package and a few other places. 

The intent was to make its usage ubiquitous. I have updated this PR with changes to make the `date` package and this change use the same code. I also discovered a bug in `date`'s implementation so I have added test for those cases.

As for why we need to handle both, the UI has two modes of specifying a time, relative and absolute. In relative mode the v.timeRangeStart/Stop variables are durations and when in absolute mode it creates them as time literals. This use case is common elsewhere too for example the range function. Generally speaking being able to reference a specific time relative to `now` is very common. The intent of the Timeable constraint it to allow for either method of specifying a time but also making it clear the argument is used as an absolute time.

Marwes:
I guess it is too late now, but it may have been better to ro have resolved the duration to a time before it is injected as `v.timeRangeStart` or have it injected as something like `v.timeRangeStart = now() + v.timeRangeStartDuration` instead :shrug:

Marwes (approved):

PR #4353, opened 2021-12-14
refactor: Merge the builtin types into an enum

Marwes:
It is pretty common that builtin types are treated the same so by splitting them up into a separate enum we can reduce the amount of matching we need to do on `MonoType`.

Did this refactoring a bit ago but stumbled on the fact that it changed the serialization of `MonoType`. Worked around that with a bit of tedious/ugly code so maybe this is worth it as a trade off. Thoughts?


wolffcm (commented):
I think I like this better, but I get concerned when the Rust, Go and FlatBuffers representations of the same things diverge. In the few places where it's happened, it's confusing.

How much work would it be to extend this new to the flatbuffer representation as well?

Marwes:
The flatbuffer representation actually use this kind of representation already https://github.com/influxdata/flux/blob/43203154ad08f768cb5bb20f425865145f601ece/internal/fbsemantic/semantic.fbs#L30-L40 and it seems to be mirrored in go (though slightly different maybe, it is a bit confusing).

wolffcm (approved):
I like this, but maybe solicit for opinions on `#flux` and get other's thoughts. If there's no objections, go for it.

Marwes:
Changes for `flux-lsp` https://github.com/influxdata/flux-lsp/tree/builtin_type

PR #4354, opened 2021-12-14
fix(stdlib/universe): optimized union needs a mutex

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/union2.go
+++ b/stdlib/universe/union2.go
@@ -1,6 +1,8 @@
 package universe
 
 import (
+	"sync"
+
 	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
@@ -12,6 +14,7 @@ import (
 type unionTransformation2 struct {
 	d       *execute.TransportDataset
 	parents int
+	mu      sync.Mutex
 }
 
 func newUnionTransformation2(id execute.DatasetID, parents []execute.DatasetID, mem memory.Allocator) (execute.Transformation, execute.Dataset, error) {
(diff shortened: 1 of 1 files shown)


stdlib/universe/union2.go
```diff
@@ -31,6 +34,9 @@ func (u *unionTransformation2) ProcessMessage(m execute.Message) error {
 		return nil
 	}
 
+	u.mu.Lock()
```
nathanielc, line 37: Can you add a comment for why a mutex is needed?

nathanielc (commented):
Look good, can you add a test that uses Union concurrently?

stdlib/universe/union2.go
jsternberg: This was caught by `test-race` which is how I know about it. It doesn't seem to be the most consistent though. Either way, I think that does mean there's a test case for this.

rockstar (approved):
I have the same request as @nathanielc, but good to merge when that's done.

stdlib/universe/union2.go
nathanielc: To expand on this request a bit, I am concerned that having a mutex within the Process message code is an anti pattern and so I want to understand why its needed as I would expect it to be the exception rather than the rule.

jsternberg: Oh whoops. I was responding to your overall comment and it looks like I'm responding to the request for a comment. I'm adding a comment. I do think it would probably be good to fix this in the engine to remove these needed locks. I don't think there's a time when the lock is not needed for concurrent parents so we might just want to do it automatically.

jsternberg: I added the comment. Not sure how to add another test since there's already a test that uses union with two parents in a concurrent way.

nathanielc (approved):

stdlib/universe/union2.go
nathanielc: Sounds good, so generally the Process method doesn't need to lock but only because we have multiple predecessors in the case of union we need to. Makes sense

PR #4355, opened 2021-12-15
fix: Split the entire pipe chain on multiple if necessary

Marwes:
This wasn't noticed before as the code will keep any already multiline pipe chain on multiple lines, so this only occured if the entire chain was on a single line, but couldn't actually fit.

### Done checklist
- [x] Test cases written

### DIFF
--- a/libflux/flux-core/src/formatter/mod.rs
+++ b/libflux/flux-core/src/formatter/mod.rs
@@ -1399,21 +1399,10 @@ impl<'doc> Formatter<'doc> {
         loop {
             match (operators.next(), arguments.next()) {
                 (Some(operator), Some(arg)) => {
-                    doc = docs![
-                        arena,
-                        doc,
-                        docs![
-                            arena,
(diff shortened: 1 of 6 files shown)


nathanielc (approved):

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -1399,21 +1399,10 @@ impl<'doc> Formatter<'doc> {
...
-                                line.clone()
-                            } else {
-                                arena.nil()
-                            }
-                        ]
-                        .group(),
-                    ];
+                    doc += docs![arena, operator, arg].group();
+                    if arguments.peek().is_some() {
+                        doc += line.clone();
```
Marwes, line 1404: This line needed to be outside of the group, otherwise the `|> call()` bit otherwise the group specified here would make the line think it fit.

jsternberg (approved):

PR #4356, opened 2021-12-15
feat: Compiler supports vectorized functions

### LINKED ISSUE #4185: Update compiler to handle vectorized functions
In issue #4183 a `vectorized` field was added to semantic graph `FunctionExpr` nodes. And in issue #4184 a pass is added to analysis to vectorize simple functions.

In this issue, the work is to enhance the `compiler` package in Go (which evaluates the functions passed to `map` and `filter`) to be able to handle vectorized functions (that is, the `r` argument will have fields that are vector rather than scalar types).

The enhancement can be limited to just those functions that can be produced by #4184, that is, functions of the form:
```
(r) => ({r with a: r.a, b: r.b, ...})
```

The representation I used for input in the proof of concept was an object with array fields:
https://github.com/influxdata/flux/pull/3930/files#diff-ca52e5e5693fff638788b406cc69092210f978d482989588503142cf525b95f4R288
This uses the `ArrayValue` implementation of `values.Value` which is a simple wrapper around Arrow arrays.

Probably we will want to change this representation to be an instance of `Vector` instead of `Array`.

This issue explicitly does *not* depend on #4184, as it should be possible for these two issues to be worked in parallel.

DOD:
- Compiler package is enhanced to execute vectorized functions as described above
- Tests are added that demonstrate this support

scbrickley:
~Closes #4185~

~This patch adds support for vectorized functions to the compiler.~

~It includes the follow changes:~
~- Creates a new value type called `Vector`~
~- Adds vector-related utility functions to the `arrowutil` package~
~- Calls the vectorization pass (see #4184) inside the analyzer after type-substitution has finished~
~- Restricts vectorized functions to include only supported types~
~- Teaches the compiler how to recognize and handle vectorized function expressions~
~- Adds some tests to confirm compiler behavior around vectors.~

~Compatibility with the new vectorized map function (see #4329) should be confirmed before this merges.~

Closed in favor of #4435 & #4427


compiler/compiler.go
```diff
@@ -256,6 +262,8 @@ func substituteTypes(subst map[uint64]semantic.MonoType, inferredType, actualTyp
 			}
 		}
 		return nil
+	case semantic.Vec:
+		return nil
```
wolffcm, line 266: I think this will want to follow a similar example to `semantic.Arr` above.

compiler/compiler.go
```diff
@@ -383,6 +391,12 @@ func apply(sub map[uint64]semantic.MonoType, props []semantic.PropertyType, t se
 			return t
 		}
 		return semantic.NewFunctionType(apply(sub, nil, retn), args)
+	case semantic.Vec:
+		element, err := t.ElemType()
+		if err != nil {
+			return t
+		}
+		return semantic.NewVectorType(apply(sub, nil, element))
```
wolffcm, line 399: I'm a little unclear on what the `props` are in this function. I can see that the `Arr` case propagates them, so maybe `Vec` should too? I'd like to understand what this parameter is for. What it's doing doesn't seem quite right to me.

semantic/graph.go
```diff
@@ -564,6 +565,25 @@ func (e *DictExpression) TypeOf() MonoType {
...
+func (e *VectorExpression) Copy() Node {
+	if e == nil {
+		return e
+	}
+
+	ne := new(VectorExpression)
+	*ne = *e
+	return ne
+}
+
```
wolffcm, line 586: I don't think a vector expression will be needed for this work, even though there will be vector values.

Vectors are unlike arrays in this way. Users can create arrays with Flux syntax like `[1, 2, 3]` (that is an `ArrayExpression`) but users will never create literal vectors. Transformations like `map` will create records with vector fields, but that doesn't require vector expressions, since the fields can just be accesed via member expressions like `r.a` which will be typed as `vector of <sometype>`.

Does that make sense, or am I missing something?

wolffcm: (correspondingly, I don't think a `VectorEvaluator` is needed either, unless I'm missing something)

semantic/monotype.go
```diff
@@ -271,17 +276,39 @@ func getArr(tbl fbTabler) (*fbsemantic.Arr, error) {
...
-	arr, err := getArr(mt.tbl)
-	if err != nil {
-		return MonoType{}, err
+	// XXX: sean (16 Dec 2021) - This err1/err2 business seems sloppy. There's
+	// probably a better way to do this. This is functional for now though.
+	// Will revisit later.
+	arr, err1 := getArr(mt.tbl)
+	vec, err2 := getVec(mt.tbl)
+	if err1 != nil && err2 != nil {
+		return MonoType{}, errors.New(codes.Internal, "MonoType is not an array or a vector")
```
wolffcm, line 295: I am regretting making all of these functions return errors. I think they should just panic, if we try to get the elem type of an int or whatever, that's always an internal error. It would just simplify all this logic so much.

wolffcm (commented):
This is a great start---I would like to make sure we're on the same page as far as vector expressions and evaluators before you go much further.

compiler/compiler.go
```diff
@@ -15,6 +15,12 @@ func Compile(scope Scope, f *semantic.FunctionExpression, in semantic.MonoType)
 		return nil, errors.Newf(codes.Invalid, "function input must be an object @ %v", f.Location())
 	}
 
+	// If the function is vectorizable, `f.Vectorized` will be populated, and
+	// we should use the FunctionExpression it points to instead of `f`
+	if f.Vectorized != nil {
```
jsternberg, line 20: What mechanism will be exposed so the caller knows they're using the vectorized version?

compiler/vectorized_test.go
```diff
@@ -0,0 +1,125 @@
...
+			input: values.NewObjectWithValues(map[string]values.Value{
+				"r": values.NewObjectWithValues(map[string]values.Value{
+					"a": arrowutil.NewVectorFromElements(1.2),
+				}),
+			}),
+			want: values.NewObjectWithValues(map[string]values.Value{
+				"a": arrowutil.NewVectorFromElements(1.2),
+				"b": arrowutil.NewVectorFromElements(1.2),
+			}),
+		},
```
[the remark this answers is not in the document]
jsternberg: This is all correct and the first thing I was looking for on reviewing.

I think the easiest way to do this without creating a full refactor of the compiler is to add retain and release to the vector interface. I think then we do two things:

1. If we store the vector in the scope, we retain it.
2. If we retrieve the vector from the scope, we retain it.
3. When the scope goes out of scope, we iterate through the items at the current level of the scope and release them.
4. When a vector is used by any of the vector operations, it releases the vector.

I _think_ that should cover everything.

internal/arrowutil/vector_values.gen.go
```diff
@@ -0,0 +1,601 @@
...
+	"regexp"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+func NewVectorValue(arr array.Interface, typ flux.ColType) values.Vector {
```
jsternberg, line 20: I think this should be in the values package but I'm assuming there's a reason it's not?

internal/arrowutil/vector_values.gen.go
```diff
@@ -0,0 +1,601 @@
...
+
+func NewIntVectorValue(arr *array.Int) values.Vector {
+	return IntVectorValue{
+		arr: arr,
+		typ: semantic.NewVectorType(semantic.BasicInt),
+		et:  semantic.BasicInt,
+	}
+}
+
+func (v IntVectorValue) ElementType() semantic.MonoType { return v.et }
```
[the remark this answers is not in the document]
jsternberg: This seems like one where a pointer receiver would be better.

internal/arrowutil/vector_values.gen.go
```diff
@@ -0,0 +1,601 @@
...
+		panic(fmt.Errorf("unsupported column data type: %s", typ))
+	}
+}
+
+var _ values.Value = IntVectorValue{}
+var _ values.Vector = IntVectorValue{}
+
+type IntVectorValue struct {
+	arr *array.Int
+	typ semantic.MonoType
```
jsternberg, line 132: Are these two attributes needed? It seems you can get `et` from `typ`.

values/vector.go
```diff
@@ -0,0 +1,13 @@
+package values
+
+import (
+	"github.com/influxdata/flux/semantic"
+)
+
+type Vector interface {
+	Value
+	ElementType() semantic.MonoType
+	Get(i int) Value
```
jsternberg, line 10: I think you can just remove most of these attributes outside of `ElementType()`. Instead, this should expose the arrow array. We're not going to use these methods anyway.

jsternberg (changes requested):

values/vector.go
scbrickley: > I think you can just remove most of these attributes outside of `ElementType()`.

Including the `Value` mixin?

jsternberg: The `Value` mixin is needed for the values interface so keep that. But, we should expose the array as much as possible since that's what we'll actually use. Access methods outside of that won't be helpful.

compiler/vectorized_test.go
```diff
@@ -24,50 +40,56 @@ func TestVectorizedFns(t *testing.T) {
...
+		input        map[string]interface{}
+		want         map[string]interface{}
 		skipComp     bool
 	}{
 		{
 			name:         "field access",
 			fn:           `(r) => ({c: r.a, d: r.b})`,
 			vectorizable: true,
+			allocated:    256, // This should be 128 (2 vectorized fields, each containing 1 int64)
+			maxAllocated: 448,
```
wolffcm, line 55: > This should be 128

I don't get what you're saying with this comment.  Can you explain? 

The numbers here are going to be dependent on the internals of Arrow arrays, so I'm not sure we want to get specific about byte counts here.

Can we write meaningful tests if we just:
- Allocate memory for the input
- Store the current number of bytes allocated
- Run the function, and collect the output
- Verify that the new current bytes allocated is what we expect; either the same as before, or possibly more if we have created new fields on the output record. Which one it is will depend on the test case I guess. For the time being, it should always be the same. Once we can handle things like `(r) => (r with a: r.a, b: r.b, ab: r.a + r.b)}` it will be more.

I would really like to be able to capture the reference counts on the arrays themselves.

@jsternberg any thoughts on this?

jsternberg: I think probably just running it and making sure the numbers make sense or just using them as-is is probably good enough. Some of these numbers can get weird because arrow allocates a minimum amount regardless of how big the array is. I forget the minimum though.

wolffcm: That makes sense. In light of this I think this is good as-is. We can always tweak these tests later.

wolffcm (approved):

PR #4357, opened 2021-12-16
refactor: Remove redundant get_err* functions

Marwes:
`ast::check::check` does the same thing but returns an `Error` type instead of a plain `String`


skartikey (approved):
Looks good to me!

Issue #4073, opened 2021-09-23
Proposal: Support Flow sensitive typing with the `exists` operator in conditional statements


nathanielc:
A common pattern is to try to use `if exists r.a then r.a else 0` to provide a default value when the property `a` is not present on the record `r`. However current type inference rules make this pattern impossible. In short the `exists r.a` expression adds a type constraint that the record `r` must have the property `a`. This defeats the purpose in many cases of using the `exists` operator.  The `exists` operator does work this way when the record is unbounded (i.e. a record from InfluxDB or csv) but not when the record is bounded. 

In the case of bounded records we would like to still be able to use the `exists` operator to provide a default. 

Take the example of the pagerDuty endpoint function, we would like to define the function like this:

```
endpoint = (url=defaultURL) => (mapFn) => (tables=<-) => tables
    |> dedupKey()
    |> map(
        fn: (r) => {
            obj = mapFn(r: r)

            return {r with
                _sent: string(
                    v: 2 == sendEvent(
                        pagerdutyURL: url,
                        // other properties hidden for simplicity
                        // ...
                        // Here since the  customDetails argument to the sendEvent function has a default value we may or may not have that data on the obj record
                        // We would like to provide a default when it doesn't exist
                        customDetails: if exists obj.customDetails then obj.customDetails else {},
                    ) / 100,
                ),
            }
        },
    )
```

If you use this pattern however type inference will state the `obj` must have the customDetails property. 

The proposal is that instead we do compile time type checking and cull the conditional expression and then finalize the type checking.

I am imagining a system like this:

For call sites for the `pagerduty.endpoint` function we determine the type of `obj`. Then statically evaluate the conditional expression. In cases where `obj` contains the `customDetails` property the code is rewritten to simply be `customDetails: obj.customDetails`. In cases where `obj` does not contain `customDetails` (not obj is bounded so we can concretely state that is does not have the property), we rewrite the code as `customDetails: {}`.

Then with this new code we finish type inference and see that it type checks correctly because when `obj` does not have the property all expressions that accessed the property were removed.

How could this work?

[...]

PR #4358, opened 2021-12-16
feat: Add optional types

Marwes:
Continued prototype of https://github.com/influxdata/flux/issues/4073#issuecomment-951892648 to try and visualize the scope.

This adds the optional type to the typesystem (`A?`, `{ A with b: B? }`), primarily to represent fields that may or may not exist. Instead of representing this directly on the record fields I represent it on `MonoType` which will complicate things a bit now, but should avoid the complexity of a potential `{ a with a?: int? }` if optional types were added later.

* We should probably prevent `A??` etc from being valid.
* Unification will need to handle subsumption (subtyping) which is added in a bit of a hacky way currently. But the short of it is that `unify` only checks that two types are equal, but under optional types (fields) we must allow `int` to be passed in place of `int?`, but not the other way around.
* I believe nulls may be generated from sources today(?). If we know whether a column of a source is strictly null/non-null, perhaps this can be used to be more precise about that. It runs the risk of increased verbosity though. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/libflux/flux-core/src/ast/mod.rs
+++ b/libflux/flux-core/src/ast/mod.rs
@@ -610,6 +610,8 @@ pub enum MonoType {
     Record(RecordType),
     #[serde(rename = "FunctionType")]
     Function(Box<FunctionType>),
+    #[serde(rename = "OptionalType")]
+    Optional(Box<OptionalType>),
 }
 
 impl MonoType {
@@ -623,6 +625,7 @@ impl MonoType {
             MonoType::Dict(t) => &t.base,
             MonoType::Record(t) => &t.base,
             MonoType::Function(t) => &t.base,
+            MonoType::Optional(t) => &t.base,
         }
     }
 }
@@ -686,6 +689,16 @@ pub struct FunctionType {
     #[serde(flatten)]
     pub base: BaseNode,
     pub parameters: Vec<ParameterType>,
+    pub monotype: MonoType,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[allow(missing_docs)]
+pub struct OptionalType {
+    #[serde(skip_serializing_if = "BaseNode::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub base: BaseNode,
     pub monotype: MonoType,
 }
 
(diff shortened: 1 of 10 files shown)


nathanielc:
I am not clear on the semantic of an optional type. How does an optional type as a property in a field compare to a record extension? For example the type `{A with a: int}` can describe both of these records: `{a: 1, b: "s"}` and `{a: 2}`. So it appears to a user that `b` is already optional. In other words the way I understand our record extension type is that we define the minimum aka required properties of the record. And the maximum bound of properties is infinite meaning that any other optional properties may exist. 

Asking my question another way: how is `{A with a: int, b: B?}` different from `{A with a: int}`? I would expect both records `{a:1, b: "s"}` and `{a: 2}` to match both types.

Marwes:
The important difference here is that the type signature of ` {A with a: int, b: B?}` limits which which type `b` may have. This is clearest in the type signature of `record.get` https://github.com/influxdata/flux/issues/4073#issuecomment-951892648

```flux
// f: (r: { A with a?: B }, default :B) => int
f = (r, default)  => if exists r.a then r.a else default
// Works
f(r: { a: 1 }, 0)
f(r: { b: "asd" }, 0)
// Error: optional field is of the wrong type
f(r: { a: "asd" }, 0)
```

If `f/record.get` were typed as `get : (r: { A with }, default: B) => A` the type (as seen by the caller) has no way to restrict that the field `a` must be `B` which would let the last case `f(r: { a: "asd" }, 0)` let you transmute the string "asd" into an integer (`B`)

Marwes:
An alternative to optional types/fields would be that `exists` checks also check that the type is correct at runtime for the branch that it is supposed to be used in.

Marwes:
Poked a bit further on this to find another issue. Index expressions (`a[x]`) expect `a` to be an array and `x` to be an integer, it is only in the case where an explicit string literal is used that index expressions work for records (`a["abc"]`) as it is handled on the parser level.

```
get : (r: { A with K: B }, key: K, default: B) => B
get = (r,key,default) => if exists r[key] then r[key] else default
```

To implement the proposed `get` function we would need index expressions to work for records as well so I think we multi parameter kinds/associated types etc. Otherwise the `r[key]` expression can't know which types to infer for `r` and `key`. If we had something like the `Index A` kind we could define.

```
Record: Index String
Array: Index Int
```

If `r` has the type `R` and `key` the type `K` then inference for index expressions would then constrain `R: Index K` which would let it work without prior knowledge.

PR #4360, opened 2021-12-16
feat(docs): add inline documentation to interpolate, json, kafka, and math packages

sanderson:

- interpolate
- json
- kafka
- math

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/math/math.flux
```diff
@@ -51,154 +75,230 @@ builtin maxuint : uint
 // >     |> map(fn: (r) => ({r with _value: math.abs(x: r._value)}))
 // ```
 //
+// ## Special cases
+//
+// ```no_run
+// math.abs(x: ±Inf) // Returns +Inf
```
wolffcm, line 81: The `±` is not valid Flux. I know there were examples like this in Flux comments before. I think it might be better to explicitly mark the cases for both `+Inf` and `-Inf` if needed for these examples and the others, sadly. Thoughts?

wolffcm: Or I guess your treating the special cases not as literal Flux code? Looking over some of the other special cases (like where the special case is just one parameter and the other parameter can be anything) I can go either way now, I'll leave it up to your discretion.

stdlib/math/math.flux
```diff
@@ -217,837 +317,1243 @@ builtin atan : (x: float) => float
...
+// import "sampledata"
+//
+// < sampledata.float()
+// >     |> map(fn: (r) => ({r with _value: math.cbrt(x: r._value)}))
+// ```
+//
+// ## Special cases
 //
+// ```no_run
 // math.cbrt(±0)   // Returns ±0
```
wolffcm, line 390: Wow, acknowledging that there is both + and - zero in special cases. Nice to see this behavior described so thoroughly!

wolffcm (approved):
This looks great. I had some comments. Nothing to block you from merging.

stdlib/math/math.flux
sanderson: Yeah, I had this same thought. These special cases are actually pulled from the Go math package documentation. Some of the other special cases would be hard to accurately represent with fully valid Flux. For example,`math.acos(x: <-1)`. Just including a value that is less than -1 doesn't accurately reflect the scope of the special case.

For what it's worth, these special cases have been in the existing math package docs since it was released and I've never heard any complaints about them not working. Doesn't mean that users haven't tried, but I haven't heard anything.

stdlib/math/math.flux
```diff
@@ -51,154 +75,230 @@ builtin maxuint : uint
...
+// ```no_run
 // import "math"
 //
 // math.asin(x: 0.22)
 // ```
-//
-// ## Special Cases
-//
+// 
+// ### Use math.asin in map
```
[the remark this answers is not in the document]
sanderson: We do get a lot of comments about how to use functions (specifically `math` functions) in `map()`. That's why I included them in this package. You're right, it is a lot of boilerplate, but I don't think we'll do it for all functions that return scalar values. It'll likely be the `math` package and the `strings` package. Those are the two most common packages brought up when users are wondering how to use them with `map`.

PR #4363, opened 2021-12-17
feat: add an internal/debug.opaque function

wolffcm:
This adds a new internal debug function that will let us write more realistic tests:
```
import "array'
import "internal/debug"
tableDataF3 = array.from(rows: [
    {_measurement: "m", _field: "f3", _value: 10.0, _time: 2021-12-16T00:00:00Z},   
  ]) |> group(columns: ["_measurement", "_field"]) |> debug.opaque()

tableDataF4 = array.from(rows: [
    {_measurement: "m", _field: "f4", _value: true, _time: 2021-12-16T00:00:00Z},   
  ]) |> group(columns: ["_measurement", "_field"]) |> debug.opaque()

tableData = union(tables: [tableDataF3, tableDataF4])
```
The columns `_value` having different types in each branch of the union is intentional. Without the call to `opaque`, type analysis will complain about the differing types.


onelson (approved):
I surely can't be the final word here, but LGTM :shipit:

PR #4366, opened 2021-12-18
test: add a table stream test for is_type, to test push downs

wolffcm:
These new tests will be extended to test out pushing down type predicates into the storage tier.

Part of #2159.


skartikey (approved):
Looks good!

PR #4373, opened 2021-12-21
feat(stdlib): update the to transformation to use narrow transformation

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/influxdata/influxdb/to.go
+++ b/stdlib/influxdata/influxdb/to.go
@@ -5,11 +5,16 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/apache/arrow/go/arrow/bitutil"
+	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/compiler"
 	"github.com/influxdata/flux/dependencies/influxdb"
 	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/internal/arrowutil"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/plan"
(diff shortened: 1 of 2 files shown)


stdlib/influxdata/influxdb/to_test.go
```diff
@@ -692,45 +692,6 @@ func TestTo_Process(t *testing.T) {
...
-						{Label: "_field", Type: flux.TString},
-						{Label: "_value", Type: flux.TFloat},
-					},
-					Data: [][]interface{}{
-						{execute.Time(11), "m", "f", 1.0},
-					},
-				},
-			},
-			wantErr: errors.New("to() found duplicate table with group key: {_field=f,_measurement=m}"),
-		},
```
wolffcm, line 733: We no longer get this error I guess due to the chunked implementation?

stdlib/influxdata/influxdb/to.go
```diff
@@ -148,48 +130,37 @@ func (t *ToTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
...
-}
+	if err := t.writeTable(chunk); err != nil {
+		return err
+	}
 
-// UpdateProcessingTime updates the processing time for the transformation for the `to` flux function.
-func (t *ToTransformation) UpdateProcessingTime(id execute.DatasetID, pt execute.Time) error {
-	return t.d.UpdateProcessingTime(pt)
+	filtered := t.filterNulls(chunk, mem)
+	return d.Process(filtered)
```
wolffcm, line 141: Could you add some comments here about the null filtering? My understanding of what you're doing here is:
- The call to `writeTable` above will skip over any points that have a null timestamp
- Because the semantics of `to()` is that we output the tables that are written, we then need to filter rows with null timestamps from the input.

Are there tests for the null filtering?

wolffcm (changes requested):
I am asking for some comments, but this looks good otherwise.

stdlib/influxdata/influxdb/to.go
jsternberg: Yes, there's a test for this. I didn't know about this behavior before I saw the failed test.

stdlib/influxdata/influxdb/to_test.go
jsternberg: We never got this error because it wasn't strictly possible, but now we don't really have a way to check for it because the chunked implementation expects us to receive table chunks with duplicate group keys.

stdlib/influxdata/influxdb/to.go
jsternberg: I'll also add a comment.

jsternberg: I updated the doc string for `filterNulls` to explain what is going on.

wolffcm (approved):

PR #4374, opened 2021-12-21
feat(docs): add inline documentation to regexp, runtime, sampledata, slack, and system packages

sanderson:

- regexp
- runtime
- sampledata
- slack
- system

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
LGTM :ship:

PR #4385, opened 2022-01-04
perf: Make SourceLocation cheaper to clone

Marwes:
## perf: Remove source for SourceLocation

The `source` field is barely used throughout the compiler which makes it a rather expensive thing to keep around when the equivalent functionality can be provided for the cases that needs by passing the original source string through as well and using `Locator` to get the source segment.

## perf: Use a shared reference for the filename in SourceLocation

Each node contains a source location so by sharing it we can save a good few allocations.

```
infer/bootstrap         time:   [59.086 ms 59.263 ms 59.476 ms]
                        change: [-21.299% -15.826% -12.469%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high mild
```


Marwes:
Can you review when you get a chance? @jsternberg

jsternberg:
I think this results in a change to the json and I think that causes the formatter to potentially become lossy as I think some elements are formatted using the source attribute (unless that was changed). I really agree with this change but I'm not sure we can without making a breaking change to the JSON API.

For that reason, I think we should discuss this at a meeting before merging something like this.

Marwes:
> The JSON API is used by the client endpoint. They can submit JSON directly instead of a script. I think there's also a way to retrieve the JSON.

:/ Can I at least hope that part of the API isn't publicly exposed? (ie we can deprecate and remove it) What is the value of submitting a JSON blob instead of just the raw flux query string considering the JSON blob is much, much larger than any query string.

Marwes:
It wouldn't be that hard to restore the source field just for the JSON at least, and if we get JSON submtited with the source field we would just ignore it.

Issue #2452, opened 2020-02-04
Iteration over values of time series with flux


sgerhardt1:
It would be very helpful for outlier detection to iterate over all values of a time series with access to previous values. So that you can do something like this:

`for(i=0; i<values.size; i++) {
    if (i > 1 && value[i-2] > value[i-1]*10 && abs(value[i]-value[i-2]) < 5) {
        listOfOutliers.add(value[i-1]);
    } 
}`

There should be an element like a for-loop that you can use to iterate over values.

nathanielc:
@sgerhardt1 Yes we agree, we have some longer term plans to add loops to the language. In the meantime we have a `reduce` function that will let you do something similar. See https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/reduce/

sanderson:
The issue with `reduce` is that it doesn't preserve the previous rows. It outputs only one row per table. `map` iterates over each row and preserves all rows.

## Proposal
What if we extended `map` so you could pass additional row objects into each row loop? For example, a `includePreviousRows` parameter that would accept an integer representing the number of previous rows to include in the mapping operation. You could then reference previous row objects using something like `p1.columnName` where `p1` is essentially `r-1`.

The example above in Flux would look something like this:

```js
import "math"
// ...
  |> map(
    includePreviousRows: 2,
    fn: (r, p1, p2) => ({ r with 
      outlier: if r._value > 1.0 and p2._value > (p1._value * 10.0) and math.abs(x: p2._value) < 5.0 then true else false
    }))
  |> filter(fn: (r) => r.outlier == true)
```

Some other notes:

- If `includePreviousRows=2`, each output table will have 2 less rows than their corresponding input table because you can't complete the loop with null values. The first `x` rows would be dropped after the transformation.
- I know this would increase the resource demands of `map`, but I don't know by how much. It would mainly depend on how many rows the user includes and the size of those rows.

sgerhardt1:
Great, that sounds very promising. I think this feature would be very helpful for many situations.

Perhaps it would be a good idea to also be able to access the loop iteration variable (like the `i` in a for-loop). In the example above, the first two values must always end up in the result, because no outlier detection can be carried out for the first two values. And the aggregation of all filtered values after that have to result in a correct value.

Thank you for your thoughts on this.

geropl:
[...]

PR #4387, opened 2022-01-04
[WIP]: first prototype of cumulative reduce. functionality is there, but no …

aanthony1243:
…params to change back to regular mode

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


sanderson:
To toggle this added functionality and output all rows from the reduce operation, we should add a new boolean parameter `emitAll` (or `outputAll` _(never a new feature without some bike-shedding 😄)_).

With `emitAll: true`, output tables include rows for each corresponding input row. The default value is `false` to preserve backwards compatibility with the existing `reduce` behavior.

aanthony1243:
I don't have a strong opinion on the parameter name, i'm not sure if there's precedent in another function that has a similar option, @sanderson maybe you have some idea if that's possibly true?

aanthony1243:
pushed a new version, this is the current behavior: 

```
import "sampledata"

stateChangeTest = (tables=<-, c) => tables
  |> reduce(identity: {prev: 0, current: 0, different: false}, fn: (r, accumulator) => ({ prev: accumulator.current, current: r._value, different: accumulator.current != r._value }), cumulative:c)

sampledata.int() |> stateChangeTest(c: false) |> yield(name: "legacy")
sampledata.int() |> stateChangeTest(c: true) |> yield(name: "cumulative")

----------------------------------
Result: legacy
Table: keys: [tag]
            tag:string                 current:int  different:bool                    prev:int
----------------------  --------------------------  --------------  --------------------------
                    t1                           4            true                          15
Table: keys: [tag]
            tag:string                 current:int  different:bool                    prev:int
----------------------  --------------------------  --------------  --------------------------
                    t2                           1            true                          13
Result: cumulative
Table: keys: [tag]
            tag:string                 current:int  different:bool                    prev:int
----------------------  --------------------------  --------------  --------------------------
                    t1                          -2            true                           0
                    t1                          10            true                          -2
                    t1                           7            true                          10
                    t1                          17            true                           7
                    t1                          15            true                          17
                    t1                           4            true                          15
Table: keys: [tag]
            tag:string                 current:int  different:bool                    prev:int
----------------------  --------------------------  --------------  --------------------------
                    t2                          19            true                           0
                    t2                           4            true                          19
                    t2                          -3            true                           4
                    t2                          19            true                          -3
                    t2                          13            true                          19
                    t2                           1            true                          13
```

sanderson:
@aanthony1243 I don't have a strong opinion either about the parameter name, although I don't know that `cumulative` is the right one. Cumulative calculations are a use case, but I don't think they are _the_ use case. I think `emitAll` or `emitEach` would work.

aanthony1243:
I will go with emitAll

lazarenkovegor:
really looking forward to this feature. Can you tell me if it is possible to collect this feature before it appears in the release? Or is there an instruction on how to build and test your own function in golang? I found only instructions on where to push my functions, but I did not find instructions on how to raise the test environment locally.

aanthony1243:
@lazarenkovegor it's possible to build+test this branch, instructions are in the README.  After you can build flux standalone, linking it into and building influxdb is a straightforward redirect of the go.mod dependency to the local flux directory.

geropl:
I happened to find this PR from this discussion: https://github.com/influxdata/flux/issues/2452#issuecomment-1007382967

Really looking forward for this to get merged, thx for working on it! :pray: 

FYI here's a slightly different (more bulky?) approach to the problem: https://github.com/influxdata/flux/pull/4395
Sharing here in case it might help learn about the design space. :+1:

stdlib/universe/reduce.go
```diff
@@ -151,60 +162,112 @@ func (t *reduceTransformation) Process(id execute.DatasetID, tbl flux.Table) err
 				return errors.Wrap(err, codes.Inherit, "failed to evaluate reduce function")
 			}
 			params[accumulatorParamName] = m
+
+			key := t.computeGroupKey(tbl.Key(), m)
```
Marwes, line 166: This seems entirely duplicated from the same code below. Should factor it out to a function if it does not make sense to keep it in the loop (as the added comment mentions).

Marwes (changes requested):

PR #4388, opened 2022-01-05
feat: Add label polymorphism 

Marwes:
POC to introduce "label polymorphism" which will allow type signatures to more precisely specify how they transform records, thereby being able to statically determine more errors. This works by letting string literals get the new `label("literal")`  type instead of just `string` which in conjunction with record fields being capable of being type variables lets functions accept a "label" as a parameter which determines which fields that it runs on.

- [x] ~Allowing multiple labels to be passed in an array? (`columns: [L]`) Probably needs to treat arrays as heterogeneous lists of some sort so they are typed like `["a", "b"]` instead of `[string]`.~  Should be a separate issue, implemented later if desired.
- [x] ~On the other hand functions still want to accept a plain dynamic `string` (as they do today). So unification needs to cope with that still. We may need "dynamic" labels for this `{ A with 'dynamic: int, 'dynamic: B }` where a record  with one (or more!) dynamic labels would accept any field being accessed.~ We are moving ahead with just statically defined labels, should be rare enough with dynamic ones that we can accept any breakages (though we do need to verify that).
- [x] Some better syntax for specifying/printing labels so it is clear when they differ. Currently the same single, uppercase leter syntax for variables are use like `{ A: int }`, but that makes it impossible to define a record type with the field `A`. Add a `"A"` syntax perhaps (same as the record creation/indexing syntax)? See #4664
- [x] ~Type variables are currently re-numbered during the `convert` pass so you can get confusing errors where it says that the field `C` does not exist, except the field were actually specified with `B` (as seen below). (This happens at the master branch today as well)~ Should be another issue, may not bother with this
- [x] The delayed unfication that is currently there is hacky. It is needed to ensure that we see which label `B` has before we do the record unfication. A better way may be to add a constraint on `B` like `B must be one of the fields in record X` which can be solved later.
- [x] Feature flag so we can release and compile code without labels.
- [x] Need more tests, or confidence that current tests in `labels.rs` have adequate coverage

Based on #4664


libflux/Polymorphic_Labels.md
```diff
@@ -0,0 +1,103 @@
...
+
+```flux
+// Unifying know fields works regardless of the order that they are defined in
+{ a: int, b: string } <=> { b: string, a: int }
+
+// Should 'column unify against `a` or `b` field on the left side?
+{ 'column: A } <=> { b: string, a: int }
+```
+
+There may be a consistent way to unify these records in the face of type variables, however an easy workaround would be to delay the unification of records with unknown fields until they have been resolved, at which point they can unify normally. If there is a field that is still unknown when type checking is done we can designate that as an error.
```
nathanielc, line 95: To me delaying unification until records with unknown fields have been resolved makes the most sense. Its how I reason about this logic in the first place. Are there any drawbacks to delaying? Multiple passes over the semantic graph?

libflux/Polymorphic_Labels.md
```diff
@@ -0,0 +1,103 @@
...
+// and then fail to unify `Label("a") <=> Label("b")`. If we keep treating `Label(..)` as a string except in cases where it unifies to a variable that actually has the `Label` type.
+add(x: "a", y: "b")
+
+builtin func : (opt: string) => int
+
+// This will work since `Label("a")` is a sub type of `string`
+func(opt: "a")
+
+// Possible extension where we only allow some specific labels to be passed in
+builtin func : (opt: "option1" | "option2") => int
```
nathanielc, line 74: Is the idea here to make a sort of enum type? For things like the `method` parameter to the `quantile` function?

Why would we use labels for this other than its convenient? Meaning I don't see a use case yet of when a label used in the context of a column name would be constrained to a specific set.

libflux/Polymorphic_Labels.md
```diff
@@ -0,0 +1,103 @@
...
+The branches will have the `Label("a")` and `Label("b")` respectively, which we still need to unify successfully and the resulting type of the expression should still be a `string`. As long as we limit  the uses of label types to direct uses in calls (like in `fill` above). We may keep the same type checking behaviour by treating `Label` as a `String` types in every instance except when type checking function calls, in which case we specialize unification to a subsumption check where we allow some limited sub-typing.
+
+```flux
+// The original fill call is allowed such that `'column` is inferred to be `Label("a")`
+fill(column: "a", value: 0)
+
+// We should (probably) still allow dynamic strings for backwards compatibility sake so the `string` type
+// also implements the `Label` Kind
+c = "a" + ""
+fill(column: c, value: 0)
```
nathanielc, line 60: I wonder how often a truly dynamic column name is used? Maybe we can get some of that from the query archive?


Also what do we consider dynamic? Anything that requires evalauting the semantic graph? What about this case:

```
c = "foo"
fill(column: c, value: 0)
```


Is that considered dynamic or static? Seems like it could be considered static. I would also expect this case to be very common as a user might have to use the column name in several different functions and so would like to put it in a variable to make changing it easier.

Marwes: Yes, that could be considered static in this context (probably harder to force it to be "dynamic" than not).

libflux/Polymorphic_Labels.md
Marwes: It adds some complexity, and could possibly give worse error messages (though as long as we take care to store all the context we need for the errors, it shouldn't be an issue).

There may be some performance implications as well as we need to know/decide when to evaluate these delayed constraints. Constraints could (in theory at least) depend on each other in complex ways which we would need to untangle to resolve them all

A bit of a tangent, but Rust's typechecker has an largely equivalent system of "obligations" which it brute forces when solving them, I tried to improve that with https://github.com/rust-lang/rust/pull/69218 which would make them know more precisely when they would be solveable, but the overhead was to much).

Marwes:
For just about all functions defined with polymorphic labels in https://github.com/influxdata/flux/issues/4410 I can think of a way to redefine them to not require polymorphic labels which may be an argument for not doing this.

```flux
builtin columns : (<-tables: [A], ?column: 'label) => [{ 'label: string }] where A: Record
// into, if we want a different name on the output that is a `map(fn: fn(r) => { mycolumn: r.value })`
builtin columns : (<-tables: [A]) => [{ value: string }] where A: Record

builtin count : (<-tables: [A], ?column: string) => [{ _value: int }] where A: Record
// A `map(fn: fn(r) => r.mycolumn)` before `count` would let us pick the column
builtin count : (<-tables: [A]) => [{ _value: int }]

builtin distinct : (<-tables: [{ A with 'column: B }], ?column: 'column) => [{ _value: B }] where A: Record, B: Equatable
// into
builtin distinct : (<-tables: [A]) => [B] where A: Equatable

// Can't write out a type without polymorphic labels as `A` is preserved in the output
// but users can write it manually as `map(fn: fn(r) => { r with as: r.column }` (vs `duplicate(column: "column", as: "as")`
builtin duplicate : (<-tables: [{ A with 'column: B }], column: 'column, as: 'as) => [{ A with 'column: B, 'as: B }]
    where A: Record

// Same as `columns` and `count`. `map` before and after can "rename" the columns to and from the format we want
builtin elapsed : (<-tables: [{ A with 'time: time }], ?unit: duration, ?timeColumn: 'time, ?columnName: 'column) => [{ A with 'time: time, 'column: duration }]
    where
    A: Record
```

I didn't go through all of the functions thoroughly but other than issues implementing these changed functions it seems pretty plausible to "only" provide functions that operate on streams of plain values instead of records (or records with specific fields and use `map` to go between).

Lists of polymorphic labels (which is not part of this PR) seem more complicated but at least some operations could be defined.

```
// map(fn: fn(r) => { column_1: r.column_1, column_2: r.column_2 } // etc 
builtin keep : (<-tables: [{ A with ['column]: C }], ?columns: ['column], ?fn: (column: string) => bool) => [B] where A: Record, B: Record

// Can't be done
builtin drop : (<-tables: [{ A with ['column]: B }], ?fn: (column: string) => bool, ?columns: ['column]) => [{ C with  }]


builtin covariance : (<-tables: [{ A with ['column]: B }], ?pearsonr: bool, ?valueDst: 'dst, columns: ['column]) => [{ A with 'dst: B }]
    where
    A: Record
// Specify two specific fields in the signature and `map` to them
builtin covariance : (<-tables: [{ A with column1: B, column2: B }], ?pearsonr: bool) => [{ A with dst: B }]
    where
    A: Record

// Can't be translated
builtin sort : (<-tables: [{ A with ['column]: B }], ?columns: ['column], ?desc: bool) => [{ A with ['column]: B }]
    where A: Record,
          B: Comparable
```

wolffcm:
@Marwes 

In general I like the idea of using smaller composable operations, and especially `map` since we are making it more performant now. I have some concerns though.

For this example with aggregate functions:
```
builtin count : (<-tables: [A], ?column: string) => [{ _value: int }] where A: Record
// A `map(fn: fn(r) => r.mycolumn)` before `count` would let us pick the column
builtin count : (<-tables: [A]) => [{ _value: int }]
```
How does this work for a typical example where there are group key columns? So say something like
``` 
import "array"
array.from(rows: [
{host: "A", region: "east", mem: 100},
{host: "B", region: "east", mem: 100},
{host: "C", region: "west", mem: 100},
{host: "D", region: "west", mem: 100},
])
  |> group(columns: ["region"])
  |> sum()
```
For this example I want to get the sum of memory used by region. How do we preserve the separation between regions if aggregates always produce `[{_value: int}]`?

A broader question, are you thinking that users will not have to change their scripts in order to get the same behavior as before? Or are you assuming that some changes will be required of users?

Marwes:
> For this example I want to get the sum of memory used by region. How do we preserve the separation between regions if aggregates always produce [{_value: int}]?

True, aggregators need to take groups into account, and the group concept is sort of shared between records and streams. The stream tracks each group, but to check the group key of a record that is accessed through the record. So returning or accepting streams of raw values is probably not that useful.

I suppose there could be some way to try and separate group keys from records, but that likely strays to far towards breaking changes and it may just end replicating what records already do. So I think we need to stick with streams of records for that reason which in turn forces aggregate functions (and other transformations) to accept which column(s) to work with as a parameter.

> A broader question, are you thinking that users will not have to change their scripts in order to get the same behavior as before? Or are you assuming that some changes will be required of users?

No, users would be able to keep everything as is, but when working with pivoted data they would need to add `map` to select the column to work with. However given that aggregators expect groups to be preserved we would need to do something like `map(fn: fn(r) => { r with value: r.mycolumn })` to preserve the group key (I think) which doesn't seem great.

Marwes:
https://github.com/influxdata/flux/pull/4664 should be reviewed and merged before this

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+//
+package universe
+
+
+// columns returns the column labels in each input table.
+//
+// For each input table, `columns` outputs a table with the same group key
```
sanderson, line 12:
```suggestion
// columns returns the column labels in each input table.
//
// **Note:** `universe.columns()` is an experimental function with a more precise type signature.
//
// For each input table, `columns` outputs a table with the same group key
```

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+// tags: transformations
+//
+builtin columns : (<-tables: stream[A], ?column: C) => stream[{C: string}] where A: Record, C: Label
+
+// fill replaces all null values in input tables with a non-null value.
+//
+// Output tables are the same as the input tables with all null values replaced
```
sanderson, line 42:
```suggestion
// fill replaces all null values in input tables with a non-null value.
//
// **Note:** `universe.fill()` is an experimental function with a more precise type signature.
//
// Output tables are the same as the input tables with all null values replaced
```

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Examples
+//
+// ### List all columns per input table
+// ```
+// import "sampledata"
+//
+// sampledata.string()
+//     |> columns(column: "labels")
+// ```
```
sanderson, line 32:
~~~suggestion
// ```
// import "experimental/universe"
// import "sampledata"
//
// sampledata.string()
//     |> universe.columns(column: "labels")
// ```
~~~

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Examples
+//
+// ### Fill null values with a specified non-null value
+// ```
+// import "sampledata"
+//
+// < sampledata.int(includeNull: true)
+// >     |> fill(value: 0)
+// ```
```
sanderson, line 64:
~~~suggestion
// ```
// import "experimental/universe"
// import "sampledata"
//
// < sampledata.int(includeNull: true)
// >     |> universe.fill(value: 0)
// ```
~~~

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ### Fill null values with the previous non-null value
+// ```
+// import "sampledata"
+//
+// < sampledata.int(includeNull: true)
+// >     |> fill(usePrevious: true)
+// ```
```
sanderson, line 72:
```suggestion
// ```
// import "experimental/universe"
// import "sampledata"
//
// < sampledata.int(includeNull: true)
// >     |> experimental.fill(usePrevious: true)
// ```
```

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+builtin fill : (<-tables: stream[{A with C: B}], ?column: C, ?value: B, ?usePrevious: bool) => stream[{A with C: B}]
+    where
+    A: Record,
+    C: Label
+
+// mean returns the average of non-null values in a specified column from each
+// input table.
+//
+// ## Parameters
```
sanderson, line 86:
```suggestion
// input table.
//
// **Note:** `universe.mean()` is an experimental function with a more precise type signature.
//
// ## Parameters
```

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Examples
+//
+// ### Return the average of values in each input table
+// ```
+// import "sampledata"
+//
+// < sampledata.int()
+// >     |> mean()
+// ```
```
sanderson, line 98:
~~~suggestion
// ```
// import "experimental/universe"
// import "sampledata"
//
// < sampledata.int()
// >     |> universe.mean()
// ```
~~~

sanderson (changes requested):

libflux/flux-core/src/semantic/tests/labels.rs
```diff
@@ -0,0 +1,226 @@
...
+            z = [{ a: 1, b: ""}] |> fill(column: b, value: 1.0)
+        "#,
+        exp: map![
+            "b" => "string",
+            "x" => "[{ a: string }]",
+            "y" => "[{ a: int, b: float }]",
+            "z" => "[{ a: int, b: float }]",
+        ],
+    }
+}
```
nathanielc, line 30: This test case is great. Its super cool seeing this actually working in Flux!

libflux/flux-core/src/semantic/tests/labels.rs
```diff
@@ -0,0 +1,226 @@
...
+                "
+        ],
+        src: r#"
+            x = [{ a: 1, b: 2.0 }] |> f(value: "x")
+        "#,
+        exp: map![
+            "x" => "[{ a: string, b: float }]",
+        ],
+    }
+}
```
nathanielc, line 51: This test case confuses me, what is the purpose of the unbounded `B` label? Since `B` is never defined what is its purpose?

Is the idea to show that it doesn't error? Should it error?

libflux/flux-core/src/semantic/tests/labels.rs
```diff
@@ -0,0 +1,226 @@
...
+
+            error: record is missing label abc
+              ┌─ main:3:17
+              │
+            3 │             y = x.abc
+              │                 ^
+
+        "#]],
+    }
+}
```
nathanielc, line 178: To make sure I understand this, we can define a builtin type that requires a label, but if we don't enable the feature you won't be able to use that builtin value? So this test case shows that the label inference logic is disabled.

libflux/flux-core/src/semantic/tests/labels.rs
```diff
@@ -0,0 +1,226 @@
...
+            ..AnalyzerConfig::default()
+        },
+        env: map![
+            "columns" => "(table: A, ?column: C) => { C: string } where A: Record, C: Label",
+        ],
+        src: r#"
+            x = columns(table: { a: 1, b: "b" })
+            y = x.abc
+        "#,
+        // TODO Improve this error
```
nathanielc, line 216: Is this still a TODO? The error below is a bit confusing. I would have expected an error about `abc` not existing on `x`.

libflux/flux-core/src/semantic/tests/labels.rs
```diff
@@ -0,0 +1,226 @@
...
+
+#[test]
+fn optional_label() {
+    test_error_msg! {
+        config: AnalyzerConfig{
+            features: vec![Feature::LabelPolymorphism],
+            ..AnalyzerConfig::default()
+        },
+        env: map![
+            "columns" => "(table: A, ?column: C) => { C: string } where A: Record, C: Label",
```
nathanielc, line 210: How do we define a default for `C`?

stdlib/experimental/universe/universe.flux
```diff
@@ -0,0 +1,104 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+// tags: transformations, aggregates
+//
+builtin mean : (<-tables: stream[{A with C: B}], ?column: C) => stream[{C: B}] where A: Record, B: Numeric, C: Label
```
nathanielc, line 104: Same question I had before, where is `_value` defined as the default?

nathanielc (commented):
I gave this a full review. I still want to go back and understand the actual implementation better. For now I am mostly what to make sure I understand the behavior changes. 

Overall looks great and its super exciting to see this be real. I have some questions below.

libflux/flux-core/src/semantic/convert.rs
```diff
@@ -2668,13 +2688,13 @@ mod tests {
 
     #[test]
     fn test_convert_monotype_record() {
-        let monotype = Parser::new("{ A with B: int }").parse_monotype();
+        let monotype = Parser::new("{ A with b: int }").parse_monotype();
```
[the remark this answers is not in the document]
Marwes: Yes, it should be fine due to that control. We may still want to consider changing what we consider a type variable though since it isn't very obvious that `A`,`B` etc are type variables. (And changing it should be fine since we control every specified type)

libflux/flux-core/src/semantic/tests/labels.rs
Marwes: Yeah, this should error. I think I wrote this test case early on and it happens to still work.

libflux/flux-core/src/semantic/tests/labels.rs
Marwes: Yes. More precisely that we aren't creating any label types without the feature (which in turn prevents the builtin from being used).

libflux/flux-core/src/semantic/tests/labels.rs
Marwes: I started tinkering on a followup for allowing default values (default types) for optional function arguments. For now it should just error

libflux/flux-core/src/semantic/tests/labels.rs
Marwes: It is due to `Property` being tweaked https://github.com/influxdata/flux/pull/4388/files#diff-d5ab296ff16df9447d31cf953c8c275a43922b7cd3b260af30e0325abbdd53eaR1886-R1889 . Since keys may contain variables they need to be `Substitutable` which forces `String` to be `Substitutable`.

libflux/flux-core/src/semantic/sub.rs
```diff
@@ -231,6 +264,12 @@ pub trait Substitutable {
     }
 }
 
+impl Substitutable for String {
```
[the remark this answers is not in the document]
Marwes: It is due to `Property` being tweaked https://github.com/influxdata/flux/pull/4388/files#diff-d5ab296ff16df9447d31cf953c8c275a43922b7cd3b260af30e0325abbdd53eaR1886-R1889 . Since keys may contain variables they need to be `Substitutable` which forces `String` to be `Substitutable`.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -1742,13 +1754,19 @@ impl RegexpLit {
 pub struct StringLit {
     pub loc: ast::SourceLocation,
     pub value: String,
+    /// The (label) type if label types are enabled. Should be removed once label polymorphism are
+    /// stabilized
+    pub typ: Option<MonoType>,
```
wolffcm, line 1759: Why would this field be removed? It seems like a string literal can be typed as either a label or a string, so wouldn't this field need to remain? I must be missing something.

libflux/flux-core/src/semantic/convert.rs
```diff
@@ -579,7 +579,21 @@ impl<'a> Converter<'a> {
                 };
                 for prop in &rec.properties {
                     let property = types::Property {
-                        k: types::Label::from(self.symbols.lookup(&prop.name.name)),
+                        k: match &prop.name {
+                            ast::PropertyKey::Identifier(id) => {
+                                if id.name.len() == 1 && id.name.starts_with(char::is_uppercase) {
```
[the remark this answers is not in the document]
wolffcm: So in order to have a record type with a key that is just a single uppercase letter, we can just use `{"A": int}`, right?

wolffcm (commented):
Sorry for the delay in my review. I'm having to refamiliarize myself with a lot of this. I just had a few questions, but I feel like this is really close. It's very cool!

I guess before we can really turn this on, we need to have default values working, we'll have something that can replace the builtin types for many of our stdlib functions?

libflux/flux-core/src/semantic/nodes.rs
Marwes: `StringLit::type_of` could construct the type dynamically as `MonoType::Label(Label::from(&self.value))` instead of having it stored. But storing it may be preferable anyway, removed the comment.

libflux/flux-core/src/semantic/convert.rs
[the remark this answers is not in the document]
Marwes: Yes

Marwes:
> I guess before we can really turn this on, we need to have default values working, we'll have something that can replace the builtin types for many of our stdlib functions?

I think we need to update the standard library to use labels and have a feature flag through which we can thoroughly check for what breakages we get. Depending on what kind of breakage we see we may then either enable this outright, enable these errors as a warning (somehow) or consider some other option.

libflux/flux-core/src/parser/mod.rs
```diff
@@ -1481,6 +1521,20 @@ impl<'input> Parser<'input> {
...
     }
+    fn parse_property_key(&mut self) -> PropertyKey {
+        let t = self.expect_one_of(&[TokenType::Ident, TokenType::String]);
+        match t.tok {
+            TokenType::Ident => PropertyKey::Identifier(Identifier {
+                base: self.base_node_from_token(&t),
+                name: t.lit,
+            }),
+            TokenType::String => PropertyKey::StringLit(self.new_string_literal(t)),
+            _ => PropertyKey::Identifier(Identifier {
```
OfTheDelmer, line 1532: Could we use this to match `TokenType::Ident` case above?

libflux/flux-core/src/parser/mod.rs
```diff
@@ -734,15 +774,15 @@ impl<'input> Parser<'input> {
...
     }
-    fn parse_property_type_suffix(&mut self, id: Identifier) -> PropertyType {
+    fn parse_property_type_suffix(&mut self, name: PropertyKey) -> PropertyType {
         self.expect(TokenType::Colon); // :
         let monotype = self.parse_monotype();
         PropertyType {
-            base: self.base_node_from_others(&id.base, monotype.base()),
-            name: id,
+            base: self.base_node_from_others(name.base(), monotype.base()),
+            name,
```
OfTheDelmer, line 785: Clean! 🧼 ✨

runtime/lookup.go
```diff
@@ -84,7 +84,10 @@ func TypeEnvMap(env *fbsemantic.TypeEnvironment) map[envKey]*fbsemantic.Prop {
 		for j := 0; j < propLen; j++ {
 			newProp := new(fbsemantic.Prop)
 			_ = record.Props(newProp, j) // this call assigns value to newProp
-			propKey := string(newProp.K())
+			propKey, err := semantic.GetPropertyName(newProp)
+			if err != nil {
+				panic(fmt.Errorf("unexpected variable label in type environment: %s", err))
```
OfTheDelmer, line 89: Do we have a test for this property name panic?

OfTheDelmer (approved):

sanderson (approved):

wolffcm (approved):
Nice to see this moving along.

PR #4393, opened 2022-01-07
refactor: Preparatory cahnges for polymorphic labels

Marwes:
A couple of commits extracted from #4388 which could be merged independently. See individual commits.


rockstar (approved):
Looking at the individual commits was helpful for context. Thanks for making that suggestion.

PR #4400, opened 2022-01-11
feat(stdlib): create an optimized sort limit

### LINKED ISSUE #4397: Create Optimized SortLimit transformation
Top/Bottom are defined as a `sort |> limit` operation. We should create an optimized transformation that can implement sort + limit efficiently. This will mean sorting individual table chunks as they are read and only keeping only the N (as defined by limit) top/bottom rows. This will greatly reduce the amount of memory needed to perform top/bottom queries. 

DOD:

- Optimized version of sortLimit is implemented
- Planner rule rewrites sort |> limit to use the new transformation
- Use a feature flag to control using this new planner rule
- Flux acceptance tests added and updated to ensure correctness

jsternberg:
The optimized sort limit will take the pattern `sort |> limit` when
limit has no offset and will replace it with an optimized version that
performs the sorting continuously as data comes in rather than
performing a sort on all of the data and then doing the limit.

This is most analogous to performing a merge sort on each incoming
chunk of data.

Fixes #4397.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/sort_limit.go
+++ b/stdlib/universe/sort_limit.go
@@ -1,0 +1,225 @@
+package universe
+
+import (
+	"context"
+
+	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/arrow"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/internal/arrowutil"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/internal/feature"
+	"github.com/influxdata/flux/plan"
+)
+
+func init() {
+	plan.RegisterPhysicalRules(SortLimitRule{})
+	execute.RegisterTransformation(SortLimitKind, createSortLimitTransformation)
+}
+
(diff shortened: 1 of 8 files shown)


stdlib/universe/sort_limit.go
```diff
@@ -0,0 +1,225 @@
...
+		SortProcedureSpec: sortSpec,
+		limit:             limitSpec.N,
+	}
+
+	n, err := plan.MergeToPhysicalNode(node, sortNode, sortLimitSpec)
+	if err != nil {
+		return nil, false, err
+	}
+	return n, true, nil
+}
```
wolffcm, line 225: Could I trouble you to add some test cases for this rule?

wolffcm (changes requested):
This looks great, I'm just requesting test cases for the planner rule and there was an extra file that I don't think you wanted to commit.

test.flux
```diff
@@ -0,0 +1,5 @@
+import "internal/gen"
+
+gen.tables(n: 100)
+|> sort()
+|> limit(n: 10)
```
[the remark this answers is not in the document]
jsternberg: Yea definitely an accident.

stdlib/universe/sort_limit.go
jsternberg: Yep, can do.

wolffcm (approved):

PR #4405, opened 2022-01-14
feat: Always provide a best effort value after compilation

Marwes:
This is helpful for the language server as it will be able to run its analysis even if compilation returns an error.

There may still be places where we can produce a more complete semantic graph in the face of errors, but other than that this is the final result of all the changes towards collecting errors and continuing despite them. The `Salvage/SalvageResult` may be a bit awkward to use so we may want some different API(s) in place of those but I mostly just copied them from `gluon` for now.


Marwes:
It may be helpful to review commit by commit for this change.

Marwes:
Closed by #4411

PR #4408, opened 2022-01-14
feat(dependencies): Add IOx Client to dependencies

onelson:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


onelson:
Seems like we're looking to manage all the coupling in IDPE to start. Not sure if/when we'll need supporting changes in flux itself.

PR #4409, opened 2022-01-14
chore: add a method to create a String given an existing array

wolffcm:
This allows us to create a `String` (which is a wrapper around the Arrow array type that also allows for a compressed representation of arrays when all the elements are the same) from a native Arrow array. This change is necessary to bring arrays coming form Arrow Flight into Flux.

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -99,6 +99,19 @@ type String struct {
 	data   *array.Binary
 }
 
+// NewStringFromBinaryArray creates an instance of String from
+// an Arrow Binary array.
+//
+// Note: Generally client code should be using the types for arrays defined in Flux.
+// This method allows string data created outside of Flux (such as from Arrow Flight)
+// to be used in Flux.
+func NewStringFromBinaryArray(data *array.Binary) *String {
+	data.Retain()
+	return &String{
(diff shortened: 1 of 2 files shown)


onelson:
Coming from a Rust perspective, I scare easy when it comes to building strings from raw buffers. Does Go account for interior nul bytes and so on? No idea if there's a concern but wanted to ask.

wolffcm:
@onelson Go's `string` type is not null-terminated, and is just a sequence of bytes, so it can contain null bytes without any unexpected behavior. In fact field values in InfluxDB are allowed to contain byte sequences that are not valid UTF8 (I would guess including the null byte) and my understanding is that this is why we represent string data in Flux not with the Arrow `String` array type, but with `Binary` instead.

onelson:
Got it. Rust's String strictly enforces the UTF-8'edness, and expects you to ensure this when building from raw bytes so that's different. I guess this is potentially a garbage-in-garbage-out situation if a nul gets passed through.

onelson (approved):
You didn't ask me for a review, but this LGTM anyway :rofl:

Marwes (approved):

PR #4411, opened 2022-01-18
perf: Take the AST by reference during semantic conversion

Marwes:
While this does forces convert to clone some `String` values it is an improvement on the whole as the caller were forced to clone the entire AST in most cases.

Based on #4405


jsternberg (commented):
This will allow a related change on the C/Go side and it was a quite onerous refactor to do. The C side will "consume" the AST when it invokes analyze which is pretty strange and causes some weird code on the Go side. It's not a thing I generally like very much.

One question I have is will this have a performance impact? It seems to me like most of the time we would take an AST and produce the semantic graph. We will then, in most cases, discard the AST as we no longer need it. It seems to me like the semantic graph will often take attributes from the AST in a direct copy/move. If we pass by reference, I assume we have to copy the values instead of moving them even though we're going to discard the original.

I just want to consider the above and, if we find the refactor still good, I can approve and merge this.

Marwes:
> One question I have is will this have a performance impact? It seems to me like most of the time we would take an AST and produce the semantic graph. We will then, in most cases, discard the AST as we no longer need it. It seems to me like the semantic graph will often take attributes from the AST in a direct copy/move. If we pass by reference, I assume we have to copy the values instead of moving them even though we're going to discard the original.

The only values that `convert` may be able to reuse is the `String` values, which isn't a lot since we use `Symbol` now for identifier so it is only string literals that may end up copied an extra time. All other values such as `Vec`, `BTreeMap` etc tend to contain an `ast` type so we can't reuse that `Vec` anyway but must allocate a new one for the `semantic` graph.

This doesn't take into account that we were forced to clone the `ast` for any imported module before checking it which is quite significant (could maybe be restructured :shrug: ).

That said, if cloning the `String` values is actually noticeable (which i doubt), we can just change the AST and semantic graph to hold `Arc<str>` instead of `String`.

jsternberg (approved):
This looks good then. Just rebase and squash the commits.

rockstar (approved):

Issue #4389, opened 2021-12-30
Unable to write data to PostgreSQL from flux task


vkhazin:
Sample code:

```go
import "sql"

option task = {name: "export-to-rdbms", every: 24h}

connectionString =
    "postgresql://postgres:<secret>@<masked>.us-east-2.rds.amazonaws.com:5432/postgres"

// Reading data works!
// data =
//     sql.from(
//         driverName: "postgres",
//         dataSourceName: connectionString,
//         query: "SELECT measurement, field, value, \"timestamp\" FROM some_table;",
//     )
data =
    from(bucket: "some-data")
        |> range(start: -30d)
        |> map(fn: (r) => ({timestamp: r._time, measurement: r._measurement, field: r._field, value: r._value}))

// Writing data does not work!
data
    |> sql.to(driverName: "postgres", dataSourceName: connectionString, table: "some_table")
```

Error:
```text
Unexpected error from query; Err: failed to initialize execute state: pq: password authentication failed for user "postgres": failed to initialize execute state: pq: password authentication failed for user "postgres"
```

nathanielc:
@vkhazin Thanks for the report, we have a large batch of updates to the `sql.to` and `sql.from` functions in the works. We'll add this bug to the testing to ensure it gets fixed.

danatinflux:
@vkhazin I was able to get this working by removing network restrictions on the AWS/Postgres side:

![image](https://user-images.githubusercontent.com/44617130/148587488-a9238912-8ce9-4c8c-a75e-ef2ed3a52303.png)

Initially, I wasn't able to make connections to my OSS box until I added the top exclusion (with the /32 subnet). That all worked fine. Then, I tried the C2 version and again, no connectivity. However, when I added the very insecure 0.0.0.0/0 source (middle) it started to work.

Unfortunately, we don't currently have a list of IPs that a customer would be able to whitelist at this time as all the IPs for Cloud are dynamic.

vkhazin:
@danatinflux thank you for your comments!
The situation is different for me: the security group is configured for 0.0.0.0/0 access with no success.
And the InfluxDB is able to read data using `sql.from`, which would not be possible with a security group not configured for wide open access.
InfluxDB task with flux is failing to execute `sql.to` command to AWS RDS Postgres and is succeeding with the same command to AWS RDS MySql.
Other than a problem with documented PostgreSql setup or with `sql.to` implementation I cannot explain the observation.

danatinflux:
[...]

PR #4413, opened 2022-01-19
chore: Allow the integration databases to be started separately

Marwes:
Testing postgres for #4389 and had issues downloading the SAP hana image so I tweaked the script so that you can chose to only start a subset of databases.

### DIFF
--- a/etc/spawn-dbs.sh
+++ b/etc/spawn-dbs.sh
@@ -262,40 +262,42 @@ function run_pg {
 function should_start {
     # Start a databases if no the script is invoked without any arguments, or if the database is
     # among the arguments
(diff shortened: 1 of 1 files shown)


onelson:
The detach happens immediately, and then there's a tail of startup activity after that point. That's why we have to "wait for" the engines to be ready as a separate step.

For many engines, it's not a huge concern, but for a select few (Vertica and HANA are particularly egregious iirc) the startup can take _minutes_.

Marwes:
Of course, database startup, not docker container startup.

onelson (approved):
```
Thanks for going the extra mile! &
wait
```

This will really help iterating on the full set. 2 or 3 minutes was already too long to wait, but it's much better than 5 or 6.

PR #4416, opened 2022-01-20
fix: update to rust 1.58.1

rockstar:
With [CVE-2022-21658](https://blog.rust-lang.org/2022/01/20/cve-2022-21658.html), rust 1.58 is no longer safe. This patch updates to use 1.58.1


onelson (approved):
:crab: ++

pierwill:
1.58.1 should be released any minute now. :)

https://bors.rust-lang.org/queue/rust

PR #4419, opened 2022-01-24
feat: Re-index all bound variables to start from 0

Marwes:
This naturally makes function signatures more readable as they will be `id: (x: A) => A` instead of `id: (x: t1234) => t1234` (for instance in the LSP) and it makes the types of exported bindings less dependent on each other (adding a new export no longer forces all exports defined after it to increment their variable numbers).  This does complicate the go side slightly since it previously would assume that variables were unique, however that might actually be a bug as the same variable can (and usually do) exist in two different packages. Not sure if it is actually an issue though.


libflux/flux-core/src/semantic/infer.rs
```diff
@@ -175,12 +175,72 @@ pub fn equal(
...
+        log::debug!("Unify error: {} <=> {}", exp, act);
+
+        Located {
+            location: loc.clone(),
+            error,
+        }
     })
 }
 
+pub(crate) fn temporary_generalize(
```
nathanielc, line 188: Can you add a doc comment here? How is temporary_generalize different from the normal generalize function?

libflux/flux-core/src/semantic/infer.rs
```diff
@@ -236,6 +334,19 @@ pub fn instantiate(
                 .collect::<Vec<Constraint>>()
                 .into()
         });
+
+    // Equivalent to `SubstitutionMap` but instantiates bound variables instead of free variables
+    struct InstantiationMap(SubstitutionMap);
```
nathanielc, line 339: Is this how free vars become bound vars? During instantiation?

libflux/flux-core/src/semantic/tests.rs
```diff
@@ -2848,44 +2848,44 @@ fn identity_function() {
...
+    // test_infer! {
+    //     src: r#"
+    //         f = (x) => 0 |> x()
+    //         f(x: (v=<-) => v)
+    //         f(x: (w=<-) => w)
+    //     "#,
+    //     exp: map![
+    //         "f" => "(x:(<-:int) => C) => C",
+    //     ]
+    // }
```
nathanielc, line 2888: Do these tests need to be uncommented now?

libflux/flux-core/src/semantic/formatter/tests.rs
```diff
@@ -111,19 +111,19 @@ fn format_function_expression() {
...
     check(
         script,
-        expect![[r#"
+        expect![[r##"
             package main
             (a) => {
-                return a:A
-            }:(a:A) => A
+                return a:#A
+            }:(a:#A) => #A
```
nathanielc, line 118: Reading the code `#` means the variable is free. Why is `A` considered free here? My expectation is that `A` would be bound to the polytype of the function. What and I misunderstanding?

nathanielc (commented):
Changes generally look good. I like the clean up of the Go code to make it more explicit that we are substituting types. 

I have some questions as I am not clear on when/how a free variable becomes bound. See below.

libflux/flux-core/src/semantic/formatter/tests.rs
Marwes: It should be, but the way the code works now we perform generalizations when it is bound to a variable, not when the function is created. I don't think this is entirely correct, but I didn't want to change the behavior here.

libflux/flux-core/src/semantic/infer.rs
Marwes: It is the reverse, we have a `PolyType` with variables bound to it and we instantiate them to free variables.

nathanielc (approved):

PR #4421, opened 2022-01-24
feat(dependency): add dependency hooks to the dependency subsystem

### LINKED ISSUE #4278: [mqtt.publish] Minimize TCP connection wait time from synchronous cal to mqtt.publish()
Further work on this PR: https://github.com/influxdata/flux/pull/4081

@alespour Regarding the quick conversation you and I had about the synchronous nature of `mqtt.publish()`, I had a conversation with @nathanielc about how to handle this.  It sounds like a good option would be to simply reuse the TCP connection as we loop through each call to `publish()` in the `map()` call.  According to Nathaniel, subsequent calls to publish can keep state on the Go side.

@alespour thoughts?

jsternberg:
This adds hooks to the dependency subsystem to run certain actions at
certain times of the execution. At present, the only hook is the
`OnFinish` hook which runs when execution has finished.

The mqtt dependency has been modified to use a pool dialer by default.
The pool dialer will retrieve an `mqtt.Client` that has the same options
and brokers that was previously "closed" to avoid establishing a new
connection. This means repeated calls to `mqtt.publish()` will use the
same client.

Pooled clients are cleaned up when `OnFinish()` is called.

Fixes #4278.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/dependencies/mqtt/dialer_test.go
+++ b/dependencies/mqtt/dialer_test.go
@@ -4,8 +4,32 @@ import (
 	"context"
 	"testing"
 
+	"github.com/influxdata/flux/dependencies/feature"
 	"github.com/influxdata/flux/dependencies/mqtt"
+	"github.com/influxdata/flux/dependency"
+	"github.com/influxdata/flux/execute/executetest"
 )
+
+type MockDialer struct {
+	DialFn func(ctx context.Context, brokers []string, options mqtt.Options) (mqtt.Client, error)
+}
+
+func (m *MockDialer) Dial(ctx context.Context, brokers []string, options mqtt.Options) (mqtt.Client, error) {
+	return m.DialFn(ctx, brokers, options)
+}
+
+type MockClient struct {
+	PublishFn func(ctx context.Context, topic string, qos byte, retain bool, payload interface{}) error
+	CloseFn   func() error
+}
+
+func (m *MockClient) Publish(ctx context.Context, topic string, qos byte, retain bool, payload interface{}) error {
+	return m.PublishFn(ctx, topic, qos, retain, payload)
+}
+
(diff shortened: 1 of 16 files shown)


dependencies/mqtt/mqtt.go
```diff
@@ -146,3 +155,87 @@ type ErrorDialer struct{}
...
+
+func (p *poolDialer) Close() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	var firstErr error
+	for _, clients := range p.clients {
+		for _, client := range *clients {
+			if err := client.Close(); err != nil && firstErr == nil {
+				firstErr = err
```
rockstar, line 212: If we only show the first error, are we potentially hiding other errors on `client.Close`?

dependencies/mqtt/dialer_test.go
```diff
@@ -4,9 +4,33 @@ import (
 	"context"
 	"testing"
 
+	"github.com/influxdata/flux/dependencies/feature"
 	"github.com/influxdata/flux/dependencies/mqtt"
+	"github.com/influxdata/flux/dependency"
+	"github.com/influxdata/flux/execute/executetest"
 )
 
+type MockDialer struct {
```
rockstar, line 13: I like this mock. It looks closer to what I expect from mocks than other mocks I've seen in our various go codebases.

rockstar (approved):
The dependency system is one of those things I have to re-learn every time I touch it or have to review it. Sorry for how long this took as a result.

alespour:
@samhld As per @jsternberg comment, if pool dialer is used by default, then I think there nothing for me to be done in `experimental/mqtt` pkg.

Issue #4381, opened 2021-12-28
Improve error feedback for invalid field access in `contains`


onelson:
When you filter using `contains` and the `value` parameter receives an invalid field (missing from the series), the error message we give today is poor:
```
array.from(rows: [{}]) |> debug.opaque() |> filter(fn: (r) => contains(value: r["undefined field"], set:[1]))
Error: runtime error @1:45-1:110: filter: keyword argument "set" should be of an array of type invalid, but got an array of type [int]
```

Instead of telling the user `set` should be _"an array of type invalid"_ it would be better if we could highlight the `value` as being missing.

Refs: https://github.com/influxdata/influxdb/issues/23025

onelson:
An alternative to improving the error message here might be to treat an invalid `value` parameter was an automatic `false` just like we do for the empty `set` case. This would align with the behavior we see with filter, where a flux program like:

```
import "array"
import "internal/debug"
array.from(rows: [{}]) |> debug.opaque() |> filter(fn: (r) => r.anything == 1)
```
will give an empty result, not an error.

onelson:
During grooming, it sounded as if we'd like it if accessing an unknown field to be falsy (like we see for the `filter()` example above). The error message won't be addressed if we can make the comparison fail quietly.

onelson:
Interesting to note that the behavior is different for filter between a case where one side of the comparison is invalid but the other isn't, versus both sides being invalid.
```
> import "array"
> import "internal/debug"
> array.from(rows:[{a:1}]) |> debug.opaque() |> filter(fn: (r) => r.foo == r.bar)
Result: _result
Error: runtime error @1:47-1:80: filter: cannot compile @ 1:58-1:79: unsupported binary expression invalid == invalid
> array.from(rows:[{a:1}]) |> debug.opaque() |> filter(fn: (r) => r.foo == 1)
Result: _result
```

It seems like `invalid == invalid` should be `false` in the way we see when comparing `NaN` to `NaN` in JS. It might be worth filing a separate issue to look at comparisons of Invalid values to each other.

onelson:
Fixed with #4424

PR #4424, opened 2022-01-25
fix(universe): return false if contains gets invalid value

onelson:
Flux treats the access of a missing field in the `value` parameter for `contains` as a hard error, whereas similar access in `filter` is treated as "falsy."

As noted in <https://github.com/influxdata/influxdb/issues/23025>, it's possible the error behavior was introduced as a breaking change accidentally in <https://github.com/influxdata/flux/pull/2696> since a user raised the issue as something that worked previously but  no longer does.

This patch checks to see if `value` is invalid and immediately returns `false` if it is, sidestepping the type check between the `value` and `set` parameters that produces the error.

Refs:
- <https://github.com/influxdata/flux/issues/4381>
- <https://github.com/influxdata/influxdb/issues/23025>
- <https://github.com/influxdata/flux/pull/2696>

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/universe/contains.go
+++ b/stdlib/universe/contains.go
@@ -7,6 +7,7 @@ import (
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/runtime"
+	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
 )
 
@@ -24,6 +25,14 @@ func MakeContainsFunc() values.Function {
 				return nil, err
 			}
 
+			// In situations where value `v` is "invalid" (like when it's a field that is missing from an incoming
+			// unknown record type, we give a surprising and confusing error message indicating
+			// `keyword argument "set" should be of an array of type invalid`.
+			// This error is normally produced by the `GetRequiredArrayAllowEmpty` call immediately below.
+			// To avoid the confusion and handle this case more intuitively, short-circuit with a "false" outcome.
+			if v.Type().Nature() == semantic.Invalid {
+				return values.NewBool(false), nil
+			}
 			setarg, err := a.GetRequiredArrayAllowEmpty("set", v.Type().Nature())
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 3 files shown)


wolffcm (commented):
Looks good, appreciate the test for it.

I wonder what should happen for a case like
`contains(value: 10, set: [r.foo, r.bar, r.unknown])`

Should that case be covered too?

onelson:
> Should that case be covered too?

It's an interesting case. Currently this fails with:

```
invalid_set_field_is_falsy...fail: runtime error @43:16-43:84: filter: failed to evaluate filter function: value type semantic.MonoType does not match set type semantic.MonoType
```

That's not a user-friendly message.

In #4381, I noted it's currently an error to compare two invalid values so I suspect there's a whole host of cases like this.

```
> import "array"
> import "internal/debug"
> array.from(rows:[{a:1}]) |> debug.opaque() |> filter(fn: (r) => r.foo == r.bar)
Result: _result
Error: runtime error @1:47-1:80: filter: cannot compile @ 1:58-1:79: unsupported binary expression invalid == invalid
```

I can certainly chase this 2nd case down since it relates directly to `contains()`, but I'm a little wary of pulling too far off track from the initial report.

stdlib/universe/contains.go
```diff
@@ -35,15 +44,19 @@ func MakeContainsFunc() values.Function {
 			if _, ok := set.(values.TableObject); ok {
 				return nil, errors.New(codes.Invalid, "set cannot be a table stream; expected an array")
 			}
+
 			if set.Len() > 0 {
-				if set.Get(0).Type() != v.Type() {
```
onelson, line 39: @wolffcm my thinking here is this strict type check is not really possible if we accept `Invalid` values can appear in the `set` array. I moved this check to be per item as we visit each one -- not sure if this is problematic somehow.

onelson:
I pushed a couple more revs to handle the case where `Invalid` appears inside the `set` array.

wolffcm (approved):
Looks good, thanks for the tests.

PR #4427, opened 2022-01-25
refactor(value): Add vector value type

scbrickley:
Related to #4185

At the request of the reviewers, #4356 has been split into 2 PRs. This one, which adds a new `Vector` type and makes changes to the `Value` interface, and another that uses these changes in the flux compiler (see #4435).

This patch introduces a new `Vector` interface that transformations can use to take advantage of the performance gains from vectorization. In addition to the new interface, the `internal/arrowutil` package now contains some concrete types that implement the `Vector` interface, as well as some helper functions for working with vectors.

This patch also adds two methods to the `Value` interface: `Retain()` and `Release()`. For most values, these methods will be a no-op, but for vectors and arrays (or types that contain vectors or arrays) they will help keep accurate reference counts on the underlying arrow arrays.


skartikey (changes requested):
The change looks good overall. Have a couple of points.
1) I can see you have added the empty Retain/Release method in a few files to support the `Value interface`. Are you going to do implementation later on those empty methods?

2) Kindly add some tests to support this change.

scbrickley:
Thanks for the review @skartikey. Sorry about the late response.

I've updated the description with some of the necessary context for this patch. Hopefully that clarifies things a bit.

>     1. I can see you have added the empty Retain/Release method in a few files to support the `Value interface`. Are you going to do implementation later on those empty methods?

It's intended that, in most cases, calls to Retain and Release will be no-ops. The only time it will actually do anything is if we call it on value types that are backed by arrow arrays (i.e., arrays and vectors), or on values that contain one of those types.

>     2. Kindly add some tests to support this change.

There are tests in the companion patch to this (#4435). If those aren't sufficient, let's talk about what kinds of tests you'd like to see, and I'd be happy to add them.

wolffcm (changes requested):
I agree with Sunil, it would be good to have some basic unit tests for the new `VectorValue` types, that just shows the `Vector` interface methods working.

wolffcm:
But I should add this this looks nice otherwise!!

values/values.go
```diff
@@ -148,6 +151,16 @@ func (v value) Equal(r Value) bool {
 	}
 }
 
+func (v value) Retain() {
```
jsternberg, line 154: I'm a bit confused by this. To my knowledge, value is only used by primitive types so it shouldn't need retain and release. I consider it a bit suspect that this calls New and then retain or release. That doesn't make much sense to me.

internal/arrowutil/vector_values.gen.go
```diff
@@ -0,0 +1,481 @@
+// Generated by tmpl
```
jsternberg, line 1: It is possible to get this file into the values package?

semantic/monotype.go
```diff
@@ -263,25 +268,53 @@ func (mt MonoType) ReturnType() (MonoType, error) {
...
+	return c.ElementType()
 }
 
-// ElemType returns the element type if this monotype is an array, and an error otherise.
-func (mt MonoType) ElemType() (MonoType, error) {
-	arr, err := getArr(mt.tbl)
-	if err != nil {
-		return MonoType{}, err
-	}
+type Collection interface {
```
jsternberg, line 280: Can you make this an unexported interface? Just make it lowercase.

jsternberg (changes requested):
Three comments but overall looks fine. I think this is probably good to merge. After you respond to the comments, I'll probably just switch this to approve.

internal/arrowutil/vector_values.gen.go
scbrickley: Yes. I originally put it here because I was using `arrowutil/array_values.gen.go` as a starting point for this, and it was convenient to just use the same template data. Now that I have something that works though, I'll move it to the values package.

values/values.go
scbrickley: You're right. I think this is the result of me just misunderstanding the code. I'll change these to be no-ops.

semantic/monotype.go
scbrickley: Can do.

scbrickley:
Addressed @jsternberg's feedback. Also added some tests, although they're pretty minimal right now. Open to feedback on any additional test coverage that might be needed.

jsternberg (approved):
I think this is good to merge.

wolffcm (approved):
Looks great.

values/vector_test.go
```diff
@@ -0,0 +1,48 @@
...
+			input:    []interface{}{true, false, true},
+			wantType: semantic.BasicBool,
+		},
+	}
+	for _, tc := range testCases {
+		got := NewVectorFromElements(nil, tc.input...)
+
+		if !got.ElementType().Equal(tc.wantType) {
+			t.Errorf("expected %v, got %v", tc.wantType, got.ElementType())
+		}
```
wolffcm, line 46: It would be good to use an allocator here and show that when the vectors are released the allocations go to zero.

skartikey (approved):

Issue #3653, opened 2021-04-28
New Full featured HTTP package


nathanielc:
Connecting to external systems over HTTP has proven to be an important part of Flux. We should create a well designed HTTP package that provides a consistent API for performing HTTP requests.

Here is the proposed package:

```
package http

_emptyBody = bytes(v: "")

// DefaultConfig is the global default for all http requests using the http package.
// Changing this config will affect all other packages using the http package.
// To change the config for a single request pass a new config directly into the corresponding function.
option defaultConfig = {
    // Timeout on the request, if the timeout is zero no timeout is applied
    timeout: 0s,
    // VerifyTLS if false TLS verification will not be performed. This is insecure.
    verifyTLS: true,
}

// Internal method used to perform the actual request
builtin _request : (
    url: string,
    method: string,
    config: {A with timeout: duration, verifyTLS: bool},
    ?headers: [string:string],
    ?body: bytes,
) => {
    statusCode: int,
    body: bytes,
    headers: [string:string],
}

// Make an HTTP request using the provided config
request = (url, method, headers=[:], config=defaultConfig, body=_emptyBody) => _request(
    url: url,
    method: method,
    headers: headers,
    config: config,
    body: body,
)

// Post makes a POST HTTP request
post = (url, headers=[:], config=defaultConfig, body=_emptyBody) => request(
    url: url,
    method: "POST",
    headers: headers,
    config: config,
    body: body,
)

// Get makes a GET HTTP request
get = (url, headers=[:], config=defaultConfig, body=_emptyBody) => request(
    url: url,
    method: "GET",
    headers: headers,
    config: config,
    body: body,
)

```

Note callers can easily overwrite the default config on a per request basis without having to specify all existing defaults.

```
package main

import "experimental/http"

http.get(
    url: "http://example.com",
    config: {http.defaultConfig with verifyTLS: false}
)
```

This work will depend on https://github.com/influxdata/flux/issues/3629 to allow for default arguments types to be generalized. It will also depend on https://github.com/influxdata/flux/issues/3655 to allow the `{http.defaultConfig with timeout: 1s}` syntax.

DOD:

* The `experimental/http` package is updated according to the above design
* Flux acceptance tests have coverage for this package. This will require some kind of httptest package equivalent that we will likely need to NOT expose in production except for testing.

nathanielc:
[...]

PR #4429, opened 2022-01-25
feat(http): adds support for generic http requests

nathanielc:
This commits add a new package experimental/http/requests API to allow for generic
request response interactions over HTTP.

This does not change existing http.get or http.post as there are
thousands of live production queries using those methods.

The plan is to use this new requests package for a few weeks/months in
experimental and shake out all the bugs. Then promote it out of
experimental and then add deprecation warnings to the existing
functions.

Fixes: https://github.com/influxdata/flux/issues/3653

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/http/requests/requests.go
```diff
@@ -0,0 +1,214 @@
...
+				// Alias the DNS lookup error so as not to disclose the
+				// DNS server address. This error is private in the net/http
+				// package, so string matching is used.
+				if strings.HasSuffix(err.Error(), "no such host") {
+					return 0, nil, nil, errors.New(codes.Invalid, "no such host")
+				}
+				return 0, nil, nil, err
+			}
+			body, err := ioutil.ReadAll(response.Body)
+			_ = response.Body.Close()
```
skartikey, line 172: if `response.Body` is nil, the code will panic with the invalid memory address or nil pointer dereference. So I guess moving this code line after `nil` error check below will make sure that the code returns if there's any error returned from `ioutil.ReadAll` or we can put a nil check on `response.Body` before calling Close on it.

skartikey (changes requested):
Looks good to me with 1 minor change mentioned below.

stdlib/experimental/http/requests/requests.go
nathanielc: @skartikey and I chatted out of band but Go guarantees the Body is non-nil, see https://pkg.go.dev/net/http#Response

skartikey (approved):

nathanielc:
One design question I had is should we make params take a list of values instead of just a single value?

Specifically should the API for `do` become?

```
builtin _do : (
        method: string,
        url: string,
        // The PR current has ?params: [string: string]
        ?params: [string:[string]],
        ?headers: [string:string],
        ?body: bytes,
        config: {A with timeout: duration, insecureSkipVerify: bool},
    ) => {statusCode: int, body: bytes, headers: [string:string]}
```

onelson (commented):
Just looking at the high level, the docs show much of what I'd expect to see.

I'm not sure what the default HTTP client behavior is, but somethings I might expect to configure:

- redirect policy (follow or not)
- loading a custom CA (which is more desirable than skipping cert verification)

There is some supporting API I'd likely reach for while using this which isn't included. I'm not sure if it belongs here or in a separate package.

Concretely:
- functions to help build urls (specifically to help URI encode path segments and query params)
- something to base64 encode strings (commonly needed to prepare headers/bodies)

I forget how python's `requests` handles encoding query params (if it's automatic or if the caller is expected to do it themselves). I'd weakly suspect it's the caller's job since the server implementation in question could be one that doesn't adhere to the HTTP spec (it happens), so callers may need ways to "do the wrong thing on purpose."

I think there is a separate ticket out there related to hashing (md5, sha) which is often needed for making REST API requests. Common need is to hash some parts of the request params _with the current timestamp_, so it has to be done as close to the request being sent as possible.

Food for thought.

onelson:
> One design question I had is should we make params take a list of values instead of just a single value?

I must have overlooked this in my skimming of the diff! Being able to support multiple occurrences of a given key is important. Commonly I've seen params expressed as a sequence of tuple pairs. If a map of `string: [string]` makes more sense for flux, then I'd take it, but in other langs I'd expect to be able to push new pairs into an indexed collection.

stdlib/experimental/http/requests/requests.flux
```diff
@@ -0,0 +1,207 @@
...
+        params: params,
+        headers: headers,
+        body: body,
+        config: config,
+    )
+
+// get makes a http GET request. This identical to calling `request.do(method: "GET", ...)`.
+//
+// ## Parameters
+// - url: url to request, this should not include any query parameters.
```
noramullen1, line 178:
```suggestion
// - url: URL to request. This should not include any query parameters.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -0,0 +1,207 @@
...
+// - params: params is a set of key value pairs to add to the URL as query parameters.
```
noramullen1, line 179:
```suggestion
// - params: Set of key value pairs to add to the URL as query parameters.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -0,0 +1,207 @@
...
+// - headers: headers is a set of key values pairs to include on the request
```
noramullen1, line 180:
```suggestion
// - headers: Set of key values pairs to include on the request.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -0,0 +1,207 @@
...
+// - body: body is the data to send with the request
```
noramullen1, line 181:
```suggestion
// - body: Data to send with the request.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -0,0 +1,207 @@
...
+// - config: config is the set of options to control how the request should be performed.
```
noramullen1, line 182:
```suggestion
// - config: Set of options to control how the request should be performed.
```

noramullen1 (commented):
Minor help text cleanup

nathanielc:
@noramullen1 Thanks updated!

@onelson Thanks for the thoughts.

>  redirect policy (follow or not)

This seems like something we can add easily when it comes up. My hope was to design a config structure that makes adding new configuration easy while not having to anticipate all use cases from the start.

> loading a custom CA (which is more desirable than skipping cert verification)

This one is also one we could add later but its a little trickier. In our cloud offering where would you load the CA keys? You could possibly store them in our secrets store but that seems a bit heavy handed. For OSS allowing the config to point to files on disk would work well. Again I expect we could add this at a later point.



> functions to help build urls (specifically to help URI encode path segments and query params)

The current `requests.do` will encode the query params for you so its not strictly needed with the current API.

>  something to base64 encode strings (commonly needed to prepare headers/bodies)

I like this idea and agree a separate package is a likely a good place to do this. 


> I think there is a separate ticket out there related to hashing (md5, sha) which is often needed for making REST API requests. Common need is to hash some parts of the request params with the current timestamp, so it has to be done as close to the request being sent as possible.

I am not sure how we would do this one well. We have the ability to get the current time of a running Flux script, however it has some unique behaviors to Flux that might make this hard. Flux likes to pretend that the entire script runs in the same instant. This way if you are query data for different ranges you don't get slightly differing time ranges because of when the query plan itself was constructed. Not sure how I would handle this case.

nathanielc:
> Commonly I've seen params expressed as a sequence of tuple pairs. If a map of string: [string] makes more sense for flux, then I'd take it, but in other langs I'd expect to be able to push new pairs into an indexed collection.

Flux doesn't have unnamed tuples, so you would either have to create a record for this purpose, like `[{key: string, value: string}]` or use a dictionary `[string:[string]]` As a result I think we should use the dictionary as it result in less verbose code and maintains good clarity for what is expected.

onelson:
I totally agree that my wishlist of batteries to include can come later.

Small nitpick on

> The current `requests.do` will encode the query params for you so its not strictly needed with the current API.

One common place not covered by this behavior is in path segments. If they are using string interpolation to build their URL, they may need a way to encode the value... but again, we don't have to address all of this now. It's just the stuff I'd expect to find as a user.

nathanielc:
I just finished the update to make `params` handle multiple values with the same key. I am realizing headers have the same behavior. Should I also change headers to be `[string:[string]]`? This one makes working with headers harder as they are both and input and output of these functions. Thoughts?

onelson:
News to me if repeated header keys are valid -- the only time I've seen them in the wild, it was due to a configuration error on a reverse proxy, and it caused clients to have issues. I thought headers were not expected to be repeated. Client APIs normally encode headers as a flat map, but now I'm not sure what to think...

onelson:
TIL https://stackoverflow.com/a/4371395

I'd only ever seen cache-control using the comma separated list form before today. Huh.

nathanielc:
Thanks for the link, given that there is an equivalent way to construct the headers as CSV values then I think we should keep our API simpler and just support `[string:string]`. Then if a user needs multiple values they will be forced to use CSV formatting of the value which is acceptable. Also since order is important using CSV values make that more clear.

nathanielc:
@onelson Thanks for all the feedback, can you give this final state a review before merging?

onelson (approved):
Looks good to me. The test cases hit the major notes and the docs read well.

PR #4430, opened 2022-01-26
feat(stdlib): add experimental/iox package and a stub for the from function

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/iox/iox.flux
```diff
@@ -0,0 +1,15 @@
+package iox
```
sanderson, line 1:
```suggestion
// Package iox provides functions for querying data from IOx.
//
package iox
```

stdlib/experimental/iox/iox.flux
```diff
@@ -0,0 +1,15 @@
...
+//
+// This function creates a source that will read data from iox. Data that is
+// returned will be in a "pivoted" format where tags will be a column with all
+// identical attributes and fields will populate a column with the name of the field
+// and the column will be values associated with the time of the point.
+//
+// ## Parameters
+// - bucket: The bucket that data will be read from.
+//
+// - measurement: The measurement that data will be read from.
```
sanderson, line 14:
```suggestion
// from reads from the selected bucket and measurement in an iox storage node.
//
// This function creates a source that reads data from IOx. Output data is
// "pivoted" on the time column and includes columns for each returned
// tag and field per time value.
//
// ## Parameters
// - bucket: IOx bucket to read data from.
// - measurement: Measurement to read data from.
```

sanderson (commented):
Just some doc suggestions.

onelson (approved):
LGTM (as fine a stub as I ever did see).

I'm wondering if we have a way to hide this for users until we're ready for them to try it?

jsternberg:
CircleCI isn't healthy for builds requiring a VM (the windows build). I'm going to merge this with that failing since I don't think there's any reason it would fail there if CI was healthy.

PR #4431, opened 2022-01-26
feat(stdlib): update limit transformation to use table chunks

### LINKED ISSUE #3939: Update limit transformation to use NarrowStateTransformation
Update the `limit()` transformation to use [NarrowStateTransformation](https://github.com/influxdata/flux/issues/3933).

The `limit()` transformation presently uses `table.Stream()` to stream tables to the downstream. The transformation should be adapted to use `NarrowStateTransformation` instead. This should involve the following steps:

1. Remove `table.Stream()`.
2. For each group key, store the state for the limit/offset variables of the table.
3. Run the normal limit algorithm.
4. Send table chunks using the `TransportDataset`.

jsternberg:
Fixes #3939.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/limit.go
+++ b/stdlib/universe/limit.go
@@ -1,15 +1,14 @@
 package universe
 
 import (
-	"context"
-
+	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/arrow"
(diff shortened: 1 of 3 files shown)


jsternberg:
I thought this would be easier but I might have to change some internals which I wasn't expecting. Converting this to a draft so it shows as a work in progress.

jsternberg:
This was done somewhere else so this PR can be closed.

PR #4435, opened 2022-01-28
feat: compiler supports vectorized functions

scbrickley:
Related to #4185 

At the request of the reviewers, #4356 has been split into 2 PRs. One that adds the `Vector` type and makes changes to the `Value` interface (see #4427), and this one.

This branch was cut from the branch used for #4427 and contains some of the same commits. As a result, this patch is a little noisy, but this will have to wait on the other patch to merge anyway. Once that happens, I'll rebase this on master and it will be much easier to read.

This patch makes the following changes:
- Invokes the vectorization pass (see #4184) in the analyzer after type substitution has finished
- Restricts the type of functions that will become vectorized
- Teaches the compiler how to handle vector types
- Adds some tests around how the compiler handles vectors
- Maintains an accurate reference count for values generated during evaluation
- Adds a short doc that outlines how to maintain reference counts in the evaluator
- Adds a `Vector` method to the `Value` interface

I have some remaining questions that I'm hoping reviewers can help me with:
- There were sections in the compiler where it wasn't clear to me whether a value needed to be retained, released, or left alone. I've left comments in those places in the hopes of getting a second set of eyes on them.
- The document I wrote on maintaining reference counts in the compiler is short, and I'm not sure if it covers everything. I could use some input on that.
- I'm not confident that the tests I've added are sufficient to ensure that vectors are not copied unnecessarily. If someone has better insight into what's happening under the hood with arrow arrays, I would appreciate a crash course.

~#4427 needs to merge before this does.~


values/values.go
```diff
@@ -148,6 +151,16 @@ func (v value) Equal(r Value) bool {
...
 
+func (v value) Retain() {
+	n := New(v.v)
+	n.Retain()
+}
+
+func (v value) Release() {
+	n := New(v.v)
+	n.Release()
+}
```
scbrickley, line 162: `make checkdocs` seems to be complaining about these implementations of retain and release. Not sure why this would cause a panic, but I will investigate on Monday.

Marwes: `New` will return a `*value`/`Value` so the `Release/Retain` called will be the same methods that are defined here, ie it will just recurse into itself. I think you need to do a manual `switch` here and dispatch.

Marwes:
The manual reference counting here worries me, can't we use finalizers here? Call `Retain` once when the value is created and `Release` in the finalizer. I guess the drawback is that the memory for the arrow arrays may live longer than necessary but it is much easier to reason about.

If having predictable memory usage throughout a query is necessary we may want to seriously consider doing the interpreter/evaluator rewrite that was discussed a little while ago in rust so that we can ditch the garbage collector entirely and not have to deal with manual reference counting.

I need to take another, proper look at the reference counting later but the rest looks good

compiler/vectorized_test.go
```diff
@@ -0,0 +1,150 @@
...
+			if err != nil {
+				t.Fatalf("unexpected error: %s", err)
+			}
+
+			want := vectorizedObjectFromMap(tc.want, &memory.Allocator{})
+			if !cmp.Equal(want, got, CmpOptions...) {
+				t.Errorf("unexpected value -want/+got\n%s", cmp.Diff(want, got, CmpOptions...))
+			}
+
+			checked.AssertSize(t, 0)
```
wolffcm, line 147: This test would definitely benefit from the `debug` build tag that would catch too many calls to `Release`.

compiler/runtime.go
```diff
@@ -418,103 +449,112 @@ func (e *unaryEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, e
...
 
 func (e *regexpEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, error) {
-	return values.NewRegexp(e.r), nil
+	e.v.Retain()
+	return e.v, nil
 }
 
 type booleanEvaluator struct {
-	b bool
+	v values.Value
```
wolffcm, line 509: Seeing the change from a native type to an interface makes me wonder about any performance impact to non-vectorized functions. What's your take on this?

Also, do we need to call `Retain` here? This change doesn't seem necessary for the current limited case that is handled by vectorization. When we do handle something like `(r) => ({r with a: 10.0})` I would expect that the vectorization pass would insert a new operation `(r) => ({r with a: vector_from_scalar(10.0)})`. Meaning even in that case, we would never have a (say) `booleanEvaluator` that produces a vector that needs to be reference counted.

(We are calling `vector_from_scalar` something different in the issues that were filed. The name we landed on is escaping me at the moment.)

Thoughts?

wolffcm (commented):
This mostly matches my expectations, and looks good. I had some questions though.

wolffcm:
@Marwes I totally agree with you on the reference counting being disturbing. I think what is happening (@jsternberg could confirm I'm sure) is that Arrow is doing some of its own pooling under the hood---perhaps it's able to reuse memory buffers sooner than the Go runtime would get around to it.

It's always been mysterious to me that we need to reference count in a garbage collected language, really the worst of both worlds. I'd like to understand better why we do it.

Marwes:
For the functions we are factorizing right now we can be sure that the `r` (`Table`) do not escape the lambda so we could just ensure that `Retain/Release` is done in `map` itself and not worry about handling it inside the executor itself.

wolffcm:
> For the functions we are factorizing right now we can be sure that the r (Table) do not escape the lambda so we could just ensure that Retain/Release is done in map itself and not worry about handling it inside the executor itself.

@Marwes I'm not sure that's quite true, since for a lambda like `(r) => ({a: r.a, b: r.a})` we want both `a` and `b` to reference the same array in the output, so it should have a reference count of two, not one. The lambda is opaque to `map` so I don't see how it can manage the references, unless I'm missing something.

Marwes:
If the vector value is passed through transparently to the next step after `map` without copying then yes that is true. Then we need to keep track of accurate reference counts throught.

```
|> map((r) => ({a: r.a, b: r.a}))
|> map((r) => ({a: r.a, b: r.a})) // r.a / r.b refer to the same vector, no copying done
```

If not, and we always copy the data we need for the output of `map` then I think we can get away with only keeping a reference to the vector on the stack for the duration of the `map` call.

compiler/compiler.go
```diff
@@ -539,7 +555,7 @@ func compile(n semantic.Node, subst map[uint64]semantic.MonoType) (Evaluator, er
 		}, nil
 	case *semantic.TextPart:
 		return &textEvaluator{
-			value: n.Value,
+			value: values.NewString(n.Value),
```
jsternberg, line 558: Why these changes?

compiler/runtime.go
```diff
@@ -35,13 +35,17 @@ func (c compiledFn) Type() semantic.MonoType {
...
 func (c compiledFn) Eval(ctx context.Context, input values.Object) (values.Value, error) {
 	inputScope := nestScope(c.parentScope)
 	input.Range(func(k string, v values.Value) {
 		inputScope.Set(k, v)
+		v.Retain()
 	})
 
-	return eval(ctx, c.root, inputScope)
+	v, err := eval(ctx, c.root, inputScope)
+	v.Retain()
```
jsternberg, line 47: I think it's possible for this to panic if err is non-nil.

compiler/runtime.go
```diff
@@ -101,7 +105,9 @@ func (e *blockEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, e
 		if err != nil {
 			return nil, err
 		}
+		value.Release()
 	}
+	value.Retain()
```
jsternberg, line 110: I think this retain is wrong. Eval should return a value with a reference count of 1. We're the only one with a reference to it. Since we pass our reference back to the caller, we would both "retain" it to hand it to outside and "release" it internally which would just be a move operation and would not involve a retain.

The release above this looks correct. Although it looks strange. I do think it's correct.

compiler/runtime.go
```diff
@@ -119,13 +125,16 @@ func (e *declarationEvaluator) Type() semantic.MonoType {
...
 
+// TODO
 func (e *declarationEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, error) {
 	v, err := eval(ctx, e.init, scope)
 	if err != nil {
 		return nil, err
 	}
 
 	scope.Set(e.id, v)
+	v.Retain()
```
jsternberg, line 136: This one is correct because we're adding it to the scope.

compiler/runtime.go
```diff
@@ -137,32 +146,37 @@ func (e *stringExpressionEvaluator) Type() semantic.MonoType {
...
 
+// TODO
 func (e *stringExpressionEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, error) {
 	var b strings.Builder
 	for _, p := range e.parts {
 		v, err := p.Eval(ctx, scope)
 		if err != nil {
 			return nil, err
 		}
+		defer v.Release()
```
jsternberg, line 157: A defer inside of a for loop is usually wrong and I believe it's wrong here too. You either have to wrap this in a function or just manually call release. I think the second is probably better.

compiler/runtime.go
```diff
@@ -137,32 +146,37 @@ func (e *stringExpressionEvaluator) Type() semantic.MonoType {
...
 
 		if v.IsNull() {
 			return nil, errors.New(codes.Invalid, "string expression evaluated to null")
 		}
 		b.WriteString(v.Str())
 	}
-	return values.NewString(b.String()), nil
+	s := values.NewString(b.String())
```
jsternberg, line 164: `values.NewString` should return a value with a reference count of one so I think this retain is incorrect. Although, in reality, the retain is a no-op and does nothing.

compiler/runtime.go
```diff
@@ -178,7 +192,10 @@ func (e *interpolatedEvaluator) Eval(ctx context.Context, scope Scope) (values.V
 	if err != nil {
 		return nil, err
 	}
+	defer o.Release()
+
 	v, err := values.Stringify(o)
+	v.Retain()
```
jsternberg, line 198: I think this is an incorrect retain too.

jsternberg (changes requested):
I see one common mistake and I think it's in most places so I'm just going to make the comment generally and rereview when it's addressed.

When a value is created, it should default to a reference count of 1. For values that aren't really reference counted, like strings, we should still assume it is a reference count of 1.

I see a bunch of retains where it doesn't seem correct. Retain should only be in the following circumstances.

1. It is stored in the scope (possibly not?)
2. It is retrieved from the scope.
3. ??? I think that's it.

Release is a bit more complicated, but I don't think retain is needed anywhere else. This is because creating a value should set the reference count to one to begin with. When we return it, we are taking the value we own and transferring our ownership to the caller which is a move operation. A move operation is a retain followed by a release so we don't need to call either.

For adding the value to the scope, I think this might be a move operation too since it is a retain to store it in the scope followed by a release. When we access the scope, we are not taking ownership of the value from the scope so there's no transfer. We need to retain the value because we now have two references to it. We then release it from the scope when the scope becomes invalid.

The releases generally seem correct. I'll re-review them afterwards.

wolffcm:
@Marwes Are you suggesting that we do array copies in this PR to avoid reference counting in the interpreter? I can't tell.

I think we should avoid copying whenever possible, that's one of the big gains we get from vectorizing this simple case. In any case we'll need to do reference counting when we handle elementwise operations, so it would be good to get it right now.

Marwes:
> I think we should avoid copying whenever possible, that's one of the big gains we get from vectorizing this simple case. In any case we'll need to do reference counting when we handle elementwise operations, so it would be good to get it right now.

Right, nevermind then.

Still is there a reason why we would not just handle `Release` in a finalizer? If we are worried about arrow reusing the memory by mutating in place (which I don't think it does after digging around a bit) we can ensure that vector creation increments the reference count to 2 and the finalizer releases twice.

wolffcm:
> Still is there a reason why we would not just handle Release in a finalizer? If we are worried about arrow reusing the memory by mutating in place (which I don't think it does after digging around a bit) we can ensure that vector creation increments the reference count to 2 and the finalizer releases twice.

@Marwes We've reference counted arrays for as long as we've used Arrow, and I don't understand why beyond "Arrow pools memory," without much more detail. My understanding is that Go's runtime doesn't provide many guarantees about how finalizers are run, so it might not be a silver bullet here. @jsternberg maybe you can shed some more light on this?

But with us now actually using the `values.Value` interface backed by Arrow arrays in production code, `Retain` and `Release` are now leaking into many new places. The cognitive load of having to think about reference counts is significant, and if we can avoid it, I would like to. I think this moment is good for having this discussion---maybe we can talk about it tomorrow.

With that said, this code is consistent with how we're treating arrays in other places, and I think it would be too much to start using finalizers in this PR. So I don't want to block Sean on merging for a discussion like this.

jsternberg:
Sorry I did a review and forgot to see the other questions being asked.

> Arrow is doing some of its own pooling under the hood---perhaps it's able to reuse memory buffers sooner than the Go runtime would get around to it.

Arrow does not do any pooling of memory buffers. We could take advantage of the custom allocators and do our own pooling. At the moment, we only utilize it to handle our own accounting of what memory we understand the user to be using.

To clarify, proper retain/releases are _not_ required for correct operation of _our_ code. If we get this wrong, there is _no_ possibility of a memory leak in production. Go will clean up the memory when it decides to and nothing Arrow does stops that or even interferes with it.

The primary reason for the retain/release is so that we can keep track of what memory the user is using so that we can restrict them if they are processing too much data or otherwise causing an issue. While retain/release might be a bit ugly to see, we're effectively writing a VM so it makes sense we would handle reference counts for user memory in the VM.

A release in the finalizer wouldn't really further that goal because the memory that was allocated is owned by Go and the release really doesn't matter at all to it. Go will do what it wants when it wants to.

compiler/runtime.go
scbrickley: This confuses me. Each instance of `value` starts with a reference count of 1, and we release it at the end of the loop, so each value would have a reference count of 0 by the time the loop is done.

...but because we're returning the last instance of `value`, I would think we'd want to retain it to bring its reference count back up to 1. Am I missing something?

Maybe I should change the code to release all but the last instance of `value`?

jsternberg: This would be correct. Also, I didn't realize this the first time, but this order of operations is incorrect. You can't release a value and then retain it as the reference count would have potentially dropped to zero.

You might just want to reorder the loop so it's not assigning values that it will ultimately discard to the variable.

compiler/runtime.go
```diff
@@ -38,7 +38,11 @@ func (c compiledFn) Type() semantic.MonoType {
 func (c compiledFn) Eval(ctx context.Context, input values.Object) (values.Value, error) {
 	inputScope := nestScope(c.parentScope)
 	input.Range(func(k string, v values.Value) {
+		// Retaining `v` here because we're storing it in the scope,
```
jsternberg, line 41: This is a good point. I'm not sure this is needed because accessing the scope should retain the value already. If there's no accompanying release associated with this retain, it likely means we should not use retain here.

compiler/runtime.go
```diff
@@ -95,14 +99,19 @@ func (e *blockEvaluator) Type() semantic.MonoType {
 
 func (e *blockEvaluator) Eval(ctx context.Context, scope Scope) (values.Value, error) {
 	var err error
-	var value values.Value
-	for _, b := range e.body {
+	var v values.Value
```
jsternberg, line 101: I was more thinking to do the following:

```go
for _, b := range e.body[:len(e.body)-1] {
    value, err := eval(ctx, b, scope)
    if err != nil {
        return nil, err
    }
    value.Release()
}
return eval(ctx, e.body[len(e.body)-1], scope)
```

I believe it is an invariant the body will have at least one statement. If that's the case, this code should be safe.

compiler/runtime.go
```diff
@@ -698,6 +722,7 @@ func (f *functionValue) Call(ctx context.Context, args values.Object) (values.Va
 			a = v
 		}
 		scope.Set(p.Key, a)
+		a.Retain()
```
jsternberg, line 724: This retain may be unnecessary since we are evaluating the arguments and then immediately transferring ownership to the scope.

We do need to release the values we have included in the scope. I think we need to debug this as this may be another circumstance similar to the one above where retain is not needed. Contrary to the one above, I think release will be needed on the values we add to the scope because in this circumstance we're the owners rather than borrowing.

jsternberg (changes requested):
Reference counting is almost there I think.

compiler/vectorized_test.go
```diff
@@ -0,0 +1,149 @@
...
+			fn:           `(r) => ({r with c: "count"})`,
+			vectorizable: false,
+			skipComp:     true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mem := &memory.Allocator{}
+			checked := arrow.NewCheckedAllocator(mem)
```
wolffcm, line 108: This won't work quite the way you want it to, I think you need to do this:
```go
checked := arrow.NewCheckedAllocator(memory.DefaultAllocator)
mem := &memory.Allocator{Allocator: checked}
```
As it's written now, the checked allocator won't be used in the call to `AssertSize` below.

I would also expect that you need to call `Release` on the input and then also on the output record for these test cases to pass.

compiler/runtime.go
wolffcm: I'm trying to understand the example of `(r) => ({foo: r.a, bar: r.a})` works, since the array `r.a` that is in the input is referenced twice in the output, so I would expect that array to have a reference count of 2 once the input has been released. This seems important to avoid releasing too many times down the line.

So how does that happen? I see that in `memberEvaluator` we extract the array from the record here:
https://github.com/influxdata/flux/blob/0ee7a903a4daa146f62cd4c4f3adc34a28c33a62/compiler/runtime.go#L574
But I don't see how that increments the reference count.

Looking at where we create the output object in `objEvaluator` I also don't see how the reference count would be incremented there. So I'm confused on this point.

wolffcm (changes requested):
I am requesting some specific changes to how the test case creates an allocator and uses `AssertSize`. 

I also had lots of other questions about how references are counted. I'd like to feel pretty solid on this stuff before moving forward to other, cooler vectorization opportunities. But I think it will get easier once we have this sorted out.

compiler/runtime.go
jsternberg: This is a good catch. Both array evaluator and object evaluator need to retain the reference it retrieves.

scbrickley: Just so I'm clear on what's being requested, we want to:

- change the `Retain` call on line 724 to a `Release` call
- add `Retain` calls to the `Get` and `Set` methods of scopes, objects, and arrays

Do I have that right?

jsternberg: I don't think you change it to a release. I may be able to walk through each of these with you tomorrow though.

The retains should be in the evaluators not the underlying interface implementation.

jsternberg (approved):

wolffcm (approved):
This looks good to me, I appreciate your patience as we get this right. Even if we need to make tweaks again to how we track references, I'm okay with this merging as-is, so we can at least get a stake in the ground.

Issue #4436, opened 2022-01-28
Disable MergeFiltersRule rule for vanilla flux


onelson:
While investigating #4378 it was found that the issue was caused by the `MergeFiltersRule` combining the filters.

The first filter should have reduced the data set to the point of having a consistent type for the `_value` field, which was to be inspected in the 2nd filter. The gist is the filters were designed to limit the comparison within a specific series, but it ended up running against multiple series.

When the filters are combined, the function is run on a wider set of records than expected. Since type checking happens earlier than any of the short-circuiting we'd expect to save us from varied typing of the `_value` field, an error is produced.

Currently this rule is _only enabled_ for the flux cli - vanilla flux - so currently these filters will work as-is in other environments. In light of this, let's disable the rule here in flux so there's at least consistent behavior.

> It does seem desirable to be able to combine filters with a rule like this, but until we can get the type system to play nice, whatever benefits we'd see in perf are probably outweighed by the surprise of errors like these.

DOD:
- Remove this line, probably https://github.com/influxdata/flux/blob/master/cmd/flux/main.go#L18

Issue #4378, opened 2021-12-23
Sample data should be modified


LLThomas:
I find something wrong when execute a query below.
```
import "influxdata/influxdb/sample"

sample.data(set: "usgs")
  |> filter(fn: (r) => r._field == "mag" and r._measurement == "earthquake" and r.id == "av91449451")
  |> filter(fn: (r) => r._value > 0.0)
```

The error information is as follows.
```
Result: _result
runtime error @4:6-4:102: filter: cannot compile @ 4:17-4:101: unsupported binary expression string > float
Error: failed to execute query
```

However, when I check the raw data, I find _value type is float.
```
import "influxdata/influxdb/sample"

sample.data(set: "usgs")
  |> filter(fn: (r) => r._field == "mag" and r._measurement == "earthquake" and r.id == "av91449451" )

```
```
Result: _result
Table: keys: [_field, _measurement, code, id, magType, net, title]
         _field:string     _measurement:string             code:string               id:string          magType:string              net:string                               title:string                      _time:time                  _value:float  
----------------------  ----------------------  ----------------------  ----------------------  ----------------------  ----------------------  -----------------------------------------  ------------------------------  ----------------------------  
                   mag              earthquake                91449451              av91449451                      ml                      av  M 1.3 - 23 km WNW of Anchor Point, Alaska  2021-12-22T23:54:41.710000000Z                          1.25  

```
I check the sample data in https://github.com/influxdata/influxdb2-sample-data/blob/master/usgs-earthquake-data/all_week-annotated.csv and find that this issue caused by different datatype in that csv file.
![image](https://user-images.githubusercontent.com/49053546/147194168-10422b1a-d352-44c8-b83f-ca607290777d.png)

The top data will make query failed.
Should we delete it ?

Related to #4084

onelson:
Part of the challenge here is in how the sample data is prepared. We've got a github action that runs periodically. The tasks fetch new data from various sources, then run them through influxdb to produce the annotated csv files that get committed to the repo.

This is to say, we'd need to add something to the workflow to omit either the tables with floats or the ones with strings.

As it turns out, the data in the repo today has an almost even split of table headers:
```
influxdb2-sample-data on  master [?] via 🐍 v3.8.10 (venv) 
❯ grep 'dateTime:RFC3339,string' usgs-earthquake-data/all_week-annotated.csv | wc -l
6

influxdb2-sample-data on  master [?] via 🐍 v3.8.10 (venv) 
[...]

PR #4437, opened 2022-01-28
chore: disable MergeFiltersRule

onelson:
The rule is _only enabled_ here in the CLI, making the behavior
inconsistent with other envs.

The rule is specifically problematic because it can cause some queries
to fail even when written explicitly to avoid type check failures.

The rule could be re-enabled at a later date if we can figure out a way
to get the type checking to be permissive enough in this case.

Refs: #4378
Fixes: #4436

To verify the patch, I didn't add any new tests since it felt wrong to base test cases on a behavior determined by which rules are registered or not. Instead I checked the work in the REPL.

```
> import "influxdata/influxdb/sample"
>
> // Looking at _value in the first filter breaks (as expected) 
> sample.data(set: "usgs") |> filter(fn: (r) => r._field == "mag" and r._measurement == "earthquake" and r._value > 0.0) |>  group() |> count()
Result: _result
Error: runtime error @1:29-1:119: filter: cannot compile @ 1:40-1:118: unsupported binary expression string > float
> 
> // Looking at _value in a 2nd filter (now) works since it's not merged
> sample.data(set: "usgs") |> filter(fn: (r) => r._field == "mag" and r._measurement == "earthquake") |> filter(fn: (r) => r._value > 0.0) |>  group() |> count()

Result: _result
Table: keys: []
                _value:int  
--------------------------  
                      1912  
>
```

### Done checklist
- [ ] Test cases written

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -15,7 +15,6 @@ func main() {
 		influxdb.DefaultFromAttributes{
 			Host: func(v string) *string { return &v }(cmd.DefaultInfluxDBHost),
 		},
-		universe.MergeFiltersRule{},
 		universe.OptimizeWindowRule{},
 	)
 	cmd.Execute()


Marwes:
An alternative fix could be to change the binary function lookup to return the error lazily when it is evaluated. That way we still report the same type error but only if evaluation actually reaches that executor.

```go
		f, err := values.LookupBinaryFunction(values.BinaryFuncSignature{
			Operator: n.Operator,
			Left:     lt,
			Right:    rt,
		})
		if err != nil {
			f = func(l, r Value) (Value, error) {
                            return nil, err
			}
		}

```

Marwes:
Also, I think `MergeFilters` might construct the new `And` expression in the wrong order which also may need to be fixed to get the short circuiting right.

onelson:
I'll step through this in the debugger and see how it plays out in terms of pushing to getting the short-circuiting working today before moving on to other stuff.

onelson:
Seems like we want to disable the rule for now until we can decide on a proper fix to get the semantics we want. There's some debate among the team as to how to achieve the best/most consistent behavior in a case where all the predicates are in a single filter.

jsternberg (approved):

PR #4440, opened 2022-02-01
chore(upgrade_arrow): upgrade to apache arrow 7

skartikey:
The latest version of Arrow contains a fix that we need to account for memory allocated from Arrow Flight:
https://github.com/apache/arrow/pull/11712

Without this fix, we can pass an Allocator when retrieving data from IOx, but any arrays allocated to hold the result will not be reflected in the allocator.

Note that the latest version Arrow makes a change to the major version and this is seen as a completely separate module (to Go Modules):
https://github.com/apache/arrow/pull/11566/files#diff-340a0a599142950de3cef119fa68e93ef2350dbab23edaa3c7e561899e9ea399R17
(note the v7 suffix)

We will need to update the version of Arrow used with Flux, idpe, and the IOx client.

As part of the upgrade, fixed issue in tests replaced checked_allocator mem.AssertSize() with mem.CurrentAlloc() as former checks two things that is mem location is freed + expected allocation size.
In our case mem.CurrentAlloc is more suitable as it checks for expected allocation size only.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


Marwes (approved):

array/array.go
```diff
@@ -195,7 +195,7 @@ func Slice(arr Interface, i, j int) Interface {
 	if arr, ok := arr.(sliceable); ok {
 		return arr.Slice(i, j)
 	}
-	if arr, ok := arr.(array.Interface); ok {
+	if arr, ok := arr.(arrow.Array); ok {
```
wolffcm, line 198: So did arrow move around and rename types? I guess they had a complete breaking change from the older version,

array/array_test.go
```diff
@@ -88,11 +89,11 @@ func TestString(t *testing.T) {
 			if want, got := len(tc.want)+2, b.Cap(); want != got {
 				t.Errorf("unexpected builder cap -want/+got:\n\t- %d\n\t+ %d", want, got)
 			}
-			mem.AssertSize(t, tc.sz)
+			assert.Equal(t, tc.sz, mem.CurrentAlloc(), "unexpected memory allocation.")
```
wolffcm, line 92: @jsternberg FYI, this is the change in Arrow's checked allocator that I was talking about. `AssertSize` now contains a check that ensures that there are no outstanding allocations, and also that the numeric argument matches the cumulative amount of memory allocated. It doesn't seem like it should do that based on the name, so this was a surprising change.

execute/aggregate_test.go
```diff
@@ -87,7 +88,7 @@ func TestAggregateTransformation_ProcessChunk(t *testing.T) {
 		t.Fatal("expected aggregate function to be invoked, but it was not")
 	}
 	// Memory should have been released since we did not retain the data.
-	mem.AssertSize(t, 0)
+	assert.Equal(t, 0, mem.CurrentAlloc(), "unexpected memory allocation.")
```
wolffcm, line 91: Did you need to make this change? I would expect anything where we're checking for `0` like in `AssertSize(t, 0)` to still pass, since there should be no outstanding allocations. Is it possible that this was actually finding a bug?

wolffcm (changes requested):
We can't merge this yet, since downstream is not compatible with the new version of gRPC. But I still have some questions that are worth looking into.

array/array_test.go
jsternberg: I think this indicates an error. I'll download the PR and take a look. It does not seem like `AssertSize` was changed based on the code.

I'm basing it off of a quick reading of these sections:

https://github.com/apache/arrow/blob/56e270fda7f5/go/arrow/memory/checked_allocator.go#L131-L134
https://github.com/apache/arrow/blob/56e270fda7f5/go/arrow/memory/checked_allocator.go#L42
https://github.com/apache/arrow/blob/56e270fda7f5/go/arrow/memory/checked_allocator.go#L73

jsternberg: Ah, so `AssertSize` still checks the current allocations (not total) but it also seems to only work when the argument is zero since if the argument wasn't zero there should be "leaked" allocations. But, they're not leaked.

array/array.go
skartikey: Interface aliases arrow.Array so that existing users don't get broken by the migration to arrow.Array. 
This alias will be removed in v8

https://pkg.go.dev/github.com/apache/arrow/go/v7@v7.0.0-20220201061922-bf0ee3f9c50e/arrow/array#Interface

wolffcm: We have an interface defined at line 27 in this file that is named after the corresponding Arrow type. Can we change the name of that to `Array` as well? That is likely to cause some breaking changes in our Go API, but there are breaking changes anyway.

wolffcm (changes requested):
This looks great, can you investigate whether we need the `Interface` type anymore? If we still need it, it can be renamed to `Array` to match what Arrow calls it.

array/array.go
skartikey: I did change the name of the interface to `Array` to match the Arrow array Interface.

Regarding removing our `Interface` completely, the current [Arrow Array interface](https://github.com/apache/arrow/blob/master/go/arrow/array.go#L83) has 2 extra members `json.Marshaler` & `Data() ArrayData` so these needs to be implemented as well at our end of implementations. I have not analyzed the code to see how much work is needed for that. If that is something simple to do I can do that in this PR itself or we can open a separate issue for that.

skartikey: Sure will do. Thanks.

wolffcm (approved):
Looks great, thanks for doing this.

PR #4441, opened 2022-02-01
fix(feature): int feature flags work properly when returned as floats

jsternberg:
The int type for feature flags has been switched from int32 to int since
it is more consistent. Additionally, if a float value is returned and is
meant to be interpreted as an int, the conversion now happens. This can
sometimes happen because feature flag libraries may transfer the data as
JSON and JSON doesn't have an integer type instead treating them as
floats.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


Marwes (approved):

adrian-thurston (approved):
Verified this allows int flags to work in the context of IDPE.

Issue #4312, opened 2021-12-02
Support for some timezones is broken


jkreileder:
We're using the rather new timezone support in Influx.  For at least one timezone this seems to be broken. With

```
import "timezone"
option location = timezone.location(name: "America/Phoenix")
```

queries don't return anything at all.

A bit of debugging in Flux shows that the problem probably is with `tzset()` returning `omega` both for `start` and `end`. The consequence is that we fall into the first truncate case in `Location.ToLocalClock()`.

Here's a test case:

```
diff --git a/interval/window_test.go b/interval/window_test.go
index 3f3a26b9..02f9209c 100644
--- a/interval/window_test.go
+++ b/interval/window_test.go
@@ -536,9 +536,11 @@ func TestWindow_GetLatestBounds(t *testing.T) {
 
 func TestWindow_GetLatestBounds_InLocation(t *testing.T) {
 	const (
-		US_Pacific     = "America/Los_Angeles"
-		Australia_East = "Australia/Sydney"
-		American_Samoa = "Pacific/Apia"
+		America_Phoenix = "America/Phoenix"
+		America_Denver  = "America/Denver"
+		US_Pacific      = "America/Los_Angeles"
+		Australia_East  = "Australia/Sydney"
+		American_Samoa  = "Pacific/Apia"
 	)
 
 	type window struct {
@@ -554,6 +556,58 @@ func TestWindow_GetLatestBounds_InLocation(t *testing.T) {
 		t    string
 		want [2]string
 	}{
+		{
+			name: "America_Phoenix",
+			loc:  America_Phoenix,
+			w: window{
+				every:  "1d",
+				period: "1d",
+			},
+			t: "2017-02-24T12:00:00-07:00",
+			want: [2]string{
+				"2017-02-24T00:00:00-07:00",
+				"2017-02-25T00:00:00-07:00",
+			},
+		},
+		{
+			name: "America_Phoenix DST", // Phoenix doesn't observer DST
+			loc:  America_Phoenix,
+			w: window{
+				every:  "1d",
+				period: "1d",
+			},
+			t: "2017-09-03T12:00:00-07:00",
+			want: [2]string{
+				"2017-09-03T00:00:00-07:00",
+				"2017-09-04T00:00:00-07:00",
+			},
+		},
+		{
+			name: "America_Denver",
+			loc:  America_Denver,
+			w: window{
+				every:  "1d",
+				period: "1d",
+			},
+			t: "2017-02-24T12:00:00-07:00",
[...]

PR #4444, opened 2022-02-03
fix(timezone): timezone is not working for 'no daylight savings time'

skartikey:
This is a draft PR as it's still in progress. 

**Issue:** 
The timezone is not working for 'No daylight savings time'.

In case of no daylight saving time for a timezone, we are hitting the end of the known zone transitions which try to use extend string here.
https://github.com/influxdata/flux/blob/bc37ddcd7f07676f6372e934999252a52b1af98c/internal/zoneinfo/zoneinfo.go#L166-L172

the problem is with [tzset()](https://github.com/influxdata/flux/blob/bc37ddcd7f07676f6372e934999252a52b1af98c/internal/zoneinfo/zoneinfo.go#L235), in case of no daylight savings timezones, tzset() is returning omega both for start and end ( see below code) . The consequence is that we fall into the first truncate case in Location.ToLocalClock() which ultimately fails.

https://github.com/influxdata/flux/blob/bc37ddcd7f07676f6372e934999252a52b1af98c/internal/zoneinfo/zoneinfo.go#L254-L257

**Fix:** 
The fix is to pass the `start` boundary in the tzset() method and return it when the code reaches no daylight saving timezone case in the method.  

Tests have been added to reproduce the issue.

fixes: https://github.com/influxdata/flux/issues/4312

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


skartikey:
@jsternberg I have made this PR a draft as the code changes which are done here to fix the issue are in the copied Go library code. Kindly have a look at the test cases and we can discuss a proper fix once you are available.

skartikey:
@jsternberg Test cases for the `ToLocalClock` are already added in this PR. 

Another point to note is when start = end = omega this [call](https://github.com/influxdata/flux/blob/b5d9d2da1c7146adbad82940fd5af781a1dbb773/internal/zoneinfo/zoneinfo_clock.go#L20) returns start and end value in a nanosecond. because of this below switch case will always fall in the first case. Also, call to next lookup function will have param with nanosecond minus 1.

jsternberg (approved):
Rebase on master but after looking at this I'm pretty convinced this is a correct solution. I find that unfortunate since I'd rather not modify this part of the code, but I don't see anywhere else to modify it that wouldn't be an equivalent change.

PR #4446, opened 2022-02-03
feat(requests): adds requests.peek for easy inspection of HTTP response

### LINKED ISSUE #4452: incude responseTime in the response object of the new http library
regarding the new library explained here: https://github.com/influxdata/flux/blob/master/stdlib/experimental/http/requests/requests.flux

In addition to statusCode, could we add the response time or latency? It would be useful to know the latency of remote call, especially when debugging slow scripts, but I can think of other uses as well.

nathanielc:
This change adds a new peek method to the requests package to make
viewing the results of an HTTP request trivial.

Additionally the examples are updated to actually consume the response of
the HTTP requests and show how to use array.from to construct tables.

The examples use real publicly and freely available APIs and can be
executed. As a result the examples are executed as part of building the
Flux docs and the results are captured.

Fixes #4452 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


nathanielc:
@sanderson This PR updates the examples a bit but all the API examples are all hypothetical. Having a real executable API would be much better example. Do you know of something like that we could add to improve these examples? 

Maybe make the examples hit our cloud API?

nathanielc:
I was sent a pointer to this https://mixedanalytics.com/blog/list-actually-free-open-no-auth-needed-apis/ has some great APIs. Going to try a few of these out.

stdlib/experimental/http/requests/requests.flux
```diff
@@ -159,6 +168,39 @@ do = (
 // resp = requests.post(url:"http://example.com", body: json.encode(v: {data: {x:1, y: 2, z:3}))
 // ```
 //
+// // ### Make a POST request with a JSON body and decode JSON response
```
sanderson, line 171:
```suggestion
// ### Make a POST request with a JSON body and decode the JSON response
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -159,6 +168,39 @@ do = (
...
+// )
+//
+// // Assuming the response body contained the following JSON:
+// //
+// // {
+// //  "color": "#456789",
+// //  "size": 10
+// // }
+// //
+// // The data variable will be a record with the same properties.
```
sanderson, line 193:
```suggestion
// // Define a data variable that parses the JSON response body into a Flux record.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -159,6 +168,39 @@ do = (
...
+// data = json.parse(data: response.body)
+//
+// // Using array.from we can construct a table of data we decoded from
+// // the HTTP POST request.
```
sanderson, line 197:
```suggestion
// // Use array.from() to construct a table of data decoded from the
// // HTTP POST request.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -192,7 +234,38 @@ post = (
 // ```no_run
 // import "experimental/http/requests"
 //
-// resp = requests.get(url:"http://example.com")
+// response = requests.get(url:"http://example.com")
+//
+// requests.peek(response: response)
+// ```
+//
+// ### Make a GET request and decode response using JSON
```
sanderson, line 242:
```suggestion
// ### Make a GET request and decode the JSON response
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -159,6 +168,39 @@ do = (
...
+//
+// body = json.encode(v:{
+//     "size": ">=8",
+// })
+// response = requests.post(
+//     url: "http://example.com",
+//     body: body,
+// )
+//
+// // Assuming the response body contained the following JSON:
```
sanderson, line 186:
```suggestion
// // Assume the response body contains the following JSON:
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -192,7 +234,38 @@ post = (
...
+// ```no_run
+// import "experimental/http/requests"
+// import "experimental/json"
+//
+// response = requests.get(
+//     url: "http://example.com",
+//     params: ["start": ["100"]],
+// )
+//
+// // Assuming the response body contained the following JSON:
```
sanderson, line 253:
```suggestion
// // Assume the response body contains the following JSON:
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -192,7 +234,38 @@ post = (
...
+// //
+// // {
+// //  "color": "#456789",
+// //  "size": 10
+// // }
+// //
+// // The data variable will be a record with the same properties.
```
sanderson, line 260:
```suggestion
// // Define a data variable that parses the JSON response body into a Flux record.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -192,7 +234,38 @@ post = (
...
+// data = json.parse(data: response.body)
+//
+// // Using array.from we can construct a table of data we decoded from
+// // the HTTP GET request.
```
sanderson, line 264:
```suggestion
// // Use array.from() to construct a table of data decoded from the
// // HTTP GET request.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +284,28 @@ get = (
         body: body,
         config: config,
     )
+
+// peek converts a response into a table for easy inspection.
+// The response body will be returned as a string column on the table.
+// The status code is returned as an integer column on the table.
```
sanderson, line 290:
```suggestion
// peek converts an HTTP response into a table for easy inspection.
// The output table includes a **body** column with the response
// body as a string and a **statusCode** column with the returned
// status code as an integer.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +284,28 @@ get = (
...
+//
+// Flux queries must produce a table this method makes that conversion easy.
+// If however you want control over how the table is constructed use `array.from`
+// in combination with a function like `json.parse`. The basic process is parse the
+// response body into a set of values and then use `array.from` to construct a 
+// table from those values. See the examples for `requests.post` and `request.get` for
+// a more complete explanation.
+//
+// ## Parameters:
+//  response: Result of making an HTTP request.
```
sanderson, line 300:
```suggestion
//  - response: Response data from an HTTP request.
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +284,28 @@ get = (
...
+//
+//  ## Examples
+//
+//  ## Using peek to quickly inspect a response
```
sanderson, line 304:
```suggestion
//  ### Inspect the response of an HTTP request
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +284,28 @@ get = (
...
+//
+//  ```no_run
+//  import "experimental/http/request"
+//
+//  request.peek(response: http.post(url:"http://example.com"))
```
sanderson, line 309:
```suggestion
//  import "experimental/http/requests"
//
//  requests.peek(response: http.post(url:"http://example.com"))
```

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +284,28 @@ get = (
...
+// peek converts a response into a table for easy inspection.
+// The response body will be returned as a string column on the table.
+// The status code is returned as an integer column on the table.
+//
+// Flux queries must produce a table this method makes that conversion easy.
+// If however you want control over how the table is constructed use `array.from`
+// in combination with a function like `json.parse`. The basic process is parse the
+// response body into a set of values and then use `array.from` to construct a 
+// table from those values. See the examples for `requests.post` and `request.get` for
+// a more complete explanation.
```
sanderson, line 297:
```suggestion
// To customize how the response data is structured in a table, use `array.from()`
// with a function like `json.parse()`. Parse the response body into a set of values
// and then use `array.from()` to construct a table from those values.
```

sanderson: I'm wondering if we should even include this content or at least move it elsewhere. It's essentially telling a user how to not use `requests.peek`.

sanderson (changes requested):

stdlib/experimental/http/requests/requests.flux
nathanielc: Yeah my thinking was the other examples have taught users how to use `peek` but they will need to learn how to move beyond using just `peek` as its not that useful beyond debugging. Happy to move this around as needed.

sanderson: It's probably fine to leave it here.

sanderson (approved):
👌🏻

Marwes (approved):

stdlib/experimental/http/requests/requests.flux
```diff
@@ -211,3 +268,30 @@ get = (
...
+//
+// ```
+// import "experimental/http/requests"
+//
+// requests.peek(response: requests.get(
+//     url: "https://api.agify.io",
+//     params: ["name": ["natalie"]],
+// > ))
+// ```
+peek = (response) => array.from(rows: [{statusCode: response.statusCode, body: string(v: response.body)}])
```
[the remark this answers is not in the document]
nathanielc: Wait, why is adding a new property to a record a breaking change? Won't that just be an additive change? If they don't use the field then its not an issue?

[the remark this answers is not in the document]
Marwes: Consider

```flux
if x then
    request.peek(response)
else
    array.from(rows: [{statusCode: 123, body: "" }])
```

this would start failing typechecks as the two branches do not return the same type if `peek` added more fields

nathanielc:
@Marwes Thanks for the heads up on the return type. There were already a few requests to add more to peek. So I have done that already. The `headers` and `duration` columns have been added. Could I get another review?

libflux/flux-core/src/doc/example.rs
```diff
@@ -132,7 +132,11 @@ fn block_mode(kind: CodeBlockKind) -> Option<BlockMode> {
 //    '# ' - hide the line in the display
 //    '< ' - mark the line as the input
 //    '> ' - mark the line as the output
+//    '#<' - mark the line as the input and hide
+//    '#>' - mark the line as the output and hide
```
nathanielc, line 136: @sanderson FYI I added this feature to fluxdoc generation.

Marwes (approved):

libflux/flux-core/src/doc/example.rs
sanderson: Awesome, do you want to add these here: https://github.com/influxdata/flux/blob/master/docs/fluxdoc.md#example-execution-input-and-output

I can open another PR and add them as well.

PR #4455, opened 2022-02-07
feat: connect lsp feature through the flux crate

rockstar:
The `lsp` feature was added previously to the `flux-core` crate, but it
wasn't exposed via `flux`. This patch exposes that, and also adds one
more `Into` implementation to simplify `flux-lsp` a bit.

### DIFF
--- a/libflux/flux-core/src/ast/mod.rs
+++ b/libflux/flux-core/src/ast/mod.rs
@@ -67,11 +67,11 @@ impl Default for Position {
 /// Convert a Position to a lsp_types::Position
 /// https://microsoft.github.io/language-server-protocol/specification#position
 #[cfg(feature = "lsp")]
-impl Into<lsp_types::Position> for Position {
-    fn into(self) -> lsp_types::Position {
-        lsp_types::Position {
-            line: self.line - 1,
-            character: self.column - 1,
+impl From<Position> for lsp_types::Position {
+    fn from(position: Position) -> Self {
+        Self {
+            line: position.line - 1,
+            character: position.column - 1,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


libflux/flux-core/src/ast/mod.rs
```diff
@@ -102,6 +102,16 @@ impl SourceLocation {
     }
 }
 
+#[cfg(feature = "lsp")]
+impl Into<lsp_types::Range> for SourceLocation {
```
Marwes, line 106: You can implement `From<SourceLocation> for lsp_types::Range` instead and you get `Into` for free, same for the existing `Position` impl.

Marwes (approved):

libflux/flux-core/src/ast/mod.rs
rockstar: Does it not work both ways? If I implement `Into`, do I not get `From` for free? If not, we should definitely change this.

skartikey (approved):

libflux/flux-core/src/ast/mod.rs
Marwes: No, there is a blanket `impl<T> Into<T> for T where T: From<T>` but there can't be one the other way around as it would overlap.

rockstar: Excellent, thanks for catching that. I'd rather implement `From` across the board. Not sure what I was thinking with the original `Into`, as I have never implemented that before. Updated both `Into` implementations into `From` implementations.

Marwes (approved):
You will have to fix the type signature/functions though since `From` is different from `Into`

PR #4456, opened 2022-02-07
fix: Set the correct type for the vectorized function

Marwes:


jsternberg (approved):
The test case looks correct to me for what I would expect.

PR #4460, opened 2022-02-08
feat: add display function to make representing any value simple

### LINKED ISSUE #4454: everything, but including dictionary, should have a string representation
When debugging Flux, it can be difficult or impossible to see values that are passed from other API calls. For example, the headers of a response object in the http library are a dictionary.

It would be useful if you could pass in any object to array.from for debugging. For example:

```flux
array.from(rows: [{"headers":string(v: response.headers}])
|> yield(name: "print headers")
```

nathanielc:
Prior to this change it was not possible to convert every Flux value
into a string for easy debugging. With this change a `display` function
has been added that converts any Flux value into a string
representation.

The display function uses the same code that the REPL uses to display
the results of executing a line.

The bytes type display was changed to be a hexidecimal string instead of
a list of decimal values.

Fixes #4454 

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/universe.flux
```diff
@@ -237,6 +237,80 @@ builtin string : (v: A) => string
...
+//
+//  // Returns
+//  // {
+//  //    array: [1, 2, 3],
+//  //    bytes: 0x616263,
+//  //    dict: [a: 1, b: 2, c: 3],
+//  //    string: str
+//  // }
+//  ```
+builtin display : (v: A) => string
```
jsternberg, line 312: Should we use the name format instead?

jsternberg (approved):

stdlib/universe/universe.flux
nathanielc: The name `format` implies some kind of control over how the value is represented and that is explicitly not the intent of this function. This function is opinionated and can change but it always gives a string representation that is human readable.

nathanielc:
@sanderson This is something we should teach users early.

stdlib/universe/universe.flux
```diff
@@ -237,6 +237,80 @@ builtin string : (v: A) => string
...
+// to a string recusively for all composite types.
+//
+//  Basic types are converted to a string directly.
+//  Bytes are represented as a string of lowercase hexadecimal characters prefixed with `0x`.
+//  Composite types (i.e. arrays, dictionaries, records) will be represented in a syntax similar
+//  to the equivalent Flux literal representation.
+//
+//  It is not expected that the representation can be parsed back into the original value.
+//  The representation may span multiple lines.
+//  The representation is not stable across Flux versions.
```
sanderson, line 251:
```suggestion
// display returns the Flux literal representation of any value as a string.
//
// Basic types are converted directly to a string.
// Bytes types are represented as a string of lowercase hexadecimal characters prefixed with `0x`.
// Composite types (arrays, dictionaries, and records) are represented in a syntax similar
// to their equivalent Flux literal representation.
//
// Note the following about the resulting string representation:
// - It cannot always be parsed back into the original value.
// - It may span multiple lines.
// - It may change between Flux versions.
//
// `display()` differs from `string()` in that `string()` recursively converts values inside
// composite types to strings.
```

sanderson (changes requested):
Some suggestions and thoughts.

stdlib/universe/universe.flux
```diff
@@ -237,6 +237,80 @@ builtin string : (v: A) => string
...
+//  display(v: x)
+//
+//  // Returns
+//  // {
+//  //    array: [1, 2, 3],
+//  //    bytes: 0x616263,
+//  //    dict: [a: 1, b: 2, c: 3],
+//  //    string: str
+//  // }
+//  ```
```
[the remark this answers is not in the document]
rickspencer3: maybe it should include something like:
```
|> yield(name: "display")
```

so users know how to actually view the values?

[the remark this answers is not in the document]
sanderson: Yields are implicit unless you have multiple. This example as-is will return a table with the string representations of composite types.

stdlib/universe/universe.flux
```diff
@@ -237,6 +237,97 @@ builtin string : (v: A) => string
...
+// composite types to strings. `string()` does not operate on composite types.
+//
+//  ## Parameters
+//  - v: Value to convert for display.
+//
+//  ## Examples
+//
+//  ### Display a value as part of a table
+//
+//  Using `array.from()` and `display() you can quickly observe any value.
```
sanderson, line 262:
```suggestion
//  Use `array.from()` and `display()` to quickly observe any value.
```

sanderson (approved):
One minor change and then it's good to go 👍🏻

stdlib/universe/universe.flux
nathanielc: Thanks I'll figure out how to do this correctly someday.

Issue #4451, opened 2022-02-05
Really need to improve the type mismatch error messages


rickspencer3:
For example, if I try to pass a dictionary into ```array.from(rows: [some_dict])```

The error I get back is:
``` 
error @13:18-13:30: expected [A] but found [string:string] (argument rows)
```

Could we not tell the user what the allowable types are, and that the wrong type was? Also, maybe suggest some casting if it is available?

Marwes:
cc https://github.com/influxdata/flux/issues/4264

Issue #4275, opened 2021-11-16
Figure out how to represent async things in the type system


onelson:
Today the typing in flux is such that "array" conflates concrete collections as well as streams (yet-to-be realized collections). This means that certain functions will fail if they actually care about a value being one or the other.

Some approaches discussed briefly (to be looked at in more depth) include:

- what if there was an implicit _await_ in functions that expect realized collections but receive a stream?
- what if there was a separate type for a _Future/Promise/Stream_ of some other type?

onelson:
See also: #4275

PR #4467, opened 2022-02-10
feat: Explain what [A], [A:B] etc means in errors

Marwes:
This should address part of https://github.com/influxdata/flux/issues/4451, by adding a name for type instead of relying on the user knowing what the symbols mean (easier to google a name, if nothing else).

It is still a bit confusing that arrays and streams of tables are both "array", but that is a different change (https://github.com/influxdata/flux/issues/4275)

cc @rickspencer3


Marwes:
See the two earlier commits for why I change the tests.

rockstar (approved):
Wooooow, much improved! Thanks for this work.

PR #4468, opened 2022-02-10
feat: Separate streams from arrays in the type system

Marwes:
Arrays and streams of tables are both represented by `[A]` in the type system, however almost all functions that specify `[A]` actually mean that they take or return a stream. Since all functions (that I am aware of) only work with one or the other we should separate the two concepts to make it type error to pass in the wrong type of value.

This uses the straw man `stream[MonoType]` to specify streams which should be relatively simple to change to something else before we merge.

We may also want add this under a feature flag which maps `stream` to `array` before typechecking to reduce the risk of breakages.

cc https://github.com/influxdata/flux/issues/4275


scbrickley:
This looks good overall, but I think the failing CI tests are related to your your changes. I'll approve once I see those turn green.

PR #4470, opened 2022-02-10
fix(schema): add stop arg to schema functions

nathanielc:
Previously the `stop` argument could not be added to the schema
functions because the default value would overly constrain the arguments
type. That has been fixed.

Note sure how to add tests for this change as we don't actually have
tests for calling the functions rather just for the push downs matching.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
@jsternberg Thoughts on how I could testthis?

stdlib/influxdata/influxdb/schema/schema.flux
```diff
@@ -240,4 +268,4 @@ measurementFieldKeys = (bucket, measurement, start=-30d) =>
 //
 // tags: metadata
 //
-measurements = (bucket) => tagValues(bucket: bucket, tag: "_measurement")
+measurements = (bucket, start=_startDefault, stop=_stopDefault) => tagValues(bucket: bucket, tag: "_measurement", start: start ,stop: stop)
```
jsternberg, line 271: A type in the signature here.

jsternberg (approved):

nathanielc:
We have tests but they just mirror the function bodies instead of calling the functions directly. This is to get around the fact that the functions call `from()` directly.

Marwes:
I guess we could add `from` as a default argument to the functions, and pass in a test `from` function for testing. Or add private variants of all functions under test that accepts `from` as a parameter, then invoke those from the current

```
tagValues = (
    bucket,
    tag,
    predicate=(r) => true,
    start=_startDefault,
    stop=_stopDefault,
)  => tagValuesInner(bucket, tag, predicate, start, stop, from)

tagValuesInner = (
    bucket,
    tag,
    predicate=(r) => true,
    start=_startDefault,
    stop=_stopDefault,
    from,
) =>
    from(bucket: bucket)
        |> filter(fn: predicate)
        |> keep(columns: [tag])
        |> group()
|> distinct(column: tag)
```

nathanielc:
@Marwes Thanks for the idea, it decided to implement something very similar by using an `option`. 

As a result I did a complete rewrite of the tests but they are much easier to understand now.

nathanielc:
Also I have run these tests against InfluxDB OSS and they pass.

noramullen1 (approved):

Marwes (approved):

PR #4473, opened 2022-02-10
chore(release): add prep-release script

nathanielc:
The Flux release process will now require a preparation step.
This still will make and commit any changes to the source code, i.e.
updating docs to mark the correct version a function was introduced.

The process will be:

1. Run ./prep-release.sh in Flux repo
2. Merge created PR
3. Run IDPE release script (which calls the Flux release script)

The Flux release script checks that the prep step was completed so it
can't be easily forgotten.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


libflux/flux/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "flux"
-version = "0.5.1"
-authors = ["jlapacik <joshua.lapacik@pacbell.net>"]
+version = "0.154.0"
```
nathanielc, line 3: I'd like to add a step to automatically bump the version in the Cargo file too. But the `cargo-bump` utility we used doesn't yet support workspace based projects. See https://github.com/wraithan/cargo-bump/blob/master/src/config.rs#L102 

We can add that behavior of the release later.

docs/fluxdoc.md
```diff
@@ -43,6 +43,9 @@ Each key-value pair must be on a single line.
   and categorize packages. _See [Metadata tags](#metadata-tags)._
 - **contributors**: Contributor GitHub usernames or other contact information.
 
+When adding a new export to a Flux package the `introduced: NEXT` metadata can be added to the value's docs.
+The release process will automatically replace the `NEXT` with the version of the Flux release.
+
```
nathanielc, line 48: @sanderson FYI we are looking to add this to our release process.

libflux/flux/Cargo.toml
rockstar: Yeah, this is one of those things that has always bothered me.

There are a couple of cargo tools that don't with workspaces that I use regularly. It makes me sad.

rockstar (commented):
> Run IDPE release script (which calls the Flux release script)

I wonder if we should also run `etc/checkprepared.sh` in the idpe script.

I'd also like to see a script that is the opposite of `etc/checkprepared.sh` so that we can assert that both scripts are only being run in their required states. This might seem like over-engineering, but footguns in the release process are guaranteed to go off more than not.

libflux/flux/Cargo.toml
nathanielc: Submitted PR to cargo-bump for workspace support https://github.com/wraithan/cargo-bump/pull/34

etc/checkprepared.sh
```diff
@@ -0,0 +1,30 @@
...
+do
+    # Check for any introduced: NEXT comments that still exist
+    grep "^//[[:space:]]*introduced:[[:space:]]\+NEXT[[:space:]]*$" $f > /dev/null
+    ret=$?
+    if [ $ret -eq 0 ]
+    then
+        EXIT=1
+        echo "$f contains 'introduced: NEXT'"
+    fi
+done < <(find ./stdlib -name '*.flux')
```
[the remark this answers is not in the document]
nathanielc: I wanted to do it this way so that if we get spaces in file names it doesn't break.

nathanielc:
> I wonder if we should also run etc/checkprepared.sh in the idpe script.

I'd rather IDPE know as little as possible about the Flux release process. So that is why the Flux release script calls the checkprepared.sh script itself. 

> I'd also like to see a script that is the opposite of etc/checkprepared.sh so that we can assert that both scripts are only being run in their required states. This might seem like over-engineering, but footguns in the release process are guaranteed to go off more than not.

I am not sure I follow. Like a checkunprepared script? What would it do? Assert that `introduced: NEXT` is present? It not always going to be present even in an unprepared repo.

rockstar:
> I'd rather IDPE know as little as possible about the Flux release process. So that is why the Flux release script calls the checkprepared.sh script itself.

I don't think it would be a _horrible_ thing if the IDPE script knew _something_ about our release process (because it is part of the release process, ultimately), but I take your point that if we can avoid it, we should.

> I am not sure I follow. Like a checkunprepared script? What would it do? Assert that introduced: NEXT is present? It not always going to be present even in an unprepared repo.

I _cannot_, for the life of me, remember what I was trying to say here. I think the thing I'm concerned about is that the state of "unprepared" becomes more complex, and maybe the check is about checking versions and tags, etc. I'm not sure that's necessary, at this point. If it gets more complicated, we should certainly think about it, but for now, I retract my statement. :)

rockstar (approved):

PR #4474, opened 2022-02-10
fix(test): remove `os.Exit` calls, allow defer `executor.Close` to run

onelson:
Today, `TestExecutor`s that depend on being able to run code on teardown by
implementing `Close()` are unable to since `os.Exit()` will terminate the
process before any deferred functions are able to run.

The exiting, if deemed necessary, should probably be the responsibility
of the actual binary rather than deep in the test runner.

While it would be preferred to simply hand errors back to the caller,
this diff avoids updating the signatures of public functions by
converting the previous `Exit` calls into `panic`s. The benefit of this
is a panic will still allow deferred functions to be called as the stack
unwinds.

~~In the future, we may elect to clean this up further by refactoring public
functions like `Finish()` to return errors.~~ The future is _now_. This has been updated to return an error instead of panicking. 

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/cmd/flux/cmd/test.go
+++ b/cmd/flux/cmd/test.go
@@ -46,7 +46,10 @@ func TestCommand(setup TestSetupFunc) *cobra.Command {
 		Long:  "Run flux tests",
 		Run: func(cmd *cobra.Command, args []string) {
 			fluxinit.FluxInit()
-			runFluxTests(setup, flags)
+			if err := runFluxTests(setup, flags); err != nil {
+				fmt.Println(err)
+				os.Exit(1)
+			}
 		},
 	}
 	testCommand.Flags().StringSliceVarP(&flags.paths, "path", "p", nil, "The root level directory for all packages.")
@@ -62,7 +65,7 @@ func init() {
 }
 
 // runFluxTests invokes the test runner.
-func runFluxTests(setup TestSetupFunc, flags testFlags) {
+func runFluxTests(setup TestSetupFunc, flags testFlags) error {
 	if len(flags.paths) == 0 {
 		flags.paths = []string{"."}
 	}
@@ -70,19 +73,17 @@ func runFluxTests(setup TestSetupFunc, flags testFlags) {
 	reporter := NewTestReporter(flags.verbosity)
 	runner := NewTestRunner(reporter)
 	if err := runner.Gather(flags.paths, flags.testNames); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		return err
 	}
 
 	executor, err := setup(context.Background())
 	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		return err
 	}
 	defer func() { _ = executor.Close() }()
 
 	runner.Run(executor, flags.verbosity, flags.skipTestCases)
-	runner.Finish()
+	return runner.Finish()
 }
 
 // Test wraps the functionality of a single testcase statement,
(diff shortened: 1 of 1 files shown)


cmd/flux/cmd/test.go
```diff
@@ -547,11 +548,13 @@ func (t *TestRunner) Run(executor TestExecutor, verbosity int, skipTestCases []s
 func (t *TestRunner) Finish() {
 	t.reporter.Summarize(t.tests)
 	for _, test := range t.tests {
-		if test.Error() != nil {
-			os.Exit(1)
+		if err := test.Error(); err != nil {
+			// TODO(onelson): would prefer to return the error but don't want to
```
jsternberg, line 552: I was wrong about this. This function is safe to change. Please do that instead unless I've missed something.

jsternberg (changes requested):

cmd/flux/cmd/test.go
onelson: I sort of took the fact it was capitalized to say someone _could be calling it_. I guess all it would mean to them is now they are ignoring an error, right? Not a big breaking change per se.

cmd/flux/cmd/test.go
```diff
@@ -46,7 +46,9 @@ func TestCommand(setup TestSetupFunc) *cobra.Command {
 		Long:  "Run flux tests",
 		Run: func(cmd *cobra.Command, args []string) {
 			fluxinit.FluxInit()
-			runFluxTests(setup, flags)
+			if err := runFluxTests(setup, flags); err != nil {
+				panic(err)
```
jsternberg, line 50: You can call `os.Exit()` here safely. That is more correct than using a panic.

jsternberg (changes requested):
Small requested change and then we're good. We should avoid explicit panics.

cmd/flux/cmd/test.go
onelson: Sure, I can try that. I ran into an edge case on the idpe side that prompted me to push the deferred cleanup closer to the top of the call stack, where an Exit here would interfere (causing a forked process to outlive the test harness).

I'll see if I can get around the issue for idpe and try  this out.

onelson: The gist of the issue is around this:

```
	executor, err := setup(context.Background())
	if err != nil {
		return err
	}
	defer func() { _ = executor.Close() }() // never called if setup gives an err

	runner.Run(executor, flags.verbosity, flags.skipTestCases)
	return runner.Finish()
}
```

For the idpe w/ IOx use case, setup will try to fork a process to run IOx mid-way through, with later parts of the impl possibly giving errors.

To ensure the forked process gets killed on the way out, we therefore need something higher in the call stack.

This `Run` callback is the callsite for this function, so if it calls `Exit`, the cleanup can't happen (since it would be waiting, deferred higher-up still).

For reference, this is how the `Command` is invoked:

```
func tryExec(cmd *cobra.Command) (err error) {
	defer func() {
		if e := recover(); e != nil {
			var ok bool
			err, ok = e.(error)
			if !ok {
				err = errors.New(fmt.Sprint(e))
			}
		}
	}()
	defer func() { _ = killIOx() }()
	err = cmd.Execute() // the Exit call would hit here, meaning the above defer will not fire
	return
}

func main() {
	c := cmd.TestCommand(NewTestExecutor)
	c.Use = "fluxtest"
	if err := tryExec(c); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}
```

onelson: For completeness -- there was room to kill the forked process inside the setup function _after detecting an error in `init`_. This meant calling `Exit` in the command's callback is no longer an issue for us.

jsternberg (approved):

PR #4476, opened 2022-02-11
test: Make TestDo_DNSFail work on ubuntu (or my computer at least)

Marwes:
Getting a different error locally so the test would fail.


rockstar (approved):

onelson:
I started seeing this yesterday too! :tada:

PR #4477, opened 2022-02-11
refactor: Merge Array and Vector into an App + Collection type

Marwes:
Makes it easier to add more parameterized types since it is mostly just adding another variant to `Collection`. `Collection` could realistically be a `MonoType` which would be a path towards higher kinded types, however that is a larger change.

Should perhaps have tried doing this before #4468 to save myself a bit of work.


scbrickley (approved):
This looks good. I have one nitpick, but it's not important enough to block this patch from merging.

The name `App` doesn't really mean much to me as someone maintaining this code. Maybe there's some context that I'm missing here, but it looks like it's just a wrapper around a collection. I think renaming `App` to `Collection` and renaming the current `Collection` to something else (maybe `CollectionType`?) would make the things clearer for others working in this code.

Not a huge deal, just thought I'd address it. Like I said, I think this is good to merge.

PR #4485, opened 2022-02-15
test(doc): Improve doc tests

Marwes:
## test: Don't early exit on errors in flux doc tests

Gathers up the result of each individual test and prints them at the top level

## test: Run doc tests in parallel

Adds rayon as a dependency to run each tested package in parallel which cuts down the test time a bit. Adds some complexity to ensure that the test results are still printed in the same order.

### DIFF
--- a/libflux/flux-core/src/bin/fluxdoc.rs
+++ b/libflux/flux-core/src/bin/fluxdoc.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::BinaryHeap,
     fs::File,
     io::{self, Write},
     path::{Path, PathBuf},
@@ -6,6 +7,7 @@ use std::{
 };
 
 use anyhow::{bail, Context, Result};
+use rayon::prelude::*;
 use structopt::StructOpt;
 
 use fluxcore::{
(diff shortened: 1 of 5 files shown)


libflux/flux-core/src/bin/fluxdoc.rs
```diff
@@ -209,24 +213,93 @@ fn lint(
...
+/// Iterates through `iter` in parallel however each item is passed to `consume` in the same order
+/// that they were produced by the iterator as long as the iterator uses `enumerate` to supply
+/// indices. Will deadlock if the indicies passed are not a complete 0..N sequence.
+fn consume_sequentially<T>(
+    iter: impl IndexedParallelIterator<Item = (usize, T)>,
+    mut consume: impl FnMut(T) + Send,
+) where
+    T: Send,
+{
+    let (sender, receiver) = std::sync::mpsc::sync_channel::<Element<T>>(12);
```
onelson, line 259: Curious about the channel bounds here - what was `12` based on?

onelson (approved):
LGTM

libflux/flux-core/src/bin/fluxdoc.rs
Marwes: https://xkcd.com/221/

onelson: Ah. :sweat_smile:

PR #4487, opened 2022-02-15
feat(stdlib): update fill to use narrow transformation

### LINKED ISSUE #3938: Update fill transformation to use NarrowTransformation
Update the `fill()` transformation to use [NarrowTransformation](https://github.com/influxdata/flux/issues/3927).

The `fill()` transformation presently uses `table.Stream()` to stream tables to the downstream. The transformation should be adapted to use `NarrowTransformation` instead. This should involve the following steps:

1. Remove `table.Stream()`.
2. Run the fill algorithm on each table chunk instead of the entire table.
3. Send the table chunk using the `TransportDataset`.

skartikey:
This updates fill to use the `NarrowStateTransformation` transformation interface instead of the stream table.

Closes #3938

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/universe/fill.gen.go.tmpl
+++ b/stdlib/universe/fill.gen.go.tmpl
@@ -3,6 +3,7 @@ package universe
 import (
 	"fmt"
 
+	arrowmem "github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/values"
@@ -47,5 +48,47 @@ func (t *fillTransformation) fill{{.Name}}Column(arr *{{.ArrowType}}, fillValue 
 		*fillValue = fillValue{{.Name}}
 	}
 	return b.NewArray()
+}
+{{end}}
+
+func (t *fillTransformation) fillColumnNT(typ flux.ColType, arr array.Interface, fillValue *interface{}, mem arrowmem.Allocator) array.Interface {
+    switch typ {
+    {{range .}}case {{.ColumnType}}:
+        return t.fill{{.Name}}ColumnNT(arr.(*{{.ArrowType}}), fillValue, mem)
+    {{end}}
+    default:
+        panic(fmt.Errorf("unsupported array data type: %s", arr.DataType()))
+    }
+}
+
+{{range .}}
+func (t *fillTransformation) fill{{.Name}}ColumnNT(arr *{{.ArrowType}}, fillValue *interface{}, mem arrowmem.Allocator) array.Interface {
+    fillValueNull := *fillValue == nil
+    var fillValue{{.Name}} {{.Type}}
+    if !fillValueNull {
+        {{if eq .Name "Time"}}fillValue{{.Name}} = {{.Type}}((*fillValue).(values.Time)){{else}}fillValue{{.Name}} = (*fillValue).({{.Type}}){{end}}
+    }
+    b := array.New{{.ArrowName}}Builder(mem)
+    b.Resize(arr.Len())
+    for i := 0; i < arr.Len(); i++ {
+        if arr.IsNull(i) {
+            if fillValueNull {
+                b.AppendNull()
+            } else {
+                b.{{.Append}}(fillValue{{.Name}})
+            }
+        } else {
+            v := arr.{{.Value}}(i)
+            b.{{.Append}}(v)
+            if t.spec.UsePrevious {
+                fillValue{{.Name}} = v
+                fillValueNull = false
+            }
+        }
+    }
+    if t.spec.UsePrevious && !fillValueNull {
+        *fillValue = fillValue{{.Name}}
+    }
+    return b.NewArray()
 }
 {{end}}
--- a/stdlib/universe/fill.gen.go.tmpl
+++ b/stdlib/universe/fill.gen.go.tmpl
@@ -9,52 +9,10 @@ import (
 	"github.com/influxdata/flux/values"
 )
 
-func (t *fillTransformation) fillColumn(typ flux.ColType, arr array.Interface, fillValue *interface{}) array.Interface {
-	switch typ {
-	{{range .}}case {{.ColumnType}}:
-		return t.fill{{.Name}}Column(arr.(*{{.ArrowType}}), fillValue)
-	{{end}}
-	default:
-		panic(fmt.Errorf("unsupported array data type: %s", arr.DataType()))
-	}
-}
-
-{{range .}}
-func (t *fillTransformation) fill{{.Name}}Column(arr *{{.ArrowType}}, fillValue *interface{}) array.Interface {
-	fillValueNull := *fillValue == nil
-	var fillValue{{.Name}} {{.Type}}
-	if !fillValueNull {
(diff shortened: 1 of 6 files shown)


wolffcm (approved):
Looks good to me!

skartikey:
Tests failing at the moment. 

```
--- FAIL: TestFill_Process_Narrow (0.00s)
    --- FAIL: TestFill_Process_Narrow/fill_previous_empty_table (0.00s)
        transformation.go:129: caught panic: runtime error: invalid memory address or nil pointer dereference
    --- FAIL: TestFill_Process_Narrow/non_null_group_key (0.00s)
        transformation.go:129: caught panic: runtime error: invalid memory address or nil pointer dereference
```

stdlib/universe/fill.go
```diff
@@ -266,6 +273,161 @@ func (t *fillTransformation) Finish(id execute.DatasetID, err error) {
...
+		return nil, false, err
+	}
+	return dstate, true, nil
+}
+
+func (t *fillTransformationAdapter) Close() error { return nil }
+
+func (t *fillTransformation) fillChunk(buffer *arrow.TableBuffer, chunk table.Chunk, colIdx int, fillValue *interface{}, mem arrowmem.Allocator) error {
+	if chunk.Len() == 0 {
+		return nil
```
Marwes, line 396: I would guess that this is the cause for the nil reference, it makes you fail to initialize the `Values` of `TableBuffer` (I had that issue)

stdlib/universe/fill_test.go
```diff
@@ -746,6 +746,658 @@ func TestFill_Process(t *testing.T) {
...
+					{execute.Time(1), 1.0, 2.0},
+					{execute.Time(2), 1.0, nil},
+					{execute.Time(3), 1.0, 4.0},
+					{execute.Time(4), 1.0, nil},
+				},
+				KeyCols:   []string{"tag0"},
+				KeyValues: []interface{}{1.0},
+			}},
+		},
+	}
```
wolffcm, line 1379: I guess all of these test cases are copies of the test cases from `TestFill_Process`? Instead of copying all these cases, could you just add a new call to `t.Run` in the original test case for the narrow version?

stdlib/universe/fill.go
```diff
@@ -266,6 +273,163 @@ func (t *fillTransformation) Finish(id execute.DatasetID, err error) {
...
+		fillTransformation,
+	}
+	return execute.NewNarrowStateTransformation(id, t, alloc)
+}
+
+type fillState struct {
+	colIdx        int
+	fillValue     interface{}
+	groupKey      flux.GroupKey
+	outputColumns []flux.ColMeta
```
wolffcm, line 295: Can we assume that different chunks of the same table must have the same schema? (@jsternberg maybe you have some insight here, I forget what the current answer to this question is)

It seems like if the schema can change, then we need to be robust to that (`colIdx` and `outputColumn` could be different between chunks and so would just need to be recomputed per-chunk)

wolffcm (changes requested):
This looks good to me, I'm just requesting some changes to how the tests are run.

I also had a question about chunk schema vs tables.

stdlib/universe/fill.go
jsternberg: You can't assume that different chunks have the same schema at the moment, but you also don't aggressively have to protect against it. If you're internally consistent and it isn't incompatible with the current state of a transformation, then you can just roll with it.

For things like the ordering of columns, you are correct that caching `colIdx` likely wouldn't do anything.

jsternberg: I'm looking more at this now and I can say that this fill state doesn't need these. The fill state will only need the previous value and the input type. The input type is to make sure that the input column type is consistent and the value is to fill in the previous value. This state isn't even necessary for things outside of fill previous.

The group key doesn't need to be stored as the state is stored with the group key by default. The `colIdx` won't be helpful since you'll have to look it up for each chunk anyway. The output columns don't have to be consistent between chunks.

stdlib/universe/fill.gen.go.tmpl
```diff
@@ -49,3 +50,45 @@ func (t *fillTransformation) fill{{.Name}}Column(arr *{{.ArrowType}}, fillValue
...
+    {{range .}}case {{.ColumnType}}:
+        return t.fill{{.Name}}ColumnNT(arr.(*{{.ArrowType}}), fillValue, mem)
+    {{end}}
+    default:
+        panic(fmt.Errorf("unsupported array data type: %s", arr.DataType()))
+    }
+}
+
+{{range .}}
+func (t *fillTransformation) fill{{.Name}}ColumnNT(arr *{{.ArrowType}}, fillValue *interface{}, mem arrowmem.Allocator) array.Interface {
```
wolffcm, line 65: I can see that the only difference between these methods and the original ones is that there is an allocator passed in here. Instead of duplicating this, can you add an allocator parameter to the old methods?

stdlib/universe/fill_test.go
skartikey: This is a temporary change till the narrow version is successfully enabled in the production. Once that is done my plan is to replace the `TestFill_Process_Narrow` with `TestFill_Process` and delete the old test completely.

stdlib/universe/fill.gen.go.tmpl
skartikey: Same as above, this is a temporary change and the original will be completely deleted.
I do not have any preference here, I can change it as suggested. Kindly let me know.

stdlib/universe/fill_test.go
wolffcm: I understand. It's helpful to reviewers if there is less duplicated code, then it would be obvious that the new transformation is just running all the same tests, and they continue to pass. Also, if something goes wrong while we test this out and we need to go back and fix something, it will be less confusing if there is not duplicated code.

There's definitely cases where duplicating code is useful, but I'd rather that it be done sparingly and there seems to be easy ways around it here.

stdlib/universe/fill.go
skartikey: In the current implementation, the pointer of `fillValue` is passed and the previous value gets populated in the same variable. So caching `fillValue` makes sense here. 
I did the changes to cache the `fillValue` only and removed index calculation and output column from the caching.

stdlib/universe/fill_test.go
skartikey: That makes sense, I have changed the test to call `t.Run` twice for existing and new cases.

stdlib/universe/fill_test.go
```diff
@@ -743,6 +812,24 @@ func TestFill_Process(t *testing.T) {
 				},
 			)
 		})
+
+		// fill narrow transformations tests
+		t.Run(tc.name, func(t *testing.T) {
```
onelson, line 817: fyi, this is the issue I mentioned making in the standup. You may need to modify the test name here to avoid confusing the test runner. I added a "narrow" suffix to mine for this case.

skartikey: Thanks Owen, will do the changes.

PR #4489, opened 2022-02-15
feat(test): add extra versbose level to dump AST of test

nathanielc:
Print the entire AST of the test case post transform with -vvv

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/cmd/flux/cmd/test.go
+++ b/cmd/flux/cmd/test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/ast/astutil"
 	"github.com/influxdata/flux/ast/testcase"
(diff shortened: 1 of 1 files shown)


cmd/flux/cmd/test.go
```diff
@@ -176,7 +193,7 @@ func (t *TestRunner) Gather(roots []string, names []string) error {
 		for _, file := range files {
 			q, err := filesystem.ReadFile(ctx, file)
 			if err != nil {
-				return err
+				return errors.Wrapf(err, codes.Invalid, "could not find test file %q", file)
```
nathanielc, line 196: @wolffcm FYI added a wrapping error so that we know which file it can't read.

jsternberg (approved):
Commit message has a typo. `versbose` -> `verbose`.

PR #4496, opened 2022-02-17
perf: Update the difference transformation to use narrow transformation

### LINKED ISSUE #3942: Refactor difference transformation to use NarrowStateTransformation
Refactor the `difference()` transformation to use [NarrowStateTransformation](https://github.com/influxdata/flux/issues/3933).

At the present moment, `difference()` uses the column list table builder to construct tables. Fortunately, the algorithm used is a column-based construction so it should be easier to refactor this to use arrow arrays directly. Tables constructs from arrow arrays are generally easier to create in column-oriented algorithms.

This issue will be very similar to work already done to `derivative()`, which underwent a similar refactoring: https://github.com/influxdata/flux/pull/4043
(NB: there was a fix up PR to this one)

The algorithm should be changed to do the following:

1. The difference state should be initialized and maintained by the `NarrowStateTransformation`.
2. The difference algorithm for each column type should be refactored into functions instead of in the switch body.
3. The functions should return the arrow array for the column.
4. For each table chunk, the process method should produce a new table chunk and send it to the transport dataset.

Marwes:
I were a bit worried about how this would handle schema mismatches but it seems like each new table resets the state anyway so the state will get reset just the same as the old implementation (I added a test since there wasn't one before).

Do we want a feature flag for this change?

Closes #3942

### DIFF
--- a/stdlib/universe/difference_test.go
+++ b/stdlib/universe/difference_test.go
@@ -104,8 +104,8 @@ func TestDifference_Process(t *testing.T) {
 					{Label: "_value", Type: flux.TString},
 				},
 				Data: [][]interface{}{
-					{execute.Time(1), int64(20)},
-					{execute.Time(2), int64(10)},
+					{execute.Time(1), "a"},
+					{execute.Time(2), "b"},
 				},
 			}},
 			wantErr: errors.New(codes.Invalid, `difference does not support column "_value" of type "string"`),
--- a/stdlib/universe/difference_test.go
+++ b/stdlib/universe/difference_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/querytest"
 	"github.com/influxdata/flux/stdlib/universe"
+	"github.com/influxdata/flux/values"
 )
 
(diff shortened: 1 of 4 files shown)


nathanielc (changes requested):
> Do we want a feature flag for this change?

Yes we do, otherwise if we do happen to have a correctness bug its easy to turn off the code and find a fix.

Otherwise LGTM

stdlib/universe/difference_test.go
```diff
@@ -23,17 +25,6 @@ func TestDifferenceOperation_Marshaling(t *testing.T) {
 	querytest.OperationMarshalingTestHelper(t, data, op)
 }
 
-func TestDifference_PassThrough(t *testing.T) {
```
Marwes, line 26: There wasn't an equivalent test helper for this for the updated `New*`signature (like there is for the other tests). Should I try to add one or is it not relevant anymore?

nathanielc (approved):

stdlib/universe/difference_test.go
jsternberg: Yea this test isn't relevant anymore

PR #4497, opened 2022-02-17
feat(feature): allow feature flags to record metrics if configured

jsternberg:
This allows feature flags to record that a metric has been accessed
through an interface that is set by the user. By default, this interface
discards the metric. This can be configured in order to allow an
application to gather things like prometheus metrics from accessing
feature flags.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


wolffcm (approved):
I guess whatever embeds Flux will need to set the metrics at startup? Looks good to me. I appreciate the tests.

PR #4498, opened 2022-02-18
feat: add second pass to physical planner for parallelization rules

adrian-thurston:
This PR adds a second pass to the physical planner that is dedicated to
running parallelization rules.

Parallelization is something we want to embark on only after all pushdowns have
succeed. Adding a second pass ensures that we can never parallelize a node and
block a pushdown.

Suppose we include parallelization rules in the set of physical rules. Consider:

```    
    range -> filter -> agg
```

And the rules happen to parallelize the filter node first:

```
    parallel-range -> parallel-filter -> merge -> agg
```

However, if the rules fail to parallelize the aggregate as well, then we have
blocked the `range-filter-agg` pushdown. Using a second pass frees us
from the challenging task of ensuring we never block a pushdown.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/plan/physical.go
+++ b/plan/physical.go
@@ -19,20 +19,30 @@ type PhysicalPlanner interface {
 // The new plan will be configured to apply any physical rules that have been registered.
 func NewPhysicalPlanner(options ...PhysicalOption) PhysicalPlanner {
 	pp := &physicalPlanner{
-		heuristicPlanner:   newHeuristicPlanner(),
-		defaultMemoryLimit: math.MaxInt64,
+		heuristicPlannerPhysical: newHeuristicPlanner(),
+		heuristicPlannerParallel: newHeuristicPlanner(),
+		defaultMemoryLimit:       math.MaxInt64,
 	}
 
-	rules := make([]Rule, len(ruleNameToPhysicalRule))
+	rulesPhysical := make([]Rule, len(ruleNameToPhysicalRule))
 	i := 0
 	for _, v := range ruleNameToPhysicalRule {
-		rules[i] = v
+		rulesPhysical[i] = v
 		i++
 	}
 
-	pp.addRules(rules...)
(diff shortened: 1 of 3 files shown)


Marwes:
Couldn't filter itself just be updated to run in parallel? Instead of requiring a separate pass for it.

wolffcm (approved):
Looks good!

wolffcm:
@Marwes What might not be obvious from this PR is that whole branches of the execution graph will run in parallel, where e.g., the source node that is reading from partition _i_ will send its data to the filter node that processes data for partition _i_, and so on up to the merge node where all the data is unioned. Doing it this way means we don't have to write as many rules that express what can be pushed through a merge node. (If I understand your question right)

adrian-thurston:
> Couldn't filter itself just be updated to run in parallel? Instead of requiring a separate pass for it.

Hi @Marwes, some broader context is missing from this PR. Just realizing now. I didn't really present the full plan along with this PR.

The goal of the parallelization work is to parallelize along sources and include transformations that don't need to see all source data in order to run. The motivating case is

```
read-window-aggregate -> to
```

Turning this into

```
read-window-aggregate -> to -> merge
read-window-aggregate -> to ----^
```

Where each `read-window-aggregate` reads from a subset of the 64 storage nodes holding partition data. For now I am just parallelizing `to`, but I would like to include non-pushable `filter` transformations, and any other transformation that is amenable to parallelization.

One of the advantages of this approach is that it requires minimal changes to the flux engine. If you replicate and specialize execution nodes, all the existing machinery of flux just handles the heavy lifting of executing in parallel. The parallelization factor is strong, near perfect up to 8 cores and a factor of 0.5 up to 32 cores. It has the potential to make some queries scream :) 

More details here: https://github.com/influxdata/idpe/issues/12693

PR #4500, opened 2022-02-18
fix(length): removes unnecessary TableObject guards

### LINKED ISSUE #4343: Remove TableObject type assertion guards
This work will remove guards that check to see if a `Value` is a `TableObject` just before we make an attempt to invoke the `Array.Len()`. 

Once we after we are able to properly represent `TableObject`s separately from standard `array` objects, the type assertions should be redundant. At that point, hopefully both the compiler and humans will be able to tell the difference between these two types.

nathanielc:
Now that the type system can distinguish between streams (aka tables)
and vanilla arrays we no longer need the TableObject interface and its
associated type guards.

Fixes #4343

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -80,9 +80,6 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 	// On specifies the columns to join on, and is required.
 	if array, err := args.GetRequiredArray("on", semantic.String); err != nil {
 		return nil, err
-	} else if _, ok := array.(values.TableObject); ok {
-		// XXX: remove when array/stream are different types <https://github.com/influxdata/flux/issues/4343>
-		return nil, errors.New(codes.Invalid, "on cannot be a table stream; expected an array")
 	} else if array.Len() == 0 {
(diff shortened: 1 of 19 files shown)


stdlib/universe/join.go
```diff
@@ -81,10 +81,6 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 	if array, err := args.GetRequiredArray("on", semantic.String); err != nil {
 		return nil, err
 	} else if _, ok := array.(values.TableObject); ok {
-		// XXX: remove when array/stream are different types <https://github.com/influxdata/flux/issues/4343>
-		return nil, errors.New(codes.Invalid, "on cannot be a table stream; expected an array")
-	} else if array.Len() == 0 {
-		return nil, errors.New(codes.Invalid, "at least one column in 'on' column list is required")
```
nathanielc, line 87: fixing this...

nathanielc: fixed

onelson (approved):
Looks great. Cheers!

big-red

PR #4503, opened 2022-02-23
feat: add an attribute-based instantiation of parallel execution nodes

adrian-thurston:
This PR expands the definition of an attribute on a physical plan node to a
collection of named simple interfaces:

```
    type PhysicalAttributes map[string]PhysicalAttr
```

The `PhysicalAttr` interface allows arbitrary data structures to be installed as
attribute values.

This PR then adds two attribute types:

 1. `ParallelRun` - indicates that a plan node should be instantiated multiple
    times when creating the execution graph. Each instantiation will handle
    some subset of the data. In the code we call this the parallel group. If
    the node is a source node, it should restrict source data to the specified
    group it has been assigned. If it is a non-source node, it should expect to
    receive parallel data (the group), and produce parallel data.

 2. `ParallelMerge` - indicates that plan node should merge the multiple
    instantiations of its predecessors. It receives parallel data and produces
    non-parallel data that can be sent to any existing transformation node.

For example, in the following graph, the `rwo` and `to` nodes will carry the
ParallelRun attribute and will be instantiated multiple times. The merge node
will carry the ParallelMerge attribute and will be responsible for unioning the
subsets.

```
    read-window-aggregate -> to -> merge
```

A node has output attributes and required attributes. There is a new validation
step after planning that ensures predecessors output the attributes that it
requires. It is also possible for an attribute to require that all successors
require it. Using these tools we can validate that we always have a sound
convergence from parallel data to non-parallel data after the planner rewrites
the graph to create parallel execution graphs.

During instantiation of the execution graph from the plan, there are three
types of instantiations we need to support, multiplied by the two existing
cases -- source and non-source.

 1. Standard instantiation. These are non-parallel, non-merge nodes.
    There is a 1:1 relationship between planner node and execution graph
    node. Only a single copy of the node is made and it references the
    single copy of the predecessor nodes.

 2. Parallel instantiation. There are multiple copies of the node and of
    all predecessors. This applies to both source and non-source nodes.

 3. Merge instantiation. There is a single copy of the node, but multiple
    copies of the predecessors. These copies merge into the node.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -72,7 +72,7 @@ func (id DatasetID) IsZero() bool {
 	return id == ZeroDatasetID
 }
 
-func DatasetIDFromNodeID(id plan.NodeID, instance int) DatasetID {
+func datasetIDFromNodeID(id plan.NodeID, instance int) DatasetID {
 	return DatasetID(uuid.NewV5(uuid.UUID{}, string(id)+"-"+strconv.Itoa(instance)))
 }
 
--- a/execute/dataset.go
+++ b/execute/dataset.go
@@ -2,6 +2,7 @@ package execute
 
 import (
 	"context"
+	"strconv"
 
 	"github.com/apache/arrow/go/arrow/memory"
 	uuid "github.com/gofrs/uuid"
@@ -71,8 +72,8 @@ func (id DatasetID) IsZero() bool {
 	return id == ZeroDatasetID
 }
 
-func DatasetIDFromNodeID(id plan.NodeID) DatasetID {
-	return DatasetID(uuid.NewV5(uuid.UUID{}, string(id)))
+func DatasetIDFromNodeID(id plan.NodeID, instance int) DatasetID {
+	return DatasetID(uuid.NewV5(uuid.UUID{}, string(id)+"-"+strconv.Itoa(instance)))
 }
 
 type dataset struct {
--- a/execute/executetest/source.go
+++ b/execute/executetest/source.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	uuid "github.com/gofrs/uuid"
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/dependencies/dependenciestest"
@@ -25,8 +26,6 @@ type FromProcedureSpec struct {
 	execute.ExecutionNode
 	data []*Table
 	ts   []execute.Transformation
-	a    execute.Administration
-	id   execute.DatasetID
 }
 
 // NewFromProcedureSpec specifies a from-test procedure with source data
@@ -55,7 +54,9 @@ func (src *FromProcedureSpec) AddTransformation(t execute.Transformation) {
 }
 
 func (src *FromProcedureSpec) Run(ctx context.Context) {
-	popts := src.a.ParallelOpts()
+	// uuid.NewV4 can return an error because of enthropy. We will stick with the previous
+	// behavior of panicing on errors when creating new uuid's
+	id := execute.DatasetID(uuid.Must(uuid.NewV4()))
 
 	if len(src.ts) == 0 {
 		return
@@ -64,12 +65,7 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
 
 		var max execute.Time
 		for _, tbl := range src.data {
-			// If restricted to a parallel run group we may need to skip the table.
-			if popts.Factor > 1 && popts.Group != tbl.ResidesOnPartition {
-				continue
-			}
-			tbl.ParallelGroup = popts.Group
-			t.Process(src.id, tbl)
+			t.Process(id, tbl)
 			stopIdx := execute.ColIdx(execute.DefaultStopColLabel, tbl.Cols())
 			if stopIdx >= 0 {
 				if s := tbl.Key().ValueTime(stopIdx); s > max {
@@ -77,17 +73,15 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
 				}
 			}
 		}
-		t.UpdateWatermark(src.id, max)
-		t.Finish(src.id, nil)
+		t.UpdateWatermark(id, max)
+		t.Finish(id, nil)
 		return
 	}
 
 	buffers := make([]flux.BufferedTable, 0, len(src.data))
-	residesOnPartition := make([]int, 0, len(src.data))
 	for _, tbl := range src.data {
 		bufTable, _ := execute.CopyTable(tbl)
 		buffers = append(buffers, bufTable)
-		residesOnPartition = append(residesOnPartition, tbl.ResidesOnPartition)
 	}
 
 	// Ensure that the buffers are released after the source has finished.
@@ -99,11 +93,8 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
 
 	for _, t := range src.ts {
 		var max execute.Time
-		for i, tbl := range buffers {
-			if popts.Factor > 1 && popts.Group != residesOnPartition[i] {
-				continue
-			}
-			t.Process(src.id, tbl.Copy())
+		for _, tbl := range buffers {
+			t.Process(id, tbl.Copy())
 			stopIdx := execute.ColIdx(execute.DefaultStopColLabel, tbl.Cols())
 			if stopIdx >= 0 {
 				if s := tbl.Key().ValueTime(stopIdx); s > max {
@@ -111,16 +102,13 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
 				}
 			}
 		}
-		t.UpdateWatermark(src.id, max)
-		t.Finish(src.id, nil)
+		t.UpdateWatermark(id, max)
+		t.Finish(id, nil)
 	}
 }
 
 func CreateFromSource(spec plan.ProcedureSpec, id execute.DatasetID, a execute.Administration) (execute.Source, error) {
-	fps := *spec.(*FromProcedureSpec)
-	fps.a = a
-	fps.id = id
-	return &fps, nil
+	return spec.(*FromProcedureSpec), nil
 }
 
 // AllocatingFromProcedureSpec is a procedure spec AND an execution node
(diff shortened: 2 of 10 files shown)


plan/physical.go
```diff
@@ -123,6 +123,54 @@ func (pp *physicalPlanner) Plan(ctx context.Context, spec *Spec) (*Spec, error)
...
+
+		for key, attr := range ppn.RequiredAttrs {
+			for _, pred := range ppn.Predecessors() {
+				ppred := pred.(*PhysicalPlanNode)
+				predAttr, ok := ppred.OutputAttrs[key]
+				if !ok {
+					return fmt.Errorf("invalid physical query plan; attribute \"%v\" required by "+
+						"\"%v\" is missing from predecessor \"%v\"", key, ppn.id, ppred.id)
+				}
+				if attr != predAttr {
```
wolffcm, line 141: If we are going to use a simple equality check here, that has some implications.  There could also be a `Equals` method or something. Seems fine for now, we might want to revisit it later.

stdlib/universe/parallel.go
```diff
@@ -0,0 +1,168 @@
...
+	ctx     context.Context
+	dataset *execute.PassthroughDataset
+	span    opentracing.Span
+	alloc   *memory.Allocator
+
+	mu          sync.Mutex
+	parentState map[execute.DatasetID]*parallelParentState
+}
+
+type parallelParentState struct {
```
wolffcm, line 68: Could we change `Parent` to `Predecessor` here? I find the former to be ambiguous and confusing in a DAG.

stdlib/universe/parallel.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	"github.com/opentracing/opentracing-go"
+)
+
+const (
+	PartitionMergeKind = "influx2x/partitionMergeKind"
```
wolffcm, line 18: Should this be called `ParallelMergeKind` instead? Since a partition is a storage concept. Or maybe you are introducing it as a Flux concept as well right now. In any case could you add some comments here describing what this node does?

plan/physical.go
```diff
@@ -308,9 +361,52 @@ func (ppn *PhysicalPlanNode) Cost(inStats []Statistics) (cost Cost, outStats Sta
...
+	return true
+}
+
+type ParallelMergeAttribute struct {
+	Factor int
+}
+
+func (ParallelMergeAttribute) SuccessorsMustRequire() bool {
+	return false
 }
```
wolffcm, line 410: It might be good if all the attributes were in a file on their own?

execute/executetest/table.go
```diff
@@ -46,6 +46,11 @@ type Table struct {
 	Alloc *memory.Allocator
 	// IsDone indicates if this table has been used.
 	IsDone bool
+	// PartitionSource indicates which partition the table is on.
+	ResidesOnPartition int
+	// ParallelGroup is assigned during execution so we can see in the
+	// execution output which group the data originated from.
+	ParallelGroup int
```
wolffcm, line 53: Is there another way to test out parallel execution of sources without adding these fields to `executetest.Table`? This class is used in many places (including outside this repo) and may also serve as an example of an implementation of `flux.Table` for some. In general a table does not need to know which partition its on (and it may even change partitions).

execute/executetest/source.go
```diff
@@ -93,22 +99,28 @@ func (src *FromProcedureSpec) Run(ctx context.Context) {
 
 	for _, t := range src.ts {
 		var max execute.Time
-		for _, tbl := range buffers {
-			t.Process(id, tbl.Copy())
+		for i, tbl := range buffers {
```
wolffcm, line 102: Rather than complicate this logic, would it make more sense to create a new kind of source in `executetest` and embed `executetest.FromProcedureSpec` inside it?

execute/parallel_test.go
```diff
@@ -0,0 +1,435 @@
...
+							{execute.Time(5), execute.Time(10), execute.Time(5), 5.0, int64(1)},
+							{execute.Time(5), execute.Time(10), execute.Time(6), 6.0, int64(1)},
+							{execute.Time(5), execute.Time(10), execute.Time(7), 7.0, int64(1)},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: `parallel-from-filter`,
```
wolffcm, line 160: _nit_ the first test case was also `parallel-from-filter` and this one just has an extra yield.

plan/physical.go
wolffcm: could I trouble you to change this error and the other errors in this file to use the modern format for generating errors?
```go
return flux.Error{
  code: codes.Internal, // flux/codes
  Msg: "...",
}
```
This message would indicate a bug in the planner or in planner rules, so it would be good to explicitly make it internal.

wolffcm (changes requested):
This looks fantastic overall. Thanks for this work!

For the changes to the product code, I had some small comments and a question about how you construct the execution graph. But other that it looks very much like I would have expected.

I have some concerns though about the changes in the `executetest` package. These implementations I think tend to serve as a reference (at least to me) for some of the important abstractions used by the runtime. It would be good to keep them simple if there is an alternative.

execute/parallel_test.go
adrian-thurston: In the first case only the `from` is parallel. In the second case the `from` and the `filter` run in parallel. I think some comments are in order to clear that up.

execute/executetest/source.go
adrian-thurston: For sure, that seems the way to go.

adrian-thurston: Done.

execute/parallel_test.go
adrian-thurston: I added some comments above the tests and also improved the test case names.

execute/executetest/table.go
adrian-thurston: I was looking for a way to verify that the data was indeed coming from multiple instantiations of the execution nodes, but couldn't think of any other way, other than modifying the data to include the parallel group. I'll keep thinking about it.

plan/physical.go
adrian-thurston: Done.

stdlib/universe/parallel.go
adrian-thurston: Ah yes. This was originally in IDPE and I was mostly using "partition" instead of "parallel". Will fix.

stdlib/universe/parallel.go
adrian-thurston: Done.

plan/physical.go
adrian-thurston: Done.

plan/physical.go
```diff
@@ -123,6 +123,54 @@ func (pp *physicalPlanner) Plan(ctx context.Context, spec *Spec) (*Spec, error)
 	return transformedSpec, nil
 }
 
+func ValidateAttributes(plan *Spec) error {
+	err := plan.BottomUpWalk(func(pn Node) error {
+		ppn, ok := pn.(*PhysicalPlanNode)
+		if !ok {
+			return fmt.Errorf("invalid physical query plan; found logical operation \"%v\"", pn.ID())
```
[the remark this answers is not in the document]
adrian-thurston: I wanted to use the attribute validator in the `parallel_test.go` file in a standalone fashion. Using the entire plan validator caused some trigger spec errors. Wondering if you could provide some insight on how to fix that. Merging the two loops and using the entire validator is probably the right way to go.

```
parallel_test.go:389: invalid physical query plan; trigger spec not set on "parallel-from-test"
```

plan/physical.go
adrian-thurston: I upgraded all errors in `plan/physical.go` to use `flux.Error`.

plan/physical.go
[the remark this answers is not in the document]
adrian-thurston: I added the setting of trigger specs in the parallel execution tests and merged the two validation functions.

execute/executor.go
```diff
@@ -173,75 +172,127 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
...
+		ec[i] = executionContext{
+			es:            v.es,
+			parents:       make([]DatasetID, len(node.Predecessors())*predCopies),
+			streamContext: streamContext,
+			parallelOpts:  ParallelOpts{Group: i, Factor: copies},
+		}
+
+		for pi, pred := range nonYieldPredecessors(node) {
+			for j := 0; j < predCopies; j++ {
+				ec[i].parents[pi*predCopies+j] = DatasetIDFromNodeID(pred.ID(), j)
```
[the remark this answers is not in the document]
adrian-thurston: This code is not so easy to follow. I really thought hard about the pros and cons of cloning and modifying it for handling the different cases. For better or worse, chose this route ...

The thing to remember here is that if `predCopies` is > `1` and that inner loop at 210 is more than just a single iteration, it means we are handling the merge node. The `i` var will always be zero in that case because the merge node is not copied. It must handle the incoming datasets from all copies of the predecessors, so that's why we supply `j`.

execute/dataset.go
```diff
@@ -71,8 +72,8 @@ func (id DatasetID) IsZero() bool {
 	return id == ZeroDatasetID
 }
 
-func DatasetIDFromNodeID(id plan.NodeID) DatasetID {
-	return DatasetID(uuid.NewV5(uuid.UUID{}, string(id)))
+func DatasetIDFromNodeID(id plan.NodeID, instance int) DatasetID {
+	return DatasetID(uuid.NewV5(uuid.UUID{}, string(id)+"-"+strconv.Itoa(instance)))
```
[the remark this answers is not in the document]
adrian-thurston: Made it non-exported.

execute/executetest/table.go
wolffcm: I wonder if you could create a new implementation of `flux.Table` called `ParallelFromTable` (that embeds `executetest.Table`) and change the test `Source` to contain a slice of them.

adrian-thurston: Yeah I could do that. I had thought cloning is not ideal, but if it's the preferred method then no problem.

execute/executor.go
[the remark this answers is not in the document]
wolffcm: This code makes sense to me for the case of the merge node, but for a parallel node that's not a source, I'm not seeing how it works right. Won't this result in every instantiation of a parallel node getting the data set from the 0th instantiation of its predecessor?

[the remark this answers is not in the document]
adrian-thurston: You're right. There is a problem. I think it needs to be `i+j`. It's not getting easier to follow.

This would mean that in my testing the incoming dataset id was not correct for the `to()` nodes. I wonder why it all appeared to work correctly? Maybe `to()` doesn't depend on a correct dataset id?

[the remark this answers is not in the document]
adrian-thurston: So it seems the call to `newConsecutiveTransport` is the official linking operation. The `parents` struct in the execution context is just an informational type of thing for transformations to use. At least that's how I'm currently understanding things.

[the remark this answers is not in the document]
adrian-thurston: Looks like the only kinds of transformations making use of `parents` are `union`, `join`, `diff`, etc.  So this explains why it didn't show up in testing. I'm going to look for some way to verify that parents is indeed correct.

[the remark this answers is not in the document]
adrian-thurston: Adding some log messages confirms we need to use `i+j` in order to ensure that `parents` is correct for both parallel-run and parallel-merge nodes.

wolffcm (approved):
Looks good!

onelson (commented):
Very glad that @wolffcm gave this a detailed look. :shipit:

adrian-thurston:
Above exercise passes.

PR #4505, opened 2022-02-23
feat(stdlib): optimized version of aggregate window for specific aggregates

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/aggregate_window.go
+++ b/stdlib/universe/aggregate_window.go
@@ -1,0 +1,906 @@
+package universe
+
+import (
+	"context"
+	"math"
+	"sort"
+
+	"github.com/apache/arrow/go/arrow/memory"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/arrow"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/internal/feature"
+	"github.com/influxdata/flux/internal/mutable"
+	"github.com/influxdata/flux/interval"
+	"github.com/influxdata/flux/plan"
+	experimentaltable "github.com/influxdata/flux/stdlib/experimental/table"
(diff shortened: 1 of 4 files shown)


scbrickley (approved):

stdlib/universe/aggregate_window.go
```diff
@@ -0,0 +1,951 @@
...
+	newNode.AddPredecessors(parentNode)
+	return newNode, true, nil
+}
+
+func (a AggregateWindowRule) isValidWindowInfSpec(spec *WindowProcedureSpec) bool {
+	return spec.TimeColumn == execute.DefaultTimeColLabel &&
+		spec.StartColumn == execute.DefaultStartColLabel &&
+		spec.StopColumn == execute.DefaultStopColLabel &&
+		!spec.CreateEmpty ||
+		spec.Window.Every == infinityVar.Duration()
```
wolffcm, line 836: Should this have parentheses around `(!spec.CreateEmpty || spec.Window.Every == infinityVar.Duration())` to get the meaning you want?

stdlib/universe/aggregate_window.go
```diff
@@ -0,0 +1,951 @@
...
+func (a AggregateWindowCreateEmptyRule) Name() string {
+	return "AggregateWindowCreateEmptyRule"
+}
+
+func (a AggregateWindowCreateEmptyRule) Pattern() plan.Pattern {
+	return plan.Pat(WindowKind,
+		plan.Pat(SchemaMutationKind,
+			plan.Pat(experimentaltable.FillKind,
+				plan.OneOf([]plan.ProcedureKind{MeanKind, SumKind, CountKind},
+					plan.Pat(WindowKind, plan.Any())))))
```
wolffcm, line 906: What is the intent of this rule? I feel like there some context I'm missing. I thought that there was an issue with selectors where we needed to use `fill` to fill empty issues, but maybe I'm misremembering.

wolffcm (commented):
Here's my after-the-fact review. I had a question about how the planner rule applies and if it will confuse the planner.

It would be great if there were more comments here, in particular for the second rule---it's pretty mysterious to me what it's doing.

PR #4507, opened 2022-02-23
feat(stdlib): allow quantile with tdigest to process any numeric

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


rockstar (approved):
From slack:

```
[rockstar] I see your review request. Is this one of those “I wanna do a bunch of little patches to try and get things working for $CUSTOMER”?
[jsternberg] this one is related to $CUSTOMER
[rockstar] Okay. I ask because I wanted to know whether to ask for the test to be updated to the more modern method or not.
[jsternberg] they want to use median and they’re doing toFloat() |> aggregateWindow(fn: median)
```

At some point we'll probably have to start enforcing this, but I don't think this is the time or the context.

PR #4508, opened 2022-02-23
test(generate): add test for generate.from based on example in the docs

onelson:
We had a signature problem show up when someone tried to use an example
from the docs.

The issue was quickly fixed with <https://github.com/influxdata/flux/pull/4502>,
but the test suite didn't catch the problem so we could see a regression
in time.

This diff adds a test based on the example from the docs here:
<https://docs.influxdata.com/flux/v0.x/stdlib/universe/union/#union-two-streams-of-tables-with-empty-group-keys>

### Done checklist
- [x] Test cases written


onelson:
Sorry for all the noise/force pushes. I badly botched this tiny diff 4 different ways in a matter of minutes.

Marwes (approved):

PR #4512, opened 2022-02-24
chore(docs): update influxdb.to documentation

sanderson:
Adds details to `to()` requested by a community member.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/influxdb.flux
```diff
@@ -170,7 +170,7 @@ builtin from : (
         ?token: string,
     ) => stream[{B with _measurement: string, _field: string, _time: time, _value: A}]
 
-// to writes data to an InfluxDB Cloud or 2.x bucket.
+// to writes data to an InfluxDB Cloud or 2.x bucket and returns the written data.
```
jstirnaman, line 173: This is confusing (in raw text, at least). Use`to()` or some other notation or formatting?

stdlib/influxdata/influxdb/influxdb.flux
```diff
@@ -287,6 +287,24 @@ builtin from : (
 // a,tag2=b hum=53.5,temp=98.6 1609459380000000000
 // ```
 //
+// ### Write to multiple InfluxDB buckets
+// The example below does the following:
+//
+// 1. Writes data to `bucket1` and returns the data as it is written.
+// 2. Ungroups the returned data.
```
jstirnaman, line 294: "Ungroups" with `group()`? That may warrant more explanation?

jstirnaman (approved):
A couple suggestions.

stdlib/influxdata/influxdb/influxdb.flux
sanderson: This is a convention of the fluxdoc format. When this gets rendered into actual documentation, it will be formatted as you suggest.

stdlib/influxdata/influxdb/influxdb.flux
sanderson: Added a clearer explanation here.

PR #4514, opened 2022-02-25
perf(limit): refactor limit to use NarrowStateTransformation

onelson:
Adds an (optimistically) optimized implementation for `limit` which leverages `NarrowStateTransformation` to read the incoming dataset as chunks.

The existing tests for limit were updated to run the established cases against both versions, the original impl _and the new Narrow_ version.

I also added "narrow" versions of the benchmarks in the `limit_test` package, though I'm not sure how strong signal these give. You can run these via `make test-bench`, then probably grep the output for `Limit`:

```
BenchmarkLimit_1N_1000-8                       1         133170217 ns/op        41040584 B/op     720007 allocs/op                                                                          
BenchmarkLimit_100N_1000-8                     1         137303909 ns/op        41040168 B/op     720006 allocs/op                                                                          
BenchmarkNarrowLimit_1N_1000-8                 1         105322194 ns/op        51309176 B/op     841101 allocs/op                                                                          
BenchmarkNarrowLimit_100N_1000-8               1         105539179 ns/op        51303848 B/op     841064 allocs/op  
```

Prior art: https://github.com/influxdata/flux/commit/fd7c66a1e9fd8cffcc9db9871222de03df11503f
Fixes: #3939

### Done checklist
- [x] Test cases written


onelson:
After getting all the existing limit-related tests I could find running against both the original and new narrow implementations, I fixed one failure, but another remains:

```
--- FAIL: TestProcess_NarrowLimit_MultiBuffer (0.00s)
    limit_test.go:491: unexpected tables -want/+got
          []*executetest.Table{
                &{
                        ... // 2 identical fields
                        KeyValues: nil,
                        ColMeta:   {{Label: "_time", Type: s"time"}, {Label: "_value", Type: s"int"}},
                        Data: [][]interface{}{
                                {s"1970-01-01T00:00:00.000000020Z", int64(2)},
                                {s"1970-01-01T00:00:00.000000030Z", int64(3)},
        -                       {s"1970-01-01T00:00:00.000000040Z", int64(4)},
        +                       {s"1970-01-01T00:00:00.000000060Z", int64(6)},
        -                       {s"1970-01-01T00:00:00.000000050Z", int64(5)},
        +                       {s"1970-01-01T00:00:00.000000070Z", int64(7)},
                        },
                        Err:    nil,
                        Alloc:  nil,
                        IsDone: false,
                },
          }
```

I'm still looking into what's going wrong, but the test has _a lot_ of setup code to it. It'll take me a minute to get up to speed on what it's actually doing.

stdlib/universe/limit.go
```diff
@@ -238,3 +246,110 @@ func (t *limitTransformation) UpdateProcessingTime(id execute.DatasetID, pt exec
...
+		// invocation of `.Process` for this transformation.
+		// When the incoming chunk is empty we can short-circuit,
+		// passing along the empty, skipping further processing.
+		if chunk.Len() == 0 {
+			chunk.Retain()
+			if err := dataset.Process(chunk); err != nil {
+				return nil, false, err
+			}
+			return nil, true, nil
+		}
```
onelson, line 285: Looking for someone to tell me if this is true :cold_sweat:. My understanding is the transform will not receive empty chunks sprinkled throughout the stream and if we start empty, things will stay empty. Is that right?

sanderson (approved):

stdlib/universe/flux_test_gen.go
```diff
@@ -152928,2715 +152928,6 @@ var FluxTestPackages = []*ast.Package{&ast.Package{
...
-						Start: ast.Position{
-							Column: 9,
-							Line:   1,
-						},
-					},
-				},
-				Name: "universe_test",
-			},
-		},
 	}, &ast.File{
```
skartikey, line 152931: I am seeing some failures in the test when ran locally with `narrowTransformationLimit: true`

```
--- FAIL: TestFluxEndToEnd (12.61s)
    --- FAIL: TestFluxEndToEnd/contrib/bonitoo-io/tickscript (0.22s)
        --- FAIL: TestFluxEndToEnd/contrib/bonitoo-io/tickscript/deadman_empty (0.08s)
            flux_test.go:146: found 1 tables that were not empty
```

skartikey (changes requested):

stdlib/universe/limit.go
```diff
@@ -238,3 +246,110 @@ func (t *limitTransformation) UpdateProcessingTime(id execute.DatasetID, pt exec
...
+	// `.Process` is reentrant, so to speak. The first invocation will not
+	// include a value for `state`. Initialization happens here then is passed
+	// in/out for the subsequent calls.
+	if state == nil {
+		state_ = &limitState{n: t.limitTransformation.n, offset: t.limitTransformation.offset}
+		// When state initialization is happening, we know this is the first
+		// invocation of `.Process` for this transformation.
+		// When the incoming chunk is empty we can short-circuit,
+		// passing along the empty, skipping further processing.
+		if chunk.Len() == 0 {
```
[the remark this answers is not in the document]
onelson: Sure. I wasn't so concerned about incoming chunks being the same length, but the tests indicated part of the contract is that limit should produce _at a minimum_ an empty chunk.

This edge case is now handled via state/offset manipulation inside `.processChunk`.

stdlib/universe/flux_test_gen.go
onelson: I added a new test `n=0` to `limit_test.go` since this tickscript test was the only place in our tests where we hit this specific case.

The test was fixed (again) with offset/state manipulation in `.processChunk`.

stdlib/universe/limit.go
```diff
@@ -238,3 +246,112 @@ func (t *limitTransformation) UpdateProcessingTime(id execute.DatasetID, pt exec
...
+	stop := chunkLen
+	count := stop - start
+	if count > state.n {
+		count = state.n
+		stop = start + count
+	}
+
+	// When n=0, clamp the offsets such that we emit an empty chunk.
+	if state.n <= 0 {
+		start = 0
```
skartikey, line 312: This seems like overwriting the variable(start, stop, count) in case of n<=0, Could you please put this corner case at the top.

something like - 
```
if n <= 0 {
   return state, true, nil
}
```

skartikey (approved):
Looks good to me!
There's just 1 change I would like you to address. Please see below comment.

stdlib/universe/limit.go
onelson: I can try this again, but I struggled with this much of the day yesterday.

For whatever reason, the tests fail if we don't produce an "empty" chunk, and when I tried to do this directly I kept seeing nil pointer panics. As a last ditch, I set all the offsets to 0 and let the main block where we produce a chunk at the bottom handle this for me.

onelson: I move this up to be handled along with the zero length chunk case at the top, but it did require me to duplicate much of the code later in the method (ref: https://github.com/influxdata/flux/pull/4514/commits/1cbf0a42873f2e1696f58a662528da755ce8a36b).

@skartikey do you know of a more terse approach for building an empty chunk? I assume the work here is essentially just to ensure we're passing along the schema information.

onelson: I guess I'll merge as-is and tighten this up down the road as needed.

PR #4520, opened 2022-02-28
  refactor(stateTracking): Implement NarrowStateTransformation

### LINKED ISSUE #3943: Refactor state tracking transformation to use NarrowStateTransformation
Refactor the `stateTracking()` transformation to use [NarrowStateTransformation](https://github.com/influxdata/flux/issues/3933).

The `stateTracking()` builtin is used by `stateCount()` and `stateDuration()` as a primitive.

At the present moment, `stateTracking()` uses the column list table builder to construct tables. I am also unclear about the exact algorithm that is used, but I suspect that it is a row-based algorithm. A row-based algorithm is that it processes a row and then appends one value for each column.

For this issue, we will need to understand what the present algorithm does and either adapt it to use arrow arrays or refactor it into an algorithm that is more amenable to arrow arrays. Generally, if possible, it is more efficient to construct each column individually when constructing a table. It can sometimes be easier to append a value to each column one at a time instead of constructing the entire column.

The following should be true after the refactor:

1. The transformation uses `NarrowStateTransformation`.
2. Each call to the `Process()` method should construct a new table chunk that will be sent using the transport dataset.
3. Memory for constructed columns should be preallocated using `Resize()` to the appropriate size of the table chunk.
4. All existing tests should pass. If more tests are needed, they should be added.

scbrickley:
Closes #3943 

Creates a new version of the `stateTracking` transformation that implements the `NarrowStateTransformation` interface, and hides the new behavior behind a feature flag.

EDIT: Feature flag has already been created in configcat and set to `off` for all environments.

### DIFF
--- a/execute/executetest/dependencies.go
+++ b/execute/executetest/dependencies.go
@@ -33,6 +33,7 @@ var testFlags = map[string]interface{}{
 	"vectorizedMap":                    true,
 	"optimizeAggregateWindow":          true,
 	"narrowTransformationLimit":        true,
+	"optimizeStateTracking":            true,
 }
 
 type TestFlagger map[string]interface{}
--- a/execute/table/chunk.go
+++ b/execute/table/chunk.go
@@ -110,7 +110,7 @@ func (v Chunk) Ints(j int) *array.Int {
 
 // Uints is a convenience function for retrieving an array
 // as a uint array.
-func (v Chunk) UInts(j int) *array.Uint {
+func (v Chunk) Uints(j int) *array.Uint {
 	return v.Values(j).(*array.Uint)
 }
 
@@ -124,12 +124,6 @@ func (v Chunk) Floats(j int) *array.Float {
 // as a string array.
 func (v Chunk) Strings(j int) *array.String {
 	return v.Values(j).(*array.String)
-}
-
-// Times is a convenience function for retrieving an array
-// of timestamps as integers
-func (v Chunk) Times(j int) *array.Int {
-	return v.Values(j).(*array.Int)
 }
 
 // Retain will retain a reference to this Chunk.
--- a/execute/table/chunk.go
+++ b/execute/table/chunk.go
@@ -110,7 +110,7 @@ func (v Chunk) Ints(j int) *array.Int {
 
 // Uints is a convenience function for retrieving an array
 // as a uint array.
-func (v Chunk) Uints(j int) *array.Uint {
+func (v Chunk) UInts(j int) *array.Uint {
 	return v.Values(j).(*array.Uint)
 }
 
@@ -124,6 +124,12 @@ func (v Chunk) Floats(j int) *array.Float {
 // as a string array.
 func (v Chunk) Strings(j int) *array.String {
 	return v.Values(j).(*array.String)
+}
+
+// Times is a convenience function for retrieving an array
+// of timestamps as integers
+func (v Chunk) Times(j int) *array.Int {
+	return v.Values(j).(*array.Int)
 }
 
 // Retain will retain a reference to this Chunk.
--- a/stdlib/universe/narrow_state_tracking.go
+++ b/stdlib/universe/narrow_state_tracking.go
@@ -1,28 +1,46 @@
 package universe
 
 import (
+	"context"
+
 	"github.com/apache/arrow/go/arrow/memory"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/compiler"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/table"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/values"
 )
 
-func NewNarrowStateTrackingTransformation(t *stateTrackingTransformation, id execute.DatasetID, mem memory.Allocator) (execute.Transformation, execute.Dataset, error) {
-	a := &narrowStateTrackingTransformationAdapter{t: t}
-	nt, d, err := execute.NewNarrowStateTransformation(id, a, mem)
+func NewNarrowStateTrackingTransformation(ctx context.Context, spec *StateTrackingProcedureSpec, id execute.DatasetID, mem memory.Allocator) (execute.Transformation, execute.Dataset, error) {
+	fn := execute.NewRowPredicateFn(spec.Fn.Fn, compiler.ToScope(spec.Fn.Scope))
+	t := &narrowStateTrackingTransformation{
+		ctx:      ctx,
+		fn:       fn,
+		timeCol:  spec.TimeCol,
+		countCol: spec.CountColumn,
+		durCol:   spec.DurationColumn,
+		unit:     int64(spec.DurationUnit.Duration()),
+	}
+	nt, d, err := execute.NewNarrowStateTransformation(id, t, mem)
 	if err != nil {
 		return nil, nil, err
 	}
 	return nt, d, nil
 }
 
-type narrowStateTrackingTransformationAdapter struct {
-	t *stateTrackingTransformation
+type narrowStateTrackingTransformation struct {
+	ctx context.Context
+	fn  *execute.RowPredicateFn
+
+	timeCol,
+	countCol,
+	durCol string
+
+	unit int64
 }
 
 type trackedState struct {
@@ -36,14 +54,18 @@ type trackedState struct {
 	duration int64
 }
 
-func (a *narrowStateTrackingTransformationAdapter) Process(chunk table.Chunk, state interface{}, d *execute.TransportDataset, mem memory.Allocator) (interface{}, bool, error) {
+func (n *narrowStateTrackingTransformation) Process(chunk table.Chunk, state interface{}, d *execute.TransportDataset, mem memory.Allocator) (interface{}, bool, error) {
 	// Track whether or not the state has been modified
 	mod := false
 
+	// Check for empty table chunk
 	if chunk.Len() == 0 {
-		return state, mod, nil
+		chunk.Retain()
+		err := d.Process(chunk)
+		return state, mod, err
 	}
 
+	// Initialize state
 	if state == nil {
 		state = trackedState{
 			count:    -1,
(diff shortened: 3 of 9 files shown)


nathanielc (approved):

execute/executetest/dependencies.go
```diff
@@ -32,6 +32,7 @@ var testFlags = map[string]interface{}{
 	"optimizeUnionTransformation":      true,
 	"vectorizedMap":                    true,
 	"optimizeAggregateWindow":          true,
+	"optimizeStateTracking":            true,
```
onelson, line 35: Is it desired to leave these toggled on? I just added this for mine, but sort of assumed it shouldn't be committed.

onelson (approved):

execute/executetest/dependencies.go
scbrickley: I have the same hesitation, but it's what I've been asked to do in the past when adding a new feature flag, and it seems to be the trend.

execute/table/chunk.go
```diff
@@ -110,7 +110,7 @@ func (v Chunk) Ints(j int) *array.Int {
 
 // Uints is a convenience function for retrieving an array
 // as a uint array.
-func (v Chunk) Uints(j int) *array.Uint {
+func (v Chunk) UInts(j int) *array.Uint {
```
jsternberg, line 113: This was intentionally named this way to match with arrow's naming convention.

execute/table/chunk.go
```diff
@@ -126,6 +126,12 @@ func (v Chunk) Strings(j int) *array.String {
 	return v.Values(j).(*array.String)
 }
 
+// Times is a convenience function for retrieving an array
+// of timestamps as integers
+func (v Chunk) Times(j int) *array.Int {
```
jsternberg, line 131: Is this needed? It's identical to `Ints` and there's no check for the column type.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/influxdata/flux/array"
+	"github.com/influxdata/flux/arrow"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/values"
+)
+
+func NewNarrowStateTrackingTransformation(t *stateTrackingTransformation, id execute.DatasetID, mem memory.Allocator) (execute.Transformation, execute.Dataset, error) {
```
jsternberg, line 15: The input to this method shouldn't be a non-exposed type. This can either be an unexported function or we can adapt this to what the signature would be in the absence of the implementation detail.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	prevTime values.Time
+
+	countInState,
+	durationInState bool
+
+	count,
+	duration int64
+}
+
+func (a *narrowStateTrackingTransformationAdapter) Process(chunk table.Chunk, state interface{}, d *execute.TransportDataset, mem memory.Allocator) (interface{}, bool, error) {
```
jsternberg, line 39: Can you logically split this method into different methods based on which part of the algorithm is being implemented? A lot of the older transformations had just massive `Process()` implementations and I'd like the refactored ones to use methods a bit more liberally to be readable. It helps general understandability.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	s := state.(trackedState)
+
+	cols := chunk.Cols()
+	fn, err := a.t.fn.Prepare(cols)
+	if err != nil {
+		return s, mod, err
+	}
+
+	// Get the index for the time column
+	timeIdx := execute.ColIdx(a.t.timeCol, cols)
```
jsternberg, line 63: `table.Chunk` has a [dedicated method](https://pkg.go.dev/github.com/influxdata/flux/execute/table#Chunk.Index) for this.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	if timeIdx < 0 {
+		// Do we need to retain and process the unmodified chunk
```
jsternberg, line 65: I don't believe this comment is needed anymore.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	for i := 0; i < nrows; i++ {
+		// Evaluate the predicate for the current row
+		match, err := fn.EvalRow(a.t.ctx, i, chunk)
+		if err != nil {
+			return s, mod, err
+		}
+
+		// Update state duration
+		if a.t.durationColumn != "" {
+			times := chunk.Times(timeIdx)
```
jsternberg, line 90: This should be pulled out into outside of the loop. Since `timeIdx` doesn't change throughout the loop, there isn't a lot of reason for us to access it repeatedly and pay the cost of the type cast.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+			if times.IsNull(i) {
```
jsternberg, line 91: This check can also be lifted. Since it's an error condition that will always happen, you can use the `NullN()` method to figure out if any of the values in the array are null.

https://pkg.go.dev/github.com/apache/arrow/go/v7@v7.0.0/arrow/array#Int64.NullN

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+
+	ncols := chunk.NCols()
+	newCols := make([]flux.ColMeta, 0, ncols+2)
+	newCols = append(newCols, cols...)
+
+	vs := make([]array.Interface, 0, ncols+2)
+	for i := 0; i < ncols; i++ {
+		vs = append(vs, chunk.Values(i))
```
jsternberg, line 143: You will need to retain these arrays before copying them over.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+		stateDurations.Release()
+	}
+
+	buffer := arrow.TableBuffer{
+		GroupKey: chunk.Key(),
+		Columns:  newCols,
+		Values:   vs,
+	}
+	c := table.ChunkFromBuffer(buffer)
+	c.Retain()
```
jsternberg, line 166: This `Retain` isn't needed if you retain in the area above. I think this retain will cause the two new columns to be retained too many times.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,171 @@
...
+	err = d.Process(c)
```
jsternberg, line 167:
```go
if err := d.Process(c); err != nil {
    return nil, false, err
}
return s, mod, nil
```

stdlib/universe/state_tracking.go
```diff
@@ -138,12 +139,18 @@ func createStateTrackingTransformation(id execute.DatasetID, mode execute.Accumu
...
-	cache := execute.NewTableBuilderCache(a.Allocator())
+	mem := a.Allocator()
+	cache := execute.NewTableBuilderCache(mem)
 	d := execute.NewDataset(id, mode, cache)
 	t, err := NewStateTrackingTransformation(a.Context(), s, d, cache)
 	if err != nil {
 		return nil, nil, err
 	}
+
+	if feature.OptimizeStateTracking().Enabled(a.Context()) {
```
jsternberg, line 150: You should move this feature flag check above the creation of the transformation. If you change `NewNarrowStateTrackingTransformation` to take in the context and spec independently, then you'll avoid creating unnecessary objects.

stdlib/universe/state_tracking_test.go
```diff
@@ -356,5 +370,32 @@ func TestStateTracking_Process(t *testing.T) {
 				},
 			)
 		})
+		t.Run(
+			fmt.Sprintf("%s narrow", tc.name),
```
jsternberg, line 374: I'd just update the old test to call into the new code. We're not changing the old code so I don't think we need that test anymore.

jsternberg (changes requested):

execute/table/chunk.go
scbrickley: I modified this, and added the `Times()` method, so that `table.Chunk` could implement the `flux.ColReader()` interface. That way we can pass it directly to `EvalRow()`. 

If there's a better way to do that though, I'm happy to undo this change.

jsternberg: The better way to do it is to do this:

```go
buf := chunk.Buffer()
EvalRow(&buf, ...)
```

That method returns `arrow.TableBuffer` which implements `flux.ColReader`. It's there to facilitate the older code that still uses `flux.ColReader`. I'd prefer to avoid adding methods to `table.Chunk` to match `flux.ColReader` as newer code should use the struct directly rather than the interface.

scbrickley: Makes sense. I'll change it.

onelson:
@scbrickley in the stand-up today, you mentioned _maybe needing to handle the empty table case_.

It seems to me that the testcase for this could probably be identical to the one seen for limit:

https://github.com/influxdata/flux/blob/master/stdlib%2Funiverse%2Flimit_test.go#L42-L67

I think you got some feedback to run the parameterized tests against only the new impl instead of both (like I was doing), which is fair, but I think the most important thing is to make sure the two impls behave the same way. As such, you may need to manually switch between the two impls in your test just to be sure it passes for both.

scbrickley:
Thanks @onelson! That's very helpful

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		ts := values.Time(times.Value(i))
+		if mod, err := a.updateState(&s, match, ts, &counts, &durations); err != nil {
+			chunk.Retain()
+			d.Process(chunk)
+			return s, mod, err
+		}
+	}
+
+	chunk.Retain()
```
jsternberg, line 102: I suspect this is to retain each of the arrays within the chunk, but it comes across as a different meaning. Can you retain the individual arrays at the point where you copy them instead of retaining it here?

stdlib/universe/narrow_state_tracking.go
jsternberg: This is the proper place to call `Retain()` rather than on the chunk itself. This is mostly just because it's easier to validate by eye that the retain is proper. You can see here, "Oh, the array was accessed and then it was stored so this retain was necessary."

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,237 @@
...
+		stateDurations.Release()
+		stateDurations = nil
+	}
+
+	buffer := arrow.TableBuffer{
+		GroupKey: chunk.Key(),
+		Columns:  newCols,
+		Values:   vs,
+	}
+	c := table.ChunkFromBuffer(buffer)
```
jsternberg, line 180: You don't need two lines here.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,237 @@
...
+	buf := chunk.Buffer()
+	*times = *buf.Times(timeIdx)
+
+	if a.durCol != "" {
+		if times.NullN() > 0 {
+			return errors.New(codes.FailedPrecondition, "got a null timestamp")
+		}
+	}
+
+	*counts = *array.NewIntBuilder(mem)
```
jsternberg, line 134: This is a C-ism that we should likely avoid. It doesn't seem like these are used in here, but if it logically makes sense to do this here, remember that it's ok for this method to call another method.

When we want to break up a process method into multiple methods logically, we can do that through a function that calls another function. If it makes sense to initialize the builders here, we can use Go's support for multiple return values to return the state we want to pass to another function. Both of these are valid. So, for example, this function might return the two builders, the prepared function, and an error if an error happens.

jsternberg (changes requested):

jsternberg (changes requested):
This looks good. Just need an answer to the one question and then I'll hit approve.

stdlib/universe/narrow_state_tracking.go
```diff
@@ -0,0 +1,211 @@
...
+	duration int64
+}
+
+func (n *narrowStateTrackingTransformation) Process(chunk table.Chunk, state interface{}, d *execute.TransportDataset, mem memory.Allocator) (interface{}, bool, error) {
+	// Track whether or not the state has been modified
+	mod := false
+
+	// Check for empty table chunk
+	if chunk.Len() == 0 {
+		chunk.Retain()
```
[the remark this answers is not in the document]
scbrickley: That question hadn't occurred to me, and I'm honestly not sure what the answer is. Does the next node expect to receive a chunk with a certain schema? Where would I look to find out?

jsternberg (approved):

PR #4527, opened 2022-03-03
refactor: Rewrite the parser with lalrpop

Marwes:
Took stab at investigating if we could rewrite the parser with https://github.com/lalrpop/lalrpop which could give us a more readable parser and allow for better recovery of syntax errors. Unfortunately statement lists seem too ambiguous for this to work (at least easily).

```
// is this a single `foo(x)` statement or two statements `foo` and `(x)` ?
foo
(x)
```

### DIFF
--- a/libflux/Cargo.lock
+++ b/libflux/Cargo.lock
@@ -48,6 +48,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
+name = "ascii-canvas"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
+dependencies = [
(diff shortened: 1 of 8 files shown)


onelson:
This is interesting. I've used pest and nom before, but never this one. Pest (PEG-based) felt novel but was very difficult for me to wield/debug.

Nom seemed pretty good, but without the benefit of having a nice grammar DSL (it's closer to parsec, combinator style). The nice thing about nom was being able to write small unit tests around each piece as I went.

Marwes:
The one reason for rewriting the parser right now would be to make it better at dealing invalid code. Currently it is quite easy for the parser to end up failing to parse something and then that first error cascades into more errors which leads to a confusing errors reported to users. I don't quite know how well parser combinators can deal with error recovery as they have a more limited knowledge of what to parse next/what was just parsed. A parser generator knows more about that so it can in theory do a better job at minimizing how many errors are generated.

rust-analyzer uses a manually written parser which may be a useful model though it seems rather complicated in some ways https://rust-analyzer.github.io/blog/2020/09/16/challeging-LR-parsing.html https://rust-analyzer.github.io/blog/2020/10/24/introducing-ungrammar.html

PR #4536, opened 2022-03-08
fix: add parallel merges with no successors to the results set

adrian-thurston:
When a transformation has side effects and it terminates a pipeline there is no
yield added to the pipeline. This is the case with the to() transformation.

https://github.com/influxdata/flux/blob/25c9a281ca27a51672e1e61a840f7eb9caea4d1f/plan/logical.go#L267

When we parallelize a to() transformation that terminates a pipeline, it no
longer terminates the pipeline. The merge node becomes the terminator.

Parallelization happens after yields are added, so we end up with a
non-side-effect node with no successors. The output is then lost because it
isn't added to the set of results.

https://github.com/influxdata/flux/blob/25c9a281ca27a51672e1e61a840f7eb9caea4d1f/execute/executor.go#L288

To handle this case we also add a node to the result set when it is a merge
node with no successors.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


wolffcm (approved):
Looks good! Just a heads up, @onelson is currently looking at an issue related to how we generate results when creating the execution graph.

PR #4537, opened 2022-03-08
fix: Retain the package for identifier referencing the prelude

Marwes:
Noticed the symbols were wrong for identifiers referencing the prelude in the LSP so chased this down.


Marwes:
ping @OfTheDelmer

OfTheDelmer (approved):

PR #4539, opened 2022-03-08
refactor(memory): convert the allocator into an interface

### LINKED ISSUE #4472: Refactor memory allocator to an interface and remove direct references to the current allocator
As part of legacy code, the `ColListTableBuilder` makes a direct reference to `memory.Allocator` and manually manages the bytes that are in use. After we did this, we started using Arrow and started tracking bytes used through passing our allocator as a `memory.Allocator` interface defined in arrow.

Some parts of our code us the interface and some parts still use the tracking struct. The tracking struct is a bit of a problem though as it prevents us from trying out different allocator interfaces or wrapping the allocator with additional behavior like per-transformation memory tracking if we wanted to experiment with something like that.

We should look into the possibility of changing the allocator reference in flux to the arrow interface and removing references to the tracking struct in places where that is possible.

jsternberg:
This converts the allocator to an interface so we can wrap it in
different locations. It expands on the arrow allocator to add a direct
access to the account method of the underlying resource allocator.

The original plan was to separate these two so the allocator would not
have an `Account()` method. This proved to be too difficult as the
`Account()` method is used in some very deep places in the code and
changing all of those locations was just not practical.

Fixes #4472.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -21,9 +21,9 @@ var (
 	BooleanType = arrow.FixedWidthTypes.Boolean
 )
 
-// Interface represents an immutable sequence of values.
+// Array represents an immutable sequence of values.
 //
-// This type is derived from the arrow array.Array interface.
+// This type is derived from the arrow.Array interface.
 type Array interface {
 	// DataType returns the type metadata for this instance.
 	DataType() DataType
--- a/array/array_test.go
+++ b/array/array_test.go
@@ -122,7 +122,7 @@ func TestString(t *testing.T) {
 }
 
 func TestNewStringFromBinaryArray(t *testing.T) {
-	alloc := &fluxmemory.Allocator{}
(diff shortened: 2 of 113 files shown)


onelson (approved):
The diff looks good. Questions for my own edification.

I was a little unsure about when we should be favoring `GoAllocator` versus `ResourceAllocator` or even relying on the `Allocator` interface. Maybe there could be a short explainer in the module-level docstring for `memory`.

In the diff, you change to use `ResourceAllocator` some of the time, but not all of the time. Some text about the rationale might help.

jsternberg:
`GoAllocator` is the name of the same type in arrow that implements a version of the allocator that just uses `make()`. Since we're adding to the interface, it's just creating a version of that that works with our interface.

PR #4540, opened 2022-03-08
fix(stdlib): use fork of gosnowflake library to prevent file transfers

### LINKED ISSUE #3965: gosnowflake dependency prints extraneous log message
We have recently been seeing that flux is causing the following message to happen whenever a binary begins to run.

```
INFO[0000]log.go:104 gosnowflake.(*defaultLogger).Infof reset OCSP cache file. /Users/macbook/Library/Caches/Snowflake/ocsp_response_cache.json
INFO[0000]log.go:104 gosnowflake.(*defaultLogger).Infof reading OCSP Response cache file. /Users/macbook/Library/Caches/Snowflake/ocsp_response_cache.json
```

While the above is a path for Mac OS X, this also appears to happen on Linux too. We believe it is likely caused by the gosnowflake dependency in Flux.

We need to determine if that dependency changed and, if it changed, how we can silence these log messages again. These messages do not seem to affect the actual functioning of anything but the text is not using our own logger and it likely needs to be silenced.

We should also look to see if this message actually indicates a real problem with the gosnowflake dependency before we just disable it.

NOTES:
 - We (influxdata) have a fork of the client with a PR https://github.com/snowflakedb/gosnowflake/issues/530 that allows us to upgrade to a version that does not log the above.
 - We should switch our build process to use our fork of the client
 - This means we should likely merge the PR into master of our fork and build of master of our fork.

DOD:
- The above message does not get printed anymore without requiring the user running it to do outside mitigation.
- All downstream builds of Flux (that we control) use the forked version of the client. Suggestion is to replace the import path to our fork so downstream projects do not have to change.

jsternberg:
The fork of the gosnowflake library adds an option to prevent file
transfers operations from being performed.

Fixes #3965.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/sql/open.go
```diff
@@ -38,6 +38,8 @@ func getOpenFunc(driverName, dataSourceName string) openFunc {
 	switch driverName {
 	case "mssql", "sqlserver":
 		return mssqlOpenFunction(driverName, dataSourceName)
+	case "snowflake":
+		return snowflakeOpenFunction(driverName, dataSourceName)
```
Marwes, line 42: How does `driverName` get here? Is it possible that a user opens a connection to snowflake with `driverName != snowflake`? (Would let them bypass `DisableFileTranfers`)

Marwes (changes requested):
I assume this needs test(s) as well (looking at slack)

stdlib/sql/open.go
jsternberg: I think this is the only way to access it: https://github.com/snowflakedb/gosnowflake/blob/v1.6.7/driver.go#L54

That indicates the library registers itself with the sql library as the "snowflake" driver. I don't think there's another name it can be accessed by. I'm not the most familiar though and it all makes me a bit nervous.

If we're making a fork anyway, maybe we should just consider changing the code itself to statically always return false instead of having it as an option?

onelson: fwiw, there's a lot of code in both `sql.from` and `sql.to` that relies on a fixed list of incoming driver name values. For example: https://github.com/influxdata/flux/blob/00ef526ad50e3fb2e7e9bee7dbcccae9ca22b34b/stdlib/sql/from.go#L116-L137

This specific block for looking up the opener could be made more explicit such that it errors when we see an unexpected name, regardless of how the `DisableFileTransfers` config is handled.

go.mod
```diff
@@ -26,12 +26,12 @@ require (
 	github.com/golang/geo v0.0.0-20190916061304-5b978397cfec
 	github.com/google/flatbuffers v2.0.5+incompatible
 	github.com/google/go-cmp v0.5.7
+	github.com/influxdata/gosnowflake v1.6.8
```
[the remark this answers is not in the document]
jsternberg: Put a comment in the snowflake.go file under notes.

nathanielc (approved):

Marwes (approved):

PR #4541, opened 2022-03-09
test: support for testing parallelization planner rules

adrian-thurston:
Moved helpers for creating plan nodes with attributes into plantest so we can
use those in planner rule tests.

Including output and required physical attributes in the want/got diff in
PhysicalRuleTestHelper. This lets us specify attributes in planner rule tests
for the parallelization rules.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


wolffcm (approved):
Looks good!

Issue #4465, opened 2022-02-09
Simplify generation of yield/result nodes


wolffcm:
We have some technical debt with how we generate implicit yield nodes.

Just before we start planning, we add them in here, if a plan node has no successors and *does not* produce side effects:
https://github.com/influxdata/flux/blob/ee80d230d11c5da9d50b8fa58951d4c61cfc3a8c/plan/logical.go#L265-L278

Later on, when we build the execution graph, we generate an implicit `yield` if there are no successors and the node *does* produce a side effect:
https://github.com/influxdata/flux/blob/ee80d230d11c5da9d50b8fa58951d4c61cfc3a8c/execute/executor.go#L237-L242
(In this case we call it a `result` instead of a `yield`)

It seems clear we need to generate a `yield` (or whatever) for any node that does not have successors. We should do this in the same place every time. My recollection is that at one point we want to implicitly yield only if there were no side effects, but then we decided that we want to yield in the side effect case too. It's confusing to have this contradictory logic like this.

DOD:
- Make it so we generate yield/result nodes in just one place, either the planner or when building the execution graph

onelson:
In an effort to better understand how we arrive at each of the two branches, I cooked up a small repro:

```
import "array"
import "experimental"
import "sql"

emptyFn = () => array.from(rows: [{x: 1}])
write = () => emptyFn() |> sql.to(dataSourceName: ":memory:", driverName: "sqlite3", table: "none")

experimental.chain(first: write(), second: emptyFn())
```

The executor branch is hit while `write()` executes, and the planner branch is hit by the `chain()`.

I'm not yet sure how the refactor should actually go, but by tracing through with the debugger I'm hoping to form some opinions.

onelson:
With breakpoints set in each of the two places we generate implicit yields today, I ran the reproducer (above) through the debugger.

First, we land in the planner branch.

![Screenshot_20220304_111918](https://user-images.githubusercontent.com/301388/156827811-04091754-96de-4877-a6b1-ad57e4bfca88.png)

The `logicalNode` here is named `array.from0`, produced by `emptyFn`.

> In future runs, I may experiment with wrapping the `array.from` calls in differently named functions to help keep better track of which expressions are being visited, and when. I'm sure the detail I need is already available, but I'm still working to better understand the execution path so more landmarks might be helpful.

Frames of (my) interest here are:
- `repl.go:196` where the program unwinds its side-effects
[...]

PR #4545, opened 2022-03-09
refactor: move implicit result generation to executor

onelson:
refs: #4465

Previously, terminal nodes (nodes without children) fell into 3
categories:

- those that _are literal yields_
- those that _have side-effects_
- everything else

In the planner, the first two cases are marked as "roots" which
contribute to the query's concurrecy quota.

The "everything else" case would add an implicit yield to the end of the
chain, which itself is marked as a "root." This essentially funnels the
"everything else" nodes into the "literal yields" category.

Later on in the executor, literal yields generate "results" and
separately, so are the nodes that _have side-effects_.

---

This diff aims to simplify the roles and responsibilities played by each
of the planner and executor.

At a high-level, the diff marks _all terminal nodes_ as "roots" in the
planner. In the executor, we handle the "everything else" nodes
similarly to how we handled the side-effecting nodes, by adding a
"result" for them to the execution state.

~~This seems to work fine for many simple cases but appears to break
things that depend on `TableObjectCompiler` which seems to no longer
produce results as expected. In tests, we'll see panics where we try to
consume the results from a "sub program" but get a `nil` back instead.~~

> Edit: it turned out this was an oversight on my part. The code for generating results we had in place in the executor previously ran only for nodes _that were not leaves_. With the full responsibility lying on the executor now, it needed to also produce results for leaves that are also roots. Details here: https://github.com/influxdata/flux/pull/4545#issuecomment-1064353746

Other tests will need to be adjusted to account for the shape of the
resultant plan having fewer yields in it, or the name of the results
being slightly different.

### Done checklist
- [x] Test cases written

### DIFF
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -159,7 +159,9 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 	// N.b. yields become results here, but terminal nodes are handled via
 	// `generateResult` further below.
 	if yieldSpec, ok := spec.(plan.YieldProcedureSpec); ok {
-		v.generateResult(yieldSpec.YieldName(), node, 0)
+		if err := v.generateResult(yieldSpec.YieldName(), node, 0); err != nil {
+			return err
+		}
 		return nil
 	}
 
@@ -243,7 +245,9 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 			// "if leaf" check.
 			if len(node.Successors()) == 0 {
 				resultName := getResultName(node, spec, isParallelMerge)
-				v.generateResult(resultName, node, i)
+				if err := v.generateResult(resultName, node, i); err != nil {
+					return err
+				}
 			}
 		}
 	} else {
@@ -302,7 +306,9 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 			// "if leaf" check.
 			if len(node.Successors()) == 0 {
 				resultName := getResultName(node, spec, isParallelMerge)
-				v.generateResult(resultName, node, i)
+				if err := v.generateResult(resultName, node, i); err != nil {
+					return err
+				}
 			}
 		}
 	}
@@ -310,10 +316,18 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 }
 
 // generateResult will attach a result to the query for the specified node.
-func (v *createExecutionNodeVisitor) generateResult(resultName string, node plan.Node, idx int) {
+func (v *createExecutionNodeVisitor) generateResult(resultName string, node plan.Node, idx int) error {
+	// if the result name is already present in the result set, that's an error.
+	if _, ok := v.es.results[resultName]; ok {
+		// XXX: we produce an error like this in the planner for duplicate yield
+		// names, but since we're generating results that aren't necessarily from
+		// yields, we need a similar check here.
+		return errors.Newf(codes.Invalid, "tried to produce more than one result with the name %q", resultName)
+	}
 	r := newResult(resultName)
 	v.es.results[resultName] = r
 	v.nodes[skipYields(node)][idx].AddTransformation(r)
+	return nil
 }
 
 // getResultName will offer a "best guess" name for a given node's result.
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -291,6 +291,19 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 		}
 	}
 	// Results should be generated for terminal nodes.
+	//
+	// If user entered `from |> range |> filter` but forgot to add the yield, we
+	// generate a result for them. In this case, the terminal node has no
+	// side-effects and the result name uses the default of `_result` (per the
+	// body of `getResultName()`).
+	//
+	// This is in contrast to other specific cases, like terminal nodes
+	// _with side-effects_ which generate results named for the node itself.
+	//
+	// All queries must have an associated result transformation, otherwise
+	// the query context will be cancelled immediately and execution
+	// will be cancelled before any work for the query has been done.
+	// TODO: understand if this (preventing cancellation) could be addressed by another means.
 	if len(node.Successors()) == 0 {
 		resultName := getResultName(node, spec, isParallelMerge)
 		if err := v.generateResult(resultName, node, 0); err != nil {
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -156,10 +156,10 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 	spec := node.ProcedureSpec()
 	kind := spec.Kind()
 
+	// N.b. yields become results here, but terminal nodes are handled via
+	// `generateResult` further below.
 	if yieldSpec, ok := spec.(plan.YieldProcedureSpec); ok {
-		r := newResult(yieldSpec.YieldName())
-		v.es.results[yieldSpec.YieldName()] = r
-		v.nodes[skipYields(node)][0].AddTransformation(r)
+		v.generateResult(yieldSpec.YieldName(), node, 0)
 		return nil
 	}
 
@@ -236,6 +236,15 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 			source.SetLabel(string(node.ID()))
 			v.es.sources = append(v.es.sources, source)
 			v.nodes[node][i] = source
+
+			// XXX: results should be generated for terminal nodes, regardless
+			// of whether or not it is also a leaf. This is why there's an
+			// identical block at the end of each loop in each case for the
+			// "if leaf" check.
+			if len(node.Successors()) == 0 {
+				resultName := getResultName(node, spec, isParallelMerge)
+				v.generateResult(resultName, node, i)
+			}
 		}
 	} else {
 		// If node is internal, create a transformation. For each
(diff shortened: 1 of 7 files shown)


onelson:
Digging into why some `TableObjectCompiler`-related tests panic, taking some notes here.

Anecdotally, at time of writing the code in this branch seems to have `TableObjectCompiler` working fine (even more correctly) for certain cases I spot-checked such as `experimental.chain()` and `tableFind() |> getRecord()`. The flux acceptance tests (run `flux test`) also all pass, so if there's a bug it might be in the various test helpers we have for dealing with `TableObjectCompiler`. As such, we may have relied on the _incorrect leakage_ of results via the insertion of implicit yields for nodes which should only be seen by the sub program.

---

Looking at specifics, one such panic in the Go test suite surfaces here:
https://github.com/influxdata/flux/blob/master/lang/compiler_test.go#L954

The panic happens when we try to access the results from the query, which end up being nil in this case.

The source for the flux program in this test looks like:

```
import "csv"
data = "#datatype,string,long,dateTime:RFC3339,long,string,string,string,string
#group,false,false,false,false,false,false,true,true
#default,_result,,,,,,,
,result,table,_time,_value,_field,_measurement,host,name
,,0,2018-05-22T19:53:26Z,15204688,io_time,diskio,host.local,disk0
,,0,2018-05-22T19:53:36Z,15204894,io_time,diskio,host.local,disk0
,,0,2018-05-22T19:53:46Z,15205102,io_time,diskio,host.local,disk0
,,0,2018-05-22T19:53:56Z,15205226,io_time,diskio,host.local,disk0
,,0,2018-05-22T19:54:06Z,15205499,io_time,diskio,host.local,disk0
,,0,2018-05-22T19:54:16Z,15205755,io_time,diskio,host.local,disk0
,,1,2018-05-22T19:53:26Z,648,io_time,diskio,host.local,disk2
,,1,2018-05-22T19:53:36Z,648,io_time,diskio,host.local,disk2
,,1,2018-05-22T19:53:46Z,648,io_time,diskio,host.local,disk2
,,1,2018-05-22T19:53:56Z,648,io_time,diskio,host.local,disk2
,,1,2018-05-22T19:54:06Z,648,io_time,diskio,host.local,disk2
,,1,2018-05-22T19:54:16Z,648,io_time,diskio,host.local,disk2
"
csv.from(csv: data)
	|> range(start: 2017-10-10T00:00:00Z, stop: 2018-05-22T19:54:00Z)
	|> filter(fn: (r) => r._value < 1000)
```

Based on the new code in this branch, when walking the graph in the executor (following along in a debugger) I see one node visited named `fromCSV0`. It has no predecessors so it is marked as a leaf (but not a root) and so does not have a result attached to it.

I'd have expected the `range` and `filter` to also be visited, ultimately so that we'd attach a result to the root node (in this case `filter`), but we never do. The query completes execution and we never produce any results (leading to the panic).

---

It's possible this is just a bug introduced by the change. There's a semantic difference here -- the planner didn't either/or based on if a node was a leaf or a root. It only checked to see if the node was a root. Differently, in the executor we only check to see if a node is a root (requiring the addition of a result) if it is not already identified as a leaf...

I'll make some changes to ensure the results are attached for roots even if they are also leaves and see if the suite turns green.

onelson:
After a little iteration on the comments/docs, I think the last item to resolve for this branch is to update existing tests that are looking for implicit yields that are no longer there in the plan.

onelson:
Something I just caught while updating the tests for the new lack of generated yields is we are no longer watching for the case of yield name clashes. Formerly this happened when adding new implicit yields in the planner.

I'll have to see if this check can be added to `generateResult` in the executor now. Hopefully we'll still have the access we need to see the full scope. As far as I know, results should be a flat map for the entire execution cycle so in theory I can just error if the name is already in the map.

plan/logical_test.go
```diff
@@ -282,10 +273,18 @@ func TestPlan_LogicalPlanFromSpec(t *testing.T) {
...
+			// To force the error case we want to test, explicit yields without
+			// specifying names have been added. These should cause 2 yields
+			// using the default name of `_result` to get caught by the planner.
+			name: "multi unnamed yields",
 			query: `
-				from(bucket: "my-bucket") |> sum()
-				from(bucket: "my-bucket") |> mean()`,
+				from(bucket: "my-bucket") |> sum() |> yield()
+				from(bucket: "my-bucket") |> mean() |> yield()`,
 			wantErr: true,
```
onelson, line 288: A new test for the equivalent name collision, but for results, was added on the executor side.

execute/executor.go
```diff
@@ -287,17 +300,49 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
...
+// For nodes that have side-effects or happen to be a Parallel Merge, the result
+// will be based on the node ID. For other cases, the default yield name of
+// `_result` is used.
+func getResultName(node plan.Node, spec plan.ProcedureSpec, isParallelMerge bool) string {
+	name := plan.DefaultYieldName
+	if plan.HasSideEffect(spec) || isParallelMerge {
+		name = string(node.ID())
+	}
+	return name
+}
```
wolffcm, line 344: Is this a departure from what we did before? It seems like this logic is carefully constructed such that it's not.

execute/executor_test.go
```diff
@@ -687,6 +688,49 @@ func TestExecutor_Execute(t *testing.T) {
...
+				},
+			},
+			// XXX(onelson): unsure if the wrapped (inner) error is ever made
+			// visible to the user.
+			// Seems like in this case we'd want the user to get this feedback...
+			wantErr: errors.Wrap(
+				errors.New(codes.Invalid, "tried to produce more than one result with the name \"_result\""),
+				codes.Inherit,
+				"failed to initialize execute state",
+			),
```
wolffcm, line 732: I'm pretty sure that it is, I've seen lots of errors like this.

plan/logical.go
```diff
@@ -263,18 +255,24 @@ func (v *fluxSpecVisitor) visitOperation(o *flux.Operation) error {
...
+		// produce a result by the time the executor is done.
+		//
+		// Now, instead of inserting a yields here we are adding a result
+		// directly in the executor for all terminal nodes. Look for usages of
+		// `generateResult()` in `executor.go` to find where this happens.
+		//
+		// N.b. the number of entries in the `Roots` map factor into the
+		// concurrency quota for the query and as such we must continue to mark
+		// these all as roots here.
+		v.plan.Roots[logicalNode] = struct{}{}
```
wolffcm, line 275: I'm not sure you need a description here of how it used to work. You've simplified it a lot, and one of the benefits of that is less cognitive load on folks reading the code. My hope is that no one ever needs to know how it used to work ever again. If for some reason it becomes important, we have the git history. This is just my subjective take.

Comments like these _can_ often be helpful as a comment on the PR though, so I don't want to discourage you from elucidating what you're doing; I just don't think this stuff needs to live on in perpetuity.

plan/logical_test.go
```diff
@@ -282,10 +273,18 @@ func TestPlan_LogicalPlanFromSpec(t *testing.T) {
...
-			name: "multi-generated yields",
+			// N.b. Formerly this test verified the behavior of yields that were
+			// injected for root nodes during planning, but we are no longer
+			// doing this!
+			// Instead terminal nodes have results produced without yields which
+			// don't show up in the plan.
+			// To force the error case we want to test, explicit yields without
+			// specifying names have been added. These should cause 2 yields
+			// using the default name of `_result` to get caught by the planner.
+			name: "multi unnamed yields",
```
wolffcm, line 284: I've a similar comment here about describing how things used to work.

wolffcm (approved):
This looks great to me. I really appreciate you digging in to the subtle details here.

It would be worth testing your branch against the downstream environments to make sure nothing breaks there, if you haven't already. Let me know if you need pointers on that.

execute/executor.go
onelson: Hard to answer as yes or no. This branching behavior matches the _outcomes_ we saw given the logic up in the planner previously _combined with_ the logic we had in the executor to start.

The old implicit yields were all named using the default yield name. The side-effects and yields that were _not implicit_ used the node id and the name associated with the yield respectively.

When we combine all this with the 3 calls to `generateResult` (yields using their own names, the rest using this logic) we should match the original outcome.

plan/logical.go
onelson: I hear you. I had the same thought but no firm feeling on leaving it in or pulling it out. I'll cut or rephrase the comment to keep it current.

execute/executor_test.go
onelson: Okay great. If it comes to it, I think we could add special handling nearer to the top to try and unwrap the inner for certain cases like this. The chain of custody in having this wrapped for now is probably healthy.

execute/executor.go
onelson: > It seems like this logic is carefully constructed such that it's not.

@wolffcm follow-up question for you on this: is there a different scheme that's been discussed that we'd like to implement?

onelson: Since we handle literal yields separately (and always use the name they carry with them) we should be able to move the `string(node.ID())` calls to the  callsite and remove this function. This would make me happy since it looks odd out of the historical context.

The net change for this would be: _"what were **formerly implicit yields** will no longer use `_result` for their name, instead using the ID of the node that produced them."_

adrian-thurston (commented):
Very nice work! I'm happy to see this simplified! I just have one suggestion based on an observation about parallel execution and results.

execute/executor.go
```diff
@@ -287,17 +300,49 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
...
+
+// getResultName will offer a "best guess" name for a given node's result.
+//
+// For nodes that have side-effects or happen to be a Parallel Merge, the result
+// will be based on the node ID. For other cases, the default yield name of
+// `_result` is used.
+func getResultName(node plan.Node, spec plan.ProcedureSpec, isParallelMerge bool) string {
+	name := plan.DefaultYieldName
+	if plan.HasSideEffect(spec) || isParallelMerge {
```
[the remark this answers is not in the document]
onelson: I [counted my chickens](https://github.com/influxdata/flux/pull/4545#discussion_r826210017) too early here. I'll leave the function in place, with a simplified body and a `TODO` mentioning this point. If you think there's a nicer way to special case the parallel names, we can do away with this function. Otherwise it'll exist mostly to encapsulate whatever you need to accomplish.

execute/executor.go
wolffcm: My reading of the code is that previously we would use `_result` only for nodes with no side effects.

I think the idea here is that if you just had `from |> range |> filter` but forgot to add the yield, we would generate one for you, and give it the default name. I am rethinking my suggestion since this seems like reasonable behavior to me.

I'm looking at the comment that added result generation in the execution graph, here:
https://github.com/influxdata/flux/commit/983a88e597de5ee5666fa23159d83e501ceef4e3

It seems like there's a technical reason that we need to add the result---the would terminate early otherwise. It seems like maybe this was a quick fix to a deep problem. Maybe for that case should add an implicit `debug.sink()`?!? I am only half-joking.

I guess for now I'd be more comfortable leaving the current behavior as-is.

wolffcm: Sorry for the flip-flopping on that... what you have here is still a great simplification! Could I trouble you to add what I wrote as a comment somewhere in `executor.go`?

> If user entered `from |> range |> filter` but forgot to add the yield, we generate one for them, and give it the default name.

> We also generate a result for terminal nodes that have side effects, since all
queries must have an associated result transformation, otherwise
the query context will be cancelled immediately and execution
will be cancelled before any work for the query has been done.
TODO: understand is this could be addressed by another means.

wolffcm: I just want to be sure that we capture the rationale for the naming.

adrian-thurston (approved):
Looks great!

execute/executor.go
onelson: Captured with 4642d255ced6870e6d028bf222b06d1debf3ff5f

wolffcm (approved):
Looks great. Thanks for this.

execute/executor.go
```diff
@@ -287,17 +300,49 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 				}
 			}
 
-			if (plan.HasSideEffect(spec) || isParallelMerge) && len(node.Successors()) == 0 {
-				name := string(node.ID())
-				r := newResult(name)
-				v.es.results[name] = r
-				v.nodes[skipYields(node)][i].AddTransformation(r)
+			// XXX: results should be generated for terminal nodes, regardless
```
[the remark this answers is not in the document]
onelson: Unsure if/when this would not hold true. It could make sense to therefore hardcode the 0th index into the function body instead of exposing it as a parameter, but I guess it's fine as is.

Nice to move it to one spot! Being able to remove a comment to explain something awkward feels good.

adrian-thurston (approved):
Looks great!

FYI, I tried it out this morning in IDPE, in conjunction with parallelization and also without. No issues found.

onelson:
Ah, thanks Adrian. I'll check things in influxdb prior to merging this afternoon.

onelson:
Assuming `go mod edit -replace` did what I expected it to, the influxdb test suite seemed happy with this patch, as did the fluxtest influxdb harness.

PR #4550, opened 2022-03-10
fix(json): dictionary encoding

scbrickley:
Replacement for https://github.com/influxdata/flux/pull/3606, rebased on latest master.

### DIFF
--- a/stdlib/json/encode_test.go
+++ b/stdlib/json/encode_test.go
@@ -25,8 +25,9 @@ o = {
     d: false,
     e: /.*/,
 	f: 2019-08-14T10:03:12Z,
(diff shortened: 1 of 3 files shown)


stdlib/json/encode_test.go
```diff
@@ -25,8 +25,9 @@ o = {
     d: false,
     e: /.*/,
 	f: 2019-08-14T10:03:12Z,
+    g: [1: "hi", 2: "there"]
```
Marwes, line 28: Nit: We could add a dictionary with string keys to the test, just to guard against someone changing `values.Display` to format strings with `"` around it to distinguish `2` from `"2"`.

Marwes (approved):

PR #4553, opened 2022-03-11
feat: Add an IOx specific unpivot

### LINKED ISSUE #4529: Add IOx specific unpivot
We need Flux queries with IOx to work over the SQL endpoint and produce unpivoted data. This is so we can rewrite this query to be backed by IOx:

```
from()
   |> range()
   |> filter()
   |> group()
   |> mean()
```

The plan would look something like:

```
IOx SQL Source -> Unpivot -> Mean
```

Possible type signature and design of the transformation:

```
// unpivot will remove any columns not in the group key or _time and output a new table with _field and _value columns pairs
// B will basically be A - the non group columns
builtin unpivot : (<-tables: stream[{A with _ time: time}]) => stream[{B with _field: string, _value: C}] : where A is Record, B is Record
```

DOD:

- Build specific IOx transformation for unpivoting
- Place `unpivot` builtin in the `internal` package to communicate it is not intended for general use

Marwes:
> We need Flux queries with IOx to work over the SQL endpoint and produce unpivoted data. This is so we can rewrite this query to be backed by IOx:
>
```
from()
   |> range()
   |> filter()
   |> group()
   |> mean()
```
> The plan would look something like:
>
> IOx SQL Source -> Unpivot -> Mean

Closes #4529

### Done checklist
- [ ] Test cases written


stdlib/internal/debug/debug.flux
```diff
@@ -52,3 +52,9 @@ builtin feature : (key: string) => A
 
 // vectorize controls whether the compiler attempts to vectorize Flux functions.
 option vectorize = false
+
+// unpivot will remove any columns not in the group key or _time and output a new table with _field and _value columns pairs
+// B will basically be A - the non group columns
+// Specialized to transform the pivoted output from `iox` into the unpivoted format
```
sanderson, line 58:
```suggestion
// unpivot removes the `_time` column and any other column not in the group key and outputs a new table with `_field` and `_value` columns pairs.
// The output stream retains the group key and all group key columns of the input stream.
// Specialized to transform the pivoted output from `iox.from()` into the unpivoted format.
```

sanderson (approved):
Just a doc suggestion, but I'll go ahead and approve this 👍🏻

stdlib/internal/debug/unpivot.go
```diff
@@ -0,0 +1,159 @@
...
+			break
+		}
+	}
+
+	if timeColumn == -1 || chunk.Cols()[timeColumn].Type != flux.TTime {
+		return errors.Newf(codes.Internal, "Expected a `_time` column in the input")
+	}
+
+	for i, c := range chunk.Cols() {
+		// for i := len(chunk.Cols()) - 1; i >= 0; i-- {
```
skartikey, line 103: it looks like commented code was left by mistake here.

skartikey (approved):

PR #4556, opened 2022-03-12
feat: add mqtt docker integration test

### LINKED ISSUE #4518: Add MQTT integration test(s) w/ Docker
In the same vein as the RDBMS integration tests, build one or more tests that run flux against a live MQTT broker which is spun up in a docker container.

To verify the MQTT client behavior, you may need to run tests that create resources to be read in a subsequent test. This is how all the RDBMS tests operate; tests for writing happen first, then separately tests read what was written.

DOD:

- update `etc/spawn-dbs.sh` to _also spin up and configure an MQTT container_.
- add new tests to interact with MQTT.
- update the `test-flux` Makefile target to skip the new MQTT tests.
- update the `test-flux-integration` Makefile target to run the new MQTT tests.
- **also update skip lists** for _other environment_ CI as needed.
- *optionally* update all the references around "DB" to be more general since MQTT is not a DB.

onelson:
Fixes #4518 

The heft of this diff:

- configures an mqtt dialer dependency for test
- adds a new integration test for `mqtt.publish`
- adds the new test to the group of "write" tests listed in `Makefile` so that the test can be included/excluded by various test make targets
- adds an mqtt broker service to the script we use to launch docker containers for the integration tests

Additionally, there are fixes for the bash script which had previously undergone some refactors that left it working only for some cases (ref: https://github.com/influxdata/flux/issues/4518#issuecomment-1067053402)

Another set of PRs will be needed to update the skip lists for various CI jobs outside of the flux repo.

### Done checklist
- [x] Test cases written


stdlib/experimental/mqtt/mqtt_test.flux
```diff
@@ -0,0 +1,28 @@
...
+                {
+                    ok:
+                        mqtt.publish(
+                            broker: "tcp://127.0.0.1:1883",
+                            topic: "test/topic",
+                            message: "smoke test",
+                            qos: 0,
+                            retain: false,
+                            clientid: "fluxtest",
+                        ),
```
nathanielc, line 21: This looks pretty straightforward, re our conversation in slack, did you end up playing any games with retain etc to ensure the message was published? Reading this test looks like we are relying on the client to report success/failure instead of validating through the server. Is that correct? 

This is likely good enough because we are using a standard client and we don't need to test if the server itself is working rather just that our client can successfully communicate with it.

onelson: The issue I ran into is we have no API in flux to subscribe or otherwise read from MQTT, so there's no reasonable way to verify the outcome from within fluxtest alone. You'd need to be subscribes/watching in another process, or... something else.

Proving the dialer pooling behavior for that other issue might be a challenge, but as a manual exercise we can at least watch the logs on the server to look at client connects/disconnects.

For an automated test, the best I could come up with was to assert we got the `true` back from the client.

This test _does show an error if MQTT isn't running_, but I don't know what the rest of the failure space looks like :shrug:

onelson: Addendum: I did use the spawner script to launch the broker and was able to watch (via a cli subscriber) messages hit the queue while using `mqtt.publish` from the REPL. (video in slack)

onelson:
@nathanielc any other questions I can answer or changes to be made here?

onelson:
@jsternberg wanted to ping you (for when you get back from holiday). We ran into some [test failures](https://app.circleci.com/pipelines/github/influxdata/flux/12382/workflows/bd7d76a8-15d4-4409-9561-6bba4c3f4145/jobs/41907) in nightly (and here) with the new "closer" dialer.

https://github.com/influxdata/flux/pull/4568 was our best guess for what should happen to make the dialer safe to use in test, but you might have a different idea for where to handle this. Seemed like this call needs to happen after the rest of the deps have been injected while also as close to the top of the call stack as possible.

PR #4559, opened 2022-03-14
feat(test): adds support for options in testacase extension

nathanielc:
The current testcase system does not allow extending a testcase that uses an option. As such we have to do odd workarounds when options are involved. We want testcases to be top level statements so option can be manipulated without concern.

This PR is an attempt at changing the design of how testcases AST transformation is performed such that options are possible.

The current implementation used anonymous functions to encapsulate the parent test case statements. This breaks options because options cannot live inside a function block. The user facing change in this PR is two fold:

1. The extends syntax must specify both a path to a test file and a testcase name within that file.
2. Instead of calling the testcase by name you must use the `testing.do()` sentinel call to indicate where the parent test case should be run. 

The implementation now uses the testcase name to extract only the preamble (i.e. statements outside any testcase statement) and the specific test case statements. These are then merged into a single file like so
```
// Test case file

// Merged imports from test case file and parent test case file

// test case preamble

// parent test case preamble

// test case statements before testing.do call

// parent test case statements

// test case statements after testing.do call

```

For example:

```
// parent_test.flux
package parent_test

import "csv"

option preamble = 0
testcase foo {
     option x = 2
     testing.diff(want, got)
}
```

```
// case_test.flux
package case_test

import "array"

testcase bar extends "parent_test.foo" {
     option before = 1
     testing.do()
     option after = 3
}
```

The above two files are transformed into

```
package main

import "csv"
import "array"

option preamble = 0

option before = 1

option x = 2
testing.diff(want, got)

option after = 3
```

With this change we no longer have any anonymous functions or abstraction through a record. Rather the generated AST is a simple concatenation of the various statement blocks.

NOTE: Use `-vv` to have the test harness output the generated source code for a given test case.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


onelson:
Sort of apropos that this likely addresses the inconsistency cited in https://github.com/influxdata/flux/pull/4488, which in turn relates to the recent work in https://github.com/influxdata/flux/issues/4465

onelson (commented):
Seems unlikely to me but would we ever need something to see through aliases for `testing.do()`? 

Feels like it'd be a heavy lift to do, and possibly needless if folks can just do the thing we'd prefer (ie, not alias it).

ast/testcase/testcase.go
```diff
@@ -105,137 +101,120 @@ func Transform(ctx context.Context, pkg *ast.Package, modules TestModules) ([]st
...
+								}
+							}
+						}
+					}
+				}
+			}
+			file.Body = append(file.Body, s)
+		}
+		if !found {
+			return nil, errors.New(codes.Invalid, "test extenstion does not contain call to 'testing.do'")
```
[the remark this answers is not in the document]
nathanielc: Yeah I hadn't thought of that, I'll change this behavior.

onelson:
> Its a trade off from the fragility of the testing.do transformation to the simplicity of pure statement set copies instead of anonymous functions.

I think it's fine. There are plenty of cases out there where strict naming conventions are required in order to achieve an effect. If someone is desperate to alias this one particular thing, I'm sure we can offer a workaround for their situation.

rockstar:
We had talked about doing this API style but I can't, for the life of me, remember why we didn't do it this way. It _might_ have been for ergonomics, e.g. `testing.do()` isn't as clear as `package.function()`, which would mean that this could probably go forward. I've asked @jsternberg to take a look as well, as he might remember specifically what it was. I have very strong memories of pairing with him for a few days straight as we bikeshedded how we should approach this.

nathanielc:
@rockstar  Yeah I have the same memory of you both working on it, hence why the ping for review. If you end up remembering why this path wasn't chosen please share. Otherwise thoughts on this approach? Given what you know now is this a better approach than the current implementation?

ast/testcase/testcase.go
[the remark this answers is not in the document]
jsternberg: `test extension` there's a typo in the error message.

ast/testcase/testcase.go
```diff
@@ -105,137 +101,117 @@ func Transform(ctx context.Context, pkg *ast.Package, modules TestModules) ([]st
...
+		file.Imports = mergeImports(file.Imports, parentImports)
+		file.Body = append(file.Body, parentPreamble...)
+		// Copy test case statements into body replacing the testing.do statement
+		// with the parent test case statements
+		found := false
+		for _, s := range tc.Block.Body {
+			if !found {
+				if es, ok := s.(*ast.ExpressionStatement); ok {
+					if call, ok := es.Expression.(*ast.CallExpression); ok {
+						if member, ok := call.Callee.(*ast.MemberExpression); ok && member.Property.Key() == "do" {
```
jsternberg, line 128: To simplify this, since this is a find/replace, I would just use `super()` as a raw name. It's common for `super()` to have a special meaning in a programming language and specifying `testing.do()` makes me want to try to find the function. Since the function doesn't exist, I think that's confusing.

stdlib/testing/testing.flux
```diff
@@ -446,3 +446,11 @@ assertEqualValues = (got, want) => {
...
             |> yield()
 }
+
+// do is a placeholder function that errors if called.
+// It is used to indicate where the parent test case should be applied.
+// The test harness replaces calls to testing.do as such it should never actually be called.
+//
+// ## Metadata
+// introduced: NEXT
+builtin do : () => bool
```
jsternberg, line 456: I'd just get rid of this.

jsternberg (changes requested):
This seems fine. If you push back on my comments I'll just change my review to approve, but I'm putting request changes because I think that these changes would improve things.

I think `testing.do()` is a bit confusing as that function can't be used outside of that context. I also find the matching to be a bit onerous.

I think `super()` has a better meaning. It's common for that function name to be used to call a parent function in inheritance and it's well known that function can't be called outside of that context since it's not a real function.

rockstar (approved):

jsternberg (approved):

ast/testcase/testcase.go
```diff
@@ -38,30 +38,26 @@ import (
 // The syntax for extending is as such:
 //
 //     import "math"
-//     testcase addition_v2 extends "math_test" {
+//     testcase addition_v2 extends "math_test.addition" {
 //         option math.enable_v2 = true
-//         math_test.test_addition()
+//         testing.do()
```
jsternberg, line 43: Need to update this comment.

Issue #4471, opened 2022-02-10
Parser treats 24h00m as duration + integer + duration


nathanielc:
The parser doesn't expect whitespace around duration or integer literals, which is probably correct. However this has a weird edge case where a typo in a duration literal will get parsed into a valid Flux script but that will have a very different meaning than what was intended.

For example

```
x = 24h01m
y = 07h01m
```

That text is parsed as a duration literal (24h) an integer literal (0) and a duration literal (1m). This is clearly not what the user intended and it is a surprising result. 

We should find a way to make this an error or otherwise inform the user about their typo.

DOD:
 - Find sane solution to parsing the above, potentially make it valid and make sense or potentially make it an error or any other sane solution. Sane meaning is doesn't violate the principal of least surprise. 
 - Add/Update test

jsternberg:
This is interesting. I agree this is probably not ideal. We might just want to change the scanner grammar to be a bit more lenient on what it considers to be an integer literal to include leading zeros. At the moment, this is the scanner definition for integer literals: `"0" | (digit - "0") digit*`.

This is to prevent `0123` from being an integer literal. The scanner definition, in my opinion, is correct. But, it has this unfortunate side effect where `00` is considered to be two integer literals instead of just an error.

I propose we change the integer literal definition to `digit+` and we catch this error when parsing the integer literal.

The reason I'm mentioning integer literal rather than duration literal is that duration literals are defined as `int_lit duration_unit` and this same use case would happen for integer literals too.

Marwes:
When I was arguing for leading zeroes that was under the assumption that `61m` would be invalid due to being out of bounds (specify it as `1h1m` instead), but that seems to not be the case and each number should be treated as an arbitrary integer. So making a leading zero be an error may make more sense then.

onelson:
Maybe I misunderstood here, but since we're talking about duration, not clock times, there shouldn't be any upper bounds (based on clocks, anyway). I'd expect `8h120m` to be equivalent to `10h` as well as `600m` (regardless of leading zeros).

jsternberg:
Whoever picks up this issue is responsible for creating a meeting for us to decide the proper way to fix this. The implementation will follow from whatever we decide should be allowed/disallowed.

PR #4561, opened 2022-03-15
fix: duration and int literal scanning

OfTheDelmer:
Part of #4471 

### Background

The scanner defined `int_lit` and thus `duration_lit` in such a way that they could not begin with a `0`, in anticipation of octal support one day. This awkwardly lead to surprises where `0` before an `int_literal` or `duration_magnitude` were treated as separate tokens.  

### Solution

We decided to scan `int_lit` and `duration_lit` with `digit+` and then handle invalid forms when converting the integer or duration magnitude. This change handles parsing an integer literal that starts with `0` and more than 1 digit as an error. 

This change also allows duration literals to be scanned with leading zeros, but does not error for leading zeros.

### Done checklist
- [ X ] docs/SPEC.md updated
- [ X ] Test cases written


libflux/flux-core/src/parser/mod.rs
```diff
@@ -1488,23 +1494,28 @@ impl<'input> Parser<'input> {
...
         let t = self.expect(TokenType::Int);
+
+        if t.lit.starts_with('0') && t.lit.len() > 1 {
+            return Err(TokenError { token: t })
+        }
+
         match (&t.lit).parse::<i64>() {
             Err(_e) => {
                 self.errs.push(format!(
                     "invalid integer literal \"{}\": value out of range",
```
nathanielc, line 1507: Why do we return an error from this function instead of push the onto the `self.errs` vector like is already done here?

OfTheDelmer: Oh true, I was caught up adding a small guard and making it clear it could have an error result with the error token, similar to float and duration literal parsing, but stepping back, it looks like just pushing to the errs vector with a descriptive message would be more helpful too. I could push an err message like `"invalid integer literal \"{}\": leading zero not supported"` or something particular to the cause.

rockstar (approved):

internal/parser/strconv_test.go
```diff
@@ -0,0 +1,167 @@
...
+				},
+				{
+					Magnitude: 7,
+					Unit:      "ns",
+				},
+			},
+		},
+		{
+			testName: "Many leading zeros duration",
+			duration: "000000000001234d",
```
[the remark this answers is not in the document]
OfTheDelmer: Yeah, the WG decided to allow an arbitrary number of leading `0`s instead of playing with crafting error messages for restricting to just 1 `0\d+` or `0` or `00`

OfTheDelmer:
@Marwes So the [internal/parser/strconv_test.go](https://github.com/influxdata/flux/pull/4561/files#diff-d0f6e83aebc99934a8909539a151da05d6491eff50bb551840410d778512d5ba) file was supposed to target the `internal/parser`, so I could make changes to `internal/parser/strconv.go`. It turned out that the `internal/parser` already had the expected duration handling behavior in master. I left these `strconv_test`  tests to demo the expected duration parsing behavior works and conforms to how we handle things on the rust side.

PR #4563, opened 2022-03-15
feat: connect lsp feature through the flux crate

rockstar:
The `lsp` feature was added previously to the `flux-core` crate, but it
wasn't exposed via `flux`. This patch exposes that, and also adds one
more `Into` implementation to simplify `flux-lsp` a bit.

Please note: there were a number of dependencies that could be updated
in `flux` and `flux-core`, but as this was driven by a specific need, I
felt it necessary to make this a separate change.


OfTheDelmer (approved):

rockstar:
Whoa. I guess I re-used an old branch name, and it appeared to use the previous PR info. I have updated it to reflect the new changes.

PR #4566, opened 2022-03-16
fix(testcase): makes testcase a semantic check instead of an error

nathanielc:
Now the AST -> Semantic convert step does not error on testcase
statements, rather an explicit semantic check is used to catch remaining
testcase statements. The TestCase semantic node behaves like a partial
file where it has a statement list instead of a block. A block is an
expression that evaluates to a value which is not appropriate for a
testcase.

Fixes https://github.com/influxdata/flux-lsp/issues/403 because the LSP
already skips semantic checks and so it doesn't report a diagnostic
about a testcase statement.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


wolffcm (approved):
Looks good!

PR #4567, opened 2022-03-16
fix: skip parallel merge nodes when selecting the result name

adrian-thurston:
If a node is a parallel merge then it was added by the planner and might be a
successor of a node with side effects. In order to preserve behaviour, skip
these parallel merge nodes and check if the predecessor has side effects. If
so, use that node name instead.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/executor.go
+++ b/execute/executor.go
@@ -305,7 +305,10 @@ func (v *createExecutionNodeVisitor) Visit(node plan.Node) error {
 	// will be cancelled before any work for the query has been done.
 	// TODO: understand if this (preventing cancellation) could be addressed by another means.
 	if len(node.Successors()) == 0 {
-		resultName := getResultName(node, spec, isParallelMerge)
+		resultName, err := getResultName(node, spec, isParallelMerge)
+		if err != nil {
+			return err
+		}
 		if err := v.generateResult(resultName, node, 0); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


execute/executor.go
```diff
@@ -330,15 +333,28 @@ func (v *createExecutionNodeVisitor) generateResult(resultName string, node plan
...
+// cases, the default yield name of `_result` is used.
+func getResultName(node plan.Node, spec plan.ProcedureSpec, isParallelMerge bool) (string, error) {
 	name := plan.DefaultYieldName
-	if plan.HasSideEffect(spec) || isParallelMerge {
+
+	// Parallel merge nodes are added by the planner and can mask the presence
+	// of a side-effect result; skip them.
+	if isParallelMerge {
+		if len( node.Predecessors() ) != 1 {
+			return "", errors.New(codes.Internal, "parallel merge must have a single predecessor")
```
onelson, line 347: I think you mentioned possibly being able to enforce this during planning. The check here should probably be here regardless, but I'm wondering if there'll be a follow-on to this follow-on.

onelson (approved):
lgtm! :shipit:

execute/executor.go
adrian-thurston: Yes, was planning on adding to the planner attributes a way to say that the node with the attribute must have at least one successor, or that it must have at least one predecessor.

PR #4568, opened 2022-03-17
fix: add span to context for all dependency tests

scbrickley:


onelson:
Thanks for this @scbrickley !

PR #4581, opened 2022-03-22
refactor: Prefer iterating over record type instead of recursing

Marwes:
Iteration is usually faster and it reduces the direct dependency on the exact shape of the `Record` enum, allowing it to be changed more easily.


nathanielc (approved):
LGTM, seems like a positive change even if we do not end up changing the Record type.

PR #4593, opened 2022-03-24
fix(stdlib): check length boundary for lower bound of substring

jsternberg:
This causes the substring function to check the lower bound to make sure
it's not greater than the length of the string. Previously, this was
just sometimes working because strings will sometimes have a larger
capacity than the actual length of the string.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
Looks right to me. We're already clamping `b <= len(v)` so I think that covers all the cases.

I thought maybe we might need to add a branch to clamp the min `b` to zero, but `a` is already clamped to min 0 so I think we're all good.

sunbryely-influxdata (approved):

Issue #4486, opened 2022-02-15
Proof of concept a memory enforcement solution using finalizers


wolffcm:
See related discussion in #445.

Requires that #4472 be implemented first.

The basic idea here is to write a proof-of-concept that uses [`SetFinalizer`](https://pkg.go.dev/runtime@go1.17.7#SetFinalizer) on the `[]byte` returned by an `Allocator` to decrement the bytes allocated when GC runs.

The solution/investigation here should focus on these requirements:
- Use finalizers to decrement memory in use by an allocator
- Provide total allocated and/or max allocated in profiler output (perhaps no changes are needed here0
- Be able to kill queries that use too much memory
- No longer any need to retain/release in Flux runtime code

DOD:
- The described solution is implemented on a branch
- We can start discussing and or testing (thorough testing will be required to merge to master)

jsternberg:
[This issue](https://github.com/influxdata/flux/issues/4472) is a prerequisite for this one.

Marwes:
> The basic idea here is to write a proof-of-concept that uses [SetFinalizer](https://pkg.go.dev/runtime@go1.17.7#SetFinalizer) on the []byte returned by an Allocator to decrement the bytes allocated when GC runs.

A bit of a wrinkle immediately, `SetFinalizer` only works with pointers so we would need to set the finalizer on the `array.Float/array.Int` pointers themselves.  Will have to add shims to all the ways we construct arrays though to inject the `SetFinalizer` call.

PR #4597, opened 2022-03-25
feat: Track freed memory with SetFinalizer

Marwes:
Instead of manual `Retain` and `Release` calls. I left those calls in right now, but since `Free` no longer actually frees the memory we can still see this works by observing the actual free being don in the finalizer.

As we can see from the repeated `GC` calls (and from being forced to manually GC in the first place), the maximum memory in use at any time will be larger and garbage collection may only clear up part of what is unreachable.

cc #4486

### DIFF
--- a/array/array_test.go
+++ b/array/array_test.go
@@ -144,6 +144,9 @@ func TestNewStringFromBinaryArray(t *testing.T) {
 
 	a.Release()
 	s.Release()
+
+	alloc.GC()
+
 	if want, got := int64(0), alloc.Allocated(); want != got {
 		t.Errorf("epxected allocated to be %v, was %v", want, got)
 	}
--- a/array/array_test.go
+++ b/array/array_test.go
@@ -121,8 +121,27 @@ func TestString(t *testing.T) {
 	}
 }
 
+type A struct {
+}
+
+type B struct {
+	A
+}
+
+type C interface {
+	C() string
+}
+
+func (*A) C() string {
+	return "A"
+}
+
+func (*B) C() string {
+	return "B"
+}
+
 func TestNewStringFromBinaryArray(t *testing.T) {
-	alloc := &fluxmemory.ResourceAllocator{}
+	alloc := &fluxmemory.GcAllocator{ResourceAllocator: &fluxmemory.ResourceAllocator{}}
 	// Need to use the Apache binary builder to be able to create an actual
 	// Arrow Binary array.
 	sb := apachearray.NewBinaryBuilder(alloc, array.StringType)
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/influxdata/flux
 
-go 1.16
+go 1.17
 
 require (
 	cloud.google.com/go v0.82.0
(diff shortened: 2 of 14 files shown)


Marwes:
> Be able to kill queries that use too much memory

Is the one requirement which is harder to fulfill as we will allocate more before freeing it (though, the earlier code didn't actually free memory either, it just decremented the counter).

The rest of the requirements in #4486 seems solved by this (if we remove all the retain and release calls).

lang/compiler.go
```diff
@@ -280,14 +280,20 @@ func (p *Program) Start(ctx context.Context, alloc memory.Allocator) (flux.Query
...
+	if !ok {
+		resourceAlloc = &memory.ResourceAllocator{
+			Allocator: alloc,
+		}
+	}
+
 	q := &query{
 		ctx:     cctx,
 		results: results,
-		alloc: &memory.ResourceAllocator{
```
Marwes, line 286: This were creating a ResourceAllocator  that is wrapping a ResourceAllocator which seems redundant at best, but was causing errors due to setting multiple finalizers on the same pointer.

memory/allocator.go
```diff
@@ -235,6 +247,76 @@ func (a *ResourceAllocator) allocator() memory.Allocator {
...
+
+func (a *GcAllocator) MaxAllocated() int64 {
+	return a.ResourceAllocator.MaxAllocated()
+}
+
+func (a *GcAllocator) TotalAllocated() int64 {
+	return a.ResourceAllocator.TotalAllocated()
+}
+
+// Forces "all" unreachable memory to be garbage collected
```
onelson, line 306: The quotes around _all_ are doing some work here, per the interior comments :laughing:

go.mod
```diff
@@ -59,3 +59,69 @@ require (
 	google.golang.org/grpc v1.44.0
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+require (
+	cloud.google.com/go/bigquery v1.8.0 // indirect
+	github.com/Azure/azure-pipeline-go v0.2.3 // indirect
+	github.com/Azure/azure-storage-blob-go v0.14.0 // indirect
```
onelson, line 66: Unsure of what prompted the 2nd require block here. Was it intended? I didn't see any new imports beyond `runtime` and `unsafe` which I assumed are part of the stdlib.

onelson (approved):
Some questions, but this looks good to me. Might be good to get someone else more familiar with Go internals to check this over, but on its face it makes sense to me as written.

go.mod
onelson: Huh. I didn't think we used `vendor` on this project. I'm lost, but this doesn't feel right to me :shrug:. More research required, but I bet  @jsternberg would have a quick thought on this.

jsternberg: We don't use vendor in the flux repository. You would use `go mod tidy` instead.

jsternberg: I think this second block is something that was added in go 1.17 as I see it in the influxdb repository.

jsternberg:
I didn't get to review this before it was merged, but there were changes I was going to request. In particular, since this is behind a feature flag, I would have liked if it didn't modify any of the tests and I would have asked for some changes that made the `StatsAllocator` interface unneeded. I don't think `GcAllocator` needs to expose these methods and I don't think it should.

Marwes:
Sorry, didn't see that you assigned yourself to this as well. I think you can still review it though and I can do a followup tomorrow.

array/array_test.go
```diff
@@ -121,8 +121,27 @@ func TestString(t *testing.T) {
 	}
 }
 
+type A struct {
```
jsternberg, line 124: These should be removed I think.

memory/allocator.go
```diff
@@ -26,6 +28,13 @@ type Allocator interface {
 	Account(size int) error
 }
 
+type StatsAllocator interface {
```
jsternberg, line 31: This stats allocator interface isn't necessary. Anywhere that needs stats should have direct access or be the owner of the resource allocator and should keep a reference to it. It's a bit of an anti-pattern to define an interface to forward method calls down the chain if we can avoid it.

lang/compiler.go
```diff
@@ -280,14 +280,27 @@ func (p *Program) Start(ctx context.Context, alloc memory.Allocator) (flux.Query
...
 	results := make(chan flux.Result)
+
+	resourceAlloc, ok := alloc.(*memory.ResourceAllocator)
+	if !ok {
+		resourceAlloc = &memory.ResourceAllocator{
+			Allocator: alloc,
+		}
+	}
+
+	var statsAlloc memory.StatsAllocator
```
jsternberg, line 291: This is the correct area to inject the GcAllocator, but I think it would be better to do this in a different way. We can define an extra field on the struct.

```
type query struct {
    alloc *memory.ResourceAllocator
    mem memory.Allocator
}
```

When we initialize this without the feature flag, they point to the same struct. When we enable the `GcAllocator`, `mem` will be pointed towards the version of `GcAllocator` that wraps another allocator.

memory/allocator.go
```diff
@@ -235,6 +247,76 @@ func (a *ResourceAllocator) allocator() memory.Allocator {
 	return a.Allocator
 }
 
+type GcAllocator struct {
+	*ResourceAllocator
```
jsternberg, line 251: With the above comment, you can change this to `mem memory.Allocator`.

For this, I would avoid using embedding. Embedding is great for some circumstances, but it automatically exposes whatever you embedded and it would be good to control exactly how the methods are implemented.

memory/allocator.go
```diff
@@ -235,6 +247,76 @@ func (a *ResourceAllocator) allocator() memory.Allocator {
...
+func (a *GcAllocator) MaxAllocated() int64 {
+	return a.ResourceAllocator.MaxAllocated()
+}
+
+func (a *GcAllocator) TotalAllocated() int64 {
+	return a.ResourceAllocator.TotalAllocated()
+}
+
+// Forces "all" unreachable memory to be garbage collected
+func (mem *GcAllocator) GC() {
```
jsternberg, line 307: I'm not completely sure what use this is but it's definitely an anti pattern. If this is only meant for testing, I'd rather this be defined in some testing code or have `*testing.T` as a parameter. I'm afraid someone may accidentally invoke this method in production code.

array/array_test.go
```diff
@@ -121,8 +121,27 @@ func TestString(t *testing.T) {
...
+	return "A"
+}
+
+func (*B) C() string {
+	return "B"
+}
+
 func TestNewStringFromBinaryArray(t *testing.T) {
-	alloc := &fluxmemory.ResourceAllocator{}
+	alloc := &fluxmemory.GcAllocator{ResourceAllocator: &fluxmemory.ResourceAllocator{}}
```
jsternberg, line 144: Can you revert all of the test changes? I don't think it adds to these tests. I'd be up for adding specific tests for `GcAllocator` but I don't want these tests to be changed.

lang/compiler.go
Marwes: Seems risky since someone could accidentally use `alloc` thereby bypassing the wrapping allocator, but ok. Named it `allocatorStats` instead to try and make it clearer that it is not to be used as an allocator directly.

jsternberg: Yea that's a good name.

PR #4612, opened 2022-03-29
feat: add cffi feature

rockstar:
When building the wasm from `flux-lsp`, the C extern functions are still
provided as part of the public interface for the wasm, despite the
library not ever using them. By putting this behind a (default) feature
flag, and then depending and this crate without the default features, we
trim about 200k from the final wasm binary size.


Marwes (approved):
I would suggest moving all the functions/impls/etc that are now under the `cffi` feature into their own module and then re-export them from the root. That way there is less risk that we add a new `extern "C"` function and forget to put it under the feature.

```rust
#[cfg(feature = "cffi")]
mod cffi;
#[cfg(feature = "cffi")]
pub use cffi::{func1, func2};
```

PR #4613, opened 2022-03-29
wip: add support to linearRegress to work with groups

nathanielc:
A bug in the planner needed to be fixed. A quick hack was put in place.
We need to create a real fix and add tests.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


nathanielc:
A more complete fix for the planner bug here https://github.com/influxdata/flux/pull/4614

PR #4614, opened 2022-03-29
fix(plan): fix panic with half diamond logical plan

nathanielc:
In the case of a half diamond of logical nodes in a plan the same
logical node could be added to the node stack during traversal, however
the visited map to avoid double traversing the same node only contained
physical nodes as such the second logical node would be visited twice.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/plan/heuristic_planner.go
+++ b/plan/heuristic_planner.go
@@ -126,7 +126,7 @@ func (p *heuristicPlanner) Plan(ctx context.Context, inputPlan *Spec) (*Spec, er
 					nodeStack = append(nodeStack, newNode.Predecessors()[i-1])
 				}
 
-				visited[newNode] = struct{}{}
+				visited[node] = struct{}{}
 			}
 		}
 	}
--- a/plan/heuristic_planner_test.go
+++ b/plan/heuristic_planner_test.go
@@ -118,6 +118,48 @@ func TestPlanTraversal(t *testing.T) {
 			},
 			nodeIDs: []plan.NodeID{"7", "6", "4", "1", "0", "3", "2", "5"},
 		},
+		{
+			name: "half diamond physical",
+			//            2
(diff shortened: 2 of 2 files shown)


plan/heuristic_planner.go
```diff
@@ -126,7 +126,7 @@ func (p *heuristicPlanner) Plan(ctx context.Context, inputPlan *Spec) (*Spec, er
 					nodeStack = append(nodeStack, newNode.Predecessors()[i-1])
 				}
 
-				visited[newNode] = struct{}{}
+				visited[node] = struct{}{}
```
nathanielc, line 129: I wasn't sure if I should be adding both new and old nodes to the visited map. Thoughts?

wolffcm: As background, this traversal starts from the sinks (nodes with no successors) and moves towards the sources, in a depth-first fashion. I still find the term "roots" to be really confusing in this context.

I'm not sure why that order was chosen (I think it was me that chose it) but it seems like we would want to traverse in the opposite direction, from the sources to the sinks, with the additional condition that we don't visit any node until we've visited all its predecessors.

That might be a big change for a bug fix. For this PR, you could keep the traversal direction the same, but only add a node to the nodestack if the current node is the last remaining successor? If that makes sense. I guess in that case you would want to add `newNode` to the `visited` map.

If I understand the problem that's occurring correctly, I think it would solve the issue and also avoid the confusion of any particular node (or new replacement of an old node) appearing more than once in the node stack.

wolffcm (commented):
I had some thoughts. Not 100% sure of the right solution though.

wolffcm (approved):
Thinking this over more, I think this will work right. We might still want to revisit how the planner traverses the graph someday.

PR #4616, opened 2022-03-30
feat: add opentracing spans to the Flux runtime

wolffcm:
This will create a new span for each Flux transformation in a query. There will be one span per for each transformation that is planned (not once for each table or each chunk) so the size of the trace should remain manageable.

### Done checklist
- [x] Test cases written


wolffcm:
One thing I noticed in examining the output for this is that the `parentId`s of the generated spans are all zeros. Perhaps this can be addressed in the future.

jsternberg (approved):

PR #4620, opened 2022-03-31
chore: Catch panics in the FFI

Marwes:
Panicking across an FFI boundary leads to crashes that are hard to understand as the backtrace seem to get suppressed. Better to catch and return a normal error to go and let it deal with it like any other error.

### DIFF
--- a/libflux/flux/src/cffi.rs
+++ b/libflux/flux/src/cffi.rs
@@ -1,4 +1,10 @@
-use std::{ffi::*, mem, os::raw::c_char};
+use std::{
+    any::Any,
+    ffi::*,
+    mem,
+    os::raw::c_char,
+    panic::{catch_unwind, resume_unwind},
+};
 
 use anyhow::anyhow;
 use fluxcore::semantic::flatbuffers::types::{build_env, build_type};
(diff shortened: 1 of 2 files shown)


libflux/flux/src/cffi.rs
```diff
@@ -1,4 +1,10 @@
-use std::{ffi::*, mem, os::raw::c_char};
+use std::{
+    any::Any,
+    ffi::*,
+    mem,
+    os::raw::c_char,
+    panic::{catch_unwind, resume_unwind},
```
rockstar, line 6: When this import style gets nested, I would prefer we use another line for it. It makes it hard to read as it gets more and more complicated, whereas single lines make it clear where something comes from. In this case, can we put `use std::panic::{catch_unwind, resume_unwind};` on a separate line?

rockstar (approved):

libflux/flux/src/cffi.rs
Marwes: Out of laziness (I guess) I tend to prefer keeping imports nested like this since it is easier to add and remove imports. I did a batch reformat a while back but I suppose we could do the inverse of that if that is generally prefered https://github.com/influxdata/flux/pull/4213/files

Marwes:
> Curious about the data race being caught. Since those errors are races they are inherently flaky, maybe we can file a bug for that and get it in the next sprint? Then we can unblock this PR, thoughts?

The tests seems to run in parallel when the race checker is enabled (though all the documentation I can find points towards sequential execution...). Since the tracing test modifies a global variable that is read from other tests I am not sure how to actually have the test run sequentially/independent of the other tests I am not sure how to fix this without just disabling the test when the race checker is on :/

Issue #4585, opened 2022-03-22
Pilot: vectorize one arithmetic operator


onelson:
Pick one (1) of the following operators:

- multiplication
- division
- modulo
- power
- addition
- subtraction

> Note that **addition** has higher complexity than the rest since we overload it for string concatenation. This means it may require extra effort to deliver compared to the rest.

> Note that this issue is specifically about multiplying a field with another field, not a field with a constant.

> Note when the code for this merges preemptively disable the feature flag

DOD:

- provide a vectorized implementation of the selected operator
- guard the implementation with a feature flag (we plan to use the existing feature flag for vectorizing map)
- temporary vectors for intermediate results need to be created and destroyed. (i.e. `a * b * c` the `a * b` temp vector needs to be created and destroyed after its consumed)
- Add Flux acceptance tests that hard codes the feature flag on to ensure we still have correctness

scbrickley:
The rust code that decides what function expressions can or can't be vectorized lives here: https://github.com/influxdata/flux/blob/b34ae446c947d332ec60a27fa72c20eee09c25be/libflux/flux-core/src/semantic/nodes.rs#L1142-L1204

PR #4623, opened 2022-04-01
refactor: Move vectorization to its own module

Marwes:
`nodes.rs` is already huge at 2000+ lines and this keeps the vectorization closer to eachother which ought to be easier to read.

cc #4585


onelson (approved):
Looks great. Seems like this will help with the continued vectorize work :clap:

PR #4636, opened 2022-04-04
feat: Allow go to pass compilation options to Rust

Marwes:
While feature flags seems hard to access from Rust, I figured we could instead add compilation options which we can map from the feature flags on the go side. I chose to serialize the options as JSON  since that seems easier than trying to define real C-structs on both sides.

This isn't tied to feature flags either so we may find other usecases for this.


skartikey:
The change looks good to me. 
Could you please add some tests to check compilation options are getting passed to Rust from Go?
Also, please describe how the code flow would be in case of feature flags something on the rust side.

Marwes:
Pulled some commits from #4641 to pass `VectorizedMap` to Rust.

skartikey (approved):
Nice work!

PR #4647, opened 2022-04-05
fix(fluxdoc): adds support for a _soft_ paragraph

nathanielc:
It is possible for a list item in Markdown to have multiple lines and
also not be a _paragraph_. I am calling these _soft_ paragraphs, this
change adds support to the parameter parsing to consume these _soft_
paragraphs as the entire headline.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


libflux/flux-core/src/doc/mod.rs
```diff
@@ -1294,6 +1299,77 @@ mod test {
         );
     }
     #[test]
+    fn test_value_doc_multiline_headline_no_desc() {
```
nathanielc, line 1302: this test passed without changes to the parser, but I wanted to validate that the normal headline didn't have the same problem as the headline for a parameter within a list item.

libflux/flux-core/src/doc/mod.rs
```diff
@@ -1294,6 +1299,77 @@ mod test {
...
+                    })),
+                ],
+                examples: Vec::new(),
+                metadata: None,
+            },
+            vec![],
+        );
+    }
+    #[test]
+    fn test_value_doc_code_headline_no_desc() {
```
nathanielc, line 1338: same thing here, the test passed without changes to the parser, but I wanted to validate that the normal headline didn't have the same problem as the headline for a parameter within a list item.

Marwes (approved):

PR #4649, opened 2022-04-06
perf: high CPU usage while parsing concurrent flux queries

skartikey:
Issue:
ran newly written analyze bench with instruments on mac to find that the bottleneck is in flux::imports (imports is a map of import path to types of packages) as it was taking maximum time to deserialize flatbuffer package exports.

Fix:
Cache the flux::imports using rust sync::Lazy once and use it in different calls

Results:
Before code changes - 

```
analyze/analyze_mean    time:   [4.8914 ms 4.9750 ms 5.0840 ms]
                        change: [+0.7985% +3.5224% +6.3640%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
```

After the code changes - 

```
analyze/analyze_mean    time:   [67.948 us 68.446 us 69.028 us]
                        change: [-98.627% -98.595% -98.567%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
```

fixes: https://github.com/influxdata/influxdb/issues/23151
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written


libflux/flux/src/cffi.rs
```diff
@@ -370,7 +370,7 @@ fn new_stateful_analyzer() -> Result<StatefulAnalyzer> {
     };
     let imports = match imports() {
```
Marwes, line 371: I think we can use `IMPORTS` here as well.

Marwes: Also in the function `flux_get_env_stdlib`

Marwes: Also in the function `flux_get_env_stdlib`

Marwes (changes requested):

PR #4650, opened 2022-04-06
fix(execute): transformation transport adapter now always invokes finish

jsternberg:
The transformationTransportAdapter would flush any buffers that were
cached and send them to be processed by the next transformation. If an
error occurred while this was happening, it would return an error
instead of invoking finish on the transformation. This caused some
transformations to not be finished when an error occurred and this would
lead to a deadlock.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


wolffcm (approved):
This makes sense. I'm okay with you merging this as-is, but it would be great if there was a unit test.

Are there other ways that a deadlock could occur, or other ways we might neglect to send a `finish` message? Just wondering if deadlocks like this could be prevented in a more general way.

PR #4651, opened 2022-04-07
refactor: Add NewFluxAllocator as an injection point for the GcAllocator

Marwes:
This removes GcAllocator from all tests but still provides a way to inject it temporarily
for testing.

@jsternberg I think something like this is what you wanted?

### DIFF
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -288,19 +288,20 @@ func (p *Program) Start(ctx context.Context, alloc memory.Allocator) (flux.Query
 		}
 	}
 
-	var statsAlloc memory.FluxAllocator
 	if feature.SetfinalizerMemoryTracking().Enabled(ctx) {
-		statsAlloc = &memory.GcAllocator{ResourceAllocator: resourceAlloc}
+		var x = memory.NewGcAllocator(resourceAlloc)
+		alloc = &x
 	} else {
-		statsAlloc = resourceAlloc
+		alloc = resourceAlloc
 	}
 
 	q := &query{
-		ctx:     cctx,
-		results: results,
-		alloc:   statsAlloc,
-		span:    s,
-		cancel:  cancel,
+		ctx:            cctx,
+		results:        results,
+		allocatorStats: resourceAlloc,
+		alloc:          alloc,
+		span:           s,
+		cancel:         cancel,
 		stats: flux.Statistics{
 			Metadata: make(metadata.Metadata),
 		},
--- a/lang/compiler.go
+++ b/lang/compiler.go
@@ -288,7 +288,7 @@ func (p *Program) Start(ctx context.Context, alloc memory.Allocator) (flux.Query
 		}
 	}
 
-	var statsAlloc memory.StatsAllocator
+	var statsAlloc memory.FluxAllocator
 	if feature.SetfinalizerMemoryTracking().Enabled(ctx) {
 		statsAlloc = &memory.GcAllocator{ResourceAllocator: resourceAlloc}
 	} else {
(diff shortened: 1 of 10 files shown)


Marwes:
@jsternberg I think I addressed you comments now

lang/compiler.go
```diff
@@ -288,19 +288,20 @@ func (p *Program) Start(ctx context.Context, alloc memory.Allocator) (flux.Query
 		}
 	}
 
-	var statsAlloc memory.StatsAllocator
 	if feature.SetfinalizerMemoryTracking().Enabled(ctx) {
-		statsAlloc = &memory.GcAllocator{ResourceAllocator: resourceAlloc}
+		var x = memory.NewGcAllocator(resourceAlloc)
```
jsternberg, line 292: Should this method be returning a value compatible with the interface?

memory/allocator.go
```diff
@@ -72,6 +65,17 @@ type ResourceAllocator struct {
 	Allocator memory.Allocator
 }
 
+func NewResourceAllocator(allocator memory.Allocator) *ResourceAllocator {
+	// Avoid nesting multiple ResourceAllocator
```
jsternberg, line 69: It's actually fine if we nest these. This could create an unintended effect. If I write this code:

```
mem := memory.NewResourceAllocator(nestedMem)
fmt.Println(mem.TotalAllocated())
```

That should absolutely print zero.

memory/allocator.go
```diff
@@ -248,19 +252,25 @@ func (a *ResourceAllocator) allocator() memory.Allocator {
 }
 
 type GcAllocator struct {
-	*ResourceAllocator
+	mem *ResourceAllocator
+}
+
+func NewGcAllocator(mem *ResourceAllocator) GcAllocator {
```
jsternberg, line 258: Return a pointer here so it's compatible with the `memory.Allocator` interface.

memory/allocator.go
```diff
@@ -293,28 +303,14 @@ func (a *GcAllocator) Free(b []byte) {
...
-func (a *GcAllocator) Account(size int) error { return a.ResourceAllocator.Account(size) }
-
-func (a *GcAllocator) MaxAllocated() int64 {
-	return a.ResourceAllocator.MaxAllocated()
-}
+func (a *GcAllocator) Account(size int) error { return a.mem.Account(size) }
 
-func (a *GcAllocator) TotalAllocated() int64 {
-	return a.ResourceAllocator.TotalAllocated()
+func (a *GcAllocator) Allocated() int64 {
```
jsternberg, line 308: These functions should be removed I think.

memory/allocator.go
```diff
@@ -248,19 +252,25 @@ func (a *ResourceAllocator) allocator() memory.Allocator {
 }
 
 type GcAllocator struct {
-	*ResourceAllocator
+	mem *ResourceAllocator
```
jsternberg, line 255: This should just be `memory.Allocator`. The location that's interested in `MaxAllocated()` and `TotalAllocated()` should keep their reference to their ResourceAllocator. `GcAllocator` doesn't need to know this.

jsternberg (changes requested):

memory/allocator.go
Marwes: They are needed for testing that we actually free things.

memory/allocator.go
Marwes: No, I need a direct reference to `ResourceAllocator` so I can call `Allocated` etc in tests to verify that we actually free the allocations.

jsternberg: You can hold onto both the `ResourceAllocator` and the `GcAllocator` to do that. There's also the `CheckedAllocator` included in arrow for this purpose.

```go
mem := &memory.ResourceAllocator{}
gc := memory.NewGcAllocator(mem)
```

Then you can check `mem` in the tests.

jsternberg (approved):

PR #4662, opened 2022-04-11
fix: prometheus.scrape should return a stream

### LINKED ISSUE #4657: prometheus.scrape has an incorrect function signature
The [`prometheus.scrape()`](https://github.com/influxdata/flux/blob/master/stdlib/experimental/prometheus/prometheus.flux#L33)  function signature doesn't appear to have been updated with the new `stream` representation and now returns the following error:

```
expected stream[#A] but found [B] (array) (argument tables)
```

For any InfluxDB user that was using InfluxDB 2.1 with a version of Flux pre-stream-update, they're scraping tasks/queries will break when they upgrade to 2.2. Fixing this, will require an InfluxDB 2.2.1 patch.

Marwes:
Fixes #4657


Marwes:
Did a search for other instances of `=> [` but the rest actually returns arrays.

sanderson (approved):

onelson (approved):
Looks good. I know we had to patch a few of these after the stream type was added. I added some testcases based on the examples to try and surface things like this in the future, but I wonder if there's a way we could check these typings against the code underneath in CI?

It'd be great if we could type-check the examples in the docs :thinking:

PR #4663, opened 2022-04-11
chore: Remove the flatbuffer representation for the AST

Marwes:
This does not appear to be used outside of tests that verify that this serialization works so we should just remove it.


wolffcm (approved):
Looks good!

PR #4664, opened 2022-04-11
feat: Accept string literals in the fields of a record type

Marwes:
For https://github.com/influxdata/flux/pull/4388 we need a way to describe a type (label) variable in place of a record field.
As type variables are described as a single, uppercase letter, mimicking that syntax will make it
impossible to describe the literal field name "A", "B", etc. By allowing a string literal in this location one can always use this syntax to describe these single letter field names and also describe any other fields which do not fit into an `identifier`, for example `"field with spaces"`, `"symbols#%^"`.

Based on #4663


nathanielc (approved):
LGTM Can you update the docs/SPEC.md  with the new grammar? I don't need to re-review once that is done.

Issue #4626, opened 2022-04-02
Investigate behavior of `exists` on nil records


scbrickley:
In the new `join` function, it's possible (in the case of outer, left, or right joins) for one of the records passed as an argument to `fn` to be empty or nil. To avoid errors that might arise from operations on such a record, it would be nice to be able to do something like the following:

```flux
join(
    left: l,
    right: r,
    predicate: (l, r) => l.a == r.a,
    fn: (l, r) => {
        // Use `exists` on a record that might be empty/nil
        lvalue = if exists l then l._value else 1
        rvalue = if exists r then r._value else 0
        return {l with _value: rvalue / lvalue}
    },
    method: "outer",
)
```

The goal of this issue is to find out if and how `exists` handles null or empty records. If we find that it does not handle them correctly (i.e., does not evaluate to `false`), then we should fix that.

DOD:
	- `exists` evaluates to `false` when operating on an empty or null record

wolffcm:
I guess it's an open question about what gets passed to the `as` lambda in the case where you have a left outer join when there is no matching record on the RHS of the join. I can envision two possibilities:
- a `null` record.  We don't have any precedent for this to my knowledge. What does that mean? Also, since the join only happens on tables with the same group key, should the group key columns have values in the RHS record or not? 
- a record where all the fields are `null`, perhaps except the group key columns. We can currently do this in our engine.

If we decide to go the latter route, it won't be useful to have expressions like `exists r` where `r` is the RHS record, since it will always be true. That is, unless we decide to add special semantics to `exists` such that it returns `false` for a record that has only `null` fields. I'm not sure we want to do that.

I don't think it's horrible though to have the user do this
```js
join(
    left: l,
    right: r,
    predicate: (l, r) => l.a == r.a,
    fn: (l, r) => {
      v = if exists r._value then r._value / l._value else 0
      return {l with _value: v}
    },
    method: "left",
)
```
This is consistent since `null` values can appear for reasons other than just outer joins (by which I mean full outer as well as left and right outer joins). It's also readable, and doesn't add any changes to the language.

scbrickley:
We've decided that we're not going to make records nullable, because doing so would force users to check for nulls, even if they don't need to for their use case.

Instead, we will populate the missing side of an outer join function with a record where all the group key columns are populated, and all other columns are null, as described in @wolffcm's comment above.

Issue #4631, opened 2022-04-02
Document `join` package


scbrickley:
DOD:
- Doc comments have been added to each item in the new `join` package, and they pass the `check-fluxdoc` CI job
- Those docs have been reviewed and approved by someone on the docs team

Issue #4629, opened 2022-04-02
Add a `join` package to stdlib


scbrickley:
Create a new package in the flux standard library called `join`. The package should expose a builtin function called `join` that implements the sort-merge join algorithm, and supports inner, outer, left, and right joins. The function signature should be as follows:

```
builtin join : (
    <-left: [L],
    right: [R],
    predicate: (l: L, r: R) => bool, // consider renaming this `on`
    fn: (l: L, r: R) => A, // consider renaming this `as`
    method: string, // can be "inner", "outer", "left", "right"
) => [A] where
    A: Record,
    L: Record,
    R: Record
```

In the interest of preserving the performance of `experimental.join()`, the underlying implementation should assume that the input tables are sorted and grouped, and should only ever compare records that have the same group key.

The package should also expose a few other convenience functions: `time()`, `inner()`, `outer()`, `left()` and `right()`, each defined in terms of the `join` builtin. They should be defined as follows:

```
// Functionally identical to `experimental.join()`
time = (left=<-, right, fn, method="inner") => join(
    left: left,
    right: right,
    predicate: (l, r) => l._time == r._time,
    fn: fn,
    method: method,
)

inner = (left=<-, right, predicate, fn) => join(
    left: left,
    right: right,
    predicate: predicate,
    fn: fn,
    method: "inner",
)

outer = (left=<-, right, predicate, fn) => join(
    left: left,
    right: right,
    predicate: predicate,
    fn: fn,
    method: "outer",
)

// You get the idea
left = ( ... ) => join( ... )
right = ( ... ) => join( ... )
```
Any go unit tests or flux acceptance tests that we use to validate the other join implementations should be adapted to validate the behavior of the new `join` functions.

New test cases should be added to confirm the correctness of outer, left, and right joins.

Additionally, the package author should confirm (either through go unit tests or flux acceptance tests) that any known and unresolved bugs in the other join implementations are not present in the new `join`.

It is not necessary to evaluate or compile the `predicate` function in the transformation itself. Rather, assume that the planner rule added in #4625 has run and has passed both of the following to the transformation:
- a bool indicating whether or not this is an equi-join
- a list of tuples, each of which contains a column name from the left table, and the name of the column in the right table that it should be compared to.

If the first is `false`, the join transformation should return an error informing the user that non-equi-joins (i.e., joins on unequal columns) are not yet supported.

If it's `true`, the transformation should use the given column names to join the table streams.

[...]

PR #4665, opened 2022-04-12
feat(planner): Add EquiJoinPredicateRule

### LINKED ISSUE #4625: Create planner rule that can analyze a predicate function
The new join package will expose a function that allows users to define a custom predicate function. Having a planner rule that can determine whether or not that predicate function follows a certain pattern would help us leverage some performance gains and avoid undefined behavior.

Specifically, we want to know whether the predicate function defines an equi-join. In other words: is the function simply comparing the equality of columns in the two input tables?

If it is, the planner rule can output a list of the column names being compared in each table, and pass it on to the transformation. This will save the transformation from having to compile the predicate function, which can be expensive.

For the initial implementation, if the predicate function is doing anything besides an equality check between two columns, that should be considered undefined behavior. In this case, the planner rule should somehow signal that to the transformation, which should then return an error.

If it turns out that adding support for non-equi-join predicate functions is easier than we thought, we can file an issue to do it later.

DOD:
- A planner rule exists that can analyze a predicate function to determine whether or not a predicate function contains any operation besides an equality comparison, and respond accordingly.

scbrickley:
This patch adds a few things:

1. A new planner rule called `EquiJoinPredicateRule` that checks if a `join` predicate defines an equi-join (Closes #4625)
2. Stubs for a new `join` package, which will be fully implemented & documented later (See https://github.com/influxdata/flux/issues/4629 and https://github.com/influxdata/flux/issues/4631)
3. A new check in `runtime.Eval` that panics if builtins are not finalized, with a clear error message describing how to fix the issue. This is the result of the difficulty I had trying to evaluate flux scripts in go unit tests within `stdlib`. The solution was to import `fluxinit/static`

I have one concern with the new planner rule that I could use some feedback on: Originally the intent was to have the `join` transformation check what kind of procedure spec it was passed, and to throw an error if it was not an `EquiJoinProcedureSpec`, but due to the scope of #4625, it was easier to just have the planner rule itself return an error. This may increase the workload for adding other implementations of join later on, but not by much. Still, I wonder if it's better to adjust the rule to not throw an error if it doesn't find an equi-join, and instead just return the original procedure spec.


stdlib/join/equijoin_test.go
```diff
@@ -0,0 +1,309 @@
...
+			name: "multiple comparison same column",
+			flux: `import "join"
+			left = from(bucket: "b1", host: "http://localhost:8086")
+				|> filter(fn: (r) => r._measurement == "a")
+			right = from(bucket: "b2", host: "http://localhost:8086")
+				|> filter(fn: (r) => r._measurement == "b")
+			join.join(
+				left: left,
+				right: right,
+				on: (l, r) => r.a == l.b and r.d == l.b,
```
OfTheDelmer, line 187: Should we explicitly test multiple comparison with an operator error for something like `on: (l, r,) => r.a == l.b and !(r.d == l.b)`?

stdlib/join/join.flux
```diff
@@ -0,0 +1,23 @@
...
+
+// join
+//
+// ## Parameters
+// - left:
+// - right:
+// - on:
+// - as:
+// - method:
+builtin join : (
```
OfTheDelmer, line 13: 🎸

stdlib/join/equijoin.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			rob, err := getObjectName(rhs)
+			if err != nil {
+				walkErr = err
+				return
+			}
+
+			// Each side of the binary expression should reference either the `l` or `r` object,
+			// but they should not reference the same object.
+			if !((lob == "l") != (rob == "l") && (lob == "r") != (rob == "r")) {
```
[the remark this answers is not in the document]
scbrickley: I don't think so. In plain english, the expression in the code reads as `one side is l and the other side is r`

Yours (I think) reads `both sides are not l OR both sides are not r`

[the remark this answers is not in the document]
scbrickley: Yes, I think so.

interpreter/interpreter_test.go
```diff
@@ -902,3 +902,32 @@ func TestStack(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	scope := values.NewScope()
+	scope.Set("r", values.Null)
+
+	intrp := interpreter.NewInterpreter(nil, nil)
+	sideEffects, _ := intrp.Eval(context.Background(), &node, scope, nil)
```
wolffcm, line 926: This is great, and it also doesn't quite capture all of the places where we handle `exists`.

We actually evaluate expressions in two places, in the interpreter as you're doing here, and also in the `Compiler` which is a really confusing name. The `Compiler` is the thing that evaluates, e.g., the lambda that's passed into `map`. You could use an acceptance test to do this, if it doesn't produce an error.

Moreover, there's really two kinds of things that should get tested here:
- When we know the names and types of all the fields statically, e.g., what the input data to a function like `map` is produced with `array.from`
- When we don't know all the fields statically, e.g., when the input data is generated via `csv.from`

If I understand write #4626 would need to test both of these cases for both the interpreter and the compiler.

runtime/runtime.go
```diff
@@ -97,6 +97,11 @@ func (r *runtime) Prelude() values.Scope {
 }
 
 func (r *runtime) Eval(ctx context.Context, astPkg flux.ASTHandle, es interpreter.ExecOptsConfig, opts ...flux.ScopeMutator) ([]interpreter.SideEffect, values.Scope, error) {
+	// Give a helpful error message for developers who are trying
+	// to evaluate flux scripts in go unit tests
+	if !r.finalized {
+		panic("runtime is not finalized - consider importing package fluxinit or fluxinit/static")
+	}
```
wolffcm, line 104: 🎉

stdlib/join/equijoin.go
```diff
@@ -0,0 +1,169 @@
...
+	s := n.ProcedureSpec()
+	spec, ok := s.(*JoinProcedureSpec)
+	if !ok {
+		return nil, false, errors.New(codes.Internal, "invalid spec type on join node")
+	}
+
+	fnBody := spec.On.Fn.Block.Body
+
+	if len(fnBody) != 1 {
+		return nil, false, errors.New(codes.Invalid, "predicate should only contain a single return statement")
```
wolffcm, line 81: Users might be confused by this error, since a valid lambda to get passed in might look something like `(l, r) => l.id == r.id and l.foo == r.bar` and doesn't need an explicit `return` statement. (I believe that the semantic graph will have a `Return` statement for that case, but it can be elided in the syntax)

Also, the message should note that there are some forms of lambda that are just not supported yet, so the user understands that they didn't do something wrong, it's just that we have more work to do.

wolffcm (changes requested):
This looks great.

My only issue is that the error messages aren't what I would want users seeing if they began using the new `join` package in anger. These can be hard to get right, especially when there are lots of cases that we can't yet support. We can work on that in this PR, or you could also file an issue to clean it up later.

stdlib/join/equijoin.go
scbrickley: Ah, I see your point. I'm open to any suggestions you have for the wording. Maybe something like this?

```
some lambda expressions are not yet supported in join predicates - function body should only contain a single logical expression comparing equality of columns from the left and right tables
```

That seems kinda long winded, but I'm having trouble thinking of a more succinct way to communicate the problem.

stdlib/join/equijoin_test.go
scbrickley: Yeah, that's probably a good idea. Adding a testcase for that now.

stdlib/join/equijoin.go
scbrickley: It would be easier if we could link them to some documentation explaining what is and isn't allowed in a predicate.

scbrickley: Maybe once #4631 is finished, we can circle back and update these errors to just link to the generated fluxdoc page for the new join package.

wolffcm: Sure, I like your message, you could also explicitly refer to the name of the parameter:
```
some expressions are not yet supported in the "on" parameter - function body should only contain a single logical expression comparing equality of columns from the left and right tables
```
(I guess the error message will have prefixed with something like `error in function join` or some such)

wolffcm (approved):
This looks good, but I don't think that this PR should close #4626, it should stay open until there are more tests.

scbrickley:
Right. Completely forgot about that. I'll move that test to a separate branch and link the issue in a new PR.

scbrickley:
@noramullen1 What are your thoughts on this PR from a docs perspective? There's a new package, but no real docs. Just the bare minimum to get the fluxdoc CI check to pass. There's an open issue to document this package that we intend to get to later: #4631 

I am operating under the assumption that we don't want to document things that are not ready for users yet, but I'm not sure how you and your team feel about this. What should we do here?

sanderson:
@scbrickley I think it's fine to not include full docs at this point. Just make sure that the existing, bare bones docs include something like: "This package is in actively development and is not currently ready to be used". Same for the functions inside the package.

Just in case someone pulls the docs assets from a Flux version where the `join` package isn't ready yet.

scbrickley:
Thanks @sanderson! Making the changes now.

Issue #4554, opened 2022-03-11
`strings.joinStr` panic when receives nil values


onelson:
Ideally we should not allow nil values to appear in the array given to `string.joinStr`, but in the event this happens we should probably give a type error rather than panicking. 

```
> import "strings"
> import "array"
> import "internal/debug"
> arr = array.from(rows: [{x: 1}])
> arr |> debug.opaque() |> map(fn: (r) => ({r with _field: strings.joinStr(arr: ["mean", r._field], v: "_")}))
Result: _result
Error: panic: unexpected kind: got "invalid" expected "string", trace: goroutine 1 [running]:
runtime/debug.Stack()
        /usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/influxdata/flux/values.unexpectedKind.As({0x203000, 0x5}, {0x1994de0, 0xc0005b55b0})
        /home/onelson/Projects/flux/values/values.go:358 +0x65
github.com/influxdata/flux/values.unexpectedKind.Error({0xc001fbfef8, 0x0})
        /home/onelson/Projects/flux/values/values.go:352 +0x2e
github.com/influxdata/flux/internal/errors.(*Error).Error(0xc002150700)
        /home/onelson/Projects/flux/internal/errors/errors.go:35 +0x184
fmt.(*pp).handleMethods(0xc000758820, 0x1c62816)
        /usr/local/go/src/fmt/print.go:620 +0x23c
fmt.(*pp).printArg(0xc000758820, {0x1a98700, 0xc002150700}, 0x76)
        /usr/local/go/src/fmt/print.go:709 +0x693
fmt.(*pp).doPrintln(0xc000758820, {0xc0005b5ab0, 0x2, 0xc001637100})
        /usr/local/go/src/fmt/print.go:1169 +0x149
fmt.Fprintln({0x2de8e80, 0xc000010018}, {0xc0005b5ab0, 0x2, 0x2})
        /usr/local/go/src/fmt/print.go:264 +0x4f
fmt.Println(...)
        /usr/local/go/src/fmt/print.go:274
github.com/influxdata/flux/repl.(*REPL).input(0xc00218fb3e, {0xc0021ffc00, 0xc000662130})
        /home/onelson/Projects/flux/repl/repl.go:138 +0x8a
github.com/c-bata/go-prompt.(*Prompt).Run(0xc001942c80)
[...]

PR #4666, opened 2022-04-13
fix(stdlib/strings): strings.joinStr panic when receives nil values

skartikey:
After the fix, the below query returns a type error rather than panicking.
```
import "strings"
import "array"
import "internal/debug"
arr = array.from(rows: [{x: 1}])
arr |> debug.opaque() |> map(fn: (r) => ({r with _field: strings.joinStr(arr: ["mean", r._field], v: "_")}))
```
```
Result: _result
Error: runtime error @5:26-5:109: map: failed to evaluate map function: expected elements of argument "arr" to be of type string, got type string value <nil>
```
fixes: https://github.com/influxdata/flux/issues/4554 https://github.com/influxdata/EAR/issues/3066

### DIFF
--- a/stdlib/strings/strings.go
+++ b/stdlib/strings/strings.go
@@ -451,10 +451,17 @@ func init() {
 					return nil, fmt.Errorf("missing argument %q", "arr")
 				}
 				arr := val.Array()
-				if arr.Len() >= 0 {
+				arrlen := arr.Len()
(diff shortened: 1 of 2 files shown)


nathanielc:
LGTM can you add a test case?

nathanielc (approved):

PR #4667, opened 2022-04-13
feat: Add contrib/mhall119/helpers

mhall119:
Adds contrib/mhall119/helpers with 3 wrapper functions to `yield()` for printing scalars and objects

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Examples
+//
+// ### Yield an option value
+// ```no_run
+// import "contrib/mhall119/helpers"
+//
+// option testData = {testValue: 125}
+// yieldValue(v: testData.testValue)
+// ```
```
sanderson, line 24:
~~~suggestion
// ```
// import "contrib/mhall119/helpers"
//
// option testData = {testValue: 125}
// option now = () => 2022-01-01T00:00:00Z
//
// > helpers.yieldValue(v: testData.testValue)
// ```
~~~

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Metadata
+// tags: 
+//
+defaultYieldValName="value"
+yieldValue = (v, name=defaultYieldValName) =>
```
sanderson, line 30: You don't need to define `defaultYeildValName` as a variable outside of the `yieldValue` function. `"value"` can just be the default for that parameter.

```suggestion
yieldValue = (v, name="value") =>
```

If you did want to make it a package level variable, make it an option (near the top of the package, below the import statements.

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+    array.from(
+      rows: [
+          {_time: now(), _value: v},
+      ]
+    )
+    |> yield(name: name)
+
+// yieldObject lets you display an object in the InfluxDB UI by
+// encoding it as String
```
sanderson, line 39: In Flux, this data type is known as a "record", not an object.

```suggestion
// yieldRecord returns a stream of tables containing a specified record encoded as a string.
// This makes it possible to display records in an InfluxDB visualization.
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Parameters
+// - o: Object to yield
+// - name: Name to pass to the subsequent yield() call.  Default is 'object'.
```
sanderson, line 43:
```suggestion
// - r: Record to yield.
// - name: Yield name. Default is `record`.
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Examples
+//
+// ### Yield an option value
+// ```no_run
+// import "contrib/mhall119/helpers"
+//
+// option testData = {testValue: 125}
+// yieldObject(v: testData)
+// ```
```
sanderson, line 53:
```suggestion
// ### Yield a record value
// ```no_run
// import "contrib/mhall119/helpers"
//
// option testData = {testValue: 125}
// option now = () => 2022-01-01T00:00:00Z
//
// > helpers.yieldRecord(r: testData)
// ```
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Metadata
+// tags:
```
sanderson, line 56:
```suggestion
// tags: outputs
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+// ### Yield an option value
+// ```no_run
+// import "contrib/mhall119/helpers"
+//
+// option testData = {testValue: 125}
+// yieldValue(v: testData.testValue)
+// ```
+//
+// ## Metadata
+// tags:
```
sanderson, line 27:
```suggestion
// tags: outputs
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// option testData = {testValue: 125}
+// yieldObject(v: testData)
+// ```
+//
+// ## Metadata
+// tags: 
+//
+defaultYieldObjName="object"
+yieldObject = (o, name=defaultYieldObjName) =>
```
sanderson, line 59:
```suggestion
yieldRecord = (r, name="record") =>
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+    array.from(
+      rows: [
+          {_time: now(), _value: display(v: o)},
+      ]
+    )
+    |> yield(name: name)
+
+// yieldObject lets you display an object in the InfluxDB UI by
+// encoding it as a JSON string and wrapping it in an Array
```
sanderson, line 68: Wrong function name.

I'm not sure I fully understand the utility of this function. What's the use case? This seems like a round about way to do the same thing as `yieldRecord`.

```suggestion
// yieldJSON returns a stream of tables containing a JSON string.
// This makes it possible to display JSON in an InfluxDB visualization.
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Parameters
+// - o: Object to yield
+// - name: Name to pass to the subsequent yield() call.  Default is 'object'.
```
sanderson, line 72:
```suggestion
// - json: JSON to yield.
// - name: Yield name. Default is `json`.
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Examples
+//
+// ### Yield an option value
+// ```no_run
+// import "contrib/mhall119/helpers"
+//
+// option testData = {testValue: 125}
+// yieldObject(v: testData)
+// ```
```
sanderson, line 82:
~~~suggestion
// ### Yield a JSON value
// ```no_run
// import "contrib/mhall119/helpers"
//
// option testData = {testValue: 125}
// option now = () => 2022-01-01T00:00:00Z
// 
// helpers.yieldJSON(json: testData)
// ```
~~~

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+//
+// ## Metadata
+// tags:
```
sanderson, line 85:
```suggestion
// tags: outputs
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+// tags: 
+//
+defaultYieldJSONName="json"
+yieldJSON = (o, name=defaultYieldJSONName) =>
+    array.from(
+      rows: [
+          {_time: now(), _value: string(v: json.encode(v: o))},
+      ]
+    )
+    |> yield(name: name)
```
sanderson, line 94:
```suggestion
yieldJSON = (json, name="json") =>
    array.from(
      rows: [
          {_time: now(), _value: string(v: json.encode(v: json))},
      ]
    )
    |> yield(name: name)
```

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,94 @@
...
+// ```
+//
+// ## Metadata
+// tags: 
+//
+defaultYieldObjName="object"
+yieldObject = (o, name=defaultYieldObjName) =>
+    array.from(
+      rows: [
+          {_time: now(), _value: display(v: o)},
```
sanderson, line 62:
```suggestion
          {_time: now(), _value: display(v: r)},
```

stdlib/contrib/mhall119/helpers/helpers_test.flux
```diff
@@ -0,0 +1,30 @@
...
+import "testing"
+import "contrib/mhall119/helpers"
+
+option testData = {testValue: 125}
+
+// Test yielding a single scalar value
+t_yieldValue = (v) =>
+    helpers.yieldValue(v:v)
+
+test _yieldValue = () =>
```
nathanielc, line 13: I would follow the `testcase` examples the `test` keyword is the older style. Look here for and example https://github.com/influxdata/flux/blob/master/stdlib/math/math_test.flux#L20

stdlib/contrib/mhall119/helpers/helpers.flux
nathanielc, line 94: As for how to test these with the explicit `yields`. How important is it that these have calls to yield? Seems the real value the function provides is the boilerplate of getting the value displayed in a table. Could the yield be implicit or explicit on the caller side? That generally fits better with the Flux idiomatic code as it makes it easier to compose these functions with other functions.

nathanielc (changes requested):
Looks great, some suggestions/questions below

stdlib/contrib/mhall119/helpers/helpers.flux
mhall119: I don't think there's many scenarios where a developer won't want to yield this data immediately, and since the wrapper functions are already very small the yield itself is like 1/3 of what they do.

I think these helper functions are more valuable if they include the yield themselves.

stdlib/contrib/mhall119/helpers/helpers_test.flux
mhall119: Is there documentation on how to use this? Trying to reverse engineer from examples is proving difficult for me

stdlib/contrib/mhall119/helpers/helpers.flux
mhall119: What does adding `now` here change? As far as I can tell it's not used anywhere.

stdlib/contrib/mhall119/helpers/helpers.flux
mhall119: This function is meant to work with any Flux object, not just records returned by a query. Unless "Record" is the term used to describe any data type that's not a table or a scalar.

stdlib/contrib/mhall119/helpers/helpers.flux
mhall119: This function turns a Flux object into JSON, so the input is an object, not JSON. Naming the input variable `json` may lead to confusion about what kind of data should be passed in.

stdlib/contrib/mhall119/helpers/helpers.flux
sanderson: Your output stream includes a `_time` column that's populated with `now()`. Setting this option will make the output consistent. If you don't set it, `now()` will return the time that the docs are generated. We can and should show the output from this functions, which is why I added the `>`. This appends a yield named "output" at the end of that line.

This also goes to Nathaniel's point using implicit yields. As-is, this example will actually output two yields. One named "value" and the other, "output".

sanderson: Something else to note is that if you don't want the `option now` line to _display_ as part of the example, you can add a `#` at the beginning of the line. For example:

```javascript
// # option now = () => 2022-01-01T00:00:00Z
```

When we display the example, that line won't be included. But it would still be used when generating the output from the example.

stdlib/contrib/mhall119/helpers/helpers.flux
sanderson: I see. These are known as [composite types](https://docs.influxdata.com/flux/v0.x/data-types/composite/) - records, arrays, dictionaries, and functions. And really, with that context, this function can display any Flux type, basic, composite, or regexp. The caveat is that it always gets output as a string.

If the purpose of this package is simply to be able to show something in the single stat viz, this might be the only function you need. I guess the downside would be losing number truncation in the UI.

stdlib/contrib/mhall119/helpers/helpers.flux
nathanielc: @jsternberg Is there a way we can trick the test engine into testing these functions with yields in them?

stdlib/contrib/mhall119/helpers/helpers_test.flux
nathanielc: Not really :( its an activity that the Flux dev team does exclusively ATM, but we are working towards making this more accessible.

stdlib/contrib/mhall119/helpers/helpers.flux
jsternberg: I don't think so.

stdlib/contrib/mhall119/helpers/helpers.flux
```diff
@@ -0,0 +1,95 @@
...
+// - v: Input value.
+// - name: Yield name. Default is `value`.
+//
+// ## Examples
+//
+// ### Yield an option value
+// ```no_run
+// import "contrib/mhall119/helpers"
+//
+// option testData = {testValue: 125}
```
Marwes, line 24:
```suggestion
// testData = {testValue: 125}
```

The option seems redundant here (and in the other tests)

PR #4670, opened 2022-04-13
fix(stdlib): remove an extraneous go generate statement

jsternberg:
This statement must have been used in development but the template file
was ultimately removed. This causes an error when you use go generate.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


rockstar (approved):
Yesssss. The less generated code we have, the better, I say!

PR #4673, opened 2022-04-14
fix(stdlib): fix memory releases and add checked allocator to end to end tests

jsternberg:
This commit adds the checked allocator to the end to end tests to ensure
that memory is actually released. It disables certain feature flags that
haven't been enabled yet that seem to have memory leak issues and
removes a package that has memory leak issues from the contrib package.

The table functions are a notable exception to these tests. The
interpreter was never updated to properly release values so the result
of the table functions does not ever get released when run outside of
the compiler. A special selection of tests have the memory tests
disabled to accomodate this.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


execute/table.go
```diff
@@ -1323,9 +1323,33 @@ func (t *ColListTable) Do(f func(flux.ColReader) error) error {
...
+			case *stringColumn:
+				buffer.Values[i] = col.data
+			case *timeColumn:
+				buffer.Values[i] = col.data
+			default:
+				return errors.Newf(codes.Internal, "unknown column type: %T", col)
+			}
+		}
+		err = f(buffer)
 		t.Release()
```
onelson, line 1353: My reading of this is we're now copying the data out of the incoming table, into a fresh buffer, passing that new buffer to `f()`, and releasing the incoming table.

For my own edification - what's the significance of this? Is it because otherwise we don't know when `t` should be released otherwise?

onelson (approved):

execute/table.go
jsternberg: There was some complicated logic happening with the old version. It implemented both `flux.Table` and `flux.ColReader`, but it caused it act weird if the column reader was retained after the table was read. So I replaced it with just our standard `flux.ColReader` implementation so there weren't any weird data flows.

PR #4674, opened 2022-04-14
feat(cmd): add trace option to the flux cli

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -2,18 +2,24 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 
 	"github.com/influxdata/flux/cmd/flux/cmd"
+	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/dependencies"
 	"github.com/influxdata/flux/dependency"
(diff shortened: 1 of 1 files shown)


nathanielc (commented):
How do you use this? Looks like you need your local env to be configured to point to a jaeger install? Is the intent that it would be an existing production install? Or one inside the isolation testing env? 

Does adding a trace client side associate the trace with the server side trace/spans?

jsternberg:
It's mostly meant to be used with a local jaeger. You run the jaeger docker image and then just use `--trace` and it will upload it to your local one on localhost. Since this is the CLI, there are no server side trace/spans. It's just taking the existing opentracing infrastructure and allowing it to be used from the CLI.

It was originally removed when we removed the opentracing components. Then @wolffcm added better designed ones in and this became useful again.

wolffcm:
Can we include quick instructions (or a link) somewhere on how to set up a local Jaeger? I can see this being useful.

onelson:
https://www.jaegertracing.io/docs/1.33/getting-started/#all-in-one

nathanielc (approved):

PR #4684, opened 2022-04-20
fix: don't set BaseLocation.file to Some("")

rockstar:
A common pattern we have is this:

    parse_string("".to_string(), "from(...) |> range(...) |> filter(...)")

That first argument is supposed to serve as the filename and, in
general, we don't care about it. However, the `BaseLocation.file` has a
`#[serde(skip_serializing_if = "skip_string_option")]` on it, which
means that when we go over the serialization boundary,
`BaseLocation.file` is no longer `Some("")` but `None`. It's probably
not a huge deal, but when the result is inconsistent going over the
serialization boundary, that becomes an issue.

This patch fixes that by only assigning `BaseLocation.file` to a `Some`
value when the string is not empty. Otherwise, assign it as a `None`.

### DIFF
--- a/libflux/flux-core/src/parser/tests.rs
+++ b/libflux/flux-core/src/parser/tests.rs
@@ -16,6 +16,20 @@ mod operator_precedence;
 mod strings;
 mod types;
 
+/// Parsed ast roundtrips across the serde boundary and generates the same ast.
+#[test]
+fn parse_ast_roundtrip() {
+    let mut p = Parser::new(
+        r#"from(bucket: "an-bucket") |> range(start: -15m) |> filter(fn: (r) => r.field == "value")"#,
+    );
+    let ast = p.parse_file("".to_string());
+
+    let serialized = serde_json::to_string(&ast).unwrap();
+    let new_ast = serde_json::from_str(&serialized).unwrap();
+
+    assert_eq!(ast, new_ast);
+}
+
 #[test]
 fn parse_array_expr_no_rbrack() {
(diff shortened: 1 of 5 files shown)


wolffcm (approved):
Very nice.

With this PR, you can serde roundtrip AST and it should be exactly the same yes? If it's not too hard to add a test for that I think it would be a good idea.

rockstar:
> With this PR, you can serde roundtrip AST and it should be exactly the same yes? If it's not too hard to add a test for that I think it would be a good idea.

Yes, exactly. I discovered this when writing a test in `flux-lsp`, so I feel a bit ashamed that I didn't think to add that here. Test is added now. Thanks for the pull in.

libflux/flux-core/src/parser/tests.rs
```diff
@@ -16,6 +16,18 @@ mod operator_precedence;
...
+#[test]
+fn parse_ast_roundtrip() {
+    let mut p = Parser::new(r#"from(bucket: "an-bucket") |> range(start: -15m) |> filter(fn: (r) => r.field == "value")"#);
+    let ast = p.parse_file("".to_string());
+
+    let serialized = serde_json::to_string(&ast).unwrap();
+    let new_ast = serde_json::from_str(&serialized).unwrap();
+
+    assert_eq!(ast, new_ast);
+}
```
wolffcm, line 29: 🎉

wolffcm (approved):

PR #4688, opened 2022-04-21
feat: Vectorize remaining arithmetic operators

### LINKED ISSUE #4586: Vectorize remaining arithmetic operators
Following on from #4585, #4587 which will have selected a single operator to vectorize from the following list:

- multiplication
- division
- modulo
- power
- addition
- subtraction

The work here is to vectorize the remaining items in the list.

> Note that **addition** has higher complexity than the rest since we overload it for string concatenation. This means it may require extra effort to deliver compared to the rest.

DOD:

- provide a vectorized implementation of the remaining operators
- guard the implementation(s) with a feature flag
- temporary vectors for intermediate results need to be created and destroyed. (i.e. a * b * c the a * b temp vector needs to be created and destroyed after its consumed)
- Add Flux acceptance tests that hard codes the feature flag on to ensure we still have correctness

Marwes:
Based on #4641 

Closes #4586


skartikey (approved):
Looks good to me. Kindly open an issue for the non-float type's implementation for the `power operator`(which can be discussed in the planning).

Marwes:
> Looks good to me. Kindly open an issue for the non-float type's implementation for the power operator(which can be discussed in the planning).

https://github.com/influxdata/flux/issues/4695

PR #4690, opened 2022-04-21
feat(stdlib): Add new join function with outer join support

### LINKED ISSUE #4629: Add a `join` package to stdlib

scbrickley:
Closes #4629 

Adds an implementation of sort-merge-join to the existing join function, which was previously just a stub. This implementation supports inner, left, right, and full-outer joins.

There are still some outstanding problems with this patch that I'm actively working on:

* ~There is a known bug where `join.join()` panics if either input table has more than 1000 rows. The panic seems to come from calling `Eval` on the compiled function used by `JoinFn`~ This was an off-by-one error. Fixed [here](https://github.com/influxdata/flux/pull/4690/commits/bf17b7dc42455eac4ec7a62c2130f47c0a54695f)
* ~The transformation does not currently do much of anything to properly account for memory used by arrow arrays~ Got the tests passing with a checked allocator with [this commit](https://github.com/influxdata/flux/pull/4690/commits/6de1e0a00638c9084de7709b34cabbcc81c0069a), though I'm surprised that's all it took.
* Some of the flux acceptance tests pass even though the `got` tables are missing a column present in the `want` tables

I also want to make sure that the code is easier to read and maintain than the previous two joins. If there's any section of code that you think could be better organized, or use more explanatory comments, please point them out.


stdlib/join/merge_join.go
```diff
@@ -0,0 +1,518 @@
...
+	s.chunks[0] = getChunkSlice(c, s.keyStart, c.Len())
+	s.chunks = append(s.chunks, getChunkSlice(c, 0, s.keyEnd))
+
+	// TODO(sean): I think rows may just point to the underlying data of s.chunks, rather than
+	// a copy of it. If so, setting s.chunks to an empty slice will cause problems. Will confirm
+	// this during testing.
+	rows := s.chunks
+	s.chunks = s.chunks[:0]
+	return rows
+}
```
scbrickley, line 336: Just now noticing some major flaws in these two functions. Working on a fix now.

stdlib/join/join_fn.go
```diff
@@ -0,0 +1,304 @@
...
+
+type JoinFn struct {
+	fn       *semantic.FunctionExpression
+	scope    compiler.Scope
+	compiled compiler.Func
+	args     values.Object
+	schema   []flux.ColMeta
+	ltyp     *semantic.MonoType
+	rtyp     *semantic.MonoType
+}
```
wolffcm, line 27: Is `JoinFn` for the `on` or the `as` parameter? It would be good to have a comment here that explains.

wolffcm, line 27: Is `JoinFn` for the `on` or the `as` parameter? It would be good to have a comment here that explains.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+	"github.com/influxdata/flux/internal/errors"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+// MergeJoinTransformation performs a sort-merge-join on two table streams.
+// It makes the following assumptions about the input tables:
+//   * They have the same group key
```
wolffcm, line 22: I guess you mean that they have the same group key columns?

wolffcm, line 22: I guess you mean that they have the same group key columns?

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+func (t *MergeJoinTransformation) initState(state interface{}) (*joinState, bool) {
+	if state != nil {
+		s, ok := state.(*joinState)
+		return s, ok
+	}
+	s := joinState{}
+	return &s, true
+}
+
+func (t *MergeJoinTransformation) processChunk(chunk table.Chunk, state interface{}, id execute.DatasetID) (*joinState, bool, error) {
```
wolffcm, line 134: A high-level description of what happens when you process a chunk would be great here.

wolffcm, line 134: A high-level description of what happens when you process a chunk would be great here.

wolffcm (commented):
Looks good to me so far, I will continue tomorrow.

stdlib/join/merge_join.go
```diff
@@ -471,7 +471,7 @@ func (r joinRows) getRow(i int, typ semantic.MonoType) values.Object {
 	var obj values.Object
 	for _, chunk := range r {
 		if i > chunk.Len()-1 {
-			i -= chunk.Len() - 1
+			i -= chunk.Len()
```
scbrickley, line 474: @wolffcm This is the fix for the panic I was talking about. It was causing `getRow` to return a `nil` object, which was then passed into `JoinFn.eval()`, which caused a panic when the compiler tried to call `Retain()` on `nil`.

https://github.com/influxdata/flux/blob/bf17b7dc42455eac4ec7a62c2130f47c0a54695f/stdlib/join/join_fn.go#L124-L125

scbrickley: It only triggered when I tried to test `join` with large inputs. Before that, all of the input tables I was using to test it were pretty small, so I imagine there was never a reason for this branch to trigger.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+		} else {
+			canJoin = prev.left.len() > 0 && prev.right.len() == 0
+		}
+	}
+	return position, canJoin
+}
+
+func (s *joinState) join(ctx context.Context, method string, fn *JoinFn, joinable int, mem memory.Allocator) ([]table.Chunk, error) {
+	if fn.compiled == nil {
+		err := fn.Prepare(s.left.schema, s.right.schema)
```
wolffcm, line 328: What happens if the schema changes here? `fn` is state that belongs to the transformation (constant across all group keys) while the group key-specific `state` might not have a consistent schema, so `Prepare` might produce a different result for different group keys.

I think some of the other transformations (`map` or `filter`) will keep a cache of compiled functions keyed on the record types in the input.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+	for i := 0; i <= joinable; i++ {
+		prod := s.products[i]
+		p, ok, err := prod.evaluate(ctx, method, *fn, mem)
+		if err != nil {
+			return joined, err
+		}
+		if !ok {
+			continue
+		}
+		joined = append(joined, *p)
```
wolffcm, line 343: I wonder if `evaluate` should produce multiple chunks? It's possible for join to produce many rows if there are many duplicate join keys on both sides---you end up with `ln * rn` rows. Maybe file an issue for it?

wolffcm (commented):
This is really fantastic.

It'd be great to see a summary of your approach written down at a high level somewhere. Maybe I missed it.

Seems like it is:
1) As data arrives keep track of when the join key changes
2) When the join key changes, you know you have one complete set of rows for the previous join key, and you store it in one side of a join product
3) When new data arrives and you end up completing both sides of a join product, you can then produce a table chunk to be processed by downstream transformations

Do I have that right?

Again, nice work, the code is definitely way more comprehensible than before!

stdlib/join/chunk_builder.go
```diff
@@ -0,0 +1,53 @@
...
+		GroupKey: key,
+		Columns:  b.cols,
+	}
+	vals := make([]array.Array, 0, len(b.builders))
+	for _, builder := range b.builders {
+		vals = append(vals, builder.NewArray())
+	}
+	buf.Values = vals
+	return table.ChunkFromBuffer(buf)
+}
```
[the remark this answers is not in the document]
scbrickley: Can do!

stdlib/join/join_fn.go
scbrickley: It's for the `as` parameter. I'll add a clarifying comment.

stdlib/join/merge_join.go
scbrickley: Yeah, I now realize that this was poorly worded. I'll amend it.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+	}
+
+	// We hit the end of the chunk without finding a new join key.
+	// Reset keyEnd to 0 since the next thing we process will be a new chunk.
+	if !complete {
+		s.keyEnd = 0
+	}
+
+	return &startKey, complete
+}
```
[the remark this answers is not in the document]
scbrickley: I just added some unit tests that do something similar to this in my most recent commit. Take a look and tell me if we're on the same page.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+	return s, true, nil
+}
+
+func (t *MergeJoinTransformation) mergeJoin(chunk table.Chunk, s *joinState, isLeft bool) error {
+	for {
+		key, rows := s.merge(chunk, isLeft, t.on)
+		if key == nil {
+			break
+		}
+		i, canJoin := s.insert(key, rows, isLeft)
```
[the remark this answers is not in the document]
scbrickley: Changed it to `scanKey`. Does that work better?

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+// If the bool == true, that means it's safe to join all the items in s.products up to and including
+// the returned position.
+//
+// Returns true under 2 circumstances:
+//   (1) Inserting `rows` completes the left and right pair for a given product
+//   (2) `rows` was inserted at an index greater than 0, and all of the products that
+//        come before it only have entries on the opposite side.
+//
+// If condition 1 is true, we can join everything up to and including the index where
+// `rows` was inserted.
```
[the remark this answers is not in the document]
scbrickley: Yes, that is correct.

stdlib/join/merge_join.go
```diff
@@ -0,0 +1,522 @@
...
+
+	found := false
+	position := 0
+	for i, product := range s.products {
+		if product.key.equal(*key) {
+			if isLeft {
+				product.left = rows
+			} else {
+				product.right = rows
+			}
```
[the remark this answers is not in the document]
scbrickley: The `processChunk` method returns an error if it receives a chunk from a dataset whose id does not match its left or right parents, which is why I felt comfortable not fully checking it here. But I can change it to a panic if you think that would be better.

stdlib/join/merge_join.go
scbrickley: Good looking out! This was a blind spot for me. I'll take a look at how map and filter handle that case and see if I can do something similar here.

scbrickley: Looking at this again, with my most recent commit, the schema is set per-side and per-state when we get a new table chunk to process, and each `joinState` calls `fn.Prepare()` before it starts joining table chunks. So I think this will be resilient to inconsistent schemas in the inputs, unless I'm missing something.

stdlib/join/merge_join.go
scbrickley: I just pushed up a commit that limits each of the output chunks to 1000 rows each. Take a look and let me know if that's what you had in mind.

stdlib/join/merge_join.go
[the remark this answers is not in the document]
scbrickley: Sorry, misunderstood your original comment. Yes, that situation should not be possible. I'll add the panic.

wolffcm (approved):
This looks great, I appreciate all the work that went into this. I just had a couple of comments for you to consider. I'm okay with you merging as-is though.

stdlib/join/join_fn.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/flux/values"
+)
+
+// JoinFn handles the logic of calling the function in the `as` parameter of join.tables()
+type JoinFn struct {
+	fn       *semantic.FunctionExpression
+	scope    compiler.Scope
+	compiled compiler.Func
```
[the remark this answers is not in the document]
scbrickley: `compiled` is still used in the `eval()` method.

There is not an issue for that, but I may be able to take care of that today. Could you point me toward some examples of how other transformations do this? I looked at the transformations you mentioned in your last review (`map` and `filter`) but couldn't find anything. Perhaps I overlooked something though.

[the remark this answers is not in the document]
scbrickley: Oh wait, are you talking about `dynamicFn`?

https://github.com/influxdata/flux/blob/1c65530064697e3b83fb908726c6ff2c4c17a9cd/execute/row_fn.go#L14-L21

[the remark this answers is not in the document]
scbrickley: I created an issue to track the work to fix this, and added it to the join epic: https://github.com/influxdata/flux/issues/4883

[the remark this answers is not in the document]
wolffcm: Ah, yes, I was confused. Thanks for filing the issue.

jsternberg (approved):
I haven't had time to look at the join code, but the overall API from the flux perspective looks good to me and the join code is going to be around awhile so we can do iterative improvements if needed in the future.

I'm looking forward to this! I like it!

PR #4697, opened 2022-04-26
refactor: Remove unnecessary code around type freshening and visitation

Marwes:
My main goal here was to remove the `Fresh` and its duplicated type traversal since it can now be implemented via a custom `Substituter`. Due to how trait objects work I were forced to change some methods to taking `impl Substituter` instead of  `dyn Substituter` (else `Substituter::visit_type` can't call `Substitutable::walk` in the default implementation).

Since I were already changing these type signatures I also changed the reference to a mutable one as there is no reason to share a `Substituter` during traversal (do `impl Substituter for &'a MyType` if you need to share a type). This let me remove all the `RefCell`/`Cell` for the `Substituter`s we have.


libflux/flux-core/src/semantic/types.rs
```diff
@@ -109,21 +110,53 @@ impl PartialEq for PolyType {
             kinds.sort();
         }
 
+        assert_eq!(a.vars, b.vars);
+        assert_eq!(a.cons, b.cons);
+        assert_eq!(a.expr, b.expr);
```
wolffcm, line 115: Maybe you added these assertions for debugging? it seems like you really just want this function to return the expression just below?

wolffcm (changes requested):
Everything here looks like a great cleanup, I just had a question about the `assert_eq!` that seems out of place.

libflux/flux-core/src/semantic/types.rs
Marwes: Woops, fixed.

wolffcm (approved):

Issue #4692, opened 2022-04-21
Panic: index out of range when extracting scalar value


Byter09:
Hello everyone!

I've been using Flux for a bit now and I hope I'm reporting this in the correct repository.

Basically, I managed to cause a panic simply by calling a function that returns a row inside of a map:

```
callsValueBelowPrice = (price) =>
  intrinsic_value 
    |> filter(fn: (r) => r.type == "call")
    |> filter(fn: (r) => float(v: r.price) == price)
    |> filter(fn: (r) => float(v: r.strike) < price)
    |> keep(columns: ["cash_value"])
    |> sum(column: "cash_value")
    |> findRecord(fn: (key) => true, idx: 0)

strikes
  |> map(fn: (r) => ({r with calls_value: callsValueBelowPrice(price: r.price).cash_value}))
  ```

I'm unsure of what I can remove above to reduce the example, but I hope that's okay.

Here's something that _does not_ do what I want, but it doesn't panic:

```
cv = callsValueBelowPrice(price: 100).cash_value

strikes
  |> map(fn: (r) => ({r with calls_value: cv}))
```

As soon as I even try to put this call in the map function in any way, I get a panic.

Even this panics:
```
callsValueBelowPrice = (price) => {
  sum = intrinsic_value 
    |> filter(fn: (r) => r.type == "call")
    |> filter(fn: (r) => float(v: r.price) == price)
    |> filter(fn: (r) => float(v: r.strike) < price)
    |> keep(columns: ["cash_value"])
    |> sum(column: "cash_value")
    |> findRecord(fn: (key) => true, idx: 0)

  return sum.cash_value
}

strikes
  |> map(fn: (r) => ({r with calls_value: callsValueBelowPrice(price: r.price)}))
```

Here's the error I get in the console:
```
ts=2022-04-21T20:09:45.371737Z lvl=info msg="Dispatcher panic" log_id=0_zSXqpl000 service=storage-reads component=dispatcher error="panic: runtime error: index out of range [2] with length 2" stacktrace="goroutine 12548271 [running]:
runtime/debug.Stack()
    /go/src/runtime/debug/stack.go:24 +0x65
github.com/influxdata/flux/execute.(*poolDispatcher).recover(0xc018a923f0)
    /go/pkg/mod/github.com/influxdata/flux@v0.161.0/execute/recover.go:53 +0x205
panic({0x3c103c0, 0xc02e13f7e8})
    /go/src/runtime/panic.go:1038 +0x215
[...]

PR #4698, opened 2022-04-26
fix(universe/join): remove `JoinOpSpec.TableNames` and favor `JoinOpSpec.params` to stay consistent inside `tableFind`

onelson:
Fixes: #4692

When a `join` is used in conjunction with a `tableFind`, the input tables
are copied and given new operation IDs. This causes problems in
`createMergeJoinTransformation` where we rely on a mapping of
`TableNames` to associate the inputs to `join` with the corresponding
datasets.

The assumption violated in this situation is we expect the number of
"parents" to be the same as the number of tracked "table names." When
this does not hold true, we panic. Formerly the panic came from an out
of bounds index, but with this diff this is more explicit.

Ideally, I think it would be better to perform a deep copy on the join's
`JoinOpSpec` for the sub-program we build via `tableFind` such that the
"sub join" is entirely different and can hold references to the "sub
inputs."

In practice this appears out of reach based on the information I saw
available in the code paths that matter. We'd need to add "join aware"
blocks to `TableObject.Operation()`, which seems to be responsible for
the "cloning" of the Operation, or something else entirely.

~~The proposed fix here is to avoid altering the `TableNames` map for a
`JoinOpSpec` when this cloning happens. In effect, once a `JoinOpSpec`
is associated with a pair of input tables, this will be fixed. The
inputs may be cloned by `tableFind` but the cloned inputs  will no longer
be referred to as they are today.~~

## A change in tactics...

~~A previous attempt to fix a panic caused by the `TableNames` map
incorrectly growing when a `JoinOpSpec` is used repeatedly via
`tableFind` had "locked" the map after the initial entries had been
added.~~

~~e83ca5b switches to solve the same problem by instead removing old entries
before adding new ones. This in effect should mean `TableNames` and
`params` should agree with each other.~~

~~While the truth is, we should not be reusing and mutating the OpSpec at
least now the fields that are changing with repeated use will agree with
each other.~~

## Third attempt...

`JoinOpSpec.TableNames` mapped operationIDs to field names for the
`tables` argument to `join`.

Meanwhile `JoinOpSpec.params` held two separate fields:
- an array of strings (the field names of the object)
- an array of "operations" (the tables bound to those fields)

The `params` are in effect separate arrays of the keys/values seen in
the `TableNames` map, but holding references to the input table objects
rather than their operation IDs.

For the purposes of doing `join`'s work, the operation IDs are not
necessary, and in fact causes issues in situations like a when `join`
is used with `tableFind`.

Since the _extra information_ contained in the `TableNames` map is not
needed it's basically redundant.

f57d93bc removes the field in favor of just referring to `params`
directly.

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/universe/covariance_test.go
+++ b/stdlib/universe/covariance_test.go
@@ -88,11 +88,7 @@ func TestCovariance_NewQuery(t *testing.T) {
 					{
 						ID: "join2",
 						Spec: &universe.JoinOpSpec{
-							On: []string{"host"},
-							TableNames: map[flux.OperationID]string{
-								"from0": "x",
-								"from1": "y",
-							},
+							On:     []string{"host"},
 							Method: "inner",
 						},
 					},
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -37,9 +37,8 @@ var methods = map[string]bool{
 
 // JoinOpSpec specifies a particular join operation
 type JoinOpSpec struct {
-	TableNames map[flux.OperationID]string `json:"tableNames"`
-	On         []string                    `json:"on"`
-	Method     string                      `json:"method"`
+	On     []string `json:"on"`
+	Method string   `json:"method"`
 
 	// Note: this field below is non-exported and is not part of the public Flux.Spec
 	// interface (used by the transpiler).  It should not be assumed to be populated
@@ -111,7 +110,6 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 		return nil, err
 	}
 
-	spec.TableNames = make(map[flux.OperationID]string, tables.Len())
 	spec.params = newJoinParams(tables.Len())
 	tables.Range(func(name string, operation values.Value) {
 		if err != nil {
@@ -144,22 +142,6 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 	return spec, nil
 }
 
-func (t *JoinOpSpec) IDer(ider flux.IDer) {
-	// When join is used inside a `tableFind`, the input table operations are
-	// "cloned" granting them new ids.
-	// This will cause a panic to occur in `createMergeJoinTransformation` when
-	// trying to associate table names with their related datasets if we don't
-	// first remove old entries from the map.
-	// Refs: <https://github.com/influxdata/flux/issues/4692>
-	for k := range t.TableNames {
-		delete(t.TableNames, k)
-	}
-	for i, name := range t.params.names {
-		operation := t.params.operations[i]
-		t.TableNames[ider.ID(operation)] = name
-	}
-}
-
 func newJoinOp() flux.OperationSpec {
 	return new(JoinOpSpec)
 }
@@ -182,11 +164,9 @@ func newMergeJoinProcedure(qs flux.OperationSpec, pa plan.Administration) (plan.
 		return nil, errors.Newf(codes.Internal, "invalid spec type %T", qs)
 	}
 
-	tableNames := make([]string, len(spec.TableNames))
-	i := 0
-	for _, name := range spec.TableNames {
+	tableNames := make([]string, len(spec.params.names))
+	for i, name := range spec.params.names {
 		tableNames[i] = name
-		i++
 	}
 	sort.Strings(tableNames)
 
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -145,19 +145,18 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 }
 
 func (t *JoinOpSpec) IDer(ider flux.IDer) {
-	// When join is used inside a `tableFind`, the input tables are cloned
-	// granting them new ids.
-	// This can cause a panic to occur in `createMergeJoinTransformation` when
-	// trying to associate table names with their related datasets.
+	// When join is used inside a `tableFind`, the input table operations are
+	// "cloned" granting them new ids.
+	// This will cause a panic to occur in `createMergeJoinTransformation` when
+	// trying to associate table names with their related datasets if we don't
+	// first remove old entries from the map.
 	// Refs: <https://github.com/influxdata/flux/issues/4692>
-
-	// Treat `TableNames` as "immutable" - don't modify them after initialization.
-	// FIXME: find a "better fix" than only initializing this once.
-	if len(t.TableNames) == 0 {
-		for i, name := range t.params.names {
-			operation := t.params.operations[i]
-			t.TableNames[ider.ID(operation)] = name
-		}
+	for k := range t.TableNames {
+		delete(t.TableNames, k)
+	}
+	for i, name := range t.params.names {
+		operation := t.params.operations[i]
+		t.TableNames[ider.ID(operation)] = name
 	}
 }
 
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -145,9 +145,19 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 }
 
 func (t *JoinOpSpec) IDer(ider flux.IDer) {
-	for i, name := range t.params.names {
-		operation := t.params.operations[i]
-		t.TableNames[ider.ID(operation)] = name
+	// When join is used inside a `tableFind`, the input tables are cloned
+	// granting them new ids.
+	// This can cause a panic to occur in `createMergeJoinTransformation` when
+	// trying to associate table names with their related datasets.
+	// Refs: <https://github.com/influxdata/flux/issues/4692>
+
+	// Treat `TableNames` as "immutable" - don't modify them after initialization.
+	// FIXME: find a "better fix" than only initializing this once.
+	if len(t.TableNames) == 0 {
+		for i, name := range t.params.names {
+			operation := t.params.operations[i]
+			t.TableNames[ider.ID(operation)] = name
+		}
 	}
 }
 
@@ -213,6 +223,22 @@ func createMergeJoinTransformation(id execute.DatasetID, mode execute.Accumulati
 	if len(parents) != 2 {
 		// TODO(nathanielc): Support n-way joins
 		return nil, nil, errors.New(codes.Unimplemented, "joins currently must only have two parents")
+	}
+
+	// In <https://github.com/influxdata/flux/issues/4692>, it was found that in
+	// certain cases `s.TableNames` could grow beyond the size of the parents
+	// _incorrectly_.
+	// This would lead to a panic in the for/range to follow.
+	// To illustrate the requirement in a more explicit way, this check ensures
+	// that the names of the tables, and their associated datasets are of equal counts.
+	if len(parents) != len(s.TableNames) {
+		panic(
+			fmt.Sprintf(
+				"expected parent and table name counts to be equal: %d != %d",
+				len(parents),
+				len(s.TableNames),
+			),
+		)
 	}
 
 	tableNames := make(map[execute.DatasetID]string, len(s.TableNames))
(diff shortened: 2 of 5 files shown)


onelson:
As an aside, I've heard the guidance that we should not be feeding the same tables into `findRecord` repeatedly, which is precisely how the panic was triggered. I'm not sure, but it seems like we could achieve the same effect (ie, not panicking) by enforcing _that_ as a requirement somewhere, but I'm not sure where.

stdlib/universe/join_test.flux
```diff
@@ -41,21 +42,59 @@ outData =
...
             |> group(columns: ["_measurement", "_field"])
 
-    return join(tables: {left: left, right: right}, on: ["_time", "_measurement", "_field"])
+    got = join(tables: {left: left, right: right}, on: ["_time", "_measurement", "_field"])
+
+    testing.diff(want: want, got: got)
 }
 
-test _join = () => ({input: testing.loadStorage(csv: inData), want: testing.loadMem(csv: outData), fn: t_join})
+testcase join_repro_4692 {
```
rockstar, line 67: Swapping `test` for `testcase`. 💯

rockstar (approved):

stdlib/universe/join_test.flux
```diff
@@ -41,21 +42,59 @@ outData =
...
+                // repeated lookups for id=2 should be okay...
+                {_value: 2},
+                {_value: 3},
+            ],
+        )
+            |> map(fn: (r) => getById(id: r._value))
+
+    want = array.from(rows: [{x: 1, y: 1}, {x: 2, y: 2}, {x: 2, y: 2}, {x: 3, y: 3}])
+
+    testing.diff(want: want, got: got)
```
wolffcm, line 99: Can you add another test case and we can see how it goes?
```js
testcase join_repro_4692 {
    xs = array.from(rows: [{id: 1, x: 1, v: "hi"}, {id: 2, x: 2, v: "hi"}, {id: 3, x: 3, v: "hi"}])
    ys = array.from(rows: [{id: 1, y: 1}, {id: 2, y: 2}, {id: 3, y: 3}])

    getById = (id) => {
        zs = if id == 2 then
            join(tables: {xs: xs |> map(fn: (r) => ({r with v: "hey"})), ys: ys}, on: ["id"])
        else
            join(tables: {xs: xs, ys: ys}, on: ["id"])
            
        r = zs |> filter(fn: (r) => r.id == id) |> findRecord(fn: (key) => true, idx: 0)
        return {x: r.x, y: r.y, v: r.v}
    }

    got =
        array.from(
            rows: [
                {_value: 1},
                {_value: 2},
                // repeated lookups for id=2 should be okay...
                {_value: 2},
                {_value: 3},
            ],
        )
            |> map(fn: (r) => getById(id: r._value))

    want = array.from(rows: [{x: 1, y: 1, v: "hi"}, {x: 2, y: 2, v: "hey"}, {x: 2, y: 2, v: "hey"}, {x: 3, y: 3, v: "hi"}])

```
I am sure that there are typos in this example but the basic idea is that the function that returns a result from `findRecord` will produce a graph whose shape depends on the parameter `id` at runtime. That is to say, the input to the join when `id == 2` will make include `map`, and you can tell if `map` was correctly executed when `v` changes from `hi` to `hey`.

My suspicion is that you're right about needing to do a deep copy, and this new test case would fail as a result.

wolffcm (changes requested):
Can you add the test case I mentioned? I'm concerned that your fix could lead to wrong answers, which would be worse than just panicking.

stdlib/universe/join_test.flux
onelson: The branching `getById` version is currently passing...
I wondered if this is only because (as written) the joins are defined within the function call, and so are 1:1 with the `findRecord` calls.

I tried moving the joins out to the script root and used the branching to alternate between those variables instead, but that also passed.

Even if these specific cases are passing, it does seem like a deep copy of the spec is desired -- if for no other reason than to justify the copies of the table inputs. I'm just not sure how to achieve this based on the flow we have today.

wolffcm: What happens if you do the join inside of the call to map?

onelson: e01529b defines the joins inside the map fn (assuming I understood you correctly). Still passes. :shrug:

onelson:
Been studying the larger flow to try and figure out if there's a place where we can make the deep copy happen as an alternative approach. I'm mostly at a loss -- unsure of how to make it happen.

Seems like what _should happen_ is we should deep copy the entire table object used to run `tableFind` here:

https://github.com/influxdata/flux/blob/ead736ee677d08df1df73a4aba097f28ae126022/stdlib/universe/table_fns.go#L87-L90

I just don't see the API to do this anywhere in the supporting `TableObject` code.

onelson:
Still looking for alternative approaches that would get us the same result without "locking" that `TableNames` when `IDer` runs repeatedly.

I started to wonder why the `params` field seemed to be "correct" in each situation whereas the `TableNames` field was incorrectly growing.

As a sanity check, I wanted to also verify that `TableNames` wasn't somehow just being copied-then-mutated (which would have led us to the same panic). I added some prints:

```diff
diff --git a/stdlib/universe/join.go b/stdlib/universe/join.go
index 7a807d66..c1b30ffb 100644
--- a/stdlib/universe/join.go
+++ b/stdlib/universe/join.go
@@ -153,12 +153,15 @@ func (t *JoinOpSpec) IDer(ider flux.IDer) {
 
        // Treat `TableNames` as "immutable" - don't modify them after initialization.
        // FIXME: find a "better fix" than only initializing this once.
-       if len(t.TableNames) == 0 {
-               for i, name := range t.params.names {
-                       operation := t.params.operations[i]
-                       t.TableNames[ider.ID(operation)] = name
-               }
+       fmt.Printf("t=%p\tt.TableNames=%p\tt.params=%p\n", t, t.TableNames, t.params)
+       for k := range t.TableNames {
+               delete(t.TableNames, k)
+       }
+       for i, name := range t.params.names {
+               operation := t.params.operations[i]
+               t.TableNames[ider.ID(operation)] = name
        }
+       fmt.Println(t.TableNames)
 }
 
 func newJoinOp() flux.OperationSpec {

```

When I run my repro script, the output shows:

```
Result: _result
t=0xc001590e00  t.TableNames=0xc00157ccf0       t.params=0xc00157cd20
map[array.from2:xs array.from3:ys]
t=0xc001590e00  t.TableNames=0xc00157ccf0       t.params=0xc00157cd20
map[array.from6:xs array.from7:ys]
t=0xc001590e00  t.TableNames=0xc00157ccf0       t.params=0xc00157cd20
map[array.from10:xs array.from11:ys]
t=0xc001590e00  t.TableNames=0xc00157ccf0       t.params=0xc00157cd20
map[array.from14:xs array.from15:ys]
t=0xc001590e00  t.TableNames=0xc00157ccf0       t.params=0xc00157cd20
map[array.from18:xs array.from19:ys]
Table: keys: []
                     x:int                       y:int  
--------------------------  --------------------------  
                         1                           1  
                         2                           2  
                         2                           2  
                         2                           2  
                         3                           3  

```

From this we can see the memory addresses for the `JoinOpSpec`, as well as its `TableNames` and `params` field are consistent each time `IDer()` is called.

For the `params` field, it appears we are simply overwriting the names/operations it holds. This makes it consistent with _only the most recent run_.

The range/delete loop I added in the above patch gives the `TableNames` map the same semantics as `params`, effectively destroying what came before each time we re-run the Op.

I'm not sure which is better, this patch or the initial "only populate `TableNames` once" tactic.
From what I can tell, the `TableNames` field _is only accessed in tests_, outside of the awkward block of code that was initially panicking... We might be able to just not care about any of this. Maybe the added `delete`s to clear the map is fine.

stdlib/universe/join.go
```diff
@@ -145,6 +145,15 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
...
 func (t *JoinOpSpec) IDer(ider flux.IDer) {
+	// When join is used inside a `tableFind`, the input table operations are
+	// "cloned" granting them new ids.
+	// This will cause a panic to occur in `createMergeJoinTransformation` when
+	// trying to associate table names with their related datasets if we don't
+	// first remove old entries from the map.
+	// Refs: <https://github.com/influxdata/flux/issues/4692>
+	for k := range t.TableNames {
+		delete(t.TableNames, k)
+	}
```
onelson, line 156: My general thinking here is internal to the OpSpec, we rely on `param` field more than the public `TableNames` field so if `param` is changing, we should "fit" `TableNames` to it.

This solves the panic problem by keeping the two aligned, so to speak.

onelson:
Looking at where the "names" of the tables even come into play, it seems as though the only impact they have is on the column renaming code.

Given that each run of the join _will have the correct datasets_ to read from, and the names are captured when constructing the cache for the join, _and are largely cosmetic_ it seems as though either solution presented here could be _good enough_.

Additionally since the names in question can't change (in my examples the names `xs` and `ys` cannot vary without becoming a type error), the list of names _should be static_ for any joins reused via `tableFind`. Ex: the type system will not allow:
```
join(tables: if x then { a: ..., b: ... } else { c: ..., d: ... })
```

At this point, I think I'm comfortable either solution.

stdlib/universe/covariance_test.go
```diff
@@ -88,11 +88,7 @@ func TestCovariance_NewQuery(t *testing.T) {
 					{
 						ID: "join2",
 						Spec: &universe.JoinOpSpec{
-							On: []string{"host"},
-							TableNames: map[flux.OperationID]string{
-								"from0": "x",
-								"from1": "y",
-							},
```
onelson, line 95: A handful of tests "want" the `TableNames` field removed in this diff. It's a bit of a shame the `params` field is not public, but _there are plenty of acceptance tests_ already in place to attest to the name-to-tables handling.

After looking at it, I don't feel like it's important to rush to rework things to get visibility on this at the OpSpec level.

stdlib/universe/join.go
```diff
@@ -144,13 +142,6 @@ func createJoinOpSpec(args flux.Arguments, a *flux.Administration) (flux.Operati
 	return spec, nil
 }
 
-func (t *JoinOpSpec) IDer(ider flux.IDer) {
-	for i, name := range t.params.names {
-		operation := t.params.operations[i]
-		t.TableNames[ider.ID(operation)] = name
-	}
-}
```
onelson, line 152: Implementing this method is no longer needed now that we aren't using the operation ids to figuratively _shoot ourselves in the foot_.

stdlib/universe/join.go
```diff
@@ -215,6 +204,22 @@ func createMergeJoinTransformation(id execute.DatasetID, mode execute.Accumulati
 		return nil, nil, errors.New(codes.Unimplemented, "joins currently must only have two parents")
 	}
 
+	// In <https://github.com/influxdata/flux/issues/4692>, it was found that in
+	// certain cases `s.TableNames` could grow beyond the size of the parents
+	// _incorrectly_.
+	// This would lead to a panic in the for/range to follow.
+	// To illustrate the requirement in a more explicit way, this check ensures
+	// that the names of the tables, and their associated datasets are of equal counts.
+	if len(parents) != len(s.TableNames) {
```
onelson, line 213: In case it's confusing -- the field removed in f57d93bc5bd36037c61860f3d948a358285a5c72 was the `TableNames` field of `JoinOpSpec` whereas this is the `TableNames` field on the `MergeJoinProcedureSpec`. Previously the latter was built from the former, but now it is built instead from `JoinOpSpec.params`.

onelson:
After discussing with @wolffcm it became apparent that keeping `params` and `TableNames` aligned is not important if we could just use _one or the other_. Since the `TableNames` map was the problematic one (in terms of the panic), I pushed f57d93bc5bd36037c61860f3d948a358285a5c72 to remove it.

sunbryely-influxdata (approved):

wolffcm (approved):
Looks great, thanks again for digging in on this one.

PR #4701, opened 2022-04-26
fix(universe): fix SortLimit for empty input group

wolffcm:
This PR adds a bit of logic to our handling of `sort |> limit` to catch the case where we have received an empty table. The code the does merging did not expect to have to deal with zero-row tables, and wound up in an infinite loop.

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/universe/sort.go
+++ b/stdlib/universe/sort.go
@@ -332,7 +332,7 @@ func (s *sortTableMergeHeap) ValueLen() int {
 
 func (s *sortTableMergeHeap) Table(limit int, mem memory.Allocator) (flux.Table, error) {
 	if s.ValueLen() == 0 {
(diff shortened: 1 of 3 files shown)


stdlib/universe/sort.go
```diff
@@ -331,6 +331,14 @@ func (s *sortTableMergeHeap) ValueLen() int {
 }
 
 func (s *sortTableMergeHeap) Table(limit int, mem memory.Allocator) (flux.Table, error) {
+	if s.ValueLen() == 0 {
+		// Degenerate case where ther are no rows to merge sort
```
Marwes, line 335:
```suggestion
		// Degenerate case where there are no rows to merge sort
```

Marwes (approved):

PR #4707, opened 2022-04-28
feat: Allow default types to be specified for default arguments

Marwes:
This is necessary for label polymorphism to work with default labels (usually `"_value"). By specifying the builtin with the default label type inference is able to pick the label up and propagate it.

Based on #4697


libflux/flux-core/src/semantic/sub.rs
```diff
@@ -264,7 +264,7 @@ pub trait Substitutable {
     /// Returns `Self` but with "fresh" type variables
     fn fresh(&self, fresher: &mut Fresher) -> Self
     where
-        Self: Sized + Clone,
+        Self: Sized + Clone + std::fmt::Debug,
```
skartikey, line 267: is this intentional or are some debug code left by mistake?

skartikey (approved):
Nice work!

Issue #4712, opened 2022-04-29
formatter will stack overflow when formatting a long chain of `or`


onelson:
**TL;DR:** long `or` chains can cause the formatter to stack overflow. To mitigate, either we need to reduce allocations or make the flow less recursive.

---

At time of writing, given a flux program like:

```flux
x = false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false or false
```

will cause `fluxcore::formatter::format()` to stack overflow. By my count, today's threshold is around 193 `or` being the breaking point.

## Reducing Allocations

There are a few places where we could reduce allocations, for example in this specific case types like `ast::LogicalOperator` are [cast to strings at each level of recursion](https://github.com/influxdata/flux/blob/a11f1207337eacfa60c5b2e22f7ce7fbd3fe9fdd/libflux/flux-core/src/formatter/mod.rs#L1529).

We could add an impl for `AsRef<str>` which returns the &'static str instead. Better might be to use only the enum variants themselves rather than dropping down to String/&str, but there might be a polymorphic concern -- are two or more enums we feed into this function?

These strings are going to be a small savings, but not really enough to make an impact. A POC to move towards using `&'static str` for this got us a single additional level of recursion only.

## Refactoring to reduce recursion

This is more likely the way we want to go if we really want to mitigate. Ideally, flux (the formatter, and the parser) should be able to happily consume programs that perhaps are rejected by limits downstream (such as in storage). We would prefer to not be the limiting factor.

[...]

PR #4721, opened 2022-05-09
perf: Avoid allocating a String for each formatted operator

Marwes:
Refactors the parse/display of operators to use a common macro which removes duplication
and ensures that we cover all variants. I changed `ToString for LogicalOperator` to a `fmt::Display`
implementation as there is a default implementation that auto implements `ToString` (`ToString for T where T: fmt::Display`).

cc #4712


onelson (approved):
A much cleaner implementation than I had in my POC 😅
The macro really helps!

Not sure how much of a savings this really is, but the allocations we were doing _did seem needless_.

Marwes:
Yeah, this won't do anything to fix the issue itself, it just removes some unnecessary work.

PR #4725, opened 2022-05-09
feat(stdlib): add experimental.preview function

### LINKED ISSUE #4719: Create experimental preview function for limiting the output
Create an experimental preview function. The preview function will limit the number of output series and the number of rows, but will not affect the sources so it will still run the entire query.

It would be the equivalent of something like the following:

```
|> slimit(n: 5)
|> limit(n: 5)
```

DOD:

- Preview function exists that limits the number of tables and number of rows per table
- Preview function has two parameters (table limit, row limit), maximum number of rows returned is (table limit * row limit). Parameters should have small defaults (i.e. 5)

jsternberg:
The preview function will limit the number of rows and the number of
tables that can be returned by a transformation. It will discard any
remaining tables/rows.

Fixes #4719.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,34 @@ builtin histogram : (
         bins: [float],
         ?normalize: bool,
     ) => stream[{T with _value: float, le: float}]
+
+// preview limits the number of rows and tables in the stream.
+// The group keys that are included are not deterministic and depends on the order
+// that the engine sends them.
```
sanderson, line 1258:
```suggestion
// preview limits the number of rows and tables in the stream.
//
// Included groups keys are not deterministic and depend on the order
// that the engine sends them.
```

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,34 @@ builtin histogram : (
...
+//
+// ## Parameters
+// - nrows: Maximum number of rows in each table.
```
sanderson, line 1261:
```suggestion
// - nrows: Maximum number of rows per table to return.
```

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,34 @@ builtin histogram : (
...
+//
+// - ntables: Maximum number of tables returned by the transformation.
```
sanderson, line 1263:
```suggestion
// - ntables: Maximum number of tables to return.
```

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,34 @@ builtin histogram : (
...
+//
+// ## Examples
+//
+// ### Preview data output
+// ```
+// import "experimental"
+// import "sampledata"
+//
+// sampledata.init()
+//   |> experimental.preview()
```
sanderson, line 1275:
```suggestion
// ```
// import "experimental"
// import "sampledata"
//
// sampledata.int()
//     |> experimental.preview()
// ```
```

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,34 @@ builtin histogram : (
...
+//
+// ## Metadata
+// introduced: next
```
sanderson, line 1278: @nathanielc Does case matter here? Does this need to be `NEXT`?

stdlib/experimental/experimental.flux
sanderson: Does this parameter have a default value? The example below suggests that it does.

stdlib/experimental/experimental.flux
sanderson: Does this parameter have a default value? The example below suggests that it does.

sanderson (changes requested):

stdlib/experimental/experimental.flux
jsternberg: It looks like it's supposed to be `NEXT`. Fixed.

wolffcm (approved):
Looks reasonable.

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,33 @@ builtin histogram : (
         bins: [float],
         ?normalize: bool,
     ) => stream[{T with _value: float, le: float}]
+
+// preview limits the number of rows and tables in the stream.
+//
+// Included group keys are not deterministic and depends on the order
+// that the engine sends them.
```
sanderson, line 1259:
```suggestion
// The function is designed to quickly provide a data preview by not
// returning the entire result set.
// Included group keys are not deterministic and depends on the order
// that the engine sends them.
```

stdlib/experimental/experimental.flux
```diff
@@ -1252,3 +1252,33 @@ builtin histogram : (
...
+//
+// ## Examples
+//
+// ### Preview data output
+// ```
+// import "experimental"
+// import "sampledata"
+//
+// sampledata.init()
+//     |> experimental.preview()
```
sanderson, line 1277: - Change `sampledata.init()` to `sampledata.int()`
- Passing parameters in the example will show exactly how the input tables are modified.
- Add input (`<`) and output (`>`) identifiers to generate the input and output examples.

```suggestion
// < sampledata.int()
// >     |> experimental.preview(nrows: 3, ntables: 1)
```

sanderson (approved):

PR #4729, opened 2022-05-10
refactor: Remove the need for the flux_test_gen files

Marwes:
Take two of https://github.com/influxdata/flux/pull/4715 . Retains the flux_test_gen.go files so that `go mod vendor` does not omit the `.flux` files  but they are now just empty dummy files. Then this exposes the test gathering such that packages that vendor us can just give the path to the `stdlib` directory inside their vendored flux dependency.


skartikey:
It looks good to me, but may I ask @wolffcm to look at it as well as the previous commit was [reverted](https://github.com/influxdata/flux/pull/4724) recently.

wolffcm (commented):
This looks reasonable to me, just a couple questions:
1. What is it about having the `_test_gen.go` files that causes `go mod vendor` to retain all the `.flux` files?
2. Did you test this out with your branch in the downstream repo?

wolffcm (approved):

wolffcm:
Thanks for taking this on, it will be a big quality of life improvement.

PR #4730, opened 2022-05-10
feat(stdlib): add scale function to add to allow for dynamic duration changes

### LINKED ISSUE #4742: Add date.scale method to multiply durations by an integer
At the moment, we can't take a base duration and then multiply it. This operation is supported by the Flux duration type and is used by the windowing logic, but it's not exposed either through an operator or the standard library.

For the operator, we generally can't do this because the operation is `duration * int` so the two sides of the operation are not equivalent types. `duration * duration` isn't allowed.

Similar to `date.add` and `date.sub`, which have a similar problem but with `time` instead of `int`, I propose we have a `date.scale(d: duration, n: int)` that performs the multiplication and add it to the standard library in the same place.

jsternberg:
feat(stdlib): add scale function to add to allow for dynamic duration changes

This adds a `scale` function to `date` so that you can functionally
do `t + (d * scale)`. This is possible in flux and is used by windowing
logic, but there's no practical way to do it in pure flux.

Fixes #4742.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


sanderson (approved):

sunbryely-influxdata (approved):

stdlib/date/date.flux
```diff
@@ -680,6 +680,31 @@ builtin _truncate : (t: T, unit: duration, location: {zone: string, offset: dura
...
+//
+// ```no_run
+// import "date"
+//
+// n = 5
+// d = date.scale(d: 1h, n: n)
+// date.add(d: d, to: 2022-05-10T00:00:00Z)
+//
+// // Returns 2022-05-10T00:00:00.000000000Z
+// ```
```
nathanielc, line 701: Can you add an example of how this works with a mixed duration?

nathanielc (approved):
One small change requested but othewise LGTM. No need for me to re-review

stdlib/date/date.flux
jsternberg: Done.

PR #4743, opened 2022-05-12
feat(testing): add support for setting feature flags via options

nathanielc:
With this change Flux acceptance tests can set the feature flags they
need via the debug.features option.

Additionally the flux test command support specifying the feature flags
on the CLI, the set of flags specified on the CLI overwrite the entire
set defined in the test case options.

Draft TODO:

- [x] Identifiers as Literals? Seems fine for now but maybe we should clean this up too? Leaving them as is for now.
- [x] ~Should we fully deprecate the testFlags with this PR?~ No I'll do this in a follow up PR.
- [x] Create InfluxDB/IDPE PRs for the breaking changes to the TestRunner interface. See https://github.com/influxdata/influxdb/pull/23355 and https://github.com/influxdata/idpe/pull/14508

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


sanderson (approved):
Just some typo fixes, then it's good from a docs perspective.

wolffcm:
> Additionally the flux test command support specifying the feature flags
on the CLI, the set of flags specified on the CLI overwrite the entire
set defined in the test case options.

This seems a bit backwards to me? I would expect flag values to be defined in order of increasing precedence:
- The default defined in the `flags.yml` file
- The value defined on the `fluxtest` command line
- The value defined with an `option` statement in the test case itself

It seems like test cases will be written with the intent to test a particular feature, so if a flag value can be overwritten by the CLI, don't we run the risk of losing coverage, or of tests becoming useless or failing?

It might be the case that for some configuration of flags provided on the CLI that we just don't want some tests to run, that seems like a good usage of labels.

Just my $.2, curious to know the reasoning behind this behavior.

lang/compiler.go
```diff
@@ -493,6 +494,12 @@ func (p *AstProgram) Start(ctx context.Context, alloc memory.Allocator) (flux.Qu
 	// execution begins.
 	deps.ExecutionOptions.ConcurrencyLimit = feature.QueryConcurrencyLimit().Int(ctx)
 
+	// Create a MutableFlagger for this program, ensuring that any changes it
+	// makes to feature flags are scoped to its own execution.
+	flagger := pkgfeature.GetFlagger(ctx)
+	mflagger := pkgfeature.NewMutableFlagger(flagger)
+	ctx = pkgfeature.Inject(ctx, mflagger)
+
```
wolffcm, line 502: I'm wondering about what would happen if we accidentally had `import "debug" debug.setFeature(...)` in a stdlib file. Would any script that imported that file have that feature flag set?

wolffcm: Also, I'm curious if this will work in the REPL. Do we create a new context for each line of input?

wolffcm (commented):
Overall I think this looks great. I just had some questions about precedence for setting flags. But looking over the code maybe I misunderstood your comment.

lang/compiler.go
nathanielc: I'll test the REPL but I don't think we need it for each new line. 

Since its a function importing won't call the function and so won't have any effect.

PR #4754, opened 2022-05-13
fix(stdlib/date): date.add and date.sub work correctly with timezones enabled

### LINKED ISSUE #4706: experimental.subduration yields wrong result in combination with now() and timezones
I have the following flux script:
```
import "array"
import "experimental"
import "timezone"

option location = timezone.location(name: "Europe/Berlin")
nowTS = now()
nowM24 = experimental.subDuration(d: 24h, from: now())

array.from(rows: 
    [
        {
            nowTS: nowTS,
            nowM24: nowM24
        }
    ]
)
```

It yields the following result (script executed 2022-04-27T22:00:XXZ):
```
_result
table   nowTS                           nowM24
0       2022-04-27T22:00:42.06168091Z   2022-04-27T00:00:42.06168091Z
```

When not defining `location`, everything works fine. However, with the `location` defined above, the subDuration yields a result 2 hours off and unexpected. There was no time zone shift or similar, so subduration of 24 hours should have resulted in 2022-04-26T22:00:XXZ

jsternberg:
When timezones were enabled, date.add and date.sub did not correctly
work. To do time math, we convert from the local time to an equivalent
clock time in utc. We perform the math and then convert back to the
local time with the same clock time.

We did not correctly convert back to the local clock time when using
`date.add` and `date.sub`. This meant that they would convert to the utc
equivalent, perform the math, and then just never go back. This resulted
in incorrect times.

This was masked by the unit tests which only tested if a specific digit
of the time was correct. These test cases tested minutes which were not
affected by this bug. The tests have been updated to check the full time
instead of just a single digit.

Fixes #4706.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


sanderson (approved):

stdlib/experimental/durations_test.flux
```diff
@@ -5,6 +5,7 @@ import "csv"
 import "testing"
```
skartikey, line 5: As durations have been moved to date from experimental, this test can be removed from experimental.

jsternberg: Removed.

skartikey (approved):
Looks good!

onelson (approved):
This all makes sense to me. I wonder if there are any other gaps where offset is not respected.

PR #4756, opened 2022-05-13
chore: make planner robust to rules that return nil values

wolffcm:
Recently an issue occurred where a planner rule returned a `nil` value for its `newNode` return when no change was made tot he plan. This is pretty reasonable to do, but the planner was not robust to it.

This change makes it so rules can return a `nil` value for the new plan node, and if no change was made to the plan, it works just fine.


nathanielc (approved):

plan/heuristic_planner.go
```diff
@@ -40,6 +42,20 @@ func (p *heuristicPlanner) clearRules() {
 	p.rules = make(map[ProcedureKind][]Rule)
 }
 
+func applyRule(ctx context.Context, rule Rule, node Node) (Node, bool, error) {
```
jsternberg, line 45: This method doesn't ever appear to return the new node?

PR #4757, opened 2022-05-16
feat(stdlib): added timezone support to hourSelection function

skartikey:
issue - [hourSelection()](https://docs.influxdata.com/flux/v0.x/stdlib/universe/hourselection/) function not respecting timezone setting

fixes: influxdata/EAR#3213

### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written


stdlib/universe/hour_selection.go
```diff
@@ -184,3 +208,29 @@ func (t *hourSelectionTransformation) UpdateProcessingTime(id execute.DatasetID,
...
+				return "UTC", values.ConvertDurationNsecs(0), errors.Newf(codes.Invalid, "offset property for location must be of type %s, got %s", semantic.Duration, got)
+			}
+		}
+	}
+	if name.IsNull() {
+		name = values.NewString("UTC")
+	}
+
+	return name.Str(), offset.Duration(), nil
+}
```
wolffcm, line 236: Is this copied and pasted from a function in the `date` package? Maybe that function could just be exported?

stdlib/universe/hour_selection_test.flux
```diff
@@ -65,3 +66,27 @@ testcase hour_selection_overnight_range {
...
+#group,false,false,true,true,true,true,false,false
+#default,_result,,,,,,,
+,result,table,_start,_stop,_measurement,_field,_time,_value
+,,0,2018-12-01T00:00:00Z,2030-01-01T00:00:00Z,Sgf,DlXwgrw,2018-12-18T17:11:25Z,33
+,,0,2018-12-01T00:00:00Z,2030-01-01T00:00:00Z,Sgf,DlXwgrw,2018-12-19T18:11:35Z,63
+",
+            )
+
+        testing.diff(got: got, want: want)
+    }
```
wolffcm, line 92: This looks good. I checked that `2018-12-19T18:11:35Z` is within the 9-10 range in LA time, and it turns out that is 10:11 in LA time. But the semantics of `hourSelection` is that the `stop` hour is inclusive. This was a surprise to me! So this looks good.

wolffcm (changes requested):
It would be good to avoid copying and pasting code if we can. Looks good otherwise.

wolffcm (approved):

Issue #4644, opened 2022-04-05
POC: Dynamic Type


onelson:
The details of what should be delivered are still TBD.

At any rate, we want to try and design a proof of concept around adding a Dynamic type to Flux to serve possible shortcomings like:

- being able to model mixed-type collections.
- modeling truly unknown values as Dynamic to avoid repeated type checks on other values when they are not actually needed.

One idea was to add strict typing to sources where schema information is available, then make all others return Dynamic.

PR #4762, opened 2022-05-16
feat(POC): add dynamic type

onelson:
Refs: #4644

So, the tl;dr might be:

> certain cases for `json.parse` can be served by `Dynamic` but it's not comprehensive. ~~The capability gain is quite limited without support of additional work.~~ More could be done to make handing of nulls more ergonomic, but even without this seems pretty okay!

This POC adds a new MonoType named `Dynamic`, a `Value` type to go along with it, then updates `json.parse` such that any array it produces is typed as `[dynamic]` regardless of whether or not the actual items inside are all the same type or not.

## What Works?

The following use cases technically work, but my have varying utility.

### Case 1

Arrays with varied object layouts, even with varied types for common field names work!

For this first case, note that:
- the first and second objects in the array have differently typed `value` fields.
- the second object has an `extra` field where the first does not.

```
import "array"
import "experimental/json"
import "types"

data = json.parse(data: bytes(v: "[{\"value\":123}, {\"value\": \"foo\", \"extra\": 123}]"))
rows = [
    {
        isFloat: types.isType(v: data[0].value, type: "float"),
        isString: types.isType(v: data[0].value, type: "string"),
        hasExtra: if exists data[0].extra then "yes" else "no",
    },
    {
        isFloat: types.isType(v: data[1].value, type: "float"),
        isString: types.isType(v: data[1].value, type: "string"),
        hasExtra: if exists data[1].extra then "yes" else "no",
    },
]
array.from(rows: rows)
```
<details>
<summary>output</summary>

```
Result: _result
Table: keys: []
isFloat:bool  isString:bool         hasExtra:string  
------------  -------------  ----------------------  
        true          false                      no  
       false           true                     yes  
```
</details>

In order for this scenario to work we must delicately marshal each object into a valid `Record`. In order to achieve this, we cannot map the `undefined` value we'd see as `extra` for that first object. This is to say we have to branch to provide a non-null default or otherwise reconcile the typing so it's consistent.

`array.from` _requires that that we give it `Record`s_, so passing along the raw `Dynamic` objects won't work.

### Case 2

For the 2nd case, I looked at how Dynamic works when an array holds a series of "basic types" including a literal `null`.

```
import "array"
import "experimental/json"

data = json.parse(data: bytes(v: "{\"items\": [1, 2.0, \"three\", false, null]}"))

rows = [
    {
        a: data.items[0],
        b: data.items[1],
        c: data.items[2],
        d: data.items[3],
        // XXX: null can't be used here...
        // e: data.items[4],
    },
]

array.from(rows: rows)
```
<details>
<summary>output</summary>

```
Result: _result
Table: keys: []
                     a:float                       b:float                c:string        d:bool  
----------------------------  ----------------------------  ----------------------  ------------  
                           1                             2                   three         false  

```
</details>

The values unpacked into our output `Record` are `values.Dynamic`, but the type information shown in the REPL output here is coming from `values.Dynamic.Type()` which delegates down to the inner `Value` it wraps. This is to say, the values "show up" as the flux type ascribed by the marshaling process internal to `json.parse`, which is sort of neat.

The `null` becomes problematic in that we can't assign it. This is to say _we must go out of our way to avoid assigning it_ to a field in our output `Record`. We could use `exists` to provide a fallback for the value, but other than that the play is "avoidance."

### Case 3
 
This case was more about verifying the behavior of `Dynamic` and how it interacts with the type system.

Given that `values.Dynamic` delegates to an inner `values.Value` which has its own Nature/Kind, lookups for various operators should "just work."

```
import "array"
import "experimental/json"

payload = json.parse(data: bytes(v: "[123, 123, 124, 124.0, 122]"))

rows = [
  {a: payload[0], b: payload[1], eq: payload[0] == payload[1], lt: payload[0] < payload[1], gt: payload[0] > payload[1] },
  {a: payload[1], b: payload[2], eq: payload[1] == payload[2], lt: payload[1] < payload[2], gt: payload[1] > payload[2] },
  {a: payload[2], b: payload[3], eq: payload[2] == payload[3], lt: payload[2] < payload[3], gt: payload[2] > payload[3] },
  {a: payload[3], b: payload[4], eq: payload[3] == payload[4], lt: payload[3] < payload[4], gt: payload[3] > payload[4] },
]
array.from(rows: rows)
```

<details>
<summary>output</summary>

```
Result: _result
Table: keys: []
                     a:float                       b:float       eq:bool       lt:bool       gt:bool  
----------------------------  ----------------------------  ------------  ------------  ------------  
                         123                           123          true         false         false  
                         123                           124         false          true         false  
                         124                           124          true         false         false  
                         124                           122         false         false          true  
```
</details>

This confirmation sort of drives home the point that, as-written, the code changes required outside of sites where we specifically construct arrays of things should be minimal. The `Dynamic` type "passes" for other basic types as it travels through the runtime.

### Case 4

**EDIT** this case formerly looked like it was a problem but now works thanks to rolling back a change I erroneously made earlier.

You'll note that in all the "working" cases so far, we're pulling values out of parsed json arrays _by index_. This is far from ideal since a more realistic case is where we'd be consuming _unknown numbers of items_.

For this, we'd need to use `experimental/array`.

Consider a case like the following:

```
import "array"
import "contrib/bonitoo-io/hex"
import "experimental/json"
import "types"
import "experimental/array"

// array with a mix of float and string values
payload = json.parse(data: bytes(v: "[123, \"aaa\", \"000\", \"FFF\", 60, 90]"))

// normalize to a hex string...
toHex = (x) => if types.isType(v: x, type: "float") then hex.string(v: int(v: x)) else x

rows = array.map(arr: payload, fn: (x) => ({value: toHex(x: x)}))
array.from(rows: rows)
```

<details>
<summary>output</summary>

```
Result: _result
Table: keys: []
          value:string  
----------------------  
                    7b  
                   aaa  
                   000  
                   FFF  
                    3c  
                    5a  

```
</details>

## What's Missing?

At this point, not too much. The implementation is spartan and in some places incomplete. A good amount of polish and scrutiny will be required before we can merge this.

For example, we might identify more places like `json.parse` which should be updated to type any arrays they produce using `Dynamic`. We should also consider if these arrays should be typed more strictly in cases where we see all the elements have a common type, only downgrading to `Dynamic` on a case by case basis.

Finally, there are some bits of the `Value` interface that need proper consideration, such as `Copy()`.

### Array literals

Probably the simplest thing to talk about as being "missing" feature-wise is a way to build an array using a mix of types. A workaround could be for scripts to employ string interpolation to build a json string, then parse it...

A more ergonomic approach would be to add a `dynamic()` builtin, like the others we provide for casts.

```
arr = [dynamic(v:false), dynamic(v:"hello")]
```

Alternatively, we could try to infer an array as `[dynamic]` when the items are
heterogeneous. Array would need to "degrade" to dynamic automatically.
This would be less explicit however, and could more easily lead to programmer errors.

### null and undefined

Generally speaking this _doesn't do much_ for handling "missing" data. It's still a little painful to work with when `null` or `undefined` is in the mix. Having to _avoid_ these values feels wrong, even while branching to provide fallbacks makes good sense. It might be nice if we could do something to allow missing/null fields to be marshaled more easily.

If we were to add the `dynamic()` builtin, we'd be able to built arrays for the purpose of producing HTTP bodies for example, but if we needed to inject a `null` into that array, we'd still be forced to use clunky workarounds...


wolffcm:
Can you elaborate a bit more on why `null`s are problematic, e.g., for case 2? Why can't you wrap a `null` value created like this:
https://github.com/influxdata/flux/blob/4996f846b917311495e1c83a9ceb4d69d1b16225/values/values.go#L256

I guess you would need to choose a type for it, but maybe it doesn't matter what you pick?

onelson:
> Can you elaborate a bit more on why `null`s are problematic, e.g., for case 2?

Happy to.

@wolffcm with the current state of the branch, if we uncomment the field `e` in that example we get the following:

```
Result: _result
Error: cannot represent the type null as column data
exit status 1
```

It might be possible to handle nulls differently in a way that will work, specifically in `json.parse`, or there might be some other approach. I'm not sure what would be best.

---

I think ideally, we should be able to pass nulls along when mapping these values to record fields without having to rely on conditional expressions to fill in defaults in the case of a null.

Part of my rationale is the fact it is possible to build a table with actual nullable fields, but only by jumping through hoops.

Ex: given a json array of object where some fields are sometimes missing or null, we'd have to
- filter out the nulls then map to records
- re-filter for only the nulls, map to records
- join the two datasets together with union (and possibly opaque).

Something like

```
items = json.parse(....) // array of obj, some fields present always, others not

partials = array.map(arr: array.filter(arr: items, fn: (x) => !exists x.c), fn: (x) => ({ a: x.a, b: x.b }))
fulls = array.map(arr: array.filter(arr: items, fn: (x) => exists x.c), fn: (x) => ({ a: x.a, b: x.b, c: x.c }))

union(tables: [array.from(rows: partials), array.from(rows: fulls)])
```

It would feel much more natural to me to map over the data once, and pass the nulls along directly.

I can see allowing this to happen could complicate our runtime further, so I'm not sure how feasible it is.

nathanielc:
### Re Nulls

> It would feel much more natural to me to map over the data once, and pass the nulls along directly.
I can see allowing this to happen could complicate our runtime further, so I'm not sure how feasible it is.

I agree and I think we should explore this. The error you report that columns cannot be `type null` has to do with @wolffcm's question about creating a new null value and its type. Similar to Go's interfaces, Flux nulls are typed, meaning their value is null and they also have a type i.e. int, float etc. We also have a `type` called `null` which we use in cases were we don't know the type. This `null` type is not a type that we allow for a column hence the error. I see two possible solutions here: 

1. Allow the type null to be a column type, the semantics would be similar to the untyped null value, i.e. we know its always a null value.
2. Somehow pick the correct type for the null value, either by checking the expected static type or by some other means.

The goal here would be to keep the same semantics Flux already has around nulls, namely: _a user doesn't have to deal with or think about nulls by default, only when the user cares about nulls should they have to check for them and add logic around them_. The implication is that if a value is null it should _just work_(i.e. not error) even if that means that it creates more nulls as its being used.


### Re Concrete vs Static Types

The following is a discussion around the future of the dynamic type when we have a runtime that is less dynamic than the current interpreter. 

>Given that values.Dynamic delegates to an inner values.Value which has its own Nature/Kind, lookups for various operators should "just work."

This seems like the correct implementation given the current design of the interpreter, however I think we would want to reconsider this design in the context of a bytecode/IR runtime. The current interpreter is dynamic, because we represent Flux values in the `values.Value` interface we always check the actual value we have been passed even if statically we know its type. This means that generally speaking the interpreter is dynamic already and so having the `dynamic` type delegate to its concrete type makes sense. The alternative way to do this is to have `dynamic` report its `Nature/Type` as `dynamic` and then let the code handle the dynamism directly (i.e. explicitly ask the dynamic type for its underlying value). But because most of the Go code in the runtime already treats values as dynamic this would be redundant and a heavy burden since all uses of a `values.Value` would have to explicitly handle `dynamic` values. So again in the context of the current interpreter the design in this POC is the correct design from my perspective.

Looking forward to a bytecode/IR runtime, we will have typed operations and so the runtime will no longer be so dynamic. In fact we hope to be able to remove most of the dynamic checks that exist currently since we will be able to rely on the static checks for that code. For the cases where we have  a `dynamic` value now we have to add special logic to convert the dynamic value into the typed value otherwise every typed instruction in the IR would have to check if its arguments are really typed as expect, thus making it dynamic again. To me this means we will have the concept of a _concrete_ type,  the type of the actual value at runtime, and the _static_ type, the type we determined a compile time. Then anytime we have a dynamic type the code generation will need to generate code to unwrap the dynamic value into its concrete value or error if that is not possible. This then gives us the concept of gradual typing where type checks for dynamic values still occur but are delay until runtime instead of at compile time. 

Its still not clear to me how much of the dynamic checks we will be able to remove as Go will force them on us in some regard. But I'd like to validate these ideas about using dynamic types for explicit runtime type checks, which implies that without dynamic there is no need for runtime type checks.

compiler/compiler.go
```diff
@@ -75,6 +75,12 @@ func Compile(scope Scope, f *semantic.FunctionExpression, in semantic.MonoType)
 // If the input type is not a type variable, it will check to ensure
 // that the type in the input matches or it will return an error.
 func substituteTypes(subst *semantic.Substitution, inferredType, actualType semantic.MonoType) error {
+	// FIXME: hack to allow `[dynamic]` to flow through `array.map`.
```
onelson, line 78: n.b. I think the original problem with `array.map` was resolved by rolling back some bad edits.

iirc _some sort of check like this is still necessary_, however.

Issue #4607, opened 2022-03-28
Return a more helpful error message to the user when an http response exceeds 100MB


scbrickley:
The flux http client [limits response sizes](https://github.com/influxdata/flux/blob/b1b083c140a2c1c12d95ee7665ae08bc5c862f4d/dependencies/http/http.go#L18) to 100 MB. If a user runs a query that results in large http response, the resulting failure can manifest in a number of different error messages for the user, not all of which are helpful in effectively communicating the actual problem.

We should return a clearer error message in the case of a large http response... one that makes it obvious that the user needs to somehow reduce the amount of data that they're querying.

wolffcm:
One possible manifestation of this is truncated CSV, which produces a confusing error.

PR #4764, opened 2022-05-17
fix: return a more helpful error message to the user when an http response exceeds maxResponseBody(100MB)

skartikey:
issue:
The flux HTTP client [limits response sizes](https://github.com/influxdata/flux/blob/master/dependencies/http/http.go#L18) to 100 MB. If a user runs a query that results in a large HTTP response, the resulting failure can manifest in a number of different error messages for the user, not all of which are helpful in effectively communicating the actual problem.
Another possible manifestation of this is truncated CSV, which produces a confusing error.

fixes: https://github.com/influxdata/flux/issues/4607

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/dependencies/http/http.go
+++ b/dependencies/http/http.go
@@ -53,6 +53,10 @@ func (l roundTripLimiter) RoundTrip(r *http.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


skartikey:
@jsternberg please review.

jsternberg (approved):

PR #4765, opened 2022-05-17
feat: Add a planner rule that expands logical join nodes

### LINKED ISSUE #4627: Add a planner rule that expands logical join nodes
For the sake of performance, it is important that the new `join` function receives sorted tables as its input.

To ensure this is the case, we should add a planner rule that finds logical `join` nodes, and expands them into two physical nodes: one `sort` followed by a `sortMergeJoin`.

A separate but related issue has been filed to add another planner rule that removes unnecessary `sort` nodes: https://github.com/influxdata/flux/issues/4628

NOTE: Decide whether to change the existing join rule (re predicate function) or add a new rule

DOD:
- The planner rule described above exists, and is used by the planner

Marwes:
Just implements the pattern matching and tests that the rule is used. There isn't an actual
transformation that makes use of this.

Closes #4627


Marwes:
@scbrickley Is this all that was needed for the issue? There isn't any way to make use of the planner rule so I were a bit confused about the scope of this.

stdlib/join/sort_merge_join.go
```diff
@@ -0,0 +1,101 @@
...
+		for i, successorNode := range n.Successors() {
+			if successorNode.ID() == n.ID() {
+				n.Successors()[i] = sortNode
+			}
+		}
+		sortNode.AddPredecessors(parentNode)
+		sortNode.AddSuccessors(n)
+	}
+
+	n.ReplaceSpec(newSortMergeJoin(spec))
```
scbrickley, line 98: I don't think there's any need to replace the node here. There's another rule in place that will take care of inserting the proper procedure spec (see #4665).

scbrickley (changes requested):
This isn't quite what I had in mind, but I realize now that I failed to outline some important details in the issue.

Happy to discuss these points further.

stdlib/join/sort_merge_join.go
Marwes: Not sure how I would stop the rewrite from just repeatedly being applied without doing a replacement (and thereby ensuring that the rewrite rule no longer matches).

stdlib/join/sortmergejoin_test.go
```diff
@@ -0,0 +1,98 @@
...
+				as: (l, r) => ({l with c: r._value}),
+				method: "inner",
+			)`,
+			wantPlan: &plantest.PlanSpec{
+				Nodes: []plan.Node{
+					plan.CreateLogicalNode("from0", &influxdb.FromProcedureSpec{}),
+					plan.CreateLogicalNode("filter1", &universe.FilterProcedureSpec{}),
+					plan.CreateLogicalNode("from2", &influxdb.FromProcedureSpec{}),
+					plan.CreateLogicalNode("filter3", &universe.FilterProcedureSpec{}),
+					plan.CreatePhysicalNode("join.join4", &join.SortMergeJoinProcedureSpec{}),
```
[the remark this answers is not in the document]
Marwes: Yeah, I copy pasted this a bit to quickly. Think I have fixed it now, though I am not entirely sure how to check that predecessor 0 and 1 matches the left and right parameter so the sorts may be the wrong way around :thinking: .

stdlib/join/sort_merge_join.go
```diff
@@ -0,0 +1,107 @@
...
+func (p *SortMergeJoinProcedureSpec) Kind() plan.ProcedureKind {
+	return plan.ProcedureKind(SortMergeJoinKind)
+}
+
+func (p *SortMergeJoinProcedureSpec) Copy() plan.ProcedureSpec {
+	return &SortMergeJoinProcedureSpec{
+		As:     p.As,
+		Left:   p.Left,
+		Right:  p.Right,
+		Method: p.Method,
```
scbrickley, line 45: Could we copy the `On` parameter here as well?

scbrickley (approved):
I think this looks good. I had one comment, but it's not worth blocking this PR over it.

I share your concerns about knowing which parent node is which, but because there's no actual transformation running yet, I don't think that's a big deal. I'm going to rebase my current branch for #4629 (which contains the actual transformation code) on this once it merges. The tests I write for that PR should be enough to confirm this works as intended.

PR #4773, opened 2022-05-19
feat: Allow features to enable builtin statements

Marwes:
This lets one write `@feature featurename` in a comment on a builtin statement to optionally include that statement in the resuling semantic graph. Since the standard library is compiled ahead of time this doesn't work to add feature flagged builtins for the go API (and therefore not for anything deployed to prod), it is enough to test two different versions of the standard library in the query log analyzer from #4776


skartikey:
@Marwes linter is throwing some error in CI, pls take a look.

skartikey (approved):

Issue #4740, opened 2022-05-11
Analyze the impact of of using polymorphic labels in the standard library


Marwes:
Polymorphic labels were added in #4388 with a couple of example function signatures that utilize the feature. To be able to change the stable builtins to use polymorphic labels we need to evaluate how large the fallout would be.

DOD:

- Provide alternate type signatures which use polymorphic labels for all builtins where it is useful (see https://github.com/influxdata/flux/issues/4410 for an incomplete list)
- When the label polymorphism feature is enabled do the typechecking with these signatures that use labels
- Evaluate the impact by compiling a representative set of real-world queries to determine the regressions

PR #4774, opened 2022-05-19
feat: Provide the comments for each Symbol from PackageExports

Marwes:
Extends the `PackageExports` type to store the comments for each exported symbol. Main use case is that LSP as I believe it should be able to use this to lookup the documentation for any exported standard library function and return it as part of a completion item (cc @rockstar , may need some work on the completion side but you may already be working on improving that(?)). Longer term I would like to also provide `symbol -> attribute` lookup through the same API.

Separated from #4773 / #4740 as I realized I didn't really need this part yet.


Marwes:
https://github.com/influxdata/flux/pull/4537 might be needed also to make symbols more accurate in the LSP

OfTheDelmer (approved):

PR #4783, opened 2022-05-23
fix(semantic): attach type when constructing logical expressions

jsternberg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


nathanielc (approved):

onelson (approved):
lgtm :+1: 

iirc this relates back to how we handle these expressions while vectorizing. Perhaps this just means we can substitute the typing to the base case of `bool` instead of having to additionally support the expr type. Am I thinking about this right?

PR #4786, opened 2022-05-24
fix: Correct several issues found during #4740

Marwes:
Various fixes and refactorings I have done so for for #4740 . Reviewing commit by commit is recommended.


libflux/flux-core/src/semantic/bootstrap.rs
```diff
@@ -43,10 +43,16 @@ pub type SemanticPackageMap = SemanticMap<String, Package>;
 /// Infers the Flux standard library given the path to the source code.
 /// The prelude and the imports are returned.
 #[allow(clippy::type_complexity)]
-pub fn infer_stdlib_dir(path: &Path) -> Result<(PackageExports, Packages, SemanticPackageMap)> {
+pub fn infer_stdlib_dir(
```
skartikey, line 46: the second argument needs to be supplied here as well. else it will return an error. 

https://github.com/influxdata/flux/blob/5a828bdd85c9cba5428c50e7f62b42382e68927f/libflux/flux/benches/basic.rs#L30

skartikey, line 46: the second argument needs to be supplied here as well. else it will return an error. 

https://github.com/influxdata/flux/blob/5a828bdd85c9cba5428c50e7f62b42382e68927f/libflux/flux/benches/basic.rs#L30

skartikey (approved):

PR #4796, opened 2022-05-26
feat(pagerduty): add a _status tag to pagerduty records

gavincabbage:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


brettbuddin:
I'd be interested to see if we could eventually make use of the newer `experimental/http` package in these notify destination packages. We'd love to capture as much information as we can about our interaction with the upstream services as we can for users to leverage in `_monitoring`.

sanderson (approved):

nathanielc (approved):

PR #4802, opened 2022-05-27
fix(integral): fix math for integral with a single value

nathanielc:
Prior to this change an integral with a single value did not scale the
result by the units of the integral. Leaving the units as nanoseconds
which meant that often a very large value was computed. This change
fixes that adds tests and some comments.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/universe/integral.go
```diff
@@ -261,18 +261,25 @@ func (t *integralTransformation) Process(id execute.DatasetID, tbl flux.Table) e
 		}
 		switch in.points {
 		case 0:
+			// We had no data therefore the area under the curve is zero.
```
wolffcm, line 264: A null value might make more sense here, applying the logic for a single point below. If there are no points, there's no basis for computing an area for a given range.

Not asking for this to be changed, but it doesn't seem quite right.

wolffcm (approved):
Looks good.

sunbryely-influxdata (approved):

stdlib/universe/integral.go
nathanielc: Yeah a null value would make more sense, although it would be a breaking change. Maybe if this comes up again we can add a parameter to change the behavior.

PR #4806, opened 2022-05-27
docs(rollups): adds initial draft of rollups proposal

nathanielc:
See proposal for details.

### DIFF
--- a/docs/rollups_proposal.md
+++ b/docs/rollups_proposal.md
@@ -1,0 +1,175 @@
+# Rollups
+
+Users want to be able to downsample their data and then query across each downsampled bucket transparently.
+Something similar to Graphite's or RRD metric databases.
+There are two key aspects to how those systems work:
+
+* Data is downsampled automatically via a predefined configuration.
+* Data is queried across the entire range of data without the user needing to know how downsampling was configured.
+
+This creates a nice user experience where queries over the data range return quickly because as the query covers a larger time period less dense data is processed.
+Instead of querying data proportional to the time range the data processed is relatively constant.
+This should then result in constant time queries regardless of the time range queried.
+
+
+## How should this use case be addressed in the InfluxDB ecosystem?
+
+This proposal is that we create a _rollups_ feature.
+There has been lots of talk over the years about _intelligent rollups_, however we can deliver on a simpler form of rollups that will more than satisfy the user's needs.
+A _rollup_ is a predefined downsampling configuration, where the user provides the platform with the information it needs to rollup data and to query data back out of the rollup.
+This simple definition eliminates some of the complexities in the _intelligent rollups_ proposals from the past that attempted to in one form or another determine how data should be downsampled automatically.
+This  _rollup_ is equivalent to how Graphite and RRD provided these same features.
+
+
+### How does one use a rollup?
+
+We should create a Flux package named `rollups` that can query data across different downsampled retentions transparently for the user.
+At a high level the Flux API would be similar to this:
+
+```flux
+import "influxdata/influxdb/rollups"
+
+
+// Query data across the entire rollup
+rollups.from(name: "mydata", start: -1y, stop: now())
+    |> filter()
+```
+
+This query returns the rolledup data across the entire time range.
+It reads similar to using a normal `from` call but we are being explicit that we want data from a rollup.
+
+### How do ones know which rollups exist?
+
+In order to use a _rollup_ as user must configure it via the API.
+At a minimum we would need information about which buckets cover which time ranges.
+At a maximum we would need information about the source raw bucket, which aggregate methods to apply and over which retentions and resolutions.
+
+We should explore what users are looking for if we want to build to full system of defining the aggregates for the rollup etc or if we want them to create those manually and then just inform Flux where data lives for which time periods.
+
+
+## How would we build this?
+
+Here is the high level design:
+
+* Store the data for each retention period into a bucket.
+* Define tasks that downsample data from one bucket to the next using a naming convention.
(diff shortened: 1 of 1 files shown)


sanderson:
@nathanielc This is something that has been on my mind lately, so I’m excited to see this proposal. The thing I’ve been mostly thinking about is the downsampling configuration and how users actually want to downsample. I’ve seen the following requests:

- Downsample by field/tag
- Downsample by field type

I think we could address each of these use cases by structuring the configuration as an array of records where each record has a predicate (a filter predicate) and an fn (aggregate/selector) property. So it would look something like this:

```javascript
import "experimental/array"
import "types"
import "rollups"

downsampleConf = [
    {predicate: (r) => types.isType(v: r._value, type: "string"), fn: last)},
    {predicate: (r) => types.isType(v: r._value, type: "bool"), fn: mode)},
    {predicate: (r) => not types.isType(v: r._value, type: "string") and not types.isType(v: r._value, type: "bool"), fn: mean)},
]

// rollups.do = (downsampleConf, every, tables=<-) => {
//     _downsampled = array.map(
//         arr: downsampleConf,
//         fn: (x) => tables
//             |> filter(fn: x.predicate)
//             |> aggregateWindow(every: every, fn: x.fn())
//     )
//
//     return union(tables: _downsampled)
// }

from(bucket: "example-bucket-1")
    |> range(start: -1y)
    |> rollups.do(downsampleConf: downsampleConf, every: 1mo)
    |> to(bucket: "example-bucket-2")
```

This doesn't address the destination bucket logic, but something like this would give users A LOT of flexibility in how they downsample their data (potentially too much flexibility).

Just some thoughts I've had related to this.

deanasheehan:
1. In this proposal you refer to it as “rollup" but I think we should consider it more generally. It could, for example, be some math between measurements that is being computed by tasks or even written directly by external systems rather than a “rollup”. With the correct semantic information supplied to Influx we could - in time - leverage that to make some queries execute significantly faster and whilst we might have a very narrow implementation to begin with, it would be a shame to paint ourselves into a corner from the outset. Whilst were only talking about using a different word here, if we use “rollup” too often then I think it causes people to think too narrowly and that would be a shame even if our initial implementation is narrow for practical reasons.

2. I would like to understand whether the “rollups.from”, or similar, is considered to be the best way or a short term solution? I believe that if we are to favour simplicity for the query author (and this might originate through our own Cloud UI, Grafana, Chronograf or other API with or without UI level support) that the user should just think "normal query" and whether there is precomputed data that could be used is entirely separate and hidden/seamless (just faster). For example, imagine a future where within Influx we notice repeated execution of, the “Hello World!” of Time Series, “bucket |> filter |> aggregateWindow(fn:mean,every:1h)” for a common bucket and measurement. Influx itself might choose to precompute the aggregation (keeping it live through a synthetic Flux Task if you will) automatically and then have it be used without any effort or awareness on behalf of the query author, just things run a lot faster and less compute resources likely used to deliver. This means that the query presented to Influx should be of its natural form over what could be considered the logical schema. Explicit use of “rollups.from” or anything else that changes the actual query presented from that canonical form would not work. I appreciate that we are up against it from a time perspective for our initial customer and that we may need to take shortcuts but I want to understand if there is a philosophical disconnect here re the long term. 50:50 whether I am missing the obvious flaw or being super insightful. I think this goes to the question you ask in the proposal of “Why not make buckets multi-retention?” and I have more on this below in point 4. Additionally, if we want to retain consistency between OSS, ENT and Cloud and then have operational differences then it creates a barrier to tell people “you need to change your queries to use rollup.from” if you want to use Cloud over your current OSS . Now, there are various ways of mitigating that (like rollup.from degenerating to not use rollup if not available) but we’re now letting the tail wag the dog.

4. “Rollup” configuration. I think we separate out the two pieces. I already have a precompute Flux Task generator which could use (was used in the recent POC) if mass pre- computation with a consistent naming convention is required - it could also be modified to register pre-commutated field semantics with an API if that is what is proposed. The solution needs a form of dictionary (implied through convention or explicit through API call) that allows it to understand that field F in measurement M of Bucket B has a pre-empted valued of 10s:mean stored in B’, M’ and F’ that was started Static Time S and trailing Now() by Seconds. 

5. To your question “Why not make buckets multi-retention? I think we are  conflating two things here - there logical view of a query re the underlying data structure/storage. If we consider the potential that the query presented to Influx (via API) represents the logical query then separate buckets can be used - no TSM or IOx re-write which would be excessive - pre-computation can be set up manually or with generator assistance quite separately - and the query may or may not benefit from precomputed data being available. This goes to my point 2 above.
    1. In essence the query engine (maybe even higher) can ask the question “does this org have any registered pre-computation?). That would be, are there any entries in the dictionary for this org. If not, normal query path.
    2. If so, look at the query (AST) and see if it matches one of the current re-write patterns. If not then continue as normal
    3. If it does then re-write the query (almost generate the union Flux on the fly - at least as a way of thinking about it) and send that Flux in instead.

I want to present our initial customer with an option where the use of rollups is explicit (per something like this proposal with use of something akin to “rollups.from”) and implicit where it is all behind the scenes (but not down at the storage layer). We might get feedback like "we like the outcome (100x performance improvement) but in its explicit form, even If sooner in timeframe X, will not be helpful and we are open to waiting until time Y to get the alternative implicit approach. I can’t present that option without a) some agreement that the alternative implicit option even exists and what X and Y timeframes might be.

nathanielc:
After internal discussion we have discovered the need is not for hierarchical rollups rather reusing downsampled/pre-computed downsampled data. Going to close this proposal for now.

PR #4809, opened 2022-05-31
feat: Add a pretty.rs based MonoType formatter

Marwes:
The `fmt::Display` based formatter only included absolutely necessary whitespace which
creates compact output which is rather hard to read. By copying over the formatter used for `ast::MonoType` we get more readable output in error messages and the LSP.

(I checked flux-lsp and the only brekage is in a test due to the addition of whitespace).

BREAKING CHANGE

Some `std::fmt::Display` implementations have been removed in the `types` module. Convert to `MonoType` and use its `Display` implementation instead or write the formatting manually.


skartikey (approved):
Could you please open an issue/PR to fix flux-lsp tests?

Issue #4307, opened 2021-11-29
Holtwinters() function not performing as expected


Anaisdg:
I used it for double exponential smoothing. That data barely has a negative slope. I would expect a much flatter line. 
<img width="1432" alt="Screen Shot 2021-11-29 at 11 09 56 AM" src="https://user-images.githubusercontent.com/30506042/143912198-4c54ed8d-b35b-41c5-bf51-a375b6e2adac.png">

```
from(bucket: "system")
  |> range(start: 2021-11-17T21:07:40.000Z, stop: 	2021-11-17T21:18:40.000Z)
  |> filter(fn: (r) => r["_measurement"] == "cpu")
  |> filter(fn: (r) => r["_field"] == "usage_system")
  |> filter(fn: (r) => r["cpu"] == "cpu-total")
  |> limit(n: 20)
  |> yield(name: "raw")
  |> holtWinters(n: 10, interval: 20s, seasonality: 0)
```
Here is the raw data with the holtwinters forecast output using seasonality=0 or double exponential smoothing:
[2021-11-29_10_11_influxdb_data.csv](https://github.com/influxdata/flux/files/7619454/2021-11-29_10_11_influxdb_data.csv)
t.
The forecasted values are:
0.15788753424689228
-0.7799915747417391
-2.0884170385340113
-3.9138518779427898
-6.460646815371196
-10.013919881448576
-14.971478972807773
-21.888361504251364
-31.53897794557487
-45.0038170160562

statsmodels confirms that there should be a forecast with much less slope:
```
import pandas as pd
import numpy as np 
import matplotlib.pyplot as plt
from statsmodels.tsa.api import ExponentialSmoothing, SimpleExpSmoothing, Holt

data = [2.4750000000057724,3.5041666666582914,3.5041666666582914,1.9586597766247285,2.1250000000035167,1.5291029540443428,3.204300179169086,3.88317153452289,2.1294328457749447,2.212131311446779,2.6789434213814647,1.9212335903300701,2.1872265966730393, 1.7298874531100965,2.5666666666631417, 2.9506147113993393, 1.8538576903864667, 1.1666666666648478, 1.3835062716186677, 0.8625000000012369]

index = pd.date_range(start="2021-11-17T21:07:40", end="2021-11-17T21:14:00", freq="20s")
data = pd.Series(data, index)
fit = Holt(data, exponential=True, initialization_method="estimated").fit()
fcast = fit.forecast(10).rename("Exponential trend")
fcast
[...]

PR #4816, opened 2022-06-01
feat(holtwinters): replaced our optimization with gonum optimize

skartikey:
Replaced our existing Nelder-Mead optimization with [gonum optimize](https://pkg.go.dev/gonum.org/v1/gonum@v0.11.0/optimize)

fixes: https://github.com/influxdata/flux/issues/4307

### Benchmark -
_go test -bench=BenchmarkHoltWinters_
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz

### Existing [Nelder-mead](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/neldermead.go) method implementation
```
BenchmarkHoltWintersWithoutFit-12                          34815             35703 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithFit-12                             37070             32289 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithoutFitSeasonality-12               38385             31545 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithFitSeasonality-12                  35692             31936 ns/op            3608 B/op         60 allocs/op
```

### Gonum [Nelder-mead](https://github.com/gonum/gonum/blob/master/optimize/neldermead.go) implementation
```
BenchmarkHoltWintersWithoutFit-12                          37761             30296 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithFit-12                             37507             30845 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithoutFitSeasonality-12               40724             30385 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithFitSeasonality-12                  36145             32799 ns/op            3672 B/op         61 allocs/op
```

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


skartikey:
benchstat after this commit [perf: allocate fix size mem and copy values](https://github.com/influxdata/flux/pull/4816/commits/b596f90353540d94d13768ed616637cfdaa7c347)

_benchstat before.bench memFix.bench_
```
name                                 old time/op    new time/op    delta
HoltWintersWithoutFit-12               20.4ms ± 0%    17.6ms ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                  19.6ms ± 0%    18.3ms ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12    27.9µs ± 0%    31.7µs ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFitSeasonality-12       28.7µs ± 0%    31.5µs ± 0%   ~     (p=1.000 n=1+1)

name                                 old alloc/op   new alloc/op   delta
HoltWintersWithoutFit-12               4.92MB ± 0%    2.46MB ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                  4.92MB ± 0%    2.46MB ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12    3.67kB ± 0%    3.67kB ± 0%   ~     (all equal)
HoltWintersWithFitSeasonality-12       3.67kB ± 0%    3.67kB ± 0%   ~     (all equal)

name                                 old allocs/op  new allocs/op  delta
HoltWintersWithoutFit-12                57.8k ± 0%     28.9k ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                   57.8k ± 0%     28.9k ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12      61.0 ± 0%      61.0 ± 0%   ~     (all equal)
HoltWintersWithFitSeasonality-12         61.0 ± 0%      61.0 ± 0%   ~     (all equal)
```

skartikey:
Ran profiler on benchmark - 

**CPU -** 

image

pprof-cpu


**Mem -**
image
pprof-mem

stdlib/universe/holt_winters_test.go
```diff
@@ -660,88 +662,88 @@ func TestHoltWinters_Process(t *testing.T) {
...
+		// 			{execute.Time(1440781800000000000), 3.243455723309489},
+		// 			{execute.Time(1440804540000000000), 3.243454638835965},
+		// 			{execute.Time(1440827280000000000), 3.2434540315472833},
+		// 			{execute.Time(1440850020000000000), 3.2434536914755303},
+		// 			{execute.Time(1440872760000000000), 3.2434535010411083},
+		// 			{execute.Time(1440895500000000000), 3.2434533944011235},
+		// 			{execute.Time(1440918240000000000), 3.2434533346845957},
+		// 		},
+		// 	}},
+		// },
```
nathanielc, line 746: Can you reenable this test?

stdlib/universe/holt_winters_test.go
```diff
@@ -856,16 +858,16 @@ func TestHoltWinters_Process(t *testing.T) {
...
+					{execute.Time(1440713580000000000), 6.517746679116747},
+					{execute.Time(1440736320000000000), 8.44990936365862},
+					{execute.Time(1440759060000000000), 11.450103919073781},
+					{execute.Time(1440781800000000000), 16.108703181906527},
+					{execute.Time(1440804540000000000), 23.342418185768715},
+					{execute.Time(1440827280000000000), 34.57468700800914},
+					{execute.Time(1440850020000000000), 52.01577652114399},
+					{execute.Time(1440872760000000000), 79.09771500282905},
+					{execute.Time(1440895500000000000), 121.14964091593322},
+					{execute.Time(1440918240000000000), 186.4464618626079},
```
nathanielc, line 870: Is it possible to add the SSE to these test cases so we can see that it improved as part of the diff?

nathanielc (changes requested):
@skartikey Looks great! If I am reading this correctly the new implementation is now slightly faster.

Looks like some clean up needs to happen re tests but otherwise LGTM

skartikey:
benchstat after this commit [perf: more mem allocation fix](https://github.com/influxdata/flux/pull/4816/commits/c4224b4b20b0e5b898546aced3969e4241859000)

_benchstat memFix.bench moreMemFix.bench_
```
name                                 old time/op    new time/op    delta
HoltWintersWithoutFit-12               17.6ms ± 0%     0.0ms ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                  18.3ms ± 0%     0.0ms ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12    31.7µs ± 0%    28.8µs ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFitSeasonality-12       31.5µs ± 0%    30.9µs ± 0%   ~     (p=1.000 n=1+1)

name                                 old alloc/op   new alloc/op   delta
HoltWintersWithoutFit-12               2.46MB ± 0%    0.00MB ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                  2.46MB ± 0%    0.00MB ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12    3.67kB ± 0%    3.67kB ± 0%   ~     (all equal)
HoltWintersWithFitSeasonality-12       3.67kB ± 0%    3.67kB ± 0%   ~     (all equal)

name                                 old allocs/op  new allocs/op  delta
HoltWintersWithoutFit-12                28.9k ± 0%      0.1k ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithFit-12                   28.9k ± 0%      0.1k ± 0%   ~     (p=1.000 n=1+1)
HoltWintersWithoutFitSeasonality-12      61.0 ± 0%      61.0 ± 0%   ~     (all equal)
HoltWintersWithFitSeasonality-12         61.0 ± 0%      61.0 ± 0%   ~     (all equal)
```

skartikey:
@nathanielc Here is the `minSSE` value I am getting running the [go test](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters_test.go#L166) with our existing and `gonum` implementation. Looking at the data below, the new one is returning better results for the `minSSE`. 

**Before -** 
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  20.208541132811906
minSSE:  38.22969757997137
minSSE:  55.02308769115148
minSSE:  30.561795293721094

**After -** 
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  3.963712566228843
minSSE:  55.00975192379251
minSSE:  30.52018686151099

 I am using a print statement to get these values.

stdlib/universe/holt_winters_test.go
skartikey: The `gonum` implementation is returning an error for this test -  `optimize: initial function value is infinite`
 I am looking into it.

nathanielc:
@skartikey The SSE's look good. Is there not an easy way to include them in the Go tests?

For the most recent benchmark results it looks like some of the tests run in 0s. That seems suspicious any ideas why those benchmarks are reporting 0s?

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -153,33 +153,24 @@ func (r *HoltWinters) Do(vs *array.Float) *array.Float {
...
+						minSSE = result.F
+						bestParams = result.X
 					}
 				}
 			}
 		}
 	}
 
 	// Final forecast
+	bestParamsF := mutable.NewFloat64Array(r.alloc)
```
nathanielc, line 166: Make sense to allocate the arrow array at the end :+1:

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -122,7 +122,7 @@ func (r *HoltWinters) Do(vs *array.Float) *array.Float {
 
 	// Determine best fit for the various parameters
 	minSSE := math.Inf(1)
-	var bestParams *mutable.Float64Array
+	bestParams := make([]float64, size)
```
nathanielc, line 125: This allocation is never used. Could you leave this line unmodified?

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -153,33 +153,24 @@ func (r *HoltWinters) Do(vs *array.Float) *array.Float {
...
+	for i := 0; i < len(bestParams); i++ {
+		bestParamsF.Set(i, bestParams[i])
+	}
 	fcast := func() *mutable.Float64Array {
-		fcast := r.forecast(bestParams, false)
-		// Now that bestParams have been used to generate the final forecast, they can be released.
-		defer bestParams.Release()
+		fcast := r.forecast(bestParamsF, false)
 		return fcast
 	}()
```
nathanielc, line 175: Do we still need this anonymous function? Seems like it was there to ensure the `defer` ran eagerly. Do we still need it?

skartikey: make sense, removed

skartikey:
> @skartikey The SSE's look good. Is there not an easy way to include them in the Go tests?

to pass the SSE in the go test, I have to make SSE part of the public API result. Not sure if that would be a good idea.

> For the most recent benchmark results it looks like some of the tests run in 0s. That seems suspicious any ideas why 
those benchmarks are reporting 0s?

Looks like `benchstat` tool is doing something here. I ran benchmark again and it looks alright to me. 

**Before -** 

```
BenchmarkHoltWintersWithoutFit-12                          29800             39355 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithFit-12                             28484             36553 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithoutFitSeasonality-12               31322             40150 ns/op            3608 B/op         60 allocs/op
BenchmarkHoltWintersWithFitSeasonality-12                  28293             38462 ns/op            3608 B/op         60 allocs/op
```

**After -**

```
BenchmarkHoltWintersWithoutFit-12                          37884             55364 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithFit-12                             33810             33034 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithoutFitSeasonality-12               35882             34245 ns/op            3672 B/op         61 allocs/op
BenchmarkHoltWintersWithFitSeasonality-12                  30438             36250 ns/op            3672 B/op         61 allocs/op
```

skartikey:
@nathanielc 
Debugged the [test case with nulls and missing data](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters_test.go#L680) today

**Input to holt winter algorithm from the test case -** 
```
seasonal 4
[]float64 [4.948,2.192,0,2.93,5.121,0,0,2.877,5.449,0.896,0,0,0,0.404,4.357,2.618,6.102,0,4.816]
```

What's happening is, in the setup phase of the algo, as [vs.Value(2)](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/holt_winters.go#L114) is invalid according to the [rule](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters.go#L254) (Rows that have a null value are considered invalid, but used by the algorithm), initParams[8] is getting set to 0, 
Due to that, `stm` is getting set to 0 [here](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/holt_winters.go#L222), which is setting[ lT to +Inf](https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/holt_winters.go#L164) and that makes bT to `Nan (Inf - Inf)`
This is introducing `Nan` in the algorithm which is messing up the rest of the processing. 
we need to discuss a way to handle the invalid input.

nathanielc:
@skartikey Ok this is starting to make sense now. That flow of calculations clearly explains why we are seeing NaN in the data. 

Look at how `stm` is used here https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/holt_winters.go#L163-L169, its only used for multiplication or division. This means that a better default for the initParam value when its null is to use `1`. This way dividing or multiplying by `1` will not introduce any Inf or NaNs. Does changing the default here https://github.com/influxdata/flux/blob/master/stdlib/universe/holt_winters/holt_winters.go#L114 to `1` fix the issue?

skartikey:
@nathanielc Yeah, It definitely makes sense to use 1 rather than 0 for the default `stm` as it's used for multiplication or division in the algorithm. 
Setting 1 for the missing/null value for `initParam` has fixed the NaN issue. The test result looks better now. I have updated the PR with my changes.

stdlib/universe/holt_winters/holt_winters.go
```diff
@@ -262,7 +278,7 @@ func (r *HoltWinters) sse(params *mutable.Float64Array) float64 {
 			// Compute error
 			if math.IsNaN(fcast.Value(i)) {
 				// Penalize fcast NaNs
-				return math.Inf(1)
+				return math.MaxFloat64
 			}
```
nathanielc, line 282: Should we just remove this check for IsNaN? And just let the NaNs continue to exist since they must have come from the input data if they do? This way we are not adding extra logic that we don't intend to actually use. If it triggers an error thats fine as the user can then find and fix the source of their own NaNs. This check as its stands will hide that fact from them and cause holt winters to optimize for the NaNs because of the overflow issue we found.

nathanielc (changes requested):
LGTM with a few small changes.

stdlib/universe/holt_winters/holt_winters.go
skartikey: That makes sense. Did a quick test with NaN in the input after removing the above `IsNaN` check. 
The code panics with a message - "`caught panic: optimize: initial function value is NaN`"
Shall I capture `NaN` case in the code and return a user error rather than letting it panic?

nathanielc (approved):

noramullen1 (commented):
LGTM

PR #4819, opened 2022-06-01
fix(stdlib): vectorized map properly handles shadowed columns

### LINKED ISSUE #4812: Vectorized map function should use results for the type instead of type inference
Because of existing bugs in the Go runtime (the interpreter and the compiler), Go doesn't always pass the correct types to transformations so, when it attempts to use the results of type inference, they sometimes fail with strange error messages.

We corrected the row-based version of map in the new map transformation to use the results, but the vectorized evaluator wasn't modified. It should be modified to use the results instead of type inference.

In reality, it would be better to use type inference. But, that's not really possible unless we refactor the interpreter/compiler so it passes the types correctly as type information is commonly lost. We should just fix map at the current moment so it works properly.

jsternberg:
The vectorized version of map did not properly handle shadowed columns
the way the row based one worked. This copies over some of that code so
that it follows the correct logic. It also copies over the same logic
that is used to determine the column type when type inference does not
properly tell us the type.

This also fixes a bug in building objects that caused a bug in the
compiler. When a `with` is evaluated, the attributes from the original
record are copied over first and then overwritten values are assigned.
This overwriting did not properly release the values that were being
overwritten resulting in a memory leak.

Fixes #4812.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/vector_fn.go
+++ b/execute/vector_fn.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/compiler"
 	"github.com/influxdata/flux/execute/table"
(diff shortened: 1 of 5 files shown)


Marwes (changes requested):
I fixed the test I mentioned on slack so that it reproduces the error so if you rebase on #4817 and add this test we have a test to cover that. You seem to have fixed another issue as well though so if you can add a test for that also this looks good :+1: 

```flux
testcase drop_fn_duplicate_column_issue_4812 {
    got = array.from(rows: [{ x: 1, y: "a" }])
        |> map(fn: (r) => ({ r with x: r.x }))
        |> drop(columns: ["y"])

    want = array.from(rows: [{ x: 1 }])

    testing.diff(got, want)
}
```

sanderson (approved):

Marwes (approved):

Issue #4813, opened 2022-05-31
Address some inconsistencies with table find functions


wolffcm:
This issue provides near-term mitigation for the first two examples in https://github.com/influxdata/flux/issues/4805.

The necessary changes would need to happen in the codepath that builds a `Spec` from a table object:
https://github.com/influxdata/flux/blob/958736c095991761f9bfdff1111c43f04641ea4e/internal/spec/build.go#L134

This is the path taken for the table functions as well as `chain`.

The other codepath that builds the `Spec` via `FromEvaluation` should remain unchanged and unaffected.

The steps for this mitigation would be:
- When building the `Spec`, skip over any `yield` nodes. They will end up being evaluated in the `FromEvaluation` path
- If any other side-effecting operations are present, produce an error. These operations would end up incorrectly getting executed twice.

DOD:
- The steps described above are implemented
- Unit tests for the non-`yield` side-effect case (it errors so cannot be an acceptance test)
- acceptance test for the `yield` case

onelson:
Wanted to note that the acceptance test mentioned in the DOD is not currently possible unless I can cook up a case where the yield is empty (otherwise the test will fail). I had the same problem in #4820. When #4300 is implemented we'll have some flexibility to write tests that yield without causing the test to fail.

onelson:
I've got sort of a weak implementation, just a sketch really, that seems to be doing the right thing for my base case:

```
import "array"

// Panic case
x = array.from(rows:[{a: 1, b: 1}])
  |> yield(name: "early")

t = x
    |> map(fn: (r) => ({r with c: 1 }))
    |> tableFind(fn: (key) => true) |> getRecord(idx: 0)

total = t.a + t.b + t.c

x |> map(fn: (r) => ({r with total: total}))
```
Output:
```
Result: early
Table: keys: []
                     a:int                       b:int  
--------------------------  --------------------------  
                         1                           1  
Result: _result
Table: keys: []
                     a:int                       b:int                   total:int  
--------------------------  --------------------------  --------------------------  
                         1                           1                           3
```

Some prints I have in the `FromEvaluation` body show whether or not we're "skipping yields", like for the case of `tableFind`. The number of results is based on the number of `TableObject` sideffects found along the way (I'm still not sure that's a correct indicator).

There's a branch in there to give an error when skip yields is true and the result count is > 1. Untested.

It also prints the spec as dot (I assume that's dot).

```
se op="map"


resultCount=1
skipYields=true
[...]

Issue #4780, opened 2022-05-23
Panic with multiple yields


nathanielc:
This query panics with multiple yields:

```
import "contrib/anaisdg/statsmodels"

from(bucket: "longterm")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "shift")
  |> filter(fn: (r) => r._field == "total" and not exists r.condition)
  |> yield(name: "context")
  |> toFloat()
  |> statsmodels.linearRegression()
  |> map(fn: (r) => ({ r with _value: r.y_hat }))
  |> yield(name: "trend")
```

```
panic during program start" log_id=0aUo7qcl000 service=storage-reads error="runtime error: invalid memory address or nil pointer dereference" stacktrace="goroutine 2139 [running]:
runtime/debug.Stack()
	/go/src/runtime/debug/stack.go:24 +0x88
github.com/influxdata/influxdb/v2/query/control.(*Controller).executeQuery.func1(0x4007226900, 0x400a97d440)
	/root/project/query/control/controller.go:465 +0x140
panic({0xffffa6be38e0, 0xffffa8169a70})
	/go/src/runtime/panic.go:1038 +0x21c
github.com/influxdata/flux/interpreter.(*Interpreter).doExpression(0x400959f180, {0xffffa6fbb4e0, 0x4017768630}, {0xffffa6fcc0d8, 0x4000fd3dd0}, {0xffffa6ff3148, 0x4017768a50})
	/go/pkg/mod/github.com/influxdata/flux@v0.161.0/interpreter/interpreter.go:374 +0x1550
github.com/influxdata/flux/interpreter.(*Interpreter).doExpression(0x400959f180, {0xffffa6fbb4e0, 0x4017768630}, {0xffffa6fcbce8, 0x4000fd3d40}, {0xffffa6ff3148, 0x4017768a50})
	/go/pkg/mod/github.com/influxdata/flux@v0.161.0/interpreter/interpreter.go:428 +0x1114
github.com/influxdata/flux/interpreter.(*Interpreter).doVariableAssignment(0x400959f180, {0xffffa6fbb4e0, 0x4017768630}, 0x40010dacc0, {0xffffa6ff3148, 0x4017768a50})
	/go/pkg/mod/github.com/influxdata/flux@v0.161.0/interpreter/interpreter.go:319 +0x5c
[...]

Issue #4300, opened 2021-11-29
Acceptance Test Harness should pass/fail based on errors


onelson:
Currently the test harness will either pass or fail a test depending on the results yielded by the flux program. Specifically, programs that yields an empty result will pass, and non-empty results will fail.

For example, this is a minimal _passing test_:

```
package sandbox

import "array"

testcase sandbox {
    // empty result
    array.from(rows: [{}]) |> filter(fn: (r) => false)
}
```

whereas this is a minimal _failing test_:

```
package sandbox

import "array"

testcase sandbox {
    // non-empty result
    array.from(rows: [{}])
}
```

Using empty and non-empty as the pass/fail metric means some tests are impossible to write.
Ideally, tests should pass or fail based on errors generated by assertions.

This would allow (implicitly or otherwise) intermediate results to yield from the program without automatically causing the test to fail. A concrete example where this would be helpful is in writing tests that leverage `experimental.chain()` where often the `first` argument will be automatically yielded before the rest of the program completes.

NOTE: We think a good way to implement this is to add a new signal to the runtime that determines if a test passes or fails. This signal is independent of errors within the runtime meaning a test can be marked as failed even without returning an error. Additionally we would make `testing.diff` mark a test as failed when its output is non-empty. This moves the non-empty logic to be withing the testing.diff function and expands the testing harness to use a more explicit means of communicating test failure.

DOD:

- A test can pass/fail independent of its output
- A test can fail even if it doesn't error
- Existing tests need not be modified

onelson:
I know there is a previous way to write acceptance tests which might be where the empty/non-empty signalling requirement comes from. If we move forward with this proposal, we may need to rephrase the older tests to conform with the assertion/error-based signalling scheme.

onelson:
- historically looking at errors/panics, no way to measure intent (#4301 might help)

---

Just this is 3-ish territory, probably...

pass/fail based on errors
print any/all results

---

how to expect an error?

use an `option` to specify a testcase _should fail_ instead of a function

PR #4820, opened 2022-06-02
fix: ensure buffers are retained when copying an already buffered table

onelson:
Refs #4780

When the tables sent into `tableFind` include a yield, the result generated by the yield can cause the tables to be released before `tableFind` can use them.

`tableFind` will make its own copy, but prior to this change the copy operation would hand back the input directly when it was already buffered. In effect, this means we produce references to the same underlying buffers without us incrementing the ref count.

~~With this diff, for the case where we try to copy an already buffered table, we call its own `Copy()` method which will `Retain()` each of the internal buffers.~~

After a few twists and turns, the fixes for this issue came from:

- making `tableBuffer`'s `Do` and `Done` idempotent such that calling `Done` after `Do` would not release the underlying arrow arrays repeatedly.
- ensuring `table.Copy()` always calls `Retain()` on the underlying arrow arrays (_previously this was not the case when the table is already buffered_).

A `tableBuffer` could be spent by calling either `Do()` or `Done()`, then later we might call `Do()` again. Prior to this patch, `Do()` would silently do nothing. Instead, it will now return an error "table already read." This is in line with other implementations of the Table APIs.

Similarly, `Copy()` will now panic if called on a table that is already "spent." Prior to the explicit panic, we'd see panics surfaced by invalid (`nil` pointer) access to underlying arrow arrays somewhere deeper into the execution of the program. _Seems as though this is a case where a `tableBuffer` lived too long via programmer error so a more immediate/explicit panic makes sense._

---

N.b this is not a total fix for #4780. We still have a separate problem related to how `tableFind` will use a random result from the input table stream when there is >1 result to be had. #4813 should address the remaining problems surfaced by the initial report.

### Done checklist
- [x] Test cases written

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ export PKG_CONFIG:=$(shell pwd)/pkg-config.sh
 
 export GOOS=$(shell go env GOOS)
 export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
+export GO_TEST=env GO111MODULE=on go test $(GO_ARGS) -tags assert
 export GO_RUN=env GO111MODULE=on go run $(GO_ARGS)
 export GO_TEST_FLAGS=
 # Do not add GO111MODULE=on to the call to go generate so it doesn't pollute the environment.
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ export PKG_CONFIG:=$(shell pwd)/pkg-config.sh
 
 export GOOS=$(shell go env GOOS)
 export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS) -tags assert
+export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
 export GO_RUN=env GO111MODULE=on go run $(GO_ARGS)
 export GO_TEST_FLAGS=
 # Do not add GO111MODULE=on to the call to go generate so it doesn't pollute the environment.
--- a/execute/table/copy.go
+++ b/execute/table/copy.go
@@ -16,7 +16,7 @@ import "github.com/influxdata/flux"
 // this happens and prevent it.
 func Copy(t flux.Table) (flux.BufferedTable, error) {
 	if tbl, ok := t.(flux.BufferedTable); ok {
-		return tbl, nil
+		return tbl.Copy(), nil
 	}
 
 	tbl := tableBuffer{
--- a/execute/table/copy.go
+++ b/execute/table/copy.go
@@ -1,11 +1,11 @@
 package table
 
 import (
-	"log"
 	"sync/atomic"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/debug"
 	"github.com/influxdata/flux/internal/errors"
 )
 
@@ -103,12 +103,10 @@ func (tb *tableBuffer) BufferN() int {
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
 
-	// TODO: make this panic instead of just logging when running tests.
-	//  Arrow has a `debug.Assert` thing which will noop unless the assert build
-	//  tag is set. We could do something similar.
-	if atomic.LoadInt32(&tb.used) == 1 {
-		log.Println("tried to copy an already used tableBuffer")
-	}
+	debug.Assert(
+		atomic.LoadInt32(&tb.used) == 0,
+		"tried to copy an already used tableBuffer",
+	)
 
 	for i := 0; i < len(tb.buffers); i++ {
 		tb.buffers[i].Retain()
--- a/execute/table/copy.go
+++ b/execute/table/copy.go
@@ -1,6 +1,7 @@
 package table
 
 import (
+	"log"
 	"sync/atomic"
 
 	"github.com/influxdata/flux"
@@ -101,8 +102,12 @@ func (tb *tableBuffer) BufferN() int {
 }
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
+
+	// TODO: make this panic instead of just logging when running tests.
+	//  Arrow has a `debug.Assert` thing which will noop unless the assert build
+	//  tag is set. We could do something similar.
 	if atomic.LoadInt32(&tb.used) == 1 {
-		panic("tried to copy an already used tableBuffer")
+		log.Println("tried to copy an already used tableBuffer")
 	}
 
 	for i := 0; i < len(tb.buffers); i++ {
--- a/execute/table/copy.go
+++ b/execute/table/copy.go
@@ -101,7 +101,7 @@ func (tb *tableBuffer) BufferN() int {
 }
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
-	if tb.used == 1 {
+	if atomic.LoadInt32(&tb.used) == 1 {
 		panic("tried to copy an already used tableBuffer")
 	}
 
--- a/execute/table/copy.go
+++ b/execute/table/copy.go
@@ -1,6 +1,12 @@
 package table
 
-import "github.com/influxdata/flux"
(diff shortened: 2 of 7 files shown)


onelson:
I had hoped to add a flux acceptance test matching the repro used while investigating #4780 but this is not possible to do since any non-empty result in a test case will mean the test will fail. Once #4300 has been completed, this should be possible.

execute/table/copy.go
```diff
@@ -16,7 +16,7 @@ import "github.com/influxdata/flux"
 // this happens and prevent it.
 func Copy(t flux.Table) (flux.BufferedTable, error) {
 	if tbl, ok := t.(flux.BufferedTable); ok {
-		return tbl, nil
+		return tbl.Copy(), nil
```
onelson, line 19: This might be a question for @jsternberg.

To mitigate the issue, we can either call `Copy()` as in the diff currently, or we can build a fresh `BufferedTable` by removing this block entirely. The semantic difference that I can see is the former appears to carry the value for `BufferedTable.i` forward, whereas the later would reset it to `0`.

I was not clear on the impact either way, but since `BufferedTable.Copy` is already a thing, I hoped that it accounted for this already.

jsternberg: These functions are weird and the naming to them is a bit of a mistake. `execute.CopyTable()` can actually be thought of more as "ensure table is buffered" while `buf.Copy()` is "copy this table so it can be used somewhere else".

The difference is that `execute.CopyTable()` _consumes_ its input. So the test in there so see "is this already a buffered table?" is mostly meant as an optimization to the function's true purpose. `buf.Copy()` will produce a copy of the table so it can be used in multiple places.

The naming isn't the best but this change would cause `execute.CopyTable()` to not work properly in the locations where it is presently used.

onelson: Based on your description here, perhaps the thing to do is make the `Copy` call closer (hopefully inside) `tableFind`. There's a slight bit of indirection via a function named `objects.NewTable`.

Regardless, I can look for another point in which to make this happen if this is contradictory to the design.

Thoughts?

onelson: I can add a type assertion to `tableFind` such that when it gets an already buffered table it calls `Copy()` on it before we hand it off to `NewTable()`. This seems sufficient to fix the bug for the identified case, but it'll put me back to square one as far as testing goes :grimacing:

jsternberg:
I had a look with this with @onelson and we discovered something I wasn't really aware of. It appears like the table find code is invoking everything correctly as-is (other than the multiple result thing). The problem seems to be that the `tableBuffer` struct, which is used for buffering data like is done here, has a bug with its `Done()` method.

`Done()` is meant to be idempotent and you can call multiple times. The table find code uses it that way. Unfortunately, there appears to be a bug in `tableBuffer` where if you invoke `Do()` (such as to read the table buffer) and then you call `Done()`, it decrements the reference count twice instead of only once as it is supposed to do.

All of the other implementations of `flux.Table` follow this pattern but `tableBuffer` isn't. I believe that table find is using the Go API in the proper way and `tableBuffer` is just not fulfilling the API the way it is supposed to be.

execute/table/copy.go
onelson: I just paired with Jonathan briefly and with his guidance I'm looking at an alternate fix.

Sounds like the implementations for `Do()` and `Done()` are not currently idempotent and they need to be, and it seems like if we truly do need to drain the table in question we need to add a missing `Copy()` (but there is doubt that this is actually needed today).

onelson:
Moving this back into draft while I rework the patch.

onelson:
Still have more digging to do. After adding atomic guards around `Do`/`Done` for `tableBuffer` I see a different panic here:

https://github.com/influxdata/flux/blob/master/stdlib/universe/table_fns.go#L261

Looks like the length check panics if the underlying arrow array has already been released, so either I screwed up the guard or there's something else contributing. Eyeballing the field I am using for the guard, `used` is 0 and 1 when I'd expect it to be, so whatever is happening is unclear.

I'll have to pick this back up next week.

execute/table/copy.go
onelson: > To mitigate the issue, we can either call Copy() as in the diff currently, or we can build a fresh BufferedTable by removing this block entirely.

After some back and forth, and the discovery of those other idempotent-related issues, the conclusion was to remove the block rather than artificially add a `Copy()` call.

What I didn't realize when I posed the question in the first place is there's a call here which will copy the table already:

https://github.com/influxdata/flux/blob/878a9f1a60837db0e28c473ea0e12a1f730650ed/stdlib/universe/table_fns.go#L251

Adding an extra copy would have been redundant, possibly leading to a leak (though this wasn't clear from the test suite).

execute/table/copy.go
```diff
@@ -89,13 +101,16 @@ func (tb *tableBuffer) BufferN() int {
 }
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
-	for i := tb.i; i < len(tb.buffers); i++ {
+	if tb.used == 1 {
```
jsternberg, line 104: Use the atomic load here.

onelson: Whoops! Makes sense.

execute/table/copy.go
```diff
@@ -89,13 +101,16 @@ func (tb *tableBuffer) BufferN() int {
 }
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
-	for i := tb.i; i < len(tb.buffers); i++ {
+	if atomic.LoadInt32(&tb.used) == 1 {
+		panic("tried to copy an already used tableBuffer")
```
jsternberg, line 105: I'm going to approve, but is there any chance you can have this panic only exist when the `assert` tag is used?

The `assert` tag is part of arrow for multiple releases and we use it when running tests.

jsternberg (approved):

execute/table/copy.go
onelson: I'll look into that. You'd rather see some later downstream error or panic in these cases, assuming they reach production somehow?

onelson: I switched this to be a logging call instead for now. Arrow's `debug.Assert()` is neat, but it lives in their `internal/` module space so I don't think we can access it.

We might want to add something like that to flux directly.

execute/table/copy.go
```diff
@@ -89,13 +102,20 @@ func (tb *tableBuffer) BufferN() int {
 }
 
 func (tb *tableBuffer) Copy() flux.BufferedTable {
-	for i := tb.i; i < len(tb.buffers); i++ {
+
+	// TODO: make this panic instead of just logging when running tests.
+	//  Arrow has a `debug.Assert` thing which will noop unless the assert build
+	//  tag is set. We could do something similar.
+	if atomic.LoadInt32(&tb.used) == 1 {
+		log.Println("tried to copy an already used tableBuffer")
```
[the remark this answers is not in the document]
onelson: I saw a random usage of `log.Println()` over in `compiler/runtime.go` :shrug:

I just didn't want to pile on adding something like that with this PR. Would rather bikeshed it separately if needed.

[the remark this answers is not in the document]
onelson: I was hesitant to add this immediately since verification seemed tricky. Added with b8c5a74786e86e93a54da5e2703323023db94eb1

I weakly tested by flipping the assertion condition and adding/removing the build tag in the Makefile to make sure panics happened or didn't happen when we wanted.

Makefile
```diff
@@ -12,7 +12,7 @@ export PKG_CONFIG:=$(shell pwd)/pkg-config.sh
 
 export GOOS=$(shell go env GOOS)
 export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
-export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
```
jsternberg, line 15: I believe this is defined in `.circleci/config.yml`. We inject `assert` through an argument.

jsternberg (approved):
One minor revert to the Makefile but otherwise good.

Makefile
onelson: Seems like if we go that route, we won't get the nice feedback when testing locally. I'm fine with reverting it, but it seems a bit odd :woozy_face:

Issue #4768, opened 2022-05-17
Convert All End to End Test to Acceptance Tests


nathanielc:
Flux has two integration type test systems.

* End to End tests: defined using the older `test` keyword.
* Acceptance tests: defined using the newer `testcase` keyword

We have been migrating many tests to the acceptance test system as we touch the code. Now it is time to aggressively convert all remaining end to end tests to acceptance tests so we can stop maintaining the end to end system.

Here is a list of packages that still have end to end tests defined.

- [x] array @jsternberg https://github.com/influxdata/flux/pull/4766
- [x] contrib/anaisdg/anomalydetection @skartikey https://github.com/influxdata/flux/pull/4785
- [x] contrib/anaisdg/statsmodels @skartikey https://github.com/influxdata/flux/pull/4785
- [x] The rest @Marwes https://github.com/influxdata/flux/pull/4825

Once the packages are all updated we need to finish the process with some cleanup:

- [x] Remove  `testing.loadStorage` and `testing.loadMem` functions #4858 #4857
- [x] Remove testing end to end Go code  https://github.com/influxdata/idpe/pull/14772
- [x] Remove code generation for the end to end Go code framework https://github.com/influxdata/flux/pull/4855

Once a package no longer has any end to end tests mark it off here.

### Notes on converting tests

Many of the end to end tests are skipped. We should still try to port these tests to see if they pass. If they do great, if not then lets simply remove the old test as its not running anyways.

End to end tests use CSV to encode the inputs and outputs, the new system can leverage array.from and we should in most cases make this change. The exceptions are if you need to encode null values in the data that is not possible with array.from.

End to end tests use the `testing.loadStorage` and `testing.loadMem` functions, the acceptances tests should only use `testing.load` not the older functions. 

Previously a test must live in its own file, the new acceptance tests do not have this constraint. We should prefer to reduce the number of test files where appropriate

DOD:

- Every package has been converted
- Only `testing.load` is used, All calls to `testing.loadStorage` and `testing.loadMem` are removed
- Where reasonable combine test files into a single file for readability.

Marwes:
Seems like there are around 380 tests that aren't converted, going to take a stab at doing an automatic conversion pass by rewriting the AST, doesn't seem to hard for 95% of the cases.

nathanielc:
Should we also remove the `test` keyword from the parser and remove the corresponding `TestStatement` from the AST?

Marwes:
[...]

PR #4822, opened 2022-06-02
chore(stdlib): convert events to acc testcase

OfTheDelmer:
Part of: [#4768](https://github.com/influxdata/flux/issues/4768)

Converts to acceptance `testcase`.


OfTheDelmer:
Included in https://github.com/influxdata/flux/pull/4825

PR #4824, opened 2022-06-02
chore(stdlib): convert aggregate to acc testing

OfTheDelmer:
Part of: [#4768](https://github.com/influxdata/flux/issues/4768)

Converts to acceptance `testcase`.

PR #4823, opened 2022-06-02
chore(stdlib/dict): convert dict to acc testcase

OfTheDelmer:
Part of: [#4768](https://github.com/influxdata/flux/issues/4768)

Converts to acceptance `testcase`.

PR #4825, opened 2022-06-03
refactor: Convert all test statements to testcase

Marwes:
By doing AST rewriting before formatting I could do a simple flux-fmt on the test files to convert all tests at once. As I don't expect us to need to do this refactor again I removed the the AST rewriting code after I ran it on the repo (so you can find it in the commit history just not when reviewing the final changes).

I copied over the skip list from `./stdlib/flux_test.go` and and made the skip unconditional in `cmd/flux test` since I don't know how to retain the comments easily if the skips were put in the Makefile like the other skips. I also had to skip one new tests after they were converted however it was already skipped upstream (just not in this repo :S ).

I didn't convert any csv input or output to use native flux array as I didn't want to get hung up on trying to parse the csv, but a similar rewrite pass could probably be done to convert `csv.from` -> `array.from`

Closes #4821 #4822 #4823 #4824 (cc @OfTheDelmer )

cc #4768 (Doesn't close just yet as we want to remove the old testing functions also)


Marwes:
I ended up using not using any testing.load here since I didn't see any other conversions using it. It looks like a no-op in its default implementation but I guess it is overridden in some test scenarios?

rockstar (commented):
Just came to say "WHOA" on that diffstat. Yayayayayayayay!

skartikey (approved):
Nice work!

skartikey:
@Marwes one of the DOD is _where reasonable combine test files into a single file for readability_, I guess this would need some manual intervention(not stopping merge of this PR, can be done separately).

PR #4835, opened 2022-06-06
fix: require an earlier minimum lsp types

rockstar:
With the switch back down to `lspower` in `flux-lsp`, `flux` is now
requiring too new of a `lsp-types` for `flux-lsp` to use. `flux-lsp`
needs `lsp-types 0.92`, and flux is using `0.93`. This patch downgrades
this requirement.

### DIFF
--- a/libflux/flux-core/Cargo.toml
+++ b/libflux/flux-core/Cargo.toml
@@ -41,7 +41,7 @@ flatbuffers = "2.1.1"
 fnv = "1.0.7"
 libflate = "1.2.0"
 log = "0.4.16"
-lsp-types = { version = "0.93", optional = true }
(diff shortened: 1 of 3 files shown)


libflux/flux-core/Cargo.toml
```diff
@@ -41,7 +41,7 @@ flatbuffers = "2.1.1"
 fnv = "1.0.7"
 libflate = "1.2.0"
 log = "0.4.16"
-lsp-types = { version = "0.93", optional = true }
+lsp-types = { version = "0.92", optional = true }
```
Marwes, line 44: We only use `Position` in `flux-core` which is very stable across `lsp-types` version so we could specify a version range here to maybe avoid some churn.

Issue #4749, opened 2022-05-12
Add a function to create a datetime from a `Timeable`


onelson:
Today, when `v.startTimeRange` is supplied to flux, it can be either a `duration` or `datetime`.

This can cause trouble for script authors who need to produce a datetime string from this value without knowing specifically which type the script will receive.

# Use case

A use case for this is when you have a script that needs a datetime regardless of how the `v.startTimeRange` is specified.

As a concrete example, if you were calling `sql.from()` and needed to string interpolate this value as ISO 8601 in your query to limit the returned rows, your script could only be written to support _either_ relative (duration-based) values, or explicit datetime values.

The script would need to be tweaked if your change the time range drop down in the UI, for example.

# Proposal

Providing a function that can receive any `Timeable` and produce a datetime could solve for this situation.

```
package date

builtin time : (t: A) => time where A: Timeable
```

Example:

```
import "date"

// normalize to datetimes regardless of incoming types
start = date.time(t: v.startTimeRange)
end = date.time(t: v.endTimeRange)

sql.from(
  ...,
  // cast to ISO 8601 for use with SQL...
  query: "SELECT * from events WHERE timestamp >= '${start}' AND timestamp < '${end}'"
) |> // snip...
```

This also makes some sense for use with JSON APIs.

Note: there's a `getTimeableTime` function in `stdlib/date/date.go` which can be used to this aim.

PR #4842, opened 2022-06-06
feat(date): add builtin function time to the date package to convert any timeable into datetime (#4749)

Rahul-Vasan:
This feature addition now addresses the need for a function as part of the date package to be able to convert any Timeable into a time.

fixes: https://github.com/influxdata/flux/issues/4749

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


sanderson (approved):

stdlib/date/date_test.go
```diff
@@ -432,6 +432,138 @@ func TestTruncate_Time(t *testing.T) {
 		})
 	}
 }
+
+func Test_Time_with_Time(t *testing.T) {
```
skartikey, line 436: nit: following function naming convention, you can rename func to `TestTime_Time, TestTime_Duration`

skartikey (commented):
@nathanielc wondering if we want to support timezone with `time` function. Can be done in separate PR as well. Let me know your thoughts.

PR #4848, opened 2022-06-08
chore: Run test cases in parallel

Marwes:
These take a fairly long time, especially when the input data is loaded externally so having some parallelization is nice.


nathanielc:
LGTM do we know if InfluxDB and IDPE will work well with parallel tests? I.e. with there be bucket conflicts are writing tests data etc? Maybe adding a flag to control the amount of parallelism would help us get around any issues there?

nathanielc (approved):
Left comment above, otherwise LGTM

Marwes:
> LGTM do we know if InfluxDB and IDPE will work well with parallel tests? I.e. with there be bucket conflicts are writing tests data etc? Maybe adding a flag to control the amount of parallelism would help us get around any issues there?

They seem to use a bucket per test, but to be safe I added a `--parallel` flag.

PR #4851, opened 2022-06-08
feat(execute): add method to parallelize aggregate transformations

jsternberg:
Aggregate transformations can now be given multiple parents and will
work in parallel. To define an aggregate transformation as
parallelizable, a `Merge()` method must be implemented on the
transformation.

This can only happen if a planner rule modifies the plan so that a node
which normally has one parent will have multiple parents instead. This
allows aggregate transformations that implement this interface to act as
merge nodes.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/aggregate.go
+++ b/execute/aggregate.go
@@ -2,6 +2,7 @@ package execute
 
 import (
 	"context"
+	"sync"
 
 	"github.com/apache/arrow/go/v7/arrow/memory"
 	"github.com/influxdata/flux"
@@ -92,6 +93,12 @@ func (t *aggregateTransformation) processChunk(chunk table.Chunk) error {
 }
 
 func (t *aggregateTransformation) computeFor(key flux.GroupKey, state interface{}) error {
+	// Special code to allow aggregate parallel to work without copying this code.
+	// Unwrap the state from the aggregate parallel state automatically.
+	if s, ok := state.(*aggregateParallelState); ok {
+		state = s.state
+	}
+
 	if err := t.t.Compute(key, state, t.d, t.d.mem); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


execute/aggregate.go
```diff
@@ -46,13 +47,33 @@ type AggregateTransformation interface {
...
+
 type aggregateTransformation struct {
 	t AggregateTransformation
 	d *TransportDataset
+
+	merge    func(into, from interface{}, mem memory.Allocator) (interface{}, error)
+	parents  map[DatasetID]*RandomAccessGroupLookup
+	finished int
+	err      error
+	mu       sync.RWMutex
```
Marwes, line 72: Would it be possible to make a separate `parallelAggregateTransformation`? The mixing of parallel and sequential concerns in the methods here makes it more difficult to read the code than necessary (IMO)

execute/aggregate.go
```diff
@@ -61,75 +82,261 @@ func NewAggregateTransformation(id DatasetID, t AggregateTransformation, mem mem
...
+				if err != nil || !done {
+					return err
+				}
+
+				// We are done. Retrieve the state.
+				var state interface{}
+				if merged, ok := t.d.Delete(key); ok {
+					state = merged.(*aggregateParallelState).state
+				}
+				return t.computeFor(key, state)
```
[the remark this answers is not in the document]
jsternberg: The original version of this didn't have these do the same thing. I've refactored this so `mergeState` can call `computeFor` so the `done` parameter isn't needed anymore.

execute/aggregate.go
jsternberg: Done. I think it's a bit cleaner now.

PR #4853, opened 2022-06-08
fix(plan): visit successors before continuing DFS on node

OfTheDelmer:
Patches the traversal from sinks to sources and implements a fix
for panics in half diamond DAGs by tracking visitation using node
instead of newNode, but also, prohibits visiting a node if
succesors were not yet visited.

Indexing visitation by node instead of the newNode fixes a panic
for half diamonds DAGs caused by revisitations of updated nodes.
However, it creates a problem of revisiting a newNode with multiple
succesors, because only the first succcesor sees the initial node.
Thus traversal is constrained to nodes with visited successors so
traversal can safely rule out one successor misidentifying a
predecessor as unvisited.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/plan/heuristic_planner_test.go
+++ b/plan/heuristic_planner_test.go
@@ -170,7 +170,7 @@ func TestHeuristicPlanner_Plan(t *testing.T) {
 					{5, 7},
 				},
 			},
-			validateFn: checkVisitedNodes([]plan.NodeID{"7", "6", "4", "1", "0", "3", "2", "5"}),
+			validateFn: checkVisitedNodes([]plan.NodeID{"7", "6", "5", "4", "1", "0", "3", "2"}),
 		},
 		{
 			name: "diamond with rewrite",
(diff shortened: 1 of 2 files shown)


plan/heuristic_planner_test.go
```diff
@@ -223,13 +223,55 @@ func TestHeuristicPlanner_Plan(t *testing.T) {
...
-					"7", "6", "new", "1", "0", "3", "2", "5", // second pass
+					"7", "6", "5", "4", "1", "0", "3", "2", // first pass
+					"7", "6", "5", "new", "1", "0", "3", "2", // second pass
 				}
 				diff := cmp.Diff(wantSeenNodes, seenNodes)
 				require.True(t, diff == "", "found difference between -want/+got nodes:\n%v", diff)
 			},
 		},
+		{
+			name: "half diamond physical",
```
OfTheDelmer, line 234: Going to add a half diamond rewrite

wolffcm (approved):
Looks good, I appreciate the new tests.

PR #4860, opened 2022-06-09
feat(cmd/flux): check memory allocations to ensure no memory leaks in testcase

jsternberg:
This changes the native flux test runner so it uses the checked
allocator to verify that no memory leaks occurred. This existed in the
old end to end tests, but was never ported over to testcase.

This only applies to the flux cli runner. It won't affect downstream
test runner implementations.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/cmd/flux/test.go
+++ b/cmd/flux/test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jsternberg:
It appears this fails because the memory retain/releases we did for the compiler were never ported over to the interpreter. I'm going to put this into draft mode and maybe make an issue for the interpreter.

PR #4876, opened 2022-06-14
fix: Don't stackoverflow on deeply nested expressions 

Marwes:
Added stacker just in case another environment has a smaller stack size than my local environment we also grow
the stack automatically. We still have the depth limit to prevent the stack from growing unbounded.

The depth limit was chosen such that the added test did not stack overflow prior to adding stacker. This limit should be a decent approximation that any previous script works and is under the limit or if it is above the limit it would have stack overflowed.

Closes https://github.com/influxdata/flux/issues/4712


Marwes:
Added back the second, original stack overflow check since the one in parser did not trigger the second test.

libflux/flux-core/src/parser/mod.rs
```diff
@@ -888,10 +904,31 @@ impl<'input> Parser<'input> {
...
 
+    fn depth_guard<T>(&mut self, f: impl FnOnce(&mut Self) -> T) -> Option<T> {
+        self.depth += 1;
+
+        let x = if self.depth > MAX_DEPTH {
+            self.errs.push(format!("Program is nested too deep"));
+            None
+        } else {
+            // Numbers are arbitrary and were just picked from the stack docs
+            Some(stacker::maybe_grow(32 * 1024, 1024 * 1024, || f(self)))
```
jsternberg, line 915: Do we know if this stacker thing is safe regarding cgo? It seems like this is intended to grow the stack, but that stack was allocated by cgo.

jsternberg (commented):
Question about how stacker interacts with cgo. The code looks fine and the depth limits seem appropriate, but I'm worried that dynamic increasing of the stack may not be well documented when happening within a cgo context which is where this code ends up being used.

libflux/flux-core/src/parser/mod.rs
Marwes: Ah damn, I just added it as an additional safety. Maybe I should just remove it and accept the risk of stack overflow and if we see another crash due to it we just continue to lower the depth limit

jsternberg: I'd really like to learn if we can expand the stack. Please don't throw out this code, but maybe we should separate it and come up with a plan to understand how this interacts with cgo?

cgo does some strange things when it comes to the stack and I'm not sure I understand it in detail enough to know how it interacts with this. In summary, Go runs with pretty small stacks because the go runtime has its own way of growing the stack interactively. Since C code generally doesn't, the Go code pre-allocates a larger stack before calling into cgo code. The Rust code runs in this context.

It _might_ be safe to use this because the stack should mostly look exactly like a C program would see it, but I just don't know enough about it.

I also think that cgo calls run in a different pthread from the goroutine that called it. So I think there's pthreads that are "cgo" pthreads (with a C stack) and goroutines run in their own pthreads. The goroutines are managed by the Go runtime to run within one of the pthreads (goroutines do not spawn new pthreads). The cgo code runs in its own pthread so when you make a cgo call you're either using one of the existing pthreads for cgo or it autospins up a new one. It then runs the code in the pthread and waits for the response.

Marwes: `stacker` seems to work as it should when I run the crashing script through `go run ./cmd/flux`. Googling a bit I can't find anything that would say otherwise. Looking at the docs for stacker https://docs.rs/stacker/latest/stacker/ it seems to  allocate a new stacck and run the closure there, returning to the orignal stack once the closure returns.

So I think it should be fine, I can't find any reason that it shouldn't work (and I wouldn't expect the `stacker` path to really execute since we also have the depth to guard against these kinds of expressions)

jsternberg: Ok that seems relatively safe. I don't think we need to do any substantial testing if it works by extending the stack with an extra thread.

jsternberg (approved):

Marwes:
Seems to be some issues with wasm might just disable it on wasm, though it seems like stacker should work with wasm

sunbryely-influxdata (approved):

PR #4879, opened 2022-06-15
fix: Only error in tests if an assertion fails

Marwes:
To guard against accidentally writing tests without any assertions this also checks something is always yield to `error`.

Closes https://github.com/influxdata/flux/issues/4300


stdlib/testing/testing.flux
```diff
@@ -132,13 +140,24 @@ builtin assertEmpty : (<-tables: stream[A]) => stream[A]
...
+    {
+        return
+            _diff(
+                got,
+                want,
+                verbose,
+                epsilon,
+                nansEqual,
+            )
+                |> yield(name: "error")
```
wolffcm, line 159: Is there a reason this needs to be called `error`? I would find `diff` more informative.

cmd/flux/test.go
```diff
@@ -52,26 +52,37 @@ func (testExecutor) Run(pkg *ast.Package) error {
 
 	var output strings.Builder
```
wolffcm, line 53: Maybe change the name of this variable to reflect that it can only contain output from `diff`?

wolffcm (changes requested):
This seems reasonable to me, but I found some of the names confusing.

stdlib/testing/testing.flux
Marwes: It may output a diff here, but since this marks the failure of a test I'd say `error` (or `assert`, `testfail` etc) is more descriptive of the `yield`. You could envision sending a plain error message here instead to indicate that the test failed.

cmd/flux/test.go
Marwes: If we add `testing.shouldError` the output could contain something else than a diff, so I think that may be too specialized.

cmd/flux/test.go
```diff
@@ -52,26 +52,37 @@ func (testExecutor) Run(pkg *ast.Package) error {
...
+			foundErrorResult = true
+
+			err := result.Tables().Do(func(tbl flux.Table) error {
+				// The data returned here is the result of `testing.diff`, so any result means that
+				// a comparison of two tables showed inequality. Capture that inequality as part of the error.
+				// XXX: rockstar (08 Dec 2020) - This could use some ergonomic work, as the diff output
+				// is not exactly "human readable."
+				_, _ = fmt.Fprint(&output, table.Stringify(tbl))
+				return nil
+			})
```
wolffcm, line 69: It would be good to run `Tables.Do()` and then `tbl.Do()` on the tables in the other results, I can't think of a scenario where it would cause a deadlock not to, but it is good practice.

wolffcm (approved):
Looks good, thanks for the explanations. What you have here makes sense in the context of a `shouldError` usage.

PR #4882, opened 2022-06-15
fix: Point to the function being piped to on argument mismatches

Marwes:
We can probably explain this even better, but just changing this to point to `yield()` instead of `findColumn` seems a good start.


onelson (approved):
Yeah, flagging the receiver in this case makes sense.

PR #4885, opened 2022-06-15
fix(table_fns): make findColumn() able to handle multi-buffer tables

### LINKED ISSUE #4884: Wrong answer when input to findColumn is spread across multiple buffers
This test case will currently fail:
```js
testcase findColumnWithGroup {
    // The result of grouping and then ungrouping
    // will be for a single table to be spread over
    // four buffers.
    input = array.from(rows: [
        {m: "m", k: "north", v: 10, _time: 2020-02-20T00:00:00Z},
        {m: "m", k: "south", v: 20, _time: 2020-02-20T00:00:00Z},
        {m: "m", k: "east", v: 30, _time: 2020-02-20T00:00:00Z},
        {m: "m", k: "west", v: 40, _time: 2020-02-20T00:00:00Z},
    ])
       |> group(columns: ["k"])
       |> range(start: -100y)
       |> group()

    arr = input |> findColumn(fn: (key) => true, column: "v")
    got = array.from(rows: [{val: display(v: arr)}])
    want = array.from(rows: [{val: display(v: [10, 20, 30, 40])}])
    testing.diff(got, want)
}
```
It fails like this on current master:
```
#  _diff=string,val=string
_diff=-,val=[
    10,
    20,
    30,
    40
]
_diff=+,val=[40]
```

That is, the result should be `[10, 20, 30, 40]` but it is `[40]`. This happens when the table input to `findColumn` and related functions is spread across multiple buffers which will happen in the query above. In this case, rather than concatenating the values in each row, it will simple take the single value in the last buffer.

wolffcm:
The code for `findColumn` was not dealing with tables with multiple buffers, and each subsequent buffer would overwrite the answer for the previous. This PR addresses that and adds some more tests.

Fixes #4884.

### Done checklist
- [x] Test cases written

### DIFF
--- a/stdlib/universe/table_fns_test.flux
+++ b/stdlib/universe/table_fns_test.flux
@@ -1,6 +1,7 @@
 package universe_test
 
 
+import "array"
 import "csv"
 import "testing"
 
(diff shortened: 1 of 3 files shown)


sanderson (approved):

stdlib/universe/table_fns_test.flux
```diff
@@ -36,3 +37,71 @@ testcase sum {
...
+    }
+    wantObj = {
+        len: 4,
+        has10: true,
+        has20: true,
+        has30: true,
+        has40: true,
+    }
+    got = array.from(rows: [{value: display(v: gotObj)}])
+    want = array.from(rows: [{value: display(v: wantObj)}])
```
OfTheDelmer, line 82: nice!

OfTheDelmer (approved):

PR #4889, opened 2022-06-16
ci: fix the docker image build for go 1.18

jsternberg:
go 1.18 no longer allows `go get` to be used for installing binaries. We
also can't use it outside of a repository without a version specifier.

This modifies the build to use the version in `go.mod` for installing
`pkg-config` and installs it in an alternate and safer way.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
Thanks for looking at this!

Issue #4836, opened 2022-06-06
Add flag to enable/disable suggestions in REPL


onelson:
The suggestions currently offered by the REPL tend to not be high-quality. Additionally, the UI that presents them seems to exacerbate other rendering issues seen in the REPL. 

As a quick fix before diving into #4793, we might add a way to control whether or not suggestions are offered by the REPL.

DOD:
- add a new CLI flag to either opt-in or opt-out of having suggestions in the REPL

PR #4891, opened 2022-06-16
feat(repl): Suggestions now start off by default and added a new flag

Westonside:
## Description

Added in a new flag to run the repl with suggestions and by default the real will start without suggestions. Addressed issue #4836. Updated docs to remove old argument in running the repl and showed how to run repl with the new flag.

#### With suggestions:

``` ./repl --enable-suggestions ``` 

#### Without suggestions :

```./repl```

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,11 @@ $ which -a pkg-config
 /usr/bin/pkg-config
 ```
 
-To compile and use the REPL, use the following command:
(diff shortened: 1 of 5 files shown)


onelson (approved):
LGTM. Was there a related update to `README.md` coming?

Issue #4678, opened 2022-04-18
POC: Introduce the concept of node attributes to the planner


scbrickley:
In order to implement #4628, the planner needs to understand what nodes do and don't disrupt sorting. It would be helpful if each node had a list of attributes that the planner could reference when making judgments about how to rewrite the query plan.

Their is some prior art for this regarding parallel queries. Using that work as a reference, and generalizing it to handle other node attributes should get us the functionality we need.

@jsternberg Has a POC that may be useful to review for whoever picks up this issue: https://github.com/influxdata/flux/issues/2131

We should be able to look at a plan and verify that `MergeJoin` is receiving data sorted by the columns defined in the `EquiJoinProcedureSpec`

We should have a way of verifying that a physical plan nodes required attributes are satisfied by its predecessors. We do not want to accidentally remove a necessary sort node, or otherwise create an invalid plan. Perhaps we need a final validation step in the planner.

We also want to make sure that we can propagate certain attributes across the query plan. Nodes should be able to "ask" their parents about the attributes they provide. If the answer is unclear, those nodes should reference their parents, and so on and so forth.

DOD:

- A `PhysicalAttr` exists that signifies whether or not a node outputs/preserves sorted data, and if so, which columns that data is sorted by
- The solution should be able to support other types of attributes too
- If needed, follow-up issues are created to track work on a finished implementation

onelson:
Probably needs a design/working group session to decide how to proceed.

onelson:
# Prior Art

The work done towards parallel execution of portions of the plan has introduced the concept of _attributes_ which are used to signal properties/qualities of the nodes in the graph.

I first came into contact with this while working on #4465 since there were some overlapping concerns around calculating the concurrency quota for a plan.

In the executor, I saw access to `ppn.OutputAttrs`, which is one of 2 attribute sets on a `PhysicalPlanNode`.

https://github.com/influxdata/flux/blob/03b81dd347830e679f169a45ee04e48ed23b8605/plan/physical.go#L282-L286

In the executor, as we visit the nodes in the plan, we can inspect the attributes and possibly extract values they might be carrying.

https://github.com/influxdata/flux/blob/03b81dd347830e679f169a45ee04e48ed23b8605/execute/executor.go#L174-L198

From what I could tell, the pattern in this case is for attributes to be initialized during planning, then consumed during execution.

wolffcm:
Here's an overview of how things currently work with attributes for parallelized execution. We can parallelize queries of the form
[...]

PR #4892, opened 2022-06-16
test(plan): make planner test helper validate by default

wolffcm:
Previously by default we would not perform validation steps
in physical planner tests since it presented a burden on the
test writer to make realistic plans and spend time on things
they are not actually testing.

This is an additive change that makes it so by default validation is
performed on the resulting plan, but it may be skipped if needed.
For tests that began failing due to validation failures, I have skipped
them.

Validation is a step performed after planning to make sure the plan
that is about to be executed is in fact sane. Validation will become more
important as we increase the use of physical attributes, as we will for #4678.


scbrickley (approved):
Seems reasonable to me. Thanks for doing this!

PR #4901, opened 2022-06-21
fix: Add required fields to the signature of experimental/to

Marwes:
The documentation implies that these fields are mandatory to call this function and looking at the implementation it will indeed error if they are missing. Changing this is still technically a breaking change so maybe we want some caution before merging this? (https://github.com/influxdata/flux/pull/4773 could perhaps be used to allow an incremental rollout, or I could run this through the query log analyzer I made for https://github.com/influxdata/flux/pull/4776)

### DIFF
--- a/libflux/go/libflux/buildinfo.gen.go
+++ b/libflux/go/libflux/buildinfo.gen.go
@@ -154,7 +154,7 @@ var sourceHashes = map[string]string{
 	"stdlib/experimental/count_test.flux":                                                         "120a145eb287266e35638cc5220f5cecf0eee5bf2eec1bb1e9852e7d382050c7",
 	"stdlib/experimental/csv/csv.flux":                                                            "94ae72b5fb50b8d65e9b384aebaa890f04bc914bb344ea751a64f729b5c849fe",
 	"stdlib/experimental/distinct_test.flux":                                                      "6d2891556d985d16ac9bd33176dd527bb34246afa1ac81dde65f1335885b6b02",
-	"stdlib/experimental/experimental.flux":                                                       "5543ad639409a8b1dcef963687eebb45c688d304fb9aa9e2d7f2186b1330d19d",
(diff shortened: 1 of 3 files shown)


wolffcm (approved):
Incidentally, I wrote the `experimental.to` code. It just seems like the user would get a different error if `_measurement` or `_time` was missing? This seems totally fine to me.

If you are still cautious, you might try out this change in remocal and verify that the new behavior isn't any worse.

Marwes:
Yeah, in theory it would just be a compiletime error instead of a runtime one. But there may be some weirdness due to type inference changing or perhaps the current runtime error doesn't fire for a script as the execution never reaches that point

wolffcm:
@marwes 
> But there may be some weirdness due to type inference changing or perhaps the current runtime error doesn't fire for a script as the execution never reaches that point

I'm not sure what you mean here. Are you thinking of a particular case? 

It does seem that a query that sends no data to `experimental.to` would succeed today. It could be that users use `to` as a way to log infrequent error conditions in their application. However, this mechanism would have never succeeded for them, if `_measurement` and `_time` were missing, as both fields are truly required to write data to InfluxDB.

We developed `experimental.to` as a way to enable the alerting framework, which uses Flux code generated by the UI. My sense is that almost all uses of `experimental.to` are for this. I think if you were to run some of these generated queries through type inference and they continue to succeed, then we are probably pretty safe here.

Marwes:
> I'm not sure what you mean here. Are you thinking of a particular case?

The case of no data is one, another could be just a random statement that never gets executued.

```
if false then
    array.from(rows: [{}]) |> experimental.to()
```
which would now fail at compile time.

> We developed experimental.to as a way to enable the alerting framework, which uses Flux code generated by the UI. My sense is that almost all uses of experimental.to are for this. I think if you were to run some of these generated queries through type inference and they continue to succeed, then we are probably pretty safe here.

Where do I find these?

PR #4902, opened 2022-06-21
feat: [POC] generalize attributes in planner

### LINKED ISSUE #4678: POC: Introduce the concept of node attributes to the planner

wolffcm:
This is just a proof of concept. Do not merge as-is!

Fixes #4678.

This PR contains some changes to allow for operations in the plan graph to model the attributes that they require and provide. In this PR there are two sets of attributes:
- Parallelization-related. I moved these into their own file.
- Collation (order of rows within tables). This is a new kind of attribute added in this PR.

The main goal of this PR is to make it so parallelization and collation can use the same framework for providing, propagating and require attributes to ensure a correct plan. The file `attributes.go` has a lenghty comment that provides an overview of how things are supposed to work.

In the example below, `SortMergeProcedureSpec` indicates that it requires sorted input by implementing an interface, and `SortProcedureSpec` indicates that it provides a `collation` attribute by implementing a different interface.

There are some breaking changes here and the parallelization attributes no longer work right. There will need to be some changes both here and cloud to make everything work. These are the relevant breaking changes:
- Attributes are no longer stored on the physical plan node itself as Go `map`s. Instead the plan node to will need to provide methods to access attributes, which it will delegate to the `ProcedureSpec` that it wraps.
- As such, it's no longer possible to modify a plan node to provide or require a certain attribute. If there is state that needs to be managed, the state should be contained within the `ProcedureSpec`, which can be manipulated via a planner rule.
- Required attributes are no longer represented as a map, but rather a slice of maps. This is because there can be different requirements for different inputs to an operation (`join` is like this)

Interestingly, this work actually caught a bug in the planner rule for `SortMergeJoin`, because the sort nodes weren't being created quite right.

### DIFF
--- a/plan/attributes.go
+++ b/plan/attributes.go
@@ -1,27 +1,81 @@
 package plan
 
-// Physical attributes used in specifying parallelization. The Run attribute
-// means the node executes in parallel. It accepts parallel data (a subset of
-// the source) and produces parallel data. The merge attribute means that the
-// node accepts parallel data, merges the streams, and produces non-parallel
-// data that covers the entire data source.
-const ParallelRunKey = "parallel-run"
-const ParallelMergeKey = "parallel-merge"
+// Attributes provide a way to model different aspects of the data
+// flowing though the physical operations in a plan graph.
+//
+// For example, if a node requires its input tables to be sorted, it
+// will have the CollationAttr among its required attributes.
+// Likewise, if a node produces sorted tables, its output attributes
+// will have CollationAttr.
+//
+// Operations can require or provide attributes by implementing interfaces
+// in the corresponding PhysicalProcedureSpec:
+// - OutputAttributer is to be implemented if the procedure will provide an attribiute. E.g.,
+//   the SortProcedureSpec will provide CollationAttr for the columns on which the data is
+//   to be sorted.
+// - PassThroughAttributeris to be implemented if a procedure will not perturb a given attribute.
+//   E.g., if data with CollationAttr flows into a filter, it will still be sorted. Therefore,
+//   FilterProcedureSpec can implement PassThroughAttributer for collation.
+// - RequiredAttributer is to be implemented by procedures that require particular attreibutes.
+//   There is one set of required physical attributes for each input, since they may be different.
+//   E.g., SortMergeJoinProcedureSpec requires that the left and right inputs be sorted on the
+//   columns being joined on (and they may in fact have different names on either side).
+//
+// It's the obligation of planner rules to ensure that required attributes are satisified by
+// a procedure's inputs. If a node has required attribute that are not satisfied, it will be
+// caught by ValidatePhysicalPlan(), and an internal error will be returned.
 
-type ParallelRunAttribute struct {
-	Factor int
+// PhysicalAttr represents an attribute (collation, parallel execution)
+// of a plan node.
+type PhysicalAttr interface {
+	Key() string
+	SuccessorsMustRequire() bool
+	SatisfiedBy(attr PhysicalAttr) bool
 }
 
-// If a node produces parallel data, then all successors must require parallel
-// data, otherwise there will be a plan error.
-func (ParallelRunAttribute) SuccessorsMustRequire() bool {
-	return true
+// PhysicalAttributes encapsulates any physical attributes of the result produced
+// by a physical plan node, such as collation, etc.
+type PhysicalAttributes map[string]PhysicalAttr
+
+// OutputAttributer is an interface to be implemented by PhysicalProcedureSpec implementations
+// that produce output that has particular attributes.
+type OutputAttributer interface {
+	OutputAttributes() PhysicalAttributes
 }
 
-type ParallelMergeAttribute struct {
-	Factor int
+// PassThroughAttributer is an interface to be implemented by PhysicalProcedureSpec implementations
+// that allow attributes to propagate from input to output.
+type PassThroughAttributer interface {
+	PassThroughAttribute(attrKey string) bool
 }
 
-func (ParallelMergeAttribute) SuccessorsMustRequire() bool {
-	return false
+// RequiredAttributer is an interface to be implemented by PhysicalProcedureSpec implementations
+// that require physical attributes to be provided by inputs. The return value here is a slice,
+// since each input may be required to have a different set of attributes.
+type RequiredAttributer interface {
+	RequiredAttributes() []PhysicalAttributes
+}
+
+// NodeSatisfiesRequiredAttribute returns true if the given node can provide the given attribute.
+// An attribute can be provided if the node provides it directly, or if the node passes through
+// an attreibute from a predecessor.
+func NodeSatisfiesRequiredAttribute(node *PhysicalPlanNode, requiredAttr PhysicalAttr) bool {
+	gotAttr := getAttribute(node, requiredAttr.Key())
+	return requiredAttr.SatisfiedBy(gotAttr)
+}
+
+func getAttribute(node *PhysicalPlanNode, attrKey string) PhysicalAttr {
+	if attr, ok := node.OutputAttrs()[attrKey]; ok {
+		return attr
+	}
+
+	if passer, ok := node.Spec.(PassThroughAttributer); ok {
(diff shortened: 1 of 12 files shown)


stdlib/join/sort_merge_join.go
```diff
@@ -72,23 +87,25 @@ func (SortMergeJoinPredicateRule) Rewrite(ctx context.Context, n plan.Node) (pla
...
 	successors = predecessors[1].Successors()
 
-	columns = make([]string, len(spec.On))
+	rightColumns := make([]string, 0, len(spec.On))
 	for _, pair := range spec.On {
-		columns = append(columns, pair.Right)
+		rightColumns = append(rightColumns, pair.Right)
 	}
-	successors[0] = makeSortNode(predecessors[1], columns)
+	successors[0] = makeSortNode(predecessors[1], rightColumns, "sortRightSideJoinInput")
```
wolffcm, line 102: Note: there was a bug here where the `columns` array contained an empty string `""` for its initial elements. This was actually caught by `ValidatePhysicalPlan()` since the sort nodes did not appear to provide the correct order to the join. I will file a separate PR to fix this.

stdlib/join/sort_merge_join.go
```diff
@@ -33,6 +33,21 @@ func (p *SortMergeJoinProcedureSpec) Copy() plan.ProcedureSpec {
...
+			},
+		},
+		{
+			plan.CollationKey: &plan.CollationAttr{
+				Columns: getJoinKeyCols(p.On, false),
+			},
+		},
+	}
+}
+
```
scbrickley, line 50: Is the list of required attributes an `or` relationship? As in "predecessors must have one of these attributes, but not all"? Or is each required attribute somehow keyed to a specific parent node?

wolffcm: Yes, each attribute (really a map of attributes) for each predecessor input to the join node. So this is requiring that the LHS of the join be sorted by the LHS join columns, and the RHS of the join must be sorted by the RHS join columns.

This assumes that the predecessors are ordered with `node.Predecessors()[0]` always being the left and `node.Predecessors()[1]` always being the right. That should hold true, right?

scbrickley: Ah, I see. Didn't realize each item was a map. That makes much more sense now.

And yes, that assumption should always hold true.

plan/attributes.go
```diff
@@ -1,27 +1,81 @@
...
 }
 
-type ParallelMergeAttribute struct {
-	Factor int
+// RequiredAttributer is an interface to be implemented by PhysicalProcedureSpec implementations
+// that require physical attributes to be provided by inputs. The return value here is a slice,
+// since each input may be required to have a different set of attributes.
+type RequiredAttributer interface {
+	RequiredAttributes() []PhysicalAttributes
 }
```
scbrickley, line 57: Will every procedure spec have to implement these `*Attributer` interfaces? Or will it be assumed that if a spec doesn't implement `RequiredAttributer` (as an example) that it has no required attributes?

wolffcm: Yes, it's the latter case. If there are no required attributes, then no changes need to be made. Likewise for the other interfaces, operations won't produce or pass on any attributes of the interfaces are unimplemented.

scbrickley: So in the case of collation, any transformation that doesn't disrupt sorting will will have to implement `PassthroughAttributer` correct?

wolffcm: If we want to be able to remove sort nodes that come after it, then yes. It won't be a wrong answer if we forget to implement it, but we might get a slower query.

I think we'll see a limited number of nodes between the sort and a source. It might work something like this for example:
```js
ReadFilter // implements OuptutAttributer with CollationAttr
  |> filter(...) // filter that could not be pushed down; implements PassThroughAttributer for collation
// |> group() // if you have this here it would disrupt collation; no need to implement anything new
  |> sort() // could be removed by a planner rule

```
Regardless of the mechanism, we'll need to look at each transformation and decide if it maintains collation and implement `PassThroughAttributer` for it if so. We definitely don't want to make it a default of any kind.

I can think of other ways to do it, like keeping a list of all the transformations that maintain collation, but this is the most generic way I could think of that could be extended to other kinds of attributes.

scbrickley: This is not a strong opinion, and I don't immediately know what it would take to implement this, but I think it might be simpler if we instead had nodes report that they disrupt sorting, rather than reporting that they preserve it. So if a node doesn't explicitly report that it disrupts sorting, we assume sorting remains unchanged. Seems like it would accomplish the same thing and require less boilerplate code.

Again, this is opinion is loosely held, and I'm interested in hearing any pushback to it. Ultimately, this patch seems like a good solution, and I'd be content if this ends up getting merged.

wolffcm: > So if a node doesn't explicitly report that it disrupts sorting, we assume sorting remains unchanged.

I have concerns with this idea. What I really want to avoid is a case where an operation needs sorted input but due to programmer error it receives unsorted data and produces a wrong answer. This would happen with merge join for example.

I would much rather we see queries be slow because we are sorting redundantly than get a wrong answer the the customer might not realize.

It could result in some boilerplate, but that could be minimized with a mixin:
```Go
type PassThroughCollation struct{}
func (ptc PassThroughCollation) PassThroughAttribute(attrKey string) bool {
  return attrKey == plan.CollationKey
}

// filter.go

type FilterProcedureSpec struct {
  PassThroughCollation
  // other fields...
}
```

With this approach, we would need to just add a single line to the procedures that pass on collation to successors.

scbrickley: >What I really want to avoid is a case where an operation needs sorted input but due to programmer error it receives unsorted data and produces a wrong answer ... I would much rather we see queries be slow because we are sorting redundantly than get a wrong answer the the customer might not realize.

That is a fair point that had not occurred to me. Consider me convinced.

In that case, this patch looks good. Assuming we can resolve the problems with the parallel attribute, I'd be happy to see something like this get merged.

wolffcm: Thanks, I'll file an issue with specifics on how to make this production-ready and put it in the epic.

adrian-thurston:
So plan nodes that compute their output attributes and their required attributes must merge from all the possible sources of that information ... which at this point would be a merge of collation-related attributes and paralllelization attributes. Is that right?

adrian-thurston:
> @adrian-thurston
> 
> > So plan nodes that compute their output attributes and their required attributes must merge from all the possible sources of that information ... which at this point would be a merge of collation-related attributes and paralllelization attributes. Is that right?
> 
> Yes, that's right. So for say for RWA, you would want it to implement `OutputAttributer` returning a map with with both `CollationAttr(_time)` and `ParallelRun` in it. Does that make sense?

@wolffcm Yeah all makes sense. It's less convenient because it will require more code in the plan nodes, but it allows for more features so it's the way to go.

adrian-thurston:
I'm excited to see the optimizations that will be possible as a result of this.

wolffcm:
Closing in favor of #4982

PR #4895, opened 2022-06-17
chore(docs): various typo fixes and doc updates

sanderson:
Various updates to the inline documentation to support the stdlib documentation that gets auto-generated. Changes include:

- Typo fixes
- Updated contributor spec
- Updated contrib packages to include contributors contact info
- Function introduced/deprecated version updates

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written **N/A**

PR #4905, opened 2022-06-22
feat: Format scripts with a trailing newline by default

Marwes:
This was already the case for the command line tool which appended its own newline, however that would only be added if the formatted file changed in other ways. I'd argue that you usually want the trailing newline as diffs where you append to a script get slightly messed up without it (github warns when you have files without this trailing newline actually).

So this makes things consistent and should hopefully reduce the risk of the LSP based formatter formatting differently from `make fmt-flux` (see https://github.com/influxdata/flux/pull/4895#discussion_r902994426). There is an option to add a trailing newline in the LSP protocol (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions) but it isn't enabled by default.

 (cc @rockstar The LSP seems to have some explicit handling around trailing newlines)

BREAKING CHANGE:

The formatter now defaults to adding a trailing newline


cmd/flux/fmt.go
```diff
@@ -87,7 +86,7 @@ func format(script string) (bool, error) {
 }
 
 func updateScript(fname string, script string) error {
-	err := ioutil.WriteFile(fname, []byte(script+"\n"), 0644)
+	err := ioutil.WriteFile(fname, []byte(script), 0644)
```
jsternberg, line 89: This is because it's done by the rust formatter now, right?

jsternberg (approved):

cmd/flux/fmt.go
Marwes: Yup

Marwes: So we did this correctly via the command line, just not in the LSP which only called into format. Since this was appended after we did the check to see if the formatting changed we didn't end up adding this newline to a script which did lack a newline unless the formatting changed for other reasons. So we ended up having a few files that without this newline also

Marwes:
Changing the default here forced me to update a bunch of tests on the go side which depends on the formatted output. I think this will cause some some downstream tests to break as well which isn't great. Should we just accept this breakage and update the tests? Add an format function which omits the newline just for these tests? Or something else?

PR #4910, opened 2022-06-23
feat: Remove test statements

### LINKED ISSUE #4768: Convert All End to End Test to Acceptance Tests

Marwes:
With all `test` statements being converted to `testcase` and neither being used outside of
testing flux (right? no one else is using either?). We can safely remove this old way of writing tests.

Closes #4768


nathanielc (approved):
LGTM I am confident no one outside the Flux repo is using the `test` statement. I checked with the edge team earlier this week to be safe.

PR #4915, opened 2022-06-23
fix(repl): replace extra boolean parameter for suggestions with repl options

jsternberg:
This uses the option pattern to configure the repl with enabling
suggestions as the currently available option. This option defaults to
false. The `repl.New()` function has also been changed back to accept
only a single argument, but to also allow an optional set of options to
configure the REPL.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
LGTM. Thanks for smoothing this over @jsternberg !

wolffcm (approved):
Looks good. I think an update for cloud will also be needed.

PR #4917, opened 2022-06-24
refactor(test): second refactor of the test executor

jsternberg:
The previous changes to the test executor worked in one environment, but
was too difficult to refactor with the defers in the second test
environment. This changes the flow back so the entire test is executed
inside of run and handles reading the result by using a callback
function instead.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/cmd/flux/cmd/test.go
+++ b/cmd/flux/cmd/test.go
@@ -162,18 +162,16 @@ func (t *Test) Error() error {
 
 // Run the test, saving the error to the err property of the struct.
 func (t *Test) Run(executor TestExecutor) {
-	results, err := executor.Run(t.ast)
-	if err != nil {
-		t.err = err
-		return
-	}
-	defer results.Release()
+	t.err = executor.Run(t.ast, t.consume)
 
+}
+
+func (t *Test) consume(ctx context.Context, results flux.ResultIterator) error {
 	var output strings.Builder
 	for results.More() {
 		result := results.Next()
 		if result.Name() == errorYield {
-			t.err = result.Tables().Do(func(tbl flux.Table) error {
+			err := result.Tables().Do(func(tbl flux.Table) error {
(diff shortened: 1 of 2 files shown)


cmd/flux/cmd/test.go
```diff
@@ -801,10 +800,13 @@ func (t *TestReporter) Summarize(tests []*Test) {
...
 
-type TestSetupFunc func(ctx context.Context) (TestExecutor, error)
+type (
+	TestSetupFunc    func(ctx context.Context) (TestExecutor, error)
+	TestResultReader func(ctx context.Context, results flux.ResultIterator) error
+)
 
 type TestExecutor interface {
-	Run(pkg *ast.Package) (flux.ResultIterator, error)
+	Run(pkg *ast.Package, read TestResultReader) error
```
nathanielc, line 809: Can you add a comment here for the purpose of the read function? Does the read function need to guarantee that the result is fully consumed? Or can it just drop it on the floor?

cmd/flux/cmd/test.go
```diff
@@ -162,46 +162,45 @@ func (t *Test) Error() error {
...
-	if err != nil {
-		t.err = err
-		return
-	}
-	defer results.Release()
+	t.err = executor.Run(t.ast, t.execute)
 
+}
+
+func (t *Test) execute(ctx context.Context, results flux.ResultIterator) error {
```
nathanielc, line 169: Why is this called execute when its supposed to just be reading the result? I found that a bit confusing. Maybe `consume` or just `read` is a better name?

nathanielc (changes requested):
Looks good but had a few questions.

cmd/flux/cmd/test.go
jsternberg: I renamed it from `execute` and forgot to change this name.

nathanielc (approved):

PR #4920, opened 2022-06-24
chore: move unpivot from internal/debug to experimental

wolffcm:
I needed to use unpivot in another repo, but since it's under
a directory called "internal" it isn't visible. This PR just moves
it out of `internal/debug`.

My understanding of `internal/debug` is that it's for functions that can only be used from elsewhere within the standard library. It's unfortunate that the Go code in there can't be used outside of this repo. Maybe we should give it a different name?


stdlib/join/sort_merge_join_test.go
```diff
@@ -115,7 +115,7 @@ func TestSortMergeJoinPredicateRule(t *testing.T) {
 			// compare the sort nodes created by the planner rule
 			err = plantest.ComparePlans(wantPlan, physicalPlan, func(p, q plan.Node) error {
 				ps, qs := getSortSpec(p), getSortSpec(q)
-				if (ps == nil)  && (qs == nil) {
```
onelson, line 118: Doesn't seem relevant to the task at hand, but welcome nonetheless :rofl:

onelson (approved):
LGTM. Small whitespace change that seems unrelated -- could take it or leave it. No big deal.

stdlib/join/sort_merge_join_test.go
wolffcm: Yes, I think that `go fmt ./...` fixed it for me, even though it wasn't part of this PR. Incidentally I don't think that our CI format check will fail a job if there are formatting inconsistencies in a test, which is how it got in there in the first place.

PR #4924, opened 2022-06-24
feat(stdlib): promote experimental/array to array

nathanielc:
Merges the experimental/array package with array. We have see widespread
use and no open bugs for the experimental/array package.

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


stdlib/array/array_test.flux
```diff
@@ -4,12 +4,92 @@ package array_test
...
+                {_value: "z"},
+                {_value: "a"},
+                {_value: "b"},
+                {_value: "c"},
+            ],
+        )
+
+    testing.diff(got, want)
+}
+testcase array_concat_to_empty {
```
OfTheDelmer, line 39: should we add a testcase with passing an empty array also?

OfTheDelmer (approved):

PR #4933, opened 2022-06-28
fix(join): return error if user modifies group key

### LINKED ISSUE #4909: Join package: report error if user attempts to modify the group key
Right now, users can define `as` functions that drop group key columns in the output records. The output tables still retain the same grouping, but the missing columns will cause problems elsewhere in the product. The join transformation should protect against this by throwing an error if the `as` function returns a record that omits any group key columns.

scbrickley:
Return an error if a user defines an `as` function that returns a record that is missing a group key column.

Fixes #4909

### DIFF
--- a/stdlib/join/join_fn.go
+++ b/stdlib/join/join_fn.go
@@ -131,7 +131,7 @@ func (f *JoinFn) crossProduct(ctx context.Context, p *joinProduct, mem memory.Al
 			}
 
 			if f.schema == nil {
-				cols, err := f.createSchema(joined)
+				cols, err := f.createSchema(joined, p.left[0].Key())
 				if err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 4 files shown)


stdlib/join/join_fn.go
```diff
@@ -211,6 +211,22 @@ func (f *JoinFn) createSchema(record values.Object) ([]flux.ColMeta, error) {
 			Type:  ty,
 		})
 	}
+	for _, gcol := range groupkey.Cols() {
+		found := false
+		for _, col := range cols {
+			if gcol == col {
+				found = true
```
OfTheDelmer, line 218: We could `break` after here. Although, if cols is sorted would a quick binary search be worthwhile?

OfTheDelmer (approved):

Issue #4930, opened 2022-06-27
Fixups for `date/boundaries`


onelson:
After discussing #4868 (which is a "wontfix"), it seemed as though there might be parts of `date/boundaries` that depend on the `offset` of a location, which could lead to _incorrect answers_ depending of the _specifics of the question_.

The current test suite deals specifically with locations with fixed offsets, and so there could be problems at the boundaries of daylight savings.

In spite of the questions raised re: correctness, I cobbled together a few testcases to check the behavior today.

<details>
<summary>
repro_test.flux
</summary>

```
package repro_test

import "array"
import "date/boundaries"
import "testing"
import "timezone"

testcase monday_test_tz_fixed_timeable {
    option now = () => 2021-12-30T00:40:44Z
    option location = timezone.fixed(offset: -8h)

    ret = boundaries.monday()

    want = array.from(rows: [{_value_a: 2021-12-27T08:00:00Z, _value_b: 2021-12-28T08:00:00Z}])
    got = array.from(rows: [{_value_a: ret.start, _value_b: ret.stop}])

    testing.diff(want: want, got: got)
}

testcase monday_test_tz_named_st {
    // DST begins March 14, ends Nov 7.
    option now = () => 2021-12-30T00:40:44Z
    option location = timezone.location(name: "America/Los_Angeles")

    ret = boundaries.monday()

    want = array.from(rows: [{_value_a: 2021-12-27T08:00:00Z, _value_b: 2021-12-28T08:00:00Z}])
    got = array.from(rows: [{_value_a: ret.start, _value_b: ret.stop}])

    testing.diff(want: want, got: got)
}

testcase monday_test_tz_named_dst {
    // DST begins March 14, ends Nov 7.
    option now = () => 2021-10-29T00:40:44Z
    option location = timezone.location(name: "America/Los_Angeles")

    ret = boundaries.monday()

    want = array.from(rows: [{_value_a: 2021-10-25T07:00:00Z, _value_b: 2021-10-26T07:00:00Z}])
    got = array.from(rows: [{_value_a: ret.start, _value_b: ret.stop}])

    testing.diff(want: want, got: got)
}

testcase month_straddling_offset_change {
    // DST begins March 14, ends Nov 7.
    option now = () => 2021-11-21T00:40:44Z
    option location = timezone.location(name: "America/Los_Angeles")

    ret = boundaries.month()

[...]

PR #4934, opened 2022-06-28
fix: Added in support for the timezone.location in boundaries package

Westonside:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### Updates
Updates in regards to #4930 

- Updated functions in the boundaries package to include support for `timezone.location()`
- Changed test names


stdlib/date/boundaries/boundaries.flux
```diff
@@ -387,7 +391,15 @@ sunday = () => {
...
-    as = _timezone_convert(s: s)
+
+    // if the timezone offset is not 0 then timezone convert else truncate
+    as =
+        if int(v: location.offset) != int(v: 0ns) then
+            _timezone_convert(s: s)
+        else
+            date.truncate(t: s, unit: 1d)
+
+    // as = _timezone_convert(s: s)
```
onelson, line 402: Better to remove unwanted lines than just comment them out.

stdlib/date/boundaries/tests/boundaries_test.flux
```diff
@@ -257,7 +257,21 @@ testcase thursday_test_tzt_timeable {
     testing.diff(want: want, got: got)
 }
 
-testcase friday_test_timeable {
+// Africa/Algiers
+testcase thursday_test_tzts_timeableasdffff {
```
onelson, line 261: This test could probably use a better name.

stdlib/date/boundaries/boundaries.flux
```diff
@@ -387,7 +391,15 @@ sunday = () => {
...
 month = (month_offset=0) => {
     s = date.truncate(t: today(), unit: 1mo)
-    as = _timezone_convert(s: s)
+
+    // if the timezone offset is not 0 then timezone convert else truncate
+    as =
+        if int(v: location.offset) != int(v: 0ns) then
+            _timezone_convert(s: s)
+        else
+            date.truncate(t: s, unit: 1d)
```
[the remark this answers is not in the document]
Westonside: You are right, it works without. Will remove

Westonside:
I added a temporary fix to the date.go file because the offset set by `timezone.offset` was being ignored by the truncate function.

stdlib/date/date.go
```diff
@@ -260,11 +260,14 @@ func init() {
 				if err != nil {
 					return nil, err
 				}
-				location, _, err := date.GetLocationFromObjArgs(args)
+				//was originally ignoring the offset _ so looked like location ,_,err Addressed in #5013
```
onelson, line 263:
```suggestion
```

stdlib/date/date.go
```diff
@@ -260,11 +260,14 @@ func init() {
...
+				location, offset, err := date.GetLocationFromObjArgs(args)
 				if err != nil {
 					return nil, err
 				}
 				intervalLocation, err := interval.LoadLocation(location)
+				//temporary fix change the location address issue #5013
```
onelson, line 269:
```suggestion
				// TODO: offset previously ignored. Is this enough of a fix?
				//  Need to check to see if setting it here is a sufficient fix. Follow-up in #5013
```

onelson (approved):
Offered a small tweak to the comments re: #5013. Looks good otherwise.

PR #4936, opened 2022-06-28
refactor(test): add integration tag for test 

nathanielc:
With this change we no longer need to maintain a list of integration
tests, rather we can declare the test as an integration test with the
testing.tags option.

Depends on https://github.com/influxdata/flux/pull/4935

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
:sparkles:

PR #4939, opened 2022-06-29
refactor: Run cargo clippy on all rust files

Marwes:
We didn't run clippy on any tests and benchmarks before.


Marwes:
ping @OfTheDelmer

skartikey (approved):

PR #4940, opened 2022-06-29
ci(docker): move the image build into circleci instead of quay triggers

jsternberg:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


Marwes (commented):
There are several `.flux` files included in this PR. Were those included by accident or is there a reason behind them?

PR #4948, opened 2022-06-30
refactor: Implement find_var_type in terms of the semantic graph

### LINKED ISSUE #4751: Refactor `FindVarType` to lookup multiple names at the same time
Today, the consumers of [`FindVarType`](https://github.com/influxdata/flux/blob/d0d82523879b39d773ec81cfe3d62aff916f9932/libflux/go/libflux/analyze.go#L150-L176
) need to use workarounds to be able to perform _multiple unknown type lookups_ at the same time.
Splitting the lookups across multiple calls seems to not be a viable option (either for perf or inference outcome reasons).

The _"here be dragons"_ text for the workaround follows:

```
	// FindVarType can only determine the type of a single unknown
	// type at a time. To work around this the query is parsed and
	// rewritten such that all of the unknown types are a records of
	// a single object that FindVarType can find the type for.
        // ... snip ...
```

This require the input flux to be rewritten, which sucks. All this could likely be avoided if we update our API perform multiple lookups together, returning the analysis results all at the same time.

Nb. since the func implementation is largely making FFI calls, we'll need to take special care with the memory handling for the new (assumed) array-based inputs and outputs.

DOD:
- pluralize the function name
- update the function such that multiple `varNames` can be received
- update the function such that an array of `MonoType`, or a mapping of names => `MonoType`s is produced

Marwes:
Changes `find_var_type` to just look at the semantic graph's type instead of an injected dummy variables. As part of this I had to make some changes to ensure that an undefined variable will always refer to the same thing, which may be a bit strange but I think it works fine (stop me otherwise!).

~While this doesn't actually change things for `find_var_type` as it could easily just define multiple variables with one call, I'd argue this exposes less of the compiler internals to users. If we can move `find_var_type` to accept a semantic graph directly we could reduce this further.~

Added `func FindVarTypes(script string, varNames []string) ([]semantic.MonoType, error)` which matches how this is used downstream. Left the current `FindVarTypes` as is to avoid a breaking change.

Alternatively we could move towards making the current `FindVarTypeSemantic` the only function we expose, which would limit our API while still letting downstream users implement the same functionality (by combining `AnalyzeString + FindVarType(Semantic)` as we do internally now.

Closes #4751


Marwes:
ping @jsternberg

jsternberg (approved):

PR #4949, opened 2022-06-30
feat: allow physical plan attributes to contribute to formatter details

adrian-thurston:
In order to determine if a query plan has been parallelized, we have been using
the presence of a `partitionMerge` node in the plan. With the emergence of a new
technique of merging in `aggregateWindow` nodes, and possibly soon `pivot` nodes as
well, we can no longer use this test.

This PR allows physical plan attributes to contribute to plan details. First
the spec is consulted, then any physical attributes that are present. Any node
with the `ParallelMerge` attribute can then indicate it is a merge node.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/plan/attributes.go
+++ b/plan/attributes.go
@@ -1,5 +1,9 @@
 package plan
 
+import (
+	"fmt"
+)
+
 // Physical attributes used in specifying parallelization. The Run attribute
 // means the node executes in parallel. It accepts parallel data (a subset of
 // the source) and produces parallel data. The merge attribute means that the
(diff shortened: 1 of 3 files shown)


onelson (approved):
LGTM. 

For my curiosity: I thought the formatter was mostly for diagnostic purposes (ie, println the plan while working on changes to verify the WIP). Does this output get consumed by something other than human eyeballs?

wolffcm:
@onelson I saw you asked this question and it never got answered. We put the plan into the profile if `query`profiling is turned on, so end users will see it as well. As far as I know nothing consumes it programatically.

PR #4953, opened 2022-07-01
fix(stdlib): http writer skips invalid floats properly

### LINKED ISSUE #4944: to function can return "no serializable fields" in a very confusing error message when field is invalid
The line protocol encoder that is used by the `to()` function can return "no serializable fields" in a location that otherwise looks like it has to have at least one field. The code inside of flux checks to make sure there's at least one field before attempting to write it, but it then uses the line protocol encoder in a way that could result in this error message.

The line protocol encoder has a method called [FailOnFieldErr](https://pkg.go.dev/github.com/influxdata/line-protocol#Encoder.FailOnFieldErr) to configure the behavior when a field is invalid in some way. The default behavior is to ignore the field. This could cause "no serializable fields" to show up as the error even though there is a field that was provided.

It would likely be better to set this value to true so the real reason why the field couldn't be encoded is bubbled up instead of the ambiguous "no serializable fields". The error message is technically correct, but also not very helpful.

There are 2 conditions where this seems to be able to happen:

1. The field key is invalid/empty.
2. The field value is a float and it is NaN or Inf.

There are possibly others but these are the ones I found.

When one of these happens, the `to()` function just skips the field and then reports the "no serializable fields" message instead of telling the user that the field they wrote had a problem.

I am not completely sure it is correct to error on this. We may want to enable the error, but also filter out float values of NaN or Inf so that they don't cause a failure.

jsternberg:
The http writer now skips invalid floats properly and errors when the
field key is wrong with a message describing the problem.

Previously, we had the line protocol writer configured to ignore errors
when writing fields. But, when that happens, the line protocol writer
will return an error if all of the fields had an error with the very
opaque message "no serializable fields".

In order to avoid this, we needed to do some validation or control the
writing of line protocol a bit better. There is a line protocol v2
library that exposes the validation behavior so this change updates to
use that library instead.

The new and intended behavior is that if no fields can be written due to
NaN or Inf, we ignore the point entirely. If we cannot write it because
the field key is invalid, we error with a message specific to that. If
some fields can be written and some can't due to NaN or Inf, we only
ignore the fields that cannot be written.

This change also updates the influxdb provider to expose the
`influxdb.Metric` type with `influxdb.Tag` and `influxdb.Field`. These
were previously exposed through the v1 line protocol library. These are
presently exposed using a type alias to keep existing code compatible,
but existing code should switch to referencing these new types so we can
remove our current references to the v1 line protocol library.

The mqtt and kafka writers also use the v1 line protocol library and
have not been touched for this change. I would assume they have the same
problem as the existing `to()` behavior.

In addition to the above, this also modifies both influxdb `to()` and
`wideTo()` so that they skip NaN and Inf values and never send them to
the underlying writer. This should help prevent the need to handle these
in the dependency provider, but it's still useful to fix the dependency
provider to give better error messages anyway.

Fixes #4944.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/influxdata/influxdb/wide_to.go
```diff
@@ -306,9 +307,17 @@ func (t *WideToTransformation) writeTable(tbl flux.Table) error {
 					continue
 				}
 
-				metric.Fields = append(metric.Fields, &lp.Field{
+				v := values.Unwrap(fieldVal)
+				if fv, ok := v.(float64); ok {
+					if math.IsNaN(fv) || math.IsInf(fv, 0) {
```
wolffcm, line 312: Why do the check here at all? If you have to do the check in the dependency to provide a good diagnostic, and both `to` and `wide_to` will use that path, isn't that enough?

I see in your header comment that it should prevent the need to check in the dependency provider, but why is that bad?

jsternberg: The checks don't provide error messages or diagnostics. They're just to skip the point.

The reason for updating the dependency is so that we don't need to add similar checks to every dependency implementation. I suspect that influxdb has its own local implementation and I don't want to go check if that one also causes an issue. The http writer is the path for most to writes, but not all.

We could feasibly remove this check from the dependency, but I just want to protect against potential problems in other sections of code that might use the dependency. I don't want to see a situation where they get an error message saying, "You can't write NaN or Inf" if we make an error in the `to()` function because then we're giving them bad information rather than learning we have an error in our own code. So I thought it was better to just do the check twice to ensure that we both protect bad dependency code and bad `to()` implementation code.

wolffcm (approved):
Looks good, thanks for the explanation.

PR #4956, opened 2022-07-01
fix(stdlib): do not load some tests

nathanielc:
Some tests that should error have a different error message when pushed
down. This is fine and make sense. To keep things simple this change no
longer uses testing.load for those tests. Equivalent tests will be added
to InfluxDB directly to ensure we still get those different errors.

### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


onelson (approved):
Seems fine. Ideally we'd enhance `shouldError` such that we can indicate a portion of the string can be ignored, or otherwise uses a regex matcher. I don't think the line/column numbers are a great signal for anything in this situation.

Python doc tests use an ellipsis to say a portion of a pattern matches anything, but the remaining portions are used verbatim. Probably that's more than we want.

PR #4964, opened 2022-07-06
Contrib print

Anaisdg:
### Done checklist
- [ ] docs/SPEC.md updated
- [x] Test cases written


Anaisdg:
I was only surprised by how the display function formats JSON:
```
testcase printJSONTest {
    got =
        print.print(val:[{val:2.0}], result_name:"JSON")
    want =
        array.from(
            rows: [
                {"_value": "[{val: 2}]"}
            ]
        )

    testing.diff(got, want)
  ```
I would expect that `display(v: "[{val:2.0}]")` would produce ` "[{val:2.0}]"` instead of  "[{val: 2}]". 

Thank you.

PR #4968, opened 2022-07-07
feat: print function to visualize other data types

Anaisdg:
### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


stdlib/contrib/anaisdg/print/print.flux
```diff
@@ -0,0 +1,23 @@
+// Package print provides function for converting values into tables
+// ## Metadata
+// introduced: NEXT
```
sanderson, line 3:
```suggestion
// Package print provides functions for displaying Flux values.
//
// ## Metadata
// introduced: NEXT
```

stdlib/contrib/anaisdg/print/print.flux
```diff
@@ -0,0 +1,23 @@
...
+package print
+
+
+import "array"
+
+// print function converts other data types to a table
+// ## Parameters
+// - val: The value you want to "print".
+// - result_name: The name you want your table to have.
+// ## Examples
```
sanderson, line 13:
```suggestion
// print outputs Flux basic and composite data types in a table.
//
// ## Parameters
// - val: Value to print.
// - result_name: Result name. Default is `_result`.
//
// ## Examples
```

stdlib/contrib/anaisdg/print/print.flux
```diff
@@ -0,0 +1,23 @@
...
+//
+// ### Perform a linear regression on a dataset
+// ```no_run
+// import "contrib/anaisdg/print"
+// import "print"
+//
+// < value = (sampledata.float() |> findRecord(fn: (key) => true,idx: 0))._value
+// >  print.print(val: value, result_name: "extracted value")
+// ```
+print = (val, result_name) => array.from(rows: [{"_value": display(v: val)}]) |> yield(name: display(v: result_name))
```
sanderson, line 23: I think `result_name` should have a default value of `_result` just so users don't have to specify a result name every time.

```suggestion
print = (val, result_name="_result") =>
    array.from(rows: [{"_value": display(v: val)}]) |> yield(name: display(v: result_name))
```

sanderson: You could also set `val` to receive piped-forward data. That way you could just pipe forward an extracted value into the function.

stdlib/contrib/anaisdg/print/print.flux
```diff
@@ -0,0 +1,23 @@
...
+// ## Examples
+//
+// ### Perform a linear regression on a dataset
+// ```no_run
+// import "contrib/anaisdg/print"
+// import "print"
+//
+// < value = (sampledata.float() |> findRecord(fn: (key) => true,idx: 0))._value
+// >  print.print(val: value, result_name: "extracted value")
+// ```
```
sanderson, line 22: - Wrong title.
- To generate the example output, this example needs to run.
- You shouldn't have the 2nd import statement.
- The input shorthand `<` needs to be a stream of tables. This one isn't. I'd use the sample data as input.

~~~suggestion
// ### Print a value extracted from a stream of tables
// ```
// import "contrib/anaisdg/print"
//
// value = (sampledata.float() |> findRecord(fn: (key) => true,idx: 0))._value
//
// >  print.print(val: value)
// ```
~~~

I think you could show a basic example of printing a literal and then show an example of printing an extracted scalar value. For printing just a basic value:

~~~js
// ```
// import "contrib/anaisdg/print"
//
// > print.print(val: 123.4)
// ```
~~~

stdlib/contrib/anaisdg/print/print.flux
Anaisdg: Is there a way for it to receive both? I don't want to limit it to pipe forward in the case where people want to display variables.

sanderson: Absolutely. With this function definition:

```js
print = (val=<-, result_name="_result") =>
    array.from(rows: [{"_value": display(v: val)}]) |> yield(name: display(v: result_name))
```

Any of these will work:

```js
import "contrib/anaisdg/print"

123.4 |> print.print()
```

```js
import "contrib/anaisdg/print"

print.print(val: 123.4)
```

```js
import "contrib/anaisdg/print"
import "sampledata"

value = (sampledata.float() |> findRecord(fn: (key) => true,idx: 0))._value

value |> print.print()
```

```js
import "contrib/anaisdg/print"
import "sampledata"

value = (sampledata.float() |> findRecord(fn: (key) => true,idx: 0))._value

print.print(val: value)
```

stdlib/contrib/anaisdg/print/print.flux
```diff
@@ -0,0 +1,28 @@
...
+// ## Parameters
+// - val: Value to print.
+// - result_name: Result name. Default is `_result`.
+//
+// ## Examples
+//
+// ### Print a value extracted from a stream of tables
+// ```
+// import "contrib/anaisdg/print"
+// import "import "sampledata""
```
sanderson, line 21:
```suggestion
// import "sampledata"
```
```suggestion
// import "import "sampledata""
```

PR #4969, opened 2022-07-07
fix(stdlib): aggregate window mean and sum work properly with null values

### LINKED ISSUE #4945: Native aggregate window mean implementation should use mean/count instead of sum/count
The native aggregate window mean implementation accumulates sums and counts for intermediate state when accumulating values. This can potentially lead to sum values that are too large and may cause NaN or Inf values instead of the proper value. This gets worse with floating points accuracy at larger values.

There is a way to use means with counts to get the same result as sums with counts. Instead of `sum1 + sum2` and `count1 + count2`, you can do `(mean1 * count1 + mean2 * count2) / (count1 + count2)`.

For anyone who picks up this issue, double check the proper math on merging means.

At the same time as this, we need to validate the behavior of what happens if the input for aggregate window is all nulls which would cause the count to be zero. The correct behavior should be whatever `|> mean()` would do with the same input.

jsternberg:
The native aggregate window implementations of mean and sum now properly
work with null values and the mean implementation now uses the mean of
means method of calculation rather than just `sum / count`.

The null values ensures that it properly acts when there are null values
or when the interval is only null values.

The change to use a mean of means method of calculation helps to prevent
infinite values from showing up in the result when the sum overflows
because we keep a mean of each interval with the count so that we can
merge two means.

Fixes #4945.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/stdlib/universe/aggregate_window.gen.go
+++ b/stdlib/universe/aggregate_window.gen.go
@@ -21,13 +21,37 @@ func (a *aggregateWindowSumInt) Aggregate(ts *array.Int, vs array.Array, start, 
 	b := array.NewIntBuilder(mem)
 	b.Resize(stop.Len())
 
+	// Check once if we should look for nulls in the input.
+	hasNulls := vs.NullN() > 0
+
 	values := vs.(*array.Int)
 	aggregateWindows(ts, start, stop, func(i, j int) {
-		var sum int64
-		for ; i < j; i++ {
-			sum += values.Value(i)
+		var (
+			sum    int64
+			isNull = hasNulls
+		)
+		if hasNulls {
+			for ; i < j; i++ {
+				// If there are nulls, check if this is null.
+				if values.IsNull(i) {
+					continue
+				}
+				sum += values.Value(i)
+				isNull = false
+			}
+		} else {
+			// Skip the extra checks if we know there are no nulls.
+			for ; i < j; i++ {
+				sum += values.Value(i)
+			}
 		}
-		b.Append(sum)
+
+		// Append a null value if there were no valid points.
+		if isNull {
+			b.AppendNull()
+		} else {
+			b.Append(sum)
+		}
 	})
 	result := b.NewIntArray()
 	a.merge(start, stop, result, mem)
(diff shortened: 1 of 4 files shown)


stdlib/universe/aggregate_window.gen.go
```diff
@@ -21,13 +21,30 @@ func (a *aggregateWindowSumInt) Aggregate(ts *array.Int, vs array.Array, start,
...
 	values := vs.(*array.Int)
 	aggregateWindows(ts, start, stop, func(i, j int) {
-		var sum int64
+		var (
+			sum    int64
+			isNull = true
+		)
 		for ; i < j; i++ {
+			// If there are nulls, check if this is null.
+			if hasNulls && values.IsNull(i) {
```
Marwes, line 35:
```suggestion
			if values.IsNull(i) {
```
Checking `hasNulls` is just a redundant extra branch on each index (unless the optimizer does some magic?). I suppose `hasNulls` could be checked outside the loop though and branch on that.

```go
		if hasNulls {
			for ; i < j; i++ {
				// If there are nulls, check if this is null.
				if values.IsNull(i) {
					continue
				}
				sum += values.Value(i)
				isNull = false
			}
		} else {
			for ; i < j; i++ {
				sum += values.Value(i)
				isNull = false
			}
		}

```

stdlib/universe/aggregate_window_test.flux
```diff
@@ -235,69 +168,355 @@ testcase count_with_nulls {
             |> group(columns: ["_measurement", "_field", "t0"])
     got = do_test(every: 10s, fn: count)
 
-    testing.diff(got, want) |> yield()
+    testing.diff(got, want)
 }
-testcase min_with_nulls {
```
Marwes, line 240: There are a bunch of pre-existing tests changed here. What is the motivation for that? I would have expected that there isn't any change in behavior except for fixing the numeric overflow issue so the tests shouldn't have to change either, right?

Marwes (changes requested):

stdlib/universe/aggregate_window_test.flux
jsternberg: I changed the test case to use a time source of `_start` instead of `_stop` because then I could reuse the function for the new tests. This just meant I had to change the `_time` column to the start instead of the stop, but it didn't meaningfully change the tests themselves.

stdlib/universe/aggregate_window.gen.go
jsternberg: Changed in both sum and mean.

skartikey (approved):

jsternberg:
Original reviewer went on vacation so I'm going to close this PR and reopen it.

PR #4970, opened 2022-07-07
chore: add .editorconfig

onelson:
The motivation here was to avoid my editor from defaulting to space
indents in go templates.

### Done checklist
- [ ] Test cases written **N/A**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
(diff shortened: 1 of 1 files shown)


onelson:
Ah, for background: https://editorconfig.org/

It's just a highly generalized config that _many editors_ can use to tweak their behavior, mostly related to white space handling.

If you're using intellij like I am, you'll likely see a ribbon appear along the top of your buffer asking you to activate the bundled editorconfig plugin (installed by default), if it's not already active.

PR #4981, opened 2022-07-08
fix(plan): update pattern matching to specify successor counts

### LINKED ISSUE #4699: Planner rules sometimes do not fire when they could
Consider this test case
```js
testcase sort_limit_zero_row_table {
        input = array.from(rows: [{foo: "bar", _value: 10}])
                |> filter(fn: (r) => r._value > 10, onEmpty: "keep")
        want = input
        got = input |> sort() |> limit(n: 5)
        testing.diff(got, want)
}
```

The plan for this looks like so:
```mermaid
graph TD
  array.from0
  filter1
  sort2
  limit3
  diff4

  array.from0 --> filter1
  filter1 --> sort2
  sort2 --> limit3
  filter1 --> diff4
  limit3 --> diff4
```

There is no reason why we should not be able to merge the `sort |> limit` part of the graph into a `SortLimit` node. However, the planner has some checks that prevent this from happening. When we check pattern for the `SortLimit` rule, we start by matching in the `Limit` node, which passes, and then it recursively checks that the `Sort` part of the pattern also matches. However, this part fails to match because the predecessor of the `Sort` has more than one successor. That causes this check to fail:
https://github.com/influxdata/flux/blob/1c35d563c72e9cf79f60666fad8881ac41fda607/plan/pattern.go#L99-L104

DOD:
- update the planner such that we can transform a graph like this
- Add a test that shows it works
- Update the `sort_limit_zero_row_table` test case.

nathanielc:
The core of the problem is that the pattern matching logic
expects there to be only a single edge between a node and
its matched predecessor. This works and is required for rules
that intend to merge the matched nodes into a single node.
For example From -> Filter, if the from has other successors
you can't merge the filter into the from. In other cases this
restriction is not needed and prevents the rule for matching
when it is safe to do so. For example a join node whose predecessor
has a different successor, since the join rule doesn't rewrite its
predecessors it doesn't matter if they have other successors.

To solve this we need to be explicit about whether a pattern match
can have multiple successors or not. I propose we replace the existing
plan.Pat function with two functions plan.Multi and plan.Single.
plan.Multi will match a node and will allow that node to have multiple
successors, while the plan.Single will match a node and that node must
have exactly one successor. This changes two things about current patterns,
first it makes the number of successors expected explicit in the pattern,
second it moves the specification of which nodes can have multiple successors
and which cannot on the pattern that matches the node instead of the
successor of the node. Additionally plan.Any will allow multiple successors.

TODO:

I have added a test that does multiple joins and with this fix the error changes from an error caused by the equijoin rule not matching to an integrity check error. We need to fix the integrity issue before we can move forward with this change.

Fixes #4699

### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written


plan/logical_test.go
```diff
@@ -331,9 +331,8 @@ func (MergeFiltersRule) Name() string {
 }
 
 func (MergeFiltersRule) Pattern() plan.Pattern {
-	return plan.Pat(universe.FilterKind,
-		plan.Pat(universe.FilterKind,
-			plan.Any()))
+	return plan.Multi(universe.FilterKind,
+		plan.Single(universe.FilterKind))
```
wolffcm, line 335: The `Multi(x, Single(y, Single(z, ...)))` pattern is so common here that I wonder if we need users to specify it in their rules. It seems like maybe the planner could take care of how a pattern is evaluated and only allow multiple successors for the root (the most successorly) part of the pattern?

The pattern for a rule describes a contract between the write of the rule and the planner. Wherever they start their pattern, rule writers are saying, "I won't mutate any nodes that succeed this one". And likewise, wherever the pattern terminates, they are saying, "I won't mutate any predecessors to nodes that match these patterns". 

We could add a boolean parameter to `Match` that is `allowMultiSuccessors` and on the initial invocation by the planner it is `true`, and then on the recursive invocations it is always `false`. Seems like this would give us everything we want without having to change any client code or require the rule writer to think about it too much. What do you think? (hope this is clear)

wolffcm (commented):
Really like the direction of this, I had some additional thoughts, let me know what you think.

plan/logical_test.go
wolffcm: Thinking this over a bit more I think the approach in this PR makes sense. Rule writers should be aware of whether their patterns match multiple successors or not. I don't think there's a one-size-fits-all approach that would work.

I'm now wondering about the effect of these changes on rules that eliminate redundant nodes.
```
p = filter
p |> yield("1")
p |> filter(fn: (r) => true) |> yield("2")
```
A rule that removes the redundant filter will need to be succeeded by both yields. I don't think this would happen currently? I think the planner needs to be updated to look at the list of successors of the eliminated node and its predecessor and concatenate them. I think this is probably the error you already saw?

wolffcm: https://github.com/influxdata/flux/blob/53c90d5abbffa5c2c91c5e64e262d083ec4dff1a/plan/heuristic_planner.go#L136

wolffcm:
I have some pending changes to make this work thoroughly, will reopen soon.

Issue #4628, opened 2022-04-02
Add a planner rule that removes unnecessary `sort` nodes


scbrickley:
Blocked by #4678

In issue #4627, a new rule was added to the planner that ensures a physical `sort` node always precedes a `sortMergeJoin` node. It's possible, however, that the data to be passed into `join` is already sorted. We should add a planner rule that detects whether or not this is the case, and if so, removes the unnecessary sort node.

The logic should go as follows:
- Data that comes from storage is already guaranteed to be sorted by time (and possibly other columns)
- The planner should be capable of determining which nodes disrupt that sorting (see #4678)
- Therefore, if we look at all the nodes between the storage read and the `join` and find none that disrupt sorting, we can remove the previously inserted `sort` call.

It is not necessary to close this issue before work on the actual implementation of the `join` package (#4629) begins.

Some implementation notes:

- We need the ability to track attributes of plan nodes, some recent work re parallel nodes has created a foundation for this already
- We need these attributes to be parameterized (i.e. a list of sorted columns)
- The mergeSortJoin node requires that left and right predecessors are sorted in the same way according to the mapping of equi join columns. 

DOD:
- The planner rule described above exists and is used by the planner

wolffcm:
This work is currently in review in a couple PRs. I am changing the estimate to `1` to reflect this moving into the next sprint.

wolffcm:
This is all done and the feature flag for it has now been removed.

PR #4982, opened 2022-07-08
feat(plan): generalize attributes in planner

wolffcm:
This set of changes creates a framework whereby plan nodes can model characteristics of the data they consume and produce. These characteristics are called `attributes` in the code and currently there are three different kinds:
- `collation` to indicate data that is collated (sorted) by particular columns
- `parallel-run` to indicate that operations are run in parallel
- `parallel-merge` to indicate a plan node that represents a merging of parallel data streams

The parallel attributes were previously existing, but with this set of changes it is the procedure specs themselves that describe whether run in parallel or if they merge parallel streams.

Different attributes being produced or required by a particular operation is expressed by implementing interfaces. The comment at the top of `attributes.go` describes this in more detail.

This change set is mostly the same as the POC that was in https://github.com/influxdata/flux/pull/4902. There is a more comprehensive description over there, which is also valid for this PR. The main differences from this PR are changes that were required to allow the tests in `parallel_test.go` to pass.

This work is part of (but does not complete) #4628.

### DIFF
--- a/plan/attributes.go
+++ b/plan/attributes.go
@@ -2,34 +2,198 @@ package plan
 
 import (
 	"fmt"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/internal/errors"
 )
 
-// Physical attributes used in specifying parallelization. The Run attribute
-// means the node executes in parallel. It accepts parallel data (a subset of
-// the source) and produces parallel data. The merge attribute means that the
-// node accepts parallel data, merges the streams, and produces non-parallel
-// data that covers the entire data source.
-const ParallelRunKey = "parallel-run"
-const ParallelMergeKey = "parallel-merge"
+// Attributes provide a way to model different aspects of the data
+// flowing though the physical operations in a plan graph.
+//
+// For example, if a node requires its input tables to be sorted, it
+// will have the CollationAttr among its required attributes.
+// Likewise, if a node produces sorted tables, its output attributes
+// will have CollationAttr.
+//
+// Operations can require or provide attributes by implementing interfaces
+// in the corresponding PhysicalProcedureSpec:
+// - OutputAttributer is to be implemented if the procedure will provide an attribiute. E.g.,
+//   the SortProcedureSpec will provide CollationAttr for the columns on which the data is
+//   to be sorted.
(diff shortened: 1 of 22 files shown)


stdlib/universe/group.go
```diff
@@ -102,6 +102,14 @@ type GroupProcedureSpec struct {
 	GroupKeys []string
 }
 
+func (s *GroupProcedureSpec) PassThroughAttribute(attrKey string) bool {
+	switch attrKey {
+	case plan.ParallelRunKey:
+		return true
+	}
+	return false
+}
```
adrian-thurston, line 111: This is cool!

plan/attributes.go
```diff
@@ -2,34 +2,198 @@ package plan
...
+
+		if len(node.Successors()) == 0 {
+			return &flux.Error{
+				Code: codes.Internal,
+				Msg: fmt.Sprintf("node %q provides attribute %v that must be required but has no "+
+					"successors to require it", node.ID(), attr.Key()),
+			}
+		}
+		for _, succ := range node.Successors() {
+			if reqd, n := requiredBySuccessor(attr, node, succ.(*PhysicalPlanNode)); !reqd {
```
OfTheDelmer, line 152: When I've seen nesting of conditionals, esp. in a loop,  I have been trying to find ways to make them go away with `guards` or `continue`  like you've done above.


```suggestion
				reqd, n := requiredBySuccessor(attr, node, succ.(*PhysicalPlanNode));
				
				if reqd {
				   continue;
				}
				
 				if n != nil {
				   return &flux.Error{
					Code: codes.Internal,
					Msg: fmt.Sprintf("plan node %q has attribute %q that must be required by successors, "+
				        "but it is not required or propagated by successor %q",
				        node.ID(), attr.Key(), n.ID(),
				   ),
			        }
				return &flux.Error{
					Code: codes.Internal,
					Msg: fmt.Sprintf("plan node %q has attribute %q that must be required by successors, "+
						"but no successors require it",
						node.ID(), attr.Key(),
					),
				}	
```

OfTheDelmer (approved):

plan/attributes.go
wolffcm: Good idea, I've made the change.

PR #4986, opened 2022-07-11
fix(stdlib): aggregate window mean and sum work properly with null values

### LINKED ISSUE #4945: Native aggregate window mean implementation should use mean/count instead of sum/count

jsternberg:
The native aggregate window implementations of mean and sum now properly
work with null values and the mean implementation now uses the mean of
means method of calculation rather than just `sum / count`.

The null values ensures that it properly acts when there are null values
or when the interval is only null values.

The change to use a mean of means method of calculation helps to prevent
infinite values from showing up in the result when the sum overflows
because we keep a mean of each interval with the count so that we can
merge two means.

Fixes #4945.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


jsternberg:
Original review is at #4969.

skartikey (approved):

PR #4999, opened 2022-07-15
fix(join): provide a default schema for unmatched group keys

scbrickley:
Fixes a bug where join could not produce any output if a given group key existed on only one side of a join. Keeps track of a union of all schemas the join transformation has seen from each parent. If a group key is missing, that schema information is used in place of the schema tracked by the `joinState` so that compilation can finish successfully.

### DIFF
--- a/stdlib/join/join_test.flux
+++ b/stdlib/join/join_test.flux
@@ -282,3 +282,103 @@ testcase right_outer_join {
 
     testing.diff(want: want, got: got)
 }
+
+testcase exclusive_group_keys1 {
+    tbl1 =
+        array.from(rows: [{_time: 2022-06-01T00:00:00Z, _value: 1, label: "a", key: 1}])
+            |> group(columns: ["key"])
+    tbl2 =
+        array.from(rows: [{_time: 2022-06-01T00:00:00Z, _value: 0.1, id: "a", key: 2}])
+            |> group(columns: ["key"])
+
+    wantOutput =
+        "
+#datatype,string,long,string,long,double,dateTime:RFC3339,long
+#group,false,false,false,false,false,false,false
+#default,_result,,,,,,
+,result,table,label,v_left,v_right,_time,key
+,,0,a,1,,2022-06-01T00:00:00Z,1
+,,0,a,,0.1,2022-06-01T00:00:00Z,2
+"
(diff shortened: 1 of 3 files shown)


wolffcm (changes requested):
This looks great, but it definitely needs an acceptance test. If you can't finish one before vacation, it would make sense to hand this work off (I would be happy to take it)

scbrickley:
Ah, fair point. Let me see what I can do.

stdlib/join/join_test.flux
```diff
@@ -282,3 +282,103 @@ testcase right_outer_join {
...
+#group,false,false,false,false,false,false,true,true
+#default,_result,,,,,,,
+,result,table,label,v_left,v_right,_time,key,group
+,,0,a,1,,2022-06-01T00:00:00Z,1,one
+
+#datatype,string,long,string,long,double,dateTime:RFC3339,long,string
+#group,false,false,false,false,false,false,true,false
+#default,_result,,,,,,,
+,result,table,label,v_left,v_right,_time,key,group
+,,0,a,,0.1,2022-06-01T00:00:00Z,2,
```
wolffcm, line 358: Should `group` be in the group key here even though there is only nulls?

wolffcm (approved):
I had the one question but thinking it over, it could go either way...

Looks good, glad you fixed this!

stdlib/join/join_test.flux
scbrickley: That's a good question. Each output table currently has the same group key as its parent tables. I'm honestly not sure what the right answer is in this weird case.

scbrickley: Group key gets set [here](https://github.com/influxdata/flux/blob/4fb8524daf39bb21b80e3ec6e0a686631021e070/stdlib/join/join_fn.go#L156) for every output table, and gets set [here](https://github.com/influxdata/flux/blob/4fb8524daf39bb21b80e3ec6e0a686631021e070/stdlib/join/join_fn.go#L100) or [here](https://github.com/influxdata/flux/blob/4fb8524daf39bb21b80e3ec6e0a686631021e070/stdlib/join/join_fn.go#L111) if one side of a join key is empty.

PR #5001, opened 2022-07-18
fix(plan): modify plan.Pat to work with no predecessors for transformations

### LINKED ISSUE #4699: Planner rules sometimes do not fire when they could

jsternberg:
This modifies the pattern produced by `plan.Pat` so that it can work
when the topmost predecessor has predecessors but without requiring
those predecessors to be listed.

This allows plans like `sort |> limit` to work without using the `any`
node. The `any` node would cause the pattern to require the `any` node
to only have one successor, but it was perfectly allowable for that node
to have multiple successors. This makes writing the planner rule more
intuitive since we don't have to use `any` as the last predecessor and
it also prevents the problem of disallowing a pattern when it could have
been used.

Fixes #4699.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/plan/pattern.go
+++ b/plan/pattern.go
@@ -92,21 +92,25 @@ func (okp UnionKindPattern) Match(node Node) bool {
 		return false
 	}
 
-	if len(okp.predecessors) != len(node.Predecessors()) {
-		return false
-	}
-
-	// Check that each predecessor does not have other successors
-	for _, pred := range node.Predecessors() {
-		if len(pred.Successors()) != 1 {
+	// If we have specified predecessors, ensure that we have
+	// the proper number and that they don't have other successors.
+	if len(okp.predecessors) > 0 {
+		if len(okp.predecessors) != len(node.Predecessors()) {
 			return false
 		}
-	}
 
-	// Recursively match each predecessor
-	for i, pattern := range okp.predecessors {
-		if !pattern.Match(node.Predecessors()[i]) {
-			return false
+		// Check that each predecessor does not have other successors
+		for _, pred := range node.Predecessors() {
+			if len(pred.Successors()) != 1 {
+				return false
+			}
+		}
+
+		// Recursively match each predecessor
+		for i, pattern := range okp.predecessors {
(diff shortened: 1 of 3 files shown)


wolffcm (commented):
I have some thoughts on this, I'm not sure this will address the problem completely. I'llmake a comment on the original issue.

wolffcm (changes requested):
I've made a comment in the original issue:
https://github.com/influxdata/flux/issues/4699#issuecomment-1189229376

Can we add a test case for the scenario that I describe there and see how it fares?

nathanielc (changes requested):
I also have a case that this change does not solve. See my comment on the original issue. https://github.com/influxdata/flux/issues/4699#issuecomment-1189447063

PR #5005, opened 2022-07-19
feat(stdlib): promote various features to defaults

jsternberg:
This promotes the narrow state tracking and limit transformations to be
the defaults and removes their old implementations.

This also turns the group transformation group and optimized aggregate
window to default to true. I'm keeping the feature flags for these two
as these ones are more likely to require an unexpected problem and I
would like to keep a way for them to be turned off quickly. These two
will likely have their old versions removed next week.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


wolffcm (commented):
All of these have been enabled in production for a while, yes? How long?

wolffcm (approved):
Looks good, thanks!

PR #5006, opened 2022-07-19
chore(pagerduty): limit length of pagerduty summary and capture response body

glinton:
By capturing the response body, debugging becomes easier in the case that pagerduty rejects the alert. This also truncates the summary if it is too large for pagerduty to accept, preemptively handling a failure scenario.

I have had difficulty with this working in practice (manually updating the vendored flux library and deploying), but running manual queries in query explorer with the changes to the `pagerduty.flux` file result in working/expected results. I can redact and post that query here if needed.

### Done checklist
- [x] docs/SPEC.md updated (**N/A**)
- [x] Test cases written

### DIFF
--- a/stdlib/pagerduty/pagerduty.flux
+++ b/stdlib/pagerduty/pagerduty.flux
@@ -233,12 +233,7 @@ sendEvent = (
             else
                 json.encode(v: {data with payload: {payload with custom_details: customDetails}})
 
-        return requests.do(
-            method: "POST",
-            url: pagerdutyURL,
-            body: enc,
-            headers: headers,
-        )
+        return requests.do(method: "POST", url: pagerdutyURL, body: enc, headers: headers)
     }
 
 // endpoint returns a function that sends a message to PagerDuty that includes output data.
@@ -335,6 +330,9 @@ endpoint = (url=defaultURL) =>
                                 customDetails: record.get(r: obj, key: "customDetails", default: record.any),
                             )
 
-                        return {r with _sent: string(v: 2 == response.statusCode / 100), _status: string(v: response.statusCode), _body: string(v: response.body)}
+                        return {r with _sent: string(v: 2 == response.statusCode / 100),
+                            _status: string(v: response.statusCode),
+                            _body: string(v: response.body),
+                        }
                     },
                 )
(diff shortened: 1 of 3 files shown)


stdlib/pagerduty/pagerduty.flux
```diff
@@ -330,6 +330,9 @@ endpoint = (url=defaultURL) =>
                                 customDetails: record.get(r: obj, key: "customDetails", default: record.any),
                             )
 
-                        return {r with _sent: string(v: 2 == status / 100), _status: string(v: status)}
+                        return {r with _sent: string(v: 2 == response.statusCode / 100),
+                            _status: string(v: response.statusCode),
+                            _body: string(v: response.body),
+                        }
```
sanderson, line 336: This does change the pattern of all the existing `endpoint` functions. I'm not against it, but it likely needs to be considered a breaking change since it will affect downstream use cases.

glinton: Technically it's not a breaking change since it's additive (only adding the response body to `_body`). Pagerduty is already unique to other endpoints in that it includes the `_sent` field in addition to the apparent default of `_status`.

sanderson: Good point.

onelson (approved):
lgtm.

To @sanderson's comment re: breaking changes.
It makes sense to me that this would _not break_ callers except for very limited cases (like if the record was being written to an SQL db where the schema was already expected to be known and fixed). I wonder if the change warrants a doc update?

glinton:
Would it be beneficial to only store the body only on a non 2XX status? On success cases, the only new bit of data returned is a dedup key which may not be valuable to store, (although, the retention policy is small, so maybe it doesn't matter).
```
{'status':'success','message':'Event processed','dedup_key':'75b7b4e26c917c16fc83bebaf95afeaa48bbbd44fc1cc839cde1a1f94a0486a7'}
```

onelson:
> Would it be beneficial to only store the body only on a non 2XX status?

I think the returned record would need to be consistent regardless of the branching of the conditional, so the field would still need to be there for both cases. Are you thinking you'd assign it an empty string for the 2xx case?

I'd probably avoid special casing this and just give the truncated body for all cases.

PR #5007, opened 2022-07-20
feat(plan): add a rule to remove redundant sort nodes

wolffcm:
This PR is part of #4628.

Like the title says, this adds a planner rule (behind a feature flag) and associated tests that demonstrate sort nodes being removed.

I did a quick pass over our transformations to find the ones that preserve collation, and added `PassThroughAttributer` implementations as needed. For the case of `aggregateWindow` it always generates rows sorted by `_time` so I implemented `OutputAttirbuter` for that transformation as well.

### DIFF
--- a/plan/attributes.go
+++ b/plan/attributes.go
@@ -141,9 +141,15 @@ func getOutputAttributeWithNode(node Node, attrKey string) (PhysicalAttr, Node) 
 }
 
 // CheckSuccessorsMustRequire will return an error if the node has an output attribute
-// that must be required by a successor, and nil otherwise.. E.g., the parallel-run
-// attribute is like this in that it must be required by a merge node.
+// that must be required by *all* successors, but there exists some node that does not
+// require it.
+//
+// E.g., the parallel-run attribute is like this in that it must be required by a merge node.
 // This function will walk forward through successors to find the requiring node.
+//
+// The desired effect here is that if an attribute must be required by successors,
+// we walk forward through the graph and ensure that it is required on every branch that
+// succeeds the given node, with ohly pass-through nodes in between.
 func CheckSuccessorsMustRequire(node *PhysicalPlanNode) error {
 	for _, attr := range node.outputAttrs() {
 		if !attr.SuccessorsMustRequire() {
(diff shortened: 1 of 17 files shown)


plan/attributes.go
```diff
@@ -182,19 +195,24 @@ func CheckSuccessorsMustRequire(node *PhysicalPlanNode) error {
...
+	if psucc.passesThroughAttr(requiredAttr.Key()) {
 		if len(succ.Successors()) == 0 {
 			return false, nil
 		}
 		// If this node does not require the attribute itself but passes it along,
 		// see if any successors require it.
-		for _, ssucc := range succ.Successors() {
-			if reqd, n := requiredBySuccessor(requiredAttr, succ, ssucc.(*PhysicalPlanNode)); !reqd {
+		for _, ssucc := range psucc.Successors() {
+			if reqd, n := requiredBySuccessor(requiredAttr, succ, ssucc); !reqd {
```
OfTheDelmer, line 215: I am trying to grok this loop and recurse. 

At first it looked like we may be able to remove the loop over `p.Successors()`. However, I see can that being problematic if two successors are on different branches. It looks like we are checking for **any** successor to require the `requiredAttr` in a successor branch of passthrough zero or more successor nodes. It also looks like the comment above wants an **any** successor check.

It looks like this loop guard will enforce that **all** successor branches to require the `requiredAttr`. For this to work as described I think we need to return `true, n`  if `reqd` is true:

```golang
if reqd, n := requiredBySuccessor(requiredAttr, succ, ssucc); reqd {
     return true, n
}
```

Current behavior:

```mermaid
flowchart TD;
   A[ requiredBySuccessor for node and succ  where `requiredAttr = 'foo'`] --> B[succ.Successors => SsuccA,  SsuccB \n does not require and passesThrough 'foo' ];
   B ---> C[range over SsuccA, SsuccB];
   C --  check requiredBySuccessor --> D[SsuccA, does not require or passthrough 'foo'];
   D -- false, SsuccA --> C;
   C -- false, SsuccA --> A;
   C -- never checked--> E[SsuccB, does require 'foo'];
```

wolffcm: It is supposed to be an *all* check. If a node has an attribute that must be required by successors, there must be a node that requires it on every successor branch.

The only use case for "successor must require" is for plans that have parallelism, eg:
```
source |> aggregateWindow |> merge
```
In this case `source` and `aggregateWindow` have the `parallelRun` attribute and the `merge` node requires it. In effect, we process 4 (or whatever) parallel streams in `source` and `aggregateWindow` and then `merge` has to bring them together into a single stream that may be returned to the client or processed further.

If `aggregateWindow` had another successor besides the `merge` that did not require the `parallelRun` attribute, that would be an indication that something went wrong with planning, because we need parallel streams to be merged into one at some point before data is returned to the client. In effect this guards from creating incorrect plans. Does that make sense?

I can definitely update the comment to be more clear.

OfTheDelmer (approved):

PR #5008, opened 2022-07-20
fix: passing epsilon value from go test to test framework

skartikey:
Some test flakiness has been reported for the holt winter test.
```
--- FAIL: TestHoltWinters_Process (35.48s)
    --- FAIL: TestHoltWinters_Process/NOAA_water_-_with_nulls_and_missing_-_no_seasonal (17.15s)
        holt_winters_test.go:903: unexpected tables -want/+got
              []*executetest.Table{
                &{
                        ... // 2 identical fields
                        KeyValues: nil,
                        ColMeta:   {{Label: "_time", Type: s"time"}, {Label: "_value", Type: s"float"}, {Label: "minSSE", Type: s"float"}},
                        Data: [][]any{
                                {
                                        s"2015-08-27T22:13:00.000000000Z",
            -                           float64(6.517746679116747),
            +                           float64(6.517761275336982),
            -                           float64(30.52018686151099),
            +                           float64(30.520187390109854),
                                },
```

In the testing framework, we have epsilon value defaults to [1e-6](https://github.com/influxdata/flux/blob/89d78e05caee30f2ff09cd6b23bf6162c65f324d/execute/executetest/transformation.go#L25) for float operations. There is no way to change it just for one Go test. So this change is to allow Go test to pass their own value of epsilon(with default value 1e-6). 

For the holt winter, epsilon of 1e-3 should suffice, as we do not need a more precise comparison.

### DIFF
--- a/execute/executetest/transformation.go
+++ b/execute/executetest/transformation.go
@@ -31,7 +31,7 @@ const ulp uint = 2
 // Comparison options for floating point values.
 // NaNs are considered equal, and float64s must
 // be sufficiently close to be considered equal.
-var floatOptions = cmp.Options{
+var defaultFloatOptions = cmp.Options{
 	cmpopts.EquateNaNs(),
 	cmp.FilterValues(func(x, y float64) bool {
 		return !math.IsNaN(x) && !math.IsNaN(y)
@@ -50,12 +50,15 @@ var floatOptions = cmp.Options{
 	})),
 }
 
+// floatOptions is a list of comparison options for floating point values.
+// if not passed by the caller, defaultFloatOptions will be used
 func ProcessTestHelper(
 	t *testing.T,
 	data []flux.Table,
 	want []*Table,
 	wantErr error,
 	create func(d execute.Dataset, c execute.TableBuilderCache) execute.Transformation,
+	floatOptions ...cmp.Option,
 ) {
 	t.Helper()
 
@@ -109,17 +112,26 @@ func ProcessTestHelper(
 	sort.Sort(SortedTables(got))
 	sort.Sort(SortedTables(want))
 
-	if !cmp.Equal(want, got, floatOptions) {
+	opts := make([]cmp.Option, 0, len(defaultFloatOptions)+len(floatOptions))
+	if len(floatOptions) > 0 {
+		opts = append(opts, floatOptions...)
+	} else {
+		opts = append(opts, defaultFloatOptions...)
+	}
+	if !cmp.Equal(want, got, opts...) {
 		t.Errorf("unexpected tables -want/+got\n%s", cmp.Diff(want, got))
 	}
 }
 
+// floatOptions is a list of comparison options for floating point values.
+// if not passed by the caller, defaultFloatOptions will be used
 func ProcessTestHelper2(
 	t *testing.T,
 	data []flux.Table,
 	want []*Table,
 	wantErr error,
 	create func(id execute.DatasetID, alloc memory.Allocator) (execute.Transformation, execute.Dataset),
+	floatOptions ...cmp.Option,
 ) {
 	t.Helper()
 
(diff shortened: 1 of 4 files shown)


execute/executetest/transformation.go
```diff
@@ -31,30 +31,36 @@ const ulp uint = 2
...
 }
 
 func ProcessTestHelper(
 	t *testing.T,
 	data []flux.Table,
 	want []*Table,
 	wantErr error,
+	epsilon float64,
 	create func(d execute.Dataset, c execute.TableBuilderCache) execute.Transformation,
 ) {
```
wolffcm, line 65: One approach I've seen taken in situations like this one is to change the signature:
```suggestion
	create func(d execute.Dataset, c execute.TableBuilderCache) execute.Transformation,
	...cmp.Option,
) {
```
And then pass any additional options at just the callsites where they're needed. This way, none of the unaffected tests need to change, and by passing a `cmp.Option` it can have a name that is clear rather than a `0` which can be opaque.

What do you think?

execute/executetest/transformation.go
```diff
@@ -109,7 +110,13 @@ func ProcessTestHelper(
...
 	sort.Sort(SortedTables(want))
 
-	if !cmp.Equal(want, got, floatOptions) {
+	opts := make([]cmp.Option, 0, len(floatOptions)+len(options))
+	if len(options) > 0 {
+		opts = append(opts, options...)
+	} else {
+		opts = append(opts, floatOptions...)
+	}
+	if !cmp.Equal(want, got, opts...) {
```
wolffcm, line 119: So this will replace the old options there were there... Can you call this out in a comment on `ProcessTestHelper`? Maybe the parameter should be named `alternateFloatOptions` to clarify that it will take precedence over the default behavior.

stdlib/universe/holt_winters_test.go
```diff
@@ -19,6 +21,12 @@ import (
 	"github.com/influxdata/flux/stdlib/universe"
 )
 
+const epsilon float64 = 1e-03
```
wolffcm, line 24: This seems really high. I wonder if the algorithm could be improved to be more stable... would it make sense to file an issue?

wolffcm (approved):
Looks good to me, I just had some comments on how to clarify the behavior of the new parameter.

stdlib/universe/holt_winters_test.go
skartikey: Looking at Jonathan's [comment](https://influxdata.slack.com/archives/C0AD7FFS9/p1658242941737489?thread_ts=1658174400.602789&cid=C0AD7FFS9), it seems le-5 would work. So I will change it to that.

wolffcm: It should be set to as small as we can and still have the test pass. `1e-05` still seems high to me to be honest, unless there is a huge tradeoff in terms of performance to get more stable answers.

execute/executetest/transformation.go
skartikey: what if we rename [floatOptions](https://github.com/influxdata/flux/blob/92b442939403cb3e159a42bc17f4e668d735af32/execute/executetest/transformation.go#L34) to defaultFloatOptions, that way it would be clear if nothing passed in the option, defaultFloatOptions will be used.

wolffcm: Sure, that works. Seems like the parameter should be named `floatOptions` in that case. Since that's all it's really used for?

stdlib/universe/holt_winters_test.go
skartikey: my mistake, what I meant was `1e-5`. we have epsilon [1e-6](https://github.com/influxdata/flux/blob/92b442939403cb3e159a42bc17f4e668d735af32/execute/executetest/transformation.go#L25) in the testing framework at the moment.

execute/executetest/transformation.go
skartikey: Yes, I will change the param name. Thanks

stdlib/universe/holt_winters_test.go
wolffcm: Ah, the epsilon is already bigger than I thought it would be. I wonder why it's so high? In any case you are only moving it by one decimal place so that seems fine.

PR #5009, opened 2022-07-20
fix(stdlib): ignore unknown messages rather than erroring

jsternberg:
The process message method is intended to ignore messages it does not
understand rather than actively return an error. This removes the error
from join for an invalid message.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written

### DIFF
--- a/execute/transport.go
+++ b/execute/transport.go
@@ -24,6 +24,11 @@ import (
 // Transport is an interface for handling raw messages.
 type Transport interface {
 	// ProcessMessage will process a message in the Transport.
+	//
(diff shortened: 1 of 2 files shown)


onelson (approved):
lgtm.

I wonder if there should be a comment added to the source to indicate the proper usage/intent. On its face, the change looks like swallowing an error and I'd naturally look for ways to avoid doing so unless there was something to indicate I should.

jsternberg:
I updated the docs for `execute.Transport` to call out this usage explicitly.

PR #5011, opened 2022-07-21
docs: fix missing "is"

jstirnaman:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

### DIFF
--- a/stdlib/http/requests/requests.flux
+++ b/stdlib/http/requests/requests.flux
@@ -150,7 +150,7 @@ do = (
         config: config,
     )
 
-// post makes a http POST request. This identical to calling `request.do(method: "POST", ...)`.
+// post makes a http POST request. This is identical to calling `request.do(method: "POST", ...)`.
 //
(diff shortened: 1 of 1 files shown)


stdlib/http/requests/requests.flux
```diff
@@ -180,7 +180,7 @@ do = (
 //
 // data = ejson.parse(data: response.body)
 //
-// > array.from(rows: [data])
+// > array.from(rows: [data]
```
sanderson, line 183: This right parenthesis shouldn't have been removed.

```suggestion
// > array.from(rows: [data])
```

sanderson (changes requested):
- One suggestion.
- Once the changes are complete, you'll need to regenerate the stdlib to get the tests to pass. I'm all set up to do it pretty quickly, so I can do it for you.
- Update the PR title to start with `chore(docs): ...` so it won't show up in the changelog.

PR #5015, opened 2022-07-22
feat(stdlib):  add vector float builtin

### LINKED ISSUE #4736: Implement vectorized float builtin
In support of #4603, write the new builtin (**giving it a leading underscore in the name**).

Where possible, try to leverage Arrow APIs for the casting of values since it may provide a more optimized path.

Note that there's some debt associated with getting access to the allocator (currently it's attached to the `Context`). If you have the appetite, think about refactoring such that we can have a more type-safe way to access this resource where needed.

DOD:
- a new builtin is defined in `universe` that accepts a vector of `T` and produces a vector of `float`
- (optionally) refactor code providing access to allocator

OfTheDelmer:
Closes #4736

This draft adds a built in to convert vectors to float vectors. It also includes handling for converting repeat vectors. There were also some changes to the vector test helpers to allow null value testing in conversions.


onelson (commented):
This is looking pretty good to me. What else is missing?

onelson:
Heads up: I had to borrow your pending changes for the supporting test code in my PR. I made some light tweaks since I was seeing panics elsewhere in semantictest with your current changes:
https://github.com/influxdata/flux/pull/5017/commits/b6a7a1f55f0deda4b99a2e20382f229891139bfc

onelson:
https://github.com/influxdata/flux/pull/5048 follows on from this one, but the sense is this PR shouldn't merge until https://github.com/influxdata/flux/pull/5048 is close to being ready. The rationale being there might be some small tweaks required on this side when the built-in is actually used by the rest of the runtime (which isn't possible until we have both PRs closer).

array/array.go
```diff
@@ -206,3 +209,74 @@ func Slice(arr Array, i, j int) Array {
 	err := errors.Newf(codes.Internal, "cannot slice array of type %T", arr)
 	panic(err)
 }
+
+func ToFloatConv(mem memory.Allocator, arr Array) (*Float, error) {
```
skartikey, line 213: This method can go in `typeconv.go` itself. Let me know if I have misunderstood something.

skartikey (approved):
Look good!

array/array.go
onelson: The thinking was "things that work with arrays directly should likely go in the array package." This sort of puts most of the underlying vectorization work here in one area of the source tree.

Let me know if you think it should really move -- we can try and shift it around if need be.

skartikey: That makes sense.

Issue #4739, opened 2022-05-11
Add a way to assert about vectorized operations


onelson:
Currently we don't have a way to write an end to end (flux acceptance) test which can make assertions about whether or not something in the query was actually vectorized.

Not exactly sure what the good solution for this is, but the closest thing I can think of available today is the API we have for asserting about pushdowns in the plan.

> related, but not necessarily on the critical path for this, but it would be interesting to be able to see information about vectorized operations from the profiler.

NOTE: We think the best way to assert about verctorized operations is to leverage the planner rules. This means we want to create/split the map transformation into two, one for row based and one for vectorized operations. Then the planner rule chooses between them when possible. This way enabling/disabling the planner rule controls vectorization. Additionally we can assert about the application of planner rules in tests already.

DOD:

- Remove old (i.e. no chunked implementation of map) this code path is currently dead in production because of the feature flag
- Split new (i.e. chunked implementation) of map transformation into two transformations, row and vectorized version.
- Create planner rule that picks the vectorized map transformation when the vectorized code is available
- Add/Update acceptance tests for each vector operation (i.e. logical, comparison, etc) that show correctness of the vectorized version asserting the planner rule was applied.

PR #5017, opened 2022-07-23
feat: support conditional expressions in vectorized map

### LINKED ISSUE #4601: Vectorize conditional expressions `if/then/else`
Not sure what the concerning details are around this, but `if/then/else` was flagged as having some complications for vectorization.
The current thinking is it could be best to deliver #4599 before beginning work on this.

Notes:

Generally the conditional, consequent and alternate will need to be evaluated for all rows in the vector. However if the conditional is either all true or all false we can evaluate only one of the consequent or alternate. But if even one value of the conditional is different then we need to evaluate both the consequent and alternate for all rows and use the conditional vector to mask the values. 

DOD:
- provide a vectorized implementation for conditional expressions `if/then/else`
- guard the new implementation with a feature flag

onelson:
Fixes #4601

Adds support for vectorized functions containing conditional expressions.
This new code path is guarded by a feature flag called `vectorizedConditionals`, enabled in test but off by default.

In cases where the `test` (ie the condition part of the expression) is all one value or another, we can skip a lot of the new code by returning either the the `consequent` or `alternate` vector directly.

For other situations, both the `consequent` and `alternate` vectors must be evaluated in their entirety after which values are selected for return based on the `test` value for the given index.
**N.b.** this _could give an inconsistent outcome_ for cases where these expressions have some sort of side-effect when the when compared to the non-vectorized version. At this time, such expressions would prevent the function from being vectorized in the first place, but we may need to revisit this detail in the future.

The flux acceptance tests pass with or without the flag enabled, but in the future when #4739 lands, these testcases should be updated to assert the vectorized code path is actually being hit.

Go unit tests verify the behavior in a more isolated way.

### Done checklist
- [x] Test cases written


compiler/vectorized_test.go
```diff
@@ -461,6 +461,8 @@ func TestVectorizedFns(t *testing.T) {
 				if fn.Vectorized == nil {
 					t.Fatal("Expected to find vectorized node, but found none")
 				}
+				// XXX: test the vectorized version instead of the row-based one.
```
OfTheDelmer, line 464: Great catch!

OfTheDelmer (commented):
It looks great!  Just adding comments from our dive into `IsValid` vs `IsNull`.

onelson:
@OfTheDelmer thanks for your notes. I'll do another pass on the null handling, hopefully with some new tests to verify that:

- when `test` is `null` it is considered `false`
- when either `consequent` or `alternate` are `null`
  - ... and they are the selected branch
  - we produce a null in the output vector

I think the blocks called out in the review may simplify further with this contract in mind.

onelson:
Tests added in https://github.com/influxdata/flux/pull/5017/commits/2afaa998513f9a5c9f3a8596370d24054942ac0c for null inputs in each of the three vectors appear to be passing as-is, so that's something.

Working to simplify the checks now.

onelson:
@OfTheDelmer I've simplified the null checks in https://github.com/influxdata/flux/pull/5017/commits/65ac9608784ba6852ddcd3ef65c39dda9d6b1872 -- if these changes seem right to you, can you resolve the remarks in your review?

array/conditional.gen.go.tmpl
```diff
@@ -0,0 +1,105 @@
...
+		b.Resize(n)
+
+		if !(c.Len() == n && a.Len() == n) {
+			return nil, errors.Newf(codes.Invalid, "vectors must be equal length")
+		}
+
+		for i := 0; i < n; i++ {
+			// nulls are considered as false
+			truthy := t.IsValid(i) && t.Value(i)
+			if truthy && c.IsValid(i) {
```
OfTheDelmer, line 22: Nice 🚀

values/vector_values.gen.go
```diff
@@ -81,47 +87,71 @@ func newVectorFromSlice(values []Value, typ semantic.MonoType, mem memory.Alloca
 	case semantic.BasicInt:
 		b := arrow.NewIntBuilder(mem)
 		for _, v := range values {
-			b.Append(v.Int())
+			if v.IsNull() {
```
OfTheDelmer, line 90: Awesome!

OfTheDelmer (approved):
This update looks great

PR #5019, opened 2022-07-25
Fix boundaries/week returning the previous week

### LINKED ISSUE #5018: boundaries/week gets last week when queried on first day of week
This testcase:

```flux
> import "date/boundaries"
> option now = () => 2022-07-25T09:50:00.0Z
> boundaries.week()
{start: 2022-07-18T00:00:00.000000000Z, stop: 2022-07-25T00:00:00.000000000Z}
```

gives me (IMHO) the wrong week.

the `now` date is a monday, so my prediction would be, that `boundaries.week()` gives me the current week, starting at `2022-07-25T00:00:00`

The same applies to a sunday query:

```flux
> import "date/boundaries"
> option now = () => 2022-07-24T09:50:00.0Z
> boundaries.week(start_sunday: true)
{start: 2022-07-17T00:00:00.000000000Z, stop: 2022-07-24T00:00:00.000000000Z}
```

Used version for this test: https://github.com/influxdata/flux/releases/tag/v0.175.0

hellow554:
### Done checklist
- [ ] docs/SPEC.md updated
- [X] Test cases written

Fixes #5018

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ test-rust:
 	$(CARGO) test --doc && \
 	$(CARGO) clippy $(CARGO_ARGS) --all-features -- -Dclippy::all -Dclippy::undocumented_unsafe_blocks
 
-test-flux:
+test-flux: libflux-go
 	$(GO_RUN) ./cmd/flux test -p stdlib -v --parallel 8
 
 test-flux-integration:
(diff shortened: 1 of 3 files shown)


scbrickley:
Thanks for the patch @hellow554!

I've discussed this issue with some of my team members, and we all agree that the package as a whole needs some more design work. The spec for this package was not as clear as it could have been, and this unexpected behavior is the result. I've filed an issue to track the work here: https://github.com/influxdata/flux/issues/5030

Again, thanks for bringing this to our attention. I'm going to leave this PR open in case this fix becomes handy in the near future.

scbrickley:
Closing because this was fixed in https://github.com/influxdata/flux/pull/5031

PR #5028, opened 2022-07-27
refactor: move lesser used packages to internal and remove unused packages

jsternberg:
This moves some lesser used packages, hence less tested, to be internal
packages. This prevents them from being used as part of the Go API and avoids
creating a contract that the API method is formally supported in its exact
signature. Since these packages are lesser used, that may mean that the API is
less than ideal or may require fixing in the future and we don't want to have
to commit to a public facing API for something we're not sure about.

This also removes some packages that are unused to both reduce the amount of
code and reduce confusion about whether it is formally supported. At the same
time, it also prevents them from being a formal part of the Go API.

### Done checklist
- [x] docs/SPEC.md updated
- [x] Test cases written


wolffcm (approved):
Looks good.

Issue #5030, opened 2022-07-27
boundaries package needs some redesign


scbrickley:
A user pointed out that the `boundaries.week()` function produces an unexpected answer: https://github.com/influxdata/flux/issues/5018

It's clear that we need to rethink the API here. We should find a way to move the boundaries package to `experimental` so that we can work out some changes to the package without the concern of breaking user scripts.

PR #5031, opened 2022-07-27
chore(stdlib): Move boundaries package to experimental

scbrickley:
Related to #5030 

This will break user queries, but the boundaries package is relatively young, so it may not be that much of a problem. I'll look through the query archiver to assess the blast radius before merging.


wolffcm (approved):
This looks good.

I know we can test for error messages in acceptance tests now... might be worth trying that out for this? I'll leave it up to you.

hellow554:
As already suggested in #5019 I would like to see, that `boundaries/week` returns the current week, if the current day is the start day of the week. Maybe it can be done in this PR?

wolffcm (approved):
This still looks good to me.

wolffcm:
@sanderson This seems like a PR that you should be aware of. For context, the `boundaries` package landed outside of `experimental/` a few weeks ago, but consensus is that we'd like to continue to iterate on it, so we're moving it back there, pending understanding of how widely it's used now.

wolffcm (approved):

PR #5035, opened 2022-07-28
fix(join): Guard message processing with mutex

scbrickley:
Previously, join was only guarding the `processChunk` method with a locked mutex. Recent changes to flux have revealed that that was not enough. This patch ensures that the handlers for the `FlushKey` and `Finish` messages are also behind a mutex.


wolffcm (approved):
Looks good.

Issue #4877, opened 2022-06-14
fluxtest test case names may conflict


wolffcm:
There are currently three tests that have the name `group`:
```
$ find . -name '*.flux' | xargs grep 'testcase group '
./stdlib/experimental/group_test.flux:testcase group {
./stdlib/universe/group_nulls_test.flux:testcase group {
./stdlib/universe/group_test.flux:testcase group {
```

I'm not sure what this means for `fluxtest`, which has relevant flags:
- `--test` will run a specific test
- `--skip` will skip a specific test

I guess if you did `--test group` it might run all three, or just one.

The usage of `--skip` is particularly problematic because we may end up skipping tests that we don't intend to.

We should consider one or both of:
- Adding a namespaces to how one specifies the test case, like package and/or the file that contains the test
- Making it an error to have duplicate test case names (perhaps within some specific namespace)

DOD:

- Conflicting test names produce an error
- Any existing conflict are fixed
- Test skip lists are updated with the new scoped names

onelson:
I think it makes a lot of sense to allow the package name to act as a stem for testcase names. I'd lean towards making duplicate testcase names within a common package an error.

skartikey:
I did a quick test.

`Without any flags -`
```
go run ./cmd/flux test
ssssss.........................................................................................................................................................................................................s..............................................................................................................................................................sssssssssssssssssssssssssss...................................................................................................................................................................................................................................s..........................................................................................................................s..................................s.
---
Found 780 tests: passed 743, failed 0, skipped 37
```

With `--skip group`, skipped all 3 `group` tests.
```
go run ./cmd/flux test --skip group
ssssss............................................................................................................................................................................................s............s..............................................................................................................................................................sssssssssssssssssssssssssss.............................................................................................................................................................................................ss....................................s..........................................................................................................................s..................................s.
---
Found 780 tests: passed 740, failed 0, skipped 40
```

With `--test group`, ran all 3 `group` tests.
```
go run ./cmd/flux test --test group  
[...]

PR #5039, opened 2022-08-02
feat(fluxtest): enhance fluxtest to use package name with test/skip flags

skartikey:
The package name can be added to the testcase name as a qualifier
Both `package.name` and `package/name` are supported.

Here is an example - 
There are 2 testcase exist with name group
```
$find . -name '*.flux' | xargs grep 'testcase group '
./stdlib/experimental/group_test.flux:testcase group {
./stdlib/universe/group_test.flux:testcase group {
```
// testcase name without qualifier will run both the test. 
<pre>go run ./cmd/flux test --test <b>group</b></pre> 
// testcase name with qualifier will run the specific test
<pre>go run ./cmd/flux test --test <b>universe.group</b>
go run ./cmd/flux test --test <b>experimental.group</b></pre>

fixes: https://github.com/influxdata/flux/issues/4877

### Checklist


cmd/flux/cmd/test.go
```diff
@@ -375,7 +412,13 @@ func (t *TestRunner) MarkSkipped(testNames, skips, tags []string, skipUntagged b
...
 			}
 		}
-		t.tests[i].skip = skip || skipMap[t.tests[i].Name()] || (skipUntagged && len(t.tests[i].tags) == 0)
+		// skip tests
+		skipTest := false
+		if !skip && len(skips) > 0 {
+			skipTest = containsWithPkgName(skips, t.tests[i])
+		}
+
+		t.tests[i].skip = skip || skipTest || (skipUntagged && len(t.tests[i].tags) == 0)
```
wolffcm, line 421: This logic is getting harder to follow, now. There is a boolean called `skip` and another called `skipTest`. Can they both be renamed to reflect what they are? Like `skipBecauseTags` and `skipBecauseSkipList`.

wolffcm (changes requested):
I had some comments but this looks good for the most part. I had a couple of comments that it would be good to address.

Let's make sure that downstream repos, which have their own skip lists, continue to work right.

cmd/flux/cmd/test.go
```diff
@@ -234,6 +241,33 @@ func contains(names []string, name string) bool {
...
 }
 
+func containsWithPkgName(names []string, test *Test) bool {
+	var aTest, tTestName string
+	for _, name := range names {
+		fTest := splitAny(name, "./")
+		// handle package.TestName or package/TestName case
+		if len(fTest) > 1 {
+			aTest = strings.Join(fTest, "")
+			tTestName = test.PackageName() + test.Name()
```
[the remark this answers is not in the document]
skartikey: Good catch! Thanks for the example. I will do the changes.

cmd/flux/cmd/test.go
```diff
@@ -300,20 +339,26 @@ func (t *TestRunner) Gather(roots []string) error {
...
 				if err != nil {
 					return err
 				}
 				if invalid := invalidTags(tags, mods.Tags(file.module)); len(invalid) != 0 {
-					return errors.Newf(codes.Invalid, "testcase %q, contains invalid tags %v, valid tags are: %v", tcnames[i], invalid, mods.Tags(file.module))
+					return errors.Newf(codes.Invalid, "testcase %q, contains invalid tags %v, valid tags are: %v", tcidens[i].Name, invalid, mods.Tags(file.module))
+				}
+				pkgTest := pkg + "." + tcidens[i].Name
+				if _, ok := seen[pkgTest]; ok {
+					return errors.Newf(codes.AlreadyExists, "duplicate testcase name %q, found in package %q, at locations %v and %v", tcidens[i].Name, pkg, seen[pkgTest].loc.String(), tcidens[i].Loc.String())
```
wolffcm, line 357: Can we have a test for this? Good use of the `AlreadyExists` code :)

wolffcm (changes requested):
You are definitely close on this, but it would be good to have test case for the error that is produced when two cases have the same name in the same package. That will keep us from having any ambiguity moving forward, to it would be good to lock it down.

skartikey:
After merging https://github.com/influxdata/flux/pull/5057 Did a manual test by duplicating test case name in the same package and got below error - 

`duplicate testcase name "group", found in package "universe", at locations stdlib/universe/group_test.flux|38:10-38:15 and stdlib/universe/group_ungroup_test.flux|48:10-48:15`

cmd/flux/cmd/test.go
skartikey: Added a test for duplicate testname case

cmd/flux/test_test.go
```diff
@@ -349,11 +380,45 @@ func Test_TestCmd_Skip(t *testing.T) {
...
+	for name, got := range got {
+		if want != got {
+			t.Errorf("%s: unexpected summary got %+v want %+v", name, got, want)
+		}
+	}
+}
+
+func Test_TestCmd_Error_Duplicate(t *testing.T) {
+	wantErr := errors.New("duplicate testcase name \"duplicate\", found in package \"test\", at locations testdataduplicate/test_test.flux|7:10-7:19 and testdataduplicate/test_test.flux|14:10-14:19")
+	runForPath(t, "./testdataduplicate", wantErr, "--test", "duplicate")
```
wolffcm, line 413: Very nice!

wolffcm (approved):

PR #5041, opened 2022-08-02
feat: add rust binary to sit on top of "headless repl" backend

Westonside:
## Description

Added in my rusty line code from my other repo. The rustyline will send the user input to the headless repl over JSON RPC via the stdin and receive output from the stdout. The rustyline also provides completion suggestions through communication over the stdio with the flux lsp. Related to #5040.

### Checklist


Marwes (commented):
Mostly just style things. Haven't gone through all of it yet so will continue tomorrow.

internal/cmd/headless_repl/rust-toolchain.toml
```diff
@@ -0,0 +1,11 @@
+[toolchain]
+channel = "1.60"
```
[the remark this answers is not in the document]
onelson: Note that, per @Westonside, there's some dependency that depends on features in the later toolchain. We can probably pin to an earlier version of whatever it is, to avoid needing the newer compiler.

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+impl From<serde_json::Error> for OutputError {
+    fn from(_: serde_json::Error) -> Self {
+        InvalidMethod
+    }
+}
+
+pub fn form_output(request_type: &str, text: &str) -> Result<String, OutputError> {
+    match request_type {
+        "Service.DidOutput" => {
+            let mut cleaned = text.to_string().replace("\"", "\\\"");
```
Marwes, line 52:
```suggestion
            let mut cleaned = text.replace("\"", "\\\"");
```
`replace` exists on `&str`, no need to heap allocated a `String` https://doc.rust-lang.org/std/primitive.str.html#method.replace

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+            let mut param = (r#"[{"input": "#).to_string();
```
Marwes, line 53:
```suggestion
            let mut param = r#"[{"input": "#.to_string();
```

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+    let num = Regex::new(r"\d").unwrap();
+    let mut buf: Vec<u8> = vec![];
+    let mut num_buf: Vec<u8> = vec![];
+    let mut x = 0;
+    let mut y = 0;
+    //indicate when to start and stop capturing numbers in the content length
+    let mut num_cap = false;
+    let mut read_exact = (false, 0);
+    for i in child_stdout.bytes() {
+        let val = i.unwrap();
```
Marwes, line 82: This will panic if there are IO errors, better to return a `Result` from this function and use the `?` operator to propagate it up.

You can use https://docs.rs/anyhow/1.0.61/anyhow/ or `Box<dyn std::error::error + Send + Sync>` as the error type if you do not want to mess with specific ones.

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+        let single = [val];
+        if read_exact.0 {
+            buf.insert(buf.len(), val);
+            read_exact.1 = read_exact.1 - 1;
+            if read_exact.1 == 0 {
+                //final result
+                let resp = str::from_utf8(&buf).unwrap();
```
Marwes, line 89: Error handling with `?` instead of panicking

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+                    let mut write_lock = storage.write().unwrap();
+                    *write_lock = val;
+                }
+                read_exact.0 = false;
+                // break;
+            }
+            continue;
+        }
+
+        let a = str::from_utf8(&single).unwrap();
```
Marwes, line 102: Ditto

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+        //if capturing numbers and the value is numeric add to number buffer
+        if num_cap && num.is_match(a) {
+            num_buf.insert(num_buf.len(), val);
+        } else {
+            if num_cap {
+                //indicate you need to take that number and read that many bytes
+                num_cap = false;
+                buf.clear();
+                let read = str::from_utf8(&num_buf).unwrap();
```
Marwes, line 111: Ditto

internal/cmd/headless_repl/src/processes/invoke_go.rs
```diff
@@ -0,0 +1,130 @@
...
+
+            // println!("{}", res);
+            Ok(res)
+        }
+        _ => Err(InvalidMethod),
+    }
+}
+
+pub fn read_json_rpc(child_stdout: ChildStdout, mut storage: Arc<RwLock<HashSet<CommandHint>>>) {
+    let re = Regex::new(r"Content-Length: ").unwrap();
```
Marwes, line 72: You don't actually need to do all this parsing, the RPC format is basically HTTP headers followed by a HTTP body. So either pull in a LSP dependency to do the parsing for you or just piggy back on a HTTP parser https://docs.rs/httparse/latest/httparse/fn.parse_headers.html .

Alternatively you can use https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line to get a full line and then you ought to be able to more easily work with it instead of decoding one digit at a time.

internal/cmd/headless_repl/src/processes/lsp_invoke.rs
```diff
@@ -0,0 +1,349 @@
...
+    let mut child = Command::new("flux-lsp")
+        // .arg("")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failure to execute");
+    child
+}
+
+pub fn join_imports(lock: MutexGuard<HashSet<String>>, delim: &str) -> String {
```
Marwes, line 318:
```suggestion
pub fn join_imports(lock: &HashSet<String>, delim: &str) -> String {
```

You don't need to pass the whole mutexguard in, just pass it by reference instead. (This will change when the mutex is unlocked, but you can add a `drop(mutex_guard)` after calling `join_imports` if that is relevant.)

internal/cmd/headless_repl/src/processes/lsp_invoke.rs
```diff
@@ -0,0 +1,349 @@
...
+pub fn characters_after_last(input: &str, charac: &str) -> Option<u32> {
+    if input == "" || charac == "" {
+        return None;
+    }
+    let mut last_occur: Option<usize> = None;
+    for (i, c) in input.chars().enumerate() {
+        if c == '\n' {
+            last_occur = Some(i)
+        }
+    }
```
Marwes, line 343:
```suggestion
    let last_occur = input.chars().position(|c| c == '\n');
```

internal/cmd/headless_repl/src/processes/process_completion.rs
```diff
@@ -0,0 +1,115 @@
...
+            3 => FunctionType,
+            5 => ArgumentType,
+            6 => MethodType,
+            9 => PackageType,
+            _ => UnimplementedType,
+        }
+    }
+}
+
+impl Clone for HintType {
```
onelson, line 55: This `Clone` impl could be derived.

internal/cmd/headless_repl/src/processes/process_completion.rs
```diff
@@ -0,0 +1,115 @@
...
+            ArgumentType => ArgumentType,
+            MethodType => MethodType,
+            UnimplementedType => UnimplementedType,
+        }
+    }
+}
+
+pub fn process_completions_response(resp: &str) -> Option<HashSet<CommandHint>> {
+    //parse the response to a value using serde then enumerate the items adding each to the new set
+    let json_bit: Value = serde_json::from_str::<Value>(resp).expect("failed to change");
```
onelson, line 69: I think ideally you'd unpack the whole thing by using the response type for a completion request from `lsp_types` here. A lot of the traversal and conversion following below would probably be simplified by doing this.

internal/cmd/headless_repl/src/processes/process_completion.rs
```diff
@@ -0,0 +1,115 @@
...
+            FunctionType => FunctionType,
+            PackageType => PackageType,
+            ArgumentType => ArgumentType,
+            MethodType => MethodType,
+            UnimplementedType => UnimplementedType,
+        }
+    }
+}
+
+pub fn process_completions_response(resp: &str) -> Option<HashSet<CommandHint>> {
```
onelson, line 67: The function returns an `Option`, but there's a lot of `expect()` and `unwrap()` going on in the body. This indicates there's a lot of things that could go wrong.

Either split out the fallible stuff as a separate step (ie, parsing the json) or update the function to return a `Result` so you can bubble failure back up to the caller.

internal/cmd/headless_repl/src/processes/invoke_go.rs
onelson: @Westonside looks like you marked the error handling feedback as resolved, but the diff doesn't show the change.

You still need to have functions where you're using `unwrap()` or `expect()` return a `anyhow::Result`, then replace the `unwrap()` and `expect()` with `?`.

onelson (changes requested):

internal/cmd/headless_repl/src/processes/invoke_go.rs
onelson: Weston looked at using `httparse` but it seemed like for there to be a real win, we probably want to make a bigger change.

My hope is we can later lift out the transport and client code from `lspower` almost verbatim. They only have an implementation for Server->Client comms, but we should be able to build our own Client->Server version based closely on theirs.

Similarly, I'd like to try and rework the transport for talking to the REPL backend to rely on a library if we can find a good "plain jsonrpc" client. https://github.com/paritytech/jsonrpsee seems to only have http and websocket transports, but we might be able to implement our own for dealing with stdin/out. Alternatively, a move to gRPC might make sense since it would afford us some interesting cloud opportunities.

internal/cmd/headless_repl/rust-toolchain.toml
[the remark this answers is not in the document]
onelson: We got the go-ahead from the edge team to bump libflux up to 1.63. Sounds like they resolved the cross-build problems on their end.

We can loop back and remove this toolchain file when we bump.

onelson (approved):
Spent some time looking things over again today and I think most of the stuff I'd like to change can be done as follow-up PRs.

The main thing will be to get this merged so we can iterate on it as time allows.

PR #5042, opened 2022-08-02
feat: add support for vectorized binary equality ops

### LINKED ISSUE #4608: Vectorize equality operators
In `flux/ast/ast.go` several operators are defined, loosely grouped into 2 categories: equality and arithmetic. Where #4585 and #4586 focus on the _arithmetic_ group, this ticket tackles the operators that are left over.

The common theme with these is they all compare 2 values and return a bool.

- LessThanEqualOperator
- LessThanOperator
- GreaterThanEqualOperator
- GreaterThanOperator
- StartsWithOperator
- InOperator
- NotOperator
- ExistsOperator
- NotEmptyOperator
- EmptyOperator
- EqualOperator
- NotEqualOperator
- RegexpMatchOperator
- NotRegexpMatchOperator

DOD:
- provide a vectorized implementation of the remaining operators
- guard the implementation(s) with a feature flag
- Some operators are either not implemented in Flux yet or not accessible via vectorization, for those operators create a separate issue and we will do them later.

onelson:
Fixes #4608

This diff provides vectorized implementations for _**binary** equality operators_ which can be currently supported (regex, and duration are not currently valid types for table column data and so operators involving these have been omitted).

Offered by this PR:

- Equal
- NotEqual
- LessThan
- LessThanEqual
- GreaterThan
- GreaterThanEqual

The supported types for these are int, uint, float, string, time, and bool.

Bool is the exception where it only supports Equal/NotEqual. If I remember correctly there's a PR out there somewhere to add `Comparable` support for bool, but I don't think it has merged yet.

The new code path is guarded by a feature flag: `vectorizedEqualityOps`

---

Of the desired operators listed in #4608, the **unary** Exists and Not have not been provided since we currently don't support any vectorized unary expressions. It would be a heavier lift to break that new ground so it will be easier to review as a separate PR. Additionally, we need support for unary arithmetic Add and Sub, so it might make sense to do these all together, ref: https://github.com/influxdata/flux/issues/5045

---

I had challenges with the template code this time around, specifically with the handling of `semantic.Time`. I opted for a pragmatic approach where I effectively hand-wrote the case for Time (you'll see some comments about this in `flux/values/binary.gen.go.tmpl`) alongside the template driven cases. This helped to avoid Time showing up places it shouldn't when introduced in the associated template data.

This could be improved, but imo the template and data is in a bit of a tangle currently and probably needs a top-down refactor to straighten things out again. As soon as we can effectively benchmark our vectorized code paths, I think we should explore moving what we can to generics just to improve the maintainability.

### Checklist


jsternberg (changes requested):
What happens if the left and right side are different types that are castable? I believe this is possible for comparison operators.

The code looks fine for everything else. Request changes is mostly because I want an answer and tests for the above before merging.

onelson:
> What happens if the left and right side are different types that are castable? I believe this is possible for comparison operators.

It's a good question. I'm not sure. I'll investigate.

onelson:
Yes, absolutely, I see the standard equality ops will perform an ad hoc cast for certain L/R type combinations.

> As an aside, it looks like when we compare int and uint, the cast is int -> uint, rather than the other way. This would produce a runtime error, which seems less useful if we're trying to smooth over the type differences.

This will require some further work, which I'll get on this. Thanks for raising this!

Looks like I went on autopilot after my last PR in the binary op space and didn't realize this was needed.

onelson:
While working on the implicit casts some fixes were needed for the row-based versions where combinations of `int` and `uint` are compared and the `int` is less than zero.

To make these ops "make sense" for these cases, I had to flip the boolean outcomes for `>` and `>=` to ensure the opposite (`<` and `<=`) can't all be `true` or `false` at the same time.

This REPL session is a good example of the problem:

```
> a = -1
> b = uint(v: 1)
> a > b
true
> a < b
true
> b > a 
false
> b < a 
false
> 
```

We discussed possibly revising the way we do these implicit casts to be aligned with how it would work if the cast were written explicitly.
I'm going to file a separate issue for this work so we can deliberate.

For now, I'm focused on removing the apparent contradiction.

jsternberg (approved):
That switch looks awful. But also necessary.

PR #5043, opened 2022-08-03
fix: check bounds using rune array instead of string len in substring

onelson:
Inputs that contain non-breaking spaces (NBSP), and assumedly a host of
other unicode characters caused `strings.substring` to panic when it
tried to index into an array of runes taken from a (seemingly larger)
string.

Failures looked like:
```
panic: runtime error: slice bounds out of range [:33] with capacity 32
```

This only seemed to panic when the inputs were "of a certain size" and
I'm not sure why that would be. Perhaps there are distinct thresholds
where the input causes the capacity to be just close enough to the
actual size of the rune count, but less than the string length in bytes.
Not sure.

The testcase added used an input seen in the wild and panicked prior to
the code change to substring.

### Checklist

### DIFF
--- a/stdlib/strings/strings.go
+++ b/stdlib/strings/strings.go
@@ -488,7 +488,6 @@ func init() {
 				return values.NewString(strings.Join(newStringArray, argVals[1].Str())), nil
 			}, false,
 		),
-		"substring": substring,
 	}
 
 	runtime.RegisterPackageValue("strings", "joinStr", SpecialFns["joinStr"])
--- a/stdlib/strings/strings.go
+++ b/stdlib/strings/strings.go
@@ -355,13 +355,14 @@ var substring = values.NewFunction(
 			}
 
 			s := []rune(v)
+			runeCount := int64(len(s))
 			if a < 0 {
 				a = 0
-			} else if a > int64(len(v)) {
-				a = int64(len(v))
+			} else if a > runeCount {
+				a = runeCount
 			}
-			if b > int64(len(v)) {
-				b = int64(len(v))
+			if b > runeCount {
+				b = runeCount
 			} else if b < a {
 				b = a
 			}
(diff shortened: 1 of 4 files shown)


stdlib/strings/strings.go
```diff
@@ -487,6 +488,7 @@ func init() {
 				return values.NewString(strings.Join(newStringArray, argVals[1].Str())), nil
 			}, false,
 		),
+		"substring": substring,
```
wolffcm, line 491: I think this map is for "special" functions that can't have their builtin bodies "generated" like with `generateDualArgStringFunctionReturnBool` for example.

It seems like `substring` also falls into this category but is being registered in a different way. 

Maybe the intent was to do something like 
```go
for k, v := range specialFns {
  runtime.RegisterPackageValue("strings",k, v)
}
```
In any case it looks like some rot has seeped in here and it might be good to make this consistent?

wolffcm (approved):
Looks good. I had a comment about how this function is registered (it was a little inconsistent before you got here)

stdlib/strings/strings.go
onelson: The function is already registered, yes.

Best I could tell the map of "special" functions exists today solely for the purpose of exposing them to the test module. I didn't see any other access to the map outside of the now 2 fns, inserted here and the test module.

If you have a recommendation for how to get access to the function in the test module without this step, I'd be happy to remove it. It wasn't clear to me what API I should use to get a handle on the function otherwise.

onelson: Actually, now that I know the repro uses NBSP instead of null bytes, I can try and replace the Go test with an acceptance test (which sidesteps the access issue for me) :sweat_smile:

I really struggled to get a null byte into the source for a test or in the repl, but in retrospect I might have succeeded many times. The null byte just wasn't contributing to the bug as originally thought.

**Update**: this seems to have worked. I'll push a new pair of revs to move the test and remove the map usage.

Issue #5044, opened 2022-08-03
Issues with deleting a node from query plan in rules


wolffcm:
Note: before starting on this issue, we should complete https://github.com/influxdata/flux/issues/5092 first.

The recent PR https://github.com/influxdata/flux/pull/4981 makes some progress that lets the planner apply rules in more situations, but does not cover an important case where we are removing a node from the graph.

Consider the case for `RemoveTrivialFilterRule` where a filter like `filter(fn: (r) => true)` can just be removed. We can handle this fine if the filter's predecessor has just one successor (the trivial filter).

However, we can't handle a case like this:
```
src = from(...)
g = from |> filter(fn:(r) => true)
g => yield("0")
g |> map(...) |> yield("1")
src |> some_other_transformations |> yield("2")
```
Here is a graph of what the planner sees. I have chosen to put the successors at the top and the predecessors at the bottom, more like an expression tree. I wish I could make the arrows go the other way. In any case I find this easier to visualize and it looks like what the planner sees.
```mermaid
graph TD
  trivial_filter --> from
  yield0 --> trivial_filter
  yield1 --> map
  map --> trivial_filter
  some_other_transformations --> from
  yield2 --> some_other_transformations
```
And we want the planner to generate this:
```mermaid
graph TD
  yield0 --> from
  yield1 --> map
  map --> from
  some_other_transformations --> from
  yield2 --> some_other_transformations
```  
In this case the planner rule would need to return the predecessor to the filter, the `from` node, and the planner in turn would need to know to merge the successors of the deleted node with the successors of the deleted node's predecessor. The planner currently has some logic for this invoked here:
https://github.com/influxdata/flux/blob/e811f6c85a48d1dc39ae62201194f971f42b427c/plan/heuristic_planner.go#L136
But I don't think it will work quite right. I'd like to see a solution that can handle this.

We need to change the contract between the planner and planner rules a bit to make this work. Currently the planner rules are responsible for for returning a node with all of its predecessors connected, and the planner itself is responsible for connecting that node to its successors. But in a case where we delete a node, we may want to add more successors to the node that is returned, in addition to those that belonged to the deleted node. The planner doesn't quite have enough context to do this, so the planner rule will need to do it.

This is tricky to do, but we only have a few rules that do this--we should define a helper in the `plan` package.

DOD:
- We can remove a node from the graph even if its predecessor has multiple successors
- We have unit tests that demonstrate this works for cases like the one above
[...]

PR #5047, opened 2022-08-04
fix(plan): update pattern matching to specify successor counts

wolffcm:
This PR builds upon @nathanielc's previous PR #4981 whose text I Include here:

> The core of the problem is that the pattern matching logic
> expects there to be only a single edge between a node and
> its matched predecessor. This works and is required for rules
> that intend to merge the matched nodes into a single node.
> For example From -> Filter, if the from has other successors
> you can't merge the filter into the from. In other cases this
> restriction is not needed and prevents the rule for matching
> when it is safe to do so. For example a join node whose predecessor
> has a different successor, since the join rule doesn't rewrite its
> predecessors it doesn't matter if they have other successors.
> 
> To solve this we need to be explicit about whether a pattern match
> can have multiple successors or not. I propose we replace the existing
> plan.Pat function with two functions plan.Multi and plan.Single.
> plan.Multi will match a node and will allow that node to have multiple
> successors, while the plan.Single will match a node and that node must
> have exactly one successor. This changes two things about current patterns,
> first it makes the number of successors expected explicit in the pattern,
> second it moves the specification of which nodes can have multiple successors
> and which cannot on the pattern that matches the node instead of the
> successor of the node. Additionally plan.Any will allow multiple successors.

Fixes https://github.com/influxdata/flux/issues/4699

In addition to this I've updated this branch with a few changes:
- Changed `Single` and `Multi` to `SingleSuccessor` etc, also updated comments in `pattern.go`
- Fixed an issue with the join planner rule
- Fixed an issue with the heuristic planner where if multiple rules fired for the same node successors would become disconnected (both of these issues were caught by the new `join` acceptance test)
- Filed #5044 for future work, and added a test case for it in this PR

### Checklist

- [x] 🧪 If **new packages** are being introduced to stdlib, link to Working Group discussion notes and ensure it lands  under `experimental/` (N/A)


scbrickley (approved):
Looks good to me! Thanks for doing this.

Issue #4735, opened 2022-05-10
Replace calls to float with vectorized version


onelson:
In support of #4603, we need to update the rust code that produces our semantic graph such that it rewrites calls to `universe.float` to the vectorized version when applicable.

As far as I know, the condition to watch for is:
- there's a call to `float()` and it's inside a call to `map()`.

DOD:
- semantic graph will rewrite calls to `universe.float` as the vectorized version where possible.
- update type of call expr
- testing:
  - end to end: vectorized and vanilla should get the same answers.
  - if we had a way to assert the decision making around when vectorization happened, would be nice...

onelson:
Having some struggles here. 

My initial approach is to look for call expressions where the callee name and source file are `float` and `universe/universe.flux`. When this condition is met, I mess with the expression to vectorize the args and update the callee name to the new builtin.

With my current WIP, the gist is:

- for `float()` calls to pass static checks the analyzer needs to "see" stdlib.
- showing stdlib to the vectorize visitor means it will attempt to vectorize everything in stdlib, or at least symbols in the prelude (at least that's what it looks like in a debugger).
- much of the stdlib will fail to vectorize, so I think this prevents the function from vectorizing.

Without additional state tracking in the graph walk, it doesn't seem possible to know the difference between float calls elsewhere in the stdlib, and those inside the function we're trying to vectorize.

I'm not yet sure how to tell "we're within the function body or not". If we knew the symbols were not inside the function, we could try to ignore them by returning the unaltered expression (though, this may distort the signaling we use right now during this process).

---

I'm going to keep hacking around to see if I can find a different path. Maybe if I target the identifier expression, rather than the call expression, I'll have different luck.

Issue #4997, opened 2022-07-14
Support boolean literals for VectorRepeatValue


onelson:
The initial work towards wrapping constants/literals for vectorization only covered a subset of literals (int, float, string time). `boolean` was intended to be included but I (@onelson) was unable to figure out how to get them working on the rust side.

Rust complained about "undefined identifier" for `true` and `false` in tests and the bool literal match arm in `vectorize()` never seemed to match.

Possibly all that's missing is some configuration of the `Analyzer` but I wasn't sure how to make it work.

DOD:
- boolean literals can "vectorize" by being wrapped with `~~vecRepeat~~()` calls
- boolean literals produce `VectorRepeatValue` instances on the Go side
- tests for both Rust and Go areas of the code

onelson:
For the "undefined identifier" problem, it seems like the fix is to define the builtin local to the test, as was done here for float:

https://github.com/influxdata/flux/pull/5048#discussion_r946017720

For booleans, this will mean adding the following to your test source:

```
builtin true : bool
builtin false : bool

// ... the function to vectorize follows
(r) => ({ r with ... })
```

Marwes:
There is also `imp` https://github.com/influxdata/flux/blob/2be980478211aa6140833947fe9f7df5c941adfb/libflux/flux-core/src/semantic/tests.rs#L4205-L4208 and `env` https://github.com/influxdata/flux/blob/2be980478211aa6140833947fe9f7df5c941adfb/libflux/flux-core/src/semantic/tests.rs#L4257-L4260 to allow for things to be defind before the script.

Issue #4603, opened 2022-03-25
Vectorize `float()` function


onelson:
Requires #4736, #4734

In the rust:
- if a func
- and the arg is a float

separate built-in? `vecFloat` (maybe not)
use the existing built-in to branch internally

new semantic node - `cast` or add another operation to unary expr (since it's typed stronger).

---

Some ideas about implementation :arrow_up: but implementer may want to hold some sort of design session before diving in.

Seems like unary-like expr might be a heavier lift, but it's unclear.

Groom (with estimate) in this session.

---

DOD:
- provide a vectorized implementation for `float()`
- guard the new implementation with a feature flag

onelson:
Discussed the options today -- new issues to come, breaking this down into 3 areas:

- updates to parser to recognize vectorized types. (3)
- creating a new builtin for `_vectorizedFloat` (nb, the leading underscore is important). (3)
- updates to rust to rewrite invocations of `universe.float` as `universe._vectorizedFloat` in semantic graph. (3)

onelson:
# Discussion Notes

# Vectorized Float Working Group

## Questions

Some ideas were pitched. Do we like any of them?

- alternate/separate built-in impl: `vecFloat`
- new semantic node: `cast`
- add some new operation to UnaryExpr (since it's typed stronger)

2 considerations:
- how do we determine when to vectorize float?
  - planner can see isType and pushdown -- prior art could help
  - universe.float is float -- hijack it 

- do we add a vectorized float type?
  - go runtime goes through `values.Function` interface
  - evaluator: whatever, compiler: whatever
  - call evaluator is slow, but for a FloatVec could be okay?

---

vectorized evaluator looks for vectorizable stuff?
Might be simpler to stay with status quo.

Rust side recognizes float: 
- go handles the branch?
- go just calls what rust says because it tweaks types?

float (the function), vectorizedFloat (diff func): unrelated to Go.

Nb. Input type is polymorphic, but not output type.

---

Comms around vectorized alternatives being available -- we only
have a weak solution in place.
Want something stronger, attribute based?

Rust can hardcode the fact `universe.float` is vectorizable.

---

Can we surface the vectorizedFloat to flux?

Pro: fits existing schemes if we do it that way: just a function call.
Con: leaky.

**Urgency to not leak stuff to the frontend.**

> When the (rust) optimizer is the only thing that can invoke a thing,
it would be nice if that thing wasn't available generally in flux.

`internal` package might be an okay to indicate this is not for general use.


[...]

PR #5048, opened 2022-08-05
feat: rewrite calls to `float()` as `_vectorizedFloat()`

onelson:
Fixes: #4735, #4603 

With this diff, the vectorization pass of the semantic graph will rewrite calls to `float()` such that they call `_vectorizedFloat()` instead (provided by #5015).

In order for the vectorization pass to type check in tests, the analyzer had to be configured to see the stdlib such that `float` would not give an "undefined identifier" error.
This was the same error which prevented boolean literals from vectorizing (in tests only, or so it seems) so #4997 is likely unblocked with these changes.

The function identifier rewrite is guarded by a feature flag: `vectorizedFloat`

### Checklist


stdlib/universe/map_vectorize_float_test.flux
```diff
@@ -0,0 +1,71 @@
...
+// FIXME: add bool literals to this once they are enabled in #4997
+testcase vec_with_float_const {
+    want = array.from(rows: [{x: 1, f: 123.456, i: 123.0}])
+
+    got =
+        array.from(rows: [{x: 1}])
+            |> map(fn: (r) => ({r with f: float(v: 123.456), i: float(v: 123)}))
+
+    testing.diff(want, got)
+}
```
wolffcm, line 71: Does it make sense to add a call to the stdlib's `toFloat` function in here?

It would be nice if we moved the vectorization logic into a planner rule so we could assert on the rule enabling vectorization.

libflux/flux-core/src/semantic/tests/vectorize.rs
```diff
@@ -1,24 +1,33 @@
...
 fn vectorize(src: &str) -> anyhow::Result<Package> {
-    let mut analyzer = Analyzer::new(Default::default(), Packages::default(), analyzer_config());
+    let stdlib_path = PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR"), "/../../stdlib"))
+        .canonicalize()
+        .expect("stdlib path");
+
+    let cfg = analyzer_config();
+    let (prelude, imports, _sem_pkgs) = bootstrap::infer_stdlib_dir(&stdlib_path, cfg.clone())?;
+
+    let mut analyzer = Analyzer::new((&prelude).into(), &imports, cfg);
```
wolffcm, line 30: Is this inferring the whole stdlib every time we receive a request?

wolffcm (commented):
Looks good to me. I just had a question about if we're inferring the stdlib when maybe we don't need to?

libflux/flux-core/src/semantic/tests/vectorize.rs
onelson: I think "yes" but more specifically, note that this is happening for regular scripts regardless.

This particular analyzer is used for these tests specific tests - the change here has no impact on the regular execution of flux otherwise.

Previously there was no need for the prelude (or any other part of the stdlib) to be visible during these test runs. Probably this was skipped 
- to be faster
- it was needless (before now)
- (my observation working on this issue) it creates a lot of noise if you're watching in a debugger

stdlib/universe/map_vectorize_float_test.flux
onelson: I was just looking at queries in lovelace to see what would be needed and I was _literally just thinking about this_. For lovelace I'm planning to have 2 queries, one with a hand-written cast (like you see here) and a 2nd relying on `toFloat()` instead.

I will expect to see the same perf increase for both when we flip the feature flag.

I'm happy to add a test here using `toFloat()` and verify we hit the vectorized path using the debugger before this merges.

libflux/flux-core/src/semantic/tests/vectorize.rs
onelson: Oh, and to speak to why this is needed -- without this (or something similar) vectorization will fail in these tests because the program will error immediately with `undefined identifier: float`.

wolffcm: Ah, ok. I missed that this was test code.

wolffcm (approved):
Looks great.

stdlib/universe/map_vectorize_float_test.flux
onelson: So, I wrote a test to compare a plain map against the tables produced by `toFloat()` and it turns out, the `toFloat()` call is _not_ vectorizing while the `map()` _is_. This is unexpected.

I can investigate further, but I wonder if this should hold up the PR from merging. I'd like to get a better understanding of why these act different just to strengthen my mental model.

libflux/flux-core/src/semantic/tests/vectorize.rs
onelson, line 30: @Marwes This is the change I made allowing my tests to include calls to `float`. If I understood you during standup, your recommendation instead is to effectively hand-write the builtin definition for the bits each test needs.

I'm not sure how that should be done. Any advice?

onelson: Huh, okay. Rewriting the flux source like the following, without the prelude configured does seem to work okay.

```
builtin float : (v: A) => float

(r) => ({ r with a: float(v: r._value) })
```

I'll make a change for this and make a note on #4997

Marwes: Yeah, there is also `imp` and `env` to let us set things like this up https://github.com/influxdata/flux/issues/4997#issuecomment-1216424229

onelson: Moved this up to where we initialize the analyzer in 889816d1d4d952864b6eb7034a3f10761ee9f660 -- feels more official to do it this way :+1:

stdlib/universe/map_vectorize_float_test.flux
onelson: I've opened https://github.com/influxdata/flux/issues/5102 to investigate this further, but it looks like we might just not be vectorizing anything that isn't defined in the main package right now.

Redefining `toFloat` in `main` like so, then using it, showed it hits the vectorized code path.

```
toFloat2 = (tables=<-) => tables |> map(fn: (r) => ({r with _value: float(v:r._value)}))
```

I'm guessing there are could be other functions in stdlib that could be vectorized but aren't.

PR #5055, opened 2022-08-09
feat: Allow any kind of AST fragment to be formatted

Marwes:
Adds `format_node` which accepts any AST fragment that `Node` can contain
and formats it. Should be useful for generating code fragments programmatically
(like that LSP wants to do) as this lets any AST be turned back into a string paving the way
for something akin to https://docs.rs/quote/latest/quote/macro.quote.html  (cc @rockstar ).

(I say string here, but it doesn't need to be a string per-se, as long as it is a common type
that the AST gets transformed to that may be passed around)

### Checklist


libflux/flux-core/src/formatter/mod.rs
```diff
@@ -329,24 +346,84 @@ impl<'doc> Formatter<'doc> {
...
+            Node::PropertyType(x) => self.format_property_type(x),
+            Node::ParameterType(x) => self.format_parameter_type(x),
+            Node::TypeConstraint(x) => self.format_type_constraint(x),
+            Node::BadExpr(_) | Node::BadStmt(_) | Node::Package(_) => {
+                return Err(anyhow!("Unable to format node"))
+            }
+        })
+    }
+
+    fn format_package_clause(&mut self, pkg: &'doc ast::PackageClause) -> Doc<'doc> {
```
Marwes, line 400: All changes from here on in this file are just boring extraction of code into functions so I could call them in `format_node`

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -52,12 +63,18 @@ pub struct Options {
     ///
     /// Default: true
     pub trailing_newline: bool,
+
+    /// How long a line is allowed to be before newlines are forced
+    ///
+    /// Default: 120
```
wolffcm, line 69: There's been some internal discussion about line length as it affects accessibility, where `100` has been suggested.

wolffcm: cc @pierwill

wolffcm (approved):
This looks good to me. Maybe change default line length to 100?

libflux/flux-core/src/formatter/mod.rs
Marwes: Sure, the only reason it was so high is that when I rewrote the formatter, having this line length kept the formatting that existed before (so less noise while getting things merged)

wolffcm: Ah ok. Maybe we should change it to 100 as a separate issue.

PR #1418, opened 2019-06-26
Full draft of Rust parser

affo:
### Done checklist
- [x] `peek_with_regex` -> the Rust scanner doesn't offer `scan_with_regex` yet
- [x] more test cases
- [ ] source location

PR #5056, opened 2022-08-09
refactor: Remove serde-aux

Marwes:
~The default attribute does the same thing so no need for this dependency. (The crate seems to have been added in the original creation of the AST, maybe `default` didn't exist then or something https://github.com/influxdata/flux/pull/1418 )~

The default attribute only handles undefined fields, not null fields. But the function is so simple
to add that we do not need an dependency really.

### Checklist

### DIFF
--- a/libflux/Cargo.lock
+++ b/libflux/Cargo.lock
@@ -441,7 +441,6 @@ dependencies = [
  "regex",
  "rusqlite",
  "serde",
- "serde-aux",
  "serde_derive",
  "serde_json",
  "structopt",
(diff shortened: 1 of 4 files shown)


jsternberg (approved):
I assume this is just pulling out some code from `serde-aux` into our own codebase so we don't need the dependency? If this is the case, can you add a comment that points to a permalink of the implementation that was copied?

If it's not copied, then you don't need to do this. Maybe just a doc comment describing what's happening.

PR #5065, opened 2022-08-10
chore: fix or skip failing integration tests

scbrickley:
This patch modifies some tests that were previously skipped in our nightly integration run in order to get them to pass. Any tests in the skiplist for the nightly run that could not be easily fixed have had the `skip` tag added to them.

I've created some issues to track the work to resolve the remaining skipped tests:

- https://github.com/influxdata/flux/issues/5068
- https://github.com/influxdata/flux/issues/5070
- https://github.com/influxdata/flux/issues/5071
- https://github.com/influxdata/flux/issues/5072
- https://github.com/influxdata/flux/issues/5073
- https://github.com/influxdata/flux/issues/5074
- https://github.com/influxdata/flux/issues/5075


stdlib/universe/cov_test.flux
```diff
@@ -40,7 +40,10 @@ outData =
 "
 
 testcase cov {
-    table = csv.from(csv: inData) |> testing.load()
+    table =
+        csv.from(csv: inData)
+            |> testing.load()
+            |> range(start: 2018-05-22T19:53:00Z, stop: 2018-05-22T19:54:00Z)
```
wolffcm, line 46: Just an observation: for some of these tests I wonder if we should just remove the call to `testing.load()`. We do not push down `cov()` for example, so I don't see much reason to load data into storage. It would also make the tests run a lot faster.

wolffcm (approved):
Looks good.

PR #5076, opened 2022-08-10
fix: fix aggregate window rules that left plan in bad state

wolffcm:
This PR fixes an issue with the planner rules that replace the composite `aggregateWindow` operation (composed of `window |> fill |> agg |> duplicate |> window(inf)`) with a unified `aggregateWindow` operation.

This rule was assuming that the predecessor to the pattern had just one successor. If this was not the case, the plan would become corrupt and planning would fail. This was being hidden before #5047 merged because a pattern that had a predecessor with multiple successors would never be reported as a match.

There are also some changes in this PR to improve diagnostics when things go wrong in the planner, and to improve the DOT representation of query plans.

### Checklist

### DIFF
--- a/plan/types.go
+++ b/plan/types.go
@@ -343,6 +343,23 @@ func ReplaceNode(oldNode, newNode Node) {
 	oldNode.ClearPredecessors()
 }
 
+// ReplacePhysicalNodes accepts a connected group of nodes that has a single output and
+// a single input, and replaces them with a single node with the predecessors of the old input node.
+// Note that the planner has a convention of connecting successors itself
+// (rather than having the rules doing it) so the old output's successors
+// remain unconnected.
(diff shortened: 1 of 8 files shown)


Marwes (approved):

plan/types.go
```diff
@@ -343,6 +343,23 @@ func ReplaceNode(oldNode, newNode Node) {
 	oldNode.ClearPredecessors()
 }
 
+// ReplacePhysicalNodes accepts a connected group of nodes that has a single output and
+// a single input, and replaces them with a single node with the predecessors of the old input node.
+// Note that the planner has a convention of connecting successors itself
+// (rather than having the rules doing it) so the old output's successors
+// remain unconnected.
+func ReplacePhysicalNodes(ctx context.Context, oldOutputNode, oldInputNode Node, name string, newSpec PhysicalProcedureSpec) Node {
```
Marwes, line 351: I would have needed this function before :heart:

PR #5077, opened 2022-08-10
chore: make flux test use pretty formatting for diffs

wolffcm:
This just makes the output of `flux test` be more readable, e.g.,
```
failures:

	stdlib/universe/aggregate_window_test.flux: aggregate_window_predecessor_multi_successor ... fail: Table: keys: [t0]
             t0:string            _diff:string                      _time:time                  _value:int
----------------------  ----------------------  ------------------------------  --------------------------
                   a-1                       -  2019-11-25T00:01:00.000000000Z                           6
                   a-1                       +  2019-11-25T00:01:00.000000000Z                           5
Table: keys: [t0]
             t0:string            _diff:string                      _time:time                  _value:int
----------------------  ----------------------  ------------------------------  --------------------------
                   a-z                       -  2019-11-25T00:01:00.000000000Z                           5
Table: keys: [t0]
             t0:string            _diff:string                      _time:time                  _value:int
----------------------  ----------------------  ------------------------------  --------------------------
                   a-0                       +  2019-11-25T00:01:00.000000000Z                           5
```
The diff currently looks like this and I find it quite hard to parse (this is a different test than the one above but hopefully you get the idea):
```
failures:

	stdlib/universe/aggregate_window_test.flux: max_null_windows ... fail: # _field=f0,_measurement=m0 _diff=string,_time=time,_value=float,t0=string
_field=f0,_measurement=m0 _diff=,_time=2019-11-25T00:00:00Z,_value=6.000,t0=a-0
_field=f0,_measurement=m0 _diff=,_time=2019-11-25T00:00:10Z,_value=2.000,t0=a-0
_field=f0,_measurement=m0 _diff=,_time=2019-11-25T00:00:20Z,_value=0.000,t0=
_field=f0,_measurement=m0 _diff=-,_time=2019-11-25T00:00:30Z,_value=3.100,t0=a-0
_field=f0,_measurement=m0 _diff=+,_time=2019-11-25T00:00:30Z,_value=3.000,t0=a-0
_field=f0,_measurement=m0 _diff=,_time=2019-11-25T00:00:40Z,_value=4.000,t0=a-0
_field=f0,_measurement=m0 _diff=,_time=2019-11-25T00:00:50Z,_value=0.000,t0=
```

### Checklist

Dear Author :wave:, the following checks should be completed (or explicitly dismissed) before merging.

- [x] ✏️ Write a PR description, regardless of triviality, to include the _value_ of this PR
- [x] 🔗 Reference related issues
[...]

PR #5080, opened 2022-08-11
feat: Add Stringify function to table.Chunk

Marwes:
Useful for getting a human readable view into a chunk while debugging. I tweaked `table.Chunk`s methods slightly so it implements `ColReader`, allowing it to be used with `stringifyRows`.

Should maybe use the formatting used in https://github.com/influxdata/flux/pull/5077 instead, but I am getting errors about circular depencies which I am not in the mood for untangling.

### Checklist

### DIFF
--- a/execute/table/chunk.go
+++ b/execute/table/chunk.go
@@ -1,6 +1,8 @@
 package table
 
 import (
+	"strings"
+
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/array"
 	"github.com/influxdata/flux/arrow"
@@ -108,9 +110,9 @@ func (v Chunk) Ints(j int) *array.Int {
 	return v.Values(j).(*array.Int)
 }
 
-// Uints is a convenience function for retrieving an array
+// UInts is a convenience function for retrieving an array
 // as a uint array.
-func (v Chunk) Uints(j int) *array.Uint {
+func (v Chunk) UInts(j int) *array.Uint {
 	return v.Values(j).(*array.Uint)
 }
 
@@ -126,6 +128,12 @@ func (v Chunk) Strings(j int) *array.String {
 	return v.Values(j).(*array.String)
 }
 
+// Times is a convenience function for retrieving an array
+// as a time array.
+func (v Chunk) Times(j int) *array.Int {
+	return v.Ints(j)
+}
+
 // Retain will retain a reference to this Chunk.
 func (v Chunk) Retain() {
 	v.buf.Retain()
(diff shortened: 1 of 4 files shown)


execute/table/chunk.go
```diff
@@ -108,9 +110,9 @@ func (v Chunk) Ints(j int) *array.Int {
 	return v.Values(j).(*array.Int)
 }
 
-// Uints is a convenience function for retrieving an array
+// UInts is a convenience function for retrieving an array
 // as a uint array.
-func (v Chunk) Uints(j int) *array.Uint {
+func (v Chunk) UInts(j int) *array.Uint {
 	return v.Values(j).(*array.Uint)
 }
```
onelson, line 117: I have a vague memory of @scbrickley proposing this exact change a couple months ago and it was rolled back for some reason. Am I remembering it wrong? If not, maybe Sean has context.

onelson (approved):
LGTM, but probably wait for @scbrickley to weigh in before merging. Sorry I can't muster a better memory of the exchange, whatever it was.

execute/table/chunk.go
scbrickley: I think I wanted to change this (or some similar method) to match the [flux column type](https://github.com/influxdata/flux/blob/a42e5be988f19519b824807992c6846792dbbc8a/result.go#L87), but @jsternberg pointed out that the method was meant to match the [type from the `array` package](https://github.com/influxdata/flux/blob/a42e5be988f19519b824807992c6846792dbbc8a/array/builder.gen.go#L16).

I don't have a strong preference for either. Just weighing in with the context.

jsternberg: `table.Chunk` is very intentionally not supposed to implement `flux.ColReader`. Part of that is because I didn't want to propagate the interface more as, in an ideal world, `flux.ColReader` would be gone. Part of it is because I wanted to normalize the method names to match the array package.

Similar to the `Times()` method below. `flux.ColReader` had code to panic if you accessed an array the wrong way which was awkward for times since they were just integers. This was intentionally changed.

jsternberg (changes requested):
I like the function but I think I gave you an implementation of this that used the existing stringify code and didn't involve changing any structs or interfaces. I would prefer that method.

`table.Chunk` is intentionally supposed to be separate from `flux.ColReader` and the two aren't meant to be interchangeable. This was mostly to prevent confusing code where they were used interchangeably instead of being treated differently and also not to force `table.Chunk` to conform to an older API.

execute/table/chunk.go
Marwes: The code you gave me caused circular dependencies so this seemed like a simpler change :shrug:

jsternberg: I think that might have been because the sample code I gave referenced the `table` package, but this is the `table` package. Which means that it referenced the internal one instead of itself and the internal one contains type aliases to this package. If you just remove the references to the `table` package, it seems to work. I'll put up a PR.

Marwes: That might be it, the error for circular dependencies are so awful I didn't realize...

jsternberg: They are. Alternate PR is up here: https://github.com/influxdata/flux/pull/5090

PR #5081, opened 2022-08-11
test: Add debug.null()

Marwes:
Allow null values to be created explicitly in flux code. We could always create nulls by using `csv.from` however that format is pretty hard to read compared to `array.from` but since the latter could not create streams with nulls we were forced to use `csv.from`. With this we can use `array.from` in cases where nulls are needed in the test.

### Checklist

### DIFF
--- a/stdlib/internal/debug/debug.flux
+++ b/stdlib/internal/debug/debug.flux
@@ -54,7 +54,14 @@ builtin getOption : (pkg: string, name: string) => A
 builtin feature : (key: string) => A
 
 // Null returns the null value with the given type
-builtin null : (?type: string) => A
+//
+// ## Parameters
+// - type: Which type to give the null
+//
+// ```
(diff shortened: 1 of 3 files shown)


stdlib/internal/debug/debug.flux
```diff
@@ -53,5 +53,8 @@ builtin getOption : (pkg: string, name: string) => A
 //
 builtin feature : (key: string) => A
 
+// Null returns the null value with the given type
+builtin null : (?type: string) => A
```
onelson, line 57: If there were an example here, I suppose it might look like:

```
import "internal/debug"
import "array"

array.from(rows: [{a: 1, b: 2, c: 3}, {a: debug.null("int"), b: 5, c: 6}])
```

Is that the idea? Neat, if so. I will be glad to use this instead of relying on combinations of `union()` and `debug.opaque()`.

onelson (approved):
LGTM, but maybe a small example could be added to the docstring.

Issue #4655, opened 2022-04-08
`debug.unpivot` can panic


onelson:
So, I accidentally stumbled into a panic while using `debug.unpivot()`. Seems like at a minimum we might adjust the code path to return an error instead, but there may also be some desired changes to its behavior.

I've got a repro, but I've really struggled to reduce it further than this. I've played around trying to get more precise a method for triggering the issue, but I think I'd have to crawl the codepath in a debugger to get any more insight.

```
import "internal/gen"
import "internal/debug"

tags = [
  { name: "t0", cardinality: 1 }, 
  { name: "t1", cardinality: 1 }, 
]

gen.tables(tags: tags,  n: 1)
  |> map(fn: (r) => ({ r with something: 1000 }))  // this map is a part of it somehow
  |> group()
  |> drop(columns: ["xyz"]) // You can name any column, present or not 
  |> debug.unpivot() // internal error
  |> debug.pass() // panics
```

The error I get back when I run this script is:

```
panic: interface conversion: array.Array is *array.Float64, not *array.String
```

So, I can't really explain how the `map()` contributes, but it does.

Of particular interest are the last 2 lines. If your remove the `debug.pass()`, the script will result in an "internal error" (which is good!)

It does seem than adding any successor to an already failing `debug.unpivot()` will cause the panic to be surfaced directly rather than the more generic error message. I've seen this with `limit()` and `count()`, but `debug.pass()` triggering this behavior makes me think it might be general to _any successor_.

Initially, I thought maybe the `drop()` had to be removing a field in the group key, but it turns out this is not the case. Any `drop()` seems to contribute.

PR #5082, opened 2022-08-11
fix: Don't include null columns when unpivoting

Marwes:
The pivoted data we get from IOx may contain fields with nulls which we must exclude from the unpivoted chunk so they do not show up as extra rows with `null` as the `_value`.

Based on #5079 #5081

### Checklist


onelson (approved):
Looks good.

I had hoped this might be what I ran into with #4655, but it still reproduces on this branch :sob:

PR #5086, opened 2022-08-12
fix: Don't report an error about testcases in the LSP

Marwes:
We should only need to care about testcases when actually serializing to flatbuffer, so just make that error instead of panicking and remove the pre-check.

Fixes https://github.com/influxdata/flux-lsp/issues/403 (again? the LSP seems to run semantic checks now which is why it was reintroced/is still reporting this error)

### Checklist


Marwes:
ping @jsternberg

jsternberg (approved):

PR #5095, opened 2022-08-12
feat(libflux): adds typecmp to compare Flux types

nathanielc:
WIP

TODO:

- [x] Choose a better name
- [x] Write tests
- [ ] Document rules
- [x] Report specifics about changes instead of just the aggregate Major,Minor,Patch diff
- [ ] Determine how we would like to use this in CI, i.e. compare master against most recent Flux release? Then what?

### Checklist

### DIFF
--- a/libflux/flux-core/Cargo.toml
+++ b/libflux/flux-core/Cargo.toml
@@ -25,6 +25,12 @@ name = "fluxc"
 test = false
 bench = false
 
+[[bin]]
+name = "typecmp"
+test = false
+bench = false
+required-features = ["doc"]
+
 [features]
 default = ["strict"]
 strict = []
--- a/libflux/flux-core/src/bin/typecmp.rs
+++ b/libflux/flux-core/src/bin/typecmp.rs
@@ -1,0 +1,144 @@
+use std::{cmp::max, fs, path::PathBuf};
+
+use anyhow::{anyhow, Result};
+use fluxcore::{
+    doc::{Doc, PackageDoc},
+    parser,
+    semantic::{
+        cmp::{self, TypeDiff},
+        convert::convert_polytype,
+        types::PolyType,
+        AnalyzerConfig,
+    },
+};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(about = "compare Flux types")]
+enum TypeCmp {
+    /// Diff an old and new Flux types
+    Diff {
+        /// Old Flux type
+        #[structopt(short, long, parse(from_os_str))]
+        old: PathBuf,
+        /// New Flux type
+        #[structopt(short, long, parse(from_os_str))]
+        new: PathBuf,
+        /// Versbose
+        #[structopt(short, long)]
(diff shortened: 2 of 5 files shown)


libflux/flux-core/src/semantic/version.rs
```diff
@@ -0,0 +1,139 @@
...
+                if let Some(nfield) = new.fields().find(|f| f.k == ofield.k) {
+                    if diff_monotype(&ofield.v, &nfield.v) == VersionDiff::Major {
+                        return VersionDiff::Major;
+                    }
+                } else {
+                    return VersionDiff::Major;
+                }
+            }
+            // TODO in some cases adding a new record field is breaking and in
+            // others its not. How do we handle this ambiquity?
```
Marwes, line 52: I think it is always a breaking change so I don't think there is any ambiguity

libflux/flux-core/src/semantic/version.rs
```diff
@@ -0,0 +1,139 @@
...
+            if old.collection != new.collection {
+                VersionDiff::Major
+            } else {
+                diff_monotype(&old.arg, &new.arg)
+            }
+        }
+        (MonoType::Record(old), MonoType::Record(new)) => {
+            for ofield in old.fields() {
+                if let Some(nfield) = new.fields().find(|f| f.k == ofield.k) {
+                    if diff_monotype(&ofield.v, &nfield.v) == VersionDiff::Major {
```
Marwes, line 44: If this returns `minor` we need to propagate that up

libflux/flux-core/src/semantic/version.rs
```diff
@@ -0,0 +1,139 @@
...
+            VersionDiff::Minor
+        }
+        (MonoType::Dict(ref old), MonoType::Dict(ref new)) => {
+            match (
+                diff_monotype(&old.key, &new.key),
+                diff_monotype(&old.val, &new.val),
+            ) {
+                (VersionDiff::Major, _) | (_, VersionDiff::Major) => VersionDiff::Major,
+                (VersionDiff::Patch, VersionDiff::Patch) => VersionDiff::Patch,
+                (_, _) => VersionDiff::Minor,
```
Marwes, line 68: Could extract this into a function to reuse as I think you need to "merge" two `VersionDiff`s else where.

Marwes: Actually, this is just `max`, which I see you have used already.

libflux/flux-core/src/semantic/version.rs
```diff
@@ -0,0 +1,139 @@
...
+        return VersionDiff::Patch;
+    }
+
+    // TODO handle constraints
+
+    diff_monotype(&old.expr, &new.expr)
+}
+
+fn diff_monotype(old: &MonoType, new: &MonoType) -> VersionDiff {
+    if old == new {
```
Marwes, line 30: Nit: This isn't really a performance sensitive code, but doing `==` at every level of the `Monotype` makes it `O(N^2)`

PR #5099, opened 2022-08-15
fix(stdlib): testing.shouldError uses a regex instead of a string

jsternberg:
The changes to the test framework necessitated this change. Since the
changes to the test framework now include the full path, we can
potentially run the tests with different paths being passed into the
test framework.

This meant that `shouldError` wasn't able to capture errors because we
can't do a single string that matches the error due to file names.

In general, it's better to do this kind of comparison with regexes so
I've changed it to use a regex.

### Checklist


onelson (approved):
I love it. This will help us keep line numbers and so on out of the assertion.

scbrickley (approved):

PR #5105, opened 2022-08-16
fix: sort columns when printing group key

### LINKED ISSUE #5104: Panic in pivot when group column order changes
When a user's query makes a call to `pivot` where there are two different tables that may have the same group key, but the columns are expressed in a different order, a panic will result.

Example:
```js
import "array"
import "influxdata/influxdb/schema"

in0 = array.from(rows: [
  {_measurement: "m0", _field: "f0", tag0: "val0", _value: 10.0, _time: 2021-01-01T00:00:00Z},
  {_measurement: "m0", _field: "f0", tag0: "val0", _value: 11.0, _time: 2021-01-01T00:00:10Z},
])
  |> group(columns: ["_measurement", "_field", "tag0"])
  |> range(start: 2021-01-01T00:00:00Z)

in1 = array.from(rows: [
  {_field: "f1", tag0: "val0", _value: 10.0, _time: 2021-01-01T00:00:00Z, _measurement: "m0"},
  {_field: "f1", tag0: "val0", _value: 11.0, _time: 2021-01-01T00:00:10Z, _measurement: "m0"},
])
  |> group(columns: ["_measurement", "_field", "tag0"])
  |> range(start: 2021-01-01T00:00:00Z)

union(tables: [in0, in1])
  |> schema.fieldsAsCols()
```
this produces
```
panic: assignment to entry in nil map
```

wolffcm:
The flux pivot() function uses a string representation of
the group key to keep track of what columns it has seen.
Sometimes the columns of the group key can appear in a different order
even if the set of columns and their values are the same. when this
happens, pivot() panics as it uses a map that assumes all group
keys have a canonical representation.

So the `groupKeyString` that is computed here:
https://github.com/influxdata/flux/blob/cf5caf06789c3e6995998a077ffc411f80582c2f/stdlib/universe/pivot.go#L264

Is later used here to access a map in a map-of-maps:
https://github.com/influxdata/flux/blob/cf5caf06789c3e6995998a077ffc411f80582c2f/stdlib/universe/pivot.go#L307

And sometimes it may happen that the string value used to create an entry in the
outer map may be different than the string value used to access that entry later on.

The fix here is to make `groupKey.String()` use the sorted order
of the group columns when returning its result, so that the group key string
is always the same.

Fixes #5104.

### Checklist

### DIFF
--- a/internal/execute/groupkey/groupkey.go
+++ b/internal/execute/groupkey/groupkey.go
@@ -92,11 +92,11 @@ func (k *groupKey) Less(o flux.GroupKey) bool {
 func (k *groupKey) String() string {
 	var b strings.Builder
 	b.WriteRune('{')
-	for j, c := range k.cols {
-		if j != 0 {
+	for i, idx := range k.sorted {
+		if i != 0 {
 			b.WriteRune(',')
 		}
-		fmt.Fprintf(&b, "%s=%v", c.Label, k.values[j])
+		fmt.Fprintf(&b, "%s=%v", k.cols[idx].Label, k.values[idx])
 	}
 	b.WriteRune('}')
 	return b.String()
--- a/internal/execute/groupkey/groupkey_test.go
+++ b/internal/execute/groupkey/groupkey_test.go
@@ -3,6 +3,7 @@ package groupkey_test
 import (
 	"testing"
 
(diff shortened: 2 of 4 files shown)


onelson (approved):
Makes perfect sense. Thanks for digging so deep to find this!

PR #5112, opened 2022-08-18
chore: upgrade Rust toolchain to 1.63

bnpfeife:
I've worked out the issues with the InfluxDB cross-builder/build system and we can now build with new versions of the Rust toolchain.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.62.1"
+channel = "1.63.0"
 components = ["rustfmt", "clippy"]
 targets = [
     "wasm32-unknown-unknown",
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,9 +1,5 @@
 [toolchain]
-# Note: we are stuck on Rust version 1.58 for the time being until
-# this issue is fixed:
-#   https://github.com/rust-lang/rust/issues/97117
-# Since it blocks the ARM64 build of InfluxDB.
-channel = "1.58"
+channel = "1.62"
 components = ["rustfmt", "clippy"]
 targets = [
     "wasm32-unknown-unknown",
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,9 +1,5 @@
 [toolchain]
-# Note: we are stuck on Rust version 1.58 for the time being until
-# this issue is fixed:
-#   https://github.com/rust-lang/rust/issues/97117
-# Since it blocks the ARM64 build of InfluxDB.
-channel = "1.58"
+channel = "1.62"
```
onelson, line 2: A couple of us ran into an issue with rustup:
https://github.com/rust-lang/rustup/issues/3061 

As a workaround, could we specify `1.62.1` _or if the cross-build is okay on `1.63`_ maybe we could bump all the way up?

```suggestion
channel = "1.62.1"
```

n.b. if we explicitly set `1.62.1` then the Dockerfile should probably match.

bnpfeife: Whew, time flies. I hadn't realized that 1.63 was out. I'm sure it will work; however, could we stick with 1.62.1 right now? I've got the cross-builder using [1.62.1](https://github.com/influxdata/edge/blob/a6d0dbfb2e0f2313cdf67ade0693047aa0690f94/dockerfiles/cross-builder/install-rust.sh#L7), and it'll take at least 6 PRs in `influxdb` and other repositories to get it updated. :sweat_smile:

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.62.1"
+channel = "1.63.0"
```
onelson, line 2:
```suggestion
channel = "1.63"
```

Best I can tell, the rustup issue only applied to `1.62` :shrug: It's weird. `1.63` seemed to work fine.

onelson (commented):
Looks like there's a bunch of new lints firing with 1.63 -- I think we can probably take it from here to clean that stuff up.

The main thing for us is just knowing we've got a green light from y'all to bump up to 1.63.

onelson:
Picked this up over in #5146

PR #5115, opened 2022-08-19
fix(plan): fix logic bug in planner helper method

### LINKED ISSUE #5114: Plan integrity violated when apply sort |> limit rule
This test case causes a `integrity violated` error:
```
testcase sort_limit_multi_successor {
    input =
        array.from(
            rows: [
                {_time: 2022-01-11T00:00:00Z, _value: 10.0},
                {_time: 2022-01-11T01:00:00Z, _value: 12.0},
                {_time: 2022-01-11T02:00:00Z, _value: 18.0},
                {_time: 2022-01-11T03:00:00Z, _value: 4.0},
                {_time: 2022-01-11T04:00:00Z, _value: 8.0},
            ],
        )
    in0 =
        input
            |> bottom(n: 2)
    in1 =
        input
            |> top(n: 2)
    got =
        union(tables: [in0, in1])
            |> sort(columns: ["_time"])

    want =
        array.from(
            rows: [
                {_time: 2022-01-11T01:00:00Z, _value: 12.0},
                {_time: 2022-01-11T02:00:00Z, _value: 18.0},
                {_time: 2022-01-11T03:00:00Z, _value: 4.0},
                {_time: 2022-01-11T04:00:00Z, _value: 8.0},
            ],
        )

    testing.diff(got: got, want: want)
}
```
This can happen when `sort |> limit` appears in a query, or `top` or `bottom`, which translate to `sort |> limit`. When a predecessor to the `sort` node has multiple successors, this internal error can occur.

wolffcm:
Fixes #5114.

This PR addresses an issue that has existed for a long time, but did not manifest until #5047 was merged. 

The issue was in a `plan` helper method that merges two nodes. If the predecessor to one of the nodes had a predecessor with multiple successors, we were not updating edges correctly and we put the plan graph into an inconsistent state.

Previously this issue could not be hit because we did not allow many rules to match in multi-successor scenarios, and this bug is particular to that situation.

### Checklist

### DIFF
--- a/plan/types.go
+++ b/plan/types.go
@@ -265,8 +265,8 @@ func mergePlanNodes(top, bottom, merged Node) (Node, error) {
 	}
 
 	merged.AddPredecessors(bottom.Predecessors()...)
-	for i, pred := range merged.Predecessors() {
-		for _, succ := range pred.Successors() {
+	for _, pred := range merged.Predecessors() {
+		for i, succ := range pred.Successors() {
 			if succ == bottom {
 				pred.Successors()[i] = merged
 			}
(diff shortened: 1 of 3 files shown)


plan/types.go
```diff
@@ -265,8 +265,8 @@ func mergePlanNodes(top, bottom, merged Node) (Node, error) {
 	}
 
 	merged.AddPredecessors(bottom.Predecessors()...)
-	for i, pred := range merged.Predecessors() {
-		for _, succ := range pred.Successors() {
+	for _, pred := range merged.Predecessors() {
+		for i, succ := range pred.Successors() {
```
scbrickley, line 269: So the bug was caused by this code block using the wrong index? Do I have that right?

wolffcm: Yes, that's right. We are looking at the predecessors of the node being replaced, and updating them to point at the new node. But we were using the index from the list of predecessors (which we are not changing) instead of the index from the predecessors' successors.

scbrickley (approved):

PR #5116, opened 2022-08-19
refactor: move spec details into an internal package

jsternberg:
The spec was something that originally existed to accomodate
transpilers. We later decided transpilers should just use the AST rather
than something internal like the spec and the repl became the only thing
that uses the spec.

This moves the spec to an internal package so it stays as an internal
detail and does not exist as part of the external Go API.

This also removes the ider interface. It didn't appear to be used for
anything other than an implementation inside of the spec, so this just
removes it. The transformations that needed to know the ids of their
parents didn't use the operation ids anyway so it didn't matter.

### Checklist


skartikey (approved):
Looks good!

PR #5117, opened 2022-08-19
refactor: remove legacy map; add rule for vec'zd map

### LINKED ISSUE #4739: Add a way to assert about vectorized operations

wolffcm:
Fixes #4739.

This PR removes the `vectorizedMap` feature flag, which has been on for a few months in production. Note that there is a deprecated parameter called `mergeKey` that was honored in the legacy implementation of `map`. It is going away altogether now, so there is some behavior change and I have updated the related Go unit tests.

In addition I added a planner rule that will enable the vectorized function to be executed when one exists. This way we can use `expect.planner` to ensure that `vectorizeMapRule` fires, and `vectorizedMap` will appear as a plan node in the graphviz rendering of a plan.

### Checklist


stdlib/universe/map.go
```diff
@@ -97,159 +105,264 @@ func createMapTransformation(id execute.DatasetID, mode execute.AccumulationMode
...
-	fn := execute.NewRowMapFn(spec.Fn.Fn, compiler.ToScope(spec.Fn.Scope))
-	return &mapTransformation{
-		d:        d,
-		cache:    cache,
-		fn:       fn,
-		ctx:      ctx,
-		mergeKey: spec.MergeKey,
-	}, nil
-}
+func (m *mapTransformation) Process(
```
wolffcm, line 129: This looks like a big change, but all of this code here was copied from `map2.go`. I did not need to alter it.

stdlib/universe/map_vectorized.go
```diff
@@ -0,0 +1,250 @@
...
+	tr := &mapTransformation{
+		ctx: a.Context(),
+		fn: &mapVectorFunc{
+			fn: execute.NewVectorMapFn(s.Fn.Fn, compiler.ToScope(s.Fn.Scope)),
+		},
+	}
+	return execute.NewGroupTransformation(id, tr, a.Allocator())
+}
+
+type mapVectorFunc struct {
```
wolffcm, line 90: Somehow Git lost track of it, but I meant to just change the name `map2.go` to `map_vectorized.go`. All of the code for the rest of the file was in `map2.go`, and I didn't need to change it.

stdlib/universe/map.go
Marwes: Nit: If you delete the previous file first, then do `git mv` you can probably preserve the history better (might need to delete in one commit, move in the next).

Marwes (commented):
I assume that any breakages from removing `mergeKey` has already been discussed and deemed to be ok? LGTM otherwise

wolffcm:
> I assume that any breakages from removing mergeKey has already been discussed and deemed to be ok? LGTM otherwise

Yes, per our discussion this morning, `mergeKey` isn't going away as part of this PR, but I'll do it as a separate one.

stdlib/universe/map.go
wolffcm: Ah, I will definitely do that next time.

Marwes (approved):

wolffcm:
@theodiefenthal Hi Theo, there was an issue with `map` related to an optimization for if/then/else. We have since turned this optimization off in Cloud. Can you check and see if the issue still exists?

onelson:
We'll hopefully know more soon, but my thinking is, to echo @wolffcm, the issue has more to do with the implementation for vectorized conditional expressions themselves, not the manner in which we select.

Investigation will likely happen over on https://github.com/influxdata/flux/issues/5155

theodiefenthal:
Hi @wolffcm , the issue now disappeared for me, so turning it off in the cloud helped. Thank you. And my bad for adding this comment to the wrong ticket. I'll follow up on reading #5155.

PR #5118, opened 2022-08-20
fix: "invalid != vector" panic for conditionals with bad values for test, consequent or alternate

onelson:
Initial work for vectorized conditionals neglected to cover cases where
the inputs for `test`, `consequent`, and `alternate` could be invalid
(ie, referencing fields which could not be statically verified and
missing at runtime).

This diff aims to handle these cases by:
- treating a null `test` the same as "all false" by returning
  `alternate` directly.
- changing the constant (vec repeat) code paths to accept pointers to
  the various primitive types so they can be optionally null.
- short circuiting for cases where both `consequent` and `alternate` are
  null (since the output will be null, regardless of the values in `test`).

The choice to adjust the constant handling code path to also handle the
null case was largely motivated by wanting to avoid adding yet another
factor to drive combinations of function signatures.

This would have required a large expansion of functions in the `array`
package, as well as higher complexity to be able to delegate to the new
functions from the `values` side.

We started with needing (per type):
- vec, const
- const, vec
- vec, vec
- const, const

By going this way, we avoid adding (per type):
- vec, null
- null, vec
- const, null
- null, const

Furthermore, the logic required for nulls in either position is more or
less identical to the const pattern, with the addition of a null check
of the pointer.

Refs:
- <https://github.com/influxdata/flux/issues/4601>
- <https://github.com/influxdata/flux/pull/5017>

### Checklist


values/conditional.gen.go.tmpl
```diff
@@ -22,71 +22,89 @@ import (
...
+// invalid (aka null), we treat this as if we have a null constant for the given
+// branch. Items selected from the branch will produce nulls in the output.
+// For cases where both `c` and `a` are null, we short circuit immediately
+// returning a new untyped null Value.
+func VectorConditional(t Vector, c, a Value, mem memory.Allocator) (Value, error) {
+	// We will know the output type only when at least one branch is not null.
+	// When both branches are null, the output can only be null (aka Invalid).
+	var elemType semantic.MonoType
+	if c.IsNull() && a.IsNull() {
+		return New(nil), nil // early return an untyped null
```
onelson, line 36: I did notice there's a sentinel value, `values.Null`, which could be used here instead of going through the trouble of calling `New()`. Spotted this while I was auditing the code path on the binary side (for the recent equality ops PR).

The similar case for binary ops is handled here

https://github.com/influxdata/flux/blob/32fe18dbcdc6ea4a30cf467231c84cdbf8d1bd54/values/binary.go#L55-L62

nathanielc (approved):

PR #5120, opened 2022-08-22
fix: Don't treat errors in SQL syntax as internal

Marwes:
Since the query is written by a user, it is not an internal error and the user should get the error reported back to them so they may correct the issue.

Added `shouldErrorWithCode` to allow tests to check the error code and tweaked `catch` to supply the code.

`shouldErrorWithCode` should maybe be in internal or experimental instead? I don't think we want to expose the error code to users.

### Checklist


stdlib/sql/from.go
```diff
@@ -173,7 +173,7 @@ func (c *sqlIterator) Do(ctx context.Context, f func(flux.Table) error) error {
 
 	rows, err := db.QueryContext(ctx, c.spec.Query)
 	if err != nil {
-		return err
+		return errors.New(codes.Invalid, err)
```
jsternberg, line 176: I think `errors.Wrap(err, codes.Invalid)` is more appropriate.

jsternberg (changes requested):
One change and otherwise it looks good.

stdlib/sql/from.go
Marwes: Fixed

jsternberg (approved):

PR #5125, opened 2022-08-22
chore(docs): add params example to requests.post, fix example in monitor package

sanderson:
Closes https://github.com/influxdata/docs-v2/issues/4374
Closes https://github.com/influxdata/DAR/issues/320

Adds an example for `requests.post` that shows how to use `params`. Updates the example for `monitor.deadman` to use `date.add` with the correct parameters.

---

### Checklist


stdlib/http/requests/requests.flux
```diff
@@ -183,6 +183,19 @@ do = (
 // > array.from(rows: [data])
 // ```
 //
+// ### Make a POST request with query parameters
+//
+// ```
+// import "http/requests"
+//
+// response = requests.post(
```
onelson, line 191: I wonder if it would be overkill to add a comment inside the example to show the full url being generated: `http://example.com?start=100`

onelson (approved):

stdlib/http/requests/requests.flux
sanderson: Add the full URL in a comment. I also added a second param with multiple values to show how those get parsed into the URL.

onelson (approved):
looks great

PR #5130, opened 2022-08-25
feat: support vectorized unary operators

### LINKED ISSUE #5045: Vectorize unary Add, Sub, Exists, Not
This issue aims to vectorize _unary_ expressions:
- Add
- Sub
- Exists
- Not

If there are other unary operators that are out there (I'm not sure what they might be), use your discretion to decide if they should be lumped in with this issue or not.

> In #4608 we vectorized equality ops that are _binary_ since we already had a lot of code in place around vectorized binary expressions, but nothing for _unary_. Since _binary vs unary_ is a more dominant feature of the work than _arithmetic vs equality_ it makes some sense to tackle these all together.

DOD:
- unary Add, Sub, Exists, Not can be vectorized
- tests
  - tests added Rust to verify functions containing these _can be vectorized_
  - tests added to Go to verify the output from the new vectorized operator impls
  - flux acceptance tests added to verify type coverage for all types supporting the named ops
  - flux acceptance tests to verify vector repeat code paths work without panicking (ie, no illegal calls to `VectorRepeatValue.Arr()`)
- add a feature flag the new unary code path

onelson:
Fixes #5045

Adds support for vectorized unary operators: `add`, `sub`, `exists`, `not`.

Differences in the supported `Value` types between the regular unary and the vectorized fall along the lines of what can be represented as column data (which is the gateway to vectorized map) and what can be freely stored in arrow arrays.

The notable omission here is Duration, which is commonly used with unary sub, ex: `-1d`. Since this type cannot appear in tables, it cannot be mapped over using vectorization.

The new code path is guarded by a feature flag: `vectorizedUnaryOps`.

### Checklist


compiler/runtime.go
```diff
@@ -644,15 +644,91 @@ func (e *unaryEvaluator) Type() semantic.MonoType {
 	return e.t
 }
 
+func doUnary(mt semantic.MonoType, op ast.OperatorKind, v values.Value) (values.Value, error) {
```
onelson, line 647: Lightly modified - this was formerly a closure defined inside the unary evaluator. Extracted so it could be shared with the vector version for the vec repeat case.

libflux/flux-core/src/semantic/mod.rs
```diff
@@ -537,6 +537,10 @@ pub enum Feature {
     /// Enables calls to map to be vectorized when the function contains
     /// float casts.
     VectorizedFloat,
+
+    /// Enables calls to map to be vectorized when the function contains
+    /// unary operators like: add, sub exists, not.
+    VectorizedUnaryOps,
```
wolffcm, line 543: It might be good to start generating Rust code for features at some point. (just an observation)

array/unary.gen.go
```diff
@@ -0,0 +1,139 @@
...
+// Arithmetic Ops
+//
+
+func IntUnarySub(v *Int, mem memory.Allocator) (*Int, error) {
+	n := v.Len()
+	b := NewIntBuilder(mem)
+	b.Resize(n)
+	for i := 0; i < n; i++ {
+		if v.IsValid(i) {
+			b.Append(-v.Value(i))
```
wolffcm, line 23: I was curious what Go does if you try to negate `math.MinInt64`:
https://go.dev/play/p/h-QiLnQU8nz

I guess we inherit this behavior...

compiler/runtime.go
```diff
@@ -644,15 +644,91 @@ func (e *unaryEvaluator) Type() semantic.MonoType {
...
+
+	// If the value is null, return it immediately.
+	if v.IsNull() {
+		return v, nil
+	}
+
+	switch op {
+	case ast.AdditionOperator:
+		// Do nothing.
+		return v, nil
```
wolffcm, line 660: I always wonder why this operator exists... at least the implementation is efficient...

libflux/flux-core/src/semantic/tests/vectorize.rs
```diff
@@ -421,3 +422,48 @@ fn vectorize_with_float_calls() -> anyhow::Result<()> {
...
+#[test]
+fn vectorize_with_unary_add_sub() -> anyhow::Result<()> {
+    let pkg = vectorize(r#"(r) => ({ r with add: +r._value, sub: -r._value })"#).unwrap();
+
+    let function = get_vectorized_function(&pkg);
+
+    expect_test::expect![[r##"
+        (r) => {
+            return {r:{#C with _value: v[#B]} with add: +r:{#C with _value: v[#B]}._value:v[#B]:v[#B], sub: -r:{#C with _value: v[#B]}._value:v[#B]:v[#B]}:{#C with add: v[#B], sub: v[#B], _value: v[#B]}
+        }:(r: {#C with _value: v[#B]}) => {#C with add: v[#B], sub: v[#B], _value: v[#B]}"##]].assert_eq(&crate::semantic::formatter::format_node(
```
wolffcm, line 435: I'm curious how you like this way of verifying vectorized code, or if you think there's a better way.

wolffcm (approved):
Looks great.

compiler/runtime.go
onelson: Yeah, I mean... :shrug:

I'm just trying to stay faithful to the source material :rofl:

libflux/flux-core/src/semantic/mod.rs
onelson: Building the enum from the yaml file? Yeah, that's an interesting idea. It could be something we do via `build.rs`

libflux/flux-core/src/semantic/tests/vectorize.rs
onelson: I have a very hard time reading the string here, so I'm glad it's generated for me. I doubt I could write this by hand correctly without much trial and error.

For me, the exercise mostly comes down to making sure I see a `v[]` around either a scalar type or a type var in the right field positions. Other than that, the test gives me pass/fail on whether or not the function expression _can be vectorized_.

Seems sufficient.

Issue #4928, opened 2022-06-27
Expose label polymorphism via an opt-in feature


Marwes:
To let type signatures take advantage of label polymorphism without breaking backwards compatibility we will add an opt-in compilation option that tells the compiler to use the label polymorphic function signatures.

### How should the option be enabled

- Via an explicit option at the command line or GUI?
- Should we enable this option if the label syntax (`.mylabel`) is used in the script? ( https://github.com/influxdata/flux/issues/4927 )
- Something like javscripts `"use strict"` in the script itself?
- If the script uses the label syntax, auto enable the feature
- Explicit Flux major versions

Since the feature changes the pre-compiled standard library we must either compile it twice, or ensure that we retain both signatures in one pre-compilation and information on how to select which (the former is easier, but less efficient).

(Based on the comment https://github.com/influxdata/flux/pull/4776#issuecomment-1167298021 )

DOD
- When compiling we can select either the old or the new function signatures to use when typehecking

PR #5131, opened 2022-08-25
feat: Add a syntax for describing label literals

### LINKED ISSUE #4927: Add a syntax to describe label values
To allow label polymorphism to be used we need a way to define a label through the syntax. Originally we tried to re-use string literals for this, but during development we did not find a backwards compatible way that does not involve sub-typing.

Instead we add a new syntax `.mylabel` to define a label and users will need to update their query to use this new syntax. Additionally we need a syntax for labels with spaces or other non identifier characters. `."my label"`.

NOTE: We may find that `.` is ambiguous  with member expressions, if so we will need a new token, but it should be trivial to change the token used to delimit labels. We leave the actual token used up to the implementer.

For testing, there are ignored tests that live in `semantic::tests::labels` which should pass after updating them to use the new syntax.

DOD:
- A `.mylabel` syntax exists. Semantically it works the same as a string literal, it just has the label type instead of `string`
- The current way of enabling labels via string literals is removed
- The documentation is updated to explain this syntax
- Plumb labels from AST to Semantic nodes to semantic type
- Update applicable tests in `semantic::tests::labels`

Marwes:
Adds the syntax and locks it behind the `LabelPolymorphism` feature flag.
(#4928 exists as the next step to allow users to actually opt-in to it).

Closes #4927

### Checklist


Marwes:
ping @jsternberg

jsternberg (changes requested):
I think you need to add the AST changes to the Go side too.

Everything else looks fine. I also can't remember if labels need to be added to the Go semantic package.

jsternberg (approved):

PR #5133, opened 2022-08-25
test: Correct the value of _stop and now in shapeData

Marwes:
`_stop` were earlier than some of the values of `_time` and `now`, which is used as the default for `range` were different
from the `_stop` specified in the CSV files which causes differences when using `diff` to compare. Correcting and normalizing these
fixes the problem when testing against influxDB.

I am not quite sure why it still works when just running the test locally though. I guess `range` never updates `_stop` in that case so it is always the default?

### Checklist


stdlib/experimental/geo/shapeDataWithFilter_test.flux
```diff
@@ -5,101 +5,115 @@ import "experimental/geo"
...
+,,2,migration,2019-01-01T00:00:00Z,2019-01-11T00:00:00Z,2019-01-05T07:00:00Z,ctrlField,91916A,21.16,39.1665,15c3af
+,,2,migration,2019-01-01T00:00:00Z,2019-01-11T00:00:00Z,2019-01-05T13:00:00Z,ctrlField,91916A,21.16633,39.16817,15c3af
+,,2,migration,2019-01-01T00:00:00Z,2019-01-11T00:00:00Z,2019-01-05T19:00:00Z,ctrlField,91916A,21.17433,39.18417,15c3af
+,,3,migration,2019-01-01T00:00:00Z,2019-01-11T00:00:00Z,2019-01-02T07:00:00Z,ctrlField,91916A,21.2935,39.15833,15c3b1
+,,4,migration,2019-01-01T00:00:00Z,2019-01-11T00:00:00Z,2019-01-02T13:00:00Z,ctrlField,91916A,21.3485,39.15083,15c3b9
 "
 
 // Passes in flux, fails in C2 and OSS
-testcase shapeDataWithFilter {
+testcase shapeData {
```
onelson, line 111: Looks like CI is calling this a duplicate.

onelson (approved):
I don't really understand why we'd see a disparity between vanilla and influxdb for this, but lgtm :shrug:

PR #5135, opened 2022-08-26
feat: Compile the standard library incrementally

Marwes:
This adds https://github.com/salsa-rs/salsa as a dependency and defines the compiler using it. With salsa the dependencies between different parts of the program are modeled implicitly which reduces or eliminates the need to run steps manually ahead of time.

For instance, we currently read all source files and compile them into their AST's before we start on any semantic checks as most scripts only require a small part of the standard library this is a lot of wasted effort. With salsa and this PR we no longer need to generate the AST for everything upfront, instead we instead ask for the semantic package for a a specific package. If the package then depends on, say the `csv` package, we recursively compile `csv` and return it, letting `salsa` take care of caching the compilation of `csv`.

Further, salsa also keeps track of when to invalidate these cached results to reduce how much that needs to be recompiled when files change. However, since scripts tend to be just a single file this isn't very interesting to us with the LSP being the only (small) user of this. The LSP should be able to remove its bespoke cache invalidation that it has at least.

## Usecases for this PR

* Removing the constant overhead of running fluxtest. `fluxtest` currently parses all `_test.flux` files, converts them to the AST, marshals the AST to JSON, umarshals to the go AST structs, constructs new ASTs for each `testcase` and passes these ASTs back to rust for final compilation. Since the AST is such a big format, especially when serialized to JSON this takes a long time
Screenshot from 2022-08-26 13-24-55 . The main improvement here would really be to move most logic of `cmd/flux/cmd/test.go` into rust so we avoid this unnecessary `rust -> go -> rust` roundtrip (especially removing the JSON roundtrip that the go bits forces on us). Once most logic is on the rust side, the go side only needs to deal with parsing the command line arguments, passing that information to rust and finally get a list of tests back. `salsa` isn't directly needed to do this change, however without the back and forth we can get some more performance for free as we would only need to run semantic checks on the tests that aren't filtered out.

* Removing the rust `build.rs` file. Since salsa only compiles what is needed, we can likely remove the ahead of time compilation done in `build.rs` and instead embed the script files directly. While it is probably a bit slower to compile source code to semantic graph on a per-file basis compared to protobuf to semantic graph, the fact that we would only compile the parts of the standard library that is used would offset some of that. (The compiler is also much faster now compared to when the ahead of time compilation were added).

* User modules can't be ahead of time compiled. Since the flux compiler doesn't know about user modules it can't ahead of time compile them like it does with the standard library. Therefore we will need a new way to inject those into the compiler and ideally we should use the same mechanism for the standard library so we don't have two implementations of the same thing. We could also implement our own model for dependency resolution and adding in new packages however salsa is certainly a way, and it does give us the other features mentioned above as well.

* Cross-file testing. We currently do some [hacks](https://github.com/influxdata/flux/blob/2879293d67c3770b4e9eda8bbbe34c8a6d276ff2/libflux/flux-core/src/semantic/tests/vectorize.rs#L23-L31) to "import" things from the standard library such that we can test against them. With salsa (or with a custom implementation with an API like salsa) we could instead point the tests towards the actual standard library and have `import` statements just work.

So some of these could definitely be implemented without salsa as we do not really need the incremental recompilation bits. However if salsa isn't seen as to complex of a dependency I think it still gives us quite a bit of nice things "for free", Note that the PR as is is quite hacky as I just wanted to get something working. This should be cleaned up before this could be merged.

- [x] There is a pre-release of the next version of salsa which we may want to update to https://smallcultfollowing.com/babysteps/blog/2022/08/18/come-contribute-to-salsa-2022/
- [x] Error handling is a bit icky in the (stable) version of salsa that I used. The pre-release seems to have gotten diagnostic support which we should probably use https://salsa-rs.github.io/salsa/overview.html#accumulators
- [ ] Lazily load the source files instead of requiring them to be supplied via `set_source` https://salsa-rs.github.io/salsa/common_patterns/on_demand_inputs.html
- [ ]  Get rid of the mess of `_inner` functions. Salsa doesn't provide a way to check if a there is a value behind a key or not by default, I added it in a fork I have used so I'd either port that or check how other users deal with invalid/missing imports.
- [x] Remove `NeverEq`. This is a hack to workaround `salsa`s `Eq` requirement and get the behaviour of "recompile only if the dependencies, ie source code, changes), should be a pretty easy addition to add custom handling of `Eq` to salsa though https://github.com/salsa-rs/salsa/issues/218 .
- [x] Remove the need for the `packages` list. Salsa doesn't provide a way to list the keys of query so I had to add a custom `set_source` and add the packages there). Should be a pretty easy addition I think.

### Checklist


nathanielc:
Re the refactor to testing so we don't have back and forth between Rust and Go. I think this also moves us in the direction where we could run tests as a first class behavior of the runtime instead of as an add-on like it currently is. The movitivation is as users write their own modules they will want to maintain tests for their code. We can't expect users to use the CLI with specialized setup to run those tests. Instead the `testcase` statements should work as of the normal compilation/execution path. (i.e. cloud users can submit testcases to be run via the API).  It seems like this change would move us closer to this goal.


> User modules can't be ahead of time compiled. Since the flux compiler doesn't know about user modules it can't ahead of time compile them like it does with the standard library. Therefore we will need a new way to inject those into the compiler and ideally we should use the same mechanism for the standard library so we don't have two implementations of the same thing. 

I was just discussing this with someone this week (can't remember who, maybe @onelson ?) Anyway this valuable to me, I think a simpler more on demand compilation of packages makes a lot of sense in the face of user modules.

Would the salsa cache be cross request? Meaning if one query comes in that uses packages X and Y from the stdlib and then later another query arrives that uses package Y from the stdlib would we need to recompile Y? This likely adds extra complexity since now we have to make sure to model all the different requirements that each different query request could bring.


Finally would using salsa also enable us to different compilation targets more transparently? For example say in the future we have a `label polymorphism` feature of the compiler, can we teach salsa about that such that it can compile and cache package A without that feature and package A with that feature?

Marwes:
> Would the salsa cache be cross request?

The cache just lives in memory inside the `Database` object so that depends on its lifetime.

> Meaning if one query comes in that uses packages X and Y from the stdlib and then later another query arrives that uses package Y from the stdlib would we need to recompile Y? This likely adds extra complexity since now we have to make sure to model all the different requirements that each different query request could bring.

A salsa query should be deterministic and only depend on the inputs given to it. So if the queries have different "requirements" that should be reflected in the salsa queries, invalidating or recording a new entry in the cache. Barring any bugs, it thus shouldn't matter that the queries are different.

> Finally would using salsa also enable us to different compilation targets more transparently?

Yes, as the PR is right now `AnalyzerConfig`, which contains the feature flags is an input query. So if the feature flags change you would then call `set_analyzer_config` which would invalidate the cache for all queries that read the `AnalyzerConfig` (read the feature flags) and any new query following that would be re-run with the new features.

Alternatively, `AnalyzerConfig` could be passed as an argument to the queries in which case the cache would be keyed by `AnalyzerConfig` and each `module path`.

libflux/flux-core/src/db.rs
```diff
@@ -0,0 +1,486 @@
...
+
+fn semantic_package(
+    db: &dyn Flux,
+    path: String,
+) -> SalvageResult<(Arc<PackageExports>, Arc<nodes::Package>), Error> {
+    // The previous standard library compiler happened to result in the prelude being incrementally
+    // added to with later packages in the prelude depending on earlier ones. This was mostly
+    // arbitrary and we should try to encode these dependencies more deliberately but these stages
+    // of no/internal/full prelude seem to do the trick in getting things to work.
+    let prelude = if !db.use_prelude() || INTERNAL_PRELUDE.contains(&&path[..]) {
```
jsternberg, line 346: Fair enough. The previous method was kind of a hack. This seems fine.

jsternberg (approved):
This overall looks good. I didn't see anything I could identify as red flags, but this code is fairly large and I'd have a difficult time reviewing it with a fine-toothed comb.

What's the risk profile for merging this change? I don't think this is something that's practical to gate behind a flag, but I'd like to understand the risk of this running in production with the only testing being integration tests.

Marwes:
> What's the risk profile for merging this change? I don't think this is something that's practical to gate behind a flag, but I'd like to understand the risk of this running in production with the only testing being integration tests.

Two things I can think of is that we cache more things now (AST, source code,...) so there may be some higher peaks in memory usage. However given that the standard library is still precompiled that *should* just be for the actual query script making it an extra constant cost. We could tell salsa to clear those bits out a bit more eagerly but I doubt this extra (constant) cost will break things.

The other is that if we have a bug in dependency resolution we may end up recursing and causing new panics, however that should all be guarded against.

Marwes:
Added a feature flag for using the salsa database in production (still using it by default for generating the precompiled stdlib).

Also fixed it so we actually use the salsa database with the `StatefulAnalyzer` which I had missed

jsternberg (approved):

PR #5137, opened 2022-08-26
test: Don't run fill tests against external storage

Marwes:
Since null values will be ignored (correct me if I am wrong) when writing to storage these tests do not pass
when using external storage. I can't think of a way to fix them so just skipping external storage for these might be the best option.

### Checklist


onelson (approved):
> Since null values will be ignored (correct me if I am wrong) when writing to storage

This sounds right, but I'm not certain. If true, I can't think of an approach where `testing.load()` helps or makes sense.

At the very least, these tests will still verify the flux-local side of the operation so that's good.

PR #5147, opened 2022-08-30
build: extend deadline for windows builds

onelson:
We see windows builds timeout all the time.
CircleCI has a deadline for how long a job can run without printing
anything. The default is 10m, so this diff doubles it at 20m.

Hopefully we will waste fewer CI cycles this way by reducing the number
of restarted jobs required for a green run.


wolffcm (approved):
Thanks for this!

PR #5149, opened 2022-08-30
feat: Allow unpivot to work when the time field is missing 

Marwes:
When pushing down aggregate functions the time field is not included in the output so we need to tweak unpivot to work without the time column. Changing the order seemed like the simplest solution however that exposed a bug in the current implementation where the full `time` column were copied over as any nulls only existed in the value column

### Checklist

### DIFF
--- a/array/array_test.go
+++ b/array/array_test.go
@@ -346,9 +346,10 @@ func TestSlice(t *testing.T) {
 
 func TestCopyValid(t *testing.T) {
 	for _, tc := range []struct {
-		name  string
-		input func(memory.Allocator) *array.Int
-		want  []int64
+		name       string
+		input      func(memory.Allocator) *array.Int
+		nullBitmap func(memory.Allocator) *array.Int
+		want       []interface{}
 	}{
 		{
 			name: "nulls",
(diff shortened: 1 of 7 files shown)


array/builder.gen.go
```diff
@@ -71,10 +71,16 @@ func (b *IntBuilder) NewArray() Array {
...
+		panic("Length mismatch between the value array and the null check array")
+	}
+	b.Reserve(values.Len() - nullCheckArray.NullN())
+
+	nullBitMapBytes := nullCheckArray.NullBitmapBytes()
+	nullOffset := nullCheckArray.Data().Offset()
 	for i := 0; i < values.Len(); i++ {
-		if values.IsValid(i) {
+		if isValid(nullBitMapBytes, nullOffset, i) {
 			b.Append(values.Value(i))
```
wolffcm, line 84: I don't think this is quite right, since you want to call `appendNull` here if the values array contains a `null`. What you have here will append a (valid) arbitrary value (probably the zero value).
```suggestion
			if values.IsValid(i) {
			    b.Append(values.Value(i))
			} else {
			    b.AppendNull()
			}
```

stdlib/experimental/unpivot.go
```diff
@@ -150,12 +148,17 @@ func (t *unpivotTransformation) Process(chunk table.Chunk, d *execute.TransportD
 			buffer.Values[toColumn] = values
 		}
 
-		buffer.Values[len(buffer.Values)-3] = array.StringRepeat(c.Label, newChunkLen, mem)
+		oldValues := chunk.Values(i)
+
+		if timeColumn != -1 {
+			// The time array does not contain any nulls so we use the value array for that information
+			times := array.CopyValidValues(mem, chunk.Values(timeColumn), oldValues)
+			buffer.Values[len(buffer.Values)-3] = times
```
wolffcm, line 156: I wonder if this fix here will address the bug that @scbrickley found with the new `testing.diff`... and this also explains why it seemed to be specific to IOx

array/array_test.go
```diff
@@ -391,7 +391,7 @@ func TestCopyValid(t *testing.T) {
 
 			input := tc.input(mem)
 
-			got := array.CopyValidValues(mem, input).(*array.Int)
+			got := array.CopyValidValues(mem, input, input).(*array.Int)
```
wolffcm, line 394: It would be good to add some additional tests here for when the null check array and the values array do not have nulls in the same places.

wolffcm (changes requested):
I agree your implementation will be faster than my attempt. Why don't you merge this first, and then I will rebase #5154 on top of this? I think it just needs to take into account nulls in the value column.

Incidentally, I think the work in #5154 is a generalization of how we are treating the `_time` column now. Sometimes it's just the `_time` column that needs to be copied into the output (despite not being a tag) and sometimes its addition columns as well.

array/builder.gen.go
Marwes: Yeah, I were just thinking of the time column which never has any nulls so I didn't handle that case.

array/builder.go
```diff
@@ -1,6 +1,7 @@
 package array
 
 import (
+	"github.com/apache/arrow/go/arrow/bitutil"
```
wolffcm, line 4: Curious if you needed to import this module without the `v7`?

go.mod
```diff
@@ -62,6 +62,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+require github.com/apache/arrow/go/arrow v0.0.0-20211112161151-bc219186db40
```
wolffcm, line 65: I don't think you wanted to import this (pre-1.0) older version of Arrow. We have it included here because a dependency of Flux needs it, but any new code should use the `v7` import path:
https://github.com/influxdata/flux/blob/6013690a714413da6eb3e44fedfce4d17971d1e3/go.mod#L15

wolffcm (changes requested):
I just think the import for the bitutils needs to use the `v7` path, but this looks good otherwise.

array/builder.go
Marwes: That was just autogenerated via the LSP , fixe

wolffcm (approved):

wolffcm:
Gonna merge this, because it looks good and I would like to build upon it.

PR #5152, opened 2022-08-31
chore: Don't report skips in fluxtest

Marwes:
Skips tend to bloat the test output and since tests are always skipped through explicit opt-in behavior I don't think it is very valuable.

### Checklist


nathanielc (approved):

onelson:
Couldn't quite tell from the diff -- are there still situations where the skipped tests _will be reported_? Looks like we'd just never report them moving forward (except for the count in the summary).

Marwes:
Individual skips never get reported now, just the count in the summary

onelson:
Got it. Sounds good to me.

I had a question the other day about if a test was still failing (and could an issue be closed now). My first thought was to look to see if the test was skipped, but in reality I got my answer from the fact _it wasn't_.

PR #5154, opened 2022-08-31
feat: add parameter to `unpivot` for non-field non-GK columns

wolffcm:
This PR is updates the `unpivot` Flux function to take an additional parameter called `ungroupedTagColumns` to be used as part of IOx SQL support in cloud. The idea behind this parameter is that there can be some columns that are not in the group key (because of a preceding `group` transformation). These columns should be specified when calling `unpivot` so it doesn't create `_field` and `_value` columns in the output for them.

In order to support this work I added a new array function called `CopyValuesWithMask`. As it turns out this function needs to be called for the `_time` column as well. This was a bug that I found in my testing.

### Checklist


array/builder.gen.go
```diff
@@ -79,6 +79,18 @@ func (b *IntBuilder) CopyValidValues(values *Int) {
 		}
 	}
 }
+func (b *IntBuilder) CopyValuesWithMask(values *Int, mask Array) {
+	b.Reserve(mask.Len() - mask.NullN())
+	for i := 0; i < values.Len(); i++ {
+		if mask.IsValid(i) {
```
Marwes, line 85: Seems we had the same issue though I tweaked the existing `CopyValidValues` function in https://github.com/influxdata/flux/pull/5149 . I also ensured that `mask.IsValid` does not go through in indirect call since this code is pretty hot.

sanderson (approved):
One minor suggestion to the inline docs.

wolffcm:
I've pushed another commit that:
- Rebases on master which includes @Marwes's PR #5149 
- Generalizes the columns parameter to be called `otherColumns` which defaults to `["_time"]`. So `_time` is not special-cased anymore, it's just like any other column that needs to "come along for ride".  

@Marwes I guess this means that pushdown code for aggregates will need to explicitly remove `_time` from `otherColumns` in a planner rule. Does that work for you?

Marwes:
> Does that work for you?

I think that would work. I guess the group pushdown will have to alter `otherColumns` in a similar manner?

wolffcm:
> I guess the group pushdown will have to alter `otherColumns` in a similar manner?

That's right. When we push down group we need a way to tell `unpivot` which columns are fields and which are not, and that is the motivation for this change.

Marwes (approved):

PR #5159, opened 2022-09-01
fix: vectorized division by zero should error not panic

onelson:
In the row-based division, we check to see if the divisor is 0 and return an error when this is the case.

This check was missed in the vectorized division, leading to a panic.

This diff adds the check for division for uint and int. For whatever reason, zero division with float produces an `Inf+` and there is no such guard in the row-based version so no updates were made there on the vectorized side.

### Checklist

### DIFF
--- a/array/binary.gen.go.tmpl
+++ b/array/binary.gen.go.tmpl
@@ -88,6 +88,12 @@ func {{$type}}{{$op.Name}}(l, r *{{$type}}, mem memory.Allocator) (*{{$type}}, e
 
             {{else}}
 
+            {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
+			if r.Value(i) == 0 {
+				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			}
+			{{end}}
+
 			b.Append(l.Value(i) {{$op.Op}} r.Value(i))
 
             {{end}}
@@ -112,6 +118,12 @@ func {{$type}}{{$op.Name}}LConst(l {{index $.TypeMap $type}}, r *{{$type}}, mem 
 
             {{else}}
 
+            {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
+			if r.Value(i) == 0 {
+				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			}
+			{{end}}
+
 			b.Append(l {{$op.Op}} r.Value(i))
 
             {{end}}
@@ -135,6 +147,12 @@ func {{$type}}{{$op.Name}}RConst(l *{{$type}}, r {{index $.TypeMap $type}}, mem 
 			b.Append(math.Mod(l.Value(i), r))
 
             {{else}}
+
+            {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
+			if r == 0 {
+				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			}
+			{{end}}
 
 			b.Append(l.Value(i) {{$op.Op}} r)
 
--- a/array/binary.gen.go
+++ b/array/binary.gen.go
@@ -598,6 +598,10 @@ func IntDiv(l, r *Int, mem memory.Allocator) (*Int, error) {
 	for i := 0; i < n; i++ {
 		if l.IsValid(i) && r.IsValid(i) {
 
+			if r.Value(i) == 0 {
+				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			}
+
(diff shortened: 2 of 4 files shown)


array/binary.gen.go.tmpl
```diff
@@ -136,6 +148,12 @@ func {{$type}}{{$op.Name}}RConst(l *{{$type}}, r {{index $.TypeMap $type}}, mem
 
             {{else}}
 
+            {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
```
skartikey, line 151: These 3 blocks of code are doing the same thing of checking value equal to zero. Can we move it to a method in the template, that can be named like `safeDivide()` or similar?

skartikey (changes requested):

array/binary.gen.go.tmpl
onelson: I'll look into the options.

I'm a little reluctant since we'd need either a `safeDivide()` per type applicable type, or maybe do it with generics.

The extra overhead for the function call is probably not a huge concern, but certainly it'd be cheaper to not introduce it.

Maybe there's something that could be done with the template lang to DRY it up and keep the generated Go source as it is now.

onelson: I did a version of this using generics, but we could just as easily skip them since there are only two types of concern.

Mostly I don't think it's worth it to make this change since we're just trading the `if r == 0` early return for a similar check for the error.

```diff
diff --git a/array/binary.gen.go.tmpl b/array/binary.gen.go.tmpl
index e39d2c53..db386d0b 100644
--- a/array/binary.gen.go.tmpl
+++ b/array/binary.gen.go.tmpl
@@ -89,12 +89,15 @@ func {{$type}}{{$op.Name}}(l, r *{{$type}}, mem memory.Allocator) (*{{$type}}, e
             {{else}}
 
             {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
-			if r.Value(i) == 0 {
-				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			x, err := safeDivide(l.Value(i), r.Value(i))
+			if err != nil {
+				return nil, err
 			}
+			{{else}}
+			x := l.Value(i) {{$op.Op}} r.Value(i)
 			{{end}}
 
-			b.Append(l.Value(i) {{$op.Op}} r.Value(i))
+			b.Append(x)
 
             {{end}}
 		} else {
@@ -119,12 +122,15 @@ func {{$type}}{{$op.Name}}LConst(l {{index $.TypeMap $type}}, r *{{$type}}, mem
             {{else}}
 
             {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
-			if r.Value(i) == 0 {
-				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			x, err := safeDivide(l, r.Value(i))
+			if err != nil {
+				return nil, err
 			}
+			{{else}}
+			x := l {{$op.Op}} r.Value(i)
 			{{end}}
 
-			b.Append(l {{$op.Op}} r.Value(i))
+			b.Append(x)
 
             {{end}}
 		} else {
@@ -149,12 +155,15 @@ func {{$type}}{{$op.Name}}RConst(l *{{$type}}, r {{index $.TypeMap $type}}, mem
             {{else}}
 
             {{if and (eq $op.Op "/") (eq $type "Int" "Uint")}}
-			if r == 0 {
-				return nil, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+			x, err := safeDivide(l.Value(i), r)
+			if err != nil {
+				return nil, err
 			}
+			{{else}}
+			x := l.Value(i) {{$op.Op}} r
 			{{end}}
 
-			b.Append(l.Value(i) {{$op.Op}} r)
+			b.Append(x)
 
             {{end}}
 		} else {
diff --git a/array/binary.go b/array/binary.go
index 84cfe8f3..b645d187 100644
--- a/array/binary.go
+++ b/array/binary.go
@@ -109,3 +109,10 @@ func OrRConst(l *Boolean, r bool, mem memory.Allocator) (*Boolean, error) {
 	b.Release()
 	return a, nil
 }
+
+func safeDivide[T int64 | uint64](l, r T) (T, error) {
+	if r == 0 {
+		return 0, errors.Newf(codes.FailedPrecondition, "cannot divide by zero")
+	}
+	return l / r, nil
+}
```

Let me know what you want to see. I'm open to making this change but I can't say for sure it's much better/different.

skartikey: Yeah, it makes sense not to do generics just for 2 types. Also, like you said there's no point moving 1 check to the method and adding another error check at all the calls. Let's keep it as it is. I will approve the PR. 
Thanks for looking into it.

skartikey (approved):

Issue #5155, opened 2022-09-01
v 0.181.0  compound logical expressions in map failing 


anthony-roy-wesdigital:
**Steps to reproduce:**  
since v 0.1.8.1  rolled out to influxDBcloud ,  queries like the below are failing. 
with error :
*runtime error: cannot use operand of type vector with logical and; expected boolean

```
import "array"

rows = [
{Brake_Code_3: 1, Brake_Code_2: 0, Brake_Code_1: 1 ,BrakeEmergency:0},
{Brake_Code_3: 1, Brake_Code_2: 0, Brake_Code_1: 1 ,BrakeEmergency:0},
{Brake_Code_3: 1, Brake_Code_2: 0, Brake_Code_1: 1 ,BrakeEmergency:0},
{Brake_Code_3: 1, Brake_Code_2: 0, Brake_Code_1: 1 ,BrakeEmergency:0},
{Brake_Code_3: 1, Brake_Code_2: 0, Brake_Code_1: 1 ,BrakeEmergency:0},
]

data =  array.from(rows: rows)
data  |> map (fn: (r) => ( { r  with  
    brakeStep: 
      if  r.BrakeEmergency == 1 then 4.0 
      else if r.Brake_Code_3 ==1  and  r.Brake_Code_2 == 1  and r.Brake_Code_1 == 1 then 0.0 
      else if r.Brake_Code_3 == 1 and  r.Brake_Code_2 == 1 and r.Brake_Code_1 == 0 then 1.0
      else if r.Brake_Code_3 == 1 and  r.Brake_Code_2 == 0 and r.Brake_Code_1 == 1 then 2.0
      else if r.Brake_Code_3 == 0 and  r.Brake_Code_2 == 0 and r.Brake_Code_1 == 0 then 3.0
      else  0.0
   }) ) 
  
```

**Expected behavior:** Query should run 

**Actual behavior:**    
runtime error: cannot use operand of type vector with logical and; expected boolean

**Environment info:**

**Config:** eu-central-1-1.aws.cloud2.influxdata.com
Flux Version : v0.181.0

onelson:
@anthony-roy-wesdigital thanks for the detailed bug report. I included a similar program as was posted here in the test suite along with the fix in #5160.

For the time being, vectorization of `map()` with conditional expressions will not be attempted in cloud, but the expectation is once this fix is rolled out we should be able to re-enable the optimization.

PR #5160, opened 2022-09-01
fix: correct inconsistent runtime typing for `logicalVectorEvaluator`

onelson:
Fixes https://github.com/influxdata/flux/issues/5155

...which shows a runtime error when nesting a logical expression inside a conditional expression.

When in the `test` position, we'd get:
```
cannot use test of type vector in conditional expression; expected boolean
```

Originally, `logicalVectorEvaluator` advertised a runtime type of bool, which is inconsistent with reality since it is actually a vector of bool.

The error message we saw in this situation stemmed from the fact we choose either the vectorized or row-based conditional evaluator with this check here:
https://github.com/influxdata/flux/blob/df2b4c404e81d39cadef7456c4a3aacf55d3638a/compiler/compiler.go#L657-L669

Since `test` in this case was a `logicalVectorEvaluator` with incorrect (bool) typing, this check failed and we fall back to the non-vectorized version for conditional.
This is why the error message states "expected boolean" -- we were calling the non-vectorized version with a vector as input.

This diff updates the impl for `logicalVectorEvaluator.Type()` such that it reports a vector of bool instead of bool.

### Checklist

### DIFF
--- a/libflux/flux-core/src/semantic/tests/vectorize.rs
+++ b/libflux/flux-core/src/semantic/tests/vectorize.rs
@@ -467,3 +467,18 @@ fn vectorize_with_unary_exists() -> anyhow::Result<()> {
     )?);
     Ok(())
 }
+
+#[test]
+fn vectorize_with_conditional_nested_logical() -> anyhow::Result<()> {
+    let pkg = vectorize(r#"(r) => ({x: if r.a and r.b then r.x else r.y})"#).unwrap();
+
+    let function = get_vectorized_function(&pkg);
+
+    expect_test::expect![[r##"
+        (r) => {
+            return {x: (if r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.a:v[bool] and:bool r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.b:v[bool] then r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.x:v[#K] else r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.y:v[#K]):v[#G]}:{x: v[#G]}
+        }:(r: {#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}) => {x: v[#G]}"##]].assert_eq(&crate::semantic::formatter::format_node(
+        Node::FunctionExpr(function),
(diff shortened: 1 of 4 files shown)


libflux/flux-core/src/semantic/tests/vectorize.rs
```diff
@@ -467,3 +467,18 @@ fn vectorize_with_unary_exists() -> anyhow::Result<()> {
...
+#[test]
+fn vectorize_with_conditional_nested_logical() -> anyhow::Result<()> {
+    let pkg = vectorize(r#"(r) => ({x: if r.a and r.b then r.x else r.y})"#).unwrap();
+
+    let function = get_vectorized_function(&pkg);
+
+    expect_test::expect![[r##"
+        (r) => {
+            return {x: (if r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.a:v[bool] and:bool r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.b:v[bool] then r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.x:v[#K] else r:{#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}.y:v[#K]):v[#G]}:{x: v[#G]}
+        }:(r: {#O with a: v[bool], b: v[bool], x: v[#K], y: v[#K]}) => {x: v[#G]}"##]].assert_eq(&crate::semantic::formatter::format_node(
```
onelson, line 480: The typing for the logical and conditional exprs here on the rust side look good. The runtime doesn't appear to rely on this info in this case, it seems.

stdlib/universe/map_vectorize_conditionals_test.flux
```diff
@@ -245,3 +245,21 @@ testcase vec_conditional_null_test_consequent_alternate {
 
     testing.diff(want: want, got: got)
 }
+
+testcase vec_nested_logical_conditional_repro {
+    // Using a logical expr in the test of the conditional expr produced a
+    // runtime error:
+    // ```
+    // cannot use test of type vector in conditional expression; expected boolean
+    // ```
```
onelson, line 254: In theory, I think we might have seen this for any nested expression where logical ops were children of a larger expression.

nathanielc (approved):

PR #5163, opened 2022-09-02
feat: Don't display nulls as 0 in diff

Marwes:
Before they would be displayed as the zero-value of whatever type the column is which is rather confusing. Now it is left empty instead (or should we display `<nil>` or something instead?)

### Checklist


wolffcm (approved):
Do we have any acceptance tests for `diff`? Maybe not I guess.

Thanks for doing this!

PR #5165, opened 2022-09-02
feat(array): add support to pipe into array.from

nathanielc:
With this change you can now pipe the `rows` argument into the
array.from function.

### Checklist


stdlib/array/array.flux
```diff
@@ -45,7 +45,7 @@ package array
 //
 // ## Metadata
 // tags: inputs
-builtin from : (rows: [A]) => stream[A] where A: Record
+builtin from : (<-rows: [A]) => stream[A] where A: Record
```
jsternberg, line 48: I'm not sure I'm the biggest fan of this. I can't really think of another example where we do this kind of thing. I saw the PR and thought I'd like it, but the test cases are a bit hard for me to read.

Is there a specific request asking for this or is this just a quality of life change? I'd like to get some opinions from others on the team if they feel it looks right before clicking approve. If others like it, I've got no problem with hitting approve from a functional standpoint.

sanderson: @jsternberg This is coming from me as I prepare content for the rollout of dynamic types. It just removes the need to assign an array to a variable to render it as a table when parsing JSON using the dynamic package:

```js
import "array"
import "experimental/dynamic"

jsonData = bytes(v: "
    {
    \"items\":  [
        {\"name\": \"a\", \"pos\": {\"x\": 10, \"y\": 10, \"z\": 20}},
        {\"name\": \"b\", \"pos\": {\"x\": 30, \"y\": 99}},
        {\"name\": \"c\", \"pos\": {}},
        {\"name\": \"d\"}
        ]
    }
")

parsed = dynamic.parseJSON(data: jsonData)

parsed.items
    |> dynamic.asArray()
    |> array.map(
        fn: (x) => (
            {
                name: string(v: x.name),
                posX: int(v: x.pos.x),
                posY: int(v: x.pos.y)
            }
        )
    )
    |> array.from()
```

jsternberg: Ok sure. I'll just say I discourage people from using it this way outside of this context.

jsternberg (approved):

sanderson (changes requested):
Just one additional update. Need to add this to the parameter description: 

```
Default is piped-forward data (`<-`).
```

stdlib/array/array.flux
nathanielc: Yes with the `json` use case its a bit awkward when every other function has a pipeforward argument.


For example:

```
dynamic.parseJSON(...)
    |> dynamic.asArray()
    |> array.filter(fn: (x) => x.bar == 42)
    |> array.map(fn: (x) => ({foo: x.foo}))
    |> array.from()
    |> to()
    
```

Without this change users must do:

```
rows = dynamic.parseJSON(...)
    |> dynamic.asArray()
    |> array.filter(fn: (x) => x.bar == 42)
    |> array.map(fn: (x) => ({foo: x.foo}))
    
array.from(rows: rows)
    |> to()
```

nathanielc: What @sanderson said. My github hadn't refreshed with the above comments before writing my reply.

sanderson (approved):

PR #5171, opened 2022-09-06
fix(join): better error when joining empty table

### LINKED ISSUE #5134: join.tables() method: "full" with an empty table returns "table is missing label _value"
`join.tables()` `method: "full"` with an empty table returns "table is missing label _value"

```js
import "array"
import "join"

left = array.from(rows: [{_value: 1, id: "a"}])

right = array.from(rows: [{_value: 0.6, id: "b"}])
          |> filter(fn: (r) => r.id == "empty table")

join.tables(
    method: "full",
    left: left,
    right: right,
    on: (l, r) => l.id == r.id,
    as: (l, r) => {
        id = if exists l.id then l.id else r.id

        return {id: id, v_left: l._value, v_right: r._value}
    },
)
```

I tested on InfluxDB Cloud v0.180.1

cc @scbrickley

scbrickley:
Fixes #5134 

### Checklist


Marwes (approved):
That is a better error message, but shouldn't joining with an empty table just output an empty table? No need to actually error here?

Marwes:
I guess we can't construct the type for the output if we try to construct an empty table? (Which is annoying and should be fixed, it is just a bit of a mess)

scbrickley:
Hmmm, that's a fair point. Maybe I can catch this condition earlier in the transformation and output an empty table before we ever check the column types. I'll tinker with it and get back to you.

scbrickley:
> I guess we can't construct the type for the output if we try to construct an empty table? (Which is annoying and should be fixed, it is just a bit of a mess)

This is the main blocker here. I'm going to merge this PR as is and we can discuss better solutions another time.

PR #5172, opened 2022-09-06
test: fixups for skipped group tests

onelson:
Fixes #15527

Tweaks to the following tests have them passing for all environments:
- [x] planner.group_one_tag_first
- [x] planner.group_all_filter_field_first
- [x] planner.group_one_tag_filter_field_first
- [x] planner.group_one_tag_last
- [x] planner.group_all_filter_field_last
- [x] planner.group_one_tag_filter_field_last
- [x] universe.group_except
- [x] universe.group
- [x] universe.group_ungroup

Still a problem:
- [ ] universe.group_nulls

In many cases, these tests were flaky because of non-deterministic row ordering when reading from storage whereas vanilla flux is able to rely on the ordering seen in the csv or array input.

As such, the fixes came in the form of adding sorts to make the ordering consistent from run to run, or to carefully alter the input data so the operation being tested can be more deterministic.

---

~~Currently, `universe.group_nulls` fails in all envs. I think the best thing to do will be to rewrite the test from scratch, with a more minimal input. Right now it's just hard to even make sense of what is or isn't aligned in the diff. I'm also not sure if the output I'm seeing is reliable since nulls specifically have been seen to report in unexpected ways recently.~~

The `universe.group_nulls` testcase has been greatly simplified and now passes in vanilla, but still needs to be skipped for C2/OSS until #5178 can be solved.

With the original version of this test, the csv data included nulls in many positions which was problematic for storage. Writes where `_time`, `_value` and so on where null seem to be dropped, so naturally we didn't see them during subsequent reading. The simplified version in this PR focuses the core case of what happens when you group on a tag which is sometimes null.

---

Another noteworthy test in this set is `universe.group` which was "fixed" by adding `_stop` to the columns to group by. It seems that `_stop` is retained in the group when we read from storage in C2, so there could be a bug we need to fix. Possibly this is an interaction with pushdowns - TBD.
The "fix" in this diff makes the test consistent with what we see when storage is in play.

---

Follow up issues filed to look at each of these two exceptions:

- https://github.com/influxdata/flux/issues/5173
- https://github.com/influxdata/flux/issues/5178

### Checklist


stdlib/planner/group_first_last_test.flux
```diff
@@ -235,13 +234,15 @@ testcase group_one_tag_first {
         testing.load(tables: inData)
             |> range(start: -100y)
             |> group(columns: ["t0"])
+            // Sort to make comparison more reliable when using storage (where
+            // row ordering can vary).
+            |> sort(columns: ["_time", "_field", "t0", "t1"])
```
onelson, line 239: I'm still not clear if the sort added to the tests in this file will impact the pushdowns applied.

I've got a concern about losing coverage, but the tradeoff here is we'd need to rework the `inData` more deliberately to give more deterministic answers when storage is used. Since `inData` is shared by _all the tests_ in this file, I opted to _not do that_ for this PR.

stdlib/universe/group_test.flux
```diff
@@ -25,24 +25,29 @@ inData =
...
             |> testing.load()
             |> range(start: 2018-05-22T19:53:26Z)
             |> filter(fn: (r) => r._measurement == "diskio" and r._field == "io_time")
-            |> group(columns: ["_measurement", "_start", "name"])
+            // XXX: C2 keeps `_stop` in the group key even when not specified here.
+            // Adding it to the group columns here, and in `outData` makes vanilla
+            // and C2 match.
+            // Could be a bug? Should flux not be able to unset `_stop` in the
+            // group key when we read from storage? If so, maybe add a separate test.
+            |> group(columns: ["_measurement", "_start", "_stop", "name"])
```
onelson, line 50: Follow-up on this filed here: https://github.com/influxdata/flux/issues/5173

stdlib/universe/group_except_test.flux
```diff
@@ -21,9 +21,9 @@ inData =
 ,,1,2018-05-22T19:53:26Z,648,io_time,diskio2,host.local,disk2
 ,,1,2018-05-22T19:53:36Z,648,io_time,diskio2,host.local,disk2
 ,,1,2018-05-22T19:53:46Z,648,io_time,diskio3,host.local,disk2
-,,1,2018-05-22T19:53:56Z,648,io_time,diskio3,host.local,disk2
-,,1,2018-05-22T19:54:06Z,648,io_time,diskio3,host.local,disk2
-,,1,2018-05-22T19:54:16Z,648,io_time,diskio3,host.local,disk2
+,,1,2018-05-22T19:53:56Z,1200,io_time,diskio3,host.local,disk2
+,,1,2018-05-22T19:54:06Z,1200,io_time,diskio3,host.local,disk2
+,,1,2018-05-22T19:54:16Z,1200,io_time,diskio3,host.local,disk2
```
onelson, line 26: Formerly, the answer we get in the test could either be a row with either diskio2 or diskio3 when reading from storage. By making diskio3 have a higher value, `max()` should be consistent across envs.

stdlib/universe/group_nulls_test.flux
```diff
@@ -1,114 +1,72 @@
...
+                    _field: "f0",
+                    _value: 2,
+                    t0: debug.null(type: "string"),
+                },
+                {
+                    _time: 2022-09-07T00:00:03Z,
+                    _measurement: "m0",
+                    _field: "f0",
+                    _value: 3,
+                    t0: debug.null(type: "string"),
```
Marwes, line 56: When we have nulls in the input we generally can't run it against C2/OSS as the nulls will be lost when written. So I think we can just skip `testing.load` and close https://github.com/influxdata/flux/issues/5178 ?

onelson: I think nulls in tags are expected to work here since tags can be added over time.

Several of the issues with the original test data fall into this trap you highlight (nulls for value, time, measurement all get dropped during ingest), but in the case of the tag I think we still want this coverage.

Marwes (approved):

stdlib/planner/group_first_last_test.flux
wolffcm: Yes, it does affect the pushdowns applied, for OSS. I filed this issue:
https://github.com/influxdata/influxdb/issues/23757

PR #5175, opened 2022-09-06
chore(docs): update unpivot documentation and add example

sanderson:
Updates the `experimental.unpivot()` documentation to describe the behavior of the function and add an example.

---

- [ x 📖 If **language features** are changing, ensure `docs/Spec.md` has been updated

### DIFF
--- a/stdlib/experimental/experimental.flux
+++ b/stdlib/experimental/experimental.flux
@@ -1309,12 +1309,46 @@ builtin histogram : (
 //
 builtin preview : (<-tables: stream[A], ?nrows: int, ?ntables: int) => stream[A] where A: Record
 
-// unpivot removes the `_time` column and any other column not in the group key and outputs a new table with `_field` and `_value` columns pairs.
+// unpivot creates `_field` and `_value` columns pairs using all columns (other than `_time`)
+// _not_ in the group key.
+// The `_field` column cotains the original column label and the `_value` column
+// contains the original column value.
+//
 // The output stream retains the group key and all group key columns of the input stream.
-// Specialized to transform the pivoted output from `iox.from()` into the unpivoted format.
+// `_field` is added to the output group key.
+// This function is designed to transform the pivoted output from `iox.from()`
+// into the unpivoted format.
 //
 // ## Parameters
 // - tables: Input data. Default is piped-forward data (`<-`).
+//
+// ## Examples
+//
+// ### Unpivot data into _field and _value columns
+//
+// ```
+// # import "array"
+// import "experimental"
+//
+// # data =
+// #     array.from(
+// #         rows: [
+// #             {_time: 2022-01-01T00:00:00Z, location: "New York", temp: 50.1, hum: 99.2},
+// #             {_time: 2022-01-02T00:00:00Z, location: "New York", temp: 55.8, hum: 97.7},
+// #             {_time: 2022-01-01T00:00:00Z, location: "Denver", temp: 10.2, hum: 81.5},
+// #             {_time: 2022-01-02T00:00:00Z, location: "Denver", temp: 12.4, hum: 41.3},
+// #         ],
+// #     )
+// #         |> group(columns: ["location"])
+// #
+// < data
+// >     |> experimental.unpivot()
+// ```
+//
+// ## Metadata
+// introduced: 0.172.0
+// tags: transformations
+//
 builtin unpivot : (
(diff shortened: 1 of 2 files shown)


stdlib/experimental/experimental.flux
```diff
@@ -1309,12 +1309,46 @@ builtin histogram : (
 //
 builtin preview : (<-tables: stream[A], ?nrows: int, ?ntables: int) => stream[A] where A: Record
 
-// unpivot removes the `_time` column and any other column not in the group key and outputs a new table with `_field` and `_value` columns pairs.
+// unpivot creates `_field` and `_value` columns pairs using all columns (other than `_time`)
+// _not_ in the group key.
+// The `_field` column cotains the original column label and the `_value` column
```
wolffcm, line 1314: Typo in "cotains"

stdlib/experimental/experimental.flux
```diff
@@ -1309,12 +1309,46 @@ builtin histogram : (
...
+// contains the original column value.
+//
 // The output stream retains the group key and all group key columns of the input stream.
-// Specialized to transform the pivoted output from `iox.from()` into the unpivoted format.
+// `_field` is added to the output group key.
+// This function is designed to transform the pivoted output from `iox.from()`
+// into the unpivoted format.
```
wolffcm, line 1320: Note that `iox.from` is going away soon, as we no longer need it. It was never callable outside of special environments that are inaccessible to external users in any case. It would probably be best to remove the reference to it here.

Do you think that `unpivot` could be actually useful for users? I think perhaps that it belongs in `internal`. The problem with our `internal` package is that it appears in a directory called `internal` which Go interprets as meaning "nothing in here is visible in other modules". This being the case we can't import the Go data structures for use in downstream repositories. So that's why I put `unpivot` here. If there's no real use case for `unpivot` outside of our IOx support, probably we should work around the limitations of `internal` somehow.

wolffcm (commented):
I had some questions for you.

stdlib/experimental/experimental.flux
sanderson: I think there's definitely a use case for it. For example, if I run a custom aggregate that gives me multiple columns that I want to display as individual graph lines I need to unpivot those columns into separate tables.

sanderson: In fact, I think a good use case for this could be a `candlestick` function. For example:

```javascript
import "sampledata"
import "date"
import "experimental"

s1 = sampledata.float() 
s2 = sampledata.float() |> map(fn: (r) => ({ r with _value: r._value * 10.0})) |> timeShift(duration: 1m)
s3 = sampledata.float() |> map(fn: (r) => ({ r with _value: r._value * 20.0})) |> timeShift(duration: 2m)

candlestick = (every, tables=<-, column="_value") => {

    _calcCandlestick = (tables=<-, column="_value") =>
        tables |> reduce(
            identity: {index: 0, open: 0.0, close: 0.0, min: 0.0, max: 0.0},
            fn: (r, accumulator) => ({
                index: accumulator.index + 1,
                open: if accumulator.index == 0.0 then r._value else accumulator.open,
                close: r._value,
                min: if accumulator.index == 0.0 then r._value else if r._value < accumulator.min then r._value else accumulator.min,
                max: if accumulator.index == 0.0 then r._value else if r._value > accumulator.max then r._value else accumulator.max,
            })
        )
            |> drop(columns: ["index"])
    
    return tables
        |> aggregateWindow(every: every, fn: _calcCandlestick)
        |> experimental.unpivot()
}

union(tables: [s1, s2, s3]) |> sort(columns: ["_time"])
    |> range(start: sampledata.start, stop: date.add(d: 2m, to: sampledata.stop))
    |> filter(fn: (r) => r.tag == "t2")
    |> candlestick(every: 40s)
```

Without `unpivot()`, you wouldn't be able to easily graph this data in the current InfluxDB UI.

wolffcm: Wow, nice custom aggregate 😎 

This makes a lot of sense, you've convinced me of its usefulness. I have an update to my open PR adding a parameter to `unpivot` that I'll be pushing shortly, I'll be curious to know what you think.

wolffcm (approved):

stdlib/experimental/experimental.flux
sanderson: Are you good if I merge this before your add the parameter?

PR #5181, opened 2022-09-08
fix: treat null inputs to `vectorAdd`, `vectorOr` as false

onelson:
Formerly, when the input to vectorized logical ops was a null, we'd see:

```
runtime error: cannot use vectorized or operation on non-vector invalid
```

The precedent set by the row-based logical ops is to treat incoming nulls as false.

This diff extracts the input validation to a separate function which is responsible for transitioning the incoming `values.Value` to a `values.Vector` giving it the chance to intercept the problem null case and convert them to a new vec repeat holding a boolean false.

> N.b this implementation has a slight semantic difference when compared to the row-based version.
>
> For the regular logical ops, we short-circuit when looking at the left-hand side, converting the return type to a boolean for these cases. There's still a case where the right-hand side is returned as-is, meaning the return type can be either null or bool.
>
> The vector code-path eagerly promotes null inputs (at the supposed vector-level) to vec repeats of `false`, meaning the output would be bool rather than null when the op happens to be one where the RHS would be returned.
>
> I believe it is possible to precisely pass along a null output the the same exact case, but it'll be a lot more messy (speculative).
 
### Checklist

### DIFF
--- a/compiler/runtime.go
+++ b/compiler/runtime.go
@@ -345,34 +345,49 @@ func (e *logicalVectorEvaluator) Eval(ctx context.Context, scope Scope) (values.
 	}
 	defer l.Release()
 
-	if typ := l.Type().Nature(); typ != semantic.Vector {
-		return nil, errors.Newf(codes.Invalid, "cannot use vectorized %s operation on non-vector %s", e.operator, typ)
-	} else if elemType := l.Vector().ElementType().Nature(); elemType != semantic.Bool {
-		return nil, errors.Newf(codes.Invalid, "cannot use operand of type %s with logical %s; expected boolean", elemType, e.operator)
+	lv, err := getLogicalVectorInput(e.operator, l)
+	if err != nil {
+		return nil, err
 	}
-
 	r, err := e.right.Eval(ctx, scope)
 	if err != nil {
 		return nil, err
 	}
+	defer r.Release()
 
-	if typ := r.Type().Nature(); typ != semantic.Vector {
-		return nil, errors.Newf(codes.Invalid, "cannot use vectorized %s operation on non-vector %s", e.operator, typ)
-	} else if elemType := r.Vector().ElementType().Nature(); elemType != semantic.Bool {
-		return nil, errors.Newf(codes.Invalid, "cannot use operand of type %s with logical %s; expected boolean", elemType, e.operator)
+	rv, err := getLogicalVectorInput(e.operator, r)
+	if err != nil {
+		return nil, err
 	}
-
 	mem := memory.GetAllocator(ctx)
 
 	switch e.operator {
 	case ast.AndOperator:
-		return vectorAnd(l.Vector(), r.Vector(), mem)
+		return vectorAnd(lv, rv, mem)
 	case ast.OrOperator:
-		return vectorOr(l.Vector(), r.Vector(), mem)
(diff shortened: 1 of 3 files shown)


nathanielc:
> The precedent set by the row-based logical ops is to treat incoming nulls as false.

I might be missing some context but we use ternary boolean logic in Flux.  See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#missing-values-null in short logical operation with nulls should produce more nulls not a true or false value. Does this logic still hold with vectorization?

onelson:
@nathanielc it does not hold true (that nulls should produce more nulls) in the proposed patch, for _some cases_.
Similarly, it does not hold true for the regular row-based implementation when the LHS prompts a short circuit.

To try and nutshell it, in the row-based:
```
null and null // gives false, LHS is converted to bool
false or null // gives null, RHS is returned as-is
```

In the proposed patch here, null inputs (ie, like an invalid member access) are converted to constant `false` up front.
For cases where nulls appear on a row-by-row basis (ie, the input is a vector with some nulls in there), but the array-based implementations handle the branching of individual nulls case by case.

Looking at the implementation for `array.Or`, it looks like it deviates slightly from the row-based version.

https://github.com/influxdata/flux/blob/edc16bf677327e55050d59e781b2657273c4f397/array/binary.go#L70-L74

Here, if we have a null on the left, we insert a null (actually... this seems like another, different bug?)
The row-based version would instead give whatever happens to be in the RHS:
https://github.com/influxdata/flux/blob/edc16bf677327e55050d59e781b2657273c4f397/compiler/runtime.go#L308-L329

Regardless, if we say nulls need to flow through and we want the rule applied consistently across the vectorized and row-based paths, I think we need to make tweaks to both.

onelson:
Sympathetically, I suspect the divergence from the advertised spec in both cases is thanks to how inconvenient it has been to write tests concerned with nulls. `debug.null` and `debug.opaque` make this easier in flux acceptance tests today.

I can push on this (both sides) for this PR if that's desired. It'll be a bit more involved, but it should be doable.

nathanielc:
@onelson OK so we seem to have drifted from the SPEC anyways here. So in this case I say lets move forward with this PR as is and comeback with a follow up PR to fix both the row-based and column-based implementation to match the spec. Does that sound good? If so can you create an issue with the above details?

onelson:
Yep! On it. I'll write up an issue to move towards conformance on both ends.

onelson:
:weary: ugh, yes. It looks like we've got a correctness problem in `vectorOr`:

```
> array.from(rows:[{x: 1, a: debug.null(type: "bool"), b: true}]) |> debug.opaque() |> map(fn: (r) => ({x: r.a or r.b }))
Result: _result
Table: keys: []
      x:bool  
------------  
              
>
```

`null or true` should absolutely give `true`, but we get the `null` instead.

onelson:
Filed the follow-up bug as https://github.com/influxdata/flux/issues/5188

onelson:
Leaving this parked here for now since reviewer won't be available until tomorrow, but launching on #5188 using this branch as the root.

scbrickley (approved):
Looks good to me. The test case really helped clarify what was being implemented here.

onelson:
Closing in favor of the more comprehensive https://github.com/influxdata/flux/pull/5192

PR #5192, opened 2022-09-12
fix: logical null handling

### LINKED ISSUE #5188: Realign logical and/or with spec regarding nulls
In the Flux spec, we detail how nulls should be treated when used in logical operators (`and`, `or`).

https://github.com/influxdata/flux/blob/edc16bf677327e55050d59e781b2657273c4f397/docs/SPEC.md?plain=1#L1439-L1444

Today we have two implementations for logical ops, the regular and the vectorized. Neither of these implementations match the spec completely when it comes to null handling. The vectorized implementation also has some correctness problems re: nulls.

For the vectorized implementation, note that nulls come in **2 flavors**:
- **individual nulls** within a vector input (ex: when `r.field` is sometimes null, sometimes not)
- **"constant" nulls** when the vector is missing (ex: when `r.field` where `r.field` just doesn't exist)

Work done in https://github.com/influxdata/flux/pull/5181 updated the vectorized code path such that it converts the latter to be a "constant false" which is not correct per the spec, but avoids a runtime error, while also weakly helping to mitigate a correctness problem in `vectorOr`.

> Vectorized conditionals ran into similar issues re: nulls. The fix for conditionals happened here: https://github.com/influxdata/flux/pull/5118 and the noteworthy bit is how the fix involved updating the constant-handling code to take pointers to scalar values so they could handle the "constant null" case. Something similar may need to happen for logical ops in order to adhere to the spec 100%.

See the comments in #5181 for more discussion on the defects found today.

DOD:
- write acceptance tests that demonstrate the table listing in the spec for:
  - vectorized (assert that the vectorized map planner rule is used)
  - non-vectorized (assert the vectorized planner rule is _not used_, you can include a non-vectorized function call in your map fn to force it to not vectorize...)
    - feature flag this breaking change
    - check both interpreter and compiler logic (i.e. boolean logic inside and outside of map)
  - include tests that demonstrate **both flavors** of nulls for the vectorized path
- update all implementations to match the null handling rules outlined in the spec such that the tests pass

onelson:
Fixes #5188

During an earlier attempt #5181 to fix a type-error problem in the vectorized logical operator implementation, we identified some deviation from the spec in how it worked with respect to `null` inputs. We also found the standard row-based version of logical ops _also deviated_ but in different ways.

This diff attempts to align all logical operator implementations (vectorized, row-based, as well as "in the interpreter") to be representative of what we have in the spec.

In addition to adhering to the spec re: nulls, this diff also adds short-circuiting support to the vectorized implementation in a similar fashion to vectorized conditionals.
This is achieved by checking the left-hand-side vector to see if all the elements are a single value. When this is the case and the value happens to be a short-circuit for the given op, we can skip evaluating the right-hand side entirely.

> In cases where all the elements are the same, but are not useful for short-circuiting, we can still use this information to select a "const" implementation for the op (where one side is fixed throughout), just like in the case of a `VectorRepeatValue`.

## Feature Flagging

Since the null handling aspect changes the behavior of _all logical op_ implementations in flux, the modifications to the conventional row-based and interpreter impls have been guarded with a new feature flag: `strictNullLogicalOps`.

The vectorized impl will use the newly aligned code with or without the `strictNullLogicalOps` flag set.

Since the status quo behavior is not aligned with the spec, we could consider this a "bug fix" but it is a **breaking change** nonetheless. Should there be users who relied on the prior behavior, the flag is our escape hatch to give the affected users a respite to update their scripts if needed.

### Checklist

### DIFF
--- a/array/binary.go
+++ b/array/binary.go
@@ -15,42 +15,60 @@ func And(l, r *Boolean, mem memory.Allocator) (*Boolean, error) {
 	b := NewBooleanBuilder(mem)
 	b.Resize(n)
 	for i := 0; i < n; i++ {
-		if l.IsValid(i) && r.IsValid(i) {
-			b.Append(l.Value(i) && r.Value(i))
-		} else {
-			b.AppendNull()
-		}
-	}
-	a := b.NewBooleanArray()
-	b.Release()
-	return a, nil
-}
-
-func AndLConst(l bool, r *Boolean, mem memory.Allocator) (*Boolean, error) {
-	n := r.Len()
-	b := NewBooleanBuilder(mem)
-	b.Resize(n)
-	for i := 0; i < n; i++ {
-		if r.IsValid(i) {
-			b.Append(l && r.Value(i))
-		} else {
-			b.AppendNull()
-		}
-	}
-	a := b.NewBooleanArray()
-	b.Release()
-	return a, nil
-}
-
-func AndRConst(l *Boolean, r bool, mem memory.Allocator) (*Boolean, error) {
-	n := l.Len()
-	b := NewBooleanBuilder(mem)
-	b.Resize(n)
-	for i := 0; i < n; i++ {
+		var elmL *bool
 		if l.IsValid(i) {
-			b.Append(l.Value(i) && r)
-		} else {
+			x := l.Value(i)
+			elmL = &x
+		}
+		var elmR *bool
+		if r.IsValid(i) {
+			x := r.Value(i)
+			elmR = &x
+		}
+
+		if elmL == nil && elmR == nil {
+			// both sides are null
 			b.AppendNull()
+		} else if elmL == nil || elmR == nil {
+			// one side is null, the other is false
+			if (elmL != nil && !*elmL) || (elmR != nil && !*elmR) {
+				b.Append(false)
+			} else {
+				b.AppendNull()
+			}
+		} else {
+			// no nulls
+			b.Append(*elmL && *elmR)
+		}
+	}
+	a := b.NewBooleanArray()
+	b.Release()
+	return a, nil
+}
+
+func AndConst(fixed *bool, arr *Boolean, mem memory.Allocator) (*Boolean, error) {
+	n := arr.Len()
+	b := NewBooleanBuilder(mem)
+	b.Resize(n)
+	for i := 0; i < n; i++ {
+		var elm *bool
+		if arr.IsValid(i) {
+			x := arr.Value(i)
+			elm = &x
+		}
+		if fixed == nil && elm == nil {
+			// both sides are null
+			b.AppendNull()
+		} else if fixed == nil || elm == nil {
+			// one side is null, the other is false
+			if (fixed != nil && !*fixed) || (elm != nil && !*elm) {
+				b.Append(false)
+			} else {
+				b.AppendNull()
+			}
+		} else {
+			// no nulls
+			b.Append(*fixed && *elm)
 		}
 	}
 	a := b.NewBooleanArray()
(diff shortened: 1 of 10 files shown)


onelson:
Initial test run. The tests might still need to be reworked since I'm not 100% sure where I should be expecting untyped null output (ie, missing columns) or typed nulls.

```
go run ./cmd/flux test --test logical_untyped_null_vectorized,logical_typed_null_vectorized,logical_untyped_null,logical_typed_null
xxxx
failures:

        stdlib/universe/map_test.flux: logical_typed_null_vectorized ... fail: Table: keys: []
          _diff:string    _true:bool   _false:bool    _null:bool  nullOrFalse:bool  nullOrTrue:bool  nullOrNull:bool  nullAndFalse:bool  nullAndTrue:bool  nullAndNull:bool  
----------------------  ------------  ------------  ------------  ----------------  ---------------  ---------------  -----------------  ----------------  ----------------  
                     -          true         false                                             true                               false                                      
                     +          true         false                                                                                                                           


        stdlib/universe/map_test.flux: logical_untyped_null_vectorized ... fail: Table: keys: []
          _diff:string    _true:bool   _false:bool  nullOrTrue:bool  nullAndFalse:bool  nullOrNull:bool  nullAndTrue:bool  nullAndNull:bool  nullOrFalse:bool  
----------------------  ------------  ------------  ---------------  -----------------  ---------------  ----------------  ----------------  ----------------  
                     -          true         false             true              false                                                                         
                     +          true         false             true              false            false             false             false             false  


        stdlib/universe/map_test.flux: logical_typed_null ... fail: Table: keys: []
          _diff:string    _true:bool   _false:bool    _null:bool  nullOrFalse:bool  nullOrTrue:bool  nullOrNull:bool  nullAndFalse:bool  nullAndTrue:bool  nullAndNull:bool  
----------------------  ------------  ------------  ------------  ----------------  ---------------  ---------------  -----------------  ----------------  ----------------  
                     -          true         false                                             true                               false                                      
                     +          true         false                           false             true                               false             false             false  


        stdlib/universe/map_test.flux: logical_untyped_null ... fail: Table: keys: []
          _diff:string    _true:bool   _false:bool  nullOrTrue:bool  nullAndFalse:bool  nullAndNull:bool  nullAndTrue:bool  nullOrFalse:bool  nullOrNull:bool  
----------------------  ------------  ------------  ---------------  -----------------  ----------------  ----------------  ----------------  ---------------  
                     -          true         false             true              false                                                                         
                     +          true         false             true              false             false             false             false                   



---
Found 864 tests: passed 0, failed 4, skipped 860
exit status 1
```

onelson:
@scbrickley this is the "likely messy" work I didn't want to dive into in #5181. Assigned you and @nathanielc since y'all both had eyes on the earlier PR (and @wolffcm, auto-assigned, happens to be unavailable).

compiler/vectorized_test.go
```diff
@@ -208,6 +208,89 @@ func TestVectorizedFns(t *testing.T) {
...
 		},
+		{
+			name:         "logical expressions",
+			fn:           "(r) => ({ r with c: r._a and r._b, d: r._a or r._b })",
+			vectorizable: true,
+			skipComp:     true,
+		},
+	}
+
+	logicalTests := []struct {
```
onelson, line 219: These go-side tests are the only place we verify that values in the vectorized code path are retained/released correctly. It appears we didn't have any for logical ops before now.

It looks like I'm testing for correctness here, but that's the aim. Mostly I'm trying to hit different short-circuit cases to verify memory accounting when (main) early returns are in play.

compiler/compiler.go
```diff
@@ -634,8 +634,11 @@ func compile(n semantic.Node, subst semantic.Substitutor) (Evaluator, error) {
 				right:    r,
 			}, nil
 		}
-
-		return &logicalEvaluator{
+		// TODO(onelson): try to select strict or regular logical here.
+		//   Need ctx to read flagger.
+		//   Currently this happens during Eval() where ctx is available and
+		//   will fallback to the regular logicalEvaluator if the flag is unset.
+		return &logicalStrictNullEvaluator{
```
onelson, line 641: This would be the ideal spot to branch, but we need the ctx to look at flags. Maybe y'all know how I can get the flagging info here without junking up the existing signatures?

array/binary.go
```diff
@@ -15,42 +15,60 @@ func And(l, r *Boolean, mem memory.Allocator) (*Boolean, error) {
...
 		}
 	}
 	a := b.NewBooleanArray()
 	b.Release()
 	return a, nil
 }
 
-func AndRConst(l *Boolean, r bool, mem memory.Allocator) (*Boolean, error) {
-	n := l.Len()
+func AndConst(fixed *bool, arr *Boolean, mem memory.Allocator) (*Boolean, error) {
```
onelson, line 49: When I added the initial "vec repeat" support for logical ops I didn't realize the op was commutative. Now that I'm wise to this, I updated this to just swap the args around at the call sites instead of having L and R versions.

stdlib/universe/map_test.flux
```diff
@@ -538,39 +538,206 @@ testcase vectorize_nested_logicals {
...
+                        trueOrNull: r._true or r._null,
+                        nullAndTrue: r._null and r._true,
+                        trueAndNull: r._true and r._null,
+                        nullAndNull: r._null and r._null,
+                        nullOrNull: r._null or r._null,
+                    }),
+            )
+            |> drop(columns: ["_true", "_false"])
+
+    testing.diff(want: want, got: got)
```
nathanielc, line 613: At first I was confused how this test could pass then I realized the missing labels in the want where the values that should be null. Maybe adding a comment to remind the reader that a missing column is an untyped null would help?

nathanielc (approved):
These tests are great! One small suggestion below.

stdlib/universe/map_test.flux
onelson: Added comments in 61f021a to explain how the untyped nulls are introduced. Yeah - it's a bit of a trick to make this happen on purpose :woozy_face:

onelson:
There'll be some conflicts on this branch until I close or merge the root branch (from the initial PR #5181). I'll clean that up once folks have had their weigh-in on this diff.

scbrickley (approved):
This looks good to me. I am missing some context because I haven't been super involved with the vec repeat work, but I think I understand what's being implemented here.

Merge away :+1:

PR #5204, opened 2022-09-15
feat: Enable codespan formatting for errors via feature flags

Marwes:
This improved error formatting was added back in https://github.com/influxdata/flux/pull/4270 but so far we have only used it for limited parts of the command line tool. This enables the new formatting everywhere through the `prettyError` feature flag which will allow us to try and roll this out everywhere else.

```
$ go run ./cmd/flux ../test.flux
error @4:33-4:35: expected int but found string
exit status 1
$ go run ./cmd/flux --features "{ \"prettyError\": true }" ../test.flux
error: expected int but found string
  ┌─ main:4:33
  │
4 │     |> map(fn: (r) => ({ x: 1 + "" }))
  │                                 ^^


exit status 1
```
(The new format should also be colored in the terminal, but the information is unfortunately lost on the go side)

I enabled the new format by default for all of our command line tools as the format fits in well there already, it is just in the GUI/server side I can see some issues.

### Checklist


Marwes:
I don't like how I had to thread the context/feature flags down through the separate AST compilation but there doesn't seem to be a way to merge that code into the later semantic package compilation (I'd rather we only pass scripts as strings to rust and get a semantic graph back and avoid passing the AST back and forth between rust and go).

Marwes:
Ping @skartikey

skartikey:
> I don't like how I had to thread the context/feature flags down through the separate AST compilation but there doesn't seem to be a way to merge that code into the later semantic package compilation (I'd rather we only pass scripts as strings to rust and get a semantic graph back and avoid passing the AST back and forth between rust and go).

Same observation while going through the code. Not sure if there's any other way of doing it. 
@jsternberg may I ask you to have a look as well and suggest if there's any alternative here?

skartikey (approved):

Issue #5158, opened 2022-09-01
Streams should reject invalid records at compile time


Marwes:
`stream` types are only allowed to contain records of types that fit in an arrow array so if you pass in a value that does not work, such as a nested record you will get a runtime error.

```flux
import "sampledata"

sampledata.int()
    |> map(fn: (r) => ({ x: {} }))
// runtime error @4:8-4:35: map: map object property "x" is object type which is not supported in a flux table
```

Waiting until runtime should not be necessary however as we already know at compile time that the code is invalid. We could make this enforment at compile time by adding a new kind that is only implement for records that contain fields that work in arrow arrays.

```flux
builtin map : (<-tables: stream[A], fn: (r: A) => B, ?mergeKey: bool) => stream[B]
    where A: RecordWithArrowValidColumns,
               B: RecordWithArrowValidColumns
```

DOD:

- :bike: :house: `RecordWithArrowValidColumns`
- Add new kind which is only implement for records with fields that work in an arrow array
- Require `A` in `stream[A]` is constrained by this kind in any type signatures
- Update all type signatures using `stream` to declare this kind

Issue #4344, opened 2021-12-13
EPIC: Arrays and Streams should be different types


onelson:
Today `TableObject` implements the `Array` interface but none of the methods are really supported.

We should refactor such that both the Flux and Go type systems can tell the difference between these types to help us avoid runtime panics down the road. 

- #4343
- #4468

nathanielc:
Notes from discussion:

* We need two different kinds of contain types. 
    * An index based container type (like the current type we have)
    * An iteration based container type, i.e. you can get next but not length or index into arbitrary locations
 
How should  we address the delayed/async nature of the new iteration based container type? Should it present a lazy iterator interface? Or should it present an explicit async/await/futures interface?

PR #5206, opened 2022-09-15
fix(SPEC): update SPEC and fix some inconsistencies

nathanielc:
The change updates the SPEC to reflect the current status of Flux.
As such the SPEC is no longer a _living_ document and instead describes
current behaviors. The few IMPL issues that were not complete are left
as feature requests for the future.

Summary of the changes:

- Specific docs about universe functions have been removed
- References to the standard library have been added
- `empty` and `in` keywords have been removed
- Missing operators added
- Numeric literals removed
- Regex escape sequences have been updated in the SPEC and the
  implementation was updated to match the SPEC
- Duration literal add/sub examples have been updated
- Shorthand date time literals have been updated
- Add docs about stream type remove generator type

### Checklist


docs/SPEC.md
```diff
@@ -569,13 +531,13 @@ Values must also be of the same type.
 ##### Function types
 
 A _function type_ represents a set of all functions with the same argument and result types.
+Functions arguments are always named, meaning there are not positional arguments.
```
pierwill, line 534:
```suggestion
Functions arguments are always named. (There are no positional arguments.)
```

docs/SPEC.md
```diff
@@ -894,6 +856,17 @@ A function may be given a name using a variable assignment.
 Function literals are _closures_: they may refer to variables defined is a surrounding block.
 Those variables are shared between the function literal and the surrounding block.
 
+
+Function arguments are named, there are no positional arguments.
```
pierwill, line 860:
```suggestion
Function arguments are named. (There are no positional arguments.)
```
or
```suggestion
Function arguments are named.
```

Might not need to repeat this.

docs/SPEC.md
```diff
@@ -1189,7 +1162,7 @@ All such values must have a corresponding builtin statement to declare the exist
 
 Example
 
-    builtin filter : (<-tables: [T], fn: (r: T) => bool) => [T]
+    builtin filter : (<-tables: stream[T], fn: (r: T) => bool) => stream[T]
```
pierwill, line 1165: Is `stream` a keyword or reserved string here? I don't see this added to the grammar anywhere yet.

docs/SPEC.md
```diff
@@ -205,64 +186,61 @@ Examples:
...
+    date.add(d: date.scale(d:1mo, n:3), to: 2018-01-31T00:00:00Z)  // 2018-04-30T00:00:00Z
 
 #### Date and time literals
 
 A date and time literal represents a specific moment in time.
 It has a date part, a time part and a time offset part.
 The format follows the RFC 3339 specification.
-The time is optional, when it is omitted the time is assumed to be midnight for the default location.
-The time_offset is optional, when it is omitted the location option is used to determine the offset.
+The time is optional, when it is omitted the time is assumed to be midnight UTC.
```
sanderson, line 236:
```suggestion
The time is optional. When it is omitted the time is assumed to be midnight UTC.
```

docs/SPEC.md
```diff
@@ -343,24 +314,22 @@ A regular expression literal represents a regular expression pattern, enclosed i
 Within the forward slashes, any unicode character may appear except for an unescaped forward slash.
 The `\x` hex byte value representation from string literals may also be present.
 
-Regular expression literals support only the following escape sequences:
+In addition standard escape sequences regular expression literals also support the following escape sequences:
```
sanderson, line 317:
```suggestion
In addition standard escape sequences, regular expression literals also support the following escape sequences:
```

docs/SPEC.md
sanderson, line 534:
```suggestion
Functions arguments are always named. There are no positional arguments.
```

docs/SPEC.md
sanderson, line 860:
```suggestion
Function arguments are named. There are no positional arguments.
```

docs/SPEC.md
sanderson: `stream[]` is a type expression that is noted later in the SPEC.

docs/SPEC.md
```diff
@@ -1501,3113 +1346,14 @@ So, you have to:
...
-    |> sum()
-```
-
-#### Multiple aggregates
-
-Multiple aggregates can be applied to the same table using the `aggregate` function.
+### Standard Library
 
-[IMPL#139](https://github.com/influxdata/platform/issues/139) Add aggregate function
+Flux provides a standard library of functions. Find documentation here https://docs.influxdata.com/flux/v0.x/stdlib/
```
sanderson, line 1351: Use the `latest` redirect to keep this current.

```suggestion
Flux provides a standard library of functions. Find documentation here https://docs.influxdata.com/flux/latest/stdlib/
```

docs/SPEC.md
pierwill: Yes, but it should be part of the BNF grammar as well.

docs/SPEC.md
```diff
@@ -205,64 +186,61 @@ Examples:
...
-The time_offset is optional, when it is omitted the location option is used to determine the offset.
+The time is optional, when it is omitted the time is assumed to be midnight UTC.
 
     date_time_lit     = date [ "T" time ] .
     date              = year_lit "-" month "-" day .
     year              = decimal_digit decimal_digit decimal_digit decimal_digit .
     month             = decimal_digit decimal_digit .
     day               = decimal_digit decimal_digit .
-    time              = hour ":" minute ":" second [ fractional_second ] [ time_offset ] .
+    time              = hour ":" minute ":" second [ fractional_second ] time_offset .
```
nathanielc, line 243: I left the implementation unchanged in that the time offset is optional so that its easier to provide a good error message to users if it is missing.

docs/SPEC.md
nathanielc, line 236: The actual behavior implemented is to use UTC midnight, and no one is asking for the default location behavior as such the SPEC has been updated to reflect reality.

docs/SPEC.md
```diff
@@ -93,35 +90,20 @@ Examples:
...
 
     +   ==   !=   (   )   =>
     -   <    !~   [   ]   ^
-    *   >    =~   {   }
-    /   <=   =    ,   :
+    *   >    =~   {   }   ?
+    /   <=   =    ,   :   "
     %   >=   <-   .   |>
 
-#### Numeric literals
```
nathanielc, line 111: We tried to implement numeric literals but were unsuccessful in making it complete through the entire implementation. We can revisit this later but for now its no longer part of the SPEC.

docs/SPEC.md
pierwill: I'm not sure I understand. The way I read this change to the grammar, the `time_offset` was optional before, but now it is required.

docs/SPEC.md
```diff
@@ -343,24 +314,22 @@ A regular expression literal represents a regular expression pattern, enclosed i
...
+
 
 Examples:
 
     /.*/
     /http:\/\/localhost:8086/
     /^\xe6\x97\xa5\xe6\x9c\xac\xe8\xaa\x9e(ZZ)?$/
     /^日本語(ZZ)?$/ // the above two lines are equivalent
-    /\\xZZ/ // this becomes the literal pattern "\xZZ"
-    /a\/b\\c\d/ // escape sequences and character class shortcuts are supported
```
nathanielc, line 363: There is no need to escape backslash in regex and it was never implemented, The SPEC has been updated to reflect that.

docs/SPEC.md
```diff
@@ -511,8 +475,6 @@ The string type name is `string`.
 The string type is nullable.
 Note that an empty string is distinct from a _null_ value.
 
-The length of a string is its size in bytes, not the number of characters, since a single character may be multiple bytes.
```
nathanielc, line 514: Nothing in the SPEC depends on the length of a string as such this is best left documented in the stdlib strings package.

docs/SPEC.md
```diff
@@ -1501,3113 +1346,14 @@ So, you have to:
...
-Example: `encode(v: {label: "abcd})` returns JSON encoded value of the record as bytes.
-
-Types `record` and `dictionary` are encoded as JSON objects.
-
-### Composite data types
+### Composite data types
 
 A composite data type is a collection of primitive data types that together have a higher meaning.
 
-### Triggers
```
nathanielc, line 4586: This concept was never exposed to users nor is it important to communicate.

docs/SPEC.md
nathanielc: Correct now it is required, however the implementation will parse without a time offset and report the error that its missing.

docs/SPEC.md
nathanielc: Nice catch, added to the grammar, now. However its done in such a way that it doesn't need to be a keyword.

docs/SPEC.md
pierwill: `?` and `"` are operators?

docs/SPEC.md
```diff
@@ -93,35 +90,20 @@ Examples:
...
 #### Operators
 
 The following character sequences represent operators:
 
     +   ==   !=   (   )   =>
     -   <    !~   [   ]   ^
-    *   >    =~   {   }
-    /   <=   =    ,   :
+    *   >    =~   {   }   ?
+    /   <=   =    ,   :   "
```
pierwill, line 104: `?` and `"` are operators?

docs/SPEC.md
```diff
@@ -569,13 +531,13 @@ Values must also be of the same type.
...
 
-##### Generator types
-
-A _generator type_ represents a value that produces an unknown number of other values.
-The generated values may be of any other type but must all be the same type.
+##### Stream types
 
-<!-- no corresponding issue [IMPL#658](https://github.com/influxdata/platform/issues/658) Implement Generators types -->
+A _stream type_ represents an unbounded collection of values.
+The values may be of any other type but must all be the same type.
```
pierwill, line 540: What does "any other" mean here? Does it mean there can't be nested stream types?

Marwes, line 540: The values of a stream can't be any type, they must be records and those records may only hold int, uint, float, string, time or bool.

cc https://github.com/influxdata/flux/issues/5158

docs/SPEC.md
```diff
@@ -1172,7 +1145,7 @@ All such values must have a corresponding builtin statement to declare the exist
     MonoType     = Tvar | BasicType | ArrayType | RecordType | FunctionType .
     Tvar         = "A" … "Z" .
     BasicType    = "int" | "uint" | "float" | "string" | "bool" | "time" | "duration" | "bytes" | "regexp" .
-    ArrayType    = "[" MonoType "]" .
+    ArrayType    = [ "stream | "vector" ] "[" MonoType "]" .
```
Marwes, line 1148: I'd move the stream and vector to their own definition

```
StreamType = "stream" "[" MonoType "]"
// etc
```

or at least rename `ArrayType` to `CollectionType`

internal/parser/strconv.go
```diff
@@ -18,10 +18,17 @@ import (
 func ParseTime(lit string) (time.Time, error) {
 	if !strings.Contains(lit, "T") {
 		// This is a date.
-		return time.Parse("2006-01-02", lit)
+		t, err := time.Parse("2006-01-02", lit)
```
Marwes, line 21: Why do we have this hardcoded data here and why do we parse it? :confused:

libflux/flux-core/src/parser/strconv.rs
```diff
@@ -90,9 +92,40 @@ pub fn parse_regex(lit: &str) -> Result<String, String> {
...
+                    // this char can have any byte length
+                    unescaped.extend_from_slice(c.to_string().as_bytes());
+                }
+                None => return Err("unterminated regex sequence".to_string()),
+            },
+            // this char can have any byte length
+            _ => unescaped.extend_from_slice(c.to_string().as_bytes()),
+        }
+    }
+    let converted = match std::str::from_utf8(&unescaped) {
```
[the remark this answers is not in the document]
nathanielc: We do not support arbitrary byes in regex, the `\x` encoding must produce a valid utf8 sequence hence the explicit from_utf8 decode. The intent of the `\x` feature is to allow easier encoding of those values in strings/regex. However I see that the SPEC isn't explicit that the hex encoding must result in valid utf8 data. I'll add that.

docs/SPEC.md
nathanielc: The `?` shows up in type signatures for optional arguments. The `"` shows up for string literals. They are not operators in the sense of binary operators but they are still operators for the language. This list was taken from the scanner.rl implementation.

docs/SPEC.md
```diff
@@ -569,13 +533,13 @@ Values must also be of the same type.
...
 
-##### Generator types
-
-A _generator type_ represents a value that produces an unknown number of other values.
-The generated values may be of any other type but must all be the same type.
+##### Stream types
 
-<!-- no corresponding issue [IMPL#658](https://github.com/influxdata/platform/issues/658) Implement Generators types -->
+A _stream type_ represents an unbounded collection of values.
+The values must be records and those records may only hold int, uint, float, string, time or bool.
```
pierwill, line 542:
```suggestion
The values must be records and those records may only hold int, uint, float, string, time or bool types.
```

docs/SPEC.md
```diff
@@ -1169,10 +1144,12 @@ All such values must have a corresponding builtin statement to declare the exist
     BuiltinStatement = "builtin" identifier ":" TypeExpression .
     TypeExpression   = MonoType ["where" Constraints] .
 
-    MonoType     = Tvar | BasicType | ArrayType | RecordType | FunctionType .
+    MonoType     = Tvar | BasicType | ArrayType | StreamType | VectorType | RecordType | FunctionType .
     Tvar         = "A" … "Z" .
     BasicType    = "int" | "uint" | "float" | "string" | "bool" | "time" | "duration" | "bytes" | "regexp" .
     ArrayType    = "[" MonoType "]" .
+    StreamType    = "stream" "[" MonoType "]" .
+    VectorType    = "vector" "[" MonoType "]" .
```
pierwill, line 1152: What about:
```suggestion
    StreamType    = "stream" ArrayType .
    VectorType    = "vector" ArrayType .
```

?

nathanielc: I think that conflates the concepts of Stream and vector types. StreamType is not a stream of arrays rather it is its own thing. Same for vector types. So while its an equivalent grammar the semantics are a bit confusing using ArrayType.

onelson: We split Array and Stream purposefully since Stream represents a collection without a known size. Ref: https://github.com/influxdata/flux/issues/4344

Marwes (approved):

PR #5208, opened 2022-09-16
chore: Convert all parser test to use expect_test

Marwes:
By shrinking the size of `Debug` printed AST's  it becomes possible to use `expect_test` for the parser tests instead of manually writing it out without ballooning the test size to much. Using expect_test to generate the expected value lets us avoid manually writing it, which is tedious when adding tests and also very brittle since any change to one of the AST types may lead to several tests needing to be manually update.

The change size here is huge, but there really isn't a need to go through and compare everything here as I have taken care to not mix changes to tests and "real code", so with each test conversion was done without changing any actual behavior.

If we would like, we might use [expect_file](https://docs.rs/expect-test/latest/expect_test/macro.expect_file.html) instead to move the expectations out to separate files instead of storing it inline.

### Checklist

### DIFF
--- a/libflux/flux-core/src/ast/mod.rs
+++ b/libflux/flux-core/src/ast/mod.rs
@@ -74,7 +74,7 @@ impl From<Position> for lsp_types::Position {
 }
 
 /// Represents the location of a node in the AST.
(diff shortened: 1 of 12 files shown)


nathanielc (approved):
Looks great and the tests read well.

PR #5209, opened 2022-09-16
fix: Prevent the metadata map from being concurrently mutated

### LINKED ISSUE #5157: fatal error: concurrent map writes (findRecord / findTable)
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

Running findRecord consecutively leads to a panic:
```
import "math"
import "array"
import "date"
start = 2022-08-24T00:00:00Z
stop = 2022-08-25T00:00:00Z
environment = "LAC"
serviceName = "Digma.Analytics"
entrySpanName = "execute continuous_endpoints_usage"
inner_spans = from(bucket: "errors")
    |> filter(fn: (r) => r._measurement == "span")
    |> filter(fn: (r) => r.Debug == "False")
    |> filter(fn: (r) => r.SpanRole != "Entry")
    |> filter(fn: (r) => r.Environment == environment)
    |> filter(fn: (r) => r.ServiceName == serviceName)
    |> filter(fn: (r) => r.EntrySpanName == entrySpanName)
inner_span_fractions = inner_spans
    |> filter(fn: (r) => r._field == "EntrySpanDurationFraction")
get_pivoted_record = (instrumentationLibrary, spanName, fraction) => {
    record_time = inner_span_fractions
        |> filter(fn: (r) => r.InstrumentationLibrary == instrumentationLibrary)
        |> filter(fn: (r) => r.SpanName == spanName)
        |> map(fn: (r) => ({r with delta: math.abs(x: float(v: r._value) - fraction)}))
        |> group()
        |> min(column: "delta")
    record_final = from(bucket: "errors")
        |> range(start: record_time._time, stop: date.add(d: 1s, to: record_time._time))
        |> filter(fn: (r) => r._time == record_time._time)
        |> filter(fn: (r) => r._measurement == "span")
        |> filter(fn: (r) => r.Debug == "False")
        |> filter(fn: (r) => r.SpanRole != "Entry")
        |> filter(fn: (r) => r.Environment == environment)
        |> filter(fn: (r) => r.ServiceName == serviceName)
        |> filter(fn: (r) => r.EntrySpanName == entrySpanName)
        |> filter(fn: (r) => r.InstrumentationLibrary == instrumentationLibrary)
        |> filter(fn: (r) => r.SpanName == spanName)
        |> pivot(rowKey: ["_time"], columnKey: ["_field"], valueColumn: "_value")
        |> filter(fn: (r) => r._measurement == r._measurement)  // doesn't work without a filter (a stupid one is enough)
    return record_final
}
p50_fraction = inner_span_fractions
    |> group(columns: ["InstrumentationLibrary", "SpanName"])
    |> quantile(q: 0.5)
    |> filter(fn: (r) => r._value > 0.02)
    |> map(fn: (r) => get_pivoted_record(instrumentationLibrary: r.InstrumentationLibrary, spanName: r.SpanName, fraction: r._value))
    |> yield(name: "P50Fraction")
```
The idea is to get quantile on some metric, then get its record (I know about "exact_selector"), then get its pivoted record

__Expected behavior:__
The query runs successfully by concurrently grabbing the row based on each map iteration. 

__Actual behavior:__
Query fails with a runtime panic
```
fatal error: concurrent map writes
goroutine 56808 [running]:
runtime.throw({0x2205150?, 0x10?})
        /go/src/runtime/panic.go:992 +0x71 fp=0xc00419c970 sp=0xc00419c940 pc=0x43f871
runtime.mapassign_faststr(0xc0028c2a00?, 0x122?, {0x21f33a7, 0xf})
        /go/src/runtime/map_faststr.go:212 +0x39c fp=0xc00419c9d8 sp=0xc00419c970 pc=0x41ca1c
github.com/influxdata/flux/metadata.Metadata.Add(...)
        /go/pkg/mod/github.com/influxdata/flux@v0.171.0/metadata/metadata.go:10
github.com/influxdata/flux/stdlib/universe.tableFind({0x33b5780, 0xc0045f1860}, 0xc0045dc800, 0xc0045f18c0)
        /go/pkg/mod/github.com/influxdata/flux@v0.171.0/stdlib/universe/table_fns.go:142 +0x6ee fp=0xc00419cbc8 sp=0xc00419c9d8 pc=0x137804e
```

__Environment info:__
* InfluxDB version: ``2.4``

__DEV notes:__

1. Seems like a bug in flux. The metadata data type should be protected by a mutex to guard against concurrent writes. The US folks would know better though (is there some reason the table functions should never be able to run concurrently?)
2. I looked at a bug earlier this year, but I forget the details exactly. Something about the op spec or procedure spec not being copied, but instead shared between instances when table functions are in play. Feels like it could be a similar thing.
3. In this case Metadata is retrieved from context.Context so I assume it is a object that is shared

Community issue: https://influxcommunity.slack.com/archives/CH8RGRW4T/p1661930280766359

Marwes:
By protecting with an `RWMutex`. I haven't been able to reproduce the crash but given the stacktrace it seems obvious enough that it is because `Metadata` is stored in the `Context` for which there is only one per query. Since queries run on multiple threads we must ensure that multiple threads can't mutated it concurrently.

Fixes #5157

### Checklist

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -7,35 +7,15 @@ import (
 
 // Metadata is made as a standalone package to avoid import cycle:
 // influxd -> flux -> flux/interpreter -> flux/execute -> flux
-type Metadata struct {
-	// Metadata is passed as a value so we must store the mutex as a pointer to ensure it does not get copied
-	lock *sync.RWMutex
-	meta map[string][]interface{}
-}
-
-func NewMetadata() Metadata {
-	return Metadata{
-		lock: &sync.RWMutex{},
-		meta: make(map[string][]interface{}),
-	}
-}
+type Metadata map[string][]interface{}
 
 func (md Metadata) Add(key string, value interface{}) {
-	md.lock.Lock()
-	defer md.lock.Unlock()
-
-	md.meta[key] = append(md.meta[key], value)
+	md[key] = append(md[key], value)
 }
 
 func (md Metadata) AddAll(other Metadata) {
-	md.lock.Lock()
-	defer md.lock.Unlock()
-
-	other.lock.RLock()
-	defer other.lock.RUnlock()
-
-	for key, values := range other.meta {
-		md.meta[key] = append(md.meta[key], values...)
+	for key, values := range other {
+		md[key] = append(md[key], values...)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Marwes:
Metadata seems to both be used as a map that can be concurrently updated and as a value type that is passed around, I think I need to separate those two concepts actually

metadata/metadata.go
```diff
@@ -1,47 +1,91 @@
...
 			if ok := fn(key, value); !ok {
-				return
+				return false
 			}
 		}
+		return true
+	})
+}
+
+func (md Metadata) RangeSlices(fn func(key string, values []interface{}) bool) {
```
onelson, line 56: I don't know if there's a reasonable rule of thumb to follow, but it would be great if there was some suggestion for when this is important to use in the docstring.

Might be hopeless since I imagine the number one reason you'd seek this out is after a panic has already been reported :shrug: :woozy_face:

onelson (commented):
The change makes sense, adding locks should surely guard against concurrent access.
I'm not currently equipped to recommend a better approach.

I do see some CI failures that might be related to the change (in test code?)

metadata/metadata.go
```diff
@@ -1,47 +1,91 @@
...
+	md.lock.RLock()
+	defer md.lock.RUnlock()
+
+	for key, values := range md.meta {
+		if ok := fn(key, values); !ok {
+			return
+		}
```
pierwill, line 63: I think this might want an unlock before returning:

```suggestion
	for key, values := range md.meta {
		if ok := fn(key, values); !ok {
		        md.lock.RUnlock()
			return
		}
```

I came across something similar in IDPE a little while back: https://github.com/influxdata/idpe/pull/15392.

metadata/metadata.go
```diff
@@ -1,47 +1,91 @@
...
 	}
 	return nil, fmt.Errorf("key %s does not exist in Metadata", key)
 }
 
 func (md Metadata) GetAll(key string) []interface{} {
-	if values, ok := md[key]; ok {
+	md.lock.RLock()
+	defer md.lock.RUnlock()
+
+	if values, ok := md.meta[key]; ok {
```
pierwill, line 88: ditto above

metadata/metadata.go
onelson: I think the `defer md.lock.RUnlock()` above (L58) should satisfy this. Even if there's a panic, the defer should still fire iirc?

pierwill: 👍

metadata/metadata.go
Marwes: I removed `RangeSlices` now after the refactor.

onelson: Ah, yeah. The refactor looks like it explains the usage patterns better.

onelson (approved):
LGTM. I think it makes good sense to tuck the real MD away as a private field in this case. Better enforcement this way. :+1:

PR #5212, opened 2022-09-19
feat: Add dynamic type

### LINKED ISSUE #5140: Type system updates for `Dynamic`
A new `Dynamic` monotype is added such that values with this type:

- allow member access, producing a new `Dynamic` value
- allow indexing, producing a new `Dynamic` value

We know that we want `Dynamic` to only be received by certain functions, so we may also need new typeclasses.

We need to be able to pass `Dynamic` into cast functions like `int()`, `float()` and so on.
This would require a way to accept "`Any` and `Dynamic`" whereas normally `Any` would _not include_ `Dynamic`.

Also determine and implement feature flagging this new work. 
While the addition of the new types themselves might be okay, special care must be given to guard breaking signature changes.

DOD:
- Add new types/typeclasses so we can accept/deny `Dynamic` as needed
- flatbuffer plumbing for the new types
- feature flagging so the new type/signatures don't break scripts by default

scbrickley:
Rebased and modified version of #4762.

Closes #5140

Added some tests to make sure that retrieving a dynamic value via index or member access still returns a dynamic value. Calling `Dynamic()` on any non-dynamic value produces an `UnexpectedKind` panic (similar to the other methods in the value interface). Similarly, attempting to call `Int()`, `UInt()`, `Float()`, etc. on a dynamic type also produces an `UnexpectedKind` panic. This is to enforce the rule that dynamic values must be explicitly cast to the appropriate type in order to be used as such.


semantic/monotype.go
```diff
@@ -906,6 +913,21 @@ func NewDictType(keyType, valueType MonoType) MonoType {
 	return mt
 }
 
+// NewDynamicType will construct a new Dynamic MonoType
+func NewDynamicType() MonoType {
+	builder := flatbuffers.NewBuilder(32) // FIXME: how big?
```
onelson, line 918: The size of this is still an open question. I'm not that familiar with flatbuffers, but it does seem to me like it should be possible to mark this as 0-sized so long as whatever fields are associated with the node size themselves.
I'm thinking in terms of the `BaseNode` fields we see like the location of the node, etc.

values/dynamic.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (d dynamic) Vector() Vector {
+	panic(UnexpectedKind(semantic.Dynamic, semantic.Vector))
+}
+
+func (d dynamic) Equal(v Value) bool {
+	// XXX(sean): Do we need a custom implementation of Equal() for dynamic?
+	// Or are we fine deferring to the inner type?
+	// Seems like a dynamic value should not equal a static value, even if they're
+	// the same under the hood.
```
onelson, line 98: It's an interesting question. I could see this returning `true` if v successfully typechecks as a `Dynamic` and the `d.inner.Equal(v.inner)` is `true`.

I think the thing to consider is when this `Equal` method is used. Do we delegate to this via the `==` binary op? I would expect not, but I don't know.

onelson: It's probably a good idea to look at some of the other value type impls and what they do for `Equal`. I bet there's a precedent well established by now. I'd look at some scalar types (int, float -- float being the one I'd expect to be more irregular because of precision concerns) as well as array or object, just to see how we compare composite types.

From there, we'd need to make a decision about which model makes more sense: basic or composite.

libflux/flux-core/src/semantic/tests.rs
```diff
@@ -4278,6 +4278,48 @@ fn pipe_error() {
...
+        ],
+        src: r#"foo = {bar: d}
+v = foo.bar
+"#,
+        exp: map![
+            "foo" => "{bar: dynamic}",
+            "v" => "dynamic",
+        ],
+    }
+}
```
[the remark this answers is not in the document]
scbrickley: Added these tests, and `v` currently registers as type `A`. Trying to fix this now.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -1556,6 +1556,11 @@ impl MemberExpr {
             }
         }
 
+        if let MonoType::Dynamic(_) = &t {
+            self.typ = t.clone();
+            return Ok(());
+        }
+
```
scbrickley, line 1563: @onelson This and the accompanying change to `Expression::Index.infer()` were my solution to get the new type inference tests to pass. This seems... hacky tough. Open to suggestions for other solutions.

onelson: I think that looks fair. I'm no expert though. My reading of this says that when `r` is dynamic, assign the type of dynamic to expression accessing the property (which is what we want).

The fact you're cloning `t` to do this is okay, I think? Otherwise you could look to assign a new instance of the type (but I'm not sure why that'd matter). There's a flatbuffer pointer inside which I'm not sure how to construct by hand :sob:

Someone else with more experience in this are should probably check this out. Maybe Jonathan has input.

jsternberg (changes requested):
I just need @Marwes to give feedback on the type inference stuff. The type inference code is solid and I don't want it to be hacky so it needs to be solid and backed by the algorithm.

The compiler code is presently incredibly hacky and is already set for a refactor. The way dynamic would be implemented in a newer compiler would just be to have it built in from the ground up as an inherent part of the reflected value which is already how the vm proposal proposes things. So I'm just not really concerned with changes to the compiler to accommodate dynamic.

scbrickley:
@Marwes could you weigh in on the questions Jonathan raised above?

libflux/flux-core/src/semantic/types.rs
```diff
@@ -906,7 +921,8 @@ impl MonoType {
             (MonoType::Record(t), MonoType::Record(s)) => t.unify(s, unifier),
 
             (MonoType::Fun(t), MonoType::Fun(s)) => t.unify(s, unifier, MonoType::clone),
-
+            // FIXME: still not sure how to reason about this, but it seems like if either side is dynamic we should just "pass"
```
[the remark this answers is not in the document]
Marwes: This is ok if we are to have a dynamic value which you can pass anywhere, however since we wanted to try and restrict dynamic to just a few operations it might be better if we could avoid this. As far as I can see the runtime checks added should handle it, but if we can do it at compile time that is still better.

Does the it not work if we remove this and just rely on the `MonoType::Dynamic(_)` checks in `nodes.rs`?

libflux/flux-core/src/semantic/nodes.rs
Marwes: Doing more targeted checks like this seems better to me generally, but I could maybe see issues depending on the order that type inference runs. That is, this requires that we know that the type is `dynamic` when we reach this point in the code, we can't discover that it is dynamic later and then pass this check. I don't know if that is actually a problem though

libflux/flux-core/src/semantic/types.rs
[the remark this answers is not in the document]
scbrickley: This is a bit that carried over unchanged from the original POC, so I'm not entirely sure. Any input @onelson?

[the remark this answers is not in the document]
scbrickley: They pass. Removed.

libflux/flux-core/src/semantic/nodes.rs
jsternberg: Did we ever figure this out? It seems to me like it's possible this condition might not work correctly if we haven't yet determined that the original object was dynamic.

Just to make sure, if it turns out this fails, we can just turn dynamic off and this section of code will just not matter, correct?

onelson: I think the deal is all the fns that will construct `Dynamic` items will be in experimental, but iirc there's no reasonable way to feature flag this stuff.

It does follow that if user's don't construct dynamic items, this match arm won't matter.

scbrickley: Building on what Owen said, nothing currently uses the dynamic type. The functions that will use it will come from the other issues in the epic. I think the patches for those issues will be easier to test.

onelson (approved):
I share some of the hesitation re: inference but I hope we will still have the room to make adjustments as we implement the other bits on top of this new type.

My feeling is it'll be much easier to smoke out problems re: inference once we can use `Dynamic` in acceptance tests.

values/dynamic.go
scbrickley: The other impls I saw seem to follow this pattern:

- Check that the types/natures are the same
- If either side of the comparison is null, the values are not equal
- Check that the values are the same

For dynamic, I figured it would be sufficient to just check that both values are dynamic, and defer to the `Equal()` method of the inner type for all of the above checks.

onelson: Lucky guess on my part then :rofl: 
Sounds great.

libflux/flux-core/src/semantic/nodes.rs
scbrickley: @jsternberg Curious about your thoughts on this. I don't want to merge this if you have serious doubts. However, this is currently blocking other work, so if there's a specific change I can make to get this PR across the line, I'm happy to make it.

jsternberg (approved):
I think my own personal concerns are fine and this can be merged. I'd still defer to @Marwes's review. If he approves, I think we're good.

Marwes (approved):

PR #5217, opened 2022-09-21
fix(join): Handle empty tables

scbrickley:


stdlib/join/join_fn.go
```diff
@@ -266,6 +266,11 @@ func defaultRow(key flux.GroupKey, objType semantic.MonoType) values.Object {
 }
 
 func getObjectType(arg *semantic.Argument, cols []flux.ColMeta) (semantic.MonoType, error) {
+	// XXX(sean): This appears to fix the empty table bug, but what's the point of the `getObjectType()`
+	// function if we have this information already? Is this type info not always available?
+	if len(cols) < 1 {
+		return arg.TypeOf()
```
nathanielc, line 272: I share your concern, its seems backwards that we need to know the output type before we have evaluated any of the output rows. I know there are cases where the semantic type is incomplete, i.e. there are type variables. What happens to the join code in those cases? For example use a `{l with ...}` in the join `as` function so that the record is unbounded. 

Generally seems like we need this code to be more robust to unknown types and we should change the paradigm to just use the actual output type of the rows.  

As an aside hopefully this is the kind of thing that becomes much more straightforward with an IR and VM design.

scbrickley: Updated the integration test in such a way that (I think) both the output record and the right input would have incomplete type information. It passes, but I'm still suspicious of it...

jsternberg: I think there's generally a problem where sometimes type information is available and sometimes it isn't and I'm not always sure why it sometimes is and sometimes isn't. We've generally had to abstain from relying on the information returned from type inference because of this.

I'd really like to be able to answer this question better but this section of code has really gone through too many hack revisions for me to remember the answer or know it with certainty.

I might try to dive into this and ensure that it's addressed in the VM proposal or I at least understand the reasoning behind it.

scbrickley: Gotcha. I'll try to do some manual validation against some real data to see if this fix holds up.

scbrickley: Yup. Was able to create the following panic by running a query similar to the testcase, but using `from()` instead of `array.from()` or `gen.tables()`

```
panic: unknown builder for type: invalid
```

scbrickley:
This was an exploratory PR to see if it was possible to improve the way join handles empty tables. Ran into the issue that I suspected we would, which is that join does not have sufficient type information to do this for every query. It relies on the type information provided by the input tables, which is not all available in the case of an empty table.

Perhaps when the compiler refactor is done, this will be easier to resolve.

nathanielc:
Before we punt on fixing empty tables I'd like to understand _why_ join needs the type information of the input tables (beyond the group key). I can see that the current implementation needs it, but is there another implementation that doesn't? I'd like us to spend a bit thinking about that and see if we can find a solution. The compiler refactor is a long ways out and we already have community users hitting this limitation of join.

scbrickley:
Join needs the type information in a few different places, primarily to create `ChunkBuilders`, which use the type information to [decide what type of arrow array it needs to construct](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/execute/chunk_builder.go#L19-L27). This happens both when we [construct the default records](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L91-L113) and when we [create the output chunk](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L151).

Join uses the [result of the compiled `as` function](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L130) to [create the schema info](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L144) for the output chunk. I believe that function relies on the types of its arguments to determine the type for its output, but I may be wrong about that. We [set the types](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L58-L61) for the arguments in the `Prepare()` function, [create an object](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L62) with that type and store it as state in the `JoinFn` struct so that we can [reuse it](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/join_fn.go#L228-L229) without having to allocate a new object every time we join two rows.

Join also uses the type information as an argument to the [`getRow()` function](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/merge_join.go#L581), which uses it to [create an object](https://github.com/influxdata/flux/blob/d1fedd9b2d47f1ff3312bbcfff638d6f2523db63/stdlib/join/merge_join.go#L630) with the appropriate type and return it to the user.

I don't know of another way to construct the arrow arrays for the output chunk or for the default records without having that type information. Maybe I'm missing something, but I hope that explanation provides enough to context to point out anything that I've overlooked.

scbrickley:
Random thought that just occurred. I assume `map()` doesn't have this same problem, otherwise I expect people would be complaining about that too. I may take a look at that code to see how it handles empty tables, and maybe adapt join to do something similar.

scbrickley:
Before looking at any code, the main difference that would get in the way (I think) is that map doesn't need to construct default records.

scbrickley:
Also the fact that, if there's no input to map, I think it can just output an empty table and call it quits. I don't think it makes sense for join to do that.

PR #5220, opened 2022-09-22
docs(SPEC): add specification of editions

nathanielc:
This change adds the details of how Flux editions will work and their intent. One small parser change is proposed to enable an edition directly within the script.

### Checklist

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1086,7 +1086,7 @@ Attributes define a set of properties on source code elements.
     AttributeParameterList = AttributeParameter { "," AttributeParameter } .
     AttributeParameter     = PrimaryExpression
 
-The set of defined attributes and their meaning is left to the runtime to specify.
+The full set of defined attributes and their meaning is left to the runtime to specify.
 
 Example
 
@@ -1105,7 +1105,6 @@ A statement controls execution.
                    | ReturnStatement
                    | ExpressionStatement
                    | TestcaseStatement .
-
 
 
 #### Return statements
@@ -1411,6 +1410,12 @@ So, you have to:
 
 Flux provides a standard library of functions. Find documentation here https://docs.influxdata.com/flux/latest/stdlib/
 
+#### Experimental namespace
+
+Within the standard library there is an `experimental` package namespace.
+Packages within this namesapce are subject to breaking changes without notice.
+See the package documentation for more details https://docs.influxdata.com/flux/latest/stdlib/experimental/
+
 ### Composite data types
 
 A composite data type is a collection of primitive data types that together have a higher meaning.
@@ -1420,6 +1425,79 @@ A composite data type is a collection of primitive data types that together have
 A query specification defines what data and operations to perform.
 The execution model reserves the right to perform those operations as efficiently as possible.
 The execution model may rewrite the query in anyway it sees fit while maintaining correctness.
+
+## Versions and editions
+
+Flux follows a [semantic versioning scheme](https://semver.org/) such that breaking changes are clearly communicated as part of the version information.
+*Editions* allow for the introduction of breaking changes via an opt-in mechanism.
+
+Flux editions are a set of features that are enabled in Flux.
+If the edition is not enabled then the features are not enabled.
+These features would otherwise be breaking changes to Flux.
+An edition is explicitly opt-in.
+The pattern of editions allows users to migrate to new Flux versions without risk that their scripts will break.
+
+An edition is separate from a version.
+A version of Flux represents a single point in the commit history of the Flux source code.
+A user can only use a single version of Flux for a given script.
+Editions represent a set of features that are enabled for a given Flux version.
+So long as the Flux version supports all the features of the edition it can be enabled.
+A user can upgrade to a newer Flux version without being required to upgrade to the newest edition of Flux.
+
+New Flux features that require a breaking change to the syntax or semantics of Flux must always be part of a new edition.
+This means that a script can regularly update to the newest Flux version without risk of breaking because any breaking changes are explicitly opt-in.
+With editions the Flux community gets both a pattern where users can always be running the latest version of Flux and the ability to introduce new useful but otherwise breaking changes to Flux.
+
+We anticipate that there will be at most one new editions of Flux a year. A slow cadence of new editions means users have ample time to migrate to a new edition if desired.
+Even being a few years behind on editions should only mean a few migration steps in order to have access to features introduced in the newest edition.
+
+### First Edition
+
+The first edition of Flux is 2022.1.
+This first edition is the only minimum required edition of Flux.
+
+### Future Editions
+
+Editions will be named after the year in which they are created, with an added sequence number if more than one edition needs to be created in a single year.
+For example the first edition is `2022.1` because it is created in the year `2022` and is the first edition created in that year.
+
+### Editions are optional
+
+A Flux script may specify directly the edition it requires to function.
+Additionally the Flux runtime will allow for the edition to be specified out of band of the script, thus allowing for deployments of Flux to have control over the edition.
+
+#### Edition attribute
+
+The edition of the current script is specified as an `edition` attribute on the package with a single parameter, the name of the edition.
+It is an error for multiple files within a package or module to specify differing editions.
+
+Examples:
+
+Specify the edition with an explicit package clause
+
+    @edition("2022.1")
+    package math
+
+    add = (x,y) => x + y
+
+
+### Flux Editions and Modules
+
+Each Flux module may specify its own edition, therefore a Flux script on an earlier edition may import and consume a module that uses a newer edition of Flux.
+Naturally if a module exposes a new edition feature via its API, consumers of that module will be required to use at least that edition in order to directly consume the module.
+
+
+### Migrating to a new edition
+
+When a new edition is created a migration process will be provided to ease the migration from an older edition to a new edition.
+
+### Editions and Experimental
+
+Editions do not change the contract of the experimental package namespace.
+Experimental packages are still subject to breaking changes without notice.
+Most new features do not require a breaking change to Flux syntax or semantics.
+As such it will remain common for new packages to be introduced as experimental packages.
+When an API is stabilized, it can be promoted out of experimental without the need to create a new edition.
 
 ## Request and Response Formats
 


docs/SPEC.md
```diff
@@ -1364,6 +1370,82 @@ A query specification defines what data and operations to perform.
 The execution model reserves the right to perform those operations as efficiently as possible.
 The execution model may rewrite the query in anyway it sees fit while maintaining correctness.
 
+## Versions and editions
+
+Flux follows a [semantic versioning scheme](https://semver.org/) such that breaking changes are clearly communicated as part of the version information.
+Editions are also supported to allow for the introduction of breaking changes via an opt-in mechanism.
```
pierwill, line 1376:
```suggestion
*Editions* allow for the introduction of breaking changes via an opt-in mechanism.
```

docs/SPEC.md
```diff
@@ -1364,6 +1370,82 @@ A query specification defines what data and operations to perform.
...
+
+Flux editions are a set of features that are enabled in Flux.
+If the edition is not enabled then the features are not enabled.
+These features would otherwise be breaking changes to Flux.
+An edition is explicitly opt-in.
+The pattern of editions allows users to migrate to new Flux versions without risk that their scripts will break.
+
+An edition is separate from a version.
+A version of Flux represents a single point in the commit in the source code for Flux.
```
pierwill, line 1385:
```suggestion
A version of Flux represents a single point in the commit history of the Flux source code.
```

docs/SPEC.md
```diff
@@ -1364,6 +1370,82 @@ A query specification defines what data and operations to perform.
...
+A user can only consume a single version of Flux for a given script.
```
pierwill, line 1386:
```suggestion
A user can only use a single version of Flux for a given script.
```

docs/SPEC.md
```diff
@@ -1364,6 +1370,82 @@ A query specification defines what data and operations to perform.
...
+Even being a few years behind on editions should only mean a few migration steps in order to have access to features introduced in the newest edition.
+
+### First Edition
+
+The first edition of Flux will be 2022.1.
+This first edition will be the only minimum required edition of Flux and represents the set of features that exist in Flux today.
+
+### Future Editions
+
+Editions will be named after the year in which they are created with a sequence number for the occasion that more than one edition needs to be created in a single year.
```
pierwill, line 1405:
```suggestion
Editions will be named after the year in which they are created, with an added sequence number if more than one edition needs to be created in a single year.
```

docs/SPEC.md
```diff
@@ -1364,6 +1370,82 @@ A query specification defines what data and operations to perform.
...
+### Migrating to a new edition
+
+When a new edition is created a migration process will be provided to ease the migration process  from an older edition to a new edition.
+
+### Editions and Experimental
+
+Editions do not change the contract of the experimental package namespace.
+Experimental packages are still subject to breaking changes without notice.
+Most new features do not require a breaking change to Flux syntax or semantics.
+As such it will remain common for new packages to be introduced as experimental packages and then when their API stabilizes be promoted out of experimental without the need to create a new edition.
```
pierwill, line 1447:
```suggestion
As such it will remain common for new packages to be introduced as experimental packages.
When an API is stabilized, it can be promoted out of experimental without the need to create a new edition.
```

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+    EditionClause = "@" edition .
+    edition = year "." decimal_digit .
+
+It is an error for multiple files within a package specify differing editions.
+
+Example
+
+    package main
+
+    @edition 2022.2
```
Marwes, line 1426: We may want to consider expanding this syntax as a general sort of "attribute" which we can attach to items. If that is the case we have some additional considerations.

* How does the syntax work with multiple arguments?
```
@attribute 123 456 789 // This syntax does not indicate when the attribute ends
@attribute(123, 456, 789) // Maybe? 
```
* We probably want the attribute to accept different values. Literals (strings, integers, floats etc), identifiers, full expressions(?)
* I'd imagine that an attribute attaches to the item after it, but in the edition example above it is places after the `package`. Maybe it should be before it? Or maybe it should be different syntax to indicate that it is free-standing/applies to the item above it?

onelson: :+1: A use case this idea of generalizing `@attribute()` brings to mind is feature flagging, which is a similar problem space in terms of (possibly) needing to restart parsing.

docs/SPEC.md
```diff
@@ -102,7 +102,7 @@ The following character sequences represent operators:
     -   <    !~   [   ]   ^
     *   >    =~   {   }   ?
     /   <=   =    ,   :   "
-    %   >=   <-   .   |>
+    %   >=   <-   .   |>  @
```
jsternberg, line 105: I think it would probably be better if this wasn't an operator. It seems important to me that the `@` is a part of a specific metadata node. I wouldn't want `@ editions` to be a thing and I'd want us to require it to be `@editions`.

docs/SPEC.md
```diff
@@ -1024,7 +1024,7 @@ Flux source is organized into packages.
 A package consists of one or more source files.
 Each source file is parsed individually and composed into a single package.
 
-    File = [ PackageClause ] [ ImportList ] StatementList .
+    File = [ PackageClause ] [ EditionClause ] [ ImportList ] StatementList .
```
jsternberg, line 1027: Should metadata be before or after the package clause?

I think we might also want this to be more general in the future (a metadata list), but I'm fine with keeping this as specific as possible for now.

docs/SPEC.md
```diff
@@ -1354,6 +1354,12 @@ So, you have to:
 
 Flux provides a standard library of functions. Find documentation here https://docs.influxdata.com/flux/latest/stdlib/
 
+#### Experimental namespace
+
+Within the standard library there is an `experimental` package namespace.
```
jsternberg, line 1359: Be more specific that anything _under_ this package namespace is included too such as `experimental/date`. Also, be sure to mark that breaking changes may occur within the same edition and we reserve the right to remove values from older editions.

Basically be as explicit as possible for all the ways that we could not support something in experimental. Some might assume that, "Well, it's in experimental, but it's still there and this is an old edition and it's now in the formal standard library in a new edition so it's guaranteed not to be removed."

I want to dispel the "it's guaranteed to not be removed" part of that sentence. Everything else is logical and makes sense. And, in practice, nothing they say is a big problem and is 99% likely to be correct.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+## Versions and editions
+
+Flux follows a [semantic versioning scheme](https://semver.org/) such that breaking changes are clearly communicated as part of the version information.
+*Editions* allow for the introduction of breaking changes via an opt-in mechanism.
+
+Flux editions are a set of features that are enabled in Flux.
+If the edition is not enabled then the features are not enabled.
+These features would otherwise be breaking changes to Flux.
+An edition is explicitly opt-in.
+The pattern of editions allows users to migrate to new Flux versions without risk that their scripts will break.
```
jsternberg, line 1382: "their scripts or their dependencies will break"

This is implicit as part of the sentence but let's be explicit.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+
+Editions will be named after the year in which they are created, with an added sequence number if more than one edition needs to be created in a single year.
+For example the first edition is `2022.1` because it is created in the year `2022` and is the first edition created in that year.
+
+### Editions are optional
+
+A Flux script may specify directly the edition it requires to function.
+Additionally the Flux runtime will allow for the edition to be specified out of band of the script, thus allowing for deployments of Flux to have control over the edition.
+
+#### Edition clause
```
jsternberg, line 1413: 👍

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+
+    @edition 2022.2
+
+    import "array"
+
+    array.from(rows: [{_value: 1.0}])
+
+### Flux Editions and Modules
+
+Each Flux module may specify its own edition, therefore a Flux script on an earlier edition may import and consume a module that uses a newer edition of Flux.
```
jsternberg, line 1434: Module, package, or file? Our current syntax really only works if it's per file which is conflated with package in our current usage.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+Naturally if a module exposes a new edition feature via its API consumers of that module will be required to use at least that edition in order to directly consume the module.
```
jsternberg, line 1435: "via its API, consumers of that module"

Needs a comma.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+
+
+### Migrating to a new edition
+
+When a new edition is created a migration process will be provided to ease the migration process  from an older edition to a new edition.
```
jsternberg, line 1440: "is created, a migration process"

Also add "an automated migration process" to that.

There's also an extra space in here.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+
+### Editions and Experimental
+
+Editions do not change the contract of the experimental package namespace.
+Experimental packages are still subject to breaking changes without notice.
+Most new features do not require a breaking change to Flux syntax or semantics.
+As such it will remain common for new packages to be introduced as experimental packages.
+When an API is stabilized, it can be promoted out of experimental without the need to create a new edition.
```
jsternberg, line 1448: Good place to clarify that we can removal or change experimental functions from non-most recent editions if needed.

docs/SPEC.md
```diff
@@ -1364,6 +1370,83 @@ A query specification defines what data and operations to perform.
...
+An edition is separate from a version.
+A version of Flux represents a single point in the commit history of the Flux source code.
+A user can only use a single version of Flux for a given script.
+Editions represent a set of features that are enabled for a given Flux version.
+So long as the Flux version supports all the features of the edition it can be enabled.
+A user can upgrade to a newer Flux version without being required to upgrade to the newest edition of Flux.
+
+New Flux features that require a breaking change to the syntax or semantics of Flux will always be part of a new edition.
+This means that a script can regularly update to the newest Flux version without risk of breaking because any breaking changes are explicitly opt-in.
+With editions the Flux community gets both a pattern where users can always be running the latest version of Flux and the ability to introduce new useful but otherwise breaking changes to Flux.
```
onelson, line 1393: This line has me idly thinking about the maintenance impact of supporting multiple editions within the stdlib.

Will we need to manage multiple versions of stdlib, per edition or just on a package by package basis?

docs/SPEC.md
wolffcm: I wonder if we should think more broadly about annotations and design something that is extensible for future use. The syntax could be something like

```
AnnotationStmt = "@" identifier expr
```
Or we could create a new kind of token if we don't want forbid a space in `@edition`. 

Allowing an arbitrary expression would allow us to be flexible in the future e.g., with an object expression
```
@some_annotation {
  foo: "bar",
  num: 100.3
}
```
For a simple case like like specifying an edition, we might want to use a string:
```
@edition "2022.1"
```
We wouldn't need to do it this way right away, but I think it would be smart to choose a syntax that could be part of a more general feature of the language in the future.

docs/SPEC.md
nathanielc: I like `@attribute(123, 456, 789)` and it would make sense to be place before the `package` clause in that case. So it would look like this:

```
@edition(2022.1)
package foo
```

I think we can support literals in the arguments to attribute but arbitrary expressions seems like to much to a simple parsing pass.



> I recall we wanted to avoid giving end-users a way to manipulate flags, so I don't know where this leaves us, but if we could exploit this in testcases it would be nice.

I think we could find a way to make the attribute meaningless in prod but still work in the test harness. Or something along those lines.

docs/SPEC.md
nathanielc: Agreed

docs/SPEC.md
nathanielc: I think re discussion above it makes sense to be above.

docs/SPEC.md
nathanielc: I want it to be per module. So my thinking was the attribute must be set on the top level package for the module and if its set anywhere else its an error. Thoughts? This reduces to the most common case of per file for one file/package modules.

jsternberg: There's not really a concept of a top level package since a package is just a directory and there's no root other than the folder itself.

We kind of have a module system like Rust but the structure of a Go package. I wonder if it might be worth updating the structure of what a module looks like before we have user-defined modules to be more similar to Rust because we treat it like Rust anyway.

docs/SPEC.md
nathanielc: Would we also use this attributes to specify the polymorphic label types for builtins as well?

Something like:

```
@label_type((<-tables: stream[{A with C: B}], ?column: C) => stream[{C: B}] where A: Record, B: Numeric, C: Label)
builtin mean : (<-tables: stream[A], ?column: string) => stream[B] where A: Record, B: Record
```

How would we parse that? Maybe this isn't the same thing?

docs/SPEC.md
jsternberg: I think I'd prefer to avoid this as the metadata system should be as simple as possible and allowing any possible expression means that the value for a module can be any expression. Once you start to open that, you start to open the question of what should be allowed (such as variables).

It might be good to have this be a string though. The primary reason is related to the scanner. In general, it's nicer if the scanner doesn't have to include parsing context. If we do `@edition 2022.1` we either have to create a new token for "edition pattern" or we have to use existing tokens. If we use existing tokens, then we're implementing the edition as a parser and whitespace will end up being ignored which would allow something like `@edition 2022 . 1`. I don't think that would be good. If we build it into the scanner, then the scanner either needs a special mode that it enters when it's parsing edition (we do this for regexes) in order for the parser to be able to recognize: `EditionsClause = EditionKey EditionValue`.

I kind of prefer the look of `@edition 2022.1` but it might not be a technically good idea unless we use some kind of preprocessor stage that causes them to be parsed differently.

We might want to consider that though. The `@` symbols are more like metadata similar to pragmas and we may just want them to be treated similar to comments by the parser rather than as part of the parser. Doing this will generally detach the system from the parser making it easier to affect the parser.

docs/SPEC.md
jsternberg: Thinking about this more I agree we can get feature flags in this at the same point.

Although maybe just something simple like `@attribute Literal` as the parsing?

```
@edition "2022.1"
@feature{a: "foo", b: "bar"}
@arrayattribute[1, 2, 3]
```

Literals should cover basically everything we would need. Array literals too? We don't need to be too strict on the exact contents of the literals because whatever interpreters the attribute would be the one to evaluate it. No need to do arbitrary expressions either.

docs/SPEC.md
nathanielc: We will need to manage multiple versions of the stdlib for each edition. This is in the case that the stdlib API changes in an edition (i.e. label polymorphism).

docs/SPEC.md
```diff
@@ -1424,6 +1430,87 @@ A query specification defines what data and operations to perform.
...
+
+    import "array"
+
+    array.from(rows: [{_value: 1.0}])
+
+
+### Flux Editions and Modules
+
+Each Flux module may specify its own edition, therefore a Flux script on an earlier edition may import and consume a module that uses a newer edition of Flux.
+Naturally if a module exposes a new edition feature via its API, consumers of that module will be required to use at least that edition in order to directly consume the module.
```
wolffcm, line 1499: I am trying to understand backwards/forwards compatibility with respect to editions.

Suppose in 2023.1 we have polymorphic labels.

If a module exposes polymorphic labels in its API, then `main` will need to be at least on `2023.1`. This makes sense to me.

What if we have a module that is on `2022.1` and uses `keep()` in its implementation (i.e., does not use polymorphic labels), and `main` declares itself to be on `2023.1`? Is the runtime able to use the older version of `keep()` on the module but still use polymorphic labels in `main`?

wolffcm (approved):
I had some questions that could perhaps be clarified. Overall this looks great to me.

docs/SPEC.md
```diff
@@ -1424,6 +1430,87 @@ A query specification defines what data and operations to perform.
...
+
+Flux editions are a set of features that are enabled in Flux.
+If the edition is not enabled then the features are not enabled.
+These features would otherwise be breaking changes to Flux.
+An edition is explicitly opt-in.
+The pattern of editions allows users to migrate to new Flux versions without risk that their scripts will break.
+
+An edition is separate from a version.
+A version of Flux represents a single point in the commit history of the Flux source code.
+A user can only use a single version of Flux for a given script.
```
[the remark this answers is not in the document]
nathanielc: In OSS the user has limited control over the Flux version, i.e. they can choose to upgrade or not. But yes for the most part with how Flux is shipped users don't have much choice.

docs/SPEC.md
```diff
@@ -1424,6 +1430,87 @@ A query specification defines what data and operations to perform.
...
+Editions represent a set of features that are enabled for a given Flux version.
+So long as the Flux version supports all the features of the edition it can be enabled.
+A user can upgrade to a newer Flux version without being required to upgrade to the newest edition of Flux.
```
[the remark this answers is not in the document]
nathanielc: The idea is that is defines the contract with users. How would you word this part?

docs/SPEC.md
nathanielc: >  Is the runtime able to use the older version of keep() on the module but still use polymorphic labels in main?

Yes, because the only difference with `keep` should be syntactic/semantic. The behavior of the old and new versions of keep is the same. The invariant of editions is that its surface layer i.e. syntactic changes only. The runtime is tied to the version and only a single version is used for any given script. 

Maybe you can expand with an example what might break in this scenario?

docs/SPEC.md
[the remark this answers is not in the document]
wolffcm: What about

> The version of Flux used to execute a script will be predetermined by the environment (e.g., Cloud, or from a local install).
> Each edition will have a minimum Flux version where it is supported, and will be supported for all subsequent versions.
> Users can decide which edition (of those that are supported for their version of Flux) they want based on compatibility and the feature set that they want.

docs/SPEC.md
wolffcm: Ah, the difference being syntactic is the piece I was missing. As long as the underlying representation stays the same, like string arguments being translated to labels or vice versa, that would be fine.

PR #5221, opened 2022-09-23
fix: skip strict null logicals

onelson:
Not yet sure this is a problem or not, but the thinking is the feature flag check we do inside the new strict null logical evaluator is too expensive to tolerate.

Selecting the original evaluator will bypass the feature flag check so we can watch the perf impact and learn if it is truly too expensive :sweat: 

Testcases that depend on the (newer) strict null version have been skipped to allow the CI to pass.

### Checklist


scbrickley (approved):

skartikey:
`make fmt` would be required.  Do you want to mention the EAR reference here just for context?

onelson:
Wasn't sure it was polite to say the EAR out here.

PR #5222, opened 2022-09-23
refactor: Pass the context to the executor compiler

Marwes:
Should make it possible to enable the strict null logical ops when compiling the executor instead of branching inside the executor itself.

### Checklist

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1,13 +1,15 @@
 package compiler
 
 import (
+	"context"
+
 	"github.com/influxdata/flux/codes"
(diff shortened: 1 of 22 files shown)


Marwes:
A bit tedious to drag the context through everywhere here since this compiler should be removed eventually, but I'd expect the context to be wanted in whatever replaces it anyway so.

onelson (approved):
On a superficial level, this looks to me as just what is needed.

I'm not sure what we can do to verify the change other than to rely on the existing suite, and maybe merge it _after the next release_ so we can have a week to run it ourselves with the hope that anything lacking will surface during that time.

Issue #4296, opened 2021-11-23
[WIP/Proposal] Flux User Modules


nathanielc:
We want the Flux ecosystem to facilitate sharing Flux code. Currently this is not possible as Flux can only import code provided as part of the standard library. The proposal is to build the APIs necessary to host and share Flux code. Our plan is to limit the scope with which code can be shared and grow that scope as we learn more. An MVP of this proposal is to enable organizations to share Flux code within their own organization.

## Technical Design

There are two parts to sharing Flux code:

* Host Flux code in a registry
* Update Flux engine to be able to load code from the filesystem and from a registry

### Registry

The registry will need to store and serve Flux modules. An API has been previously designed and a PDF dump of its swagger doc can bee seen [here](https://github.com/influxdata/flux/files/7573125/registry_api.pdf). This API will very likely change but that gives an idea of what we are thinking.

The rough API will look like this:

| Method | Route | Description |
| ---------- | --------- |---------------- |
| GET | /modules | Return all modules owned by the user/org |
| POST | /modules | Create a new module by uploading a zip archive |
| GET | /module/{path} | Download a specific module |
| GET | /versions/{path} | Get a list of versions for a specific module |

#### Versioning

Flux modules will be immutable, i.e. cannot be deleted. To make a change to a module create a new version of that module. Version number will follow semantic versioning where a breaking change must increment the major version number. 

### Flux Engine

The Flux engine will need to support loading Flux code. This will be the process of downloading the module from the registry and including it during the execution of the Flux script.

We need to design a few aspects of this problem.

#### Contents of the module itself

A module must contain the code to run and other metadata in order integrate it into the runtime. This will likely exists as a zip archive of the semantic graph of the packages included in a module. Note a module may contain multiple packages.

#### Specification of dependencies

How does a Flux package define which versions of a module it wants to import? Some design considerations:

Flux is lightweight and a scripting language, we do not want to make it onerous on the author of a Flux script to define versions of a dependency. This means we will want clear sane defaults when a version is not specified.

#### Resolving versions of dependencies

Resolving a tree of dependencies will need to solve some important questions:

* Can we have multiple versions of the same Flux module included into the engine simultaneously?
* If not, how do we pick a version if we need to pick just one?
* How do we keep this simple?

[...]

PR #5224, opened 2022-09-23
docs(SPEC): add specification of modules

nathanielc:
The implementation of this work is still in-progress see https://github.com/influxdata/flux/issues/4296

### Checklist

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -1051,9 +1051,10 @@ The _main_ package is special for a few reasons:
 
     ImportDeclaration = [ Attributes ] "import" [identifier] string_lit
 
+
 Associated with every package is a package name and an import path.
 The import statement takes a package's import path and brings all of the identifiers defined in that package into the current scope under a namespace.
-The import statment defines the namespace through which to access the imported identifiers.
+The import statement defines the namespace through which to access the imported identifiers.
 By default the identifier of this namespace is the package name unless otherwise specified.
 For example, given a variable `x` declared in package `foo`, importing `foo` and referencing `x` would look like this:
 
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -1364,6 +1371,191 @@ A query specification defines what data and operations to perform.
...
+Flux modules are imported using an import declaration.
+The import path may contain specifiers about which versions of a module should be imported.
+
+When no version information is provided the latest version of the module is used.
+A _minimum_ version may be specified using `@` followed by a version number.
+An import may also specify the version `pre` which is the most recent pre-release version of the module.
+
+Examples
+
+    import "foo/module"             // imports the latest version 0.x or 1.x version of the module
```
Marwes, line 1410: This will be ambiguous with standard library modules as it is currently (did you mean the user module "array" or the standard library "array"?). For the existing standard library we can just reserve the names for all existing packages, but that won't work if we want to add a new packages so maybe we also want to add and reserve a `std` prefix for the standard library?

onelson: I've seen things in the JS space where importers use like `~/...` to say, "relative to the project's `node_modules` dir."

It could be useful to have something as a prefix to indicate the importer to use (registry vs stdlib).

docs/SPEC.md
```diff
@@ -1370,6 +1371,115 @@ A query specification defines what data and operations to perform.
...
+    foo/bar    // module path of foo/bar for version zero or one
+    foo/bar/v2 // module path of foo/bar for major version two
+
+### Module versions
+
+All modules are versioned using a [semantic version number](https://semver.org/) prefixed with a `v`, i.e. `vMAJOR.MINOR.PATCH`.
+Once a module version has been published it cannot be modified.
+A new version of the module must be published.
+
+### Importing modules
```
onelson, line 1399: I wonder about how users would elect to load from local disk. Would it be some prefix or edit in the import path (which would be a tedious workflow) or would we have a way to set custom "resolve" (as with npm) or "registry patches" (as with cargo)?

nathanielc: I think with the new specification this would look like

```
@registry("modules", "file://./fluxmods")
package main

import "modules/foo"
```

Now the URL for `foo` will be `file://./fluxmods/foo` and we can implement the logic to grab it off disk. The granularity here might be too coarse since its the entire registry but a `registry_patch` or similar attribute could be used in those cases. In short the fundamentals is we map registry names to URLs and those URLs can point to the local filesystem.
```

File: docs/SPEC.md [hunk 3]
```diff
@@ -1426,6 +1427,176 @@ A query specification defines what data and operations to perform.
...
+| GET    | $base/$module/@v/list         | Returns a list of known versions of the given module in plain text, one per line.                                                                          |
+| GET    | $base/$module/@v/$version.zip | Returns a zip file of the contents of the module at a specific version.                                                                                    |
+| GET    | $base/$module/@latest         | Returns the highest released version, or if no released versions exist the highest pre-release version of the given module in plain text on a single line. |
+| GET    | $base/$module/@pre            | Returns the highest pre-released version of the given module in plain text on a single line.                                                               |
+| POST   | $base/$module/@v/$version     | Publish a new version of the module where the POST body contains multipart formdata for the contents of the module.                                        |
```
nathanielc, line 1582: The `@v` delimiter is no longer necessary since a module path is at most two path elements, of the form `name/v#` where `#` is a number >= 2.  We needed a rule to distinguish module paths from package paths and it so happens we can reuse that rule for the registry API. 

Should we remove it?

nathanielc:
NOTE: This specifies two changes to the current implementation of the registry API.

* the `v` is required anywhere a version is used 
* the POST no longer expects a `.zip` extension.

docs/SPEC.md
onelson: Yeah, I think this works for me.

Though I have some misgivings about the file protocol appearing anywhere in the flux source code since _I think_ we'd need to start thinking about security implications.

Less concerned about this if there is a patching mechanism that has some guardrails around what paths are allowed for reading (before we even look at the source code).

docs/SPEC.md
wolffcm: How does `POST` and `@pre` work---is there a way to specify that the version you are publishing is release or pre-release?

wolffcm (commented):
This looks reasonable to me, but I had some questions about what it means to use `@pre`.

docs/SPEC.md
```diff
@@ -1426,6 +1427,180 @@ A query specification defines what data and operations to perform.
...
+
+When resolving an import path the first path element of the module path is compared against the defined registry names.
+If a match is found the import path is understood to be relative to the registry.
+If no match is found the import path is understood to be an import from the standard library.
+It is an error to specify a version on imports from the standard library.
+The standard library version is implicit to the runtime.
+
+When no version information is provided the latest version of the module is used.
+A _minimum_ version may be specified.
+An import may also specify the version `pre` which is the most recent pre-release version of the module.
```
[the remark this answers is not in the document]
nathanielc: I am expecting that `@pre` refers to an explicit pre-release using semantic versioning for [pre-release versions](https://semver.org/#spec-item-9). I am expecting that clients (i.e. the UI) will leverage pre-releases to `pre-publish` or `preview` a module before doing a _real_ publish of the module. This way users can test their modules without risk of breaking existing consumers.

I imagine a workflow like this:

1. Publish a module `foo`
2. Consume module `foo` in various other scripts
3. Decide to improve `foo`
4. Publish a pre-release of `foo`.
5. Use `@pre` is specific scripts to ensure its working
6. Publish a normal release of foo, all other scripts are now upgraded

This work across major versions as well as `@pre` applies to within a single major version. Meaning there can be a different pre-release for each major version.

nathanielc:
> How does POST and @pre work---is there a way to specify that the version you are publishing is release or pre-release?

The `pre-release` is part of the semantic version spec. Therefore it becomes straightforward to create a pre-release when publishing by specifying a version string that is a pre-release.

docs/SPEC.md
```diff
@@ -1460,7 +1460,7 @@ Module paths need not be unique across different registries.
 
 #### Registry attribute
 
-The `registry` attribute defines the available registries.
+The `registry` attribute defines the available registries and must preceed a package clause.
```
wolffcm, line 1463: Just a nit, typo in the word "precede".

wolffcm (approved):
Looks great.

PR #5230, opened 2022-09-26
fix: vectorized float should propagate nulls

onelson:
### Checklist

### DIFF
--- a/stdlib/universe/map_vectorize_float_test.flux
+++ b/stdlib/universe/map_vectorize_float_test.flux
@@ -4,6 +4,7 @@ package universe_test
 import "array"
 import "testing"
 import "testing/expect"
+import "internal/debug"
 
 testcase vec_with_float {
     expect.planner(rules: ["vectorizeMapRule": 1])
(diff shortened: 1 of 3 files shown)


stdlib/universe/map_vectorize_float_test.flux
```diff
@@ -91,3 +92,38 @@ testcase vec_with_float_const {
...
+    want =
+        array.from(rows: [{x: 1}])
+            |> debug.opaque()
+
+    got =
+        array.from(rows: [{x: 1}])
+            |> debug.opaque()
+            |> map(fn: (r) => ({r with output: float(v: r._null)}))
+
+    testing.diff(want, got)
```
Marwes, line 128: You could use `assertEqualValues` instead so you do not have to construct the streams manually just to compare one value.

Marwes (approved):

stdlib/universe/map_vectorize_float_test.flux
onelson: Yeah, it seems like if I wanted to use `assertEqualValues` for this case, I'd need to also leverage `tableFind` et al. I think I'd rather leave it as-is.

Largely the point of these two testcases is to test the null interaction of `_vecFloat` which is only possible via `map()` right now.

PR #5234, opened 2022-09-28
docs(SPEC): add specification of attributes

nathanielc:
Attributes are a generic way to add early metadata to a Flux script. Attributes are only supported before package clauses and testcase statements As testcase statements were not defined in the SPEC they have been added.

This will be the mechanism we use to define editions. See https://github.com/influxdata/flux/pull/5220

### Checklist


docs/SPEC.md
```diff
@@ -1105,6 +1128,41 @@ Examples:
...
+
+    import "testing"
+
+    testcase addition {
+        testing.assertEqualValues(got: 1 + 1, want: 2)
+    }
+
+An example of test case extention to validate a feature does not regress existing behavior.
+
+    @feature({vectorization: true})
```
nathanielc, line 1161: Should we make features of Flux explicitly boolean? Meaning you can't have a feature accept an arbitrary value like our system feature flags do? 

This would change to be
```
@feature("vectorization")
```

Naming the feature is then sufficient to know its should be set to true. 

Thoughts?

docs/SPEC.md
```diff
@@ -102,7 +102,7 @@ The following character sequences represent operators:
     -   <    !~   [   ]   ^
     *   >    =~   {   }   ?
     /   <=   =    ,   :   "
-    %   >=   <-   .   |>
+    %   >=   <-   .   |>  @
```
jsternberg, line 105: I think we should avoid calling this an operator but also why are quotes in here too?

We might want to review exactly what we're trying to communicate with this section. It seems we're trying to say that these are characters that have special meaning lexically but calling them operators.

docs/SPEC.md
```diff
@@ -1086,6 +1078,37 @@ A package may reassign a new value to an option identifier declared in one of it
 A package cannot access nor modify the identifiers belonging to the imported packages of its imported packages.
 Every statement contained in an imported package is evaluated.
 
+### Attributes
+
+Attributes define a set of properties on the subsequent source code element.
+
+    Attributes             = { Attribute } .
+    Attribute              = "@" identifier AttributeParameters .
+    AttributeParameters    = "(" [ AttributeParameterList [ "," ] ] ")" .
```
jsternberg, line 1087: Thinking about this more, I think this should be forced to be a property list. At the moment, we wouldn't be able to update the edition attribute without it being its own breaking change because it takes a string literal. Maybe we should do it like:

```
@edition(name: "2022.1")
```

Thoughts?

docs/SPEC.md
jsternberg: Features can be arbitrary types so I think it's better as is.

docs/SPEC.md
nathanielc: yeah its a bit confusing, these are the set of lexical tokens. Perhaps we should refactor that section.

docs/SPEC.md
nathanielc: Hmm still thinking, my initial thoughts are that it make the edition specifically quite verbose. Also once we introduce the first edition parsing we can't change it since we need to support old parsers being able to understand it.

jsternberg: Maybe it's fine as-is? Since object literals exist we can still support the property lists and the parser just needs to be very flexible.

I'm going to say it's fine as-is now that I think about it. I was also going to propose removing the commas to have multiple attribute parameters, but let's keep that in the grammar for extra flexibility and then never use it.

jsternberg (approved):

docs/SPEC.md
onelson: I think I agree.

Furthermore, for vectorization the need was often to test with multiple related flags together. If we opted for `@feature("name")`, I think we should be able to stack these attributes:

```
@feature("vectorizedMap")
@feature("vectorizedLogicalOps")
@feature("vectorizedConditionals")
testcase something_with_vec_cond_and_logical {
  // ..
}
```

I like this less than being able to specify a map of features I want to enable.

jsternberg (approved):

Issue #5142, opened 2022-08-27
Add `dynamic.asArray` function


onelson:
```
builtin asArray(<-v: Dynamic) => [Dynamic]
```

In order to allow the functions in the `array` package (`map`, `filter`) to consume `Dynamic` values, they should first be cast to an array: `[Dynamic]`.

The functions in the `array` package themselves will need to be updated to handle `Dynamic` elements.

This is to say, `array.map` and `array.filter` will require arrays as input, but the element type may be `Dynamic`. Output of `[Dynamic]` should also be possible.

> **TBD:** need to give some consideration to the error handling story. In #5143 the plan is to introduce casting functions which can offer a default value in the case of an error.
> A case can be made for both producing hard errors when the input _is not an array_ versus producing an empty `[Dynamic]`. Need to figure out how to present this choice in our API.

DOD:
- Add new builtin to cast a `Dynamic` to `[Dynamic]`.
- ~~Update implementation of stdlib functions accepting arrays.~~ Assumed to not be needed since just making sure the outer element is an array should make these work.
- **TBD** Determine the error handling story and implement accordingly.

PR #5239, opened 2022-09-28
feat(dynamic): add `dynamic.asArray` builtin

onelson:
Fixes: #5142

This diff adds a new builtin to the `experimental/dynamic` package which converts a `dynamic` value into an array of `dynamic` elements.

In order to test this, I had to get a little creative and added an additional builtin called `dynamic._isNotDistinct` which uses the ~~`Value.Equal` method~~ `display()` function to compare 2 dynamics. This was made necessary by the lack of other cast functions (still pending implementation) and the fact `dynamic` is not permitted as table column data. This means we _must convert_ dynamic values to _something else_ before we can assert via `testing.diff`.

While testing, I ran into issue with the initial type-system work for dynamic (#5212); multiple dynamic values could not be used together because of a missing bit (they failed to unify). A small test was added at the type-system level, and the new rule added. This allowed tests like the "heterogeneous array" one to be written.

Some small refactors were also made to `dynamic.dynamic` which landed in #5242. Specifically it was updated to not rewrap values which were already dynamic. The original rewrap behavior caused me a confusion earlier in this PR, though I don't think you'd see evidence of it at this stage.

### Checklist


stdlib/experimental/dynamic/dynamic.go
```diff
@@ -0,0 +1,54 @@
+package dynamic
```
wolffcm, line 1: I have a similarly named file on my branch, but it's just for JSON-related functions, I'll rename my file to `json.go` to avoid conflicts.

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -0,0 +1,29 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+builtin asArray : (<-v: dynamic) => [dynamic]
+
+// Convert a value into a dynamic.
+//
+// ## Metadata
+// introduced: NEXT
+builtin dynamic : (v: A) => dynamic
```
wolffcm, line 29: Looking at some other packages in the stdlib, if there is a package-scoped `introduced` doc attribute, the package members seem inherit it if they don't have an `introduced` attribute of their own.

stdlib/experimental/dynamic/dynamic.go
onelson: I actually stubbed these files out in separate commits so I could drop them (assuming your PR landed ahead of mine). Might now be an issue :shrug:

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,13 @@ package dynamic
 // ## Metadata
 // tags: type-conversions
 builtin dynamic : (v: A) => dynamic
+
+// asArray attempts to convert a dynamic value into an array of dynamic elements.
+// If the value is not an array, `asArray` will result in a fatal error.
```
sanderson, line 19:
```suggestion
// asArray converts a dynamic value into an array of dynamic elements.
//
// The dynamic input value must be an array. If it is not an array, `dynamic.asArray()` returns an error.
```

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,13 @@ package dynamic
...
+//
+// ## Parameters
+// - v: Dynamic value to convert. Defaults is the piped-forward value (`<-`).
```
sanderson, line 22:
```suggestion
// - v: Dynamic value to convert. Default is the piped-forward value (`<-`).
```

sanderson (approved):
👌

stdlib/experimental/dynamic/dynamic_test.flux
```diff
@@ -10,3 +11,56 @@ testcase dynamic_not_comparable {
...
+
+    testing.assertEqualValues(
+        want: true,
+        got: dynamic._equal(a: dynamic.dynamic(v: 123), b: arr[0]),
+    )
+}
+
+// This is similar to the "actual array" test except that the elements in the
+// array are not wrapped in dynamic ahead of time. asArray therefore needs to
+// wrap the elements prior to producing the `[dynamic]` it'll return.
```
wolffcm, line 42: I would expect it to be a bug for `dynamic.jsonParse` a dynamic array whose elements weren't also wrapped. Maybe `asArray` should assert in this case?

On the other hand, I can also imagine other sources of dynamic data that, unlike JSON, are able to guarantee homogeneity of arrays. In that case, _maybe_ it might make sense for `asArray` to leave the elements themselves unwrapped and produce a `[int]` or whatever?

I guess what I'm saying is that it seems like auto-wrapping in `asArray` could cover up bugs elsewhere in the stdlib (if we're being strict about wrapping), or it could be making things needlessly difficult (if we're not). Thoughts?

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +27,94 @@ var dynamicConv = values.NewFunction(
...
+	runtime.MustLookupBuiltinType("experimental/dynamic", "asArray"),
+	func(ctx context.Context, args values.Object) (values.Value, error) {
+		arguments := interpreter.NewArguments(args)
+		v, err := arguments.GetRequired("v")
+		if err != nil {
+			return nil, err
+		}
+
+		if v.IsNull() {
+			panic("TODO")
```
wolffcm, line 55: Should this just be an error as below? Or propagate the `null`?

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +27,94 @@ var dynamicConv = values.NewFunction(
...
+				v := arr.Get(i)
+				v.Retain()
+				elems[i] = values.NewDynamic(v)
+			}
+			dynArr := values.NewArrayWithBacking(semantic.NewArrayType(semantic.NewDynamicType()), elems)
+			return dynArr, nil
+		}
+	},
+	false,
+)
```
wolffcm, line 90: Seeing calls to `Retain` here makes me wonder if we would catch bugs related to reference counts in our acceptance tests? I know that we can catch them with Go unit tests by using a special allocator and assert there is no more allocated bytes at the end of the test. We should do something similar when we run our acceptance tests. I don't know if we do that or not.

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +27,94 @@ var dynamicConv = values.NewFunction(
...
+	"_equal",
+	runtime.MustLookupBuiltinType("experimental/dynamic", "_equal"),
+	func(ctx context.Context, args values.Object) (values.Value, error) {
+		arguments := interpreter.NewArguments(args)
+		a, err := arguments.GetRequired("a")
+		if err != nil {
+			return nil, err
+		}
+		if a.IsNull() {
+			return values.NewBool(false), nil
```
wolffcm, line 102: It seems like for testing purposes, you might want to consider nulls as equal?

wolffcm (commented):
This looks great, but I had questions about whether or not `asArray` should autowrap its elements.

stdlib/experimental/dynamic/dynamic.go
onelson: Well color me embarrassed. Let me clean this up :cold_sweat: Yeah, it should be something like that.

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -2,13 +2,19 @@ package dynamic
...
 	"github.com/influxdata/flux/interpreter"
 	"github.com/influxdata/flux/runtime"
+	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
 )
 
 func init() {
 	runtime.RegisterPackageValue("experimental/dynamic", "dynamic", dynamicConv)
+	runtime.RegisterPackageValue("experimental/dynamic", "asArray", asArray)
+	runtime.RegisterPackageValue("experimental/dynamic", "_equal", equal)
```
scbrickley, line 17: It seems odd to me that there's no op spec or procedure spec for these new functions. I wonder how the planner handles the nodes without them? Do we know how these show up in a planner graph?

stdlib/experimental/dynamic/dynamic.go
onelson: iirc in the spec we say that nulls are never equal, but you're right that it may make sense in this case :thinking:

stdlib/experimental/dynamic/dynamic.go
onelson: I think those are only used for functions that operate on tables, but I could be totally wrong.

My intuition was that they are the bits that connect the stream of operations together. Since dynamic cannot be added to tables, it can't appear in table streams, therefore no op spec or what have you. All this work happens in the interpreter, which I guess is handled differently.

stdlib/experimental/dynamic/dynamic_test.flux
wolffcm: Thinking some more about this, the signature of `asArray` is `(<-v: dynamic) => [dynamic]` so it _needs_ to produce a `[dynamic]`. I think `asArray` should produce an error rather than wrap---but still not 100%.

stdlib/experimental/dynamic/dynamic.go
wolffcm: Yes, you're right @onelson. No need for an OpSpec for these, since they operate only on scalar values. They are just like the function in the `strings` package for example, `strings.substr`, etc.

scbrickley: Ah, I see. Just a misunderstanding on my part then. I don't think I've ever added a function that wasn't manipulating table streams, so that stuck out to me. But it makes sense that we wouldn't need those for functions that manipulative individual scalars/collections.

stdlib/experimental/dynamic/dynamic_test.flux
onelson: I've been mulling this over - the problem of who's responsibility is is to wrap. 

From my side I was trying to weigh whether or not the wrapping should happen in the initial `dynamic()` call when it receives an `[int]`. I opted to do the check on the `toArray` side to defer the extra work of wrapping each element until it would be needed.

If we say that `toArray` returns `[A]` instead of `[dynamic]` the extra wrapping could be skipped entirely for the homogeneous case but we sacrifice enforcement of the static types matching reality, I think.

The one thing I can say is if the caller sends in an `[int]` I would like for them to not need to manually wrap each element themselves in the script. That would be punitive.

stdlib/experimental/dynamic/dynamic.go
onelson: This is a good idea. Right now a lot of the vectorization tests are partially duplicated in Go just to verify the memory accounting. We currently don't.

I don't know that it'll be a big effort, but I'd like to file a separate issue to set that up. I suspect the smaller lift is just to add a set of Go tests, but it'd be great to not have to do that.

scbrickley (approved):
Apart from what @wolffcm has already pointed out, this looks good to me.

stdlib/experimental/dynamic/dynamic.go
onelson: Starting out with an error following the "error if the input is not an array" policy. 
I'm not sure our static checks save us from untyped nulls getting through to a receiver like this, so I guess the check is still necessary.

stdlib/experimental/dynamic/dynamic.go
onelson: I filed this as https://github.com/influxdata/flux/issues/5251

stdlib/experimental/dynamic/dynamic.go
onelson: Okay, so here's what I have rationale-wise.

The `a.IsNull()` check here is for the "untyped null" case where the argument `a` is null and didn't come from some known-typed column value. This is to say, the argument isn't even a `dynamic`.

I would expect the outcome to match whatever I got from nulls that are properly wrapped in a `dynamic`.

```
$ go run ./cmd/flux
> import dyn "experimental/dynamic"
> import "internal/debug"
> untypedNull = dyn.dynamic(v: debug.null())
> dyn._equal(a: untypedNull, b: untypedNull)
false
> intNull = dyn.dynamic(v: debug.null(type: "int"))
> dyn._equal(a: intNull, b: intNull)
false
```

For dynamic-wrapped nulls, both typed and untyped, I see `false` where `a` and `b` are the same flavor of null.

I think this makes sense if we're talking about value equality while disregarding identity. My discussion around "equivalency" earlier during show and tell obviously doesn't hold true here since I'd argue these are indeed equivalent but not equal.

stdlib/experimental/dynamic/dynamic_test.flux
onelson: > It seems like you're suggesting that we can think of dynamic(v: ...) as producing a dynamic value where any collections also have their values wrapped with dynamic, but that as an implementation detail we are not wrapping them. We are wrapping them lazily upon access as an optimization.

This is precisely what I was thinking. Clearly there are situations where we can't afford to be lazy, specifically the `json.parse` which needs to establish the element type for arrays it produces as it builds them.

I think I agree that symmetry is desirable and so I think that means adding a recursive structure to `dynamic(v: ...)`. Perhaps there should be some code share between this and the json parser since I imagine the job is effectively the same in both cases.

If this sounds right, I can relocate my naive array wrapping code into `dynamic(v:...)` as a stop-gap and we can coalesce once both our branches have landed.

> I'm wondering about the use case of dynamic(v: ...). Do we ever expect it to be used outside of testing?

The original use case I identified was in building json bodies to send out via http POST.

stdlib/experimental/dynamic/dynamic.go
wolffcm: It shouldn't be possible to pass a non-dynamic value to `_equal` because of its type signature, right?

Is `_equal` explicitly just for testing? If so, this gets an empty diff, which i think suggests `dynamic._equal` should follow suit:
```
> import "internal/debug"
> import "testing"
> testing.assertEqualValues(want: debug.null(type: "int"), got: debug.null(type: "int"))
Result: _result
>
```

onelson: Interesting.

Yes, my understanding is static checks would have failed us if we pass values that are not dynamic and reach that particular branch in the impl. Today I can only think it would have to be some sort of programmer error on the Go side, or otherwise refactors to how we treat dynamic statically.

`_equal` is explicitly for testing, yes. I'd expect once we have the rest of the cast functions in place, we'd use those to write more regular acceptance tests that compare tables.

_If we want to say nulls are equal_ we would need to be ready to no longer rely on `Value.Equal()` method and do something else. I can patch in a naive check to see if a and b themselves are both null, but I think if the nulls are deeper down in the value we'll get the wrong answer via `Value.Equal()`.

wolffcm: What about:
```js
_equal = (a, b) => display(v: a) == display(v: b)
```
Two dynamic values would be equal if their display strings are equal. We could call it something less loaded like `_isNotDistinct` to avoid confusion with `==` semantics in Flux.

As an aside, do we want to allow `==` to be used with dynamic values? It seems like maybe we want to produce an error for that, to force users to cast first?

onelson: Yeah, I think leveraging display is a good way to go. One issue might be field ordering of records in the output but if we hit that case for some reason, fixing it in `display()` will have bigger benefit.

I'll rebase and get that going. I can even move the definition of this into the test package, rather than exposing it. Win-win.

onelson: (actually, I'll leave it in the main package for now since it might be useful in the REPL as we hack through this stuff)

stdlib/experimental/dynamic/dynamic_test.flux
wolffcm: It sounds right to me. We can change things later on if we need, since I imagine all this will be in `experimental` for a while.

wolffcm (commented):
I still have some questions about what the behavior of `dynamic` should be. But maybe that's a part of a follow-up? I wasn't clear on that.

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +26,72 @@ var dynamicConv = values.NewFunction(
...
+		// Nothing to do if the incoming value is already a dynamic
+		if v.Type().Nature() == semantic.Dynamic {
+			v.Retain()
+			return v, nil
+		}
+
+		// FIXME(onelson): wrap the value recursively, not just if v is an Array
+		//  We want to produce a dynamic where every single value contained
+		//  within it is also wrapped with a dynamic.
+		if v.Type().Nature() == semantic.Array {
```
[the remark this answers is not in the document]
onelson: You are correct, but I'm calling this out of scope for the `asArray` issue. I only added an initial impl of the `dynamic()` builtin to allow me to write tests in the first place :weary: 

My hope was to address this fully in the follow-up #5252

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +26,72 @@ var dynamicConv = values.NewFunction(
...
+			arr := v.Array()
+			elmType, err := arr.Type().ElemType()
+			if err != nil {
+				return nil, err
+			}
+			if elmType.Nature() == semantic.Dynamic {
+				arr.Retain()
```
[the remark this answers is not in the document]
onelson: That's interesting to learn. My understanding was that `Value`s in flux should be retained if they are passed along in the output, but maybe that's just a special case for vectorization. I've had less exposure to the interpreter side so maybe the rules are different here.

I'll rip these calls out. If we happen to see a panic arise in the coming weeks we'll know better.

stdlib/experimental/dynamic/dynamic.go
```diff
@@ -21,10 +26,72 @@ var dynamicConv = values.NewFunction(
...
+				return values.NewDynamic(arr), nil
+			} else {
+				elems := make([]values.Value, arr.Len())
+				for i := 0; i < arr.Len(); i++ {
+					v := arr.Get(i)
+					v.Retain()
+					elems[i] = values.NewDynamic(v)
```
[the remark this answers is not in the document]
onelson: Oh, I see what you're saying. I was mostly hoping we wouldn't need to both write the same recursive wrapper code then pick a winner. I was more thinking I'd patch in what I needed to make `asArray` work then we'd refactor to leverage work done in your branch, but move it into `NewDynamic` so we'd both benefit.

I can built this out for a full implementation if you want, but I wasn't planning on it.

wolffcm (approved):
Looks good. I'm totally fine with saving any further to the `dynamic` cast for later.

PR #5256, opened 2022-10-03
feat: iox.sql function

### LINKED ISSUE #5203: IOx SQL function
in order to provide cross-compatibility with automated tasks and visualizations, a iox.SQL function would be useful: 

```
iox.SQL(bucket: "my-bucket", query: "select (input - capacity) AS difference from measurement1 where field1 > 0") 
|> map( fn: (r) => {
  return if r.difference < 0  
  then sendSlackMessage(message: "input exceeding current capacity") 
  else sendSlackMessage(message: "input OK") 
})
```
the intent here is for users who have a collection of SQL queries to be able to use them directly without having to translate them to flux.  

This may also be a way for the UI to utilize SQL in visualizations that are configured to read flux's annotated CSV output.  

A difference between this function and the regular sql.from()  function is that we will use our internal authorization mechanisms to confirm that the user has privileges to read the bucket. 

DOD:

 - A function experimental/iox.sql exists that can submit SQL to IOx over the Arrow Flight API
 - The iox.sql function will authorize the query based on bucket read permissions only (hence the proposal to have an explicit bucket parameter on the function)
 - Add Flux acceptance tests within the iox integeration tags

jsternberg:
This adds an `iox.sql` function to `experimental/iox`. This function is
powered by the iox dependency which defaults to an error provider.

There's presently no default implementation of this dependency in this
repository.

Fixes #5203.

### Checklist


stdlib/experimental/iox/iox.flux
```diff
@@ -18,3 +18,16 @@ package iox
 // ## Metadata
 // tags: inputs
 builtin from : (bucket: string, measurement: string) => stream[{A with _time: time}] where A: Record
+
+// sql will execute an sql query against the selected bucket in an IOx storage node.
```
sanderson, line 22:
```suggestion
// sql executes an SQL query against a bucket in an IOx storage node.
```

sanderson (approved):
Approved, with one minor suggestion to the docs.

Marwes (approved):

PR #5259, opened 2022-10-03
feat: add JSON functions that work with dynamic values

### LINKED ISSUE #5141: Add `Dynamic` version of `json` functions
The `experimental/json` package has some functions which will need to be adjusted to account for the new `Dynamic` type.

In order to avoid breaking changes during this experimental period for `Dynamic`, new functions should be added under `experiemental/dynamic` to mirror the functions in `experimental/json` and `json`.

Note that the work for the new `json.parse()` can't be a wrapper around the existing with an updated signature -- the implementation for `toValue()` will differ. It needs to be updated to build arrays with Dynamic elements instead using the element type of the first (non-invalid) element.

Similar work may also be required for the new dynamic version of `json.encode()`.

Ref: https://github.com/influxdata/flux/blob/d6a97eff2123bf75323bad1319ff10751177c445/stdlib/experimental/json/json.go#L56-L72

DOD:
- add a version of `json.parse()` which returns a `Dynamic`
- add a version of `json.encode()` can receive `Dynamic`

wolffcm:
This PR adds two new functions to the `experimental/dynamic` stdlib package:
- `jsonParse` which produces dynamic values.
- `jsonEncode` which encodes JSON bytes into dynamic values.

I had to cherry-pick a commit from #5239 to fix type inference to allow calling functions that accept `dynamic` arguments. I also found an issue where `map` does not deal with `dynamic` types, but this was an easy one-line fix.

Fixes #5141.

### Checklist


stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,24 @@ package dynamic
 // ## Metadata
 // tags: type-conversions
 builtin dynamic : (v: A) => dynamic
+
+// jsonParse takes JSON data as bytes and returns dynamic values.
+//
+// JSON input is converted to dynamic-typed values which may be converted to
+// a statically typed value with asArray or casting functions in this package.
```
sanderson, line 21:
```suggestion
// a statically typed value with `dynamic.asArray()` or casting functions in the `dynamic` package.
```

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,24 @@ package dynamic
...
+// tags: type-conversions
+builtin jsonParse : (data: bytes) => dynamic
+
+// jsonEncode converts a dynamic value into JSON bytes.
+//
+// ## Parameters
+// - v: value to be encoded into JSON
+//
+// ## Metadata
+// tags: type-conversions
```
nathanielc, line 36: nit, can you add the `introduced` metadata entry here.

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,24 @@ package dynamic
...
+// jsonParse takes JSON data as bytes and returns dynamic values.
+//
+// JSON input is converted to dynamic-typed values which may be converted to
+// a statically typed value with asArray or casting functions in this package.
+//
+// ## Parameters
+// - data: JSON data (as bytes) to parse.
+//
+// ## Metadata
+// tags: type-conversions
```
nathanielc, line 27: and same here.

nathanielc (approved):
LGTM, with two small fixes

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -14,3 +14,24 @@ package dynamic
...
+builtin jsonParse : (data: bytes) => dynamic
+
+// jsonEncode converts a dynamic value into JSON bytes.
+//
+// ## Parameters
+// - v: value to be encoded into JSON
```
sanderson, line 33:
```suggestion
// - v: Value to encode into JSON.
```

stdlib/experimental/dynamic/dynamic.flux
sanderson, line 27:
```suggestion
// tags: type-conversions
// introduced: NEXT
```

stdlib/experimental/dynamic/dynamic.flux
sanderson, line 36:
```suggestion
// tags: type-conversions
// introduced: NEXT
```

sanderson (commented):
Some minor doc suggestions.

libflux/flux-core/src/semantic/types.rs
```diff
@@ -882,6 +882,8 @@ impl MonoType {
             // An error has already occurred so assume everything is ok here so that we do not
             // create additional, spurious errors
             (MonoType::Error, _) | (_, MonoType::Error) => (),
+            // Similarly if both sides are dynamic, there's nothing more to do.
+            (MonoType::Dynamic(_), MonoType::Dynamic(_)) => (),
```
onelson: #5239 landed moments ago, so you should be able to drop that cherry-picked commit :crossed_fingers:

stdlib/experimental/dynamic/json.go
```diff
@@ -0,0 +1,229 @@
...
+		return v.Time().Time().Format(time.RFC3339Nano), nil
+	case semantic.Duration:
+		return v.Duration(), nil
+	case semantic.Regexp:
+		return v.Regexp().String(), nil
+	case semantic.Array:
+		arr := v.Array()
+		a := make([]interface{}, arr.Len())
+		var rangeErr error
+		arr.Range(func(i int, v values.Value) {
```
onelson, line 155: I wondered whether I should use `Array.Range` or just use a plain loop while I was looking at `asArray`. I opted for "plain loop" just because it seemed more minimal. What made you prefer it?

My guess: in this function we're iterating over different collection types (array, dict, object) and the `Range()` method might make this work more uniform.

onelson (approved):
Looks great. Love to see these good tests!

sanderson (approved):
👌🏻

stdlib/experimental/dynamic/json.go
wolffcm: This code I mostly copied from the non-dynamic implementation of `encode`. I only had to add the case to `semantic.Dynamic` above. So it was like this already. Looking at the alternative, I can't say that I have a strong preference.

I considered just changing the regular version of `json.encode` to be able to handle `dynamic`, since it's an additive change that shouldn't break with existing behavior, but decided it would be more prudent to copy it.

PR #5264, opened 2022-10-04
refactor(deps): Add private http client

scbrickley:
Adds the `PrivateClient` struct which wraps the default client. Behaves identically to the default client except that it obscures error messages that may contain sensitive information.

Also exposes a `PrivateHTTPClient()` method to the `Dependencies` struct. If an `HttpProvider` is created and no host is specified in the config, this method is called in place of `HTTPClient()`.

### DIFF
--- a/dependencies.go
+++ b/dependencies.go
@@ -3,6 +3,7 @@ package flux
 import (
 	"context"
 	"net"
+	nethttp "net/http"
 	"syscall"
 	"time"
 
@@ -26,6 +27,7 @@ const dependenciesKey key = iota
 type Dependencies interface {
 	Dependency
 	HTTPClient() (http.Client, error)
+	PrivateHTTPClient() (http.Client, error)
 	FilesystemService() (filesystem.Service, error)
 	SecretService() (secret.Service, error)
 	URLValidator() (url.Validator, error)
@@ -49,6 +51,14 @@ func (d Deps) HTTPClient() (http.Client, error) {
 		return d.Deps.HTTPClient, nil
 	}
 	return nil, errors.New(codes.Unimplemented, "http client uninitialized in dependencies")
+}
+
+func (d Deps) PrivateHTTPClient() (http.Client, error) {
+	c, err := d.HTTPClient()
+	if err != nil {
+		return nil, err
+	}
+	return newPrivateClient(c), nil
 }
 
 func (d Deps) FilesystemService() (filesystem.Service, error) {
@@ -135,4 +145,22 @@ func GetDialer(ctx context.Context) (*net.Dialer, error) {
 		KeepAlive: 30 * time.Second,
 		Control:   control,
 	}, nil
+}
+
+// privateClient is an http client that obscures error messages that may contain
+// sensitive information
+type privateClient struct {
+	client http.Client
+}
+
+func newPrivateClient(c http.Client) http.Client {
+	return &privateClient{client: c}
+}
+
+func (c *privateClient) Do(req *nethttp.Request) (*nethttp.Response, error) {
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.New(codes.Internal, "an internal error has occurred")
+	}
+	return resp, nil
 }
--- a/dependencies.go
+++ b/dependencies.go
@@ -160,7 +160,7 @@ func newPrivateClient(c http.Client) http.Client {
 func (c *privateClient) Do(req *nethttp.Request) (*nethttp.Response, error) {
 	resp, err := c.client.Do(req)
 	if err != nil {
-		return nil, errors.Wrap(codes.Internal, "an internal error has occurred", err)
+		return nil, errors.Wrap(err, codes.Internal, "an internal error has occurred")
 	}
 	return resp, nil
 }
--- a/dependencies.go
+++ b/dependencies.go
@@ -160,7 +160,7 @@ func newPrivateClient(c http.Client) http.Client {
 func (c *privateClient) Do(req *nethttp.Request) (*nethttp.Response, error) {
 	resp, err := c.client.Do(req)
 	if err != nil {
-		return nil, errors.New(codes.Internal, "an internal error has occurred")
+		return nil, errors.Wrap(codes.Internal, "an internal error has occurred", err)
 	}
 	return resp, nil
 }
--- a/dependencies/influxdb/http_provider_test.go
+++ b/dependencies/influxdb/http_provider_test.go
@@ -38,7 +38,7 @@ func TestPrivateClient(t *testing.T) {
 	}
 	_, err = c.Client.Do(&nethttp.Request{})
 	if err != nil {
-		if err.Error() != "an internal error has occurred" {
+		if err.Error() != "an internal error has occurred: mock error" {
 			t.Fatalf("got unexpected error: %s", err)
 		}
 	} else {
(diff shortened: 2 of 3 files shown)


dependencies/http/http.go
```diff
@@ -142,3 +142,28 @@ func WithTLSConfig(c Client, config *tls.Config) (Client, error) {
 	}
 	return &newClient, nil
 }
+
+// PrivateClient is an http client that obscures error messages that may contain
+// sensitive information
+type PrivateClient struct {
```
wolffcm, line 148: Could you make this type unexported (lowercase letter) since you have a public constructor that returns an interface below?

Is it possible to write a unit test that shows that it always returns the opaque error?

wolffcm (changes requested):
Seems like this new type can be private, and perhaps it can be tested.

scbrickley:
@wolffcm Could use some feedback here. There are definitely changes that still need to happen, including adding some test coverage. I haven't spent much time in this area of code though, so I still have questions about whether or not this is an appropriate fix, and if so, what specifically needs to be tested.

scbrickley:
@wolffcm I believe I found all the places that create an http client. If specifying the host was optional, I added a branch that used the new private client if the host argument is missing.

I had some difficulty testing the changes to `HttpProvider` because the things I needed to test were private. The test I added fails with nil pointer dereference panic. Trying to fix that now, but I'm not even sure if this test is something we want to keep.

scbrickley:
Test should pass now. I think this is ready for another round of review?

dependencies/http/http.go
scbrickley: Had to make this public again because there's a test in a separate package that references this type.

dependencies/influxdb/http_provider_test.go
```diff
@@ -0,0 +1,32 @@
...
+	ctx, _ := dependency.Inject(context.Background(), deps)
+	c, err := h.clientFor(ctx, Config{})
+	if err != nil {
+		t.Error(err)
+	}
+	client := c.Client
+	if _, ok := client.(*http.PrivateClient); !ok {
+		t.Error("reader does not use private client")
+	}
+}
```
wolffcm, line 32: The `HTTPClient` field in `Deps` is an interface, so you can pass anything you want in there.

What if you passed a mock client into the `Deps` that looks something like this
```go
import nethttp "net/http"
import "github.com/influxdata/flux/dependencies/http"

type struct mockClient {}

var _ http.Client = mockClient{}

func (c mockClient) Do(req *nethttp.Request) (*nethttp.Response, error) {
  return <some error>
}
```

Then create an instance of `Deps` with the mock client. You should be able to show that when the client returns an error you get the `an internal error...` message if there was no URL supplied.

This way both `PrivateClient` and `NewPrivateClient` don't need to be exported, but you can still test the new logic you added.

wolffcm (changes requested):
I had some suggestions on how to test this without adding to our public Go API.

dependencies/influxdb/http_provider_test.go
scbrickley: I was able to make `PrivateClient` private again using your suggestions, but I don't see how to make `NewPrivateClient` private, since it needs to be imported to a separate package in order for `Deps.PrivateHTTPClient()` to work.

https://github.com/influxdata/flux/blob/648369db57a595e9bcaab076c9f5815c373b4c37/dependencies.go#L55-L62

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -151,6 +151,9 @@ func api(ctx context.Context, a values.Object) (values.Value, error) {
 		req.Header.Set("Authorization", "Token "+token)
 	}
 
+	if host == "" {
+		client = fluxhttp.NewPrivateClient(client)
```
wolffcm: Instead of this can't you call `deps.PrivateHTTPClient` here?

stdlib/experimental/influxdb/influxdb.go
```diff
@@ -43,27 +43,39 @@ func api(ctx context.Context, a values.Object) (values.Value, error) {
...
 	{
 		if validator, err = deps.URLValidator(); err != nil {
 			return nil, fluxerrors.New(codes.Internal, "missing dependencies")
 		}
-		if client, err = deps.HTTPClient(); err != nil {
+
+		if host, ok, err = args.GetString("host"); err != nil {
+			return nil, err
+		} else if !ok {
+			return nil, fluxerrors.New(codes.Invalid, `keyword argument "host" is required when executing outside InfluxDB`)
```
wolffcm, line 58: I am pretty sure that we will hit this path if `host` is an empty string, so I think the first branch below is dead code. I think this is okay though, since it's defensive.

(There is a separate implementation of this that is used for cloud)

stdlib/influxdata/influxdb/source.go
```diff
@@ -75,7 +76,12 @@ func (s *source) run(ctx context.Context) error {
 		return err
 	}
 
-	client, err := s.deps.HTTPClient()
+	var client fluxhttp.Client
+	if *s.spec.GetHost() == "" {
+		client, err = s.deps.PrivateHTTPClient()
+	} else {
+		client, err = s.deps.HTTPClient()
+	}
```
wolffcm, line 84: I don't see where a default `host` would be provided here, so I think that this is dead code; `host` will always be defined. I believe that `validateHost` would produce an error otherwise.

This implementation of `from` is for when we it's invoked with an explicit host, so by default, the user is supposed to have provided one. There is planner rule in cloud that will convert a `from` that is destined for a local cluster.

wolffcm (approved):
This looks good to me. I saw some minor issues where there are branches where I think `host` would always be defined, I'll leave it to your discretion.

stdlib/experimental/influxdb/influxdb.go
scbrickley: For my own clarity, do you know where the other implementation is? I ran into the cloud vs. oss difference while I was debugging this, and I couldn't find the version where specifying the host is optional. Maybe I can move this fix there.

stdlib/experimental/influxdb/influxdb.go
scbrickley: The most recent commit does that. I must have pushed it up after you added this comment.

stdlib/influxdata/influxdb/source.go
scbrickley: If it's dead code, I'll just remove it. Maybe the converted `from()` has a section of code where this branch would be more appropriate?

wolffcm (changes requested):
This looks good, I just had one last-minute change request.

dependencies.go
```diff
@@ -136,3 +146,21 @@ func GetDialer(ctx context.Context) (*net.Dialer, error) {
...
+}
+
+func newPrivateClient(c http.Client) http.Client {
+	return &privateClient{client: c}
+}
+
+func (c *privateClient) Do(req *nethttp.Request) (*nethttp.Response, error) {
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.New(codes.Internal, "an internal error has occurred")
```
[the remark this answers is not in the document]
scbrickley: Can do!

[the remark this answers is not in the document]
wolffcm: So something like
```go
return nil, errors.Wrap(codes.Internal, "an internal error has occurred", err)
```
would be fine.

wolffcm (approved):

PR #5267, opened 2022-10-06
feat(dynamic): add `isType` builtin

### LINKED ISSUE #5145: `Dynamic` and `typing.isType`
I still have questions about this one.

If we want to let folks check the inner type of a `Dynamic` we could say `typing.isType()` should be able to return `true` when there's a match.

If we build individual functions for `asInt`, `asFloat`, etc then they could be implemented in terms in pure flux like with a common pattern of: 

```
// assuming `castFns` or similar lets us lookup the right cast function
(type, v, default) => if typing.isType(v: v, type: type) then castFns[type](v) else default
```

It seems possible to build, but still the question remains: _should `isType` report `true` based on the inner type?_

My intuition currently is we should not report `true` for the inner type and `Dynamic` should be opaque until a conversion is attempted.

Instead callers should use `dynamic.as*` or `dynamic.typecase` etc to convert.

Create a new `isType` function in the dynamic package:
```
builtin isType : (v: dynamic, type: string) => bool
```

DOD:

- New `dynamic.isType` function exists, which operates on the inner type.
- Create a new issue to explore how dynamic and `typing.isType` should behave.

onelson:
Fixes #5145

### Checklist


skartikey (approved):
Looks good!

stdlib/experimental/dynamic/dynamic.flux
```diff
@@ -52,3 +52,30 @@ builtin jsonParse : (data: bytes) => dynamic
 // tags: type-conversions
 // introduced: NEXT
 builtin jsonEncode : (v: dynamic) => bytes
+
+// isType tests if a `dynamic` holds a value is a specified type.
```
sanderson, line 56:
```suggestion
// isType tests if a dynamic type holds a value of a specified type.
```

sanderson (approved):
Looks good. Just one minor suggestion.

Issue #5244, opened 2022-09-30
Remove the need for `NeverEq` in the salsa queries


Marwes:
The salsa POC uses the wrapper `NeverEq` type allow values which do not implement `Eq`  to be returned by queries. Since the wrapper is never equal to any other value, the behavior that we get is that any time something in a dependent query changes, a query using `NeverEq` also need to be recomputed. This does lose the finer grained dependency tracking that salsa is capable of but will be fine for regular end-to-end compilation (where we don't expect that anything changes within a compilation).

Salsa used to have the `volatile` attribute which had this same behavior but it was removed as it was thought to be unnecessary https://github.com/salsa-rs/salsa/issues/218 . Adding it back should be fairly straightforward by adding a new `MemoizationPolicy` and select that through a `volatile` attribute (just like `NeverMemoizeValue` is used)

https://github.com/salsa-rs/salsa/blob/30b5e9760aadc3570dc2ba176f4d74448c4152ed/src/derived.rs#L63-L99

```rust
pub enum VolatileMemoizeValue {}
impl<Q> MemoizationPolicy<Q> for VolatileMemoizeValue
where
    Q: QueryFunction,
{
    fn should_memoize_value(_key: &Q::Key) -> bool {
       // Do memoize the value
        true
    }

    fn memoized_value_eq(_old_value: &Q::Value, _new_value: &Q::Value) -> bool {
       // We had to recompute the value as its dependencies change, just replace it
        false
    }
}
```

### Alternative

We could instead implement `Eq` for all values we cache in salsa which would let us just use queries without any extra attributes. The only drawbacks is

* We need to deal with floats not implementing `Eq` (use https://docs.rs/ordered-float/latest/ordered_float/struct.NotNan.html ?)
* It isn't really necessary for us to track changes that closely (yet) and checking equality between an entire AST or semantic graph is relatively expensive

DOD
- We do not need to use the `NeverEq` workaround to represent volatile queries

Marwes:
Salsa PR https://github.com/salsa-rs/salsa/pull/421

PR #5146, opened 2022-08-30
chore: update rust toolchain to 1.63

onelson:
Updates the rust toolchain to 1.63 which is the current latest.

The hold at v1.58 is no longer required since the cross-build issue has been resolved.
Context around the initial reason for the hold is mentioned here https://github.com/influxdata/flux/pull/5084, citing https://github.com/rust-lang/rust/issues/97117 

InfluxDB updated to 1.63 already with this PR: https://github.com/influxdata/influxdb/pull/23660

This also means the extra toolchain introduced for the new rust repl work is no longer needed.

With the toolchain bump came a new lint which required updates to the codebase.
Looks like it was all a single lint: [derive_partial_eq_without_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq)

There was a [redundant_clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone) showing up in `convert.rs` but it seems to be a false positive.

Existing tests pass, so I think we're good to go.

### Checklist

Dear Author :wave:, the following checks should be completed (or explicitly dismissed) before merging.

- [x] ✏️ Write a PR description, regardless of triviality, to include the _value_ of this PR
- [x] 🔗 Reference related issues
- [x] 🏃 Test cases are included to exercise the new code
- [x] 🧪 If **new packages** are being introduced to stdlib, link to Working Group discussion notes and ensure it lands under `experimental/` **N/A**
- [x] 📖 If **language features** are changing, ensure `docs/Spec.md` has been updated **N/A**

Dear Reviewer(s) :wave:, you are responsible (among others) for ensuring the completeness and quality of the above before approval.

PR #5268, opened 2022-10-06
chore: Implement Eq for values we need to store in the salsa database

Marwes:
See https://github.com/influxdata/flux/pull/5135

Another way of solving this would be with volatile salsa queries (see https://github.com/influxdata/flux/issues/5244). But due to the changes in salsa 2022 it is unclear to me right now if that path is possible. So instead we implement `Eq` on everything. The only real change here is that we use `NotNan` for the `f64` in `FloatLiteral`. Since we do not have a way to write a NaN literal that should not change things however.

### Checklist


onelson (approved):
Looks fair to me. Might conflict a bit with #5146 but there's already a lot to fixup there since so much time has passed.

PR #5272, opened 2022-10-11
fix(unpivot): _field should not be added as a group key by default

skartikey:
if `_field` is in otherColumns parameter, `_field` should not be made group key by default
The issue was discovered during the [iox group push down](https://github.com/influxdata/idpe/pull/15867) work

### Checklist


stdlib/experimental/experimental_test.flux
```diff
@@ -388,6 +388,69 @@ testcase unpivot_other_columns {
...
+                    _time: 2018-12-01T00:00:00Z,
+                    _field: "load1",
+                },
+                {
+                    _measurement: "m",
+                    tag: "t1",
+                    f0: 20.1,
+                    f1: 20.2,
+                    _time: 2018-12-01T00:00:10Z,
+                    _field: "load1",
```
wolffcm, line 409: For this to be a representative test case, there shouldn't be a `_field` column in the input to `unpivot`, since pivoted data does generally not have a `_field` column.

wolffcm: Thinking it through, if there _is_ a `_field` column in the input an error seems to be appropriate? Since there is no way to produce output that does not overwrite the data that already exists in `_field`. Thoughts?

stdlib/experimental/unpivot.go
```diff
@@ -146,22 +147,34 @@ func (t *unpivotTransformation) Process(chunk table.Chunk, d *execute.TransportD
...
 		groupKey := chunk.Key()
-		nCols := len(groupKey.Cols()) + len(otherCols) + 2
+		nCols := len(groupKey.Cols()) + len(otherCols) + defaultValueColLen + defaultFieldColLen
 
 		columns := make([]flux.ColMeta, 0, nCols)
 		columns = append(columns, groupKey.Cols()...)
-		columns = append(columns, flux.ColMeta{Label: "_field", Type: flux.TString})
+		if defaultFieldColLen != 0 {
+			columns = append(columns, flux.ColMeta{Label: influxdb.DefaultFieldColLabel, Type: flux.TString})
+		}
```
wolffcm, line 163: This logic doesn't seem quite right to me. If there is no `_field` column in the input, then the `_field` column will never get created. We want to create it regardless, we just want to omit the column from `groupCols` and `groupValues` below if `_field` is in the `otherCols` parameter.

wolffcm (changes requested):
I think this is pretty close, but the logic is not quite right.

stdlib/experimental/experimental_test.flux
skartikey: I think it makes sense to show an error if the input to unpivot contains `_field` in it. wondering if we have to do the same for `_value` column?

The reason I have to put up this PR is that the [pushdown group test](https://github.com/influxdata/idpe/pull/15867/files#diff-0eafdd805b72b81b52f134721a2a1718e3230d507a8a7705c4e3541d1aab7d95R644) was failing because it contains `_field` in the input.

The fix would be to throw a user error when there is `_field` or `_value` column in the input to unpivot.  

At the moment, if a column is present in the otherColumns and missing in the input column, the code throws an error. 

```
import "array"
import "experimental"
import "testing"

 array.from(
    rows: [
        {
            _measurement: "m",
            tag: "t1",
            f0: 10.1,
            f1: 10.2,
            _time: 2018-12-01T00:00:00Z,
        },
        {
            _measurement: "m",
            tag: "t1",
            f0: 20.1,
            f1: 20.2,
            _time: 2018-12-01T00:00:10Z,
        },
    ],
)
    |> group(columns: ["_measurement"])
    |> experimental.unpivot(otherColumns: ["_time", "tag", "_field"])
```
```
Result: _result
Error: runtime error @24:8-24:70: unpivot: unpivot could not find column named "_field"
```

wolffcm (approved):
Looks good, thanks.

Issue #5069, opened 2022-08-10
implicit cast of uint/int in equality ops riddle


onelson:
In https://github.com/influxdata/flux/pull/5042 a bug in the `>=` and `>` operators for `int`/`uint` and `uint`/`int` arguments prompted a "reality check" discussion.

The the riddle before us is: should `-1 > uint(v: 1)` be true or false?

In a situation where we wrap the (signed) int `-1` in a call to `uint()`, we can see the value "wraps around":

```
> uint(v: -1)
18446744073709551615
```

With this, a `uint` cast of `-1` actually could be considered larger than `uint` `1`.

The question: should `uint` on either side force the cast of _the other side_ to `uint`, or should `int` on either side force the cast to `int`?

Consider the precedent set by `float`. When either side is a `float`, the _other side_ is cast to `float` before comparing. To me this means we should be selecting the weaker/more permissive type in these situations. A "race to the bottom."

The implementation we have today avoids casting at all by reasoning that a negative `int` value _must be smaller than even the smallest `uint`_, and conversely that a negative `int` _cannot be larger than any valid `uint`_.

---

In summary this might be a #wontfix. We might not want to change a thing, but we had a brief discussion on slack about it so it seemed worth recording here for posterity.

onelson:
While this issue was mostly concerned with implicit casting during equality ops _in general_, there's a related concern about `uint` (where these casts are the most questionable) behaving differently between x86 and arm architectures. See: #5274

PR #5274, opened 2022-10-11
fix(stdlib): array touint test utilized conversion behavior incorrectly

jsternberg:
The `array.toUInt()` test utilized conversion behavior and assumed that
conversion behavior was the same for floats to uints as ints to uints.

Floats to uints is implemented differently than ints to uints especially
on the ARM platform. On ARM, a negative float converts to zero. While we
may want to change Flux to define a specific behavior for this
conversion in the long run, this is just going to fix the test so it
consistently uses floats for now so the test passes appropriately on all
architectures.

### Checklist


onelson (approved):
Yeah, this looks like a convenient "fix" to sidestep the issue. Linking to #5069 since it relates to `uint` behavior.

PR #5278, opened 2022-10-12
feat(libflux): add attribute parsing to the parser

### LINKED ISSUE #5235: Implement Attribute Syntax
See https://github.com/influxdata/flux/pull/5234 for the definition.

TL;DR

```
@attribute(1,2,3)
@otherattribute({x: 5})
```

Maybe add a generic Attribute section to the base node so that any node in the future can have attributes?
The initial specification is that attributes are only allowed as part of a package or testcase.

DOD:
- Attributes can be parsed
- Parsed Attributes are added to the AST
- Test cases added to parser

jsternberg:
This adds attributes to both the file and the statements. The parser
understands attributes as arbitrary and allow them to be attached to
nodes that don't necessarily have any supported attributes.

It's up to the semantic pass to deal with interpreting the attributes.
This is just about the parser.

Fixes #5235.

### Checklist


libflux/flux-core/src/parser/tests/attributes.rs
```diff
@@ -0,0 +1,395 @@
...
+                                Attribute {
+                                    base: BaseNode {
+                                        location: SourceLocation {
+                                            start: "line: 3, column: 1",
+                                            end: "line: 3, column: 12",
+                                            source: "@deprecated",
+                                        },
+                                        comments: [
+                                            Comment {
+                                                text: "// My documentation comment.\n",
```
jsternberg, line 315: This result was concerning to me. The code that it represents seems normal, a comment, an attribute, then a variable assignment. But the comment is attached to the first token which gets attached to the attribute which is a base node itself. This seems correct, but also seems like a breaking change for anything consuming doc comments.

jsternberg:
I also need to update the formatter for this.

libflux/flux-core/src/parser/tests/attributes.rs
jsternberg: If we're fine with this change it seems the code to extract the package doc comment is also in rust so we can probably just change that to search for the appropriate comment.

libflux/flux-core/src/parser/tests/attributes.rs
```diff
@@ -0,0 +1,405 @@
+use super::*;
+
+#[test]
+fn parse_package_attribute() {
+    let mut p = Parser::new(
+        r#"@attribute
```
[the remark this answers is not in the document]
jsternberg: I don't think we decided, but I figured that package could be a special exception to this. There's really no other "mandatory optional" components of the grammar and the attributes that you would apply to a package/file aren't going to be the same as the ones you would apply to a variable assignment so I figured we could just look in both places for the package level ones.

libflux/flux-core/src/parser/tests.rs
```diff
@@ -73,93 +74,93 @@ fn parse_invalid_unicode_bare() {
     test_file(
         r#"®some string®"#,
         expect![[r#"
-        File {
-            base: BaseNode {
-                location: SourceLocation {
-                    start: "line: 1, column: 1",
-                    end: "line: 1, column: 16",
-                    source: "®some string®",
+            File {
```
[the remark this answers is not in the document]
jsternberg: I couldn't figure out why. It was very strange.

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -1819,6 +1824,56 @@ impl<'doc> Formatter<'doc> {
...
+        doc
+    }
+
+    fn format_attribute(&mut self, n: &'doc ast::Attribute) -> Doc<'doc> {
+        let arena = self.arena;
+        let mut doc = docs![
+            arena,
+            self.format_comments(&n.base.comments),
+            arena.text(format!("@{}", n.name)),
+        ];
```
Marwes, line 1844:
```suggestion
        let mut doc = docs![
            arena,
            self.format_comments(&n.base.comments),
            "@",
            &n.name,
        ];
```

jsternberg: My understanding is this change will allow `@` and the name to be split from each other. The scanner grammar doesn't allow `@ attribute` so these two are a single atom. If the suggested change ensures the two will be together, then this change is fine. Otherwise, I think the previous one needs to remain.

Marwes: No, you must add softline/hardline/etc to get explicit whitespace the formatter will always lay them out as `@attribute` with the change.

jsternberg: Ok updated this then.

libflux/flux-core/src/parser/mod.rs
```diff
@@ -370,14 +374,22 @@ impl<'input> Parser<'input> {
...
         }
-        let body = self.parse_statement_list();
+        let mut body = self.parse_statement_list();
         if let Some(stmt) = body.last() {
             end = stmt.base().location.end;
         }
+
+        // If no package or imports were specified, then attach
+        // the attributes to the first statement if it exists.
+        if let (None, 0, Some(stmt)) = (&pkg, imports.len(), body.first_mut()) {
```
nathanielc, line 384: Why do we need attach them to the first statement? Since we directly attach them to the `File` struct can't we leave them there? In other words what happens if you just delete this condition? I would expect that the attributes then always exist on the `File` struct directly but maybe I am missing something?

nathanielc: Ok reading through all the test cases I see the ambiguity. I had it backwards in my head, its not that we can't put them on the File struct rather we don't want to put them on the File struct since it would be confusing that its not on the statement. 

For example:

```
@foo
x = "foo"
```

We want `@foo` to apply to the statement not to the whole package.

What if we required a package clause in those cases?

```
package bar

@foo 
x = "foo"
```

In other words the first attributes in a file always apply to the whole file scope and if you don't want that add a package clause.

Alternatively we could do the `inner/outer` differing syntax.

jsternberg: There's no reason we can't do that, but I suspect that would be more confusing since you could write what you had there:

```
@foo
x = "foo"
```

If I read this, I'm not going to be thinking about how the parser interprets things. I'm going to just say "well I put an annotation on a statement, why isn't it working?"

At the same time, it occurred to me that attributes that would be attached to a statement aren't going to correspond to ones attached to the file itself so we can just check both locations and avoid an extra syntax that would be fairly confusing in my opinion.

Another alternative is just say that `package` is required if you use the attribute in the file. In all honesty, we'll have ways to specify the edition outside of the script (like the http param or the default for the organization) so I think it might be overkill to add a new syntax specifically for the package attributes. I think rust needs the syntax more because it's possible to put a crate macro anywhere in the file and it's not restricted to the beginning of the file. I think making package required if you want to use attributes would be my favorite and I think we can find a way to make it so it's not painful through the UX by just filling it in as the default.

nathanielc: Makes sense lets move forward with what you have now.

Marwes (approved):
I can approve this so we move forward but I still think it is a mistake to special case attributes such that they sometimes attach to the package.

Attaches to the package
```
@feature 
x = 1
```

Attaches to the statement
```
package x
 
@feature
x = 1
```

jsternberg:
@Marwes just as an FYI, the example you gave:

```
@feature 
x = 1
```

This does not attach to the package. It attaches to the statement. When we implement the logic, we'll have to search for `@edition` on both the package and the first statement, but the parser itself isn't creating any special case.

Issue #5276, opened 2022-10-11
Last mile dynamic work


onelson:
A test was added while working on the final bit of feature dev towards dynamic in #5271 which surfaced a gap in the inference for dynamic.

DOD:
- [x] fix up the type inference so that `dynamic_cast_from_json_deep` passes 
- [x] fix up dynamic member expressions during evaluation
- [x] refine member expression runtime checks: error when not object
- ...try to identify any other gaps and provide fixes so we can close out the epic.

onelson:
For a use case where a dynamic is created within a `universe.map()`, not an `array.map()`, but converted to some scalar before the `map()` returns, we're seeing a runtime error:

```
runtime error ... map: failed to evaluate map function: cannot access property of a value with type dynamic; expected record
```

Looks like the inference is okay in this case, but maybe we're missing more runtime stuff - do we need updates to the evaluators to support this case?

I need to try and create a small reproducer for this before digging in. If we can't support this use case, folks might instead be able to leverage `findColumn` to pull the raw data back into the interpreter space.

onelson:
Yeah, this makes sense. #5246 was effectively incomplete. At the time I thought I was off the hook for this since dynamic values can't be in tables, but being in tables isn't the only way values appear in this space, clearly :sweat_smile: 

Should be a small change to get this working.

onelson:
https://github.com/influxdata/flux/pull/5286 covers the 2 concrete bullet points.
I have not yet found anything new to add to sneak into the scope of this issue.

onelson:
https://github.com/influxdata/flux/pull/5290 covers the tuned-up runtime checks for member expressions. 

It'll mean folks will need to navigate objects more carefully using `exists` (which is fine, though a bit noisy) but also may require them to use `debug.null()` for fallback cases.

Potentially it'll force folks to face some of the nuance (which until recently I found daunting) of nulls in flux :thinking: :cry:.

PR #5286, opened 2022-10-13
fix(dynamic): allow functions receiving dynamic args to do member expressions

onelson:
Moves towards fixing #5276 

When a function accepts an unbounded argument, we fail to unify the argument with the record-like usage in the function body.

Example:
```
n = (d) => {
  return int(v: 
    d.number // compile-time error
  )
}(d: dynamic.dynamic(v: { number: 123 }))
```

Today, the REPL shows this output for the above:

```
error: expected {A with number: B} (record) but found dynamic (dynamic) (argument d)
  ┌─ main:5:6
  │
5 │ }(d: dynamic.dynamic(v: { number: 123 }))
  │      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

The new match arm added to `MonoType#unify` in this diff lets this compile as expected.

### Checklist


onelson:
I just remembered there was a test to unskip as a part of this PR, and the test is still failing :cold_sweat: 

Pulling this back into draft until I get it passing.

onelson:
@Marwes tapping you for a look in case there's something terrible being proposed here. I have mostly been making guesses and I think you're the resident expert in this area :sweat:

compiler/compiler.go
```diff
@@ -84,6 +84,9 @@ func substituteTypes(subst *semantic.Substitution, inferredType, actualType sema
 	// which isn't represented in type inference.
 	if actualType.Nature() == semantic.Invalid {
 		return nil
+	} else if actualType.Nature() == semantic.Dynamic {
+		// Dynamic values are as reduced as they are going to get.
+		return nil
```
onelson, line 89: Seems like ideally this substitution would happen in rust and flow back through here.

I'm not yet sure how to make that happen.

libflux/flux-core/src/semantic/types.rs
```diff
@@ -904,7 +904,11 @@ impl MonoType {
             (MonoType::Error, _) | (_, MonoType::Error) => (),
             // Similarly if both sides are dynamic, there's nothing more to do.
             (MonoType::Dynamic(_), MonoType::Dynamic(_)) => (),
-
+            // In situations where a function receives a dynamic value as an
+            // unbounded parameter, this allows the value to be used like a
+            // record (ie, granting member access).
+            (MonoType::Dynamic(_), MonoType::Record(_))
+            | (MonoType::Record(_), MonoType::Dynamic(_)) => (),
```
Marwes, line 911: I don't think this makes sense under the assumption we have discussed with dynamic. This effectively allow `dynamic` to be passed anywhere we expect a record (and vice versa). If we go with a more permissive dynamic where dynamic may be accepted everywhere (and therefore "bleeds out more easily" in the code) I think  this change makes sense (though it would be more general and just do this for all types, effectively.

onelson: Also, I suppose this is a testcase I should add on the rust side; something to show that dynamic is rejected from a builtin that has a Record in the constraint.

I'll try to add that with the continued work here.

onelson: Sure enough, a test like the following fails by passing typechecks. I'm still tinkering to try and get it green without spoiling the rest.

```rust
#[test]
fn dynamic_rejected_where_record_expected() {
    test_error_msg! {
        env: map![
            "d" => "dynamic",
            "fn" => "(v: A) => bool where A: Record",
        ],
        src: r#"
        x = fn(v: d)
        "#,
        expect: expect![[r#"tbd"#]]
    }
}
```

libflux/flux-core/src/semantic/types.rs
```diff
@@ -1256,10 +1256,14 @@ impl Substitutable for Dynamic {
...
-            exp: with,
-        })
+        match with {
+            // Dynamic needs to be able to go into the various cast functions
+            Kind::Basic => Ok(()),
+            _ => Err(Error::CannotConstrain {
+                act: MonoType::Dynamic(self.clone().into()),
+                exp: with,
+            }),
+        }
```
onelson, line 1266: This prevents `dynamic` being passed into functions that specify `Record` (or anything else). `Basic` was needed to continue to allow the standard cast fns to work, although I didn't actually find where they specify this. Still, the tests failed complaining the "arg must be `Basic`".

iirc when we were discussing the design, we raised the possibility of needing a new Kind for `Basic + Dynamic`. Not sure it's necessary to add a new Kind at this stage, but I just now remembered it was mentioned.

stdlib/experimental/dynamic/dynamic_test.flux
```diff
@@ -6,13 +6,6 @@ import "testing"
 import "experimental/dynamic"
 import "internal/debug"
 
-testcase dynamic_not_comparable {
-    testing.shouldError(
-        fn: () => dynamic.dynamic(v: 123) == dynamic.dynamic(v: 123),
-        want: /unsupported/,
-    )
-}
```
onelson, line 14: With the new `constrain()` implementation, this no longer compiles with `dynamic (dynamic) is not Equatable`. Seems better than catching it at runtime.

libflux/flux-core/src/semantic/types.rs
onelson: _I think_ 2efb7affabf93e885b7e10da62d0a983a171cd00 should cover the specific concern of Record/dynamic being interchangeable. For places where the `Record` constraint is used, `dynamic` should now be rejected.

Again, this is mostly guesswork. I only know what the test suite is telling me. If there are more edges to this, I'll probably need more tests to capture the scenario.

Marwes: 2efb7af seems fine (though it would be better to have `ToInt`, `ToString`, etc kinds rather than an all, encompassing `Basic` kind).

The `Record <=> Dynamic` is  still wrong though as I understand from what we intended with dynamic

(That is, this test should not compile under the current direction we have for dynamic)
```
fn = (x) => ({ f: x.f0 })
v = fn(x: d)
```

onelson: Yeah, okay. Unfortunately, I think that test will have to work as-is if we want to be able to send `dynamic` into function args like those in `array.map` and `array.filter`. Maybe I've missed something here.

Part of the my hope for this version of dynamic was to be _more strict at the leaves_ while still _permissive at the branches_, so having a close affinity with `Record` makes sense to me.

The only way I can think to handle things differently would be to add a new kind which is `RecordLike` which both `Record` and perhaps a new `DynamicRecord` could constrain to. We could then offer a `dynamic.asObject` converter, like we have for array. 

Either that or we deny member access entirely and emulate the [API we have for dictionary](https://docs.influxdata.com/flux/v0.x/stdlib/dict/get/) which I find to be very clunky.

skartikey (approved):

Issue #5265, opened 2022-10-05
Union a single data set to each group key in a data set - internal error: panic: arrow/array: index out of range


jonathandavis805:
Flux Query as suggested

```
mapped_state_changes = state_changes
  |> set(key: "on", value: "x")

mapped_last_value_per_id = last_value_per_id
  |> set(key: "on", value: "x")

join.inner(left: mapped_last_value_per_id,  right: mapped_state_changes,  on: (l, r) => l.on == r.on,  as: (l, r) => { return {r with id: l.id} } )
```

Flux returns the error `internal error: panic: arrow/array: index out of range`

Data:

Data to union - state_changes
_start | _stop | _time | bool_value
-- | -- | -- | --
2022-09-14 12:24:30 | 2022-09-28 12:24:30 | 2022-09-28 12:24:30 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-09-28 17:00:00 | FALSE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-09-29 7:00:00 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-09-29 17:00:00 | FALSE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-09-30 8:00:00 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-09-30 16:00:00 | FALSE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-03 7:00:00 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-03 17:00:00 | FALSE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-04 7:00:00 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-04 17:00:00 | FALSE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-05 7:00:00 | TRUE
2022-09-28 12:24:30 | 2022-10-05 12:24:30 | 2022-10-05 12:24:30 | TRUE
2022-09-14 12:24:30 | 2022-09-28 12:24:30 | 2022-10-05 12:24:30 | TRUE

Grouped Data - last_value_per_id
_start | _stop | _time | _value | id
-- | -- | -- | -- | --
[...]

PR #5303, opened 2022-10-20
fix: join with large input crossing buffer size

skartikey:
**Issue -**
when the input table size is more than the max buffer size in the join function, the code was throwing the error `internal error: panic: arrow/array: index out of range
`

**Fix -**
Added the logic in split chunk to break the chunk with correct start and stop indexes

fixes: https://github.com/influxdata/flux/issues/5265

### Checklist


stdlib/join/join_fn.go
```diff
@@ -326,14 +326,25 @@ func colsFromObjectType(t semantic.MonoType) []flux.ColMeta {
...
+		if curSize > bufferSize {
+			start = i * bufferSize
+			stop = bufferSize - 1
+			curSize -= bufferSize
+		} else {
+			start = i * bufferSize
+			stop = start + curSize - 1
+			curSize = 0
+		}
+		slice := getChunkSlice(c, start, stop)
```
wolffcm, line 347: Is stop inclusive or exclusive here? Based on Go indexing semantics I might expect it to be exclusive, in which case I would not expect you to need to subtract one from `stop` above.

stdlib/join/join_test.flux
```diff
@@ -442,3 +443,153 @@ testcase join_empty_table {
...
+    got =
+        join.inner(
+            left: mappedLeft,
+            right: mappedRight,
+            on: (l, r) => l.on == r.on,
+            as: (l, r) => {
+                return {r with id: l.id}
+            },
+        )
+            |> limit(n: 13)
```
wolffcm, line 592: Since you limit the number of rows here, it's hard to tell if there is an off-by-one error in the fix. Maybe verify the row count as well?

wolffcm (changes requested):
I had some questions about a possible off-by-error.

stdlib/join/join_fn.go
skartikey: Thanks for pointing that out, yes stop param is exclusive to `getChunkSlice` I have corrected the index.

stdlib/join/join_test.flux
skartikey: I did limit it so that I do not have to add a lot of rows in the `want`. 
I wrote it in a different way now, as each `id` from the left table will join with 13 rows of the right table producing 13 rows per id for the join result, I have added `tail(n:13)` to verify their is no off-by-one error.

scbrickley:
Did this only happen under specific conditions? I remember testing join with input datasets that were quite large and I never ran into this problem.

skartikey:
@scbrickley I followed the example given in the issue to reproduce this bug. In this particular scenario, the table size is more than the [BufferSize](https://github.com/influxdata/flux/blob/master/execute/table/buffered.go#L23).

wolffcm:
@scbrickley Maybe the issue in this case is that the output size is larger than the buffer size? The example I looked at it was effectively doing a cross join, where the predicate always matches.

wolffcm (approved):

MarcoPignati:
Even though this is marked as fixed in Oct 2022, it is still an issue. See  #5402 and on the [community](https://community.influxdata.com/t/limit-on-join-function/27616)

Issue #5289, opened 2022-10-17
`testing.shouldError` should fail the test when no error is produced by the fn


onelson:
Currently it looks like a test will pass if the supplied function doesn't error at all. The `want` error message is never checked against anything in this situation.

PR #5319, opened 2022-10-28
fix: shouldError should error when no error occurs

Marwes:
When `catch` did not get an error the `msg` field would be null, which would make the regexp match operator to just propagate `null` instead of trying to evaluate. Since `null` gets translated to false in the filter we would get an empty stream, causing us to not emit an error.

Fixes https://github.com/influxdata/flux/issues/5289

### Checklist


stdlib/testing/testing_test.flux
```diff
@@ -29,3 +31,9 @@ testcase succeed_on_non_empty_result {
 testcase test_should_error {
     testing.shouldError(fn: () => die(msg: "error message"), want: /error message$/)
 }
+
+testcase test_should_error_should_error_when_no_error {
```
kelseiv, line 35: `should_error` twice intentional?

kelseiv (approved):

stdlib/testing/testing_test.flux
Marwes: Yes :)

onelson (approved):
Looks great. Thanks for patching this :pray: 

This is much more simple than I thought it would be on my first look! (I thought we'd need to change how the stream/filter was written, but this is more direct :tada:)

Issue #5299, opened 2022-10-19
Collect metrics on when `window` and `aggregateWindow` are used with `_time` in the group key


wolffcm:
Having `_time` in the group key for these functions tends to produce output that is not all that useful, and moreover the semantics can be hard to implement, resulting in bugs like this one: #5260.

We should collect metrics in `Statistics.Metadata` to get an idea of how often this happens in the wild, with an eye to making these functions return an error in the future.

Issue #5248, opened 2022-09-30
Define behavior of yield/table find and prevent misuse


jsternberg:
At the current moment, some of the behavior between yield and table find can be undefined and result in some undesired behavior. This mostly happens when the two mix with each other. Consider the following:

```
src = gen.tables(...)
a = src |> getRecord(idx: 0)
src |> yield()
```

This is an oversimplification of the scenario, but the problem here is that `getRecord` forces `gen.tables` to be evaluated and then yield has it evaluated again. This results in `gen.tables` being created and executed twice and, in the case of random data like `gen.tables`, results in different outputs.

The defined mechanic for this though is that this should be a branch of the same stream. If we do this:

```
src = gen.tables(...)
src |> sum() |> yield(name: "sum")
src |> count() |> yield(name: "count")
```

This results in `gen.tables` being created once and passed to multiple areas. 

This can be represented as part of a graph of the plan nodes. The yield is one plan graph that exists for every flux script, but table find functions create additional plan graphs. When a plan node is potentially registered as part of multiple plan graphs, we get this ambiguity. We want to create some documentation that identifies what this internally looks like and also adds some protections to the interpreter to avoid a table object being associated with multiple plan graphs. We also want to add some protection to prevent hanging nodes or at least define the behavior. These are nodes that are included in a plan graph, but are not a root node anywhere. We likely want to error because it could be ambiguous what should be done with them but we could also just ignore them.

As an example:

```
src = gen.tables(...)
a = src |> sum() // stored in a variable, but never actually yielded so it is part of the plan graph but never used
src |> count()
```

PR #5320, opened 2022-10-28
feat: detect certain undesirable patterns so we can remove them later

jsternberg:
This detects two undesirable patterns so that we can remove them later.

The first is the `_time` parameter being in the group key when window is
used. This wasn't really intended and should have been an error, but it
may have accidentally been done and it's not explicitly an error. We
want to detect it happening so we can communicate those scripts need to
be fixed so we can turn it into an error without disruption.

The second is an undesirable interaction with table find. When table
find is used, the plan is immediately created and evaluated. But, when
this interacts with another plan evalutation, the operation will be
duplicated. This can create inconsistencies from how other parts of the
engine work. We've chosen to have this interaction be an error, but we
want to detect if it's happening so that we can make it an error and so
we can fix any scripts that may be using the interaction inadvertantly.

Related to #5248 and #5299.

### Checklist

### DIFF
--- a/internal/operation/spec.go
+++ b/internal/operation/spec.go
@@ -32,6 +32,12 @@ type Spec struct {
 	Resources  flux.ResourceManagement `json:"resources"`
 	Now        time.Time               `json:"now"`
 
+	// HasConflict is true if one of the operations in this spec
+	// was previously used in another spec. This indicates
+	// that two nodes from distinct specs relate to the same
+	// operation.
+	HasConflict bool
+
 	sorted   []*Node
 	children map[NodeID][]*Node
 	parents  map[NodeID][]*Node
--- a/internal/spec/build.go
+++ b/internal/spec/build.go
@@ -125,16 +125,21 @@ func isDuplicateTableObject(ctx context.Context, op *flux.TableObject, objs []*f
 }
 
 func buildSpecWithTrace(ctx context.Context, t *flux.TableObject, ider *ider, spec *operation.Spec, visited map[*flux.TableObject]bool, skipYields bool) {
-	s, _ := opentracing.StartSpanFromContext(ctx, "buildSpec")
+	s, cctx := opentracing.StartSpanFromContext(ctx, "buildSpec")
 	s.SetTag("opKind", t.Kind)
-	buildSpec(t, ider, spec, visited, skipYields)
+	buildSpec(cctx, t, ider, spec, visited, skipYields)
 	s.Finish()
 }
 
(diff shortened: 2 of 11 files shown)


internal/operation/spec.go
```diff
@@ -31,6 +31,7 @@ type Spec struct {
 	Edges      []Edge                  `json:"edges"`
 	Resources  flux.ResourceManagement `json:"resources"`
 	Now        time.Time               `json:"now"`
+	Conflicts  bool
```
wolffcm, line 34: Can you add a comment here that describes what `Conflicts` indicates? I know this is temporary code, but I would like to understand.

lang/compiler_test.go
```diff
@@ -1126,3 +1126,75 @@ var crlfPattern = regexp.MustCompile(`\r?\n`)
...
+		{
+			name: "disjoint_tablefind",
+			src: `import "internal/gen"
+a = gen.tables(n: 10)
+b = a
+	|> tableFind(fn: (key) => true)
+	|> getRecord(idx: 0)
+a
+	|> map(fn: (r) => ({_time: r._time, _value: r._value + b._value}))
+	|> yield(name: "_result")
```
wolffcm, line 1166: I'm trying to reconcile why this Flux code would trigger the event. I can see that the call to `gen.tables()` would be executed more than once for this plan because it is consumed by `tableFind` and also yielded.

It seems to me that because `buildSpec` checks that parents are not yet visited before it recurses to them, doesn't that mean it would fail to find the conflict? But this test still passes, so I must be missing something.

I would expect that:
-

wolffcm (commented):
I had some questions about how the changes to spec building find the table find event.

lang/compiler_test.go
jsternberg: `buildSpec` is invoked multiple times so the visited map only lives through each construction of a spec rather than the construction of all specs.

table find forces a spec to be built and another spec is built with yield. So `gen.tables()` is visited twice.

internal/spec/build.go
```diff
@@ -171,6 +176,7 @@ func buildSpec(t *flux.TableObject, ider *ider, spec *operation.Spec, visited ma
 	}
 
 	visited[t] = true
+	t.Owned = true
```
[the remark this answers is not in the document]
wolffcm: Ok, I get it now, can you add a comment to the `Owned` field that says this/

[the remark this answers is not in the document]
jsternberg: Done.

internal/operation/spec.go
jsternberg: Done.

wolffcm (approved):

PR #5321, opened 2022-10-28
feat: add outer attribute syntax 

nathanielc:
With this change it is now possible to attach attributes to an _outer_
context. The only place currently this is allowed is at the file scope,
where attributes are added to the File AST base node.

The syntax chosen in the PR current is `#attribute` is for outer attributes and `@attribute` is for inner attributes.

Some examples:

```
// Comment on file base node
#edition("2022.1")

x = 1
```

```

// Comment on package clause base node
@edition("2022.1")
package main

x = 1
```

// Comment on file base node
```
#edition("2022.1")

// Comment on package clause base node
@mount("fluxlang.dev", "....")
package main


// Comment on statement base node
@deprecated
x = 1
```

```

// Comment on file base node
#edition("2022.1")
package main

x = 1
```

```
// Comment on file base node
#edition("2022.1")


// Comment on package clause node
package main

x = 1
```

### Checklist


onelson:
During discussions earlier today, I mentioned _"I wish there was **one way** to specify the edition,"_ but I think I've come around on the whole inner/outer syntax split.

I think it'd be fine to have both inner and outer. In fact I might even prefer them to land together.

If the education around this is really just about attaching to the _following item_ vs the _containing item_ where containing item is going to be an actual file, or the implied file for a given testcase, then I think that's probably fine.

It's possibly still awkward to explain the implications of "file" vs "package", or how they relate.

nathanielc:
Inner only syntax support is added here https://github.com/influxdata/flux/pull/5325 this PR now only extends the syntax to add outer. We can add this later once we determine we need it.

Issue #5310, opened 2022-10-25
Download Flux modules within runtime as part of Salsa query


nathanielc:
Now that we can incrementally compile the stdlib (see https://github.com/influxdata/flux/pull/5135) next steps are to dynamically download and compile Flux modules.

This issue is about adding the ability for the Rust side to download Flux modules using the module API (see https://github.com/influxdata/flux/pull/5224).

Some implementation notes:

- We plan to reuse the token from the query request when making the Flux modules API request. This means passing the token from Go to Rust
- Downloading the module will be part of the `source()` salsa query (see https://github.com/influxdata/flux/blob/master/libflux/flux-core/src/db.rs#L57). 
- We do not need to worry about WASM/LSP as part of this issue. We may need an abstraction in the future to download modules but that can wait until we better understand how the LSP/WASM intend to download modules.

Version selection (i.e. finding the max of mins) is out of scope for this issue. The behavior will be to always download and compile the latest version of the module.

DOD:

- Flux modules can be downloaded from the registry and dynamically compiled
- Support for downloading latest version of a module.
- Flux scripts work with modules

PR #5324, opened 2022-11-01
feat: Download Flux modules within runtime as part of Salsa query

Marwes:
Adds a branch in the salsa database to try and lookup a given import through fluxmod. Since registries aren't implemented yet the code will just always try to query fluxmod first and fallback to the stdlib otherwise.

- [x] Flux modules can be downloaded from the registry and dynamically compiled
- [x] Support for downloading latest version of a module.
- [x] Flux scripts work with modules in tests
- [x] Cache the downloaded module
- [x] Limit depth of module dependencies
- [x] Test for module cycles
- [x] Remove debug url

The token is provided by the go side, but only through an environment variable for testing as I haven't figured out how to retrieve it (yet).

Testcases can be run with or without the `integration_test` rust feature flag however the test for modules with directories is not working until fluxmod handles those.

Part of #5310 

### Checklist


libflux/flux-core/src/db.rs
```diff
@@ -35,6 +36,151 @@ pub enum Error {
...
+    ) -> Result<()> {
+        let mut multipart = multipart::client::lazy::Multipart::new();
+        for (name, contents) in &files {
+            multipart.add_stream("module", contents.as_bytes(), Some(name), None);
+        }
+        let body = multipart
+            .prepare()
+            .map_err(|err| Error::Message(err.to_string()))?;
+
+        let response = ureq::post(&format!("{}/{}/@v/v{}.zip", Self::BASE, module, version))
```
nathanielc, line 72: Heads up, we expect the `.zip` extention to be removed from the API route since we are not posting zip files. We will use the `Content-Type` header to indicate how the data is being uploaded. Obviously this works now, but when we change the server we will need to update this line.

nathanielc (approved):
LGTM minus the hard coded bits to remocal. I like that the client is quite simple.

PR #5325, opened 2022-11-01
fix: update attribute syntax to require following element

nathanielc:
This change makes is so that attributes must only exist preceding another syntax that allows attributes. Attributes can now always be found as part of the base node of the attached element (i.e. package clause or import declaration or statement).

### Checklist


docs/SPEC.md
```diff
@@ -1081,7 +1078,8 @@ Every statement contained in an imported package is evaluated.
 
 ### Attributes
 
-Attributes define a set of properties on the subsequent source code element.
+Attributes define a set of properties on source code elements.
+An outer attribute applies to the outer context it is found in, while an inner attribute applies to the subsequent context.
```
wolffcm, line 1082: I thought the purpose of this change would be to remove the distinction of inner vs. outer attributes. Hence attributes always requiring a subsequent element to attach to. Am I missing something?

libflux/flux-core/src/parser/mod.rs
```diff
@@ -363,18 +377,28 @@ impl<'input> Parser<'input> {
         let start_pos = ast::Position::from(&self.peek().start_pos);
         let mut end = ast::Position::invalid();
 
-        // Parse attributes at the beginning of the file.
-        let attributes = self.parse_attribute_list();
+        // Parse inner attributes at the beginning of the file and hand them off to the first
+        // clause, declaration or statement that exists
+        let inner_attributes = self.parse_attribute_inner_list();
```
wolffcm, line 382: There is no such thing as "inner" or "outer" attributes now, right? I wonder if it will be clearer to omit `inner_` here?

wolffcm (approved):
This looks good to me. I think there is some language left over from an earlier iteration that refers to `inner` attributes, it might make things more clear to remove it.

libflux/flux-core/src/parser/mod.rs
nathanielc: I left the `inner` vocabulary because its still accurate in that these attributes can only apply to an inner context. Unlike before this change the attributes could apply to either an inner or outer context. Even if we never add an explicit outer syntax it makes it more specific that its about the inner context only.

docs/SPEC.md
nathanielc: This line was left in during the split, I'll remove/reword it.

PR #5330, opened 2022-11-02
feat: add PartialOrd to ast::Position

rockstar:
In the flux-lsp, we often have need to figure out if an `ast::Position`
is within an `ast::Range`. The spelling that would be best for this is
`(range.start..range.end).contains(position)`. Unfortunately, we can't
do this today because `ast::Position` doesn't implement `PartialOrd`.
This patch corrects that.

I'm not sure if the test is _really_ all that needed, since the
implementation is derived, but I added it because it's functionality
that we don't want to break in the future.

### Checklist


libflux/flux-core/src/ast/mod.rs
```diff
@@ -20,7 +20,7 @@ pub const DEFAULT_PACKAGE_NAME: &str = "main";
 
 /// Position is the AST counterpart of [`scanner::Position`].
 /// It adds serde capabilities.
-#[derive(Debug, PartialEq, Eq, Copy, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Serialize, Deserialize, PartialOrd)]
```
Marwes, line 23:
```suggestion
#[derive(Debug, PartialEq, Eq, Copy, Clone, Serialize, Deserialize, PartialOrd, Ord)]
```

Marwes (approved):
Thought this impl existed already, weird.

PR #5332, opened 2022-11-07
feat(libflux): associate registry attributes to import statements

### LINKED ISSUE #5311: Design and Implement Import Namespacing
We need Flux module imports to be unique from stdlib import paths. The design is to assume that paths are unique to a specific registry and that initially there are two conceptual registries, the standard library itself is a registry and the private organization registry.

### MVP

An import declaration will need metadata about which registry the module should be imported from. We will use attributes `@registry` to define the registry for a given import.

For example:

```
@registry("modules")
import "date" // Imports the `date` module from the `modules` registry. 
```

The absence of a registry means to import from the stdlib. The stdlib can be specified explicitly but there is no need to do so:

```
import "math" // Imports math module from the standard library

@registry("stdlib")
import "date" // Imports date module from the standard library
```

Grouping imports, it _may_ be useful to implement import grouping to reduce the amount of text needed to import many modules from a registry. 

```
@registry("modules")
import (
    "date"
    foo "bar"
) // Import both date and bar from the modules registry
```

### Future Extensions

We may want to introduce a simpler syntax in the future to make working with imports easier. One proposal is to define a set of registries and then allow import paths to use the names of those registries as the root of the import path. For example:

```
registry "fluxlang.dev" "https://fluxlang.dev/modules"

import "fluxlang.dev/foo" // Import foo from fluxlang.dev
```

Additionally there will be two implicit registries, one for the standard library and one for the private organization registry. The org registry can be defined with a specific name, lets use `modules`.  This means there is an implicit `registry "modules" "https://gateway/api/v2private/modules"` defined in the script. Then users can import there modules with this minimal syntax.

```
import "date" // Imports date from stdlib
import "modules/date" // Imports date from the private org registry
```

This means whatever name we choose for the private org registry will be a reserved name and no stdlib package will ever use that name. This will be the only reserved name that the stdlib cannot use.

### Import Semantics

Given the above here are the full import semantics.

Each import declaration has the following properties (rust psuedo code)

```rust
struct ImportDeclaration {
    pub registry: string,
    pub path: string,
    pub alias: string,
}
```

Additionally a mapping of registry names to registry URLs will be provided.

Resolving an import declaration involves combining the above information into a full module path. A full module path may be represented as a URL. For example `stdlib:///date` or `https://fluxlang.dev/modules/foo`. We can decide the details of the URL or if we want to use a URL at all later. The important part is that we have both a URL for the registry and a path for the module within that specific registry.

When resolving an import declaration the following logic is applied:

1. If the registry is an empty string, match the root path element to the list of registry names. If a registry is found use that registry URL and the module path is the remaining import path.
2. If the registry is not empty lookup the registry in the list of registry names, if it is missing error. If not use the registry URL and the full import path is the module path.

Examples showing import syntax and what fully qualified module URL is resolved

```
registry "fluxlang.dev" "https://fluxlang.dev/modules"

import "date" // stdlib:///date
import "influxdata/influxdb/schema" // stdlib:///influxdata/influxdb/schema
@registry("modules")
import ( 
   "date" // https://gateway/api/v2private/modules/date
   "data/buckets" // https://gateway/api/v2private/modules/data/buckets
)

import "fluxlang.dev/foo" // https://fluxlang.dev/modules/foo

import "modules/date"  // https://gateway/api/v2private/modules/date

@registry("stdlib")
import "fluxlang.dev/date" // stdlib:///fluxlang.dev/date
```

## TODO

- Decide on the name of the private organization registry.  Some proposed names are `modules`, `self`, `extern`, `local`. 

## DOD

- Module and Registry semantics are documented in the SPEC
- `@registry` attribute is added
- Create the logic for import resolution to a full path/URL.

jsternberg:
This change retrieves the `@registry` attribute from the package and
associates those definitions with corresponding imports in the same
package.

This gets encoded in the semantic graph and the flatbuffer so it is
accessible in both Rust and Go.

Fixes #5311.

### Checklist


libflux/flux-core/src/semantic/convert.rs
```diff
@@ -450,16 +501,54 @@ impl<'a> Converter<'a> {
...
     }
 
+    fn convert_import_path(&mut self, path: &ast::StringLit) -> (Option<String>, String) {
+        let path_str = path.value.as_str();
+        // Empty components or components completely composed of dots are just not allowed.
+        for p in path_str.split('/') {
+            if p.is_empty() || p == "." || p == ".." {
+                self.errors.push(located(
+                    path.base.location.clone(),
+                    ErrorKind::InvalidImportPath(path_str.to_string()),
```
Marwes, line 522: This could explain why the path is invalid more precisely, either by pointing precisely to the invalid segment, or explaining why a segment is invalid.

libflux/flux-core/src/semantic/nodes.rs
```diff
@@ -610,7 +610,8 @@ pub struct ImportDeclaration {
     pub loc: ast::SourceLocation,
 
     pub alias: Option<Identifier>,
-    pub path: StringLit,
+    pub registry: Option<String>,
```
Marwes, line 613: Nit: This could maybe be an enum to make it clearer

```
enum Registry {
    Stdlib,
    Other(String),
}
```

semantic/graph.go
```diff
@@ -220,8 +220,9 @@ func (p *PackageClause) Copy() Node {
 type ImportDeclaration struct {
 	Loc
 
-	As   *Identifier
-	Path *StringLiteral
+	As       *Identifier
+	Registry *string
```
Marwes, line 224: Nit: When resolving the import here on the go side you could add an error case for when a registry exist, to serve as a point to fill in the actual implementation later if nothing else.

Marwes (approved):

libflux/flux-core/src/semantic/nodes.rs
jsternberg: I like this change, but just wanted to note that this isn't going to end up influencing the flatbuffers representation. I think it adds clarity to the Rust code but it doesn't really add anything to flatbuffers in my opinion and the flatbuffers representation will likely be removed if we refactor the execution to happen in Rust anyway so I don't want to spend too much time on adding an enum to flatbuffers.

libflux/flux-core/src/semantic/convert.rs
jsternberg: I'll give this a try.

libflux/flux-core/src/semantic/nodes.rs
jsternberg: I was trying to integrate this and I think I'll just merge the current version and we can reconsider this in the future.

The reason is because I started to make the change and thought, "Well the standard library isn't really a registry. So let me use a different name." Then I started going down a rabbit hole of changing names and I'd like to get this merged to unblock other work.

libflux/flux-core/src/semantic/convert.rs
jsternberg: Ok I changed the error message to be specific about which path segment. I couldn't really get it to point to the exact location, but it should have the location of the import itself and the path segment that was invalid.

semantic/graph.go
jsternberg: This is going to follow as the next PR for the Go side.

libflux/flux-core/src/semantic/nodes.rs
Marwes: Yeah, I don't think this needs to be represented in the flatbuffer, just would be nice to be clearer on the rust side.

Marwes (approved):

Issue #5298, opened 2022-10-19
Order of operations or formatting problem 


Anaisdg:
A user was experiencing different results with the following two queries:
```
join(tables: {billed: billed, current: current}, on: ["type"])
    |> map(fn: (r) => ({r with balance: -r.peak_used_current - r.peak_used_billed + (r.peak_generated_current - r.peak_generated_billed)}))
    |> map(fn: (r) => ({r with join: 1}))

join(tables: {billed: billed, current: current}, on: ["type"])
  |> map(fn: (r) => ({ r with balance: -(r.peak_used_current-r.peak_used_billed)+(r.peak_generated_current-r.peak_generated_billed) }))
  |> map(fn: (r) => ({ r with join: 1 }))
```
From: 
https://community.influxdata.com/t/flux-query-reformatted-in-task/26840/2

Thank you!

sanderson:
@Anaisdg This is _kind of_ an order of operations thing, but more of a signed-numeric-value thing. Let me explain:

Let's assign some arbitrary values to the variables used in the `map` call:

- `r.peak_used_current` = 10
- `r.peak_used_billed` = 8
- `r.peak_generated_current` = 12
- `r.peak_generated_billed` = 6

In the first query, the expression, `-r.peak_used_current - r.peak_used_billed`—`r.peak_used_current` is signed as a negative number (prepended with `-`). So the expression resolves to:

```js
-10 - 8 + (12 - 6)
// -18 + 6
// -12
```

In the second query, the expression, `-(r.peak_used_current - r.peak_used_billed)`—The result of `r.peak_used_current - r.peak_used_billed` is signed as negative (because the group `()` is prepended with `-`). So the expression resolves to:

```js
-(10 - 8) + (12 - 6)
// -2 + 6
// 4
```

Marwes:
This is a bug in the formatter where the parenthesis on the left hand side gets removed. I thought it could be fixed easily by correcting the precedences for unary `+/-` but I couldn't get it right which I think partly is because the precedence table we have seems wrong, or at least odd? https://github.com/influxdata/flux/blob/master/docs/SPEC.md#operators-1

> Operators with a lower number have higher precedence.

[...]

PR #5337, opened 2022-11-09
fix: order of operations for unary operator

skartikey:
**Issue:**

The formatter was formatting expression with unary operator wrong `-(a + b)` to `-a + b`

**Fix:**
Corrected the [precedence](https://github.com/influxdata/flux/blob/master/docs/SPEC.md#operators-1) of the `unary operator` to fix the order of operation issue.
Flux has 4 unary operator `+/-/not/exists`, the existing implementation was not treating `+/-` different than binary `+/-`
So to correct the precedence I have given higher precedence to the unary `+/-`
Also kept the precedence of not and exists untouched to satisfy the current behaviour. 

fixes: https://github.com/influxdata/flux/issues/5298

### Checklist


skartikey:
Just realised that [SPEC](https://github.com/influxdata/flux/blob/master/docs/SPEC.md#operators-1) needs to be updated too for the unary `+/-` operator.

stdlib/generate/from_test.flux
```diff
@@ -21,7 +21,7 @@ testcase unionTwoStreamsWithEmptyGroupKeys {
     t2 =
         generate.from(
             count: 4,
-            fn: (n) => n * (-1),
+            fn: (n) => n * -1,
```
onelson, line 24: This is sort of an interesting change.

Were the parens previously required for some reason or is this just the formatter stepping in to remove them now that the bind strength has been adjusted?

libflux/flux-core/src/formatter/mod.rs
```diff
@@ -1914,28 +1914,28 @@ fn get_precedences(parent: &Node, child: &Node) -> (u32, u32) {
     let pvp: u32 = match parent {
         Node::BinaryExpr(p) => Operator::new(&p.operator).get_precedence(),
         Node::LogicalExpr(p) => Operator::new_logical(&p.operator).get_precedence(),
-        Node::UnaryExpr(p) => Operator::new(&p.operator).get_precedence(),
-        Node::FunctionExpr(_) => 3,
-        Node::PipeExpr(_) => 2,
+        Node::UnaryExpr(p) => Operator::new(&p.operator).get_precedence_unary_operator(),
```
onelson, line 1917: This is more of a nit than anything, re: consistency.

I notice in the case of logical expressions, we use `Operator::new_logical()` which sets a bool field inside the `Operator` struct, which is then used to branch inside `Operator::get_precedence()`.

I wonder if that design could work for the unary case rather than delegating to a new method. Are there barriers to this? Maybe it just makes the flow more complicated.

onelson (commented):
Had a question about whether the new `get_precedence_unary_operator()` method is needed or if it can be addressed within the existing method.

libflux/flux-core/src/formatter/mod.rs
skartikey: @onelson That makes sense to follow the same design here as well. I will do the changes.

stdlib/generate/from_test.flux
skartikey: Parenthesis is needed if a child node has lower precedence than its parent node. The same stands for the left child of a binary expression; while, for the right child, we need parenthesis if its precedence is lower or equal to its parent's.

As unary `-` has higher precedence now(earlier it was the same as binary `-`), formatter removed the parenthesis.

onelson: Yep. Makes sense. It just looks so odd as it was originally and probably should have been an indication something was awry. Thanks :pray:

PR #5353, opened 2022-12-20
feat(contrib/qxip): add contrib/qxip/logql package

lmangani:
This PR adds contrib/qxip/logql package to interact with LogQL HTTP APIs such as Grafana Loki or qryn. 
The contrib adds no go code and only wraps flux functions. Included readme explains the functionality.

Done checklist

- [x] No additional Code.


stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+// )
+// ```
+//
+// ## Metadata
+// tags: inputs
+//
+query_range = (
+        url="http://127.0.0.1:3100",
+        path="/loki/api/v1/query_range",
+        query="",
```
sanderson, line 60: This should be a required parameter so it shouldn't have a default value.

```suggestion
        query,
```

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+        limit=100,
+        step=10,
+        start=-1h,
+        end=now(),
+        orgid="",
+    ) =>
+    {
+        dstart = date.sub(from: start, d: 0d)
+        dend = date.sub(from: end, d: 0d)
```
sanderson, line 69: The way this is written, the `start` and `end` won't really do anything. This is currently subtracting a `0d` duration from each of the provided timestamps, so they return the timestamps unchanged. Also, this makes it so the `start` and `stop` parameters can only be times. Durations won't work this way.

To get the functionality you're looking for, I think instead of using `date.sub` here, you should use [`date.time`](https://docs.influxdata.com/flux/v0.x/stdlib/date/time/). This will return a time value when provided either a date or a duration. This will _not_ support unix timestamps (ints or uints).

```suggestion
        dstart = date.time(t: start)
        dend = date.time(t: end)
```

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+//   Results include points that match the specified start time.
+//   Use a relative duration, absolute time, or integer (Unix timestamp in nanoseconds).
+//   For example, `-1h`, `2019-08-28T22:00:00.801064Z`, or `1545136086801064`.
+//
+// - end: Latest time to include in results. Default is `now()`.
+//
+//   Results exclude points that match the specified stop time.
+//   Use a relative duration, absolute time, or integer (Unix timestamp in nanoseconds).
+//   For example, `now()`, `2019-08-28T22:00:00.801064Z`, or `1545136086801064`.
+//
```
sanderson, line 35: Related to my comment below about these two parameters. You won't be able to use unix timestamps here. You can use durations and timestamps.

```suggestion
// - start: Earliest time to include in results. Default is `-1h`.
//
//   Results include points that match the specified start time.
//   Use a relative duration or absolute time.
//   For example, `-1h` or `2022-01-01T22:00:00.801064Z`.
//
// - end: Latest time to include in results. Default is `now()`.
//
//   Results exclude points that match the specified stop time.
//   Use a relative duration or absolute time.
//   For example, `-1h` or `2022-01-01T22:00:00.801064Z`.
//
```

And I'm not sure about the behavior of these boundaries (if they're inclusive or exclusive). This is how the `start` and `stop` boundaries work in `range()`, but you may need to confirm this behavior in Loki.

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+// - step: Query stepping. Default is 10.
```
sanderson, line 36: I'd just copy part of the description from the Loki documentation here.

```suggestion
// - step: Query resolution step width in seconds. Default is 10.
//
//   Only applies to query types which produce a matrix response.
```

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+// - orgid: Optional Organization Id for partitioning.
```
sanderson, line 37:
```suggestion
// - orgid: Optional Loki organization ID for partitioning. Default is `""`.
```

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+                params:
+                    [
+                        "query": [query],
+                        "limit": ["${limit}"],
+                        "start": [string(v: uint(v: dstart))],
+                        "end": [string(v: uint(v: dend))],
+                        "step": ["${step}"],
+                        "csv": ["1"],
+                    ],
+                headers: if orgid != "" then ["X-Scope-OrgID": orgid] else ["X-Scope-OrgID": "0"],
```
sanderson, line 82: This seems odd to me since it will still send an OrgID in the headers, it just won't be the users' org ID. Rather than sending a `0` OrgID, why not set this to an empty dictionary? Then the header won't be sent at all:

```suggestion
                headers: if orgid != "" then ["X-Scope-OrgID": orgid] else [:],
```

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+        dend = date.sub(from: end, d: 0d)
+        response =
+            requests.get(
+                url: url + path,
+                params:
+                    [
+                        "query": [query],
+                        "limit": ["${limit}"],
+                        "start": [string(v: uint(v: dstart))],
+                        "end": [string(v: uint(v: dend))],
```
sanderson, line 78: Just a nit here. The Loki API does support RFC3339 timestamps, so you don't necessarily need to convert these to Unix nanosecond timestamps, but you can.

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
+// Package logql provides functions meant to simplify common logql queries.
+//
+// The primary function in this package is `logql.query_range()`
```
sanderson, line 3: It might be worth noting in the description here that all values are returned as string values (using `raw` mode in `csv.from`)

sanderson (changes requested):

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
...
+//     limit: 100,
+// )
+// ```
+//
+// ## Metadata
+// tags: inputs
+//
+query_range = (
+        url="http://127.0.0.1:3100",
+        path="/loki/api/v1/query_range",
```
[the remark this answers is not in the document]
lmangani: Will switch to options.

stdlib/contrib/qxip/logql/logql.flux
lmangani: This was/is intended as a formatting workaround for the date fields to work through the input. I'm actually only using the conversion to get the desired format output in ` [string(v: uint(v: dstart))]`, hence the 0d subtraction. Using RFC strings is not supported uniformly yet, so the microsecond format is the current requirement. If there's a better method to avoid the double-conversion, I'd love to learn how!

stdlib/contrib/qxip/logql/logql.flux
lmangani: See comments in the above discussion. At this time, we must use nanosecond ts for compatibility. 
> dstart = date.sub(from: start, d: 0d)
        dend = date.sub(from: end, d: 0d)

stdlib/contrib/qxip/logql/logql.flux
lmangani: Thanks! Indeed, this is way more elegant - I didn't know this was possible - thanks for the suggestion!

stdlib/contrib/qxip/logql/logql.flux
```diff
@@ -0,0 +1,87 @@
+// Package logql provides functions meant to simplify common logql queries.
```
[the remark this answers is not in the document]
lmangani: Our project [qryn](https://qryn.dev) is also in scope and this integration attempts to be neutral, hence the generic LogQL definition rather than limiting this to Loki. We have not included our own link for the same reason. 
This said, If you think this would be best for the documentation, we have absolutely no objections.

sanderson (approved):

PR #5355, opened 2023-01-01
feat(contrib/qxip): add contrib/qxip/hash package

lmangani:
This PR adds `contrib/qxip/hash` package exposing basic go hashing functions to flux scripts.
Included readme explains the functionality.

This proposal might require adaptions and is very much intended for discussion and extension as core maintainers see fit.

Done checklist:

- [x] Documentation
- [x] Basic Tests

### DIFF
--- a/stdlib/contrib/qxip/hash/hash.flux
+++ b/stdlib/contrib/qxip/hash/hash.flux
@@ -1,0 +1,68 @@
+// Package hash provides functions that convert string values to hashes.
+//
+// ## Metadata
+// introduced: NEXT
+// contributors: **GitHub**: [@lmangani](https://github.com/lmangani)
+//
+package hash
+
+
+// sha256 converts a string value to a hexadecimal hash using the SHA 256 hash algorithm.
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert a string to a SHA 256 hash
+// ```no_run
+// import "contrib/qxip/hash"
+//
+// hash.sha256(v: "Hello, world!")
+//
+// // Returns 315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin sha256 : (v: A) => string
+
+// xxhash64 converts a string value to a 64-bit hexadecimal hash using the xxHash algorithm.
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert a string to 64-bit hash using xxHash
+// ```no_run
+// import "contrib/qxip/hash"
+//
+// hash.xxhash64(v: "Hello, world!")
+//
+// // Returns 17691043854468224118
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin xxhash64 : (v: A) => string
+
+// cityhash64 converts a string value to a 64-bit hexadecimal hash using the CityHash64 algorithm.
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert a string to a 64-bit hash using CityHash64
+// ```no_run
+// import "contrib/qxip/hash"
+//
+// hash.cityhash64(v: "Hello, world!")
+//
+// // Returns 2359500134450972198
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin cityhash64 : (v: A) => string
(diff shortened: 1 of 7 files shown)


stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
+// Package hash provides functions that perform string to hash conversion
```
sanderson, line 1:
```suggestion
// Package hash provides functions that convert string values to hashes.
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+// contributors: **GitHub**: [@lmangani](https://github.com/lmangani)
+//
+package hash
+
+
+// sha256 converts a string to an hex using sha256 .
```
sanderson, line 10:
```suggestion
// sha256 converts a string value to a hexadecimal hash using the SHA 256 hash algorithm.
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+// hash.sha256(v: "Hello, world!")
+//
+// // Returns 315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin sha256 : (v: A) => string
+
+// xxhash64 converts a string to an hex using xxhash.
```
sanderson, line 30:
```suggestion
// xxhash64 converts a string value to a 64-bit hexadecimal hash using the xxHash algorithm.
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+
+
+// sha256 converts a string to an hex using sha256 .
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert hexadecimal string to sha256 hash
```
sanderson, line 17:
```suggestion
// ### Convert a string to a SHA 256 hash
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+builtin sha256 : (v: A) => string
+
+// xxhash64 converts a string to an hex using xxhash.
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert hexadecimal string to hash using xxhash64.
```
sanderson, line 37:
```suggestion
// ### Convert a string to 64-bit hash using xxHash
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+// hash.xxhash64(v: "Hello, world!")
+//
+// // Returns 17691043854468224118
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin xxhash64 : (v: A) => string
+
+// cityhash64 converts a string to an hex using CityHash64.
```
sanderson, line 50:
```suggestion
// cityhash64 converts a string value to a 64-bit hexadecimal hash using the CityHash64 algorithm.
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -0,0 +1,68 @@
...
+//
+// ## Parameters
+//
+// - v: String to hash.
+//
+// ## Examples
+// ### Convert hexadecimal string to hash using xxhash64.
```
sanderson, line 57:
```suggestion
// ### Convert a string to a 64-bit hash using CityHash64
```

sanderson (changes requested):
@lmangani Looks great. Some minor doc suggestions, but that's it. Thanks!

lmangani:
Thanks @sanderson for the review and for the doc corrections! 
Just to be sure - is the cityhash source include method considered "proper" (as opposed to using the library) for inclusion in Fluxlib and as reference for future contribs with go code?

wolffcm:
@lmangani Did you copy the code for `CityHash` instead of using the library for any particular reason? It would be preferable to import it in the normal way, with an `import` statement and an entry in `go.mod`.

sanderson (approved):

wolffcm (changes requested):
I'd like to understand how/if we can import `CityHash` in the normal way.

lmangani:
> @lmangani Did you copy the code for `CityHash` instead of using the library for any particular reason? It would be preferable to import it in the normal way, with an `import` statement and an entry in `go.mod`.

Just following the [contribution guidelines](https://github.com/influxdata/flux/blob/master/CONTRIBUTING.md#use-of-third-party-packages) 

> A third-party package is defined as one that is not part of the standard Go distribution. Generally speaking, we prefer to minimize our use of third-party packages and avoid them unless absolutely necessarily. We'll often write a little bit of code rather than pull in a third-party package. To maximize the chance your change will be accepted, use only the standard libraries, or the third-party packages we have decided to use.

wolffcm:
@lmangani This makes sense. Thanks for the rationale there. It's been a while since I've looked at the contribution guidelines.

wolffcm (approved):

lmangani:
Thanks! Shall CityHash or any other hashing library become part of the Flux core, I'll update the code to use it/them.

PR #5367, opened 2023-01-18
Support non-uniformly spaced data in regression

jat255:
Prior to this change, the statsmodels linear regression calculation assumed that the input table data was uniformly spaced (the "x" column was just a uniformly increasing integer). This resulted in incorrect "slope" calculations on data that was nonuniformly-spaced. This change modifies the function's behavior to map the input `_time` columns to a time difference in nanoseconds, and use that value for the calculation of the regression, meaning its output now matches other tools such as Excel's or Libreoffice Calc's "trend lines" analysis.

I'm a very basic user of Influx and am totally unfamiliar with its development, so I'm not sure how or where to put test cases and will need some direction on that.

@Anaisdg will probably want to take a look

### Checklist

- [x] 🔗 ~~Reference related issues~~ (did not create one)
- [ ] 🏃 Test cases are included to exercise the new code (I'm not really sure how to do this)

### DIFF
--- a/stdlib/contrib/anaisdg/statsmodels/linearreg.flux
+++ b/stdlib/contrib/anaisdg/statsmodels/linearreg.flux
@@ -17,14 +17,14 @@ import "generate"
 // Output data includes the following columns:
 //
 // - **N**: Number of points in the calculation.
-// - **slope**: Slope of the calculated regression.
+// - **slope**: Slope of the calculated regression (slope "x" unit is nanoseconds).
 // - **sx**: Sum of x.
 // - **sxx**: Sum of x squared.
 // - **sxy**: Sum of x*y.
 // - **sy**: Sum of y.
 // - **errors**: Residual sum of squares.
 //   Defined by `(r.y - r.y_hat) ^ 2` in this context
-// - **x**: An index [1,2,3,4...n], with the assumption that the timestamps are regularly spaced.
+// - **x**: An index [1,2,3,4...n] of the input _time column in nanoseconds.
 // - **y**: Field value
 // - **y\_hat**: Linear regression values
 //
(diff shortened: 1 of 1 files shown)


stdlib/contrib/anaisdg/statsmodels/linearreg.flux
```diff
@@ -43,11 +43,17 @@ import "generate"
...
+            fn: (key) => true,
+            idx: 0
+        )
     renameAndSum =
         tables
             |> rename(columns: {_value: "y"})
-            |> map(fn: (r) => ({r with x: 1.0}))
-            |> cumulativeSum(columns: ["x"])
+            // calculate x as difference from first value (in seconds)
+            |> map(fn: (r) => ({ r with x : float(v: uint(v: r._time)  - uint(v: firstRecord._time)) }))
```
sanderson, line 56: This method assumes that the first timestamp in the first table in the stream of table should be used as the "origin" time for all other tables. I don't think that's a safe assumption. I think a safer assumption would be to use the `_start` value of the row as the origin timestamp. This time represents the query boundary, or if the data is windowed, the window boundary. This would also negate the need to use `findRecord` at the beginning of the function.

```suggestion
            |> map(fn: (r) => ({ r with x : float(v: uint(v: r._time)  - uint(v: r._start)) }))
```

sanderson (changes requested):

stdlib/contrib/anaisdg/statsmodels/linearreg.flux
```diff
@@ -43,11 +43,17 @@ import "generate"
...
+        |> findRecord(
+            fn: (key) => true,
+            idx: 0
+        )
     renameAndSum =
         tables
             |> rename(columns: {_value: "y"})
-            |> map(fn: (r) => ({r with x: 1.0}))
-            |> cumulativeSum(columns: ["x"])
+            // calculate x as difference from first value (in seconds)
```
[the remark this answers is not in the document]
jat255: Good catch, this was a comment left over from a previous version where I did things in seconds

stdlib/contrib/anaisdg/statsmodels/linearreg.flux
jat255: I suppose I'm fine with either, but if you use `r._start`, then your regression will be relative to an arbitrary view point (i.e. the beginning of your window/query), rather than the actual data. Playing around with this a bit, if I use a relative query window (i.e. last thirty days), then the slope of my regression changes (very slightly, but still) every time I run the query, since its calculating relative to the query rather than the actual first data point, even though the data contained within my query window has not changed.

> ...first table in the stream of table should be used as the "origin" time for all other tables. I don't think that's a safe assumption.

I use Influx for very simple data, so I don't think I appreciate the consequences of this. I'm willing to defer to your suggestion, but my gut feeling is that it's not "technically" correct, since the regression should be performed on just the data itself, rather than as a function of the viewport.

jat255:
This is stale because I was waiting for further feedback

jat255:
Bump again for feedback from @sanderson  or @Anaisdg

Issue #1764, opened 2019-08-23
Remove EnvironmentSecretService


nathanielc:
We should be be exposing arbitrary env vars. This may be OK in the context of secrets, but there is not immediate need for it so I would rather add it once we have a concrete need and can address the security implications.

aanthony1243:
I think it's reasonable to say that there's a difference between `flux` on the command line as a standalone scripting language and `flux` embedded in `influxdb`.   We can segregate the packages so that flux-local packages are not accidentally made a part of the internal influxdb.

PR #1753, opened 2019-08-23
feat(dependencies): added secret service interface

sauren-khosla:
### Done checklist
- [ ] docs/SPEC.md updated
- [ ] Test cases written

PR #5370, opened 2023-01-21
feat(contrib/qxip): add contrib/qxip/env package

lmangani:
This PR adds `contrib/qxip/env` package exposing a limited scope of OS `ENV` variables to flux scripts.

Included readme explains the functionality.

#### PR Note
The original ENV reading functionality available in Secretservice seems to have been intentionally [removed](https://github.com/influxdata/flux/issues/1764) due to various [security concerns](https://github.com/influxdata/flux/pull/1753#issuecomment-524838180) such as allowing scripts to read InfluxDB Cloud ENVs, etc. This package limits the allowed ENV scope _(`FLUX_` or any other prefix considered safe by the Influx Team)_ to attempt complying with the previous criteria. 

Done checklist:

- [x]  Documentation

### DIFF
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -1,0 +1,39 @@
+// Package env provides a function for reading ENV variables starting with the FLUX_ prefix.
+//
+// ## Metadata
+// introduced: NEXT
+// tags: secrets,security,env
+//
+package env
+
+
+// get retrieves a key variable from the process ENV.
+//
+// ## Parameters
+// - key: ENV key to retrieve.
+//
+// ## Examples
+//
+// ### Retrive a key from ENV variables
+// ```no_run
+// import "contrib/qxip/env"
+//
+// env.get(key: "KEY_NAME")
+// ```
+//
+// ### Populate sensitive credentials with ENV variables//
+// ```no_run
+// import "sql"
+// import "contrib/qxip/env"
+//
+// username = env.get(key: "FLUX_USERNAME")
+// password = env.get(key: "FLUX_PASSWORD")
+//
+// sql.from(
+//     driverName: "postgres",
+//     dataSourceName: "postgresql://${username}:${password}@localhost",
+//     query: "SELECT * FROM example-table",
+// )
+// ```
+//
+builtin get : (key: string) => string
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -14,7 +14,7 @@ package env
 //
 // ## Examples
 //
-// ### Retrive a key from ENV variables
+// ### Retrieve an environment variable
 // ```no_run
 // import "contrib/qxip/env"
 //
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -18,7 +18,7 @@ package env
 // ```no_run
 // import "contrib/qxip/env"
 //
-// env.get(key: "KEY_NAME")
+// env.get(key: "FLUX_KEY_NAME")
 // ```
 //
 // ### Populate sensitive credentials with ENV variables//
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -21,7 +21,7 @@ package env
 // env.get(key: "FLUX_KEY_NAME")
 // ```
 //
-// ### Populate sensitive credentials with ENV variables//
+// ### Populate sensitive credentials with ENV variables
 // ```no_run
 // import "sql"
 // import "contrib/qxip/env"
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -10,7 +10,7 @@ package env
 // get retrieves an environment variable from the process ENV.
 //
 // ## Parameters
-// - key: ENV key to retrieve.
+// - key: Environment variable key to retrieve.
 //
 // ## Examples
 //
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -1,4 +1,4 @@
-// Package env provides a function for reading ENV variables starting with the FLUX_ prefix.
+// Package env provides a function for reading environment variables starting with the `FLUX_` prefix.
 //
 // ## Metadata
 // introduced: NEXT
--- a/stdlib/contrib/qxip/env/env.flux
+++ b/stdlib/contrib/qxip/env/env.flux
@@ -7,7 +7,7 @@
 package env
 
 
-// get retrieves a key variable from the process ENV.
+// get retrieves an environment variable from the process ENV.
 //
 // ## Parameters
 // - key: ENV key to retrieve.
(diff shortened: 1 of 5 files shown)


stdlib/contrib/qxip/env/env.flux
```diff
@@ -0,0 +1,39 @@
...
+
+
+// get retrieves a key variable from the process ENV.
+//
+// ## Parameters
+// - key: ENV key to retrieve.
+//
+// ## Examples
+//
+// ### Retrive a key from ENV variables
```
sanderson, line 17:
```suggestion
// ### Retrieve an environment variable
```

stdlib/contrib/qxip/env/env.flux
```diff
@@ -0,0 +1,39 @@
...
+// ```no_run
+// import "contrib/qxip/env"
+//
+// env.get(key: "FLUX_KEY_NAME")
+// ```
+//
+// ### Populate sensitive credentials with ENV variables//
```
sanderson, line 24:
```suggestion
// ### Populate sensitive credentials with ENV variables
```

stdlib/contrib/qxip/env/env.flux
```diff
@@ -0,0 +1,39 @@
+// Package env provides a function for reading ENV variables starting with the FLUX_ prefix.
```
sanderson, line 1:
```suggestion
// Package env provides a function for reading environment variables starting with the `FLUX_` prefix.
```

stdlib/contrib/qxip/env/env.flux
```diff
@@ -0,0 +1,39 @@
...
+//
+// ## Metadata
+// introduced: NEXT
+// tags: secrets,security,env
+//
+package env
+
+
+// get retrieves a key variable from the process ENV.
```
sanderson, line 10:
```suggestion
// get retrieves an environment variable from the process ENV.
```

stdlib/contrib/qxip/env/env.flux
```diff
@@ -0,0 +1,39 @@
...
+//
+// ## Parameters
+// - key: ENV key to retrieve.
```
sanderson, line 13:
```suggestion
// - key: Environment variable key to retrieve.
```

sanderson (changes requested):
Thanks for this @lmangani. I have a few doc suggestions.

While this does take into account the security concern that originally led to the removal of access to environment variables, I do still feel this could potentially pose a risk. With that said, I'm not that well-versed in OS-level security to definitively say this would or wouldn't. I think @wolffcm or @influxdata/secops should weigh in here.

My gut says that the safer approach would be to build an implementation for a 3rd party secret store like Vault or Bitwarden, but these have their downsides as well.

lmangani:
> Thanks for this @lmangani. I have a few doc suggestions.
> 
> While this does take into account the security concern that originally led to the removal of access to environment variables, I do still feel this could potentially pose a risk. 

Thanks! As per PR note, I most definitely agree this requires a deeper discussion to make sure it poses zero risks,

> With that said, I'm not that well-versed in OS-level security to definitively say this would or wouldn't. I think @wolffcm or @influxdata/secops should weigh in here.

I believe I've read about ways to exclude specific modules from being part in the InfluxDB bundled Flux versions without impacting Fluxlib. Could this be one such case where the contrib really only benefits Flux integrations as a library or console executions?

> My gut says that the safer approach would be to build an implementation for a 3rd party secret store like Vault or Bitwarden, but these have their downsides as well.

Our use case is much simpler (passing ENVs to Flux instances as secret alternative for standalone executions) and would ideally benefit a simple solution introducing no 3rd party components.

Thanks for considering this!

gavincabbage:
+1 re needing @influxdata/secops insight on this

wolffcm:
I don't think that this PR (even if only exposing environment variables that start with `FLUX`) would be acceptable, unfortunately.

> I believe I've read about ways to exclude specific modules from being part in the InfluxDB bundled Flux versions without impacting Fluxlib. Could this be one such case where the contrib really only benefits Flux integrations as a library or console executions?

@lmangani Do you happen to remember where you read that?

We _do_ have a secret service that is accessible in cloud:
https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/secrets/get/

And it does happen to have an implementation that just uses the environment:
https://github.com/influxdata/flux/blob/master/dependencies/secret/environment_secret_service.go

However this implementation is not used anywhere. It would be possible to configure a Go program where Flux is embedded to use the `EnvironmentSecretService` instead of the the default. However this is inherently insecure and I don't know that we would want to make it available, even in OSS. I'm sure @influxdata/secops will have an opinion on this.

lmangani:
> I don't think that this PR (even if only exposing environment variables that start with `FLUX`) would be acceptable, unfortunately.

Thanks for the review @wolffcm I appreciate the time looking into this and I cannot say this was unexpected but still worth opening to spark some discussions. 

>> I believe I've read about ways to exclude specific modules from being part in the InfluxDB bundled Flux versions without impacting Fluxlib. Could this be one such case where the contrib really only benefits Flux integrations as a library or console executions?
>@lmangani Do you happen to remember where you read that?

This is the post: https://github.com/influxdata/flux/issues/1764#issuecomment-524854463 by @aanthony1243 

```
I think it's reasonable to say that there's a difference between flux on the command line as a standalone scripting language and flux embedded in influxdb. We can segregate the packages so that flux-local packages are not accidentally made a part of the internal influxdb.
```

Purely to play devil's advocate, while I perfectly understand the influxdb integration security concern, I would agree with the comment and assume this would have to be dealt with by the affected application integrating flux, rather than be itself a limit in the extensibility of the flux library itself. To make a case for this, in my specific [use-case](https://github.com/metrico/fluXpipe) I'm only using fluxlib and environment variables would not represent a security issue but rather a resourceful feature to build security in of itself _(in absence of secrets)_. Same would apply to command line executions. 

Just my 2 cents as an external user and integrator, with much love and respect for all team at InfluxData!

wolffcm:
@lmangani 

> I would agree with the comment and assume this would have to be dealt with by the affected application integrating flux, rather than being a limit in the extensibility of the flux library itself. 

By the "Flux library" do you mean the Go library and its API? For what it's worth, you _can_ extend it to use environment variables as secrets, and you would not have to change any code in the Flux repo to do so.

In your case, you inject the dependencies here:
https://github.com/metrico/fluXpipe/blob/21e76626c9ee24463fb41873fd1d7b0e6d4d65ca/fluxpipe-server.go#L103

which ultimately creates an empty secret store here:
https://github.com/influxdata/flux/blob/1eba4ff2ba6002756333eed3f4c95e6f6c79e9cc/dependencies.go#L109

But instead in `flux-pipe` you could create a set of dependencies that use the `EnvironmentSecretService` that I link above. If you did this, then I would expect you to be able to access env vars from Flux, as in [this example](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/secrets/get/#retrive-a-key-from-the-influxdb-secret-store)

lmangani:
Thanks @wolffcm I appreciate the input and completely understand the decision. On to the next module!

PS: Thanks for the advice! I most definitely would like to try, but I didn't understand how to inject dependencies to get this effect

lmangani:
@wolffcm I just wanted to thank you once again for pointing me in the right direction. It works a beauty! I've never been this enthusiastic about a PR not being accepted before 😄 
image

wolffcm:
@lmangani Really glad it worked out! 🎉

PR #5371, opened 2023-01-21
feat(contrib/qxip/hash): add sha1, md5, base64, hmac

lmangani:
This PR extends the `contrib/qxip/hash` package with additional functions for `sha1`, `base64`, `md5`, `hmac`

Note: The primary goal of this extension is powering flux native extensions dealing with auth token calculation etc. _(ie: s3 client we're working on)_ without having to use or import additional go code. 

- [x] Updated Documentation
- [x] Extended Test Converage

### DIFF
--- a/stdlib/contrib/qxip/hash/hash.flux
+++ b/stdlib/contrib/qxip/hash/hash.flux
@@ -34,7 +34,7 @@ builtin sha256 : (v: A) => string
 // - v: String to hash.
 //
 // ## Examples
-// ### Convert a string to a SHA 256 hash
+// ### Convert a string to a SHA-1 hash
 // ```no_run
 // import "contrib/qxip/hash"
 //
--- a/stdlib/contrib/qxip/hash/hash.flux
+++ b/stdlib/contrib/qxip/hash/hash.flux
@@ -114,7 +114,7 @@ builtin b64 : (v: A) => string
 // - v: String to hash.
 //
 // ## Examples
-// ### Convert a string to a 64-bit hash using CityHash64
+// ### Convert a string to an MD5 hash
 // ```no_run
 // import "contrib/qxip/hash"
 //
--- a/stdlib/contrib/qxip/hash/hash.flux
+++ b/stdlib/contrib/qxip/hash/hash.flux
@@ -135,7 +135,7 @@ builtin md5 : (v: A) => string
 // - k: Key to sign hash.
 //
 // ## Examples
-// ### Converts a string, key to base64 signed hash for S3 auth.
+// ### Convert a string and key to a base64-signed hash
 // ```no_run
 // import "contrib/qxip/hash"
 //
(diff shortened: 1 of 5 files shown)


stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -66,3 +86,64 @@ builtin xxhash64 : (v: A) => string
 // ## Metadata
 // tag: type-conversion
 builtin cityhash64 : (v: A) => string
+
+// b64 converts a string value to Base64 string.
```
sanderson, line 90:
```suggestion
// b64 converts a string value to a Base64 string.
```

stdlib/contrib/qxip/hash/hash.flux
```diff
@@ -66,3 +86,64 @@ builtin xxhash64 : (v: A) => string
...
+// hash.md5(v: "Hello, world!")
+//
+// // Returns 2359500134450972198
+// ```
+//
+// ## Metadata
+// tag: type-conversion
+builtin md5 : (v: A) => string
+
+// hmac converts a string value to an MD5 signed sha1 hash.
```
sanderson, line 130:
```suggestion
// hmac converts a string value to an MD5-signed SHA-1 hash.
```

sanderson (changes requested):
Thanks @lmangani! A few minor suggestions on the docs, but this looks great.

lmangani:
Thank you so much @sanderson! Next attempt in line is an S3 HTTP client for flux 100% made in flux 🤞

sanderson (approved):

gavincabbage (approved):

lmangani:
@sanderson going paranoid mode temporarily: building with the latest flux master causes the following error, which related to this PR. I'm not sure if this is just about how I'm building my binary, but I'd rather be safe than sorry:

```
panic: missing builtin values [sha1 b64 md5 hmac], extra builtin values []

goroutine 1 [running]:
github.com/influxdata/flux/runtime.FinalizeBuiltIns(...)
        /home/runner/go/pkg/mod/github.com/influxdata/flux@v0.192.0/runtime/global.go:97
github.com/influxdata/flux/fluxinit.FluxInit(...)
        /home/runner/go/pkg/mod/github.com/influxdata/flux@v0.192.0/fluxinit/init.go:23
github.com/influxdata/flux/fluxinit/static.init.0()
        /home/runner/go/pkg/mod/github.com/influxdata/flux@v0.192.0/fluxinit/static/static.go:10 +0x3f
```

Did I miss something in the new code? The previous hash contrib was working fine. Apologies in advance!

sanderson:
@lmangani I think you need to regenerate the stdlib before you build.

```sh
> go generate ./stdlib ./libflux/go/libflux
> make generate
```

lmangani:
I think this could be caused by me building using the master for the go library and a release (v0.192.0) as the library (or vice versa) which does not include yet the changes? As long as this is a build issue only on my side, all is well... Thanks @sanderson

PR #5375, opened 2023-01-30
Create stale.yml

gavincabbage:
Add a stalebot to make issue management easier.


brettbuddin (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '21 1 * * *'
```
appletreeisyellow, line 10: Just curious, does this mean check any PR that has no activity at 1:21 am? What time zone is it?

gavincabbage: Yes, good catch. I'm not sure why that is the default time, I didn't set it. The timezone will be UTC for GH action cron schedules, based on a quick Google. In the end, it doesn't matter much when it happens as long as it checks once a day or so.

appletreeisyellow: Got it. Thanks!

appletreeisyellow (approved):

helenosheaa:
just checked the defaults so this will trigger after 60 days of inactivity with a label and then 7 days after that it will close the issue/pr - seems fair to me.

we may want to add [stale issue message](https://github.com/actions/stale#stale-issue-message) and [stale pr message](https://github.com/actions/stale#stale-issue-message) to add a comment with the label that warns what will happen

you'll need to edit your commit message to be semantic

PR #5377, opened 2023-01-31
fix(array): use-after-free in optimized String array

mhilton:
The optimized string array, which is used whan all column values are identical, was caching the string value that was passed to it. If this value came from an arrow String array then the string's data would be in the buffer backing that array. Once the array is finished with and freed the new string array containing the reference to the data buffer remains and prevents the entire buffer from being garbage collected.

The changes to the executetest package show that this is the case by scribbling over a buffer's memory when it is reported as free. This caused a number of tests to fail with strings that contain the data pattern rather than the expected result.

This could cause a correctness error if one uses a more sophisticated allocator that maintains a backing buffer. This would be a legitimate design choice in order to relieve pressure on the garbage collector.

In order to correct this the StringBuilder has been enhanced to always copy the input string into a new buffer whether it is in the optimised mode, or not. This buffer is allocated from the memory allocator. Once the builder has been converted into a string array retrieving a string from this array always makes a string copy of the data, this time allocated by the standard runtime, so that the string can be used with other structures without the possibility that it will keep a larger data buffer from being garbage collected.

These changes are likely to have an impact on the performance of certain operations. Increasing the amount of memory copying will mean that some transform heavy workloads take longer. I don't know how to ascertain the impact yet. In order to alleviate this slightly a new byte array based method (ValueBytes) has been added to the string array. This does not copy the data but instead returns a slice of the data backing the array. This is to be used when the caller knows that they will be using the data without storing a reference, or immediately copying it to a different location.

### Checklist

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -2,9 +2,11 @@ package array
 
 import (
 	"strconv"
+	"sync/atomic"
 
 	"github.com/apache/arrow/go/v7/arrow"
 	"github.com/apache/arrow/go/v7/arrow/array"
+	arrowmem "github.com/apache/arrow/go/v7/arrow/memory"
 	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/internal/errors"
 	"github.com/influxdata/flux/memory"
@@ -102,9 +104,9 @@ type Builder interface {
 }
 
 type String struct {
-	value  string
 	length int
 	data   *array.Binary
+	value  *stringValue
 }
 
 // NewStringFromBinaryArray creates an instance of String from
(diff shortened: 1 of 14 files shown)


brettbuddin (approved):
This all seems reasonable, @mhilton. We'll need to cut a pre-release build to canary in Cloud 2 so we can assess the performance impact.

gavincabbage:
@brettbuddin I spoke with @mhilton today and general plan, as you've mentioned, will be to canary in staging tomorrow morning and then in the region affected by the EAR Friday morning (EU morning). Then GA next week if that all goes well.

PR #5386, opened 2023-02-22
feat: add parameter to histogramQuantile

### LINKED ISSUE #5385: Add parameter to `histogramQuantile` to control non-monotonic behavior
Often when users invoke `histogramQuantile` especially in cloud, they may be aggregating data from several sources. Due to the eventually consistent nature of InfluxDB, it's possible that missing data makes the bins not be monotonically increasing when sorted by upper bound.

The current behavior for this is to produce an error:
```
histogram record counts are not monotonic
```

We should instead provide a way for users to specify different behavior if that's what they want:
- Produce no output
- Force the bins to be monotonic

wolffcm:
Closes #5385.

This adds a parameter to `histogramQuantile` that will let the user decide upon an alternative behavior when bin counts are not monotonically increasing. ~~Currently only `"error"` is supported in this PR, I just wanted to get a bit of feedback before I do more.~~ 

As described in the Flux stdlib doc comments, this adds a parameter called `onNonmonotonic` that accepts three values:
- `error` (default)
- `force`
- `drop`

@sanderson Do you have any thoughts on this approach? My thinking is that we have a similar parameter for `filter` called `onEmpty`, so I'm kind of following that pattern.

### Checklist


sanderson:
@wolffcm This looks good to me. I notice `force` and `drop` as other potential options in the source. Are these just placeholders?

wolffcm:
> I notice force and drop as other potential options in the source. Are these just placeholders?

Yes, that's right. They are unused, but I just wanted to indicate that they would be potential values.

sanderson:
> Yes, that's right. They are unused, but I just wanted to indicate that they would be potential values.

Sounds good. Just wanted to make sure.

mhilton (approved):

PR #5401, opened 2023-03-27
chore: Update CODEOWNERS

gavincabbage:


gavincabbage:
I think I can merge despite the semantic check failing so I am going to ignore that

PR #5404, opened 2023-03-29
fix(stdlib): panic in chunking outer joins

mhilton:
An outer join that creates large tables (more than 3000 rows) caused a panic when splitting the resulting table into chunks. This was due to the stop parameter becoming lower than that start parameter, attempting to use nonsensical ranges. Tables larger than 2000 rows also potentially dropped rows.

The fix is to ensure that the stop parameter is set relative to the start.

### Checklist


wolffcm (approved):
This looks good. Nice work.

I can see why this was never tested. It would have been good to be able to very the buffer size to be very small for tests to catch stuff like this. (I have been writing tests with batch size set to 1 for IOx and it has kept me from writing tons of bugs)

PR #5408, opened 2023-04-11
fix: make C FFI more robust with respect to malformed input

wolffcm:
An audit discovered a few places where we could be more rigorous about checking inputs for valid data. This PR addresses the issues and adds tests for them.


crepererum (approved):

crepererum:
BTW: another way to guard your FFI interfaces against all sorts of panics is [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html). Just use it within the top-level FFI method, try to downcast the FFI payload to `String` and `&'static str` (both are possible) and create your FFI error type from that.

PR #5419, opened 2023-05-26
fix: make histogramQuantile handle case of zero samples

### LINKED ISSUE #5415: The histogramQuantile function returns an incorrect value when there are no observations in the histogram
## Description

There is an issue with the Flux `histogramQuantile()` function that causes it to display incorrect results. Specifically, when trying to calculate the quantile for a histogram with no new observations (therefore, a total count of observations in the buckets is zero), the function fails to return the expected result.

## Steps to Reproduce

1. There is a Prometheus histogram in InfluxDB (v2 mapping) that is not currently being incremented due to zero traffic condition.
2. Try to plot the quantile of the histogram using the `histogramQuantile()` function.
3. Notice that the resulting plot defaults the result to the value of the highest bucket boundary in the histogram, even though there have been no observations in this period.

## Expected Behaviour

The `histogramQuantile()` function should return nothing if the total count of the buckets is zero. 

## Actual Behaviour

The `histogramQuantile()` function returns the value of the highest bucket. This is because `rankIdx` will always be the highest bucket, as `rank ≥ b.count` will always be true if `rank := t.spec.Quantile * totalCount` is zero and `b.count` is also zero. The `rankIdx` switch case then assumes that the quantile lies above the highest upper bounds and returns the highest upper bound as the result.

## Possible Solution

Add a return condition to the `computeQuantile()` function in `histogram_quantile.go` that checks if the `totalCount` is zero.

```go
if totalCount == 0 {
    return 0, false, nil
  }
```

wolffcm:
Closes #5415 

When there are no observations/samples in a histogram (all zeros for each bucket) produce a `null` value.

### Checklist


gavincabbage (approved):

stdlib/universe/histogram_quantile.go
```diff
@@ -251,51 +251,97 @@ func (t histogramQuantileTransformation) Process(id execute.DatasetID, tbl flux.
...
+				return false
+			}
+		} else {
+			prevCount = cdf[i].count
+		}
+	}
+	return true
+}
+
+func (t *histogramQuantileTransformation) computeQuantile(cdf []bucket) (quantileResult, error) {
```
wolffcm, line 313: The issue here was that the Flux stdlib function `histogram_quantile` gets a wrong answer for some input data.

This function accepts a cumulative distribution function (a cumulative histogram produced from the input table data) and produces the requested quantile.

When the cdf contains all zeroes, this function would return the bound of the last histogram bucket, which is incorrect. The right thing to do for that case is to return a null value, since we can't compute a quantile if we didn't actually receive any observations.

stdlib/universe/histogram_quantile.go
```diff
@@ -251,51 +251,97 @@ func (t histogramQuantileTransformation) Process(id execute.DatasetID, tbl flux.
...
+		switch t.spec.OnNonmonotonic {
+		case onNonmonotonicError:
+			return quantileResult{}, errors.New(codes.FailedPrecondition, "histogram records counts are not monotonic")
+		case onNonmonotonicDrop:
+			return quantileResult{action: drop}, nil
+		default:
+			// "force" is not possible because isMonotonic will fix the buckets
+			return quantileResult{}, errors.Newf(codes.Internal, "unknown or unexpected value for onNonmonotonic: %q", t.spec.OnNonmonotonic)
+		}
+	}
```
wolffcm, line 328: Sometimes the histogram buckets are not monotonic (which they should be if they are cumulative) due to late-arriving data on the edge. The `OnNonmonotonic` parameter describes what to do in this case.

Checking for monotonicity first (and fixing if needed and requested by the user) avoids a bug that occurred when the total observation count was pulled from the last bucket before it was "fixed" in the case of forcing monotonicity.

This is not really related to the issue the user found but I saw it here and fixed it. The test case `histogramQuantileOnNonmonotonicForceLastBucket` below verifies this fix.

stdlib/universe/histogram_quantile.go
```diff
@@ -251,51 +251,97 @@ func (t histogramQuantileTransformation) Process(id execute.DatasetID, tbl flux.
...
+
 	// Find rank index and check counts are monotonic
-	prevCount := 0.0
 	totalCount := cdf[len(cdf)-1].count
+	if totalCount == 0 {
+		// Produce a null value if there were no samples
+		return quantileResult{action: appendNil}, nil
+	}
```
wolffcm, line 335: Here is where we bail and produce a null value for the case of zero observations.

PR #5436, opened 2023-10-25
fix(flux-core): infinite loop protection in the parser

mhilton:
Update all loops using self.more in the parser to detect if they get stuck attempting to process the same token multiple times. This has been observed to cause the parser to get into an infinite loop with some erroneus inputs.

The protection code was copied from the parse_array_items_rest and applied everywhere the parser could conceivably get stuck. It is not clear that it is possible to get stuck in all the places that the protection was added.

### Checklist

### DIFF
--- a/libflux/flux-core/src/parser/mod.rs
+++ b/libflux/flux-core/src/parser/mod.rs
@@ -734,12 +734,23 @@ impl<'input> Parser<'input> {
     // Parameters = Parameter { "," Parameter } .
     fn parse_parameters(&mut self) -> Vec<ParameterType> {
         let mut params = Vec::<ParameterType>::new();
+        // keep track of the last token's byte offsets
+        let mut last = self.peek().start_offset;
         while self.more() {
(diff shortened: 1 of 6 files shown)


btasker (approved):
LGTM - tested against the original testcase and it definitely resolves that.

libflux/flux-core/src/parser/mod.rs
```diff
@@ -734,12 +734,23 @@ impl<'input> Parser<'input> {
...
             }
+
+            // If we parse the same token twice in a row,
+            // it means we've hit a parse error, and that
+            // we're now in an infinite loop.
+            let this = self.peek().start_offset;
+            if last == this {
+                break;
+            }
+            last = this;
```
appletreeisyellow, line 753: Thanks for the comment 👍

appletreeisyellow (approved):
Thanks for the fix! The new test cases are very helpful for review!

Issue #5349, opened 2022-12-09
timeWeightedAverage interpolation off?


Anaisdg:
When I use the function I don't get an expected value. 
```
import "array"
array.from(rows: [{_time: 2022-01-02, _value: 1}, {_time: 2022-01-03, _value: 2}, {_time: 2022-01-04, _value: 3}])
|>range(start: 2022-01-01)
  |> timeWeightedAvg(unit: 1d)
```
gives 171.40402444422162

When I picture a straight line and take the integral for (1,1)(2,2)(3,3) I don't get a value remotely close to that. 
Thank you!

sanderson:
I think the function is working as expected. `timeWeightedAvg()` use the `_start` and `_stop` times to interpolate values between. In this query, the stop time is `now()`, so `timeWeightedAvg()` is interpolating all the values between `2022-01-04T00:00:00Z` and `now()`. That's why the returned average is so high. If you explicitly set the stop time in `range()` to `2022-01-05`, it should give you the expected value.

```js
import "array"

array.from(
    rows: [
        {_time: 2022-01-02, _value: 1},
        {_time: 2022-01-03, _value: 2},
        {_time: 2022-01-04, _value: 3}
    ]
)
    |>range(start: 2022-01-01, stop: 2022-01-05)
    |> timeWeightedAvg(unit: 1d)
```

Anaisdg:
Thanks @sanderson

alberanid:
Thanks @Anaisdg and @sanderson for your analysis, but I in my experience the `timeWeightedAvg` function also produces unexpected values when you select a safe start and stop range **if there are gaps in the data**. That's not the case in your test data, so I agree that the result is sound in your case, but only because its data are stored at regular intervals.

There are various issues about this problem, like https://github.com/influxdata/influxdb/issues/7445 and the more recent https://github.com/influxdata/influxdb/issues/23929

A dataset that present this problem can be found here: https://community.influxdata.com/t/how-to-aggregate-non-regular-values-using-a-time-weighted-average/27619

As others are saying, in the IoT world is quite common to have sensors that send data only on changes, or when they wake up responding to certain events, and in that cases `timeWeightedAvg` can't be used. I agree that this probably opens various questions, like "what should be used as 'previous' value when you have a gap in the data", but then again the current implementation can't be used on series that are not regularly timed.

Thanks!

sanderson:
[...]

Issue #4782, opened 2022-05-23
Integral function not returning expected results 


Anaisdg:
https://community.influxdata.com/t/integral-function-not-returning-expected/25061
Thank you

amykyta3:
@Anaisdg I have a more extreme example of integral() mis-calculating the result.
It seems to be an issue with the linear interpolation mode.

For example:
```flux
import "experimental/array"

array.from(rows: [
    {_value: 0, _start: 2022-12-04T00:00:00.000Z, _stop: 2022-12-04T13:00:00.000Z, _time: 2022-12-04T12:00:00.000Z},
    {_value: 1, _start: 2022-12-04T00:00:00.000Z, _stop: 2022-12-04T13:00:00.000Z, _time: 2022-12-04T12:01:00.000Z},
    {_value: 1, _start: 2022-12-04T00:00:00.000Z, _stop: 2022-12-04T13:00:00.000Z, _time: 2022-12-04T12:02:00.000Z},
    {_value: 0, _start: 2022-12-04T00:00:00.000Z, _stop: 2022-12-04T13:00:00.000Z, _time: 2022-12-04T12:03:00.000Z},
])
  |>  group(columns: ["_start", "_stop"])
  |> integral(unit: 1m, interpolate: "linear")
```
Using linear interpolation, I would expect the above integration of values `0,1,1,0` to return `2`. Instead, the flux query somehow comes up with `-260822.5`

Without interpolation, I get the expected result (which also happens to be `2` in this case)

amykyta3:
After some experimentation, I think I understand what is going on.
Instead of simply interpolating *between* points, It appears to be also *extrapolating* the linear slope from the leading and trailing pair of points all the way to the beginning/end of the window.

This is extremely surprising and seems inappropriate. When dealing with sparse data, I would have expected it to clamp values to 0, or at least assume a zero-slope at the boundaries of the data.

default-student:
This is still real bad and not fixed

Issue #4619, opened 2022-03-31
Handle window boundaries in transformations like integral, difference, derivative 


scbrickley:
Transformations like `integral` don't work correctly inside of `aggregateWindow`. This is because windows are isolated from one another, and functions like integral need to understand the boundary conditions of each window.

As can be seen in [this issue]( https://github.com/influxdata/influxdb/issues/19522), even trivial examples produce unexpected results when applied to time windows.

The core problem here is that the points that land on or near the window boundaries need to exist in both windows in order for integral to compute the correct result.

There are a few potential solutions to this.

1. Modify `window` to optionally enable some cross-window logic

2. Add an `every` argument to `integral`, `derivative`, and `difference` and allow them to do their own windowing.

Some more design work may be needed for this issue.

PR #5468, opened 2024-03-15
[Fix]: Integral calculation 

Nocccer:
# Description
The current integral function with interpolate is not working if points are in between the time range. The first point and the last point are wrong calculated. This is a big problem in the IoT world where devices send data based on events and not in a defined time interval. Even if they send it in an interval, this time is mostly not realtime, so timestamp have some time difference to the actual configured interval.

Lets asume we have two points: 
- Time: 6, Value: 6
- Time: 7, Value: 7

And a time Range:
- Start: 0
- Stop: 8

With the current `integral` and `interpolate: true` this leads to false calculations especially if you use the `timeWeightedAvg` which is based on `integral`. 
Interpolate und Time old

By stretching the first point to the start of the time range and the last point to the stop of the time range, we get a more precized integral:
Interpolate und Time

With `interpolate:false` we get the following:
Step und Time

This PR adds this solution to the integral function.

# Open Discussion:
- The stretching could be a opt-in or opt-out option. Anyway the stretching is needed for `timeWeigthedAvg` to work.

### Checklist


Nocccer:
Related tickets:
- #5349
- #4782
- #4619
- https://github.com/influxdata/influxdb/issues/23929
- https://github.com/influxdata/influxdb/issues/20790

helenosheaa:
Thank you for your contribution! There is a community owned [fork](https://github.com/InfluxCommunity/flux) of Flux that would be a great place to open this PR. More context:

We are still supporting Flux, but are not doing active development on it or adding any new features to it. We still do security patches and will address any critical defects through the maintenance period. Our focus now is on features related to our latest database engine, InfluxDB 3.0, and its associated products (Cloud Serverless, Cloud Dedicated, Clustered, Edge).

jhovell:
@Nocccer Did you ever find or implement a solution? The linked community repo has 8 PRs ever, and none in the past year... doesn't seem super active.

Issue #4900, opened 2022-06-21
Booleans are not comparable


Marwes:
Noticed that `max`,`min` etc do not work with `bool` and `bool` doesn't implement the `Comparable` kind. Comparing bools isn't that common but there isn't any practical reason to omit their inclusion either.

```flux
false < true // ERROR: bool is not comparable
```

ypnos:
max/min aggregations would be really helpful, for example in case of windowing one can decide to let false or true prevail.
let's say the boolean expresses an error state, I woud like to aggregate in a way that any error (true) within the window is preserved.
not having this made the boolean type pretty useless for us so we convert to int and insert as int instead..

PR #5469, opened 2024-03-18
[Enhancement]: Add boolean support for `min()` and `max()`

Nocccer:
# Description

This PR adds boolean support for the following aggregation functions:

- `max()`: Detects the first rising edge or returns the first value, if all false
- `min()`: Detects the first falling edge or returns the first value, if all true

I added this because writing this inside flux itself is really slow. You need to first split all values into numerics and bool (inefficient). Also using the reduce function is always 20% or more slower than the internal method.

### Checklist


Nocccer:
related issues:
- #4900

PR #5483, opened 2024-05-21
feat(array): less leaky string array

mhilton:
Change the behviour of the string array back to the old behaviour where accessing the Value function returns a string that is backed by the arrow memory buffer. This avoids data allocations to memory outside of the memory allocator.

The implementation of array.String has been simplified somewhat as part of the new behaviour.

There are a number of places where correct behviour relies on copies of the data being made. To avoid having to fix all of these in the same PR a temporary ValueCopy function has been added to maintain the old semantics. This is being used everywhere the Value function was previously, except for cases where the value is obviously immediately processed, then discarded.

The cases where the `VisitCopy` function is being used will be address one at a time until we can avoid significant levels of unaccounted memory.

### Checklist

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -2,11 +2,11 @@ package array
 
 import (
 	"strconv"
-	"sync/atomic"
 
 	"github.com/apache/arrow/go/v7/arrow"
 	"github.com/apache/arrow/go/v7/arrow/array"
 	arrowmem "github.com/apache/arrow/go/v7/arrow/memory"
+
 	"github.com/influxdata/flux/codes"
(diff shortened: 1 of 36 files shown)


appletreeisyellow (approved):
It seems a lot of files got touched in this PR, but the main change was adding the interface `binaryArray` and adapting it to different types. Many of the files changes are just refactoring the method name. I left some unblocking comments and questions ✅

I would prefer an other pair of eyes for review!

jeffreyssmith2nd (approved):

PR #5503, opened 2024-08-13
fix: Make fewer copies when building a string array

mhilton:
Change the StringBuilder to only create an arrow builder when it becomes necessary. Whist the StringBuilder has only a single distinct value it will just record how long the array is. Creating an arrow.StringBuilder when a second distinct value is added.

### Checklist

### DIFF
--- a/array/builder.go
+++ b/array/builder.go
@@ -88,8 +88,9 @@ func (b *StringBuilder) AppendBytes(buf []byte) {
 
 }
 
-// Append appends a string to the array being built. The input string
-// will always be copied.
+// Append appends a string to the array being built. A reference
+// to the input string will not be retained by the builder. The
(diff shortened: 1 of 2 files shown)


array/builder.go
```diff
@@ -2,53 +2,99 @@ package array
...
+	if bytes.Equal(b.buffer.Bytes(), buf) {
+		b.len++
+		return
+	}
+	b.makeBuilder(buf)
+
 }
 
 // Append appends a string to the array being built. The input string
 // will always be copied.
```
btasker, line 92: Minor point: the second half of this comment was true prior to this change, but now isn't

btasker (approved):

crepererum (approved):

PR #5511, opened 2024-11-22
feat: upgrade go to 1.22.7

devanbenz:
Upgrades Go to version 1.22.7 to address changes required in influxdb
https://github.com/influxdata/edge/issues/741


davidby-influx (approved):
LGTM

jdstrand:
Fyi, I've not done a golang upgrade for flux, but I think you need to (at least) also adjust https://github.com/influxdata/flux/blob/master/.circleci/config.yml#L113

PR #5524, opened 2025-02-19
build(flux): prepare Flux release for v0.196.0

mhilton:


prep-release.sh
```diff
@@ -24,19 +24,19 @@ make generate
 
 message="build(flux): prepare Flux release for $version"
 
-git commit -am "$message"
+git commit --allow-empty -am "$message"
 git push
 
-if ! command -v hub &> /dev/null
+if ! command -v gh &> /dev/null
```
mhilton, line 30: `gh` has replaced `hub` so I've updated the script, rather than spend seconds creating a PR by hand.

btasker (approved):

PR #5545, opened 2025-06-24
feat(array): update String array type

mhilton:
Update flux's String array type to either be an arrow Binary array or an arrow Dictionary with Binary values. This removes the non-arrow compatible single string variant, instead using a dictionary to provide the low memory version for repeated values. A dictionary provides a more general purpose implementation of the same idea to not keep repeating identical values.

The StringBuilder still swaps to a standard Binary array after a second unique value of a String is observed, but does not do so just because a NULL is added to the array. In the future the heuristic could be changed to provide memory efficient string representations in other contexts.

Moving to a completely arrow-compatible interface makes the String array type much less fragile. It is now possible to use the String array in any context that an arrow Array can be used, and removes the special-case code previously required to split a String array.

### Checklist

### DIFF
--- a/array/builder.go
+++ b/array/builder.go
@@ -5,31 +5,25 @@ import (
 	"sync/atomic"
 	"unsafe"
 
-	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 )
 
 type StringBuilder struct {
-	mem         memory.Allocator
-	len         int
-	cap         int
-	reserveData int
-	buffer      *memory.Buffer
-	builder     *array.BinaryBuilder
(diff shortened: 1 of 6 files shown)


array/builder.go
```diff
@@ -38,54 +32,64 @@ func (b *StringBuilder) Retain() {
...
+		b.values.ReserveData(values.ValueLen(0) * indices.Len())
+		for i := 0; i < indices.Len(); i++ {
+			if indices.IsNull(i) {
+				b.values.AppendNull()
+			} else {
+				b.values.Append(values.Value(int(indices.Value(i))))
+			}
+		}
+		values.Release()
+		indices.Release()
```
jeffreyssmith2nd, line 90: Feels like I'm back in my C++ days 😄

jeffreyssmith2nd (approved):

array/builder.go
mhilton: That is because the go API is modelled off of the C++ one.

PR #5550, opened 2025-08-12
feat(array): Run-end encoded string arrays

mhilton:
It is very common for flux to be using arrays of strings containing a single repeated value. This is often the case when processing InfluxDB tag types. Ad support for run-end encoded string arrays to optimize for this case. Such arrays are quick to access as they consist of a single run, but use less memory than a full string array, or even a dictionary which still needs a full length array to hold the dictionary indexes.

### Checklist

### DIFF
--- a/array/array.go
+++ b/array/array.go
@@ -115,7 +115,8 @@ type Builder interface {
 }
 
 // String holds an array of flux string values. The arrow data must be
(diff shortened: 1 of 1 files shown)


appletreeisyellow (approved):
Nice refactor! It would be helpful to link the [run-end encoded layout](https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout) doc somewhere in the comment for future references.

PR #5583, opened 2025-12-03
fix(stdlib): mark HTTP POST examples not to run

mhilton:
It seems reqres.in has stopped working the way it used to, so the examples in the http/request.post and experimental/http/request.post documentation no longer work. Mark them with no_run to prevent test failures.

### Checklist


mhilton:
> Is there a requirement to follow-up making this work again or changing the example to use a different library?

I don't think so. The example is still fine for documentation, it just doesn't work if executed. It seems a bit dodgy relying on a third-party service for ones tests to pass in the first place.

PR #5638, opened 2026-05-25
ci(semantic): cap GITHUB_TOKEN to contents: read

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at workflow level. No GitHub API writes from the workflow.

Post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/semantic.yml
+++ b/.github/workflows/semantic.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
     types: [opened, reopened, synchronize, edited]
 
+permissions:
+  contents: read
+
 jobs:
   semantic:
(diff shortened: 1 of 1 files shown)


mhilton:
@jdstrand I'd be interested in your thoughts for this one.

jdstrand:
> @jdstrand I'd be interested in your thoughts for this one.

I'm a fan of this sort of thing generally for security in depth. The token will ideally already be scoped to just what it needs, but this helps enforce that. I've used it in a couple of places and encourage we use it more.

mhilton (approved):

mhilton:
Thanks @arpitjain099

